PR #6559, opened 2018-12-08
Fixed ruby warnings

walf443:
Some warnings occured when running with --warnings option

```
$ bundle exec ruby -w  ./exe/rubocop --cache false
./Users/yoshimin/src/github.com/rubocop-hq/rubocop/lib/rubocop/cop/style/percent_literal_delimiters.rb:82: warning: instance variable @preferred_delimiters not initialized
./Users/yoshimin/src/github.com/rubocop-hq/rubocop/lib/rubocop/cop/layout/empty_lines_around_access_modifier.rb:109: warning: instance variable @block_line not initialized
```

It seems to be better to run test with --warnings mode.

-----------------

* [ ] Added tests.


lib/rubocop/cop/layout/empty_lines_around_access_modifier.rb
```diff
@@ -106,7 +106,7 @@ def class_def?(line)
         end
 
         def block_start?(line)
-          return false unless @block_line
+          return false unless defined? @block_line
```
koic: Ah, this is my oversight at #6361. Thanks for catching up.

I think that it is possible to clarify the code intention by using `instance_variable_defined?` method than `define?` method.

```ruby
return false unless instance_variable_defined?(:@block_line)
```

However, using the `initialize` method may be simplest and easier to understand. I think that it is better to use the `initialize` method approach.

```diff
         MSG_BEFORE_AND_AFTER = 'Keep a blank line before and after ' \
                                '`%<modifier>s`.'.freeze

+        def initialize(config = nil, options = nil)
+          super
+
+          @block_line = nil
+        end
+
         def on_class(node)
```

walf443: @koic I fixed it.

PR #6562, opened 2018-12-08
Fix a false positive parentheses omission in safe navigation calls

gsamokovarov:
The safe navigation produces false positives for the `omit_parentheses`
style of the `Style/MethodCallWithArgsParentheses` cop.

Here is an example:

```
› bundle exec rubocop app/services/request_accountant_change.rb
Inspecting 1 file
C

Offenses:

app/services/request_accountant_change.rb:7:34: C: Style/MethodCallWithArgsParentheses: Omit parentheses for method calls with arguments.
    @accountant_id = User.find_by(email: email)&.id || SecondaryEmail.find_by(address: email)&.user_id
                                 ^^^^^^^^^^^^^^
app/services/request_accountant_change.rb:7:78: C: Style/MethodCallWithArgsParentheses: Omit parentheses for method calls with arguments.
    @accountant_id = User.find_by(email: email)&.id || SecondaryEmail.find_by(address: email)&.user_id
                                                                             ^^^^^^^^^^^^^^^^
```

Handling the safe navigation operator fixes those bad reports.

### DIFF
--- a/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
+++ b/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
@@ -248,7 +248,7 @@ module RuboCop
 
         def call_with_ambiguous_arguments?(node)
           call_with_braced_block?(node) ||
-            call_as_argument?(node) ||
+            call_as_argument_or_chain?(node) ||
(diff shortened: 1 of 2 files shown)


bbatsov:
👍

PR #6568, opened 2018-12-11
New cop Lint/SafeNavigationWithEmpty

rspeicher:
This cop checks to make sure safe navigation isn't used with `empty?` in
a conditional.

While the safe navigation operator is generally a good idea, when
checking `foo&.empty?` in a conditional, `foo` being `nil` will actually
do the opposite of what the author intends.

This was suggested by my GitLab colleague @grzesiek at https://gitlab.com/gitlab-org/gitlab-ce/issues/54853.

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/safe_navigation_with_empty.rb
```diff
@@ -0,0 +1,38 @@
...
+      # a conditional.
+      #
+      # While the safe navigation operator is generally a good idea, when
+      # checking `foo&.empty?` in a conditional, `foo` being `nil` will actually
+      # do the opposite of what the author intends.
+      #
+      # @example
+      #   # bad
+      #   return if foo&.empty?
+      #   return unless foo&.empty?
```
grzesiek: What about `blank?`. `foo.blank?` does not need safe navigation operator, so `return if foo&.blank?` is not only incorrect because the problem described here, but it is also completely unnecessary to use safe navigation operator because `nil` also responds to `blank?`. This is Rails-specific, though, do we have a different repository with Rails-specific cops? :thinking:

manual/cops_lint.md
```diff
@@ -1728,6 +1728,31 @@ Name | Default value | Configurable values
 --- | --- | ---
 Whitelist | `present?`, `blank?`, `presence`, `try`, `try!` | Array
 
+## Lint/SafeNavigationWithEmpty
+
+Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
+--- | --- | --- | --- | ---
+Enabled | Yes | No | 0.62 | -
```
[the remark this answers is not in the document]
grzesiek: @rspeicher Actually I think that what we are suggest in `good` basically behaves exactly the same as the `bad` behavior:

```ruby
def test_unsafe_empty(value)
  return if value&.empty?
 
  'xxx'
end  

pry(main)> test_unsafe_empty([])
=> nil
pry(main)> test_unsafe_empty(nil)
=> "xxx"

def test_safe_empty(value)
  return if value && value.empty?
 
  'yyy'
end  

pry(main)> test_safe_empty([])
=> nil
pry(main)> test_safe_empty(nil)
=> "yyy"
```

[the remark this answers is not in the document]
grzesiek: Cross that, it seems this is not a good behavior too :arrow_right: https://github.com/rubocop-hq/rubocop/pull/6568#discussion_r240919953

lib/rubocop/cop/lint/safe_navigation_with_empty.rb
Drenmi: > This is Rails-specific, though, do we have a different repository with Rails-specific cops? 🤔

We have extracted `rubocop-rails`, but the Rails specific cops haven't been moved there yet. But you are correct that it shouldn't go in this cop. 🙂

lib/rubocop/cop/lint/safe_navigation_with_empty.rb
```diff
@@ -0,0 +1,38 @@
...
+      #
+      #   # good
+      #   return if foo && foo.empty?
+      #   return unless foo && foo.empty?
+      #
+      class SafeNavigationWithEmpty < Cop
+        MSG = 'Avoid calling `empty?` with the safe navigation operator ' \
```
Drenmi: I'm trying to standardise a leading pound sign when referring to method names in messages, i.e. `#empty?`. 🙂

lib/rubocop/cop/lint/safe_navigation_with_empty.rb
```diff
@@ -0,0 +1,38 @@
...
+          'in conditionals.'.freeze
+
+        def_node_matcher :safe_navigation_empty_in_conditional?, <<-PATTERN
```
Drenmi: Do you think this check would be helpful even if it's not used directly in a conditional? It could be in a named predicate method, e.g.:

```ruby
def foo?
  bar&.empty?
end

baz if foo?
```

bbatsov:
@rspeicher PIng :-)

rspeicher:
@bbatsov Yep sorry, work got in the way. 😃

lib/rubocop/cop/lint/safe_navigation_with_empty.rb
rrosenblum: Even once we are done extracting `rubocop-rails`, that project won't be able to change the behavior of this cop without redefining functionality. I think it could be reasonable to have a configuration setting to optionally use `blank?`.

manual/cops_lint.md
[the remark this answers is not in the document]
rrosenblum: Adding auto-correction should be pretty straight forward for this cop.

Drenmi:
> Do you have any time to take this over? I'm starting vacation this week and don't know when I'll be able to get back to it.

@rspeicher We have a two day public holiday in Singapore starting tomorrow. I think I can find time to make the required changes to get this merged. 🙂

mikegee:
I think this cop is valuable without configurability or autocorrection. Those could be followup PRs.

bbatsov:
I'll merge this as is with the hope that someone will tweak it later.

mvz:
I'm confused by the documentation of this cop, since in my case `foo&.empty?` does exactly what I intend.

(To be specific, I have the equivalent of `foo = nil if foo&.empty?`. If `foo` is already `nil`, it won't be set to `nil` again and that's intentional.)

grzesiek:
@mvz I think that using `foo = nil if foo && foo.empty?` is actually more readable and less confusing, so perhaps this cop still makes sense in this case?

mvz:
@grzesiek I agree and it actually made me reconsider the coda and change it to `foo = nil if foo.blank?`. It's the documentation that I find problematic. I'll see if I can make a pull request to improve it.

dbackeus:
I came across a violation of this cop today and am utterly confused. The documentation states:

> when checking foo&.empty? in a conditional, foo being nil will actually do the opposite of what the author intends.

How would it do the opposite of what I intend? What are you assuming that I'm intending? And how would changing to `foo && foo.empty?` stop doing the opposite of what I intended with `foo&.empty?` when they logically execute exactly the same way?

I can understand the cop in rubocop-rails related to `.blank?` since that exists on `NilClass` and doesn't make sense to use safe navigation for. But `empty?` doesn't exist on NilClass. So if I eg. want to verify that a variable is defined as well as being empty, `foo&.empty?` is the most consise way of expressing that.

rspeicher:
> So if I eg. want to verify that variable `foo` is defined as well as empty, `foo&.empty?` is the most consise way of expressing that. It's also consistent with how safe navigation is prefered to `&&` in the style guide in general.

@dbackeus In the case where you want to verify that something is defined _and_ empty, `foo&.empty?` won't do it:

1. If it's truly undefined, you should get a `NameError`.
2. If it's `nil`, the safe navigation will return `nil`, which is false-y.
3. If it's defined and empty (e.g., `[]`), it will return `true`.

They're mutually exclusive. For your specific example I think you'd need `defined?(foo) && foo.empty?`.

Apologies for the "what the author intends" language, that's presumptuous, but I think the intent of the cop is still valid.

dbackeus:
To be clear, I didn't intend to raise the "truly undefined" case, that would indeed return a NameError and should not be a surprise to anyone.

With regards to 2 and 3. Yes that's the expected outcome. And exactly what should be intended with `foo&.empty?`.

So do I understand you correctly in that you are assuming that Ruby developers writing `foo&.empty?` would actually intend a truthy outcome if foo was nil?

If so shouldn't the "fix" be `foo.nil? || foo.empty?` rather than `foo && foo.empty?` (which at the end of the day is just a more verbose and inconsistent way of making the same statement).

rspeicher:
@dbackeus I can't remember exactly but I suspect the specific case that lead us to write this cop was when using `empty?` in a guard clause, but the value passed to the method wasn't guaranteed to be non-nil:

```ruby
def some_method(val)
  return if val&.empty?

  delete_everything!
end
```

In this (contrived) case, if the method receives `nil` instead of `[]`, we're not returning early as we probably intended. Arguably that's not very good defensive coding, but this is a subtle bug that may be easily missed, and the cop protects against it. In this case we probably actually want to take advantage of ActiveSupport and do something like `return unless val.blank?`.

PR #6570, opened 2018-12-12
False positive for parentheses removal in splatted calls

gsamokovarov:
Ahem, this is yet-another-false-positive for the `omit_parentheses`
style of the `Style/MethodCallWithArgsParentheses` cop.

If you are splatting the result of a call, you may end up with the
following:

```ruby
app/controllers/foo/bar_controller.rb:4:68: C: Style/MethodCallWithArgsParentheses: Omit parentheses for method calls with arguments.
      render json: FieldSearch.with(search_params).map(&method(:serialize))
                                                              ^^^^^^^^^^^^
```


Drenmi:
Thanks for your work on this cop @gsamokovarov! 🙏 I've been wanting to enable this cop in my projects for the longest time, but the number of false positives has made it not worth it. Excited to see us getting this up to an acceptable standard. 🙂

spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
```diff
@@ -414,6 +414,14 @@ def foo
       RUBY
     end
 
+    it 'accepts parens in splat calls' do
+      expect_no_offenses(<<-RUBY)
```
Drenmi: Can we split these up into separate examples, please? 🙇

gsamokovarov: Sure!

gsamokovarov:
@Drenmi I think we're getting to a better state! 😅 I think I also addressed your comments.

PR #6572, opened 2018-12-12
[Fix #6500] Add offense to deprecated to_time_in_current_zone

### LINKED ISSUE #6500: to_time_in_current_zone does not exist anymore
When checking `RuboCop::Cop::Rails::Date` there is suggested an example with `to_time_in_current_zone` https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/Date which does not exist anymore in Rails. https://apidock.com/rails/Date/to_time_in_current_zone

Instead `Date#in_time_zone` should be used.

nadiyaka:
- Remove suggested example with `to_time_in_current_zone ` https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/Date

- Add a deprecation warning offense when using this method.


lib/rubocop/cop/rails/date.rb
```diff
@@ -68,12 +74,25 @@ def on_send(node)
...
+          check_deprecated_method(node)
+
           add_offense(node, location: :selector,
                             message: format(MSG_SEND, method: node.method_name))
         end
 
         private
 
+        def check_deprecated_method(node)
+          DEPRECATED_METHODS.each do |info|
```
Drenmi: Ruby 2.0 introduced block keyword arguments, so we can do assignment there:

```ruby
DEPRECATED_METHODS.each do |relevant:, deprecated:|
  # relevant and deprecated are now available as local variables in this block
  ...
end
```

nadiyaka: Thanks! I've just updated the code.

PR #6577, opened 2018-12-13
Rails/Blank skips offense when defining the `blank?` method

jonatas:
We found a case that we need to disable the `Rails/Blank` cop when we're creating the `blank?` method.

So, it adds an offense when we have:

```ruby
def blank?
  !present?
end
```

But it should accept because it's the method definition.

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/rails/blank_spec.rb
+++ b/spec/rubocop/cop/rails/blank_spec.rb
@@ -129,6 +129,10 @@ RSpec.describe RuboCop::Cop::Rails::Blank, :config do
     it_behaves_like 'offense', '!present?',
                     'blank?',
                     'Use `blank?` instead of `!present?`.'
+
+    it 'accepts !present? if its in the body of a `blank?` method' do
+      expect_no_offenses('def blank?; !present? end')
+    end
   end
 
   context 'UnlessPresent set to true' do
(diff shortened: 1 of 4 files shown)


mikegee:
The Rails cops are supposed to be enforced on Rails apps, where the `present?` method is defined on every object. Are you enforcing Rails cops on a non-Rails application? I feel like I'm missing something here.

jonatas:
I use this cop enabled for the entire rails project using the default configuration.

I have some methods that are supporting my cucumber features and I define a `blank?` method there. 

Should we set some default configuration to ignore non-rails folders?

mikegee:
> Should we set some default configuration to ignore non-rails folders?

Perhaps not, I was just a bit confused. This PR looks good to me.

spec/rubocop/cop/rails/blank_spec.rb
```diff
@@ -129,6 +129,10 @@
     it_behaves_like 'offense', '!present?',
                     'blank?',
                     'Use `blank?` instead of `!present?`.'
+
+    it 'accepts define blank? method' do
```
bbatsov: accepts `!present?` if its in the body of a blank? method definition.

bbatsov:
Thanks!

PR #6580, opened 2018-12-14
Add new `Rails/LinkToBlank` cop

Intrepidd:
This cop checks for calls to `link_to` that contain a `target: '_blank'` but no `rel: 'noopener'`. This can be a security risk as the loaded page will have control over the previous page and could change its location for phishing purposes.

Demo resource for reference : https://mathiasbynens.github.io/rel-noopener/


lib/rubocop/cop/rails/link_to_blank.rb
```diff
@@ -0,0 +1,51 @@
...
+        PATTERN
+
+        def_node_matcher :includes_noopener?, <<-PATTERN
+          (pair {(sym :rel) (str "rel")} (str #contains_noopener?))
+        PATTERN
+
+        def on_send(node)
+          return unless node.method?(:link_to)
+
+          option_nodes = [node.children.last, node.children[3]].compact
```
Intrepidd: Not sure about the `[3]` here, any better idea ?

Drenmi: I would probably go with something like:

```ruby
node.arguments.each_child_node(:hash)
```

lib/rubocop/cop/rails/link_to_blank.rb
```diff
@@ -0,0 +1,51 @@
...
+
+          option_nodes.map(&:children).each do |options|
```
Drenmi: You can go with `#each_pair` here, instead of `map(&:children)`.

lib/rubocop/cop/rails/link_to_blank.rb
Intrepidd: Thanks for the tip ! I managed to make it work with `node.each_child_node(:hash)`

lib/rubocop/cop/rails/link_to_blank.rb
Intrepidd: Unfortunately at this point `option_nodes` is an `Enumerator` of Hash nodes

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Change log
 
 ## master (unreleased)
+* New cop `Rails/LinkToBlank` checks for `link_to` calls with `target: '_blank'` and no `rel: 'noopener'`. ([@Intrepidd][])
```
bbatsov: You have to put this under "New features".

lib/rubocop/cop/rails/link_to_blank.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Rails
+      # This cop checks for calls to `link_to` that contain a
+      # `target: '_blank'` but no `rel: 'noopener'`. This can be a security
+      #  risk as the loaded page will have control over the previous page
```
bbatsov: This line seems misaligned.

config/default.yml
```diff
@@ -2399,6 +2399,11 @@ Rails/LexicallyScopedActionFilter:
   Include:
     - app/controllers/**/*.rb
 
+Rails/LinkToBlank:
+  Description: 'Checks that `link_to` with a `target: "_blank"` have a `rel: "noopener"` option passed to them.'
```
bbatsov: Might be a good idea to add some `Reference:` link here as well.

Intrepidd:
Thanks @bbatsov ! I applied your comments an rebased

bbatsov:
Thanks!

koic:
FYI, although it has not been merged yet, there is a PR opened in rails/rails that adds `rel="noopener"` default to `link_to` when particular conditions are met 😉 
https://github.com/rails/rails/pull/28035

PR #6584, opened 2018-12-18
[Fix #6150] Add runtime option to ignore cop disable comments

### LINKED ISSUE #6150: Add support to enforce disabled cops to be executed
According to the [documentation](http://docs.rubocop.org/en/latest/configuration/#disabling-cops-within-source-code), you can disable a cop by adding a comment to your source code:

```ruby
# rubocop:disable Metrics/LineLength, Style/StringLiterals
def foo
  bar
end
# rubocop:enable Metrics/LineLength, Style/StringLiterals
```
This is a useful feature indeed. Problem is that it enables people to ignore some easy to fix issues instead of actually fixing them.

It would be nice to have a way to enforce cops to be always evaluated, even if you disabled them within your source code.

The idea is to have a new configuration where you could disallow disabling cops using comments:

- Add new config `AllowSourceCodeDisabling` -  default is `true`
- Change cops to understand this new configuration
- Add new cop to check if you disabled a cop when it was not allowed to - or extend `Lint/UnneededCopDisableDirective` to do that

I would help with PRs if it's something that would be useful for other people.

roooodcastro:
This PR adds the ability to run and enforce cops that were disabled using code comments such as:

```ruby
# rubocop:disable Metrics/LineLength, Style/StringLiterals
def foo
  bar
end
# rubocop:enable Metrics/LineLength, Style/StringLiterals
```

The issue #6150 asked for this functionality to be added, but in a slightly different way. Instead of a runtime option, it would add a configuration option to control this. In my opinion this is not the best solution, as we sometimes do need to ignore a few offences, and flat out blocking this would mean that no one would be able to actually use this feature.

Instead, this adds a runtime option `--ignore-disable-comments`, or `-i`, that will run all cops even if they're disabled, but as a one-off execution.

What I've done to accomplish this was basically to add the option to the list and check for it when a cop checks if the offending line of code is enabled or not. Now it will always say it's enabled if this option is on.

-----------------

* [x] Added tests.


pocke:
Thank you for your pull request! Sorry for the delayed review.
I really like this solution, which is an option instead of configuration. I think it is more useful than configutaion.



Personaly I think `-i` short option is unnecessary, `--ignore-disable-comments` is enough.
Already RuboCop has `--ginore-parent-exclusion` option. User cannot distinguish between these options if it has `-i` short option.

roooodcastro:
Ok, thanks for the review,  I'll remove the `-i` shortcut then. Is it good to go otherwise?

roooodcastro:
I've removed the `-i` shortcut, are we good to merge?

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 * [#6100](https://github.com/rubocop-hq/rubocop/issues/6100): Fix a false positive in `Naming/ConstantName` cop when rhs is a conditional expression. ([@tatsuyafw][])
 * [#6526](https://github.com/rubocop-hq/rubocop/issues/6526): Fix a wrong line highlight in `Lint/ShadowedException` cop. ([@tatsuyafw][])
 
+### New features
```
pocke: [nitpick] Can you move `New fetures` section above `Bug fixes` section?

roooodcastro: Done

pocke:
I have a small comment about CHANGELOG, but I think it is ready to merge excep the comment.

I'll merge it after a while unless anyone disagrees with this pull request. Thank you.

roooodcastro:
Hey, I rebased with master and got one spec failing, that I believe was introduced by #6650.

PR #6587, opened 2018-12-19
[Fix #6349] MemoizedInstanceVariableName allows underscored method name

### LINKED ISSUE #6349: MemoizedInstanceVariableName Still complaining about leading underscores
## Expected behavior

Not to report the cop (not present pre 0.58), fixed in 0.58.0

## Actual behavior

rubocop 0.57.2 - 1 file inspected 0 failures
rubocop 0.58.2 - 1 file inspected 1 failure
rubocop 0.59.2 - 1 file inspected 1 failure

```
# This is fine
def foo
  @foo ||= :foo
end

# This isn't fine
def _foo
  @_foo ||= :foo
end
```

```
features/support/cop_test.rb:6:3: C: Naming/MemoizedInstanceVariableName: Memoized variable @_foo does not match method name _foo. Use @_foo instead.
```
## RuboCop version('s)

```
0.57.2 (using Parser 2.5.1.2, running on ruby 2.2.10 x86_64-linux)
0.58.2 (using Parser 2.5.1.2, running on ruby 2.2.10 x86_64-linux)
0.59.2 (using Parser 2.5.1.2, running on ruby 2.2.10 x86_64-linux)
```

matthewrudy:
We should be allowed to write code like this:

``` ruby
def _foo
  @_foo ||= :foo
end
```
namely if the method name has a leading `_`, the memoized variable name should have it too.
Currently it gives  an error, saying the variable should be named `@foo`.

This should fix #6349

### DIFF
--- a/lib/rubocop/cop/naming/memoized_instance_variable_name.rb
+++ b/lib/rubocop/cop/naming/memoized_instance_variable_name.rb
@@ -132,9 +132,7 @@ module RuboCop
           variable = ivar_assign.children.first
           variable_name = variable.to_s.sub('@', '')
 
-          return false unless valid_leading_underscore?(variable_name)
-
-          variable_name.sub(/\A_/, '') == method_name.sub(/\A_/, '')
+          valid_variable_name?(method_name, variable_name)
         end
 
         def message(variable)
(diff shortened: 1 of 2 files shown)


matthewrudy:
I will address the failing spec after lunch.

bbatsov:
Hmm, I thought we have fixed this already. Seems to me that's the 3rd time I see some issue about fixing vars with underscores. 😆

matthewrudy:
@bbatsov maybe so.

Was still a problem when I did this
(Which was prompted by an earlier issue with the same thing that I worked around)

If it's fixed on master, then fairdos.

Appreciate you taking a look.

bbatsov:
The changes look good, but your branch has to be rebased on top of the current `master` branch due to merge conflicts. And add an entry in the changelog.

PR #6590, opened 2018-12-20
Fix bug extend_self with private methods

### LINKED ISSUE #6370: Style/ModuleFunction autocorrection breaks code
## Expected behavior

The Style/ModuleFunction autocorrection should either not happen or not break code.

## Actual behavior

This

    module X
      extend self
      
      def foo; bar; end
  
      private

      def bar; end
    end

gets autocorrected into 

    module X
      module_function
      
      def foo; bar; end
  
      private

      def bar; end
    end

which - in contrast to the previous definition - does not define `X.bar`

Run rubocop 0.59.2 with this code.

## RuboCop version

0.59.2

Ruffeng:
Fix #6370. Cop extend self to turn into module_function should
not be invoked in case there are private methods. Because of that, the
flow of the code gets broken.

My idea: Since the switch from `extend self` into `module_function` forces you to change the code in the scenario you have private methods, I just added a small validation to check if there's any private declaration before. In case there are no private methods, proceed with the offense. 

It works as well for the correction ( I am not sure if I need tests there since the warning is not written at all) 

This is a very first draft PR. All refactors or wrong code is more than welcome 😄 

-----------------

* [x] Added tests.


Ruffeng:
@Drenmi I hope it looks good now 😄

bbatsov:
Thanks for working on this! I really appreciate it! 🙇

PR #6594, opened 2018-12-21
Allow Rails html_safe for non-interpolated String literal receiver

amatsuda:
Current Rails/OutputSafety cop seems to be always warning any kinds of `html_safe` method call as possibly insecure code.

But in fact, `html_safe` is not such a dangerous thing. It's a part of Rails core API, and it's not something that is discouraged to use that way. It can just be a security risk only when the string could contain random user input values.

Actually, the document of this cop indicates the following code as a "good" code, so I guess this is just an implementation bug. https://github.com/rubocop-hq/rubocop/blob/a5abbde/manual/cops_rails.md#examples-27

```
out = "<h1>trusted content</h1>".html_safe
```

And so attached is a patch that allows `html_safe` call on non-interpolated string literals, just as documented.

-----------------

* [x] Added tests.


bbatsov:
Thanks for working on this, @amatsuda! I really appreciate it! 🙇

PR #6595, opened 2018-12-21
Fix a false positive parentheses omission for slash regexps

gsamokovarov:
Allow parentheses in calls with slash regexp literals as first
argument for the `Style/MethodCallWithArgsParentheses` cop's
`omit_parentheses` style.

Ruby issues a warning for those and if we want to omit the
parens in this case and we're also fighting with other cops.

```ruby
         method_call(/foo/)
                    ^^^^^^^ Omit parentheses for method calls with arguments.
```


bbatsov:
This has to be mentioned in the changelog.

PR #6599, opened 2018-12-23
Move CHANGELOG entry to correct section

garettarrowood:
A recent merge added an `CHANGELOG` entry to an earlier release. This just moves that entry to the `unreleased` section.


bbatsov:
Thanks for spotting and fixing this!

PR #6600, opened 2018-12-23
[Fix #3904] Build config & sync cop descriptions

### LINKED ISSUE #3904: Deduplicate cop documentation and cop descriptions like rubocop-rspec
I refactored part of our configuration a while back for rubocop-rspec and I'm pretty happy with it. See https://github.com/backus/rubocop-rspec/pull/174. Basically, all of the cop documentation had something like this

```ruby
module RuboCop
  module Cop
    module RSpec
      # Check for test expectations that can use `be` instead of `eql`
      #
      # (longer description snipped)
      #
      # (examples section snipped)
      #
      # (more long form description snipped)
      #
      class BeEql < Cop
      end
    end
  end
end
```

and the configuration description is subtly different for (IMO) no good reason:

```yaml
RSpec/BeEql:
  Description: 'Check for expectations where `be(...)` can be used instead of `eql(...)`'
  Enabled: true
```

We now have a build command you can execute (`bin/build_config`) that will parse the YARD documentation and write the leading docstring to the configuration. This is run in CI too so these don't fall out of sync. I think this is a nice change because it removes a lot of duplicated work.

I would be happy to eventually help upstream this functionality to rubocop. It would be harder since your configuration files are heavily commented and spread across multiple config files. It still would be worth it though I think.

garettarrowood:
This PR addresses #3904. It deduplicates `Cop` descriptions by loading the config and replacing its `Description`s with the ones held in `yard`. That makes the source code the single source of truth for descriptions. This was heavily inspired by @backus 's implementation in `rubocop-rspec`. (Thank you sir!)

There are a lot of consequences of loading and dumping `YAML`. Here are some of the changes you will find in `config/default.yml` that are noteworthy but (imo) seem ok.

* Lists are not indented - ref: https://stackoverflow.com/questions/17014460/yaml-indentation-for-array-in-hash
* single quotes are generally replaced with double quotes. This appears to be when the string contains special characters like `*` and `#`. They are not replaced in `VersionAdded` and `VersionChanged` settings.
* Settings such as `CacheRootDirectory: ~` are replaced with `CacheRootDirectory: `. I don't like this, but it's still valid yaml. It seems to boil down to aesthetic, though.
* [Descriptions include line feeds](https://github.com/rubocop-hq/rubocop/compare/master...garettarrowood:deduplicate_cop_documentation?expand=1#diff-db360415c82dc1c3dd5a5c2872d0cd04L31). This doesn't seem like a big deal because those line feeds are following the 80 character `Metrics/LineLength` setting enforced in the repo already. So comments will appear uniform in the file.

But. There is a deal breaking change here. All comments are stripped from the config. I looked around for a solution and only found ones involving adding comments in one line at a time. While searching for alternative yaml parsers besides `YAML`/`Pysch`,  I found [this blog post which suggests the only comment preserving YAML parser is ruamel, which is for Python](https://kev.inburke.com/kevin/more-comment-preserving-configuration-parsers/).

Possible solutions:

1. Move the comments into the config. Add keys such as `AllowSymlinksInCacheRootDirectoryDescription` and give it the associated comment as a value to save.
1. Don't mix these concerns. Put documentation in something else, designed for it, instead of a configuration file.
1. Explore trying to make all of these updates using other Ruby libraries such as `File`, instead of loading the config itself.
1. Something I haven't thought of. Have an idea?

Last note: I believe it is to our benefit to load config, make a change, and the dump it back out. After syncing descriptions, we could explore syncing any/all other values. Procedurally updating our configuration feels like a step in the right direction.

Thoughts and ideas?

-----------------

* [x] Added tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -33,9 +33,12 @@ RuboCop::RakeTask.new(:internal_investigation).tap do |task|
 end
 
 task default: %i[
-  documentation_syntax_check generate_cops_documentation
+  build_config
+  documentation_syntax_check
+  generate_cops_documentation
   spec ascii_spec
   internal_investigation
+  confirm_config
 ]
 
 require 'yard'
@@ -118,4 +121,24 @@ task documentation_syntax_check: :yard_for_generate_documentation do
     end
   end
   abort unless ok
+end
+
+desc 'Build config/default.yml'
+task :build_config do
+  sh('bin/build_config')
+end
+
+desc 'Confirm config/default.yml is up to date'
+task confirm_config: :build_config do
+  # Do not print diff and yield whether exit code was zero
+  sh('git diff --quiet config') do |outcome, _|
+    return if outcome
+
+    # Output diff before raising error
+    sh('GIT_PAGER=cat git diff config')
+
+    warn 'The config/default.yml is out of sync. ' \
+      'Run `rake build_config` and commit the results.'
+    exit!
+  end
 end
--- a/lib/rubocop/config_formatter.rb
+++ b/lib/rubocop/config_formatter.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+module RuboCop
+  # Builds YAML config file from two hashes
+  class ConfigFormatter
+    DEPARTMENTS = /^(#{Cop::Cop.registry.departments.join('|')})/x.freeze
+    STANDALONE_DEPARTMENT_EXCLUSIONS = ['Rails'].freeze
+
+    def initialize(config, descriptions)
+      @config = config
+      @descriptions = descriptions
+    end
+
+    def dump
+      YAML.dump(unified_config).gsub(DEPARTMENTS, "\n\\1")
+    end
+
+    private
+
+    def unified_config
+      cops.each_with_object(config.dup) do |cop, unified|
+        unified[cop] = config.fetch(cop).merge(descriptions.fetch(cop))
+      end
+    end
+
+    def cops
+      (descriptions.keys | config.keys).grep(DEPARTMENTS) -
+        STANDALONE_DEPARTMENT_EXCLUSIONS
+    end
+
+    attr_reader :config, :descriptions
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -123,6 +123,7 @@
 ### Changes
 
 * [#595](https://github.com/rubocop-hq/rubocop/issues/595): Exclude files ignored by `git`. ([@AlexWayfer][])
+* [#3904](https://github.com/rubocop-hq/rubocop/issues/3904): Deduplicate descriptions by copying cop descriptions into config. ([@garettarrowood][])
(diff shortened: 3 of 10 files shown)


Rakefile
```diff
@@ -119,3 +123,17 @@ task documentation_syntax_check: :yard_for_generate_documentation do
...
 end
+
+desc 'Build config/default.yml'
+task :build_config do
+  sh('bin/build_config')
+end
+
+desc 'Confirm config/default.yml is up to date'
+task confirm_config: :build_config do
+  _, status = Open3.capture2('git diff --exit-code config/default.yml')
```
Darhazer: the documentation task uses
```
sh('git diff --quiet manual') do |outcome, _|
      return if outcome
```
so no need of Open3

lib/rubocop/config_formatter.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+module RuboCop
+  # Builds YAML config file from two hashes
+  class ConfigFormatter
+    NAMESPACES =
```
Darhazer: I wonder whether this can be generated from `Cop.registry.departments`

lib/rubocop/config_formatter.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def dump
+      YAML.dump(unified_config).gsub(NAMESPACES, "\n\\1")
+    end
+
+    private
+
+    def unified_config
+      cops.each_with_object(config.dup) do |cop, unified|
+        next if cop == 'Rails'
```
Darhazer: a comment why Rails was excluded would be nice

garettarrowood: Addressed this by refactoring in 8fb2d70.

lib/rubocop/config_formatter.rb
garettarrowood: Thanks! Done.

garettarrowood:
@Darhazer - Thanks for the comments! Any opinion on an approach to preserving the config comments? -- I'm leaning towards moving them directly into the config (option 1), but would love a thumbs up from a collaborator/owner before making such an update. -- @bbatsov , what do you think?

bbatsov:
> There are a lot of consequences of loading and dumping YAML. Here are some of the changes you will find in config/default.yml that are noteworthy but (imo) seem ok.

Why do you even load and dump it? I'd just merge the data from the config with whatever we get from the source file without actually regenerating the file itself.

bbatsov:
@garettarrowood What's the state here?

garettarrowood:
@bbatsov -
I'm finding it difficult to update the yaml config file using Ruby File classes and not YAML libraries (because they all wipe away the code comments when the config is regenerated). It's possible to find the line in the config file to insert the cop's `Description`, and then insert it. But I have not yet found how to then detect the formatting/how many lines the outdated description is, and then remove it.

Perhaps there is a technique to go about this that has escaped me so far. I'm open to suggestions :).

garettarrowood:
Perhaps I could cycle through the file twice. The first time removing all the Descriptions. Then cycle through it again and add the ones from the cop files themselves.

It seems inefficient, and a lot of extra file operations than would be necessary. But I can give it a go again with that idea.

PR #6606, opened 2018-12-27
[Fix #6603] - Testing RescueEnsureAlignment

### LINKED ISSUE #6603: Layout/RescueEnsureAlignment issue with nested begin
Looks similar to #6433 

## Expected behavior

No error emitted

## Actual behavior

Cop expects `rescue` to be aligned with `begin`

```ruby
def some_method
  @ivar ||= begin
     1
  rescue
    2
  end
end
```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.61.1 (using Parser 2.5.1.2, running on ruby 2.5.1 x86_64-darwin17)
```

jaredbeck:
Add a test to demonstrate the correct styles of a begin-rescue as the
RHS of an assignment.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/layout/rescue_ensure_alignment_spec.rb
+++ b/spec/rubocop/cop/layout/rescue_ensure_alignment_spec.rb
@@ -36,6 +36,40 @@ RSpec.describe RuboCop::Cop::Layout::RescueEnsureAlignment, :config do
         end
       RUBY
     end
+
+    context 'as RHS of assignment' do
+      it 'accepts multi-line, aligned' do
+        expect_no_offenses(<<-RUBY.strip_indent)
+          x ||= begin
+                  1
+                rescue
+                  2
+                end
+        RUBY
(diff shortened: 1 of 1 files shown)


spec/rubocop/cop/layout/rescue_ensure_alignment_spec.rb
```diff
@@ -36,6 +36,30 @@
         end
       RUBY
     end
+
+    context 'as RHS of assignment' do
+      it 'accepts multi-line assignment' do
+        expect_no_offenses(<<-RUBY.strip_indent)
```
bbatsov: I think we also need a similar example when the `begin` is on the same line as the assignment op.

jaredbeck:
> I think we also need a similar example when the begin is on the same line as the assignment op.

Example added.

bbatsov:
Thanks!

PR #6607, opened 2018-12-27
Improve CLI usage message for --stdin option

jaredbeck:
When I encountered this message for the first time I thought "what
the heck? Why do you need a path when I said --stdin". Of course,
we do need a path, but the reason was not obvious to me at first.
Maybe this improved message will help someone else.

-----------------

* [ ] Added tests.


lib/rubocop/options.rb
```diff
@@ -9,6 +9,9 @@ class OptionArgumentError < StandardError; end
 
   # This class handles command line options.
   class Options
+    E_STDIN_NO_PATH = '-s/--stdin requires exactly one path, so that rubocop ' \
```
bbatsov: I wonder if longer messages shouldn't link to the documentation instead.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New features
 
 * [#6580](https://github.com/rubocop-hq/rubocop/pull/6580): New cop `Rails/LinkToBlank` checks for `link_to` calls with `target: '_blank'` and no `rel: 'noopener'`. ([@Intrepidd][])
+* [#6607](https://github.com/rubocop-hq/rubocop/pull/6607): Improve CLI usage message for --stdin option. ([@jaredbeck][])
```
bbatsov: This should be under "Changes".

lib/rubocop/options.rb
bitst0rm: I still do not get it, what the hell it means. `'--stdin' '-' `or `'--stdin' 'xxx.rb'` product the same error. this improved msg does not help much:
```
['ruby', '-S', 'rubocop', '--format', 'emacs', '--force-exclusion', '--stdin', '/Users/admin/Downloads/rubocop-master/lib/rubocop/file_finder.rb']
SublimeLinter: WARNING: rubocop output:
-s/--stdin requires exactly one path.
```

jaredbeck: > I wonder if longer messages shouldn't link to the documentation instead.

I think, for a few dozen words, it is convenient to have it directly in the error message. If it were a few hundred words, I'd move it to the docs. That's just my opinion. 

Pro: docs can have formatting and code examples
Con: links can break in the future

> I still do not get it, what the hell it means. '--stdin' '-' or '--stdin' 'xxx.rb' product the same error. this improved msg does not help much:

Well, it will be compared against `Include` rules, so I think it's supposed to be a path, relative to the root of the project, right? Maybe we should also say that? "relative to the root of the project"

jaredbeck:
Rebased. All review comments addressed.

jaredbeck:
> I am as confused as you are @jaredbeck.

Haha OK good :grin: I won't worry about it, thanks.

PR #6608, opened 2018-12-27
[Fix #6586] Add DisjunctiveAssignmentInConstructor

### LINKED ISSUE #6586: Idea: Cop for unnecessary `||=` operator in the initializer

I often stumble across code like this:

```ruby
class Person
  def initialize(full_name)
    @full_name ||= full_name
  end

  # ...
end
```

In this case, I find the `||=` usage unnecessary as the `#initialize` method is called just once.

I believe that this is an offence and deserves a dedicated cop.

Thank you for your time and consideration. Would be happy to hear back from you. 

Blank

**Additional context**

Blank

jaredbeck:
This new cop checks constructors, warns about disjunctive assignment
to instance variables, suggests using plain assignment.

This initial implementation is intended to be conservative, aware of
the limitations of static analysis. IOW, it only registers the most
obvious offenses.

-----------------

* [x] Added tests.


koic:
This cop seems to be suitable for `Lint` department than `Style` department.

spec/rubocop/cop/style/disjunctive_assignment_in_constructor_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      context 'constructor calls super after assignment' do
+        it 'registers an offense' do
+          expect_offense(<<-RUBY.strip_indent)
+          class Banana
+            def initialize
+              @delicious ||= true
+                         ^^^ Unnecessary disjunctive assignment. Use plain assignment.
+              super
+            end
+          end
```
koic: This indentation is misaligned.

lib/rubocop/cop/style/disjunctive_assignment_in_constructor.rb
```diff
@@ -0,0 +1,83 @@
...
+      #   end
+      class DisjunctiveAssignmentInConstructor < Cop
+        include DefNode
+
+        MSG = 'Unnecessary disjunctive assignment. Use plain assignment.'.freeze
+
+        def on_def(node)
+          check(node)
+        end
+        alias on_defs on_def
```
koic: I think that aliasing to `defs` seems unnecessary because `initialize` method is an instance method.

lib/rubocop/cop/style/disjunctive_assignment_in_constructor.rb
```diff
@@ -0,0 +1,83 @@
...
+      #   def initialize
+      #     @x ||= 1
+      #   end
+      #
+      #   # good
+      #   def initialize
+      #     @x = 1
+      #   end
+      class DisjunctiveAssignmentInConstructor < Cop
+        include DefNode
```
koic: Is this `include DefNode` necessary?

koic:
Below is an edge case, it seems difficult to prevent this false positive.

```ruby
class Base
  def initialize
    @config ||= 'base'
  end
end

class Derived < Base
  def initialize
    @config = 'derived'

    super
  end

  def show
    puts @config
  end
end

Derived.new.show #=> 'derived'
```

It seems good to consider making it `Safe: false` 🤔

bbatsov:
I'm fine with the proposed cop, but it needs a matching entry in the style guide.

jaredbeck:
> I'm fine with the proposed cop, but it needs a matching entry in the style guide.

Style guide entry added: https://github.com/rubocop-hq/ruby-style-guide/pull/732

mikegee:
Since this cop is technically unsafe even with this very conservative approach, I would prefer to let it check all the lines in `initialize`.

Drenmi:
While "disjunctive assignment" is more specific, I never heard anyone use this name before. Would it be better to opt for the less specific, but more colloquial "conditional assignment"?

jaredbeck:
> While "disjunctive assignment" is more specific, I never heard anyone use this name before. Would it be better to opt for the less specific, but more colloquial "conditional assignment"?

I considered both. I slightly prefer "disjunctive" but happy to change it if the rubocop team reaches that decision.

PR #6611, opened 2018-12-27
Various omit parentheses fixes volume n+1

gsamokovarov:
🎶 He's making a bug-list
🎶 And checking it twice
🎶 Gonna find out if the bugs are naughty or nice
🎶 `omit_parentheses` fixes are coming to Rubocop 🎄 

---

Before:

```ruby
def foo
  if cond.present? && verify?(:something)
    h.do_with(kw: value)
  elsif cond.present? || verify?(:something_else)
    h.do_with(kw: value)
  elsif whatevs?
    h.do_with(kw: value)
             ^^^^^^^^^^^ Omit parentheses for method calls with arguments.
  end
end
```

Now:

```ruby
def foo
  if cond.present? && verify?(:something)
    h.do_with(kw: value)
             ^^^^^^^^^^^ Omit parentheses for method calls with arguments.
  elsif cond.present? || verify?(:something_else)
    h.do_with(kw: value)
             ^^^^^^^^^^^ Omit parentheses for method calls with arguments.
  elsif whatevs?
    h.do_with(kw: value)
             ^^^^^^^^^^^ Omit parentheses for method calls with arguments.
  end
end
```

---

Before:

```ruby
bar.foo.quux = A::B.new(c) # No offense here!
bar.foo(42).quux = A::B.new(c)
       ^^^^ Omit parentheses for method calls with arguments.
```

Now:

```ruby
bar.foo.quux = A::B.new(c)
bar.foo(42).quux = A::B.new(c)
```

### DIFF
--- a/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
+++ b/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
@@ -245,7 +245,8 @@ module RuboCop
         def call_in_logical_operators?(node)
           node.parent &&
             (logical_operator?(node.parent) ||
-             node.parent.descendants.any?(&method(:logical_operator?)))
+             node.parent.send_type? &&
+             node.parent.arguments.any?(&method(:logical_operator?)))
         end
 
         def call_in_optional_arguments?(node)
--- a/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
+++ b/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
@@ -268,7 +268,9 @@ module RuboCop
         end
 
         def call_as_argument_or_chain?(node)
-          node.parent && (node.parent.send_type? || node.parent.csend_type?)
+          node.parent &&
+            (node.parent.send_type? && !assigned_before?(node.parent, node) ||
+             node.parent.csend_type?)
         end
 
         def hash_literal_in_arguments?(node)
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/style/method_call_with_args_parentheses.rb
```diff
@@ -245,7 +245,8 @@ def call_in_literals?(node)
         def call_in_logical_operators?(node)
           node.parent &&
             (logical_operator?(node.parent) ||
-             node.parent.descendants.any?(&method(:logical_operator?)))
+             node.parent.send_type? &&
+             node.parent.arguments.any?(&method(:logical_operator?)))
```
bbatsov: I think some other cop(s) had logic for complex conditions as well, so you might do some grepping and reuse/duplicate it to be on the safe side. I guess it should be `AssignmentInCondition` or something like this.

bbatsov:
I think you officially broke all the bugginess records. 🤣

gsamokovarov:
And we're still finding broken cases! 😅

lib/rubocop/cop/style/method_call_with_args_parentheses.rb
gsamokovarov: The problem was around the use of logic operators in send arguments in the conditions and the bad check for it before this change. I haven't found any similar logic in `AssingmentInCondition` though. Maybe close enough for the other change that checks for assignment in chained calls? 

I'll see what I can dedup from this change. We have common utilities to the `MethodCallWithoutArgsParamethers` cop as well.

gsamokovarov:
Sorry, didn't have the time to get back to this sooner. May try to refactor the rule a bit when I get more time.

PR #6622, opened 2019-01-02
windows fix

ahorek:
PR https://github.com/rubocop-hq/rubocop/pull/6546 introduced a regression

sh command is usually not available on Windows (it is on appveyor)

it fails like this
```
No such file or directory - sh -c 'command -v git'
```


AlexWayfer:
I think, condition like this:

```ruby
Gem.win_platform?
```

is better than rescuing exception.

---

> I am a little suprised to see git related code in RuboCop. I thought we were avoiding tool specific features because this potentially opens the flood gates to start support Mercurial, SVN, and any other source control tool.

I'm a little surprised to see warnings from RuboCop about files ignored by `git`. Sometimes there are hundreds of warnings.

ahorek:
@AlexWayfer ok, but then we need to find an alternative for
\`sh -c 'command -v git``
to check that git is available

IMO rescuing an exception is better and safer alternative that shelling an additional external command

bbatsov:
> I am a little suprised to see git related code in RuboCop. I thought we were avoiding tool specific features because this potentially opens the flood gates to start support Mercurial, SVN, and any other source control tool.

@rrosenblum Yeah, I resisted adding support for this for a while for that same reason. Recently I decided to make an exception given the wide usage of git, but in light of the reported problems I'm considering revering this.

PR #6623, opened 2019-01-03
Fix heredoc detection in trailing comma

palkan:
Fixes the regression (false negatives) introduced in https://github.com/rubocop-hq/rubocop/commit/27341536e6cbea0105a6e3ee887a2f292517538c.

When `#` appeared inside a heredoc, we assume it's not a heredoc.

Example failing code:

```ruby
test(
  <<-CODE.squiggly
    ,
    # and a "comment"
  CODE
)
```

(See failed run https://travis-ci.org/rubocop-hq/rubocop-md/jobs/474056101#L469)

This PR changes the way we detect heredocs: from regexeps to AST "analysis".

**NOTE:** I'm not very happy with the implementation (it's not easy to cover all the cases), but, IMO, it is much better than regexp-based checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#6623](https://github.com/rubocop-hq/rubocop/pull/6623): Fix heredoc detection in trailing comma. ([@palkan][])
+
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #6626, opened 2019-01-03
Fix passing directories as arguments to rubocop

### LINKED ISSUE #6625: Passing directory to rubocop doesn't work
On CI we only want to run rubocop in a few directories, to achieve this we pass the directories paths as arguments, as specified in the manual:

```
Alternatively you can pass rubocop a list of files and directories to check:

$ rubocop app spec lib/something.rb
```
[source](https://docs.rubocop.org/en/latest/basic_usage/)

Since we updated rubocop 0.61.1 → 0.62.0 this stopped working, as you can see bellow, if we pass a directory as an argument rubocop will stop inspecting all files.

I've narrowed down the breaking change to this PR https://github.com/rubocop-hq/rubocop/pull/6546

I will poke around and try to fix it.

--------

## Expected behavior

```
=> rubocop app
Inspecting 911 files
```
`app` is one of the directories of the project

## Actual behavior

```
=> rubocop app
Inspecting 0 files
```
`app` is one of the directories of the project

I did some experiments and using `../main_directory/app` somehow fixes the problem.`./app` still has the same problem

```
=> rubocop ../project/app
Inspecting 911 files
``` 

Run `rubocop directory_path`

## RuboCop version

```
0.62.0 (using Parser 2.5.3.0, running on ruby 2.5.3 x86_64-darwin18)
```

tiagotex:
The regression seems to be caused by not adding the correct prefix to the list of git files. 

I created tests and made the code pass for all the ways I thought the user might send us a directory:

`.`
`./`
`directory`
`./directory`
`pwd/directory`

Is there another form to pass a directory that I might have missed?

My implementation to add the prefix seems a bit clumsy to me, I have the feeling there must be a better way to achieve this. I'm open for suggestions on how to improve it.

[Fix #6625]


rrosenblum:
> Is there another form to pass a directory that I might have missed?

I assume a trailing `/` has no impact on this?

```
directory/
```

What about some weird ones like
```
..
directory/.. (current dir)
directory/foo/../bar
~/something
.*
.**
```

tiagotex:
After looking at the examples you suggested a better solution finally appeared 😃

If we always expand the path we guarantee we are checking the right folder and that the files we find using `Dir.glob` and `git ls-files` always match.

I added test for your suggestions except for `~/something`, `.*` and `.**`, as far as I can see this was not previously supported as a directory for rubocop, should it be supported?

Since we are now transforming every relative path into full paths this might have some performance impact, do we have some benchmarks in place?

tiagotex:
I just noticed that we were already expanding the paths in `find(args)` https://github.com/rubocop-hq/rubocop/blob/b980b6be09bd8d3e58234a81704125769fc4cc56/lib/rubocop/target_finder.rb#L44
this means we are currently repeating the work, so I will change `process_explicit_path` to also return full path and remove it from `find`

rrosenblum:
> I added test for your suggestions except for ~/something, .* and .**, as far as I can see this was not previously supported as a directory for rubocop, should it be supported?

I personally feel that we should attempt to support as much as possible. If it wasn't supported before, it isn't high priority to support right now. If it seems easy to support, go for it. If not, we can wait until someone requests that functionality. Regardless of our support for it, a user can still pass whatever they want so it is good for us to understand what happens in the application.

> Since we are now transforming every relative path into full paths this might have some performance impact, do we have some benchmarks in place?

I don't think we have any forms of performance testing.

This sounds relatively similar to a bug that I fixed several years ago in commit 322579d81. There was (possibly still is) a bug with jRuby and dir globbing that caused the entire filesystem (or massive parts of it) to be globbed. When this happened, the jRuby CI builds would time out for taking too long (30+ minutes).

tiagotex:
Removed file expansion from `find` and made `target_files_in_dir` and `process_explicit_path` return full paths.

lib/rubocop/target_finder.rb
```diff
@@ -94,7 +94,9 @@ def find_files(base_dir, flags)
                   # on the top level, which would otherwise not be found.
                   wanted_toplevel_dirs.unshift("#{base_dir}/*")
                 end
-      Dir.glob(pattern, flags).select { |path| FileTest.file?(path) }
+      Dir.glob(pattern, flags)
+         .select { |path| FileTest.file?(path) }
+         .map { |relative_path| File.expand_path(relative_path) }
```
rrosenblum: Nitpicky, but if you are worried about performance, using `each_with_object` is better than chaining `select...map`.
```suggestion
Dir.glob(pattern, flags).each_with_object([]) do |path, files|
  next unless FileTest.file?(path)
  files << File.expand_path(relative_path)
end
```

... I don't think github's suggested change feature will work here since my suggestion spans multiple lines.

tiagotex:
> I personally feel that we should attempt to support as much as possible. If it wasn't supported before, it isn't high priority to support right now. If it seems easy to support, go for it. If not, we can wait until someone requests that functionality. Regardless of our support for it, a user can still pass whatever they want so it is good for us to understand what happens in the application.

 I'll have a look on how to support them but for now it might be better to release this as a bug fix release, since I imagine I'm not the only one that had rubocop silently stop working on CI.

> This sounds relatively similar to a bug that I fixed several years ago in commit [322579d](https://github.com/rubocop-hq/rubocop/commit/322579d817dab398c194cd97b442be5b947bd4ca). There was (possibly still is) a bug with jRuby and dir globbing that caused the entire filesystem (or massive parts of it) to be globbed. When this happened, the jRuby CI builds would time out for taking too long (30+ minutes).

Yeah, the reason we encountered this bug is that we have a CI server with a big file system and for some reason when we run rubocop without any directories in the args it would take a long time (10+ minutes) finding files. I will try to debug this on our CI server and post the results in another issue.

rrosenblum:
> I'll have a look on how to support them but for now it might be better to release this as a bug fix release, since I imagine I'm not the only one that had rubocop silently stop working on CI.

I think that is the best plan.

> Yeah, we encountered this bug because we have a CI server with a big file system and for some reason when we run rubocop without any directories in the args it would take a long time (10+ minutes) finding files, adding directories as arguments fixed this issue. I will try to debug this on our CI server and post the results in another issue.

I am a little confused by this. Is the code in this PR a fix for slow dir globbing on you CI server or is your CI server issue unrelated?

tiagotex:
The code in this PR is only fixing passing directories as arguments.  We were passing directories as arguments on CI because when we didn't pass any arguments the dir globbing was too slow.

Our CI server issue is unrelated to this PR.

PR #6629, opened 2019-01-06
[Fix #6627] Fix handling of hashes (with heredoc) in Style/TrailingCommaInArguments

### LINKED ISSUE #6627: Regression in 0.62.0: Style/TrailingCommaInArguments erroneously corrects heredoc
## Expected behavior

Rubocop should not register any offenses for the following:

```ruby
order(
  <<-SQL.squish
    #{redacted}.a ASC,
    #{redacted}.b ASC,
    #{redacted}.c ASC
  SQL
)
```

## Actual behavior

Registers erroneous offense.

```plain
redacted.rb:3:22: C: Style/TrailingCommaInArguments: Avoid comma after the last parameter of a method call.
    #{redacted}.a ASC,
                     ^
```

`--auto-correct` erroneously removes all commas in the heredoc.

Put the above in a file and run rubocop 0.62.0 (default config) on it.

## RuboCop version

```
bundle exec rubocop -V
0.62.0 (using Parser 2.5.3.0, running on ruby 2.5.3 x86_64-darwin18)
```

## Analysis

This is a regression between 0.61.1 and 0.62.0. Just looking at the changelog, it was probably introduced by https://github.com/rubocop-hq/rubocop/pull/6581 but I have not confirmed that yet.

abrom:
Extends #6623 fix for regression added in #6389 where string interpolation within heredoc would cause following commas (within the string) to be treated as code. This change addresses use case where heredoc is a parameter of a hash which contains string interpolation

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug fixes
 
+* [#6627](https://github.com/rubocop-hq/rubocop/pull/6627): Fix handling of hashes in trailing comma. ([@abrom][])
 * [#6623](https://github.com/rubocop-hq/rubocop/pull/6623): Fix heredoc detection in trailing comma. ([@palkan][])
 * [#6100](https://github.com/rubocop-hq/rubocop/issues/6100): Fix a false positive in `Naming/ConstantName` cop when rhs is a conditional expression. ([@tatsuyafw][])
 
--- a/lib/rubocop/cop/mixin/trailing_comma.rb
+++ b/lib/rubocop/cop/mixin/trailing_comma.rb
@@ -180,7 +180,9 @@ module RuboCop
         #       ...
         #     SOURCE
         #   })
-        return heredoc?(node.children.last) if node.pair_type?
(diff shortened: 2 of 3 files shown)


abrom:
It looks like there are still some issues not addressed @palkan from #6623

For example, it looks like the auto-correct is removing the comma from within a method call:

```
       expected: "some_method(\n  bar: <<-BAR\n    \#{other_method(a, b)} foo, bar\n  BAR\n)\n"
            got: "some_method(\n  bar: <<-BAR\n    \#{other_method(a b)} foo, bar\n  BAR\n)\n"
     
       (compared using ==)
     
       Diff:
       @@ -1,6 +1,6 @@
        some_method(
          bar: <<-BAR
       -    #{other_method(a, b)} foo, bar
       +    #{other_method(a b)} foo, bar
          BAR
        )
```

It is also treating this as an offense:

```
            some_method(
              bar: <<-BAR
                #{variable} foo, bar
              BAR
            )
```

Thoughts?

FYI @jaredbeck

PR #6640, opened 2019-01-08
Fix grammar

sodapopcan:


manual/development.md
```diff
@@ -107,7 +107,7 @@ NodePattern.new('send').match(node) # => true
 ```
 
 It matches because the root is a `send` type. Now lets match it deeply using
-parens to define details for sub-nodes. If you don't care about what a internal
+parens to define details for sub-nodes. If you don't care about what an internal
```
bbatsov: I guess it shouldn't be parens (plural) as well. :-)

rrosenblum: I think parens (plural) is fine. The change here a to an because it being followed by a word that starts with a vowel. This isn't a pluralization change.

bbatsov: Yeah, I misread the text here.

bbatsov:
Thanks!

PR #6647, opened 2019-01-09
[Fix #6641] Specify `Performance/RangeInclude` as unsafe

### LINKED ISSUE #6641: Performance/RangeInclude is not safe
`Performance/RangeInclude` should not suggest to use `Range#cover?` instead of `Range#include?`  in safe mode, nor should it auto-correct it in safe mode, because they don't provide equal results.

```
#!/usr/bin/env ruby
# frozen_string_literal: true

p ('a'..'b').include?('aa') # false
p ('a'..'b').cover?('aa') # true
```

--------

## Expected behavior

`Performance/RangeInclude` should not run in safe mode.

## Actual behavior

`Performance/RangeInclude` replaces `Range#include?` with `Range#cover?`.

See example script above.

## RuboCop version

```
$ rubocop -V
0.62.0 (using Parser 2.5.3.0, running on ruby 2.5.3 x86_64-darwin17)
```

koic:
Fixes #6641.

This PR specifies `Performance/RangeInclude` as unsafe because `Range#include?` and `Range#cover?` are not equivalent behaviour.

-----------------

* [ ] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2074,6 +2074,7 @@ Performance/RangeInclude:
   Reference: 'https://github.com/JuanitoFatas/fast-ruby#cover-vs-include-code'
   Enabled: true
   VersionAdded: '0.36'
+  Safe: false
 
 Performance/RedundantBlockCall:
   Description: 'Use `yield` instead of `block.call`.'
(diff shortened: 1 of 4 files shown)


config/default.yml
```diff
@@ -2074,6 +2074,7 @@ Performance/RangeInclude:
   Reference: 'https://github.com/JuanitoFatas/fast-ruby#cover-vs-include-code'
   Enabled: true
   VersionAdded: '0.36'
+  Safe: false
```
koic: @bbatsov Is it necessary to add `VersionChanged` metadata when changing `Safe`?

bbatsov:
Thanks!

Issue #6410, opened 2018-10-27
Feature request: Layout/AlignHash option to allow for `key` and `table` styles to both be permissible


mattbrictson:
**Is your feature request related to a problem? Please describe.**

The changes to `Layout/AlignHash` in 0.60 mean that you can no longer mix the `key` and `table` style without one of them being reported as a violation. For example:

```ruby
key_style = {
  key: :value,
  another: :value,
  yet_another: :value
}
table_style = {
  key:         :value,
  another:     :value
  yet_another: :value
}
```

Before 0.60, if you chose the `key` style, both hashes would be allowed, because in both cases the keys are left-aligned.

However starting with 0.60, RuboCop strictly enforces the whitespace to the right of the keys. If you set the enforced style to `key`, the second hash will get marked as a violation. If you choose `table`, then the first will be a violation.

I believe that there is room for both styles in a single codebase. For small hashes, `key` is appropriate. But for large hashes the `table` style is much easier to read. Take a look at [the changes introduced by 0.60 to RuboCop's own codebase](https://github.com/rubocop-hq/rubocop/pull/6205/files?utf8=%E2%9C%93&diff=split#diff-a533b255d2dfd2cb88d774f52b8a19e2). I think many people would agree that readability has been decreased.

**Describe the solution you'd like**

I am open to ideas. A naive solution would be a `key_or_table` option that allows for both styles of whitespace to be accepted. I.e.

```yaml
Layout/AlignHash:
  EnforcedColonStyle: key_or_table
  EnforcedHashRocketStyle: key_or_table
```

**Describe alternatives you've considered**

Currently the only way to mix key and table styles in the same codebase is to turn off `AlignHash` altogether. This is a big step backwards, though, because now even blatantly bad alignments will be allowed without any violations, like this:

```ruby
{
  key: :value,
    another:   :value
}
```

Or I could set the enforced style to `key`, and then use a `# rubocop:disable` comment around every hash where I want to use table syntax. That is a bit tedious, and again, it disables alignment checking completely. To my knowledge there is not a way to change the enforced style with a `# rubocop` comment; you can only disable the cop.

bquorning:
[...]

PR #6649, opened 2019-01-09
Layout/AlignHash allow two options

stoivo:
I my company we like to use Hashes in both table and key format. We would like to have rubucop allow both so I make a path an it seams to work for me.
What I want is 
```ruby
# bad
{
  :foo => bar,
   :ba => baz
}

# good
{
  :foo => bar,
  :ba => baz
}
{
  :foo => bar,
  :ba  => baz
}
```
I wanted to open the PR to check if it would be accepted if I can submitt a nice PR for it. I can add more spec. I wil do so if you think you want this feature

Right now I made it an option not_<any_of_the_setting>. I think it would be nicer if you could spesifi it as a list in .`rubocop.yml`
```yml
Layout/AlignHash:
  EnforcedHashRocketStyle: 
    - table
    - key
  # vs
  EnforcedHashRocketStyle: not_separator
```

Thoughts? Do you have an example where you do this?  
If I can change it to a list the first one can be the format to autofix to.  

**Edit**: after a little more testing is should autofix to the the format how as the least amoint of offences.

-----------------

* [x] Added tests.


stoivo:
I should add some spec to this.

stoivo:
I have written spec. I would very much like some feedback on this.

stoivo:
Hi, @pocke 
I see you have merges some other MR here would you review this MR too?

stoivo:
I will rebase after it is review so I don't wast to much time rebasing and force pushing

bbatsov:
Sorry about the slow turnaround on this. I'll take a look a closer look soon. In general I'm wary of adding 3 more styles to this cop, if we probably need just one. From what I gathered most people want something that allows a combination of table or key, right?

stoivo:
Thank you for taking a look at it.
Yes, we want a combination of table or key. I think we kan remove the two others, I added those because it was't too much work. We wnt to enforce one style in one hash, either table or key.

pocke:
Sorry for the late reply.
I have the same concern. If a style is added to the cop in the future, `no_*` styles behaviour is changed.



How about making EnforcedStyle receiving multiple options?
exmaple:

```yaml
EnforcedStyle:
  - table
  - key
```


I think it is easy to use, but it is inconsistent with other cops because other cops do not support multiple options.


If you and many users need to allow `table` and `key`, I think adding one option is good. The option allows `table` and `key` styles.
And I'd like to rename the option to `table_and_key`. Because `no_*` option will be changed the behaviour if an option is added to the cop in the future.

bbatsov:
I think `table_or_key` might be a better name, unless we come up with a single word to describe this. 

> How about making EnforcedStyle receiving multiple options?

That's a pretty interesting idea in general - make it possible to accept several styles where it makes sense. It would probably be reasonable simple to do this, and it might be my preference. @rubocop-hq/rubocop-core What do you think about this idea?

stoivo:
I think the  idea of supporting either a string or a list would be very cool and flexible.  And when you enter all thee formats, any format is valid but only one format in one hash.

I did one attempt at making it a list in the beginning but faced some dificulities. I would be happy diving it a second time, with some mentoring and pointing me where to look.

bbatsov:
The only small problem is which style to suggest in case a hash doesn't match any of them. Probably the one that was specified first.

stoivo:
@bbatsov, yes to the one with leaset offences. But thats allreade fixed 
https://github.com/stoivo/rubocop/blob/master/lib/rubocop/cop/layout/align_hash.rb#L213

stoivo:
I think thats better becuase If you have 
```ruby 
{
  ab:    123,
  assdf: 12332,
  asf:   12332,
}
```
And you add one element 

```ruby
{
  ab:    123,
  assdf: 12332,
  asf:   12332,
  b: true,
}
```

You probably want to format it to table if you allow both key and table

bbatsov:
So, what are doing here? Do we agree to try the multiple enforced styles approach?

stoivo:
I can use the name `table_or_key` instead of `not_separator`.
@bbatsov  What do you think of autocorrection to the one with least offences?

CHANGELOG.md
```diff
@@ -15,7 +16,7 @@
 * [#6858](https://github.com/rubocop-hq/rubocop/issues/6858): Fix an incorrect auto-correct for `Lint/ToJSON` when there are no `to_json` arguments. ([@koic][])
 * [#6865](https://github.com/rubocop-hq/rubocop/pull/6865): Fix deactivated `StyleGuideBaseURL` for `Layout/ClassStructure`. ([@aeroastro][])
 * [#6868](https://github.com/rubocop-hq/rubocop/pull/6868): Fix `Rails/LinkToBlank` auto-correct bug when using symbol for target. ([@r7kamura][])
-* [#6869](https://github.com/rubocop-hq/rubocop/pull/6869): Fix false positive for `Rails/LinkToBlank` when rel is a symbol value. ([@r7kamura][])
+* [#6869](https://github.com/rubocop-hq/rubocop/pull/6869): Fix false positive for `Rails/LinkToBlank` when rel is a symbol value. ([@r7kamura][]
```
jonas054: It looks like you removed a parenthesis by mistake here.

bbatsov:
I actually prefer @pocke's idea and autocorrecting using whatever is first in the `EnforcedStyle` list. I don't like relying on offenses, because that's non deterministic.

CHANGELOG.md
stoivo: Yes, will fix when rebase to remove merge commits, Thanks.

stoivo:
> I actually prefer @pocke's idea and autocorrecting using whatever is first in the `EnforcedStyle` list. I don't like relying on offenses, because that's non deterministic.

Did you se my comment about what happens if you add something to an existing hash but without 100% match? 

> I think thats better becuase If you have
> 
> ```ruby
> {
>   ab:    123,
>   assdf: 12332,
>   asf:   12332,
> }
> ```
> 
> And you add one element
> 
> ```ruby
> {
>   ab:    123,
>   assdf: 12332,
>   asf:   12332,
>   b: true,
> }
> ```
> 
> You probably want to format it to table if you allow both key and table

I just want to say, you have more experence maintaining rubocop so I will do just want to say :)
I pushed a commit where I merged the commits and rebased away the merge commits.
It is't working 100% and I dont have time to fix it today. Will do it shortly.

lib/rubocop/cop/layout/align_hash.rb
```diff
@@ -289,11 +321,13 @@ def correct_key_value(key_delta, key, value, separator)
...
-          case cop_config[key]
-          when 'key'       then KeyAlignment.new
-          when 'table'     then TableAlignment.new
-          when 'separator' then SeparatorAlignment.new
-          else raise "Unknown #{key}: #{cop_config[key]}"
+          {
+            'key' => [KeyAlignment.new],
+            'table' => [TableAlignment.new],
+            'separator' => [SeparatorAlignment.new],
+            'table_or_key' => [TableAlignment.new, KeyAlignment.new]
```
stoivo: shoudl I split the string by `or`and use both sides of the or. So you can do either `table_or_key` or `key_or_table`, if it shoudl autofix to the one spesified first we might want both?

bbatsov:
> I just want to say, you have more experence maintaining rubocop so I will do just want to say :)
I pushed a commit where I merged the commits and rebased away the merge commits.
It is't working 100% and I dont have time to fix it today. Will do it shortly.

Yeah, I did. I don't consider this a problem as in your editor you'd know what to do - you see some warning, but you also know it will disappear once you change it. The auto-correct should just pick the first option:

```
EnforcedStyle:
  - table
  - key
```

In this case - `table`. Just to be clear - I strongly prefer to have a list of options for `EnforcedStyle` than a dedicated `table_or_key` style (although I can live with it). I this case I'd default to `key` i guess.

bbatsov:
@stoivo Ping :-)

stoivo:
@bbatsov, thanks for beein patient with me. 

Could to review it now? 

I am facing a minor issue with rubocop 
```
$ rubocop                                                                                                                                                    15:09:20
Inspecting 1172 files
..................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................C.................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Offenses:

lib/rubocop/config.rb:528:5: C: Metrics/AbcSize: Assignment Branch Condition size for validate_enforced_styles is too high. [20.05/17]
    def validate_enforced_styles(valid_cop_names) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/rubocop/config.rb:528:5: C: Metrics/MethodLength: Method has too many lines. [16/14]
    def validate_enforced_styles(valid_cop_names) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1172 files inspected, 2 offenses detected
```
https://github.com/stoivo/rubocop/blob/0306ae96739b4de743ddb3935adc6611e1134b31/lib/rubocop/config.rb#L528-L548

```ruby
    def validate_enforced_styles(valid_cop_names)
      valid_cop_names.each do |name|
        styles = self[name].select { |key, _| key.start_with?('Enforced') }

        styles.each do |style_name, style|
          supported_key = RuboCop::Cop::Util.to_supported_styles(style_name)
          valid = ConfigLoader.default_configuration[name][supported_key]

          next unless valid
          next if valid.include?(style)
          next if ConfigLoader.default_configuration[name]['AllowListStyles'] &&
                  style.is_a?(Array) &&
                  style.all? { |a_style| valid.include?(a_style) }

          msg = "invalid #{style_name} '#{style}' for #{name} found in " \
            "#{smart_loaded_path}\n" \
            "Valid choices are: #{valid.join(', ')}"
          raise ValidationError, msg
        end
      end
    end
```

How do you want me to solve this?

jonas054:
> How do you want me to solve this?

I think you should extract either the contents of the inner `each` loop, or just the complex expression
```ruby
ConfigLoader.default_configuration[name]['AllowListStyles'] &&
                  style.is_a?(Array) &&
                  style.all? { |a_style| valid.include?(a_style) }
```
into a new method. Come up with a name for the new method and you're good to go. :slightly_smiling_face:

CHANGELOG.md
```diff
@@ -84,6 +89,7 @@
 ### New features
 
 * [#5184](https://github.com/rubocop-hq/rubocop/issues/5184): Add new multiline element line break cops. ([@maxh][])
+* [#6649](https://github.com/rubocop-hq/rubocop/pull/6649): Layout/AlignHash supports table_or_key. ([@stoivo][])
```
bbatsov: I think you forgot to remove this.

config/default.yml
```diff
@@ -260,6 +260,7 @@ Layout/AlignHash:
                  Align the elements of a hash literal if they span more than
                  one line.
   Enabled: true
+  AllowListStyles: true
```
bbatsov: I think `AllowMultipleStyles` would be a better name.

lib/rubocop/cop/layout/align_hash.rb
```diff
@@ -10,6 +10,7 @@ module Layout
       #   - key (left align keys, one space before hash rockets and values)
       #   - separator (align hash rockets and colons, right align keys)
       #   - table (left align keys, hash rockets, and values)
+      #   - or a list of the above options
```
bbatsov: I think this should say something like "Alternatively you can specify multiple allowed styles. That's done by passing a list of styles to `EnforcedStyles`. " (or something along those lines)

spec/rubocop/config_spec.rb
```diff
@@ -174,6 +174,37 @@
       end
     end
 
+    context 'when the configuration includes multiple valid EnforcedStyle' do
```
bbatsov: Should we raise an error if the cop doesn't support multiple style explicitly?

bbatsov:
@stoivo Added a bit of small remarks. Generally the code looks pretty good and we're very close to finally merging this.

bquorning:
Should we consider reverting https://github.com/rubocop-hq/rubocop/pull/6205/commits/807b362db653be8253b7bf0500edab93ac8d6dfa as part of this PR?

bquorning:
This will fix #6410, right?

bbatsov:
> Should we consider reverting 807b362 as part of this PR?

Yes.

stoivo:
Ready for an onther review.

I reverted the commit and did some small modifications to resolve conflicts.

bbatsov:
Good job! There are some small things that'd I'd like us to polish down the road and we have to document somewhere the ability to have cops that support multiple styles, but I'll merge the PR in its current state as it's pretty big and you've been wasting a lot of efforts keeping it up to date. Thanks! 🙇

stoivo:
Thank you too. I wm happy to help out a more but I think is is better if you do the polish and add the documentation where it belongs. Very happy with submitting a patch to rubocop. Love the tool ❤️

PR #6650, opened 2019-01-10
fix Lint/AmbiguousBlockAssociation cites unambiguous lambda

ayumu838:
fix AmbiguousBlockAssociation accepts  to rspec `to change lambda`

With this fix the following code is ok for AmbiguousBlockAssociation cop:

This method is also used in the [official document](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher)


bbatsov:
Thanks!

PR #6652, opened 2019-01-10
[Fix #6651] Make Style/RegexpLiteral correctly auto-correct inner slashes when there's interpolation

### LINKED ISSUE #6651: Style/RegexpLiteral autocorrect not parsing interpolated string
Cleaning up some code and I ran across an issue with regexes that contain string interpolation.

This is the regex in question:

`/users\/#{user.id}\/forms/`

## Expected behavior

Autocorrect should convert the regex above to

`%r{users/#{user.id}/forms}`

## Actual behavior

Autocorrect converts the regex to

`%r{users//user.id}\/forms}`

```
# echo "/users\/#{user.id}\/forms/" > tmp.rb
# rubocop -a tmp.rb
# cat tmp.rb
%r{users//user.id}\/forms}
```

## RuboCop version

```
# rubocop -V
0.62.0 (using Parser 2.5.3.0, running on ruby 2.3.5 x86_64-darwin17)
```

roooodcastro:
Let me start this by saying that I don't know much about ASTs and never messed with them in Ruby, so this fix might be all wrong.

The issue this PR intends to fix is one where the `Style/RegexpLiteral` cop fails to correctly replace and fix inner slashes after changing a regex from `/users\/#{user.id}\/forms/` to `%r{users/#{user.id}/forms}`.

Analysing the execution flow of this cop, I noticed that when there's no interpolation, the resulting AST's regexp node has only a single child, of type `srt`. The cop correctly corrects this case. When there is an interpolation in the mix, however, the AST node now has 4 children, of types: `[:str, :begin, :str, :regopt]`.

Example output of `node.child_nodes`:

* when regex is `/users\/user.id\/forms/`: (no interpolation)

```ruby
> node.child_nodes
=> [s(:str, "users/user.id/forms"), s(:regopt)]
```

* when regex is `/users\/#{user.id}\/forms/`: (interpolation):

```ruby
> node.children
=> [s(:str, "users/"), s(:begin,
  s(:send,
    s(:send, nil, :user), :id)), s(:str, "/forms"), s(:regopt)]
```

Specifically, the cop method `node_body` filters these children and only considers the `str` children. This is useful to avoid trying to detect and fix stuff inside the interpolation which isn't part of the regex, but at the same time, in this specific case, the resulting body will be `users\/\/forms`. Rubocop can correctly identify both offenses there, but for some reason it cannot correct them. The fix for this was to have `node_body` ignore the `begin` node for the detection, but include it for the auto-correction. This way the cop works as intended for all (tested) cases.

I just don't know if including the `begin` node type in there is enough, or if there's some other type that should be included as well. Please advise if this fix is not adequate or safe.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [#6617](https://github.com/rubocop-hq/rubocop/issues/6617): Prevent traversal error on infinite ranges. ([@drenmi][])
 * [#6625](https://github.com/rubocop-hq/rubocop/issues/6625): Revert the "auto-exclusion of files ignored by git" feature. ([@bbatsov][])
 * [#4460](https://github.com/rubocop-hq/rubocop/issues/4460): Fix the determination of unsafe auto-correct in `Style/TernaryParentheses`. ([@jonas054][])
+* [#6651](https://github.com/rubocop-hq/rubocop/issues/6651): Fix auto-correct issue in `Style/RegexpLiteral` cop when there is string interpolation. ([@roooodcastro][])
 
 ### Changes
 
@@ -3746,3 +3747,4 @@
 [@amatsuda]: https://github.com/amatsuda
 [@Intrepidd]: https://github.com/Intrepidd
 [@Ruffeng]: https://github.com/Ruffeng
+[@roooodcastro]: https://github.com/roooodcastro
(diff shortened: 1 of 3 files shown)


Drenmi:
Thanks for your contribution!

This looks like a reasonable fix to me. I can't think of any other cases off the back of my head, but I'm sure if there are any, they will eventually bubble up in the issues. 🙂 

🚀

roooodcastro:
Thank you! I just pushed a rebase now to resolve some possible conflicts with master.

bbatsov:
Thanks!

PR #6657, opened 2019-01-11
Single constant initialization cop

jbernardo95:
Adds a new cop to verify that constants are only initialized once.


wata727:
This feature looks good. But, unfortunately, it is difficult to perform *flow-sensitive* analysis like this in RuboCop yet. For example, the following example is perfectly correct:

```ruby
if ENV["ENABLED"]
  ENABLED = true
else
  ENABLED = false
end

puts ENABLED
```

However, this cop reports offences like the following:

```
test.rb:4:3: W: Lint/SingleConstantInitialization: Constant ENABLED initialized more than once.
  ENABLED = false
  ^^^^^^^^^^^^^^^
```

In order to prevent this, we need to perform data-flow analysis, but RuboCop doesn't have that mechanism...

jbernardo95:
Hum 🤔 I never thought of that case. I was always thinking about constants defined on top of the file/module. But this example is perfectly valid.

Closing for now. Until data flow analysis is possible.

PR #6658, opened 2019-01-12
Adds parser gem as a development dependency

jbernardo95:
The first recommended in the [development guide](https://rubocop.readthedocs.io/en/latest/development/) is to install de `parser` gem. So, why not have it as a development dependency ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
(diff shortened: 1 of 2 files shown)


pocke:
https://github.com/rubocop-hq/rubocop/blob/f174fe62d1a59125bd3ae9de2d4b0a4d98ce2c81/rubocop.gemspec#L39
The gemspec already has parser as a runtime dependency, so `bundle install` command installs parser gem.

Probably we can update the documentation to avoid the same confusion. `gem install parser` is unnecessary if `bundle install` is executed. Probably we can just replace `gem install parser` with `bundle install`.

PR #6659, opened 2019-01-13
[Fix #4460] Improve TernaryParentheses#unparenthesized_method_call?

### LINKED ISSUE #4460: Style/TernaryParentheses autocorrect changes code intent
The [`Style/TernaryParentheses`](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/ternary_parentheses.rb) cop removes parentheses around ternary conditions when auto-correcting.

This week we had a case where applying this cop's autocorrect resulted in a change in code behaviour. I understand from other PRs/issues that autocorrects are intended to not change code's behaviour, only it's style.

Given code like `(odd? x) ? "A" : "B"`, this cop transforms it into `odd? x ? "A" : "B"` which will return a different result.

--------

## Expected behavior

Given: `(odd? x) ? "A" : "B"`

I would expect the cop to do one of:
- Leave it as-is
- Register an offense
- Autocorrect to `odd?(x) ? "A" : "B"`

## Actual behavior

The code is auto-corrected to `odd? x ? "A" : "B"` (interpreted as `odd?(x ? "A" : "B")`)

I've reproduced in a test which you can run: https://github.com/petehamilton/rubocop/commit/a9ec2ee966d989d223d80e244ef183878c83cd3e

## RuboCop version

```
$ rubocop -V
0.49.1 (using Parser 2.4.0.0, running on ruby 2.3.3 x86_64-darwin16)
```

_Note: I originally thought this might be a similar issue to https://github.com/bbatsov/rubocop/issues/4118 but on re-reading I think it's actually different so I've opened this issue to discuss_

jonas054:
We say that a method call is not safe for removing outer parentheses if the method takes an argument but doesn't enclose the argument in parentheses. E.g., `(func x) ? 1 : 0`, but the previous solution didn't get all cases right.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [#6526](https://github.com/rubocop-hq/rubocop/issues/6526): Fix a wrong line highlight in `Lint/ShadowedException` cop. ([@tatsuyafw][])
 * [#6617](https://github.com/rubocop-hq/rubocop/issues/6617): Prevent traversal error on infinite ranges. ([@drenmi][])
 * [#6625](https://github.com/rubocop-hq/rubocop/issues/6625): Revert the "auto-exclusion of files ignored by git" feature. ([@bbatsov][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Nice improvement! Thanks!

PR #6660, opened 2019-01-13
Add new `Rails/IgnoredSkipActionFilterOption` cop

wata727:
Fixes https://github.com/rubocop-hq/rubocop-rails/issues/35

This cop checks that `if` and `only` (or `except`) are not used together as options of `skip_*` action filter.

The `if` option will be ignored when `if` and `only` are used together.
Similarly, the `except` option will be ignored when `if` and `except` are used together.

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -2391,6 +2391,13 @@ Rails/HttpStatus:
     - numeric
     - symbolic
 
+Rails/IgnoredSkipActionFilterOption:
+  Description: 'Checks that `if` and `only` (or `except`) are not used together as options of `skip_*` action filter.'
```
koic: It may be useful if API reference is mentioned.

```yaml
Reference: 'https://api.rubyonrails.org/classes/AbstractController/Callbacks/ClassMethods.html#method-i-_normalize_callback_options'
```

bbatsov:
Nicely done! Thanks!

bquorning:
Shouldn’t this cop also apply when defining the callbacks, not just when skipping them? I.e. `(prepend_|skip_)?(before|after|around)_action`

wata727:
@bquorning I've confirmed this behavior in only `skip_*` method.  See also https://github.com/rails/rails/pull/18404#issuecomment-172446870

bquorning:
Interesting. The documentation doesn’t mention that at all: https://api.rubyonrails.org/classes/AbstractController/Callbacks/ClassMethods.html#method-i-_normalize_callback_options

PR #6663, opened 2019-01-14
[Fix #6575] Fix `Naming/PredicateName` cop suggesting invalid rename

### LINKED ISSUE #6575: Naming/PredicateName suggest invalid rename
Naming/PredicateName can currently suggest an invalid rename of a method.

Example code snippet:

```ruby
class BoundingBox
  # ...
  def is_2d?
    @points.size == 4
  end
end
```

--------

## Expected behavior

I expect RuboCop to suggest a method name which is valid Ruby syntax. Or simply allow `is_` prefix when the remainder starts with a number.

## Actual behavior

```
modules/boundingbox.rb:35:9: C: Naming/PredicateName: Rename is_2d? to 2d?.
```

However, if one were to perform the suggested rename you'll get SyntaxError:

```
SyntaxError: (irb):1: syntax error, unexpected tINTEGER
```

```ruby
class BoundingBox
  # ...
  def is_2d?
    @points.size == 4
  end
end
```

## RuboCop version

```
0.61.1 (using Parser 2.5.3.0, running on ruby 2.2.6 x64-mingw32)
```

tejasbubane:
Fixes #6575 

Do not suggest change if the suggested identifier is invalid.

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/naming/predicate_name.rb
+++ b/lib/rubocop/cop/naming/predicate_name.rb
@@ -61,7 +61,7 @@ module RuboCop
         private
 
         def allowed_method_name?(method_name, prefix)
-          !method_name.start_with?(prefix) ||
+          !method_name.match(/^#{prefix}[^0-9]/) ||
             method_name == expected_name(method_name, prefix) ||
             method_name.end_with?('=') ||
             predicate_whitelist.include?(method_name)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * [#6651](https://github.com/rubocop-hq/rubocop/issues/6651): Fix auto-correct issue in `Style/RegexpLiteral` cop when there is string interpolation. ([@roooodcastro][])
 * [#6670](https://github.com/rubocop-hq/rubocop/issues/6670): Fix a false positive for `Style/SafeNavigation` when a method call safeguarded with a negative check for the object. ([@koic][])
 * [#6633](https://github.com/rubocop-hq/rubocop/issues/6633): Fix `Lint/SafeNavigation` complaining about use of `to_d`. ([@tejasbubane][])
+* [#6575](https://github.com/rubocop-hq/rubocop/issues/6575): Fix `Naming/PredicateName` suggesting invalid rename. ([@tejasbubane][])
 
 ### Changes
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Can you rebase this?

lib/rubocop/cop/naming/predicate_name.rb
```diff
@@ -61,7 +61,7 @@ def on_def(node)
         private
 
         def allowed_method_name?(method_name, prefix)
-          !method_name.start_with?(prefix) ||
+          !method_name.match(/^#{prefix}[^0-9]/) ||
```
bbatsov: I'm not sure digits were the only characters that we should be mindful of. Perhaps you can check what exactly was allowed there.

thomthom: https://docs.ruby-lang.org/en/trunk/syntax/methods_rdoc.html#label-Method+Names

> Method names may be one of the operators or must start a letter or a character with the eight bit set. It may contain letters, numbers, an _ (underscore or low line) or a character with the eight bit set. 

> Ruby programs must be written in a US-ASCII-compatible character set such as UTF-8, ISO-8859-1 etc. In such character sets if the eight bit is set it indicates an extended character. Ruby allows method names and other identifiers to contain such characters. Ruby programs cannot contain some characters like ASCII NUL (\x00).

The description isn't very precise to what they mean by a "letter".  But seeing that the eight bit is used to denote an extended character set I think that means within the US-ASCII range; `a-z` and `A-Z`.

So the recipe for valid starting character is `a-z`, `A-Z`, any character with the eight bit set. Also one would have to consider any of the valid operators. At least that is how I interpret the description from the documentation.

tejasbubane: > Identifiers consist of alphabets, decimal digits, and the underscore character, and begin with a alphabets(including underscore).

> Method_name should be either identifier or re-definable operators (e.g. ==, +, -, etc.).

Ref: https://ruby-doc.org/docs/ruby-doc-bundle/Manual/man-1.4/syntax.html

Which means after removing the prefix `is_` from `is_<predicate>?` the remaining `<predicate>` should start with either `a-z`, `A-Z` or `_`. Or in other words should not start with `0-9`.

> Perhaps you can check what exactly was allowed there.

The cop removed prefix and showed the rest as suggested method-name.

bbatsov:
Can you rebase again?

tejasbubane:
@bbatsov Rebased.

thomthom:
Thank you for the fix @tejasbubane!

PR #6664, opened 2019-01-14
[Fix #6633] Fix `Lint/SafeNavigation` complaining about use of `to_d`

### LINKED ISSUE #6633: Lint/SafeNavigationChain shouldn't complain about chaining `to_d`
## Expected behavior

`nil.to_d` returns `0.0`, so Rubocop should **not** complain about this case in particular.

## Actual behavior

It complains:
```
# Lint/SafeNavigationChain: Do not chain ordinary method call after safe navigation operator.
```

**Note:** Rubocop doesn't complain about the  same code with `to_f`.

Just write the following code and run Rubocop:

```ruby
amount&.slice.to_d
```

## RuboCop version

```
0.62.0 (using Parser 2.5.3.0, running on ruby 2.6.0 x86_64-darwin18)
```

tejasbubane:
Fixes: https://github.com/rubocop-hq/rubocop/issues/6633

`to_d` is defined in `bigdecimal/util` in stdlib and hence should be handled like other default methods on `nil`. Refer https://github.com/rubocop-hq/rubocop/issues/6633#issuecomment-452189702

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/nil_methods.rb
+++ b/lib/rubocop/cop/mixin/nil_methods.rb
@@ -2,13 +2,19 @@
 
 module RuboCop
   module Cop
-    # This module provides a list of methods that are either in the NilClass,
-    # or in the cop's configuration parameter Whitelist.
+    # This module provides a list of methods that are:
+    # 1. In the NilClass by default
+    # 2. Added to NilClass by explicitly requiring any standard libraries
+    # 3. Cop's configuration parameter Whitelist.
     module NilMethods
       private
 
       def nil_methods
-        nil.methods + whitelist
+        nil.methods + other_stdlib_methods + whitelist
+      end
+
+      def other_stdlib_methods
+        [:to_d]
       end
 
       def whitelist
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/mixin/nil_methods.rb
```diff
@@ -8,12 +8,16 @@ module NilMethods
...
-        nil.methods + whitelist
+        nil.methods + other_stdlib_methods + whitelist
       end
 
       def whitelist
         cop_config['Whitelist'].map(&:to_sym)
       end
+
+      def other_stdlib_methods
+        [:to_d]
```
bbatsov: Why just not add this to the default whitelist?

bbatsov: (if we keep the current approach we'll also need to update the description of the class)

tejasbubane: I find keeping the three separate slightly cleaner. Updated class description accordingly. Also rebased the PR.

bbatsov:
Can you rebase again?

tejasbubane:
@bbatsov Rebased.

bbatsov:
Thanks!

PR #6667, opened 2019-01-14
In Rails >= 5, use optional: true, not required: false

petehamilton:
The `required` option on belongs_to associations has been deprecated,
instead the recommended (and default) approach is to set this to true by
default and disable it using `optional: true`.

The developer can override this default globally, however, so we can't
say that things like `optional: false` are superfluous, or indeed that
`required: true` isn't necessary.

This change therefore converts any use of `required` to it's inverse
`optional` equivalent.

See: https://guides.rubyonrails.org/5_0_release_notes.html

-----------------

This was originally added in https://github.com/rubocop-hq/rubocop/pull/6596, accidentally merged and then reverted. Let's try this again! :)

* [x] Added tests.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * [#6604](https://github.com/rubocop-hq/rubocop/pull/6604): Add auto-correct support to `Rails/LinkToBlank`. ([@Intrepidd][])
 * [#6660](https://github.com/rubocop-hq/rubocop/pull/6660): Add new `Rails/IgnoredSkipActionFilterOption` cop. ([@wata727][])
 * [#6363](https://github.com/rubocop-hq/rubocop/issues/6363): Allow `Style/YodaCondition` cop to be configured to enforce yoda conditions. ([@tejasbubane][])
+* [#6596](https://github.com/rubocop-hq/rubocop/pull/6596): Add new `Rails/BelongsTo` cop with auto-correct for Rails >= 5. ([@petehamilton][])
```
petehamilton: Am already listed in contributors, so only adding this line.

lib/rubocop/cop/rails/belongs_to.rb
```diff
@@ -0,0 +1,120 @@
...
+      # In the case that the developer is doing `required: false`, we
+      # definitely want to autocorrect to `optional: true`.
+      #
+      # However, without knowing whether they've set overriden the default
+      # value of `config.active_record.belongs_to_required_by_default`, we
+      # can't say whether it's safe to remove `required: true` or whether we
+      # should replace it with `optional: false` (or, similarly, remove a
+      # superfluous `optional: false`). Therefore, in the cases we're using
+      # `required: true`, we'll simply invert it to `optional: false` and the
+      # user can remove depending on their defaults.
```
petehamilton: We could make this configurable (i.e., user indicates their app's default in config), but this will at least do a sensible thing for now.

CHANGELOG.md
bbatsov: 👍

lib/rubocop/cop/rails/belongs_to.rb
bbatsov: 👍

PR #6671, opened 2019-01-16
[Fix #6254] Fix RescueEnsureAlignment for non-local assignments

### LINKED ISSUE #6254: Bug encountered with `Layout/RescueEnsureAlignment` Cop and a memoized method
## Expected behavior

Expected RuboCop to **not flag** the method.

## Actual behavior

```
$ rubocop test.rb
Inspecting 1 file
C

Offenses:

test.rb:8:5: C: Layout/RescueEnsureAlignment: rescue at 8, 4 is not aligned with begin at 6, 13.
    rescue StandardError
    ^^^^^^
```

1. Create `test.rb`:

```ruby
# frozen_string_literal: true

# Dummy Class
class Foo
  def bar
    @bar ||= begin
      expensive_method
    rescue StandardError
      fall_back
    end
  end
end
```
2. Run `rubocop test.rb`

## RuboCop version

```
$ [bundle exec] rubocop -V
0.59.0
```

## Inference

Rubocop recommending to place `rescue` at position 13 in this example is a distasteful :bug:
/cc @rrosenblum

marcotc:
This PR fixes the alignment of `rescue` and `ensure` keywords when those are part of an assignment do-end block.

The behaviour is currently correct when assigning into local variable, but not for instance variables, class variables, global variables or class constants. 

One thing to notice in this PR is that I used all applicable assignment nodes available on [`whitequark/parser`](
https://github.com/whitequark/parser/blob/fbe0e8cbec557c96b0e0f4a8c5201155ee478284/lib/parser/tree_rewriter.rb#L80). Let me know if we only want a subset of that list.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/rescue_ensure_alignment.rb
+++ b/lib/rubocop/cop/layout/rescue_ensure_alignment.rb
@@ -30,7 +30,17 @@ module RuboCop
         ANCESTOR_TYPES = %i[kwbegin def defs class module].freeze
         RUBY_2_5_ANCESTOR_TYPES = (ANCESTOR_TYPES + %i[block]).freeze
         ANCESTOR_TYPES_WITH_ACCESS_MODIFIERS = %i[def defs].freeze
-        ASSIGNMENT_TYPES = %i[lvasgn].freeze
+        ASSIGNMENT_TYPES = %i[
+          lvasgn
+          ivasgn
+          cvasgn
+          gvasgn
+          casgn
+          masgn
+          op_asgn
+          and_asgn
+          or_asgn
+        ].freeze
 
         def on_resbody(node)
           check(node) unless modifier?(node)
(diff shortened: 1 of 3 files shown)


pocke:
Thank you for your pull request!


Probably this list also should include `masgn`, `op_asgn`, `or_asgn` and `and_asgn`. They are `a, b = x`, `a += x`, `a ||= x` and `a &&= x`. I'm not sure why the parser's list does not have these types, but I guess the cop should treat these types.



I think we need a helper method to check assignment type. For instance: `node.asgn?` or `assign_node?(node)`. But I cannot find it from our codebase. If someone knows the method, please tell me it :pray:

lib/rubocop/cop/layout/rescue_ensure_alignment.rb
```diff
@@ -30,7 +30,17 @@ class RescueEnsureAlignment < Cop
         ANCESTOR_TYPES = %i[kwbegin def defs class module].freeze
         RUBY_2_5_ANCESTOR_TYPES = (ANCESTOR_TYPES + %i[block]).freeze
         ANCESTOR_TYPES_WITH_ACCESS_MODIFIERS = %i[def defs].freeze
-        ASSIGNMENT_TYPES = %i[lvasgn].freeze
+        ASSIGNMENT_TYPES = %i[
```
Drenmi: Instead of this, you can use `Node::ASSIGNMENTS`, defined in `node.rb`. It's quite helpful to keep these kinds of lists to a single place. 🙂

marcotc: Thank you @Drenmi, will do!

marcotc: Done! Using `node.assignment?` now as it has the same list of types.

Drenmi:
Thank you @marcotc! 🙇

PR #6672, opened 2019-01-16
[Fix #6670] Fix a false positive for `Style/SafeNavigation`

### LINKED ISSUE #6670: Another wrong auto-correct by Style/SafeNavigation
rubocop 0.62.0 autocorrects
```ruby
foo.bar unless foo
```
into
```ruby
foo&.bar
```
which changes the behaviour of the code. (In fact, `foo.bar if foo` is auto-corrected to the very same result.)

Admittedly, the `foo.method unless foo` construct seems very rare because if effectively calls the method on `nil` or `false` only. (In our case, this wrong auto-correction actually revealed a long-standing bug). But rubocop should avoid changing the meaning of code.

koic:
Fixes #6670.

This PR fixes a false positive for `Style/SafeNavigation` when a method call safeguarded with a negative check for the object.

The following is a reproduction step.

```console
% rubocop -v
0.62.0

% cat example.rb
foo.bar unless foo

% rubocop --only Style/SafeNavigation
Inspecting 1 file
C

Offenses:

example.rb:1:1: C: Style/SafeNavigation: Use safe navigation (&.)
instead of checking if an object exists before calling the method.
foo.bar unless foo
^^^^^^^^^^^^^^^^^^

```

-----------------

* [x] Added tests.


bbatsov:
Good work!

PR #6676, opened 2019-01-16
[Fix #6387] Prevent `Lint/NumberConversion` from reporting error with `Time.now`

### LINKED ISSUE #6387: Lint/NumberConversion complains about Time.now.to_i
`Time.now.to_i` cannot and need not be converted to `Integer(Time.now, 10)` -- when I try that in an irb console, I get an `ArgumentError`: `base specified for non string value`

However, rubocop suggests I make that change:

--------

## Expected behavior

No rubocop error

## Actual behavior

$ rubocop test.rb  --only Lint/NumberConversion
Inspecting 1 file
W

Offenses:

test.rb:1:1: W: Lint/NumberConversion: Replace unsafe number conversion with number class parsing, instead of using Time.now.to_i, use stricter Integer(Time.now, 10).
Time.now.to_i
^^^^^^^^^^^^^

$ cat test.rb 
Time.now.to_i

## RuboCop version

```
$ rubocop -V
0.59.2 (using Parser 2.5.1.2, running on ruby 2.4.0 x86_64-linux)
```

tejasbubane:
Fixes: https://github.com/rubocop-hq/rubocop/issues/6387

* [x] Added tests.


tejasbubane:
Since `Time.now` is stdlib, I think this should be allowed by default. We can add a config option `AllowedReceivers` for adding other exceptional cases which are not ruby stdlib - but that would be separate PR.

lib/rubocop/cop/lint/number_conversion.rb
```diff
@@ -49,6 +51,10 @@ def on_send(node)
 
         private
 
+        def allowed_receivers
+          ['Time.now']
```
rrosenblum: Does the same thing happen with `Time.current` and other similar time methods?

This would be better off as a constant.

tejasbubane: Ok now that I think about it properly (rather than rushing onto a solution) - this happens for any function that returns a `Time` object - which is almost all methods [here](https://ruby-doc.org/core-2.6/Time.html) plus the ones added by ActiveSupport.

Maybe we can check for receiver's parent being Time? Fortunately `Date` and `DateTime` don't support `to_i`, `to_f` etc.

rrosenblum:
`DateTime` does support `to_i`

```ruby
[2] pry(main)> DateTime.now
=> Wed, 16 Jan 2019 18:34:11 +0000
[3] pry(main)> DateTime.now.to_i
=> 1547663653
```

> Maybe we can check for receiver's parent being `Time` class?

This sounds like a good solution to me. Make sure to include `DateTime` in that check.

It looks like the same issue happens with `ActiveSupport::Duration` as well

```ruby
[8] pry(main)> 1.second.to_i
=> 1
[9] pry(main)> Integer(1.second, 10)
ArgumentError: base specified for non string value
from (pry):9:in `Integer'
[10] pry(main)> 1.second.class
=> ActiveSupport::Duration
```

tejasbubane:
`to_i` on DateTime is [added by ActiveSupport](https://api.rubyonrails.org/classes/DateTime.html#method-i-to_i). I was checking with `irb`.

lib/rubocop/cop/lint/number_conversion.rb
```diff
@@ -49,6 +59,11 @@ def on_send(node)
 
         private
 
+        def date_time_object?(node)
+          time?(node) || date_time?(node) ||
+            time?(node.children[0]) || date_time?(node.children[0])
```
tejasbubane: @rrosenblum Is there a way to define matcher that will check this recursively for children? Currently this works for `Time.now.to_i` and `Time.zone.now.to_i` but will fail for `Time.zone.now.to_datetime.to_i`.

rrosenblum: This is always a bit tricky to handle. You can define a node matcher that can check a node's ancestors' pattern, but you still have to tell it exactly how to match so it doesn't work for checking both parent and grandparent. 

I added a method in `Style/SafeNavigationChain` to return the full method chain. This method should probably be moved to `SendNode` or somewhere similar so that it is more accessible.

```ruby
        def method_chain(node)    
          chain = node    
          while chain.send_type?    
            chain = chain.parent if chain.parent &&    
                                    %i[send csend].include?(chain.parent.type)    
            break                                                                                                                                  
          end           
          chain                     
        end
```

You can combine this with a `def_node_matcher` or `def_node_search` that can then check if the start of the chain matches your pattern. I normally wind up using `def_node_matcher` because it can return match groups. 

Let me know if you need anymore help with this.

tejasbubane: Thanks @rrosenblum! That helped.

tejasbubane:
Now `ActiveSupport::Duration` is one thing remaining which I am not really sure how to handle.

tejasbubane:
Rebased on latest master. Please review.

bbatsov:
Thanks!

PR #6680, opened 2019-01-16
[Fix #6648] EmptyLiteral - wrap arguments in parenthese when correcting Hash.new is the fist argument to super

### LINKED ISSUE #6648: Style/EmptyLiteral is (possibly) not safe
I've encountered a situation where this cop is possibly not safe.

I have a class with subclasses. The constructor for the parent class accepts a hash, and I want this to be an empty hash in one of the child classes.

```ruby
class Parent
  def initialize(hash)
    @data = hash
  end
end

class Child < Parent
  def initialize(string)
    super Hash.new
    @string = string
  end
end
```

When I autocorrect the child class it gets changed to:

```ruby
class Child < Parent
  def initialize(string)
    super {}
    @string = string
  end
end
```
Which is not equivalent. This results in the argument `string` being passed to the super call (i.e. it behaves as a 'bare' `super` call).  Changing this line to `super({})` causes it to work as expected.

--------

```
$ [bundle exec] rubocop -V
0.62.0 (using Parser 2.5.1.2, running on ruby 2.5.1 x86_64-linux)
```

rrosenblum:
This fixes #6648.


lib/rubocop/cop/style/empty_literal.rb
```diff
@@ -67,10 +67,13 @@ def string_literals_config
...
         def first_argument_unparenthesized?(node)
-          return false unless node.parent && node.parent.send_type?
+          parent = node.parent
+          unless parent && %i[send super zsuper].include?(parent.type)
+            return false
+          end
 
-          _receiver, _method_name, *args = *node.parent
-          node.object_id == args.first.object_id && !parentheses?(node.parent)
+          node.object_id == parent.arguments.first.object_id &&
```
Drenmi: I think we have been using `#equal?` for identity checks. (Perhaps a cop opportunity. 🙂)

Issue #6662, opened 2019-01-14
Load user-specific settings from XDG_CONFIG_HOME


FranklinYu:
**Is your feature request related to a problem? Please describe.**

I don’t like to clutter my home directory, and I only synchronize my `XDG_CONFIG_HOME`.

**Describe the solution you'd like**

If the file `~/.rubocop.yml` is missing, check whether environment variable `$XDG_CONFIG_HOME` is defined. If yes, then load from `$XDG_CONFIG_HOME/rubocop/config.yml`; otherwise load from `~/.config/rubocop/config.yml`.

**Describe alternatives you've considered**

N/A

**Additional context**

Latest standard specification: https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.8.html
Arch Linux Wiki: https://wiki.archlinux.org/index.php/XDG_Base_Directory

I can try PR if anyone is interested.

bbatsov:
PRs welcome!

FranklinYu:
I'm not familiar with the code base, but a simple search indicates that I should do it here:

https://github.com/rubocop-hq/rubocop/blob/19845704d148918eb9d7e8d00e1892fc7111854d/lib/rubocop/file_finder.rb#L43-L44

I'm still wondering whether `FileFinder` is used anywhere besides configuration loading? If so, modifying this file may affect other functions.

FranklinYu:
This is closed by #6895 IIUC. Thanks!

PR #6682, opened 2019-01-17
Load config from `~/rubocop/` and `$XDG_CONFIG_HOME`

tejasbubane:
if `$HOME` not set

Spec: https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.8.html

Closes: #6662

* [x] Added tests.

### DIFF
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -15,6 +15,7 @@ module RuboCop
   # directories are inspected.
   class ConfigLoader
     DOTFILE = '.rubocop.yml'.freeze
+    CONFIGFILE = 'rubocop/config.yml'.freeze
     RUBOCOP_HOME = File.realpath(File.join(File.dirname(__FILE__), '..', '..'))
     DEFAULT_FILE = File.join(RUBOCOP_HOME, 'config', 'default.yml')
     AUTO_GENERATED_FILE = '.rubocop_todo.yml'.freeze
@@ -75,7 +76,9 @@ module RuboCop
       # user's home directory is checked. If there's no .rubocop.yml
       # there either, the path to the default file is returned.
       def configuration_file_for(target_dir)
-        find_file_upwards(DOTFILE, target_dir, use_home: true) || DEFAULT_FILE
+        find_file_upwards(DOTFILE, target_dir, use_home: true) ||
+          find_file_upwards(CONFIGFILE, target_dir, use_home: true) ||
+          DEFAULT_FILE
       end
 
       def configuration_from_file(config_file)
--- a/lib/rubocop/file_finder.rb
+++ b/lib/rubocop/file_finder.rb
@@ -31,17 +31,26 @@ module RuboCop
     private
 
     def traverse_files_upwards(filename, start_dir, use_home)
+      traverse_file(filename, start_dir).each { |f| yield(f.to_s) }
+      return unless use_home && (ENV.key?('HOME') ||
+                                 ENV.key?('XDG_CONFIG_HOME'))
+
+      xdg_file = File.join(ENV['XDG_CONFIG_HOME'].to_s, filename)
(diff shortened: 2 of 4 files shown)


Mange:
In which context would `$HOME` not be set but a `$XDG_CONFIG_HOME` is set?

I went here to see if there was an issue for XDG support already. My take on it would be this:

- Check for `$XDG_CONFIG_HOME/rubocop/config.yml` and load it if it exists, else go to next step
- Check for `$HOME/.rubocop.yml` and load it if it exits.

Then existing users of the dotfile would not see any breakages, but XDG-focused users could move their existing file to the correct location and have it load automatically.

tejasbubane:
@Mange Gotcha, I did it the other way round. You are right, I will change the implementation. Thanks!

tejasbubane:
@Mange Fixed. I made some changes, this is the order being checked now:

```
RuboCop::ConfigLoader
  .configuration_file_for
    when a config file exists in the parent directory
      returns the path to that configuration file
    when multiple config files exist in ancestor directories
      prefers closer config file
    when no config file exists in ancestor directories
      ENV has `XDG_HOME` defined
        config file exists in `XDG_HOME` directory
          returns path to the file in `XDG_HOME` directory
        no config file in `XDG_HOME` but in rubocop directory
          returns path to the file in XDG_HOME/rubocop directory
      when ENV does not have `XDG_HOME` defined
        and a config file exists in home directory
          returns the path to the file in home directory
        and a config file exists in rubocop directory
          returns the path to the file in home/rubocop directory
      and no config file exists in `XDG_HOME` or `HOME` directories
        falls back to the provided default file
```

Mange:
I cannot find any references to a `XDG_HOME` directory in the specification: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

Is it from some other specification, or is it a mistake? Are you mixing up XDG_CONFIG_HOME with HOME?

Normal config path according to the specification would be `$XDG_CONFIG_HOME/rubocop/config.yml`, aka `$HOME/.config/rubocop/config.yml` (if XDG_CONFIG_HOME isn't defined).

tejasbubane:
@Mange That was indeed a mistake. I changed it to `XDG_CONFIG_HOME`.

Mange:
I think this looks good now overall. I hope the maintainers agree.

lib/rubocop/file_finder.rb
```diff
@@ -31,17 +31,26 @@ def find_files_upwards(filename, start_dir, use_home: false)
     private
 
     def traverse_files_upwards(filename, start_dir, use_home)
+      traverse_file(filename, start_dir).each { |f| yield(f.to_s) }
+      return unless use_home && (ENV.key?('HOME') ||
```
bbatsov: Seems to me that we're mixing different concerns in this method. Ideally the file traversal should be separate from checks for which directory contains the config file.

lib/rubocop/config_loader.rb
```diff
@@ -15,6 +15,7 @@ class ConfigNotFoundError < Error
   # directories are inspected.
   class ConfigLoader
     DOTFILE = '.rubocop.yml'.freeze
+    CONFIGFILE = 'rubocop/config.yml'.freeze
```
bbatsov: Probably this should be `.rubocop`, not `rubocop`. Usually config dirs are "hidden". I'd also name the constant `CONFIG_FILE`.

bbatsov:
I left a few small comments. Generally I'm fine with the proposed solution.

lib/rubocop/config_loader.rb
Mange: Files in XDG_CONFIG_HOME should not be hidden according to specs. The most common convention is `appname/config`.

bbatsov:

Mange:
I re-opened this in #6895.

PR #6683, opened 2019-01-17
Fix super edge cases for omit_parentheses style of Style/MethodCallWithArgsParentheses

gsamokovarov:
1. Calls with arguments as `super` arguments.
```ruby
super foo(bar)
         ^^^^^ Omit parentheses for method calls with arguments.
```

2. Braced `super` calls.
```ruby
super(foo(bar)) { yield }
     ^^^^^^^^^^ Omit parentheses for method calls with arguments.
```


Drenmi:
Gotta catch 'em all, @gsamokovarov! 😉 Thank you!

Issue #6655, opened 2019-01-11
Node Pattern Search as an CLI to search code


jonatas:
During my early days implementing new cops, I created [rubocopular search](https://github.com/jonatas/rubocopular/blob/master/bin/search) to allow me to understand if my node patterns are clear and matching with the nodes that I expected.

I'm curious how other developers are dealing with it and validating their node patterns. 

The feature would be we have some `rubocop-search` or something that we can just enter with a node pattern and grep in all the code for it.

I'd like to know if the team has the interest to put it in the core. I can provide the PR for the feature.

ps: I'm actively working on [fast](https://github.com/jonatas/fast) that is a tool only with this purpose but my node pattern implementation is not complete and fully functional as the RuboCop one.

bbatsov:
I think a good first step would be to extra the pattern matching library into its own gem, perhaps our note extensions as well. We discussed this back in the day, but we never made any progress. 

I'm not strongly opposed to the idea of adding more scripts in the main gem, but it seems to me that something that searching for node patterns in somewhat orthogonal to the main purpose of RuboCop.

Darhazer:
It could be useful if there is a tool that searches for a given pattern, extract the line that matches, run a given cop on that line and aggregates report with:
* lines matching that were flagged by the cop (in order to inspect for false positives)
* lines matching that were not flagged (in order to inspect for false negatives)
removing any duplicates in the process.

Drenmi:
> I think a good first step would be to extra the pattern matching library into its own gem, perhaps our note extensions as well. We discussed this back in the day, but we never made any progress.

I was thinking about this as well, while working on the node extensions. Something like `parser-tools`.

jonatas:
Great! I started working on it ☝️ 

Let me know if the name is good or we can run a poll to find the proper name. I personally like NodePattern but I also thought about the following names:

* AST extensions
* AST grep (reminds regular expressions or grep)
* Node Extensions
* Node Grep

Drenmi:
> Let me know if the name is good or we can run a poll to find the proper name.

The gem contains more than one part (node patterns and node extensions), so I'm hesitant to name it after only one of those parts.

koic:
Humm, I'm not sure if it is good to do Gemify. I think that it is API depending on the use of RuboCop. Extracting from RuboCop makes code tracking inconvenient. I don't have confidence that the gain is larger than pain.

[...]

PR #6686, opened 2019-01-18
Extract NodePattern to a separated gem

jonatas:
I started the work related to #6655.

The gem extracted is under my namespace and I can move it soon some maintainer transfer it.

https://github.com/jonatas/rubocop-node_pattern

The name of the gem is also in the TODO  list to be discussed and probably we can find a better name since all AST utilities are going to the repo too.

## TODO

* [ ] Move the related project to `rubocop-hq` namespace
* [ ] Update the Gemfile to `rubocop-hq` namespace 
* [ ] Move NodePattern implementation to `NodePattern::Parser` or something to make the gem clear and avoid [requires in the bottom](https://github.com/jonatas/rubocop-node_pattern/blob/master/lib/rubocop/node_pattern.rb#L567)
* [ ] Define a better name to the gem?
* [ ] Setup RuboCop config for the project

* [x] Added tests.

## Notes

I'm not sure how can we maintain the ownership of the current git files that are moving to another repo. Is it something common or we just move and have a new author?


Gemfile
```diff
@@ -14,6 +14,9 @@ gem 'simplecov', '~> 0.10'
 gem 'test-queue'
 gem 'yard', '~> 0.9'
 
+gem 'rubocop-node_pattern',
+    git: 'git://github.com/jonatas/rubocop-node_pattern.git'
```
jonatas: Soon we migrate the gem to the proper repo, we can also publish as a gem and remove the git reference.

wevtimoteo: Did you consider keeping git history using filter-branch feature for those files?

https://help.github.com/articles/splitting-a-subfolder-out-into-a-new-repository/

jonatas: That  is pretty cool @wevtimoteo, I'll use the tool to keep the git history intact 👍

bbatsov:
Keep in mind that the node extensions and the note pattern matcher are somewhat orthogonal - you don’t necessary always need the two of them. This means that potentially we can have an `astx` gem and some other gem about the pattern matcher. On the other hand this would mean more overhead from a maintenance perspective, so perhaps a single gem would do. I also think it might be convenient to simply have those 2 or 3 gems in the same repo (RuboCop’s) for simplicity’s sake. We will have to think this over very carefully.

jonatas:
Got it @bbatsov. Looking at the code, we can see `node_pattern` depends on `astx`. We can do it in two gems or only one.

@baweaver let me know how can we schedule a pairing session or work together on it. I'm open to both cases. We can go with one or two gems. I'd start with one and we can split the external gem later if needed.

Gemfile
wevtimoteo: If you need some help with that I'll happy to pair with you!

pirj:
With the movement of extracting parts of RuboCop out of it (e.g. `rubocop-performance`), while keeping those extracted repositories first-class citizen of RuboCopHQ, it makes sense to think about how a change to main `rubocop`, and specifically to its node pattern repository would affect the dependent repositories.

An approach to consider is [the one taken by RSpec]( https://github.com/rspec/rspec-dev#rspec-dev), when all the repositories are checked out, and before submitting a change one would check if the dependent officially supported gems are not broken.
```
# rspec-expectations' Gemfile
branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp # `master`, or some cross-repository branch
%w[rspec rspec-core rspec-mocks rspec-support].each do |lib|
  library_path = File.expand_path("../../#{lib}", __FILE__)
  if File.exist?(library_path) && !ENV['USE_GIT_REPOS']
    gem lib, :path => library_path
  else
    gem lib, :git => "https://github.com/rspec/#{lib}.git", :branch => branch
  end
end
```

With that modular yet atomically safe approach, it doesn't matter that much anymore if `rubocop-node_pattern` is in the same repository or not, and breakages can be spotted early, during local development. That also kind of simplifies the changes that need to be made in several repositories at once to fix a bug/introduce a feature.

WDYT?

bbatsov:
Yeah, that's a reasonable approach. We should first finish the isolation/extraction of something though. :-) Right now everything's work in progress...

@jonas054 Any updates here?

jonas054:
@bbatsov It's very nice to hear from you again! :smile: But did you in this case mean @jonatas?

bbatsov:
Ops! Sorry about that! I totally meant to ping @jonatas, but I love you so much that I type your name every time I start typing `@jona...`. 😄

jonatas:
I like the idea and I guess we can try to follow the approach @pirj suggested.

The hardest part is that we have `ast/node` that is a dependency for `node_pattern`. I'll try to pair with @pirj tomorrow and see if we can handle it.

bbatsov:
Sounds good. Ping me if you need my help with anyone.

jonatas:
Thanks, @bbatsov. Should we keep the same module name for now or already go with `astx`?

Gemfile
wevtimoteo: Done in https://github.com/jonatas/rubocop-astx/tree/move-ast-keeping-history

bbatsov:
I guess for now it can be the same module name.

marcandre:
Let me know if you need help...

jonatas:
I'm not having the proper time to finish it. I'll close it for now and revisit in a near future.

jonatas:
@marcandre feel free to ping me I can try to pair with you and help to review it.

PR #4702, opened 2017-09-05
Add new `Lint/UriEscapeUnescape` cop

koic:
## Feature

This cop identifies places where `URI.escape` can be replaced by `CGI.escape`, `URI.encode_www_form` or `URI.encode_www_form_component` depending on your specific use case.
Also this cop identifies places where `URI.unescape` can be replaced by `CGI.unescape`, `URI.decode_www_form` or `URI.decode_www_form_component` depending on your specific use case.

```console
% cat /tmp/example.rb
# frozen_string_literal: true

require 'uri'

URI.escape('http://example.com')
URI.encode('http://example.com')
URI.unescape('http://example.com')
URI.decode('http://example.com')
```

```console
% bundle exec rubocop /tmp/example.rb
Inspecting 1 file
W

Offenses:

/tmp/example.rb:5:1: W: URI.escape method is obsolete and should not be used. Instead, use CGI.escape, URI.encode_www_form or URI.encode_www_form_component depending on your specific use case.
URI.escape('http://example.com')
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/tmp/example.rb:6:1: W: URI.encode method is obsolete and should not be used. Instead, use CGI.escape, URI.encode_www_form or URI.encode_www_form_component depending on your specific use case.
URI.encode('http://example.com')
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/tmp/example.rb:7:1: W: URI.unescape method is obsolete and should not be used. Instead, use CGI.unescape, URI.decode_www_form or URI.decode_www_form_component depending on your specific use case.
URI.unescape('http://example.com')
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/tmp/example.rb:8:1: W: URI.decode method is obsolete and should not be used. Instead, use CGI.unescape, URI.decode_www_form or URI.decode_www_form_component depending on your specific use case.
URI.decode('http://example.com')
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 4 offenses detected
```

## Target Problem

Emulate the following Ruby's warnings.

```console
% ruby -w -ruri -e "URI.escape('http://example.com')"
-e:1:in `<main>': warning: URI.escape is obsolete

% ruby -w -ruri -e "URI.encode('http://example.com')"
-e:1:in `<main>': warning: URI.escape is obsolete

% ruby -w -ruri -e "URI.unescape('http://example.com')"
[...]

PR #6687, opened 2019-01-18
update helper comment url encode

bauerjon:
A change related to https://github.com/rubocop-hq/rubocop/issues/6610.

This change updates the helper output to include `ERB::Util.url_encode` as another suggestion for the deprecated `URI.encode` and `URI.escape`.

-----------------

* [x] Added tests.
* [x] Added an entry to the [Changelog](../blob/g) if the new code introduces user-observable changes. See [changelog entry format](../blob/master/CONTRIBUTING.md#changelog-entry-format).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#6610](https://github.com/rubocop-hq/rubocop/issues/6610): Extend `Lint/UriEscapeUnescape` to include ERB::Util.url_encode in suggested alternatives. ([@bauerjon][])
+
 ### Bug fixes
 
 * [#6678](https://github.com/rubocop-hq/rubocop/issues/6678): Fix `Lint/DisjunctiveAssignmentInConstructor` when it finds an empty constructor. ([@rmm5t][])
@@ -3763,3 +3765,4 @@
 [@Ruffeng]: https://github.com/Ruffeng
 [@roooodcastro]: https://github.com/roooodcastro
 [@rmm5t]: https://github.com/rmm5t
+[@bauerjon]: https://github.com/bauerjon
--- a/lib/rubocop/cop/lint/uri_escape_unescape.rb
+++ b/lib/rubocop/cop/lint/uri_escape_unescape.rb
@@ -4,8 +4,8 @@ module RuboCop
   module Cop
     module Lint
       # This cop identifies places where `URI.escape` can be replaced by
-      # `CGI.escape`, `URI.encode_www_form`, or `URI.encode_www_form_component`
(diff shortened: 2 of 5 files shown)


lib/rubocop/cop/lint/uri_escape_unescape.rb
```diff
@@ -53,23 +55,31 @@ class UriEscapeUnescape < Cop
...
-                             ALTERNATE_METHODS_OF_URI_UNESCAPE
-                           end
-
             double_colon = top_level ? '::' : ''
 
             message = format(
               MSG, uri_method: "#{double_colon}URI.#{obsolete_method}",
-                   replacements: "`#{replacements[0]}`, `#{replacements[1]}` " \
-                                 "or `#{replacements[2]}`"
+                   replacements: replacements_string(obsolete_method)
```
bauerjon: happy to change this if desired. we could do something similar to some of the suggestions here https://codereview.stackexchange.com/questions/5863/ruby-function-to-join-array-with-commas-and-a-conjunction/5890

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#6610](https://github.com/rubocop-hq/rubocop/issues/6610): Extend `Lint/UriEscapeUnescape` to include ERB::Util.url_encode in suggested alternatives. ([@bauerjon][])
```
Drenmi: This doesn't really require a CHANGELOG entry. 🙂

Drenmi: This doesn't really require a CHANGELOG entry. 🙂

Drenmi:
Looks reasonable to me. 🚀

koic:
I wrote an API in the document of `Lint/UriEscapeUnescape` cop that officially suggested alternatives at #4702.

The following official API document of `CGI.escape` and `CGI.unescape`.

> This method is obsolete and should not be used. Instead, use CGI.escape, URI.encode_www_form or URI.encode_www_form_component depending on your specific use case.

- https://ruby-doc.org/stdlib-2.6/libdoc/uri/rdoc/URI/Escape.html#method-i-escape
- https://ruby-doc.org/stdlib-2.6/libdoc/uri/rdoc/URI/Escape.html#method-i-unescape

I'm not sure if it is better to present alternatives that are not written in upstream.

The following is upstream. If you open a PR and accept it, the value will be more clear :-)
https://github.com/ruby/ruby/blob/v2_6_0/lib/uri/common.rb#L80-L86

bauerjon:
Probs not gonna carry this across the finish line. If someone wants to reopen and do so feel free!

PR #6702, opened 2019-01-23
[Fix #6701] Trailing comma detection regression with HEREDOC

### LINKED ISSUE #6701: TrailingComma false positives in heredoc
## Expected behavior

I've found two cases where commas inside HEREDOC is tripping up the TrailingComma module. I've written spec tests that highlight the issue(s) below but not sure what the best solution is going to be. 

I'm wondering if a better solution for handling HEREDOC would be to strip it out or replace it with a regular string before trying to do the comma detection?

## Actual behavior

The specs below fail :(

The following example specs fail:

In spec/rubocop/cop/style/trailing_comma_in_array_literal_spec.rb
```
expect_no_offenses(<<-RUBY.strip_indent)
  [
    <<-TEXT, 123
      Something with a , in it  # <-- Fails on this comma
    TEXT
  ]
RUBY
```
In this case it appears to be because the HEREDOC isn't the last parameter in the array

In spec/rubocop/cop/style/trailing_comma_in_arguments_spec.rb
```
expect_no_offenses(<<-RUBY.strip_indent)
  some_method(
    <<-LOREM.delete("\\n")
      Something with a , in it # <-- Fails on this comma
    LOREM
  )
RUBY
```
In this case, it seems to be the `delete("\n")` that is tripping up the HEREDOC detection

## RuboCop version

RuboCop 0.63.1  (current master)
https://github.com/rubocop-hq/rubocop/commit/42351e6d5ae73654a2318530a3fcc3861e94fa80

abrom:
The TrailingCommaInArguments and TrailingCommaInArrayLiteral cops have a false positive when the parent node contains HEREDOC with commas in them. The issue appears to be due to lax comma detection.

Instead of only matching the last item for heredoc, match all items and search for optional whitespace followed by trailing commas, excluding newlines in the match when heredoc is present. 

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * [#6382](https://github.com/rubocop-hq/rubocop/issues/6382): Fix `Layout/IndentationWidth` with `Layout/EndAlignment` set to start_of_line. ([@dischorde][], [@siegfault][], [@mhelmetag][])
 * [#6710](https://github.com/rubocop-hq/rubocop/issues/6710): Fix `Naming/MemoizedInstanceVariableName` on method starts with underscore. ([@pocke][])
 * [#6722](https://github.com/rubocop-hq/rubocop/issues/6722): Fix an error for `Style/OneLineConditional` when `then` branch has no body. ([@koic][])
+* [#6702](https://github.com/rubocop-hq/rubocop/pull/6702): Fix `TrailingComma` regression where heredoc with commas caused false positives. ([@abrom][])
 
 ### Changes
 
--- a/lib/rubocop/cop/mixin/trailing_comma.rb
+++ b/lib/rubocop/cop/mixin/trailing_comma.rb
@@ -17,11 +17,13 @@ module RuboCop
       end
 
       def check(node, items, kind, begin_pos, end_pos)
-        return if heredoc?(items.last)
-
         after_last_item = range_between(begin_pos, end_pos)
 
-        comma_offset = after_last_item.source =~ /,/
+        # If there is any heredoc in items, then match the comma succeeding
+        # any whitespace (except newlines), otherwise allow for newlines
+        comma_regex = any_heredoc?(items) ? /\A[^\S\n]*,/ : /\A\s*,/
+        comma_offset = after_last_item.source =~ comma_regex &&
+                       after_last_item.source.index(',')
 
         if comma_offset && !inside_comment?(after_last_item, comma_offset)
           check_comma(node, kind, after_last_item.begin_pos + comma_offset)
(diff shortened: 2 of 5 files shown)


abrom:
This solution still has issues if the HEREDOC starts with a comma

abrom:
I've tried to solve this problem a different way. Instead of skipping the trailing comma check if the *last* item is heredoc, we look to see if any item in the node is heredoc and if so we check for a trailing comma directly preceding the last item source (with only horizontal whitespace). If there is no heredoc then we allow horizontal and vertical whitespace before the comma.

This did raise some potential issues with the existing specs, specifically those titled
```
accepts missing comma after a heredoc
```

The comment at the top of these specs is:
```
# A heredoc that's the last item in a literal or parameter list can not
# have a trailing comma. It's a syntax error.
```

Although this isn't true?? 

```
some_method(
  <<-FOO,
    ...
  FOO
)
```
is valid syntax?! 

As is
```
[
  <<-FOO,
    ...
  FOO
]
```

abrom:
...as such, my belief is that the specs failing on this branch are actually incorrect. If others agree, I'll modify those specs as needed

Drenmi:
@abrom Since there have been no objections to this, I'm ready to merge it in. Please rebase on latest `master`, and make sure the CHANGELOG is merged correctly, then I will accept the PR. 🙇

abrom:
Done 👍

wilg:
It seems like after this PR if I have:

```
Style/TrailingCommaInArguments:
  Enabled: true
  EnforcedStyleForMultiline: comma
```

(or the other trailing comma cops), it forces a trailing comma in the heredoc, like in the test cases:

```
[
  <<-FOO,
    ...
  FOO
]
```

This ends up looking really confusing and bizarre to me. It's particularly bizarre when a one-argument function that takes a multiline heredoc is required to have a comma. Is this intended behavior? Would it make sense to have an option to disable it for heredocs?

abrom:
But that's the expected result with `EnforcedStyleForMultiline: comma` and your array being defined across multiple lines?!

wilg:
I think my example was poor, it makes more sense for array/hash literals, but less for:

```
my_func(
 <<-FOO,
    ...
  FOO
)
```

abrom:
But that's exactly what `EnforcedStyleForMultiline: comma` is expected to do there too. If you wanted to have the cop configured as such, but not have the trailing comma you'd have something like:
```
my_func(<<-FOO)  # <-- the method definition is now on one line
  ...
FOO
```

wilg:
Yep, I think you're right. The codebase I'm looking at just uses heredocs and multiline comma enforcement in a way that makes things look and feel weird. I'll open an issue if I can figure out a way to have my cake and eat it too. Thanks!

PR #6704, opened 2019-01-23
Added ReflectionClassNameString cop

Bhacaz:
This PR add a cop to check if the value of the option `class_name` of a [ActiveRecord::Reflection](https://apidock.com/rails/ActiveRecord/Reflection/AbstractReflection/class_name) is a string.

_Why this cop:_

I was working in a Rails project where often there was a constant used for `class_name` value. That caused a issue which was to preload unnecessary models use in the definition of a reflection.

-----------------

* [x] Added tests.


lib/rubocop/cop/rails/reflection_class_name_string.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Rails
+      # @example
```
koic: I think that this cop makes sense. Please write a description about the cop.

koic:
FYI, since Rails 5.2 app if a string isn't used, the following error occurs.

```console
/Users/koic/.rbenv/versions/2.6.0/lib/ruby/gems/2.6.0/gems/activerecord-5.2.2/lib/active_record/reflection.rb:436:in `initialize': A class was passed to `:class_name` but we are expecting a string. (ArgumentError)
```

https://github.com/rails/rails/commit/e65aff70696be52b46ebe57207ebd8bb2cfcdbb6

Having auto-correct in the future will be useful when migrating from old Rails to new Rails.

lib/rubocop/cop/rails/reflection_class_name_string.rb
```diff
@@ -0,0 +1,38 @@
...
+      # the definition of a reflection is a string.
+      #
+      # @example
+      #   # bad
+      #   has_many :accounts, class_name: Account
+      #   has_many :accounts, class_name: Account.name
+      #
+      #   # good
+      #   has_many :accounts, class_name: 'Account'
+      class ReflectionClassNameString < Cop
```
koic: Like [`Rails/ActionFilter` cop](https://github.com/rubocop-hq/rubocop/blob/v0.63.1/lib/rubocop/cop/rails/action_filter.rb), I think the name `Rails/ReflectionClassName` is simple and clear.
The name inspired by original proposal name and https://github.com/rails/rails/pull/27551.

lib/rubocop/cop/rails/reflection_class_name_string.rb
```diff
@@ -0,0 +1,38 @@
...
+        MSG = 'Use a string has value for a class_name.'.freeze
+
+        def_node_matcher :association_with_options?, <<-PATTERN
+          (send nil? {:has_many :has_one :belongs_to} _ (hash $...))
+        PATTERN
+
+        def_node_search :class_name_string, <<-PATTERN
```
koic: Does this mean `:reflection_class_name` node pattern?

lib/rubocop/cop/rails/reflection_class_name_string.rb
Bhacaz: You're right, `Rails/ReflectionClassName` is simple and clear enough.

lib/rubocop/cop/rails/reflection_class_name_string.rb
Bhacaz: This pattern match for the key (`class_name`) and his value of a hash. The reflection matcher itself is defined [here](https://github.com/rubocop-hq/rubocop/pull/6704/files/7693b518d72dc4b8f7e6584f0b119b0db3d203e1#diff-21df1c369efb890c11c7d6ecba1911f0R19).

lib/rubocop/cop/rails/reflection_class_name_string.rb
Bhacaz: And thanks for those links. 😄

lib/rubocop.rb
```diff
@@ -592,6 +592,7 @@
 require_relative 'rubocop/cop/rails/present'
 require_relative 'rubocop/cop/rails/read_write_attribute'
 require_relative 'rubocop/cop/rails/redundant_receiver_in_with_options'
+require_relative 'rubocop/cop/rails/reflection_class_name_string'
```
koic: Thank you for renewing the cop name. Please change the file name according to cop class name.

```diff
-require_relative 'rubocop/cop/rails/reflection_class_name_string'
+require_relative 'rubocop/cop/rails/reflection_class_name'
```

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## master (unreleased)
 
 ### New features
-
```
koic: Keeping the one blank line will be consistent in CHANGELOG.md format.

lib/rubocop/cop/rails/reflection_class_name_string.rb
koic: This s-expression pattern seems to match a pattern using constants like `class_name: Account`, for example. This represents a case where the value is not a string (`class_name: 'Account'`).
In other words it seems to match a reflection class name other than a string class name.

Bhacaz: I see what you mean. It's true that the name of the pattern is confusing. I can change it for something like `class_name_not_string`. 

Suggestions are welcome.

koic: I like `reflaction_class_name` as I think of it.

The current code base change when using this is as follows.

```diff
% g diff
diff --git a/lib/rubocop/cop/rails/reflection_class_name_string.rb b/lib/rubocop/cop/rails/reflection_class_name_string.rb
index c3dfa96c5..542a89f7e 100644
--- a/lib/rubocop/cop/rails/reflection_class_name_string.rb
+++ b/lib/rubocop/cop/rails/reflection_class_name_string.rb
@@ -20,17 +20,17 @@ module RuboCop
           (send nil? {:has_many :has_one :belongs_to} _ (hash $...))
         PATTERN

-        def_node_search :class_name_string, <<-PATTERN
+        def_node_search :reflection_class_name, <<-PATTERN
           (pair (sym :class_name) !str)
         PATTERN

         def on_send(node)
-          return unless association_with_options? node
+          return unless association_with_options?(node)

-          class_name_node = class_name_string(node).first
-          return unless class_name_node
+          reflection_class_name = reflection_class_name(node).first
+          return unless reflection_class_name

-          add_offense(node, location: class_name_node.loc.expression)
+          add_offense(node, location: reflection_class_name.loc.expression)
         end
       end
     end
```

"`return unless reflaction_class_name`" can represent the intent to reach `add_offense` when a node matches the reflaction class name.

koic:

koic:
If you do not mind, could you please update the body of commit message with the description of the PR that you wrote using `git commit --amend`?

After merging this PR I'd like to forward porting it to the RuboCop Rails repository (Rails cops are extracted in the near future to this repository). Moving the repository will weaken the linkage with PR, so it will be helpful if details are written in the body of commit message.

bquorning:
Using e.g. `Account.name` also returns a string, *and* protects against typos. I much prefer that style. Could the cop be made configurable?

Bhacaz:
Using `Account.name` will force load the class, instance of lazy loading it, which was the original reason to deprecate passing a class since Rails 5.2.  [rails/rails#27551](https://github.com/rails/rails/pull/27551)

ryym:
Shouldn't this cop allow a symbol value as well as a string?
`class_name: :Account`

Bhacaz:
@ryym We can discuss about that (maybe in an issue), but using a symbol can quickly have its limits when using module.

Example: `class_name: :Account::Admin` will not work.

cheerfulstoic:
I just came here to ask the same a @ryym 

For your example you could use `class_name: :'Account::Admin'`.

The advice that I've generally heard is that symbols are better for memory usage than strings when you can use them

Bhacaz:
@cheerfulstoic You're right that can works. But [Reflection#class_name](https://github.com/rails/rails/blob/6127b8d9203d7ce852d103ff1686621f9aca292f/activerecord/lib/active_record/reflection.rb#L166) call `.to_s` on the value of `class_name`, so it will create a new `String` object unless it's already a string. I don't know what it's the best in term of memory but while using a symbol there will be one more object.

In the following example there will be 1 symbol and 2 strings (when using `.to_s`)
```ruby
# ActiveRecord #1
has_one :account, class_name: :Account

# ActiveRecord #2
has_many :account, class_name: :Account
```

OR

2 strings
```ruby
# ActiveRecord #1
has_one :account, class_name: 'Account'

# ActiveRecord #2
has_many :account, class_name: 'Account'
```

Proof that `String#to_s` return the same object.
```ruby
options = { class_name: 'Account' }
p options[:class_name].object_id # => 70101446453420
@class_name = options[:class_name].to_s
p @class_name.object_id # => 70101446453420

options = { class_name: :Account }
p options[:class_name].object_id # => 6419108
@class_name2 = options[:class_name].to_s
p @class_name2.object_id # => 2347999460
```

_Correct me if there something wrong_

ryym:
> using a symbol can quickly have its limits when using module

> [Reflection#class_name](https://github.com/rails/rails/blob/6127b8d9203d7ce852d103ff1686621f9aca292f/activerecord/lib/active_record/reflection.rb#L166) call `.to_s` on the value of `class_name`

@Bhacaz I see. I didn't notice them. Thanks!

cheerfulstoic:
I didn't know that it called `to_s` behind the scenes, but I suppose it makes sense.  I guess I would have hoped that it went directly from a string/symbol to storing on the association object the class.  But maybe it can't always do that

:+1:

PR #6351, opened 2018-10-03
Fix a false positive for `Layout/ClosingParenthesisIndentation`

antonzaytsev:
This PR fixes a false negative for `Layout/ClosingParenthesisIndentation` when first argument is multiline.

The following is a reproduction step.
```bash
% rubocop -V
0.59.2 (using Parser 2.5.1.2, running on ruby 2.5.0 x86_64-darwin17)

% cat app/models/users.rb
class User < ApplicationRecord
  def self.complex_find
    where(
      "users.approved_at < ? OR
       users.approved_at IS NULL", 3.days.ago
    )
  end
end

% rubocop app/models/users.rb --only Layout/ClosingParenthesisIndentation -d
or /Users/anton/code/rubocop: configuration from /Users/anton/code/rubocop/.rubocop.yml
configuration from /Users/anton/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/rubocop-rspec-1.29.1/config/default.yml
configuration from /Users/anton/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/rubocop-rspec-1.29.1/config/default.yml
Default configuration from /Users/anton/code/rubocop/config/default.yml
Inheriting configuration from /Users/anton/code/rubocop/.rubocop_todo.yml
Inspecting 1 file
Scanning /Users/anton/code/rubocop/users.rb
C

Offenses:

users.rb:6:5: C: Layout/ClosingParenthesisIndentation: Indent ) to column 5 (not 4)
    )
    ^

1 file inspected, 1 offense detected
Finished in 0.5669199999974808 seconds
```

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](../blob/master/CONTRIBUTING.md#changelog-entry-format).
* [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.
* [x] Run `rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: http://chris.beams.io/posts/git-commit/

PR #6712, opened 2019-01-27
Fix a bug in Layout/ClosingParenthesisIndentation

bquorning:
ClosingParenthesisIndentation would complain about
```rb
block_adds_extra_indentation do
  some_method(a,
    x: 1,
    y: 2
  )
end
```
saying it should be indented as:
```
block_adds_extra_indentation do
  some_method(a,
    x: 1,
    y: 2
)
end
```
This issue appeared after the bugfix commit 16531120c0a05d824c2ff1bc6ccb2bf070ad3f89 in #6351.

* I did not add an entry to the Changelog because the issue I’m fixing hasn’t been released yet.
* The first commit in the PR is somewhat unrelated (moving one spec example to another section of the spec file), but I found it too small a change for a separate PR.

/cc @antonzaytsev 

-----------------

* [x] Added tests.


lib/rubocop/cop/layout/closing_parenthesis_indentation.rb
```diff
@@ -142,15 +142,17 @@ def check_for_no_elements(node)
...
 
         def expected_column(left_paren, elements)
-          if !line_break_after_left_paren?(left_paren, elements) &&
-             all_elements_aligned?(elements)
-            left_paren.column
-          else
+          if line_break_after_left_paren?(left_paren, elements)
             source_indent = processed_source
                             .line_indentation(first_argument_line(elements))
+
```
bquorning: Why did I add a blank line here? I’ll remove it as push again.

antonzaytsev:
@bquorning thank you for fix and new test case to prevent such issue 👍

PR #6713, opened 2019-01-28
Fix error message for `Performance/ChainArrayAllocation` cop

### LINKED ISSUE #6684: Performance/ChainArrayAllocation -> map!, flatten!, and compact! return nil if unmodified
The [Performance/ChainArrayAllocation cop](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/performance/chain_array_allocation.rb) suggests `map!`, `flatten!`, and `compact!` over their non-mutating counterparts when following another new array allocation. However, each of these methods returns `nil` if they don't modify the array, rather than returning the array itself. This seems unsafe.

--------

## Expected behavior

I expect it to suggest performance improvements that don't change behavior. In this case, the non-mutating methods always return the resulting array.

## Actual behavior

Following the cop's suggestions might return the modified array, but might return `nil` if the array was not modified. See documentation for the methods, e.g. here: https://ruby-doc.org/core-2.5.1/Array.html#method-i-uniq-21

I don't remember if this behavior was different at some point, but not recently.

```
[1,2].map{|i|i}.uniq
=> [1,2]
[1,2].map{|i|i}.uniq!
=> nil
```

## RuboCop version

```
rubocop -V
0.63.0 (using Parser 2.6.0.0, running on ruby 2.5.0 x86_64-darwin17)
```

## THANK YOU FOR ALL YOUR HARD WORK 🚀 🦄 💖 
I can submit a PR if you'd like, haven't gotten to contribute here before.

tejasbubane:
Closes #6684

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/performance/chain_array_allocation.rb
+++ b/lib/rubocop/cop/performance/chain_array_allocation.rb
@@ -47,8 +47,9 @@ module RuboCop
         HAS_MUTATION_ALTERNATIVE = ':collect :compact :flatten :map :reject '\
                                    ':reverse :rotate :select :shuffle :sort '\
                                    ':uniq '.freeze
-        MSG = 'Use `%<method>s...%<second_method>s!` instead of `%<method>s' \
-              '...%<second_method>s`.'.freeze
+        MSG = 'Use unchained `%<method>s!` and `%<second_method>s!` '\
+              '(followed by `return array` if required) instead of chaining '\
+              '`%<method>s...%<second_method>s`.'.freeze
 
         def_node_matcher :flat_map_candidate?, <<-PATTERN
           {
--- a/spec/rubocop/cop/performance/chain_array_allocation_spec.rb
+++ b/spec/rubocop/cop/performance/chain_array_allocation_spec.rb
@@ -4,8 +4,9 @@ RSpec.describe RuboCop::Cop::Performance::ChainArrayAllocation, :config do
   subject(:cop) { described_class.new(config) }
 
   def generate_message(method_one, method_two)
-    "Use `#{method_one}...#{method_two}!` instead of `#{method_one}"\
-    "...#{method_two}`."
+    "Use unchained `#{method_one}!` and `#{method_two}!` "\
+    '(followed by `return array` if required) instead of '\
+    "chaining `#{method_one}...#{method_two}`."
   end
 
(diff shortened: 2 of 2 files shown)


lib/rubocop/cop/performance/chain_array_allocation.rb
```diff
@@ -47,8 +47,9 @@ class ChainArrayAllocation < Cop
         HAS_MUTATION_ALTERNATIVE = ':collect :compact :flatten :map :reject '\
                                    ':reverse :rotate :select :shuffle :sort '\
                                    ':uniq '.freeze
-        MSG = 'Use `%<method>s...%<second_method>s!` instead of `%<method>s' \
-              '...%<second_method>s`.'.freeze
+        MSG = 'Use unchained `%<method>s!` and `%<second_method>s!` '\
+              '(followed by `return array` if required) instead of chaining '\
+              '`%<method>s...%<second_method>s`.'.freeze
```
tejasbubane:
```
Use unchained `map!` and `flatten!` (followed by `return array` if required) 
instead of chaining `map...flatten`.`
```

This is the message I came up with for now, let me know if you need a better one.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 * [#6668](https://github.com/rubocop-hq/rubocop/issues/6668): Fix autocorrection for `Style/UnneededCondition` when conditional has the `unless` form. ([@mvz][])
 * [#6382](https://github.com/rubocop-hq/rubocop/issues/6382): Fix `Layout/IndentationWidth` with `Layout/EndAlignment` set to start_of_line. ([@dischorde][], [@siegfault][], [@mhelmetag][])
 * [#6710](https://github.com/rubocop-hq/rubocop/issues/6710): Fix `Naming/MemoizedInstanceVariableName` on method starts with underscore. ([@pocke][])
+* [#6684](https://github.com/rubocop-hq/rubocop/issues/6684): Fix error message for  `Performance/ChainArrayAllocation`. ([@tejasbubane][])
```
Drenmi: This doesn't really require a `CHANGELOG` entry. 🙂

tejasbubane:
@Drenmi Changelog entry removed.

Drenmi:
Thank you for your contribution! 🙇

PR #6714, opened 2019-01-28
 Add `iterator?` to deprecated methods and prefer `block_given?`

### LINKED ISSUE #6688: Prefer `block_given?` to `iterator?`

[The <code>iterator?</code> form is mildly deprecated](https://github.com/ruby/ruby/blob/d3c0b20949916aa7b432a0498a8016d67a7822ad/vm_eval.c#L2101-L2102).

Recommend to use `block_given?` instead of `iterator?`.

**Additional context**

- [Prefer block_given? to iterator?](https://github.com/ruby/ruby/commit/c20aae965e2e79fcf4c443b266f7012157d5b23b)
- [Feature #15547 deprecate iterator?](https://bugs.ruby-lang.org/issues/15547)

tejasbubane:
Closes #6688 

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/deprecated_class_methods_spec.rb
+++ b/spec/rubocop/cop/lint/deprecated_class_methods_spec.rb
@@ -3,38 +3,81 @@
 RSpec.describe RuboCop::Cop::Lint::DeprecatedClassMethods do
   subject(:cop) { described_class.new }
 
-  it 'registers an offense for File.exists?' do
-    expect_offense(<<-RUBY.strip_indent)
-      File.exists?(o)
-           ^^^^^^^ `File.exists?` is deprecated in favor of `File.exist?`.
-    RUBY
-  end
-
-  it 'registers an offense for ::File.exists?' do
-    expect_offense(<<-RUBY.strip_indent)
-      ::File.exists?(o)
+  context 'prefer `File.exist?` over `File.exists?`' do
+    it 'registers an offense for File.exists?' do
+      expect_offense(<<-RUBY.strip_indent)
+        File.exists?(o)
              ^^^^^^^ `File.exists?` is deprecated in favor of `File.exist?`.
-    RUBY
+      RUBY
+    end
+
+    it 'registers an offense for ::File.exists?' do
+      expect_offense(<<-RUBY.strip_indent)
+        ::File.exists?(o)
+               ^^^^^^^ `File.exists?` is deprecated in favor of `File.exist?`.
+      RUBY
+    end
+
+    it 'does not register an offense for File.exist?' do
+      expect_no_offenses('File.exist?(o)')
+    end
+
+    it 'auto-corrects File.exists? with File.exist?' do
+      new_source = autocorrect_source('File.exists?(something)')
+      expect(new_source).to eq('File.exist?(something)')
+    end
   end
 
-  it 'does not register an offense for File.exist?' do
-    expect_no_offenses('File.exist?(o)')
+  context 'prefer `Dir.exist?` over `Dir.exists?`' do
+    it 'registers an offense for Dir.exists?' do
+      expect_offense(<<-RUBY.strip_indent)
+        Dir.exists?(o)
+            ^^^^^^^ `Dir.exists?` is deprecated in favor of `Dir.exist?`.
+      RUBY
+    end
+
+    it 'registers an offense for ::Dir.exists?' do
+      expect_offense(<<-RUBY.strip_indent)
+        ::Dir.exists?(o)
+              ^^^^^^^ `Dir.exists?` is deprecated in favor of `Dir.exist?`.
+      RUBY
+    end
+
+    it 'does not register an offense for Dir.exist?' do
+      expect_no_offenses('Dir.exist?(o)')
+    end
+
+    it 'auto-corrects Dir.exists? with Dir.exist?' do
+      new_source = autocorrect_source('Dir.exists?(something)')
+      expect(new_source).to eq('Dir.exist?(something)')
+    end
+
+    it 'does not register an offense for offensive method `exists?`'\
+       'on other receivers' do
+      expect_no_offenses('Foo.exists?(o)')
+    end
   end
 
-  it 'registers an offense for Dir.exists?' do
-    expect_offense(<<-RUBY.strip_indent)
-      Dir.exists?(o)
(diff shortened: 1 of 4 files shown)


Drenmi:
> The code looks good to me. I think we should add autocorrect support to the cop.

Would you be open to adding auto-correct @tejasbubane? It involves defining an `#autocorrect` method on the cop. Something like this should do the trick:

```ruby
def autocorrect(node)
  lambda do |corrector|
    corrector.replace(node.source_range, 'block_given?')
  end
end
```

I haven't tested it though. 🙂

tejasbubane:
@Drenmi yes I will add autocorrect and make all the changes today. Thanks!

bbatsov:
I don’t think this warrants a dedicated cop. Probably it should be part of some generic cop about preferred methods. I know we have some offenders in the current codebase, but we shouldn’t grow that list.

tejasbubane:
> I don’t think this warrants a dedicated cop. Probably it should be part of some generic cop about preferred methods. I know we have some offenders in the current codebase, but we shouldn’t grow that list.

Which cop do you think this should go? Create a new one called `PreferredMethods`?

bquorning:
We have an existing cop called `DeprecatedClassMethods` where this code could go.

rrosenblum:
`DeprecatedClassMethods` will need to be refactored for this to work. That cop was specifically designed to handle class methods so it looks for both the receiver and the method. `iterator?` has a `nil` receiver.

tejasbubane:
I have made the changes to use `DeprecatedClassMethods` instead of adding a new cop.

spec/rubocop/cop/lint/deprecated_class_methods_spec.rb
```diff
@@ -3,38 +3,64 @@
...
 
-  it 'registers an offense for File.exists?' do
-    expect_offense(<<-RUBY.strip_indent)
-      File.exists?(o)
-           ^^^^^^^ `File.exists?` is deprecated in favor of `File.exist?`.
-    RUBY
-  end
+  before { inspect_source(source) }
+
+  shared_examples 'code with offense' do |code, highlights, expected|
```
bquorning: This looks like a re-implementation of `expect_offense`. Can’t we please use `expect_offense` instead?

Same comment applies to `'code without offense'` and `expect_no_offenses`.

tejasbubane: Thanks. I did not know about `expect_offense`. I simply searched for shared examples and found [these](https://github.com/rubocop-hq/rubocop/blob/7a1baaee07129123b0a4d6cc283064f6bda30838/spec/rubocop/cop/lint/empty_when_spec.rb#L10-L29) and used the same. I'll change them.

bquorning: Thanks. I’ll take a look at those other specs :-)

tejasbubane: Changed.

spec/rubocop/cop/lint/deprecated_class_methods_spec.rb
```diff
@@ -3,38 +3,67 @@
...
   subject(:cop) { described_class.new }
 
-  it 'registers an offense for File.exists?' do
-    expect_offense(<<-RUBY.strip_indent)
-      File.exists?(o)
-           ^^^^^^^ `File.exists?` is deprecated in favor of `File.exist?`.
-    RUBY
-  end
+  context 'prefer `File.exist?` over `File.exists?`' do
+    let(:message) { '`File.exists?` is deprecated in favor of `File.exist?`.' }
```
bquorning: We don’t need the `message` `let`s anymore.

tejasbubane: Ahh. Missed them. Removed now.

spec/rubocop/cop/lint/deprecated_class_methods_spec.rb
```diff
@@ -3,38 +3,61 @@
...
+              ^^^^^^^ `Dir.exists?` is deprecated in favor of `Dir.exist?`.
+      RUBY
+    end
+
+    it 'does not register an offense for Dir.exist?' do
+      expect_no_offenses('Dir.exist?(o)')
+    end
   end
 
-  it 'auto-corrects File.exists? with File.exist?' do
```
bquorning: One final thing: It looks like these auto-correct specs were removed.

tejasbubane: Good catch! Added autocorrection specs for all 3 cases.

Drenmi:
Looks good @tejasbubane! I'm ready to accept the code as is, but it would be cool, since it's a `Kernel` method, to only register an offense when there is no receiver. I.e. `foo.iterator?` shouldn't be an offense.

Maybe the code already does this, but the unit test seems to be missing. 🙂

tejasbubane:
> Maybe the code already does this, but the unit test seems to be missing. 🙂

Yes the code already checked for `nil` receiver. I added the test case.

dgollahon:
This is less about this specific change and might require further discussion outside of this PR, but I noticed that this cop is marked as `safe`. Any time a method selector is changed this is not _safe_ (as I mean it: known to be semantically equivalent). On something like `File.exists?` it's still not _technically_ safe, but there the odds of a semantic change are extremely low so that concerns me a bit less. In this particular case it seems rather unlikely that someone would define `#iterator?`, but it's definitely more likely than redefining `File.exists?` and certainly not _provably_ safe.

Maybe it's good to define `safe` as "safeish" but I, personally, would prefer to be able to trust that `safe` is truly _safe_ under all circumstances. Perhaps we need a distinction between cops with (in theory) provably safe autocorrects (maybe even assert the AST is unchanged?) v.s. things that almost always do the right thing, even if there are edge cases that rubocop can't reasonably detect via static analysis (for example, the `#iterator?` -> `#block_given?`).

Has someone defined a specific policy for what counts as `safe` anywhere?

Drenmi:
> Has someone defined a specific policy for what counts as `safe` anywhere?

You're asking good questions, @dgollahon. We don't have a formal definition of this at the moment. It was also relatively recently we started marking cops as unsafe, so I haven't given this a lot of thought. On first thought, I also prefer that "safe" is truly safe, although it could turn out (I don't have a good feel of this) that it would disqualify a very large subset of cops, which might be the wrong trade-off. 🤔

PR #6718, opened 2019-01-30
Drop old Ruby version matrix from Appveyor CI

koic:
Appveyor is very slow in the CI matrix.

Actually, Appveyor have not recently found a problem.
This PR proposes to drop old Ruby version matrix from Appveyor CI.

-----------------

* [ ] Added tests.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,8 +4,8 @@ version: "{build}"
 
 install:
   - set PATH=C:\Ruby%ruby_version%\bin;%PATH%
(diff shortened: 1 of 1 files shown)


koic:
If you are uncertain about testing for Windows platform, there is another plan to check only the latest Ruby version in Appveyor.

pocke:
I agree the second option which is testing only the latest Ruby on Windows. RuboCop supports Windows. But sometimes patch break Windows support unfortunately. So I think we need to test on Windows with at least the latest Ruby.

PR #6723, opened 2019-02-01
Use h2 for ISSUE_TEMPLATE header

pocke:
The ISSUE_TEMPLATE for a feature request uses `**` to highlight the section header.
I think `##` is more clearly than `**`. `**` only makes it bolder, so I think it is a bit of hard to find the section headers from an issue.

The ISSUE_TEMPLATE for a bug report uses `##` as a header.

I put an example below.

------

# before

---
name: Feature Request
about: Suggest new RuboCop features or improvements to existing features.
---

**Additional context**

# after


koic:
That makes sense!

PR #6735, opened 2019-02-05
[Fix #6708] Style/CommentedKeyword knows :yields:

### LINKED ISSUE #6708: Style/CommentedKeyword should be aware of RDoc Directives
The following “comment” is an RDoc directive. It is *required* to be on the same line as the `def`.

```
class Foo
  def initialize # :yields self:
    yield self if block_given?
  end
end
```

--------

## Expected behavior

RuboCop should not warn about commented keywords.

## Actual behavior

RuboCop warns about commented keywords.

Run Rubocop over the following code:

```
class Foo
  def initialize # :yields self:
    yield self if block_given?
  end
end
```

## RuboCop version

`0.63.1 (using Parser 2.5.1.0, running on ruby 2.6.0 x86_64-darwin18)`

bquorning:
The `Style/CommentedKeyword` cop would not allow the `:yields:` RDoc directive as a trailing comment (to `def`). The documentation https://docs.ruby-lang.org/en/2.1.0/RDoc/Markup.html#class-RDoc::Markup-label-Directives lists many, many directives, but only `:nodoc:` and `:yields:` are shown as trailing comments.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 
 * [#6597](https://github.com/rubocop-hq/rubocop/issues/6597): `Style/LineEndConcatenation` is now known to be unsafe for auto-correct. ([@jaredbeck][])
 * [#6725](https://github.com/rubocop-hq/rubocop/issues/6725): Mark `Style/SymbolProc` as unsafe for auto-correct. ([@drenmi][])
+* [#6708](https://github.com/rubocop-hq/rubocop/issues/6708): Make `Style/CommentedKeyword` allow the `:yields:` RDoc comment. ([@bquorning][])
 
 ## 0.63.1 (2019-01-22)
 
--- a/lib/rubocop/cop/style/commented_keyword.rb
+++ b/lib/rubocop/cop/style/commented_keyword.rb
@@ -6,8 +6,8 @@ module RuboCop
       # This cop checks for comments put on the same line as some keywords.
       # These keywords are: `begin`, `class`, `def`, `end`, `module`.
       #
-      # Note that some comments (such as `:nodoc:` and `rubocop:disable`) are
-      # allowed.
+      # Note that some comments (`:nodoc:`, `:yields:, and `rubocop:disable`)
+      # are allowed.
       #
       # @example
       #   # bad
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * [#6382](https://github.com/rubocop-hq/rubocop/issues/6382): Fix `Layout/IndentationWidth` with `Layout/EndAlignment` set to start_of_line. ([@dischorde][], [@siegfault][], [@mhelmetag][])
 * [#6710](https://github.com/rubocop-hq/rubocop/issues/6710): Fix `Naming/MemoizedInstanceVariableName` on method starts with underscore. ([@pocke][])
 * [#6722](https://github.com/rubocop-hq/rubocop/issues/6722): Fix an error for `Style/OneLineConditional` when `then` branch has no body. ([@koic][])
+* [#6708](https://github.com/rubocop-hq/rubocop/issues/6708): Make `Style/CommentedKeyword` allow the `:yields:` RDoc comment. ([@bquorning][])
```
koic: I think this belongs to `New features` or `Changes`.

lib/rubocop/cop/style/commented_keyword.rb
```diff
@@ -59,7 +59,7 @@ def investigate(processed_source)
         private
 
         KEYWORDS = %w[begin class def end module].freeze
-        ALLOWED_COMMENTS = %w[:nodoc: rubocop:disable].freeze
+        ALLOWED_COMMENTS = %w[:nodoc: :yields: rubocop:disable].freeze
```
koic: This change also update the document.
https://github.com/rubocop-hq/rubocop/blob/v0.63.1/lib/rubocop/cop/style/commented_keyword.rb#L9

And it may be better to be able to customize with default.yml (.rubocop.yml) in the future :-)

bquorning: The document says “such as `:nodoc:` and `rubocop:disable`”, which I thought means it doesn’t need to list *all* allowed comments. Would you prefer listing all allowed comments there? E.g.
```
# Note that some comments (`:nodoc:`, `:yields:, and `rubocop:disable`)
# are allowed.
```
I wouldn’t go for full customization (giving an array in default.yml) just yet. I’d be interested to see – when a bug is reported – which kind of comment would need to be added to the list.

koic: > Would you prefer listing all allowed comments there? E.g.

Yes. I think that's fine.

> when a bug is reported – which kind of comment would need to be added to the list.

OK. Let's think about it at that time.

bquorning: I’ve pushed the new comments.

koic: Thanks for quick updating!

PR #6736, opened 2019-02-06
Detect OS more consistently

### LINKED ISSUE #6720: Layout/EndOfLine does not choose correct default line ending
The Layout/EndOfLine cop automatically decides the 'correct' line ending when set to `:native`. However, the logic used to make that decision does not come to the correct answer when using JRuby. The check being used is `RuboCop::Platform.windows?` which just does an inspection of `RUBY_PLATFORM`:

`RUBY_PLATFORM =~ /cygwin|mswin|mingw|bccwin|wince|emx/`

However, when using JRuby the value of `RUBY_PLATFORM` will be `java`, even when using Windows. This results in the cop deciding that it is in a non-Windows environment and thus check for 'lf' line endings instead of 'crlf'.

--------

## Expected behavior

The Layout/EndOfLine cop should natively use 'crlf' line endings for Windows with JRuby.

## Actual behavior

The Layout/EndOfLine cop natively uses 'lf' line endings for Windows with JRuby.

- use Windows
- use JRuby (9.1.17, in my case)
- have a Ruby file with crlf line endings

Run rubocop against the file and the generated report will contain violations for Layout/EndOfLine.

## RuboCop version

`0.63.1 (using Parser 2.6.0.0, running on jruby 2.3.3 java)`
--

enkessler:
This should fix #6720 and implements the suggested solution.

I did not add new tests, as testing this issue would require adding additional matrix combinations to the AppVeyor build (there is no JRuby/Windows combo being tested in CI) and I have no desire at this time to debug any failures on an entirely new build combination. I did run the default rake task locally and the results are no worse after than before the changes in this PR.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/platform.rb
+++ b/lib/rubocop/platform.rb
@@ -5,7 +5,7 @@ module RuboCop
   # on.
   module Platform
     def self.windows?
-      RUBY_PLATFORM =~ /cygwin|mswin|mingw|bccwin|wince|emx/
+      RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/
     end
   end
 end


koic:
Thanks!

PR #6740, opened 2019-02-06
Start converting tests to use expect_correction

rrosenblum:
Making use of `expect_correction` in some of the smaller groups of tests. For the most part the change went smooth. These were all manual updates. The names of some of the tests feel a bit weird now. I really enjoy having a small compact test that check for an offense, the message, and correction.

I need to open up an issue for the message in `Rails/ReadWriteAttribute`. It uses generic examples rather than actual variables names from the code.


Drenmi:
Beautiful! Great job! 🙇

PR #6762, opened 2019-02-12
Specify `stable` to build image on Read the Docs

koic:
http://www.rubocop.org seems to be redirecting to Read the Docs.
https://docs.rubocop.org

By default, Read the Docs refers to `latest` (i.e. master branch).
https://docs.readthedocs.io/en/latest/config-file/v1.html#build-image

Here is the `latest` docs.
https://docs.rubocop.org/en/latest

However I think RuboCop users expect `stable` docs.
https://docs.rubocop.org/en/stable

This idea was what @pocke taught me before. Thanks!

-----------------

* [ ] Added tests.

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,2 +1,5 @@
+build:
+    image: stable
+
 formats:
     - htmlzip


pocke:
Thank you! It sounds great :+1:
But I'm not familiar Read the Docs, and I'm not sure what `build.image` option is.
I (and users) expect it redirects to `/stable` instead of `/latest` when I access http://docs.rubocop.org . If it will work, I think we should merge this pull request :+1:

koic:
Thanks! I will try it first.

koic:
It has not affected https://rubocop.readthedocs.io/en/latest yet. I will check it for a while whether the site will be updated.

PR #6763, opened 2019-02-12
Allow parens for calls in range literals (Style/MethodCallWithArgsParentheses)

gsamokovarov:
With the style of `omit_parentheses` in the `Style/MethodCallWithArgsParentheses` cop, we got:

Before:

```ruby
1..limit(n)
```

After:

```ruby
1..limit(n)
```

👌

### DIFF
--- a/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
+++ b/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
@@ -259,13 +259,16 @@ module RuboCop
             allowed_chained_call_with_parentheses?(node)
         end
 
+        # rubocop:disable Metrics/CyclomaticComplexity
         def call_in_literals?(node)
           node.parent &&
             (node.parent.pair_type? ||
              node.parent.array_type? ||
+             node.parent.irange_type? || node.parent.erange_type? ||
              splat?(node.parent) ||
              ternary_if?(node.parent))
         end
+        # rubocop:enable Metrics/CyclomaticComplexity
 
         def call_in_logical_operators?(node)
           node.parent &&
--- a/spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
+++ b/spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
@@ -512,6 +512,13 @@ RSpec.describe RuboCop::Cop::Style::MethodCallWithArgsParentheses, :config do
       RUBY
     end
 
+    it 'accepts parens in range literals' do
+      expect_no_offenses(<<-RUBY)
+        1..limit(n)
+        1...limit(n)
+      RUBY
+    end
+
     it 'auto-corrects single-line calls' do
       original = <<-RUBY.strip_indent
         top.test(1, 2, foo: bar(3))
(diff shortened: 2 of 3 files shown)


spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
```diff
@@ -512,6 +512,12 @@ def seatle_style arg = default(42)
       RUBY
     end
 
+    it 'accepts parens in range literals' do
+      expect_no_offenses(<<-RUBY)
+        1..limit(n)
+      RUBY
+    end
```
koic: Looks good! On the other hand, there is also `erange` in the similar case :-)

```ruby
1...limit(n)
```

Would you please add along with the CHANGELOG entry?

lib/rubocop/cop/style/method_call_with_args_parentheses.rb
```diff
@@ -263,6 +263,7 @@ def call_in_literals?(node)
           node.parent &&
             (node.parent.pair_type? ||
              node.parent.array_type? ||
+             node.parent.irange_type? ||
```
pocke: Could you add `erange_type` also? `erange` is `...`.

```bash
$ ruby-parse -e 1..2
(irange
  (int 1)
  (int 2))

$ ruby-parse -e 1...2
(erange
  (int 1)
  (int 2))

```

spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
gsamokovarov: Oh, didn't know that. Thanks for the catch!

lib/rubocop/cop/style/method_call_with_args_parentheses.rb
```diff
@@ -259,13 +259,16 @@ def legitimate_call_with_parentheses?(node)
             allowed_chained_call_with_parentheses?(node)
         end
 
+        # rubocop:disable Metrics/CyclomaticComplexity
```
gsamokovarov: The whole `omit_parentheses` style is based around the many edge-cases where we actually _do_ need the parentheses. I have tried to cater to the `Metrics/CyclomaticComplexity` until now, but the extra helpers I introduce to hide the conditions blow up the line length cop. I think the code will be clearer if we don't measure the cyclomatic complexity for this whole cop and eventually inline the helpers where they are used.

koic: Introduction of `range_type?` meaning both `irange_type?` and `erange_type?` is proposed.
https://github.com/rubocop-hq/rubocop/pull/6126#discussion_r255288410

We will be able to remove the disabling cop comment when replaced by this proposal . So I think this is fine for this.

koic:
Thanks!

lib/rubocop/cop/style/method_call_with_args_parentheses.rb
koic: https://github.com/rubocop-hq/rubocop/pull/6773 will remove the disabling cop comment.

gsamokovarov: Great!

PR #6765, opened 2019-02-12
Allow parens for optional keyword value calls in Style/MethodCallWithArgsParentheses

gsamokovarov:
Yet-another-edge-case for the `omit_parentheses` style.

---

Before:

```ruby
def foo(bar: baz(42))
                ^^^^ Omit parentheses for method calls with arguments.
end
```

After:

```ruby
def foo(bar: baz(42))
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Bug fixes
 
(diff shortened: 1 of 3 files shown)


koic:
This looks good 🌟 Would you add the CHANGELOG entry?

gsamokovarov:
@koic, thanks for the reminder. Added the changelog entry, the change should be good now.

koic:
Thank you @gsamokovarov!

PR #6733, opened 2019-02-05
Warn duplicated keys in .rubocop.yml

pocke:
Problem
===

Sometimes `.rubocop.yml` has duplicated keys, but RuboCop ignores them. It confuses us.
For example https://github.com/rubocop-hq/rubocop/issues/6728

```yaml
# RuboCop ignores this configuration.
Metrics/BlockLength:
  Max: 50

# long long lines

Metrics/BlockLength:
  Max: 25
```

Solution
===

Warn duplicated keys in .rubocop.yml.
For example

```bash
$ rubocop 
.rubocop.yml:2: `Metrics/BlockLength` is concealed by line 7
Inspecting 0 files


0 files inspected, no offenses detected
```

TODOs
===

- [x] Write test for YAMLDuplicationChecker
- [x] Fix existing tests that are failed
  * RuboCop uses wrong cache of `PathUtil.pwd`, so they are failed.

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](../blob/master/CONTRIBUTING.md#changelog-entry-format).
* [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.
* [x] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: https://chris.beams.io/posts/git-commit/

PR #6766, opened 2019-02-13
Drop Ruby 2.2.0 and 2.2.1 to use Psych 3.1

pocke:
see https://github.com/rubocop-hq/rubocop/pull/6733#discussion_r254700697

This pull request drops supporting Ruby 2.2.0 and 2.2.1 to use Psych 3.1.

Parameter style of `YAML.load` has been changed since Psych 3.1. Psych 3.1 still supports the old parameter style, but it displays a warning. So we need to change the behaviour between Psych 3.1 and others to avoid the warning. We can delete the branch by dropping old versions of Psych.

And Psych 3.1 also has a location feature. I'd like to use the feature in #6733.

Note: I'd like to drop Ruby 2.2 also. Personally, I think we can drop it after Ruby 2.3 EOL, it is March 2019.
I'm looking forward to using Safe Navigation Operator in RuboCop! :wink:  

-----------------

* [ ] Added tests.

### DIFF
--- a/rubocop.gemspec
+++ b/rubocop.gemspec
@@ -9,7 +9,7 @@ Gem::Specification.new do |s|
   s.name = 'rubocop'
   s.version = RuboCop::Version::STRING
   s.platform = Gem::Platform::RUBY
-  s.required_ruby_version = '>= 2.2.0'
+  s.required_ruby_version = '>= 2.2.2'
   s.authors = ['Bozhidar Batsov', 'Jonas Arvidsson', 'Yuji Nakayama']
   s.description = <<-DESCRIPTION
     Automatic Ruby code style checking tool.
@@ -38,6 +38,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('parallel', '~> 1.10')
   s.add_runtime_dependency('parser', '>= 2.5', '!= 2.5.1.1')
   s.add_runtime_dependency('powerpack', '~> 0.1')
+  s.add_runtime_dependency('psych', '>= 3.1.0')
   s.add_runtime_dependency('rainbow', '>= 2.2.2', '< 4.0')
   s.add_runtime_dependency('ruby-progressbar', '~> 1.7')
   s.add_runtime_dependency('unicode-display_width', '~> 1.4.0')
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

koic:
> I'd like to drop Ruby 2.2 also. Personally, I think we can drop it after Ruby 2.3 EOL, it is March 2019.

I think so too :-)

pocke:
https://github.com/ruby/psych/issues/385

Installing RuboCop 0.65.0 has a problem on Windows by using Psych 3.1.
I guess it is not RuboCop matter, but I just put a link to the issue.

rubocop.gemspec
```diff
@@ -38,6 +38,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('parallel', '~> 1.10')
   s.add_runtime_dependency('parser', '>= 2.5', '!= 2.5.1.1')
   s.add_runtime_dependency('powerpack', '~> 0.1')
+  s.add_runtime_dependency('psych', '>= 3.1.0')
```
Drenmi: @pocke Adding this as a runtime dependency seems to break a lot of people's environments. See #6781. Can we get around this somehow?

rubocop.gemspec
```diff
@@ -9,7 +9,7 @@ Gem::Specification.new do |s|
   s.name = 'rubocop'
   s.version = RuboCop::Version::STRING
   s.platform = Gem::Platform::RUBY
-  s.required_ruby_version = '>= 2.2.0'
+  s.required_ruby_version = '>= 2.2.2'
```
coldnebo: I would add to @Drenmi's comment by saying that if you are going to require psych 3.1.0, you should probably require ruby 2.5.0 or greater as that seems to be when [ruby core made the switch](https://github.com/ruby/ruby/commit/6d77e28763ed17f75edf3b4072701b4dbd7644bb).

coldnebo:
Unfortunately the windows issue and the other psych issues are related to your PR.

See [my comment](https://github.com/rubocop-hq/rubocop/issues/6781#issuecomment-471061171).

PR #6769, opened 2019-02-15
FIPS: replace MD5 with SHA1

dirtyharrycallahan:
* rubocop fails when host has enabled FIPS mode because MD5 will
  be disabled. Replaced MD5 with SHA1 which my benchmark is also
  shown to be faster at least on SkyLake + Red Hat Enterprise Linux
  7 across Ruby 2.0.0, Ruby 2.3.8, Ruby 2.4.5, Ruby 2.5.3, and
  Ruby 2.6.1.

-----------------

* [ ] Added tests.


mikegee:
What impact do we expect for users? The caching mechanism will "miss" all lookups on the next run after upgrade, right? Anything else?

dirtyharrycallahan:
Correct. The change will cause cache "misses" on the next run. That could be avoided by including MD5 within rubocop or at least not use digest/md5 which relies on openssl.
Granted this probably does not impact too many users but we hit this as soon as we rolled out required FIPS mode on all of servers in production and development.

bquorning:
FIPA is mentioned in the “Digest algorithms” section of the `Digest` documentation: https://ruby-doc.org/stdlib-2.5.3/libdoc/digest/rdoc/Digest.html#module-Digest-label-Digest+algorithms

I cannot measure any performance difference between `Digest::MD5` and `Digest::SHA1`. For our purposes, I think cryptographic strength and collision guarantees aren’t *that* important, as long as the hashing is fast and deterministic.

This change will invalidate the all users’ RuboCop caches, but I’m not sure that is enough of an inconvenience that it should stop this PR.

koic:
Rather than changing the default algorithm, it is better to provide some option that can be changed to another algorithm. I think that impact should not be given to users by measurement of implementation-dependent.

bquorning:
I also considered making it user configurable whether to use `Digest::MD5` or `Digest::SHA1`. But since all we really want is a reliable, fast hash function, our users shouldn’t care which digest algorithm is used internally – as long as it works.

koic:
> But since all we really want is a reliable, fast hash function, our users shouldn’t care which digest algorithm is used internally – as long as it works.

Ah, sure. I agree!

bbatsov:
I agree with the proposed change. The impact on the end users is acceptable IMO.

koic:
@dirtyharrycallahan Could you write the change log entry to `### Changes` section of CHANGELOG.md?
https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md#changes

Change log entry format is as follows.

dirtyharrycallahan:
Working on it.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 * [#6766](https://github.com/rubocop-hq/rubocop/pull/6766): Drop support for Ruby 2.2.0 and 2.2.1. ([@pocke][])
 * [#6733](https://github.com/rubocop-hq/rubocop/pull/6733): Warn duplicated keys in `.rubocop.yml`. ([@pocke][])
+* [#6769](https://github.com/rubocop-hq/rubocop/pull/6769): Replace md5 with sha1 for FIPS compliance. ([@dirtyharrycallahan][])
```
koic: RuboCop 0.65.0 has been released. Please rebase with the latest master branch.

koic: And CI has failed.

```console
  1) RuboCop Project changelog has link definitions for all implicit links
     Failure/Error:
       expect(changelog.include?("[#{name}]: http"))
         .to be(true), "CHANGELOG.md is missing a link for #{name}. " \
                       'Please add this link to the bottom of the file.'
```

https://circleci.com/gh/rubocop-hq/rubocop/37727

Please add a link definition for the link to the bottom of the changelog as `[@dirtyharrycallahan]: https://github.com/dirtyharrycallahan`.

```diff
+[@dirtyharrycallahan]: https://github.com/dirtyharrycallahan
```

lib/rubocop/version.rb
```diff
@@ -3,7 +3,7 @@
 module RuboCop
   # This module holds the RuboCop version information.
   module Version
-    STRING = '0.64.0'.freeze
+    STRING = '0.64.1'.freeze
```
koic: This change is unnecessary. This is what a maintainer updates when releasing.

koic:
I wrote some review comments. Please squash your commits into one when these are done :-)

dirtyharrycallahan:
Been out of pocket. I messed up the PR by including some internal changes. Let me try to fix this up tonight.

bbatsov:
@dirtyharrycallahan Ping :-)

dirtyharrycallahan:
My github/git foo is not strong. Let me see if I can figure how to fix this up without create a new PR.

mikegee:
> My github/git foo is not strong.

Feel free to post specific questions. I can help. I'm sure others here would love to help too.

bbatsov:
I've committed this myself.

PR #6772, opened 2019-02-16
New cop:  Layout/OrderedMethods

shanecav84:
New cop to enforce that methods are ordered alphabetically per visibility block.

5587453 was accomplished using a corrector that's still a work-in-progress.

-----------------

* [x] Added tests.


koic:
I think that it is good to have some kind of some consistency style like that. Unfortunately, this change to force uniformly alphabetically makes it harder to see git blame and history. In the many changes to RuboCop repository, the benefit which exceeds the loss of making git history difficult to read cannot be considered from this diff.  Also it will be hard to maintain even if default is set to disable, so I'm sorry to close this PR. It might be PoC to try making it as 3rd party gem.

shanecav84:
https://github.com/shanecav84/rubocop-ordered_methods

PR #6782, opened 2019-02-20
README tweaks

sgerrand:
💁 Some small grammatical edits:
* "financially supporting" makes more sense.
* either "use a conservative version lock" or "use conservative version locking" makes sense, the previous version didn't.


Drenmi:
Thanks! 🙇

PR #6126, opened 2018-07-24
Add a Strict mode to Style/MutableConstants

rrosenblum:
This has been something that I have wanted to do for a while. Right now, we only freeze literals that are assigned to constants. This leaves many constants unfrozen and therefore potentially unsafe.

I consider this to be experimental functionality. I'm not sure how many other people would be interested in this feature. I think I have all of the main use cases covered. I am trying to account for any edge cases now. 

Something interesting that I learned from this is that concatenating frozen strings produces an unfrozen string.

PR #6784, opened 2019-02-20
Make `Style/RedundantFreeze` aware of methods that will produce frozen objects

rrosenblum:
This was extract out of the PR to add strict mode to `Style/MutableConstants`, #6126. This isn't an exhaustive list of methods that return immutable values, but it is at least a start.


config/default.yml
```diff
@@ -3867,6 +3867,7 @@ Style/RedundantFreeze:
   Description: "Checks usages of Object#freeze on immutable objects."
   Enabled: true
   VersionAdded: '0.34'
+  VersionAdded: '0.66'
```
mikegee: Is it ok to have two `VersionAdded` or was this one supposed to be `VersionChanged`?

rrosenblum: It's supposed to be `VersionChanged `. Copy/paste error.

CHANGELOG.md
```diff
@@ -48,6 +52,7 @@
 
 * [#6704](https://github.com/rubocop-hq/rubocop/pull/6704): Add new `Rails/ReflectionClassName` cop. ([@Bhacaz][])
 * [#6643](https://github.com/rubocop-hq/rubocop/pull/6643): Support `AllowParenthesesInCamelCaseMethod` option on `Style/MethodCallWithArgsParentheses` `omit_parentheses`. ([@dazuma][])
+* Add an experimental strict mode to `Style/MutableConstant` that will freeze all constants, rather than just literals. ([@rrosenblum][])
```
bbatsov: I think you've added this here accidentally.

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@
 * [#6806](https://github.com/rubocop-hq/rubocop/pull/6806): Remove `powerpack` dependency. ([@dduugg][])
 * [#6810](https://github.com/rubocop-hq/rubocop/pull/6810): Exclude gemspec file by default for `Metrics/BlockLength` cop. ([@koic][])
 
+### New features
```
bbatsov: This should be moved to changes probably.

spec/rubocop/cop/style/redundant_freeze_spec.rb
```diff
@@ -23,6 +23,11 @@
   it_behaves_like 'immutable objects', '1.5'
   it_behaves_like 'immutable objects', ':sym'
   it_behaves_like 'immutable objects', ':""'
+  it_behaves_like 'immutable objects', "ENV['foo']"
+  it_behaves_like 'immutable objects', "'foo'.count"
+  it_behaves_like 'immutable objects', '(1 + 2)'
```
bbatsov: I guess we also need tests which show what happens for objects other than numbers.

rrosenblum: I added tests for comparing strings, comparing variables, adding string, and adding variables.

PR #6786, opened 2019-02-21
Improve specs thanks to rspectre

bquorning:
I tried running the amazing :sparkles:[rspectre](https://github.com/dgollahon/rspectre):sparkles: tool against our spec suite, and discovered a few different problems:

 - A few `let`s were no longer being used.
 - A `shared_context` was only being used once.
 - The same `let` was being defined twice within the same context.
 - One `let` had its `end` misplaced, and was thus wrapping an entire block of examples that have therefore never been run…
 - Some `shared_examples` called by `include_examples` caused the issue that [the RSpec documentation](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples) warns against:

> When you include parameterized examples in the current context multiple times, you may override previous method definitions and last declaration wins.

@dgollahon @Darhazer Would some of these issues be discoverable by rubocop-rspec? We could e.g. warn on a `let` inside a `shared_example`…

-----------------

* [x] Added tests.

### DIFF
--- a/spec/project_spec.rb
+++ b/spec/project_spec.rb
@@ -10,15 +10,10 @@ RSpec.describe 'RuboCop Project', type: :feature do
       .map(&:cop_name)
   end
 
-  shared_context 'configuration file' do |config_path|
-    subject(:config) { RuboCop::ConfigLoader.load_file(config_path) }
+  describe 'default configuration file' do
+    subject(:config) { RuboCop::ConfigLoader.load_file('config/default.yml') }
 
     let(:configuration_keys) { config.keys }
-    let(:raw_configuration) { config.to_h.values }
-  end
-
-  describe 'default configuration file' do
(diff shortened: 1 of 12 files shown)


dgollahon:
> warn on a `let` inside a `shared_example`

We should definitely do this. I thought I had an opened an issue to this effect but I suppose not. I also wondered at the time if this was possible to open an issue on `rspec` about this but I'm not sure if that would be easy to detect or not for them.

> The same `let` was being defined twice within the same context.

Don't we have a cop for this already? Any idea why it wasn't caught here?

I think the others are not something we can detect in `rubocop-*`.

dgollahon:
Follow up: I noticed from your commit message that `rspec` does already warn on clobbering methods so I doubt they'd want to bother with a specific warning for using the block associated with `include_examples` but someone could ask.

Drenmi:
Nice! 🚀

PR #6787, opened 2019-02-22
Add InstanceVarsInClass cop

anujbiyani:
Inspired by https://github.com/rubocop-hq/ruby-style-guide/issues/538

This cop checks for instance variables used in classes outside of the initializer. It encourages using `attr_reader` or `attr_accessor` to set up safer methods to read from or write to your instance variable.

Things I didn't do but are worth calling out:
1. The cop doesn't look at whether or not the `attr_reader` / `attr_accessor` should be made private, though the linked issue does go into that too. I felt that'd be too complicated to implement accurately.
1. Anonymous classes or dynamically defined classes (`Class.new`, essentially) are not checked. I spiked on implementing this but I'm not sure how to differentiate between arbitrary blocks and blocks that are used in a class definition.
1. I didn't use `def_node_matcher`, I couldn't figure out how to code this up using that approach. Happy to refactor if someone can help.
1. When I tried this in my Rails application, I ignored a couple directories straightaway (controllers, jobs, mailers...) because this cop doesn't really apply there. Feels awkward to code this into the cop, but I called it out in the documentation.

-----------------

* [x] Added tests.


spec/rubocop/cop/style/instance_vars_in_class_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  context 'in a class' do
+    context 'when using an instance variable instead of an attr_reader' do
+      it 'registers an offense' do
+        expect_offense(<<-RUBY.strip_indent)
+          class TestClass
+            def foo
+              @instance_variable
+              ^^^^^^^^^^^^^^^^^^ Outside of the initializer, use an `attr_reader :instance_variable` instead of `@instance_variable`.
```
marcotc: In case there's no `attr_*` declared for `@instance_variable`, is it sill fair to enforce this cop?

This also brings up another point which is private instance variables, as one might not have an `attr_*` for them just for internal use.

anujbiyani: My phrasing probably isn't very good; this is meant to tell users to define an `attr_*` and then use that to interact with the `@instance_variable`.

It's meant to apply in the case of private instance variables as well. Basically this is the goal: https://github.com/rubocop-hq/ruby-style-guide/issues/538#issuecomment-208771509

I wasn't sure how to convey the private part concisely, though, so I left it out. I'll try and come up with a better message.

anujbiyani: @marcotc Updated the message. It's longer now, unfortunately, but more descriptive.

Drenmi:
I have a feeling this might produce false positives for class instance variables, e.g.:

```ruby
class Foo
  @bar
end
```

Perhaps the name of the cop should be `InstanceVariableOutsideInitializer`? 🙂

koic:
Thanks for opening your PR. However I believe it is hard to add it to the core cops.

- This will destroy the encapsulation which is one of the fundamentals of object-oriented programming when prepared unnecessary public interface against the private status. And publish interfaces that collaborate with other objects should be limited to stable ones. When all instance variables are unconditionally published, they become unstable on dependent objects.
- Described in the document, however it seems to be incompatible with Rails application with a lot of usage scenes.

> Perhaps the name of the cop should be `InstanceVariableOutsideInitializer`? 🙂

Alternatively, there may be different ideas, such as suppressing instance variables at the top level, but I'm not sure yet. In any case it will be a different cop. Also it will be hard to maintain even if default is set to disable so I'm sorry I will close this PR. Thanks for your proposal to share your idea. And If you need it in a limited context, it may be good to start from a 3rd party gem. Thank you again!

lib/rubocop/cop/style/instance_vars_in_class.rb
```diff
@@ -0,0 +1,128 @@
...
+      #   # good
+      #   class Foo
+      #     attr_accessor :foo
+      #
+      #     def initialize
+      #       @foo = "bar"
+      #     end
+      #
+      #     def instance_method
+      #       foo = "baz"
```
ujihisa: This doesn't work as `Foo#foo=`, but just as a local var assignment. This has to be `self.foo = "baz"` instead

anujbiyani: Thanks for catching!

anujbiyani:
@Drenmi What's the use case for an instance var floating in a class? My instinct is that's better off being either a constant or defined in an initializer, but maybe I'm missing something.

@koic Thanks for the feedback, I understand if this cop is too unmaintainable to be merged. A couple points I'd like considered before closing the book on this:
> * This will destroy the encapsulation which is one of the fundamentals of object-oriented programming when prepared unnecessary public interface against the private status. And publish interfaces that collaborate with other objects should be limited to stable ones. When all instance variables are unconditionally published, they become unstable on dependent objects.

Unless I'm misunderstanding, this can be resolved by just recommending the `attr_*` be made `private` no? https://github.com/onemedical/rubocop/commit/6edbb936ca4a2da740c77807ae46161d86854534

> * Described in the document, however it seems to be incompatible with Rails application with a lot of usage scenes.

Yeah, if recommending `Exclude:` isn't good enough then I could either hardcode the exclusions and/or move this into the `Rails` category?

PR #6791, opened 2019-02-25
Allow Rails/ReflectionClassName to use symbol argument

unasuke:
```ruby
belongs_to :account, class_name: :Account
```

Rails/ReflectionClassName ( introduced from #6704 ) offences that code, but symbol is also uses as string so I think this code has no offences by the cop.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * [#6800](https://github.com/rubocop-hq/rubocop/issues/6800): Fix an incorrect auto-correct for `Rails/Validation` when method arguments are enclosed in parentheses. ([@koic][])
 * [#6808](https://github.com/rubocop-hq/rubocop/issues/6808): Prevent false positive in `Naming/ConstantName` when assigning a frozen range. ([@drenmi][])
 * Fix the calculation of `Metrics/AbcSize`. Comparison methods and `else` branches add to the comparison count. ([@rrosenblum][])
+* [#6791](https://github.com/rubocop-hq/rubocop/pull/6791): Allow `Rails/ReflectionClassName` to use symbol argument for `class_name`. ([@unasuke][])
 
 ### Changes
 
@@ -3853,3 +3854,4 @@
 [@Bhacaz]: https://github.com/bhacaz
 [@enkessler]: https://github.com/enkessler
 [@tagliala]: https://github.com/tagliala
+[@unasuke]: https://github.com/unasuke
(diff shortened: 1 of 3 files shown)


koic:
That makes sense. Can you write the CHANGELOG entry?

unasuke:
@koic I see!

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changes
 
 * [#6688](https://github.com/rubocop-hq/rubocop/pull/6688): Add `iterator?` to deprecated methods and prefer `block_given?` instead. ([@tejasbubane][])
+* [#6791](https://github.com/rubocop-hq/rubocop/pull/6791): Allow `Rails/ReflectionClassName` to use symbol argument. ([@unasuke][])
```
koic: Ah, probably this change is a bug fix. The symbol object is a valid argument that will not cause the following error.
https://github.com/rubocop-hq/rubocop/pull/6704#issuecomment-457121739

unasuke: Wow :exclamation: I moved the line to `fix bug` section, thx :wink:

koic:
Unfortunately I had overlooked the discussion below 💦 
https://github.com/rubocop-hq/rubocop/pull/6704#issuecomment-462247177

I have an opportunity to meet Rails committer soon. I'd like to make a final judgment by listening to opinion as Rails. Please wait for a little while longer.

Drenmi:
Looks like a number of different test cases have been squashed into one (even before you made this change.) Could you please separate the test cases so we test one thing per test? 🙂

koic:
Correct 💦 The test case of String object and the test case of Symbol object should have been independent.

mikegee:
I disagree with allowing a symbol here. It is immediately converted to a string with `to_s`, so we are ok requiring a string in the first place.

koic:
I think that using String for `class_name` will be more unified than using Symbol. However I'd like to accept this change. Because the following test case that accepts Symbol with `class_name` is written in test cases of Rails.
https://github.com/rails/rails/blob/v6.0.0.beta2/activerecord/test/cases/associations/has_and_belongs_to_many_associations_test.rb#L950-L955

On the other hand, If you propose to Rails that it does not accept Symbol with `class_name` and accept it, let's change it so that `Rails/ReflectionClassName` cop does not accept Symbol at that time.

Cc @amatsuda @kamipo @y-yagi

koic:
Thanks!

unasuke:
Thank you too!

PR #6795, opened 2019-02-28
[Fix #6777] Fix a false positive for `Style/TrivialAccessors`

### LINKED ISSUE #6777: Style/TrivialAccessors wrongly auto-corrects methods outside of classes
We have a method inside our `spec_helper.rb` file like this

```
def response
  @response
end
```
It's detected by `Style/TrivialAccessors` and replaced with `attr_reader` leading to error 

```
Failure/Error: attr_reader :response

NoMethodError:
  undefined method `attr_reader' for main:Object
```

## Expected behavior

`Style/TrivialAccessors` cop should not auto-correct methods outside classes.

## Actual behavior

It auto-corrects methods outside classes, e.g. in `spec/spec_helper.rb` to use `attr_reader` breaking the code.

```
# frozen_string_literal: true

def foo
  @foo
end

puts foo
```

This gets auto-corrected to
```
# frozen_string_literal: true

attr_reader :foo

puts foo
```

`ruby foo.rb` then gives the error `undefined method attr_reader for main:Object (NoMethodError)`

## RuboCop version

```
0.64.0 (using Parser 2.6.0.0, running on ruby 2.6.1 x86_64-darwin18)
```

koic:
Fixes #6777.

This PR fixes a false positive for `Style/TrivialAccessors` when using reader / writer at the top level.

The following is a reproduction code.

```ruby
# example.rb
def foo
  @foo
end
```

```console
% rubocop example.rb --only Style/TrivialAccessors -a
Inspecting 1 file
C

Offenses:

example.rb:1:1: C: [Corrected] Style/TrivialAccessors: Use attr_reader
to define trivial reader methods.
def response
^^^

```

An error occurs in the auto-corrected code.

```diff
% git diff
diff --git a/example.rb b/example.rb
index b26fec0..1868b1d 100644
--- a/example.rb
+++ b/example.rb
@@ -1,5 +1,3 @@
-def response
-  @response
-end
+attr_reader :response
```

```console
% ruby example.rb
example.rb:1:in `<main>': undefined method `attr_reader' for main:Object (NoMethodError)
```

This also occurs with writer. This PR will allow these cases.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/trivial_accessors_spec.rb
+++ b/spec/rubocop/cop/style/trivial_accessors_spec.rb
@@ -7,50 +7,64 @@ RSpec.describe RuboCop::Cop::Style::TrivialAccessors, :config do
 
   it 'registers an offense on instance reader' do
     expect_offense(<<-RUBY.strip_indent)
-      def foo
-      ^^^ Use `attr_reader` to define trivial reader methods.
-        @foo
+      class Foo
+        def foo
+        ^^^ Use `attr_reader` to define trivial reader methods.
+          @foo
+        end
       end
     RUBY
 
     expect_correction(<<-RUBY.strip_indent)
-      attr_reader :foo
+      class Foo
+        attr_reader :foo
+      end
     RUBY
   end
 
   it 'registers an offense on instance writer' do
     expect_offense(<<-RUBY.strip_indent)
-      def foo=(val)
-      ^^^ Use `attr_writer` to define trivial writer methods.
-        @foo = val
+      class Foo
+        def foo=(val)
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/trivial_accessors_spec.rb
```diff
@@ -234,6 +284,22 @@ def bar
...
+    RUBY
+  end
+
+  it 'accepts writer using top level' do
+    expect_no_offenses(<<-RUBY.strip_indent)
+      def bar=(bar)
+        @bar = bar
+      end
+    RUBY
+  end
```
koic: 📝 These two test cases are the main changes.

PR #6806, opened 2019-03-03
rm powerpack dependency

dduugg:
Rubocop adds a powerpack depenendency for five core extensions in production (as well as an additional one for tests). These extensions can cause issues when rubocop is executed in an environment that requires different versions of these core extensions. Further, two of them can be replaced by more concise code (`Array#butfirst` and `Enumerabale#drop_last`). An additional one (`Hash#symbolize_keys`) has only a single call site, which has been replaced with an inline version.
The remaining two (as well as the test dependency), are replaced by core extensions. (One of these, `String#strip_indent`, can be replaced by switching callers to use squiggly heredocs, once the minimum ruby version is 2.3.). The core extensions are preferred to the powerpack dependency, as it raises the barrier to the introduction of further core extensions.

Since there is no change to user behavior, this change relies on existing tests (though I'm happy to add more on request).

-----------------

* [ ] Added tests.
* [X] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](../blob/master/CONTRIBUTING.md#changelog-entry-format).


bbatsov:
I’ll think about this. Powerpack started as I wanted to extract some code out of RuboCop (and I didn’t want to use one of the huge libraries that existed with tons of extensions), but these days I don’t mind duplicating code that much. Never had much time for Powerpack to refine the idea of minimal extension lib that adds just really useful stuff, which should have probably been in Ruby in the first place. I was also ruminating on whether instead of extensions those should be module functions or something like this.

PR #6771, opened 2019-02-16
add specs for Layout/RescueEnsureAlignment fix c8ab8d9 from #6254

oehlschl:
This PR demonstrates that there are still unexpected issues with `Layout/RescueEnsureAlignment` even after issue #6254 was resolved. I'm not familiar enough with Rubocop internals to resolve the issue, but the specs still failing include the original test case from @ashmaroli.

-----------------

Before submitting the PR make sure the following are checked:

* [ ] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [ ] Feature branch is up-to-date with `master` (if not - rebase it).
* [ ] Squashed related commits together.
* [ ] Added tests.
* [ ] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](../blob/master/CONTRIBUTING.md#changelog-entry-format).
* [ ] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.
* [ ] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: https://chris.beams.io/posts/git-commit/

PR #6818, opened 2019-03-06
Layout/RescueEnsureAlignment continuation of #6254 #6771

formigarafa:
This PR fix unexpected issues with Layout/RescueEnsureAlignment raised by #6771.
By implementing the solution for the related problems one pre-existing spec come to fail but I am not so sure if that has to pass as is or if that simply highlight the existence of two other possible conflicting layouts for rescue that were not being considered.

1.
```
a_variable = begin
               something
             rescue
               log_some_failure
             end
```

2.
```
a_variable = begin
  something
rescue
  log_some_failure
end
```
One of these 2 cases passes and the other fail after my changes but none of them fail before. I don't know if one of those styles should be selected over the other or if they should be considered equivalent and keep passing without error.
-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/rescue_ensure_alignment.rb
+++ b/lib/rubocop/cop/layout/rescue_ensure_alignment.rb
@@ -101,6 +101,8 @@ module RuboCop
               node.loc.begin
             when :def, :defs, :class, :module
               node.loc.name
+            when :lvasgn
+              node.child_nodes.first.loc.begin
             else
               # It is a wrapper with access modifier.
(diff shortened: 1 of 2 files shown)


formigarafa:
@Drenmi , 
Would I be correct to think that each of the styles I mentioned above would be defined by `Layout/EndAlignment` the same way as `Layout/ElseAlignment` does at https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/layout/else_alignment.rb#L109 ???

PR #6819, opened 2019-03-07
Add new Style/ConstantVisibility cop

Drenmi:
This cop enables enforcing visibility scope for class- and module constants. Constants without a visibility declaration, e.g.:

```
class Foo
  BAR = 42
end
```

will register an offense, prompting the user to declare visibility:

```
class Foo
  BAR = 42
  private_constant :BAR
end
```

I decided to make it disabled by default, because I don't think many users have spent time caring about this. 😅 

-----------------

* [X] Added tests.
* [X] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](../blob/master/CONTRIBUTING.md#changelog-entry-format).


bbatsov:
> I decided to make it disabled by default, because I don't think many users have spent time caring about this. 😅

I know I certainly didn't. :-)

bbatsov:
Btw, you should totally add something about this to the style guide, although there I would suggest using `private_constant` when needed, but not having to explicitly mark the public ones.

PR #6821, opened 2019-03-07
Fix false negative for Rails/LinkToBlank when _blank is a symbol

Intrepidd:
when using `target: :_blank`, the `Rails/LinkToBlank` cop does not register an offence when it should.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug fixes
 
+* [#6821](https://github.com/rubocop-hq/rubocop/pull/6821): Fix false negative for Rails/LinkToBlank when `_blank` is a symbol. ([@Intrepidd][])
 * [#6699](https://github.com/rubocop-hq/rubocop/issues/6699): Fix infinite loop for `Layout/IndentationWidth` and `Layout/IndentationConsistency` when bad modifier indentation before good method definition. ([@koic][])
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Bug fixes
 
+* [#6821](https://github.com/rubocop-hq/rubocop/pull/6821): Fix false positive for Rails/LinkToBlank when _blank is a symbol. ([@Intrepidd][])
```
koic: What is not able to find a problem is "false negative" (not "false positive") :-)

Intrepidd: Ok I thought as a false positive as in "your code is ok" when it is not.

I'll update my PR right away

Thanks !

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Bug fixes
 
+* [#6821](https://github.com/rubocop-hq/rubocop/pull/6821): Fix false negative for Rails/LinkToBlank when _blank is a symbol. ([@Intrepidd][])
```
koic: This is a nitpick. It seems that `_` intents italic start if `_blank` is not enclosed in backticks.

```diff
-* [#6821](https://github.com/rubocop-hq/rubocop/pull/6821): Fix false negative for Rails/LinkToBlank when _blank is a symbol. ([@Intrepidd][])
+* [#6821](https://github.com/rubocop-hq/rubocop/pull/6821): Fix false negative for Rails/LinkToBlank when `_blank` is a symbol. ([@Intrepidd][])
```

Intrepidd: Thanks, fixed

koic:
Thanks!

PR #6824, opened 2019-03-10
[Fix #5184] Add new multiline element line break cops

### LINKED ISSUE #5184: Enforce one parameter per line when method parameters span more than one line
Not sure if this should be considered a bug in the `Layout/AlignParameters` cop, a new style for it, or a separate cop, but I'd certainly like this to be signaled as a style offense.

Given the following code

```ruby
a(1,
  2, 3,
  4)
```

## Expected behavior

I would expect rubocop to tell me

```
Layout/AlignParameters: Align the parameters of a method call if they span more than one line.
```

for the third parameter, since it's not aligned with the rest of them.

## Actual behavior

No offenses are registered.

```shell
$ echo "
> a(1,
>   2, 3,
>   4)" | bundle exec rubocop --stdin hola.rb
Inspecting 1 file
.

```

## RuboCop version

```
$ rubocop -V
0.51.0 (using Parser 2.4.0.2, running on ruby 2.4.2 x86_64-linux)
```

maxh:
Fixes https://github.com/rubocop-hq/rubocop/issues/5184 by adding support for autocorrecting cops that ensure that individual elements in multiline arrays, hashes, and method calls each start on separate lines.

```rb
# bad:
[
  a, b,
  c
]

# good:
[
  a,
  b,
  c
]
```

-----------------

* [x] Added tests.


TSMMark:
This looks great! Thank you for doing this. I managed to find some time today to run a mass auto-correct from this branch on our medium sized ruby codebase. It works very well!

I did notice one thing, which I'm not sure if it is intended or not. Ptal:

```ruby
# bad:
hash[:foo] = {
  :bar => true,
  :baz => false,
}
```

```ruby
# good:
hash[:foo] =
  {
    :bar => true,
    :baz => false,
  }
```

I'm fine with this ruling but wanted to point it out in case it was unintentional.

maxh:
Thanks for trying it out! I've seen that as well. Just wrote some quick unit tests locally and confirmed this cop is working as intended:

image

I think it might be due to unexpected interaction during autocorrect with this cop:

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/FirstHashElementLineBreak

Two questions:

1) If you disable that cop, does the issue still appear? My hypothesis: no.
2) If you manually fix it to the "bad" case, does the code get autocorrected back to the good? My hypothesis: no, the "bad" is allowed after manual fixing.

TSMMark:
@maxh I will have to try that when I get a chance, but if your local unit test passes then I'm confident you're right. Good to know thank you

deivid-rodriguez:
Thanks so much for adding this @maxh, I really like this new family of cops! :)

bbatsov:
Nicely done! And sorry for the slow turnaround here!

PR #6826, opened 2019-03-11
Add new `Layout/IndentMultilineClosingBrace` cop

maxh:
Add new `Layout/IndentMultilineClosingBrace` cop for aligning the closing brace of multi-line arrays, hashes, and method calls.

```
 # bad
foo(
  a,
  b
               )
      
# good
foo(
  a,
  b
)
```

-----------------

* [x] Added tests.


bbatsov:
@jonas054 Didn't we have something like this already?

jonas054:
Yes, there's `Layout/ClosingParenthesisIndentation`. Look:

```
rubocop -v && rubocop
0.65.0
Inspecting 1 file
C

Offenses:

file.rb:5:16: C: Layout/ClosingParenthesisIndentation: Indent ) to column 0 (not 15)
```

maxh:
Ah, I remember now. That cop wasn't yielding the behavior we wanted so we wrote our own. The proper way to approach this would be update that cop to support our use case. Noted; closing this.

maxh:
I'd like to reopen this for consideration.

The reason we added this cop internally is because `Layout/ClosingParenthesisIndentation` does not work on hash literals (`}`) or array literals (`]`), though conceptually their formatting needs are similar to parenthesis in method calls and expressions (`)`).

I propose that we replace the existing cop with this new one. The behavior of the old cop is a subset of the new behavior, with the following exception:

```
# good: when x follows opening parenthesis, align parentheses
a = b * (x +
         y
        )
```

In the new cop, "good" means "the closing brace of a multi-line expression aligns with the beginning of the first line containing the expression". One option would be to make this behavior configurable.

Before I move forward, I'd like to get some feedback on this idea. I'm not sure what the process is for deprecating cops.

What do you think?

jonas054:
@maxh Deprecation is tricky. It involves extra code to maintain, and it requires users to update their configuration.

I think a better idea is to extend the cops we have, and add new ones if something is missing.

- `Layout/ClosingParenthesisIndentation` mostly works the way you want for `)`, so add an optional style for your preference as you suggested.
- `Layout/IndentHash` works for `}`, but doesn't support the same cases as for `)`, so look at that cop and fix it.
- There's currently no checking for indentation of `]`, so a new cop would be in order there.

maxh:
@jonas054 sounds good! Thanks for the suggestions. I'll probably create a mixin to share the logic across these three. It bothers me there won't be symmetry among the cop names, but I suppose that's better than forcing users to change their configurations :)

Stay tuned.

maxh:
I've created a new PR to fix https://github.com/rubocop-hq/rubocop/issues/5975 ("Layout/ClosingParenthesisIndentation is incorrect in 0.57.1 even on basic case").

For the most part, the combination of cops below is able achieve to the style we are looking for.

```
Layout/ClosingParenthesisIndentation:
  Enabled: true

Layout/FirstParameterIndentation:
  Enabled: true
  EnforcedStyle: consistent

Layout/IndentArray:
  Enabled: true
  EnforcedStyle: consistent

Layout/IndentHash:
  Enabled: true
  EnforcedStyle: consistent
```

Only after further experimentation did I discover that `FirstParameterIndentation` would impact the behavior of `ClosingParenthesisIndentation`. I've added a note to the docs to explain that in the PR linked above. Thanks!

Issue #5470, opened 2018-01-15
NodePattern fails extracting node types when presented with nil


shalecraig:
### Ruby Code:
A short test case for this problem:
```ruby
pattern = RuboCop::NodePattern.new('(send ($send ...) _method)')
source = 'a' # "(send nil :a)"
node = RuboCop::ProcessedSource.new(source, RUBY_VERSION.to_f, nil).ast
type = pattern.match(node)
```
#### Expected behavior

`pattern.match(node)` returns `nil`

#### Actual behavior

`pattern.match(node)` raises:

```
NoMethodError: undefined method `type' for nil:NilClass
```

### Test variant that works:
```ruby
pattern = RuboCop::NodePattern.new('(send $(send ...) _method)')
source = 'a' # "(send nil :a)"
node = RuboCop::ProcessedSource.new(source, RUBY_VERSION.to_f, nil).ast
result = pattern.match(node)
type = result&.type
```

#### Why this is relevant:

I'm trying to write a `NodePattern` that captures the difference between a `send` and `csend`.

#### Workaround:
For now, I can avoid this problem by:
* Specifying `RuboCop::NodePattern.new('(send $({send csend} ...) _method)')` instead of `RuboCop::NodePattern.new('(send (${send csend} ...) _method)')`
* Extracting the node's type based on `result.type`

## RuboCop version

```
$ bundle exec rubocop -V
0.49.1 (using Parser 2.4.0.2, running on ruby 2.4.1 x86_64-darwin16)
```

Drenmi:
Labelling this as a feature request, as I don't think this is something that was intended to work. 🙂

thomthom:
I'm running into similar error:

The custom cop I'm working on;

```ruby
        def_node_matcher :sketchup_require, <<-PATTERN
          (send
            (const nil? :Sketchup) :require
            (str $_)
          )
        PATTERN

        def_node_matcher :sketchup_require?, <<-PATTERN
          (send
            (const nil? :Sketchup) :require
            (str _)
          )
        PATTERN


        def_node_matcher :sketchup_extension_new, <<-PATTERN
          (:send
            (:const nil? :SketchupExtension) :new
            _
            (str $_))
        PATTERN

        def_node_search :sketchup_extension_new?, <<-PATTERN
          (:send
            (:const nil? :SketchupExtension) :new
            _
            (str _))
        PATTERN


        def on_send(node)
          if sketchup_require?(node)
            filename = sketchup_require(node)
[...]

PR #6828, opened 2019-03-12
NodePattern improvements

marcandre:
This make `NodePattern` a better Ruby citizen:

* provides valuable information with `to_s` / `inspect`
* compatible with `dup`, `YAML` and `Marshal`.
* method `pattern` gives access to the original pattern given when calling `new`
* add comparison (`==`)

### DIFF
--- a/lib/rubocop/node_pattern.rb
+++ b/lib/rubocop/node_pattern.rb
@@ -562,6 +562,14 @@ module RuboCop
       instance_eval(src)
     end
 
+    def marshal_load(pattern)
+      initialize pattern
+    end
+
+    def marshal_dump
+      pattern
+    end
+
     def to_s
       "#<#{self.class} #{pattern}>"
     end
--- a/lib/rubocop/node_pattern.rb
+++ b/lib/rubocop/node_pattern.rb
@@ -133,8 +133,7 @@ module RuboCop
       end
 
       def run(node_var)
-        tokens =
-          @string.scan(TOKEN).reject { |token| token =~ /\A#{SEPARATORS}\Z/ }
+        tokens = Compiler.tokens(@string)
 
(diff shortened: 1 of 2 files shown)


marcandre:
The one environment failing is having trouble with `bundle`; clearly not related to these changes.

lib/rubocop/node_pattern.rb
```diff
@@ -577,6 +577,11 @@ def marshal_dump
       pattern
     end
 
+    def ==(other)
+      other.is_a?(NodePattern) && other.pattern == pattern
```
Drenmi: Node patterns allow arbitrary use of whitespace without changing semantics. We might want to strip out all spaces before comparing them. 🙂

marcandre: Good point :+1:

Stripping all spaces would be too permissive though, as `"12 34"` and `"1 234"` aren't the same. I've reused the right code so it should deal correctly with all whitespace now.

Drenmi:
Thanks @marcandre! All these improvements are really nice. 🙇 

If you're interested in working more on the node patterns, there is an [open issue](#5470).

bbatsov:
Can you fix the name of the `+` prefix commit?

marcandre:
> Can you fix the name of the `+` prefix commit?

Sorry, I don't understand

bbatsov:
You've got a commit message titled "+ Make NodePattern serializable and al.". Likely you didn't mean to put the plus there.

marcandre:
> You've got a commit message titled "+ Make NodePattern serializable and al.". Likely you didn't mean to put the plus there.

Oh, I see where Github shows that in the discussion above, but I've amended the PR since. Check the "commits" tab for the up to date list.

bbatsov:
My bad. Thanks for working on this!

PR #6833, opened 2019-03-14
Add examples for rule PredicateName to better show its behavior 

fuksito:
Makes it more obvious that the rule also applies to the methods that don't end with question mark, but to all the one starting with a specific prefix.

Having both "is_" and "?" question mark is undoubtedly redundant, but example don't reveal other aspect of this rule.

-----------------

* [ ] Added tests.


bbatsov:
You should actually change this in the source file (the manual in generated from the comments in the Ruby files).

PR #6834, opened 2019-03-14
Fix NodePattern raising errors on some inputs

### LINKED ISSUE #5470: NodePattern fails extracting node types when presented with nil

marcandre:
This fixes #5470 

Note: the first two commits are not directly related, but make things a bit easier to debug.

### DIFF
--- a/lib/rubocop/node_pattern.rb
+++ b/lib/rubocop/node_pattern.rb
@@ -180,9 +180,14 @@ module RuboCop
         # temp variable as 'cur_node'
         with_temp_node(cur_node) do |init, temp_node|
           terms = compile_seq_terms(tokens, temp_node)
+          terms.unshift(compile_guard_clause(temp_node))
 
           join_terms(init, terms, " &&\n")
         end
+      end
+
+      def compile_guard_clause(cur_node)
+        "#{cur_node}.is_a?(RuboCop::AST::Node)"
       end
 
       def compile_seq_terms(tokens, cur_node)
@@ -444,12 +449,6 @@ module RuboCop
         params.empty? ? '' : ",#{params}"
       end
 
-      def emit_guard_clause
-        <<-RUBY
-          return unless node.is_a?(RuboCop::AST::Node)
-        RUBY
-      end
-
       def emit_method_code
         <<-RUBY
           return unless #{@match_code}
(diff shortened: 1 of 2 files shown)


Drenmi:
If you're hungry for more, there's a long standing request to add an "any child" matcher. This would be especially helpful when matching on hashes. Consider:

```
ruby-parse -e '{ foo: 1, bar: 2 }'

(hash
  (pair
    (sym :foo)
    (int 1))
  (pair
    (sym :bar)
    (int 2)))
```

We may be interested in hashes that have a `foo: 1` pair, but the pairs can appear in any order, so we can't reliably match that in a pattern. 🙂 

What is needed is a suggested extension to the syntax, and the actual implementation.

marcandre:
@Drenmi  Is there an issue already open for this? Doesn't seem very hard to do...

Drenmi:
@marcandre I don't think there's an issue, but feel free to open one. 🙂

marcandre:
> @marcandre I don't think there's an issue, but feel free to open one. 🙂

Created #6841

PR #6836, opened 2019-03-14
[Fix #6746] Avoid offense on `$stderr.puts` with no arguments.

### LINKED ISSUE #6746: Style/StderrPuts isn't the same thing when there is no message
## Expected behavior

Not to replace `$stderr.puts` (with no argument) with `warn` (with no argument) because they don't do the same thing - the former prints a new line, the latter prints nothing.

## Actual behavior

`$stderr.puts` (with no argument) is replaced with `warn` (with no argument).

A file with just `$stderr.puts` (with no argument) in, compared to a file with just `warn` (with no argument) in. They don't do the same thing, but Rubocop suggests I swap from the former to the latter, and does so automatically if asked to fix.

This breaks real specs I have.

## RuboCop version

```
$ rubocop -V
0.63.1 (using Parser 2.6.0.0, running on ruby 2.6.1 x86_64-darwin18)
```

luciamo:
This PR closes issue #6746. The applied changes will avoid offense on `$stderr.puts` with no arguments.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/stderr_puts.rb
+++ b/lib/rubocop/cop/style/stderr_puts.rb
@@ -22,7 +22,7 @@ module RuboCop
 
         def_node_matcher :stderr_puts?, <<-PATTERN
           (send
-            (gvar #stderr_gvar?) :puts
+            (gvar #stderr_gvar?) :puts $_
             ...)
         PATTERN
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/stderr_puts.rb
```diff
@@ -29,6 +29,9 @@ class StderrPuts < Cop
         def on_send(node)
           return unless stderr_puts?(node)
 
+          arguments = node.arguments
```
bbatsov: I'd update the pattern instead. It would make the code more compact.

luciamo: Thanks for the reply @bbatsov! I changed the pattern like you suggested. Could you check again? 😉

bbatsov:
Thanks!

PR #6839, opened 2019-03-16
[Fix #4321] Allow `self` to receive methods also defined on `Kernel`

### LINKED ISSUE #4321: Style/RedundantSelf confused when `extending` a scope
## Expected behavior

```ruby
module MyClassQuery
  def initialize(scope: MyClass.all)
    @scope = scope.extending(Scopes)
  end

  module Scopes
    def active
      self.open.where(active: true)
    end
  end
end

class MyClass  < ActiveRecord::Base
  scope :open, -> { where(open: true) }
end
```

The `self.open` should not trigger the redundant self cop.  Without the `self` ruby calls https://ruby-doc.org/core-2.1.4/File.html#method-c-open, with the `self` it calls `MyClass.instance_method(:open)` which is the intended behavior.

## Actual behavior

The cop is triggered

See Expected behavior

## RuboCop version

```
$ rubocop -V
0.48.1 (using Parser 2.4.0.0, running on ruby 2.3.1 x86_64-linux)
```

mikegee:
Method calls with an explicit receiver of `self` are ok when a method with the same name is also defined on `Kernel`. Rubocop can't know if the explicit `self` is disambiguating a method call or not.

Fixes #4321 

(Perhaps there should be another cop warning about defining methods that are also defined on `Kernel`? 🤔 )

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_self.rb
+++ b/lib/rubocop/cop/style/redundant_self.rb
@@ -43,6 +43,7 @@ module RuboCop
       #   end
       class RedundantSelf < Cop
         MSG = 'Redundant `self` detected.'.freeze
+        KERNEL_METHODS = Kernel.methods(false)
 
         def self.autocorrect_incompatible_with
           [ColonMethodCall]
@@ -117,7 +118,8 @@ module RuboCop
 
         def allowed_send_node?(node)
           @allowed_send_nodes.include?(node) ||
-            @local_variables_scopes[node].include?(node.method_name)
+            @local_variables_scopes[node].include?(node.method_name) ||
+            KERNEL_METHODS.include?(node.method_name)
         end
 
         def regular_method_call?(node)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Bug fixes
 
+* [#4321](https://github.com/rubocop-hq/rubocop/pull/4321): Fix false offense for Style/RedundantSelf when the method is also defined on `Kernel`. ([@mikegee][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/redundant_self.rb
```diff
@@ -43,6 +43,7 @@ module Style
       #   end
       class RedundantSelf < Cop
         MSG = 'Redundant `self` detected.'.freeze
+        KERNEL_METHODS = Kernel.methods(false)
```
bbatsov: The problem with this approach is that the methods you'll get might not match the target Ruby version (e.g. if you're running RuboCop with a different version of Ruby).

bbatsov:
Apart from my small remark, I think that's a good improvement. 

> (Perhaps there should be another cop warning about defining methods that are also defined on Kernel? 🤔 )

Yeah, I totally agree. No idea why we haven't implemented this yet. Should be super simple.

lib/rubocop/cop/style/redundant_self.rb
mikegee: Good point. Do we need to maintain a list for each version? I’ll try to think of something simpler.

bbatsov: I'd rather avoid that. I was just pointing out that this solution is not completely reliable.

bbatsov:
I just realized it would have also been nice to update the docs/examples.

mikegee:
I figured this was an edge case that didn’t need the visibility but I didn’t think about it much.

bbatsov:
Well, that's not that different from having local variables named the same way as a method. :-)

PR #6847, opened 2019-03-19
Allow chained Hash#[] when BlockDelimiters braces_for_chaining is true

att14:
Sometimes the return of a block is a Hash and Ruby allows you to call `[]` without a dot. This means only checking `dot?` in `return_value_chaining?` does not allow chaining Hash element reference. I wasn't sure if checking the explicit method name was the best choice or if there is something that I missed in the documentation.

I did not add an entry to the Changelog in order to get feedback first. If this is an acceptable change I will do so.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/block_delimiters_spec.rb
+++ b/spec/rubocop/cop/style/block_delimiters_spec.rb
@@ -490,6 +490,45 @@ RSpec.describe RuboCop::Cop::Style::BlockDelimiters, :config do
       RUBY
     end
 
+    it 'allows when :[] is chained' do
+      expect_no_offenses(<<-RUBY.strip_indent)
+        foo = [{foo: :bar}].find { |h|
+          h.key?(:foo)
+        }[:foo]
+      RUBY
+    end
+
+    it 'allows do/end inside Hash[]' do
+      expect_no_offenses(<<-RUBY.strip_indent)
+        Hash[
+          {foo: :bar}.map do |k, v|
+            [k, v]
+          end
+        ]
+      RUBY
+    end
+
+    it 'allows chaining to } inside of Hash[]' do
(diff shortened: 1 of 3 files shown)


att14:
This doesn't work for the case,

```
        Hash[
          {foo: :bar}.map do |k, v|
            [k, v]
          end
        ]
```
 Maybe someone with more knowledge of the AST could point me in the right direction for fixing this.

att14:
`chained?` appears to work and the tests are passing. Is there any reason not to use `chained?` instead of `return_value_chaining?`?

att14:
Is anyone able to look at this? I'm pretty sure this is a bug. My company has been using this for some time with no issues.

bbatsov:
@att14 Sorry about the slow response - huge backlog of other tasks... :-( Rebase and add a changelog entry and we're good to go.

spec/rubocop/cop/style/block_delimiters_spec.rb
```diff
@@ -490,6 +490,36 @@
       RUBY
     end
 
+    it 'allows when :[] is chained' do
+      expect_no_offenses(<<-RUBY.strip_indent)
+        foo = [{foo: :bar}].find { |h|
+          h.key?(:foo)
+        }[:foo]
+      RUBY
+      expect_no_offenses(<<-RUBY.strip_indent)
```
bbatsov: This example doesn't match the `it` description. In general I'd have one it per example.

att14: 👌

att14:
Any ideas on why the 2.2 build is failing?

bbatsov:
Seems that the newest version of byebug dropped support for Ruby 2.2.

att14:
:+1: thank you!

PR #6848, opened 2019-03-20
Add `Rails/RedundantAllowNil` cop

elebow:
This cop checks for a situation in a Rails model validation where the value of `allow_blank` makes `allow_nil` redundant.

In Rails, `#blank?` includes nil checking. When both options are set to the same value, `allow_nil` is entirely subsumed within `allow_blank`.

When `allow_nil: false, allow_blank: true`, again `allow_blank` controls because Rails performs an `||` between the two. See https://github.com/rails/rails/blob/d5699198a45a91250e1adb3ed899b0b46b4ac879/activemodel/lib/active_model/validator.rb#L151

When `allow_nil: true, allow_blank: false`, both options are meaningful. In this case, a value of `nil` would be valid, but a value of `''` would not be.

-----------------

* [x] Added tests.


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 ### New features
 
+* Add new cop `Rails/RedundantAllowNil` that checks for cases when `allow_blank` makes `allow_nil` unnecessary in model validations. ([@elebow][])
```
bbatsov: Please move this to the proper section of the changelog. 0.65 was released a while ago.

lib/rubocop/cop/rails/redundant_allow_nil.rb
```diff
@@ -0,0 +1,104 @@
...
+      #
+      #   # good
+      #   # Here, `nil` is valid but `''` is not
+      #   validates :x, length: { is: 5 }, allow_nil: true, allow_blank: false
+      #
+      class RedundantAllowNil < Cop
+        include RangeHelp
+
+        MSG_SAME =
+          'allow_nil is redundant when allow_blank has the same value'.freeze
```
bbatsov: Surround identifiers with backticks and end messages with `.`.

bbatsov:
Apart from my small remarks your changes look good.

config/default.yml
```diff
@@ -2488,6 +2488,13 @@ Rails/ReadWriteAttribute:
   Include:
     - app/models/**/*.rb
 
+Rails/RedundantAllowNil:
+  Description: >-
+                 Finds redundant use of `allow_nil` when `allow_blank` is set to
+                 certain values in model validations.
+  Enabled: true
+  VersionAdded: '0.67'
```
bbatsov: I think you should limit this check to `app/models/**/*.rb`.

elebow:
Updated to address comments, and rebased on latest master.

bbatsov:
Thanks!

PR #6849, opened 2019-03-20
Add `STDERR.puts` check to `Style/StderrPuts` cop

elebow:
`STDERR` is often equal to `$stderr` in Ruby programs. Writing to either one should be replaced by `warn`.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/stderr_puts.rb
+++ b/lib/rubocop/cop/style/stderr_puts.rb
@@ -17,12 +17,14 @@ module RuboCop
       class StderrPuts < Cop
         include RangeHelp
 
-        MSG = 'Use `warn` instead of `$stderr.puts` to allow such output ' \
-              'to be disabled.'.freeze
+        MSG =
+          'Use `warn` instead of `%<bad>s` to allow such output to be disabled.'
+          .freeze
 
         def_node_matcher :stderr_puts?, <<-PATTERN
           (send
-            (gvar #stderr_gvar?) :puts $_
+            {(gvar #stderr_gvar?) (const nil? :STDERR)}
+            :puts $_
             ...)
         PATTERN
 
@@ -39,6 +41,10 @@ module RuboCop
         end
 
         private
+
+        def message(node)
+          format(MSG, bad: "#{node.receiver.source}.#{node.method_name}")
+        end
 
         def stderr_gvar?(sym)
           sym == :$stderr
--- a/spec/rubocop/cop/style/stderr_puts_spec.rb
+++ b/spec/rubocop/cop/style/stderr_puts_spec.rb
@@ -12,7 +12,7 @@ RSpec.describe RuboCop::Cop::Style::StderrPuts do
     RUBY
   end
 
-  it "autocorrects `warn('hello')`" do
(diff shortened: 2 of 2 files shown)


pocke:
link: https://github.com/rubocop-hq/rubocop/pull/4813#discussion_r142003554
We have discussed it, but I guess the new cop that is discussed in the link has not been implemented yet.

@koic what do you think about this pull-request?

spec/rubocop/cop/style/stderr_puts_spec.rb
```diff
@@ -8,11 +8,11 @@
   it "registers an offense when using `$stderr.puts('hello')`" do
     expect_offense(<<-RUBY.strip_indent)
       $stderr.puts('hello')
-      ^^^^^^^^^^^^ Use `warn` instead of `$stderr.puts` to allow such output to be disabled.
+      ^^^^^^^^^^^^ Use `warn` instead of `$stderr.puts` or `STDERR.puts` to allow such output to be disabled.
```
jaredbeck: A very minor thing, but adding the word "either" here might help ("instead of either").

spec/rubocop/cop/style/stderr_puts_spec.rb
```diff
@@ -23,4 +23,23 @@
...
+    new_source = autocorrect_source("STDERR.puts('hello')")
+
+    expect(new_source).to eq "warn('hello')"
+  end
+
+  it 'registers no offense when using `STDERR.puts` with no arguments' do
+    expect_no_offenses(<<-RUBY.strip_indent)
+      STDERR.puts
+    RUBY
+  end
```
jaredbeck: I believe that `STDERR.puts` with no arguments *should* register an offense. Printing a newline should be no different that printing anything else. Perhaps this should be discussed in a separate PR, though?

spec/rubocop/cop/style/stderr_puts_spec.rb
elebow: Thanks. Text updated.

spec/rubocop/cop/style/stderr_puts_spec.rb
elebow: Agreed on both counts. I can't think of a situation where it would be useful to allow `STDERR.puts` without arguments but prohibit it with arguments. But for this PR, I think it's best to follow the example of the legacy logic.

lib/rubocop/cop/style/stderr_puts.rb
```diff
@@ -17,12 +17,13 @@ module Style
       class StderrPuts < Cop
         include RangeHelp
 
-        MSG = 'Use `warn` instead of `$stderr.puts` to allow such output ' \
-              'to be disabled.'.freeze
+        MSG = 'Use `warn` instead of either `$stderr.puts` or `STDERR.puts` ' \
```
bbatsov: I think in the message we should show exactly what we detected (meaning we should make the message dynamic).

elebow:
Updated to address comments.

bbatsov:
Thanks!

PR #6851, opened 2019-03-20
Fix NodePattern of node type on a literal

marcandre:
Without this fix, one can get errors raised like:

```
     NoMethodError:
       undefined method `send_type?' for :bar:Symbol
```


Drenmi:
This is great! Thank you, @marcandre. 🙇

PR #6853, opened 2019-03-21
Add new cop `Rails/ActiveRecordOverride`

elebow:
This cop checks for cases where an Active Record method like `save`, `update`, etc is overridden and includes a `super` statement. Instead, the user should use Active Record callbacks.

---

There is a limitation in this PR currently, where the cop will not register any offense if the class in question inherits indirectly from `ApplicationRecord` or `ActiveModel::Base`.  For example:

```ruby
class Book < ApplicationRecord
end

class Novel < Book
  # this should be an offense, but it will not be detected
  def update
    this.x = 5
    super
  end
end
```
It would be much more complicated to evaluate all the parent classes looking for  `ApplicationRecord` or `ActiveModel::Base`, so it seems more practical to accept this limitation. I'm open to advice, of course.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2229,6 +2229,15 @@ Rails/ActiveRecordAliases:
   Enabled: true
   VersionAdded: '0.53'
 
+Rails/ActiveRecordOverride:
+  Description: >-
+                  Check for overriding Active Record methods instead of using
+                  callbacks.
+  Enabled: true
+  VersionAdded: '0.67'
+  Include:
+    - app/models/**/*.rb
+
 Rails/ActiveSupportAliases:
   Description: >-
                   Avoid ActiveSupport aliases of standard ruby methods:
--- a/lib/rubocop/cop/rails/active_record_override.rb
+++ b/lib/rubocop/cop/rails/active_record_override.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Rails
+      # Checks for overriding built-in Active Record methods instead of using
+      # callbacks.
+      #
+      # @example
+      #   # bad
+      #   class Book < ApplicationRecord
+      #     def save
+      #       self.title = title.upcase!
+      #       super
+      #     end
+      #   end
+      #
+      #   # good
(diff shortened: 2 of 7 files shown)


lib/rubocop/cop/rails/active_record_override.rb
```diff
@@ -0,0 +1,49 @@
...
+      #   class X < ApplicationRecord
+      #     def save
+      #       self.title = title.upcase!
+      #       super
+      #     end
+      #   end
+      #
+      #   # good
+      #   class X < ApplicationRecord
+      #     before_save upcase_title
```
bbatsov: I think you missed a `:` here.

lib/rubocop/cop/rails/active_record_override.rb
```diff
@@ -0,0 +1,49 @@
...
+      #
+      #     def upcase_title
+      #       self.title = title.upcase!
+      #     end
+      #   end
+      #
+      class ActiveRecordOverride < Cop
+        MSG = 'Use Active Record callbacks instead of overriding methods.'
```
bbatsov: This message could be a bit more specific - Use X callback instead of overriding Y method.

bbatsov:
The changes look good (apart from my small remarks), but your branch has to be rebased on top of the current `master` branch due to merge conflicts.

config/default.yml
```diff
@@ -2234,6 +2234,13 @@ Rails/ActiveRecordAliases:
   Enabled: true
   VersionAdded: '0.53'
 
+Rails/ActiveRecordOverride:
+  Description: >-
+                  Check for overriding Active Record methods instead of using
+                  callbacks.
+  Enabled: true
+  VersionAdded: '0.67'
```
bbatsov: Don't forget to limit the cop to `app/controllers/**/*.rb`.

elebow:
Updated to address comments, and rebased on latest master.

config/default.yml
elebow: It should actually be `app/models/**/*.rb`.

bbatsov:
Thanks!

PR #6854, opened 2019-03-21
Clarify dangers of Rails/LexicallyScopedActionFilter

urbanautomaton:
This cop insists that users define action methods in the same controller class as any `before_`, `after_` or `around_action`s that refer to the action.

However, the examples given don't account for the fact that behaviour might be defined in the superclass. Indeed, this is one of the main use cases for action decorators: customisations can be applied to superclass behaviour without redefining the whole action in the subclass.

If this is the case, then the empty action definitions shown in the docs will override the desired behaviour, and lead users to introduce bugs.

```ruby
class ContentController < ApplicationController
  def update
    @content.update(content_attributes)
  end
end

class ArticlesController < ContentController
  before_action :load_article, only: [:update]

  # the cop says add this, but it breaks the update
  def update
  end

  private

  def load_article
    @content = Article.find(params[:article_id])
  end
end
```

In this PR I've tried to emphasise this in the docs, and have added an example to encourage users to think about the inheritance chain. The example is slightly awkward because the superclass in this case is `ApplicationController`, which is pretty unlikely to define actions in most apps; I wasn't sure about this, but also felt that introducing a different named superclass might also be confusing. An alternative might be to add the above snippet as a third, more realistic example case. What do people think?

I'd also be interested in people's thoughts about whether this cop should be disabled by default. My view is that it encourages semantically meaningful changes, and is therefore quite risky no matter how clear the documentation is, so I'd be tempted to disable it.

-----------------

* [ ] ~Added tests.~
* [ ] ~Added an entry to the [Changelog](https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format).~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 ### Changes
 
+* [#6854](https://github.com/rubocop-hq/rubocop/pull/6854): Mark Rails/LexicallyScopedActionFilter as unsafe and document risks. ([@urbanautomaton][])
 * [#5977](https://github.com/rubocop-hq/rubocop/issues/5977): Warn for Performance Cops. ([@koic][])
 * [#6637](https://github.com/rubocop-hq/rubocop/issues/6637): Move `LstripRstrip` from `Performance` to `Style` department and rename it to `Strip`. ([@anuja-joshi][])
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -2445,6 +2445,7 @@ Rails/LexicallyScopedActionFilter:
   Description: "Checks that methods specified in the filter's `only` or `except` options are explicitly defined in the controller."
   StyleGuide: 'https://github.com/rubocop-hq/rails-style-guide#lexically-scoped-action-filter'
   Enabled: true
+  Safe: false
   VersionAdded: '0.52'
   Include:
     - app/controllers/**/*.rb
--- a/lib/rubocop/cop/rails/lexically_scoped_action_filter.rb
+++ b/lib/rubocop/cop/rails/lexically_scoped_action_filter.rb
@@ -6,10 +6,13 @@ module RuboCop
       # This cop checks that methods specified in the filter's `only` or
       # `except` options are defined within the same class or module.
       #
-      # You can technically specify methods of superclass or methods added
-      # by mixins on the filter, but these confuse developers. If you
-      # specify methods that are defined in other classes or modules, you
-      # should define the filter in that class or module.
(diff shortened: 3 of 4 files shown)


urbanautomaton:
Oh - should I have opened this on [rubocop-rails](https://github.com/rubocop-hq/rubocop-rails/issues/47) instead? Happy to reopen there if so.

wata727:
Interesting. I wasn't considering the possibility of this cop would such mislead.

> An alternative might be to add the above snippet as a third, more realistic example case. What do people think?

I agree with this idea :+1: 

> I'd also be interested in people's thoughts about whether this cop should be disabled by default. My view is that it encourages semantically meaningful changes, and is therefore quite risky no matter how clear the documentation is, so I'd be tempted to disable it.

Hmm... This is a very difficult problem. I believe that this Cop will help a lot, but we should avoid introducing such bugs. One idea is to make this cop as unsafe.

urbanautomaton:
@wata727 Thanks for the feedback - I've added the inheritance example as a separate case.

What does marking the cop unsafe involve?

wata727:
Cops marked as unsafe will not run when executed with the `--safe` flag. Originally, it seems to indicate a Cop which can yield false positives, but it may be suitable for such a case. See https://github.com/rubocop-hq/rubocop/issues/5978#issuecomment-395958738

urbanautomaton:
Ah okay, interesting - thanks very much for the extra context. 👍

As you say, we'd be changing the definition of "safety" somewhat if we used that here, and while I certainly feel that the suggestions this cop makes are "unsafe" in some general sense, I'd be interested  in whether other people think this flag is an appropriate way to indicate that. @bbatsov, what do you think about using the "Safe: false" flag for cops that suggest semantically meaningful edits, as well as those that emit false positives?

(On a side note: the expanded `default.yml` metadata and elimination of the enabled / disabled configs is a great improvement, thank you to those who worked on it. 😄)

bbatsov:
> @bbatsov, what do you think about using the "Safe: false" flag for cops that suggest semantically meaningful edits, as well as those that emit false positives?

That's what we usually do. I guess we should mark this one as unsafe indeed.

urbanautomaton:
Okay, thanks very much both - I've marked the cop as unsafe and updated the changelog. From my perspective this is ready to go, but please let me know if you'd like anything changed.

bbatsov:
Thanks!

PR #6859, opened 2019-03-22
[Fix #6858] Fix an incorrect auto-correct for `Lint/ToJSON`

### LINKED ISSUE #6858: Lint/ToJSON suggest wrong code
## Expected behavior

With following suggestion at `Lint/ToJSON`, not raise error.
https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/lint/to_json.rb#L17

```
@example
  # bad
  def to_json
  end

  # good
  def to_json(_opts)
  end
```

## Actual behavior

Because I get `Lint/ToJSON`, I changed my code like below

```rb
# Before
class Ex
  def to_json
    {a: 1}.to_json
  end
end

# After
class Ex
  def to_json(_opts)
    {a: 1}.to_json
  end
end
```

Then, `Ex.new.to_json` start raising error saying `ArgumentError: wrong number of arguments (given 0, expected 1)`

As talked https://github.com/rubocop-hq/rubocop/pull/6379#discussion_r236021167, document should suggest the argument to be optional

## RuboCop version

```
$ bundle exec rubocop -V
0.66.0 (using Parser 2.6.0.0, running on ruby 2.4.1 x86_64-linux)
```

koic:
Fixes #6858.

I think it was a mistake of multiple arguments.

## Before

```ruby
def to_json(_opts)
end
```

## After

```ruby
def to_json(*args)
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/to_json.rb
+++ b/lib/rubocop/cop/lint/to_json.rb
@@ -14,7 +14,7 @@ module RuboCop
       #   end
       #
       #   # good
-      #   def to_json(_opts)
+      #   def to_json(*_args)
       #   end
       #
       class ToJSON < Cop
@@ -29,7 +29,10 @@ module RuboCop
 
         def autocorrect(node)
           lambda do |corrector|
-            corrector.insert_after(node.loc.name, '(_opts)')
+            # The following used `*_args` because `to_json(*args)` has
+            # an offense of `Lint/UnusedMethodArgument` cop if `*args`
+            # is not used.
+            corrector.insert_after(node.loc.name, '(*_args)')
           end
         end
       end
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/lint/to_json.rb
```diff
@@ -29,7 +29,7 @@ def on_def(node)
 
         def autocorrect(node)
           lambda do |corrector|
-            corrector.insert_after(node.loc.name, '(_opts)')
+            corrector.insert_after(node.loc.name, '(*args)')
```
pocke: `def to_json(*args)` has an offense of `Lint/UnusedMethodArgument` cop if `args` is not used.
So I think it should be named `_args`.

koic: Sure! I updated it and added a source code comment.

PR #6862, opened 2019-03-25
Allow Rails associations to reference `class_name: self`

avit:
When used in concern mixins, the class name that the association is included in
may be declared dynamically, instead of a static string.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rails/reflection_class_name.rb
+++ b/lib/rubocop/cop/rails/reflection_class_name.rb
@@ -13,6 +13,7 @@ module RuboCop
       #
       #   # good
       #   has_many :accounts, class_name: 'Account'
+      #   has_many :children, class_name: self.name
       class ReflectionClassName < Cop
         MSG = 'Use a string value for `class_name`.'.freeze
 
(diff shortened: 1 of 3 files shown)


mikegee:
seems legit 👍

koic:
Interesting. However the following error occurs in Rails 5.2 using `class_name: self`. I think it is not good to not register an offense that is disallowed in the new Rails. It will be a Rails application that won't work in the future.

```console
/Users/koic/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/activerecord-5.2.3/
lib/active_record/reflection.rb:436:in `initialize': A class was passed to `:class_name` 
but we are expecting a string. (ArgumentError)
```

On the other hand, if you can suggest this issue to rails/rails repo and accept it for the Rails team, I'd like to merge this suggestion.

bbatsov:
I agree with @koic.

avit:
Thanks, I hadn't noticed this new change. I will submit a PR to rails/rails and we can decide here based on the outcome.

avit:
Thinking about this further, I don't think there's any reason for Rails to change their validation: they expect a String, and that's that. 

I think there are other valid ways to pass a string, other than a string literal, such as `self.name`, so I'm suggesting that instead.

The new commit rejects any class constant references. This matches the Rails logic.

bbatsov:
We're in the process of moving all Rails-related functionality to a standalone library (gem) named `rubocop-rails`. Please, migrate this issue to its issue tracker https://github.com/rubocop-hq/rubocop-rails/issues/

Issue #5977, opened 2018-06-09
Extract Performance Cops into a separate gem


bbatsov:
Performance cops have to extracted into a standalone gem (named something like rubocop-performance or whatever).

The reason for this decision is simple. I want RuboCop to be focused just on Ruby and have no framework/library/runtime dependant pieces for the sake of being easier to focus on evolving it. Creating more extensions should also help us come up with a good official API for them. Performance cops are all MRI focused and are highly dependent of the version of MRI you're using, so they don't belong in the core RuboCop. 

Any volunteers to do this? I was always ambivalent towards the performance checks (as they often come at the expense of readability), so I'm not particularly interested in working on this.

Drenmi:
> Any volunteers to do this? I was always ambivalent towards the performance checks (as they often come at the expense of readability), so I'm not particularly interested in working on this.

Agreed. Language level optimizations also suffer from being subject to the current VM implementation. So a big task for whoever wants to take this would be to benchmark on every Ruby version and see that the improvements still hold.

bbatsov:
Although they just extract it as-is in the beginning and improve things later. Probably performance cops should have explicit versions on which they make sense as part of their config.

gizotti:
Extracting as it is into a gem seems simple enough. I could handle this one if nobody is on it yet.

bbatsov:
@gizotti Happy to hear this! 

I've created a placeholder repo for the extraction https://github.com/rubocop-hq/rubocop-performance

bbatsov:
Done.

gizotti:
No worries @composerinteralia, I got to where you were last night so not reason to push my stuff if you already did.
I'll leave you to it, I think if we both work in the same branch we might step on each others toes.

koic:
I'm investigating the same git history problem as the current RuboCop Rails repository. The current situation is below.
https://github.com/rubocop-hq/rubocop/issues/5976#issuecomment-425884216

koic:
@bbatsov @composerinteralia 
Cc @rubocop-hq/rubocop-core 

I extracted `lib/rubocop/cop/performance/*` and `spec/rubocop/cop/performance/*` along with git history from RuboCop Core (I used `git cherry-pick` for this extraction) . 
I think that `lib/rubocop/cop/performance` and `spec/rubocop/cop/performance` will be the most important history in maintenance.

The following branch including previous work done in RuboCop Performance repo, and it is synchronized with the latest performance cops of RuboCop Core.
[...]

PR #6870, opened 2019-04-01
[Fix #6637] Promoted Performance/LstripRstrip to Style/LstripRstrip

### LINKED ISSUE #6637: promote Performance/LstripRstrip to Style/LstripRstrip
Unlike many other Performance cops `Performance/LstripRstrip` is not making code less readable and is a good idea independent from performance improvements.

As  #5977 will cause performance cops to be disabled by default I suggest to rescue this one as it is useful to avoid `lstrip.rstrip` where `strip` that is more readable and shorter exists.

It may be also useful to expand description with a note that this change may improve performance (as it would no longer be implied)

------

Other usual issue fields are not applicable - I spotted it during reviewing https://github.com/rubocop-hq/rubocop/blob/master/config/default.yml

anuja-joshi:
Moved `Performance/LstripRstrip` to `Style/LstripRstrip`, as per [#6637](https://github.com/rubocop-hq/rubocop/issues/6637)


bbatsov:
I also wonder if we shouldn't rename this to `Style/Strip`. @rubocop-hq/rubocop-core Any thoughts?

Drenmi:
Plus one for renaming. 👍

bbatsov:
I'll do the renaming myself. 

@anuja-joshi Thanks!

matkoniecz:
@anuja-joshi Thank you for implementing this! I opened #6871 ("promote Performance/RedundantSortBy to Style/RedundantSortBy") as there is at least one more cop that is worth rescuing before #5977 happens.

bbatsov:
FYI - the removal of the performance cops is scheduled to happen 2 releases from now. Next release will deprecate them and point to the new gem. 0.68 is going to delete them.

PR #6872, opened 2019-04-01
[Fix #6761] Make UncommunicativeMethodParamName account for underscores.

### LINKED ISSUE #6761: Naming/UncommunicativeMethodParamName doesn't take into account underscore prefix
I use Ruby in the domain of 3D modelling. So common short variables include `x`, `y`, `z` and `u`, `v`. These are flagged by the default rules of `Naming/UncommunicativeMethodParamName` so I add them to the list of allowed names:

```yml
Naming/UncommunicativeMethodParamName:
  AllowedNames:
    - x
    - y
    - z
    - u
    - v
    - uv
```

However, some times these names are also used as argument names in callbacks of the API. And you don't always use all the arguments being passed in the callback. For instance:

```ruby
 def onLButtonDoubleClick(_flags, _x, _y, _view)
  # ...
end
```

However this then will again flag the `x` and `y` arguments:

```
src/tt_truebend/tools/truebend.rb:195:42: C: Naming/UncommunicativeMethodParamName: Method parameter must be at least 3 characters long.
    def onLButtonDoubleClick(_flags, _x, _y, _view)
```

I could add `_x` and `_y` to the list of allowed names. But this still feels like a bug - where the cop should take into account the convention of using underscore prefix to denote unused arguments.

--------

## Expected behavior

I would expect `Naming/UncommunicativeMethodParamName` to take into account the `_` prefix convention and strip it away when it checks against the list of allowed short names.

## Actual behavior

Currently `_x` is flagged even if `x` is an allowed variable name.

```yml
Naming/UncommunicativeMethodParamName:
  AllowedNames:
    - x
    - y
    - z
```

```ruby
 def onLButtonDoubleClick(_flags, _x, _y, _view)
  # ...
end
```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.63.1 (using Parser 2.6.0.0, running on ruby 2.2.6 x64-mingw32)
```

thomthom:
This change makes `Naming/UncommunicativeMethodParamName` understand the `_`-prefix convention annotation unused parameters. It also ignores parameters named `_`. This allowing usage of API interfaces where you don't use all the parameters provided by the interface.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug fixes
 
+* [#6761](https://github.com/rubocop-hq/rubocop/issues/6761): Make `Naming/UncommunicativeMethodParamName` account for param names prefixed with underscores. ([@thomthom][])
 * [#6855](https://github.com/rubocop-hq/rubocop/pull/6855): Fix an exception in `Rails/RedundantReceiverInWithOptions` when the body is empty. ([@ericsullivan][])
 * [#6856](https://github.com/rubocop-hq/rubocop/pull/6856): Fix auto-correction for `Style/BlockComments` when the file is missing a trailing blank line. ([@ericsullivan][])
 * [#6858](https://github.com/rubocop-hq/rubocop/issues/6858): Fix an incorrect auto-correct for `Lint/ToJSON` when there are no `to_json` arguments. ([@koic][])
@@ -3895,3 +3896,4 @@
 [@ericsullivan]: https://github.com/ericsullivan
 [@aeroastro]: https://github.com/aeroastro
 [@anuja-joshi]: https://github.com/anuja-joshi
+[@thomthom]: https://github.com/thomthom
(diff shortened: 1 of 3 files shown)


thomthom:
Sorry, my bad. I made a last moment change before pushing the PR. I'll clean up the RuboCop warnings.

thomthom:
The checks failing now is not related to this PR. I see this message:

```
spec/rubocop/cop/style/lstrip_rstrip_spec.rb:3:1: C: RSpec/FilePath: Spec path should end with rubocop/cop/style/strip*_spec.rb.
RSpec.describe RuboCop::Cop::Style::Strip do
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

bbatsov:
Rebase on top of `master` and the error will go away.

bbatsov:
Thanks!

PR #6875, opened 2019-04-01
Mention block form of Struct.new in Style/StructInheritance

XrXr:
As the documentation for `Struct` recommends using the block
form over inheriting, I think it is worth mentioning it.

> If a block is given it will be evaluated in the context of
StructClass, passing the created class as a parameter

> This is the recommended way to customize a struct.

Adjust examples in the documentation and add a test which uses
the block form.

-----------------

* [x] Added tests.


bbatsov:
Thanks! Probably it's worth extending the rationale of the cop as well.

Issue #5999, opened 2018-06-13
on_index, on_indexasgn, on_lambda callbacks seem to be missing?


alfonsoperez:
--------

## Expected behavior

Have callbacks available for https://github.com/whitequark/parser/commit/7d72eba571b8684ff452dd9c1885ea8c43698442 and `_type?`

## Actual behavior

Describe here what actually happened.

## Steps to reproduce the problem

```
module RuboCop
  module Cop
    module Deprecation
       class MutatedStrings < Cop
           MSG = "Don't mutate strings!"
           
           def on_indexasgn(node)
              binding.pry
           end
        end
       end
    end
  end
end
```

And a quick test with:
```
  it 'register an offense when indexed assignment is being used on a string' do
    expect_offense(<<~RUBY)
      foo[1] = 'c'
    RUBY
 end
```

It appears that the callback is not being called, nor a `node` as `indexasgn_type?` available

## RuboCop version

```
$ rubocop -V
0.57.2 (using Parser 2.5.1.0, running on ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
```

I am fairly new to ruby and definitely to rubocop, so excuse the issue / vs potential PR, I would love to contribute with some tips!.

bbatsov:
First time I hear about those callbacks. :-) I guess we'll need to update `Commissioner` to support them.

alfonsoperez:
> First time I hear about those callbacks. :-) I guess we'll need to update Commissioner to support them.

I will totally try to take a stab at this then!

Ruffeng:
Hi! I can try to have a look!

Drenmi:
> First time I hear about those callbacks.

They were only added a few months back. A lot of good work being done on parser. 🙂

Ruffeng:
After checking a while, I am not feeling confident to understand what's going on with AST, and therefore apply these 3 rules. Sorry

Drenmi:
Thanks anyway @Ruffeng. Feel free to chip in on other parts of the codebase. 🙂

paolobrasolin:
Hi @bbatsov and @Ruffeng , I'm trying to tackle this.

I think the reason for the missing callbacks is that the most recent AST format is opt-in: https://github.com/whitequark/parser#usage

This has two consequences:
[...]

PR #6877, opened 2019-04-03
NodePattern: Allow numerals in identifiers

marcandre:
This PR allows using digits in NodePatterns (e.g. `_arg0`, `custom_node_matcher_42?`).

It also paves the way to support the node type `procarg0`, which are not emitted by default (see #5999).

I didn't test for it, but it also forbids non sensical identifiers starting with `-` or consisting only of `-`.

Note that `Node` already states that [numerals are allowed](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/ast/node.rb#L15-L16)


bbatsov:
🙇

PR #6883, opened 2019-04-04
Add new `Layout/AutocorrectableLineLength` cop

maxh:
Similar to https://github.com/rubocop-hq/rubocop/pull/6824 and https://github.com/rubocop-hq/rubocop/pull/6826, this is a cop that we developed internally so that we could achieve more opinionated Ruby formatting using Rubocop.

It programmatically shortens certain lines by conservatively inserting line breaks into long method calls, hashes, and arrays. It's a bit complex, but I think the code comments and documentation do a reasonable job explaining both the implementation and intended usage.

I'm open to suggestions as how best to upstream this. I'm not sure, for example, if I should try to integrate it into `Metrics/LineLength`, or vice versa.

-----------------

* [x] Added tests.


bbatsov:
That's a really cool cop overall! 

> I'm open to suggestions as how best to upstream this. I'm not sure, for example, if I should try to integrate it into Metrics/LineLength, or vice versa.

I'll have to think about this. I'd also be curious to hear what the rest of @rubocop-hq/rubocop-core think on the subject of having a separate cop or merging this in the existing one.

jonas054:
It's an awesome new feature, @maxh!

I'm leaning towards incorporation into the existing `Layout/LineLength` cop as the best alternative. We'd get rid of double reporting of the same offense, and also of the trouble with two `Max` parameters or a shared `Max` parameter (both have their drawbacks). Do you think it's doable?

jonas054:
> This new logic cannot autocorrect all instances of Layout/LineLength -- only certain long lines that have breakable hashs, method calls, arrays, etc. Is it ok to have the LineLength cop be partially but not completely auto-correctable?

Yes, that's no problem. There are already several cops that autocorrect only some (or perhaps most) of the offenses they report.

> LineLength is in the namespace Metrics. The other cops in that namespace don't support autocorrect and tend to be computed summary stats of code. It might seem more natural to me for this cop to live in Layout or Style.

No matter where you put it, the `LineLength` cop will be a bit different from its department colleagues. It would fit better in `Layout` than in `Style` since it only makes whitespace changes, but it has a `Max` parameter that can be set by `--auto-gen-config` and that's why it's been put in the `Metrics` department. It's better to let it stay there IMO.

So, green light from me at least, to go ahead with the merger. I hope @bbatsov agrees.

maxh:
Please take a look at the new PR "Add initial autocorrection support to LineLength cop" https://github.com/rubocop-hq/rubocop/pull/6927. Thanks!

PR #6885, opened 2019-04-04
revert adding psych as runtime dependency

### LINKED ISSUE #6781: Upgrading 0.64.0 to 0.65.0 causes 'TypeError: superclass mismatch for class Mark'
Seems to be related to psych – please see https://github.com/ruby/psych/issues/386 for details.

andreaseger:
This reverts parts of #6766 by removing the psych runtime dependency again. This in term should fix #6781 and other issues regarding installation of psych from source on jruby.

From what I understand and is also written [here](https://github.com/rubocop-hq/rubocop/issues/6781#issuecomment-471061171) the psych version is tightly coupled to the ruby version. Especially in this case I do not really see a good enough reason to explicitly require a newer version of psych.

Note I had to make a small change in the YAMLDuplicationChecker

### DIFF
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -206,10 +206,16 @@ module RuboCop
         smart_path = PathUtil.smart_path(absolute_path)
         YAMLDuplicationChecker.check(yaml_code, absolute_path) do |key1, key2|
           value = key1.value
-          line1 = key1.start_line + 1
-          line2 = key2.start_line + 1
-          message = "#{smart_path}:#{line1}: " \
-                    "`#{value}` is concealed by line #{line2}"
+          # .start_line is only available since ruby 2.5 / psych 3.0
+          message = if key1.respond_to? :start_line
+                      line1 = key1.start_line + 1
+                      line2 = key2.start_line + 1
+                      "#{smart_path}:#{line1}: " \
+                      "`#{value}` is concealed by line #{line2}"
+                    else
+                      "#{smart_path}: " \
+                        "`#{value}` is concealed by duplicate"
+                    end
           warn Rainbow(message).yellow
         end
       end
@@ -228,7 +234,8 @@ module RuboCop
         if defined?(SafeYAML) && SafeYAML.respond_to?(:load)
           SafeYAML.load(yaml_code, filename,
                         whitelisted_tags: %w[!ruby/regexp])
-        else
+        # Ruby 2.6+
+        elsif Gem::Version.new(Psych::VERSION) >= Gem::Version.new('3.1.0')
           YAML.safe_load(
             yaml_code,
             permitted_classes: [Regexp, Symbol],
@@ -236,6 +243,8 @@ module RuboCop
             aliases: false,
             filename: filename
           )
+        else
+          YAML.safe_load(yaml_code, [Regexp, Symbol], [], false, filename)
         end
       end
     end
--- a/lib/rubocop/yaml_duplication_checker.rb
+++ b/lib/rubocop/yaml_duplication_checker.rb
@@ -5,7 +5,7 @@ module RuboCop
   module YAMLDuplicationChecker
     def self.check(yaml_string, filename, &on_duplicated)
       # Specify filename to display helpful message when it raises an error.
-      tree = YAML.parse(yaml_string, filename: filename)
+      tree = YAML.parse(yaml_string, filename)
       return unless tree
 
       traverse(tree, &on_duplicated)
--- a/spec/rubocop/yaml_duplication_checker_spec.rb
+++ b/spec/rubocop/yaml_duplication_checker_spec.rb
@@ -26,17 +26,32 @@ RSpec.describe RuboCop::YAMLDuplicationChecker do
 
     include_examples 'call block'
 
-    it 'calls block with keys' do
-      key1 = nil
-      key2 = nil
-      check(yaml) do |key_a, key_b|
-        key1 = key_a
-        key2 = key_b
+    context '< Ruby 2.5', if: RUBY_VERSION < '2.5' do
+      it 'calls block with keys' do
+        key1 = nil
+        key2 = nil
+        check(yaml) do |key_a, key_b|
+          key1 = key_a
+          key2 = key_b
+        end
+        expect(key1.value).to eq('Layout/Tab')
+        expect(key2.value).to eq('Layout/Tab')
       end
-      expect(key1.start_line).to eq(0)
-      expect(key2.start_line).to eq(3)
-      expect(key1.value).to eq('Layout/Tab')
-      expect(key2.value).to eq('Layout/Tab')
+    end
+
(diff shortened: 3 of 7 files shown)


andreaseger:
hmm ok now I see why this was probably done. To be able to who a warning when there are duplicate keys in the yaml. No actually to be able to show the line_numbers...

Not sure there is a way to get this information pre ruby 2.5. Changed it to simply not show the line numbers if `start_line` is not available on the psych nodes.

bbatsov:
I don't remember why this was needed to begin with. @rubocop-hq/rubocop-core Anyone knows?

bquorning:
cc/ @pocke re. https://github.com/rubocop-hq/rubocop/pull/6766

lib/rubocop/config_loader.rb
```diff
@@ -174,10 +174,15 @@ def check_duplication(yaml_code, absolute_path)
         smart_path = PathUtil.smart_path(absolute_path)
         YAMLDuplicationChecker.check(yaml_code, absolute_path) do |key1, key2|
           value = key1.value
-          line1 = key1.start_line + 1
-          line2 = key2.start_line + 1
-          message = "#{smart_path}:#{line1}: " \
-                    "`#{value}` is concealed by line #{line2}"
+          message = if key1.respond_to? :start_line
```
bbatsov: Please add some comment here about the need for the extra check, so we can delete it down the road.

lib/rubocop/config_loader.rb
```diff
@@ -196,14 +201,17 @@ def yaml_safe_load(yaml_code, filename)
         if defined?(SafeYAML) && SafeYAML.respond_to?(:load)
           SafeYAML.load(yaml_code, filename,
                         whitelisted_tags: %w[!ruby/regexp])
-        else
+        # Ruby 2.6+
```
bbatsov: Didn't you say this would be ok on 2.5+?

lib/rubocop/yaml_duplication_checker.rb
```diff
@@ -5,7 +5,7 @@ module RuboCop
   module YAMLDuplicationChecker
     def self.check(yaml_string, filename, &on_duplicated)
       # Specify filename to display helpful message when it raises an error.
-      tree = YAML.parse(yaml_string, filename: filename)
+      tree = YAML.parse(yaml_string, filename)
```
bbatsov: I think we need to make this conditional here as well.

spec/rubocop/yaml_duplication_checker_spec.rb
```diff
@@ -33,8 +33,10 @@ def check(yaml, &block)
         key1 = key_a
         key2 = key_b
       end
-      expect(key1.start_line).to eq(0)
-      expect(key2.start_line).to eq(3)
+      if key1.respond_to?(:start_line)
```
bbatsov: We also have mechanism to run different specs on different Ruby versions. I think it'd be better to use it.

lib/rubocop/config_loader.rb
andreaseger: I just reverted the exact comment that was removed in #6766 
I only checked the `start_time` bit and saw that it is available since 2.5

lib/rubocop/yaml_duplication_checker.rb
andreaseger: At least given that the specs work in all ruby versions used on CI can't we leave it like this? I doubt ruby-core would introduce a breaking change to the YAML module

spec/rubocop/yaml_duplication_checker_spec.rb
andreaseger: can you point me to a spec which uses this? then I can adapt the PR

lib/rubocop/config_loader.rb
andreaseger: will do 👍

lib/rubocop/yaml_duplication_checker.rb
bbatsov: OK.

spec/rubocop/yaml_duplication_checker_spec.rb
bbatsov: See `erb_new_arguments_spec.rb`.

lib/rubocop/config_loader.rb
andreaseger: ok I checked and the kwargs used here where introduced in psych 3.1 (which is used since ruby 2.6)

spec/rubocop/yaml_duplication_checker_spec.rb
andreaseger: hmm do I need to do anything more then set the rspec tags to make this work? it still seems to run all tests regardless which ruby version i'm using and therefore I get failing tests on old ruby version and errors on newer ones because of the changed output

jonas054: That's weird. I get the same results. I wonder if the RSpec functionality for specifying which Ruby version a certain scope is valid for has changed. This works:
```diff
diff --git a/spec/rubocop/yaml_duplication_checker_spec.rb b/spec/rubocop/yaml_duplication_checker_spec.rb
index 068270128..b6dc8548f 100644
--- a/spec/rubocop/yaml_duplication_checker_spec.rb
+++ b/spec/rubocop/yaml_duplication_checker_spec.rb
@@ -26,7 +26,7 @@ RSpec.describe RuboCop::YAMLDuplicationChecker do
 
     include_examples 'call block'
 
-    context '< Ruby 2.5', :ruby24 do
+    context '< Ruby 2.5', if: RUBY_VERSION >= '2.4' do
       it 'calls block with keys' do
         key1 = nil
         key2 = nil
@@ -39,7 +39,7 @@ RSpec.describe RuboCop::YAMLDuplicationChecker do
       end
     end
 
-    context '>= Ruby 2.5', :ruby25 do
+    context '>= Ruby 2.5', if: RUBY_VERSION >= '2.5' do
       it 'calls block with keys' do
         key1 = nil
         key2 = nil
```
so maybe we should change `:ruby2x` everywhere?

andreaseger: ok I'll change my tests then to this syntax

bbatsov: > That's weird. I get the same results. I wonder if the RSpec functionality for specifying which Ruby version a certain scope is valid for has changed. This works:

@Darhazer @bquorning ^^

andreaseger: fixed for the tests I added/touched. For the other cases it doesn't seem to be a problem

Darhazer: Setting the ruby version via `:ruby25` actually just sets the `ruby_version` variable, which is used by the ProcessedSource, e.g. what parser version to be used. It's useful for cops specs, but here you are not testing a cop.

PR #6892, opened 2019-04-06
Mark Rails/DynamicFindBy not safe

vfonic:
In the ruby stdlib, there's `Gem::Specification.find_by_name('...')` method which works without rails, but breaks if `Rails/DynamicFindBy` cop autocorrects it to `Gem::Specification.find_by(name: '...')`. This cop is therefore not safe.

From the rubydoc [find_by_name(name, *requirements)](https://ruby-doc.org/stdlib-2.6.1/libdoc/rubygems/rdoc/Gem/Specification.html#method-c-find_by_name):
> Find the best specification matching a name and requirements. Raises if the dependency doesn't resolve to a valid specification.

I'm using this method in my gem to find a path to another gem, from a file which doesn't necessarily load rails (think of a "fast" rspec test of a PORO object that doesn't need to load entire rails application; or in my case: a rake task that actually runs rubocop, and doesn't need to load rails to do so).

When I run `rubocop --safe-auto-correct`, the line:

```ruby
Gem::Specification.find_by_name('xxx_magic_gem').gem_dir
```

Gets autocorrected to:

```ruby
Gem::Specification.find_by(name: 'xxx_magic_gem').gem_dir
```

This breaks the code with the following error:
> NoMethodError: undefined method `find_by' for Gem::Specification:Class

I wanted to add to `Whitelist` the exact method name: `Gem::Specification.find_by_name`, but it seems like `Whitelist` config option only accepts method names, not the "full path" with module and class name.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2345,7 +2345,9 @@ Rails/DynamicFindBy:
   Description: 'Use `find_by` instead of dynamic `find_by_*`.'
   StyleGuide: 'https://github.com/rubocop-hq/rails-style-guide#find_by'
   Enabled: true
+  Safe: false
   VersionAdded: '0.44'
+  VersionChanged: '0.68'
   Whitelist:
     - find_by_sql
 


mikegee:
> I wanted to add to `Whitelist` the exact method name: `Gem::Specification.find_by_name`

I think adding that functionality is worth it. The volume of `ActiveRecord` usage that this cop can autocorrect outweighs the minority of `Gem::Specification` usage, in my opinion.

Perhaps we can add a `SafeReceivers` configuration that allows `find_by_*` methods to be called on `Gem::Specification`?

vfonic:
This cop can still be run with `--auto-correct` instead of `--safe-auto-correct`.

What's `SafeReceivers `? I can't find a single references it in the repo.

Considering the fact that users can still run the autocorrect, understanding the risks, I'd say this fix is good enough.

mikegee:
> What's SafeReceivers ? I can't find a single references it in the repo.

I suggested adding it. 

> I'd say this fix is good enough.

Maybe. I’ll defer to any third opinion that shows up here.

vfonic:
Hmmmm, I'm not sure we should introduce completely new config option that hasn't been used before. Perhaps we could just make `Whitelist` support regexes, just like `Blacklist` here supports regexes:
https://github.com/rubocop-hq/rubocop/blob/master/config/default.yml#L1884

After thinking more, no matter the solution we choose, this cop will never be safe, because we can never be sure on what type of object this method is called on.
For example:

```ruby
# clearly shouldn't be autocorrected
Gem::Specification.find_by_name('xxx_magic_gem')

# is spec an instance of ActiveRecord or Gem::Specification?
spec.find_by_name('xxx_magic_gem')
```

I believe, we should be conservative here and mark this cop as not safe.

bbatsov:
I think a big part of this problem can be addressed by scoping this cop only to models and controllers out of the box.

vfonic:
@bbatsov I like this idea! I understand that fixing `find_by_*` would probably never be 100% safe, as anyone can create a PORO that implements such a method.

If the usual practice is to still keep these cops safe, even though they are not, I'd say the best thing to do is to scope these to controllers and models, like you said.

On the other hand...I'd personally lean towards `--safe-auto-correct` always being safe and `--auto-correct` being unsafe (`--auto-correct` would autocorrect `find_by_*` while `--safe-auto-correct` wouldn't).

Thinking further, what about the `app/services` directory? I checked some of my apps and some of the classes in there know about Active Record (for example: I use one class to query some records and group them by day for `created_at` column).

If I'd have any `find_by_*` methods there, I'd prefer `--safe-auto-correct` to NOT correct it and `--auto-correct` to DO correct it. Scoping this cop to only controllers and models doesn't solve (actually creates) this issue.

bbatsov:
> If I'd have any find_by_* methods there, I'd prefer --safe-auto-correct to NOT correct it and --auto-correct to DO correct it. Scoping this cop to only controllers and models doesn't solve (actually creates) this issue.

No argument from me. Just keep in mind that almost all Rails cops are unsafe technically speaking, so if we really want to do things by the book we should audit all of them. The problem is always the same - their heuristics are naming based and therefore not reliable. Scoping the cops generally makes them safer, but there's always some potential for mistakes.

vfonic:
> Just keep in mind that almost all Rails cops are unsafe technically speaking, so if we really want to do things by the book we should audit all of them.

Yeah, I understand. 😕
What do you suggest? Should I just close this PR and leave the things as they are now?

> Scoping the cops generally makes them safer, but there's always some potential for mistakes.

How would I go about using autocorrect outside of scoped directories? Is the "scope" a config thing or cop implementation thing? Quick search didn't yield any results.

I'm going to close this PR now. My suggestion would be to mark all unsafe cops as unsafe (except some really edge-case ones, when user, for example, monkey patched something completely unexpectedly), and allow people to choose whether to use --auto-correct or --safe-auto-correct. I'd, for example, first run `--safe-auto-correct`, commit those changes, and then run `--auto-correct` and check each of the changes whether they are safe/properly corrected or not.

PR #6894, opened 2019-04-06
NodePattern: Fix complex ascend in sequence-head position

marcandre:
I noticed a subtle issue with `NodePattern` compiling ascend. If we write `^<something>`, the `<something>` should be about the parent's node, never about the parent's node's type.

So `seq_head` shouldn't be passed down.

This error wasn't too noticeable because matching on `node_type` works exactly the same in `seq_head` or not.

The added tests try to illustrate this. The first three were already passing, but the last one wasn't.


Drenmi:
Very subtle indeed. Sharp eye! 🙂

PR #6895, opened 2019-04-06
Load config from XDG config home

Mange:
Users now have an additional place they can have their own config file, if they prefer to not have dotfiles inside their home directory.

Based on initial work by @tejasbubane in #6682.

---

I chose `rubocop/config.yml` as the config file name since it seems to be the closes to how people use the specification currently.

In order to make this non-controversial, I made the XDG file last in the prioritization order instead of in front of `$HOME`. This could be changed if you prefer, or possibly changed later on (although that would then be a breaking change, while it would only be a minor bump doing it now).

Documentation and changelog has been updated. I made the documentation pretty verbose, but I can dial it back if you wish. Hopefully having the algorithm spelled out in a clear example makes it easier for people to grasp how it works.

`FileFinder` lost the `use_home` option as it now seems unrelated to traversing directories, and more to do with where the config files are located unless they are found by traversing the directories.

The new private methods in `ConfigLoader` is written in a way to make the prioritization fairly clear and easy to change and relate to in the future.

### DIFF
--- a/manual/configuration.md
+++ b/manual/configuration.md
@@ -4,10 +4,8 @@ The behavior of RuboCop can be controlled via the
 [.rubocop.yml](https://github.com/rubocop-hq/rubocop/blob/master/.rubocop.yml)
 configuration file. It makes it possible to enable/disable certain cops
 (checks) and to alter their behavior if they accept any parameters. The file
-can be placed either in your home directory or in some project directory.
-
-RuboCop will start looking for the configuration file in the directory
-where the inspected file is and continue its way up to the root directory.
+can be placed in your home directory, XDG config directory, or in some project
+directory.
 
 The file has the following format:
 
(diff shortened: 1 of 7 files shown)


bbatsov:
The build seems to be failing.

Mange:
I don't understand why because the faulures don't appear to be related to the code changes from what I can tell. I'm trying a rebuild since CircleCI had some issues recently.

Mange:
That fixed it. :smile:

bbatsov:
Thanks!

manual/configuration.md
```diff
@@ -26,6 +24,34 @@ Metrics/LineLength:
...
+RuboCop will start looking for the configuration file in the directory
+where the inspected file is and continue its way up to the root directory.
+
+If it cannot be found until reaching the project's root directory, then it will
+be searched for in the user's global config locations, which consists of a
+dotfile or a config file inside the [XDG Base Directory
+specification][xdg-basedir-spec].
+
+* `~/.rubocop.yml`
+* `$XDG_HOME/rubocop/config.yml` (expands to `~/.config/rubocop/config.yml` if
```
FranklinYu:
```diff
-* `$XDG_HOME/rubocop/config.yml` (expands to `~/.config/rubocop/config.yml` if
+* `$XDG_CONFIG_HOME/rubocop/config.yml` (expands to `~/.config/rubocop/config.yml` if
```

Mange: 🤦‍♂️

Should I do another PR?

Mange: Opened it here: #7045.

PR #6897, opened 2019-04-07
Add link to Development docs

vfonic:
I'm looking to add a cop so I went to `CONTRIBUTING.md` and couldn't find much information about adding a new cop. I found it in the docs site and thought it would be a good idea to include it in the `CONTRIBUTING.md` guide as well.


Drenmi:
🙇

PR #6906, opened 2019-04-08
[Fix #6903] Handle variables with `_` for `Naming/RescuedExceptionsVariableName`

### LINKED ISSUE #6903: `Naming/RescuedExceptionsVariableName` does not handle variables with underscore
After upgrading to Rubocop `v0.67.2`, if we use a variable name with a leading `_`, the offense message ask to rename it without the underscore which flag a new `Lint/UselessAssignment` offense.

--------

## Expected behavior

The cop should deal with underscored variable name and ask to rename it keeping the `_`.

## Actual behavior

Rubocop ask me to rename my variable `e` whereas it should be `_e`.

```ruby
class MyController < ApplicationController
  def create
    # some code here
  rescue MyException => _exception
    render_my_exception
  end
end
```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.67.2 (using Parser 2.6.2.1, running on ruby 2.6.1 x86_64-darwin18)
```

anthony-robin:
Currently, this cop do not play nicely with variables that are prefixed by an `_`.

The message should take into consideration if the variable is going to be
used or not when showing the error.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/naming/rescued_exceptions_variable_name.rb
+++ b/lib/rubocop/cop/naming/rescued_exceptions_variable_name.rb
@@ -24,6 +24,13 @@ module RuboCop
       #     # do something
       #   end
       #
+      #   # good
+      #   begin
+      #     # do something
+      #   rescue MyException => _e
+      #     # do something
+      #   end
+      #
       # @example PreferredName: exception
       #   # bad
       #   begin
@@ -36,6 +43,13 @@ module RuboCop
       #   begin
       #     # do something
       #   rescue MyException => exception
+      #     # do something
+      #   end
+      #
+      #   # good
+      #   begin
+      #     # do something
+      #   rescue MyException => _exception
       #     # do something
       #   end
       #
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/naming/rescued_exceptions_variable_name.rb
```diff
@@ -56,7 +56,11 @@ def on_resbody(node)
         private
 
         def preferred_name
-          @preferred_name ||= cop_config.fetch('PreferredName', 'e')
+          @preferred_name ||= begin
+            name = cop_config.fetch('PreferredName', 'e')
+            name = "_#{name}" if variable_name.to_s.start_with?('_')
```
anthony-robin: I'm not sure how I could use [this existing method][1] to avoid repeating code here 🤔 

[1]: https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/variable_force/variable.rb#L78-L80

bbatsov:
@anthony-robin The changes look good, but you'll need to rebase on top of `master`.

lib/rubocop/cop/naming/rescued_exceptions_variable_name.rb
bbatsov: The small amount of repetition is fine here.

anthony-robin:
- Rebased with master ✅ 
- Added more specs with auto-correct now that is has been merged ✅ 

Thanks !

bbatsov:
Thanks!

PR #6908, opened 2019-04-08
[Fix #6893] FIX bug with splat or methods exceptions on `RescuedExceptionsVariableName`

### LINKED ISSUE #6893: Naming/RescuedExceptionsVariableName invalid report
When I run rubocop __0.67.2__ with following code:
```ruby
handled = [StandardError]

begin
  raise StandardError
rescue *handled
  false
end
```

I get the following error:

```
test.rb:5:8: C: Naming/RescuedExceptionsVariableName: Use e instead of *handled.
rescue *handled
       ^^^^^^^^
```
--------

## Expected behavior

I should not get this offense.

## Actual behavior

I do get the offense.

- Create a file with the code above
- Run Rubocop  on it

## RuboCop version

```
$ [bundle exec] rubocop -V
0.67.2 (using Parser 2.6.2.1, running on ruby 2.6.2 x86_64-linux)
```

### LINKED ISSUE #6907: Naming/RescuedExceptionsVariableName does not handle method call instead of an error class name
When I run rubocop `0.67.2` with following code:

```ruby
begin
  obj = ActiveSupport::JSON.decode(some_string)
rescue ActiveSupport::JSON.parse_error
  Rails.logger.warn("Attempted to decode invalid JSON: #{some_string}")
end
```

I get the following error:

```
C: Naming/RescuedExceptionsVariableName: Use e instead of ActiveSupport::JSON.parse_error.
    rescue ActiveSupport::JSON.parse_error
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

--------

## Expected behavior

I should not get this offense.

## Actual behavior

I do get the offense.

* Create a file with the code above
* Run Rubocop on it

## RuboCop version

```
$ [bundle exec] rubocop -V
0.67.2 (using Parser 2.6.2.1, running on ruby 2.6.1 x86_64-darwin18)
```

### LINKED ISSUE #6912: RescuedExceptionsVariableName don't understand methods-as-rescue
In Ruby you can call a method in the rescue statement, and have it return an exception.

```ruby
module Errors
  def self.errors_matching_code(code)
    m = Module.new
    (class << m; self; end).instance_eval do
      define_method(:===, ->(e) { e.respond_to?(:code?) && e.code?(code) })
    end
    m
  end
end

begin
    ...
rescue Errors.errors_matching_code(1)
    ...
end
```

Rubocop 0.67.2 is currently unable to properly handle this case.

```
Naming/RescuedExceptionsVariableName: Use e instead of Errors.errors_matching_code(ERROR_CODE_SYSTEM_FAILEDUPDATE).
  rescue Errors.errors_matching_code(ERROR_CODE_SYSTEM_FAILEDUPDATE)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

anthony-robin:
This PR makes the `Naming/RescuedExceptionsVariableName` cop to play friendly
with exceptions listed using the splat operator or calling a method on a class.

```ruby
handled = [StandardError]

begin
  something
rescue *handled
  # do something
end
```

```ruby
begin
  something
rescue ActiveSupport::JSON.my_method
  # do something
end
```

Fix #6893
Fix #6907
Fix #6912

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/naming/rescued_exceptions_variable_name.rb
+++ b/lib/rubocop/cop/naming/rescued_exceptions_variable_name.rb
@@ -61,7 +61,8 @@ module RuboCop
           return unless exception_type || @exception_name
 
           @exception_name ||= exception_type.children.first
-          return if @exception_name.const_type? ||
+
+          return if !@exception_name.lvasgn_type? ||
                     variable_name == preferred_name
 
           add_offense(node, location: offense_range(node))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * [#5977](https://github.com/rubocop-hq/rubocop/issues/5977): Remove Performance cops. ([@koic][])
 * Add auto-correction to `Naming/RescuedExceptionsVariableName`. ([@anthony-robin][])
 * [#6903](https://github.com/rubocop-hq/rubocop/issues/6903): Handle variables prefixed with `_` in `Naming/RescuedExceptionsVariableName` cop. ([@anthony-robin][])
+* [#6893](https://github.com/rubocop-hq/rubocop/issues/6893): Fix bug with splat or method exceptions on `Naming/RescuedExceptionsVariableName` cop. ([@anthony-robin][])
 * [#6917](https://github.com/rubocop-hq/rubocop/issues/6917): Bump Bundler dependency to >= 1.15.0. ([@koic][])
 * Add `--auto-gen-only-exclude` to the command outputted in `rubocop_todo.yml` if the option is specified. ([@dvandersluis][])
 
--- a/spec/rubocop/cop/naming/rescued_exceptions_variable_name_spec.rb
+++ b/spec/rubocop/cop/naming/rescued_exceptions_variable_name_spec.rb
@@ -101,6 +101,88 @@ RSpec.describe RuboCop::Cop::Naming::RescuedExceptionsVariableName, :config do
           RUBY
         end
       end
+
+      context 'with method as `Exception`' do
+        it 'does not register an offense without variable name' do
+          expect_no_offenses(<<-RUBY.strip_indent)
+            begin
+              something
+            rescue ActiveSupport::JSON.my_method
+              # do something
+            end
+          RUBY
+        end
+
+        it 'does not register an offense with expected variable name' do
+          expect_no_offenses(<<-RUBY.strip_indent)
+            begin
+              something
+            rescue ActiveSupport::JSON.my_method => e
+              # do something
+            end
+          RUBY
+        end
+
+        it 'registers an offense with unexpected variable name' do
+          expect_offense(<<-RUBY.strip_indent)
+            begin
+              something
+            rescue ActiveSupport::JSON.my_method => exc
+                                                    ^^^ Use `e` instead of `exc`.
+              # do something
+            end
+          RUBY
+
+          expect_correction(<<-RUBY.strip_indent)
+            begin
+              something
+            rescue ActiveSupport::JSON.my_method => e
+              # do something
+            end
+          RUBY
+        end
+      end
+
+      context 'with splat operator as `Exception` list' do
+        it 'does not register an offense without variable name' do
+          expect_no_offenses(<<-RUBY.strip_indent)
+            begin
+              something
+            rescue *handled
+              # do something
+            end
+          RUBY
+        end
+
+        it 'does not register an offense with expected variable name' do
+          expect_no_offenses(<<-RUBY.strip_indent)
+            begin
+              something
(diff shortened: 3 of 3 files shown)


lib/rubocop/cop/naming/rescued_exceptions_variable_name.rb
```diff
@@ -70,6 +71,10 @@ def location
         def message(_node = nil)
           format(MSG, preferred: preferred_name, bad: variable_name)
         end
+
+        def splat_exceptions?(location)
+          location.source.to_s.start_with?('*')
```
bbatsov: It's best to check the type of the node instead of the source code, and filter out stuff based on it. Other things that should be filtered out are constants and method calls with params, that are producing some list of exceptions.

lib/rubocop/cop/naming/rescued_exceptions_variable_name.rb
```diff
@@ -76,6 +76,19 @@ def location
...
+            @exception_name.splat_type? ||
+            variable_name == preferred_name ||
+            exception_with_method?
+        end
+
+        def exception_with_method?
+          return unless @exception_name.children.first.respond_to?(:const_type?)
+
+          @exception_name.children.first.const_type?
+        end
```
anthony-robin: I'm not proud if this method... It feels hacky.

Am I right writing this ? Is there a way to clean it ?

bbatsov: Generally for situations like this it's much better to use just node pattern matching. Take a look at the docs or the code for other cops. With the right pattern you can potentially eliminate this method and `correct_code?` altogether.

anthony-robin:
Can somebody point me in the right direction to easily make the distinction between the constant name and the variable name with AST ?

When using the `send_type` it matches both of them so I have to add an extra condition that use the `&.` operation and this make the CI to fail.

Thanks for your help on this :)

bbatsov:
@anthony-robin What's the code you're trying to process?

anthony-robin:
I'm trying to enhance this code to detect exceptions defined from a method or if there is only the rescued variable name.

Both are matching the `send_type?` so I have to add an extra condition that use the `&.` which breaks the CI:

```ruby
@exception_name.send_type? && @exception_name.children.first&.const_type?
```

bbatsov:
> @exception_name.children

Should return `nil` for `e`. I can't imagine how this can be giving some result. In general what you need to check is that whatever was passed to `rescue` is not `lvasgn` node. That's everything else is out the scope of the cop.

lib/rubocop/cop/naming/rescued_exceptions_variable_name.rb
```diff
@@ -94,6 +94,14 @@ def location
         def message(_node = nil)
           format(MSG, preferred: preferred_name, bad: variable_name)
         end
+
+        def valid_code?
```
bbatsov: This whole method is redundant.

lib/rubocop/cop/naming/rescued_exceptions_variable_name.rb
```diff
@@ -61,8 +61,8 @@ def on_resbody(node)
           return unless exception_type || @exception_name
 
           @exception_name ||= exception_type.children.first
-          return if @exception_name.const_type? ||
-                    variable_name == preferred_name
+
+          return if valid_code?
```
bbatsov: return unless lvar assignment and you're done :-)

anthony-robin:
Thanks for the tips !

I've been able to refactor using `lvasgn_type?` but it now breaks specs with implicit rescue :(

bbatsov:
What exactly gets broken?

anthony-robin:
It breaks specs with implicit rescues 🤔 

```diff
  begin
    something
  rescue exc
-        ^^^ Use `e` instead of `exc`.
    # do something
  end
```

How can I avoid returning on this case as it matches the `lvasgn_type?` as well ?

Thanks !

Drenmi:
It's a question of whether we want to force people to change their variable that was assigned somewhere else, but for auto-correction, this should 100% be disabled for this case:

```ruby
begin
  ...
rescue foo
  ...
end
```

because it will break their code unless `foo` is also renamed in it's definition, which we can't reliably do (and probably shouldn't.)

I would probably suggest this cop only handle cases like:

```ruby
begin
  ...
rescue SomeException => foo
  ...
end
```

anthony-robin:
What should be the direction of this PR so ?

pocke:
Thank you for your pull request.
The problem has been fixed with #7122, so I'll close this pull request.
Thanks.

PR #6843, opened 2019-03-17
Unlimited terms after ellipsis

marcandre:
This makes it possible to have node patterns with as many terms as wanted after an `...` or an `$...`.

The initial commits are refactoring the code to make small improvements, factorize common code. Overall code is shorter.

Fixes #6840. Should help for #6841 too.

Issue #6841, opened 2019-03-16
Matching any child with NodePattern


marcandre:
From @Drenmi :
>  there's a long standing request to add an "any child" matcher. This would be especially helpful when matching on hashes. Consider:
> ```
> ruby-parse -e '{ foo: 1, bar: 2 }'
> 
> (hash
>   (pair
>     (sym :foo)
>     (int 1))
>   (pair
>     (sym :bar)
>     (int 2)))
> ```
> We may be interested in hashes that have a `foo: 1` pair, but the pairs can appear in any order, so we can't reliably match that in a pattern. 🙂
> 
> What is needed is a suggested extension to the syntax, and the actual implementation.

Suggested syntax extension would be `<...>` as it seems the easiest remaining "brackety" symbol left.

# Simple:
```ruby
(hash <(pair (sym :foo) (int 1)) (pair (sym :bar) (int 2))>)
```
This would match `{bar: 2, foo: 1}` and `{foo: 1, bar: 2}` but nothing else. To match hashes containing these two value-key pairs, use `...`:

# Ellipsis:
```ruby
(hash <(pair (sym :foo) (int 1)) ...>)
```

This would match any literal hash with a `foo: 1` key-value pair. Ellipsis must be at end of `<>` sequence. 

*Note*: Patterns are attempted to be matched in the order given. `<(pair _ (int 1)) (pair (sym :foo) _)>` would match `{bar: 1, foo: 1}` but not `{foo: 1, bar: 1}`. Captures are always returned in the order of the pattern.

Comments?

marcandre:
Description edited to match #6965

PR #6909, opened 2019-04-08
NodePattern: Optimize by using placeholders for post processing

marcandre:
The improved handling of `...` had to use some post-processing of the `cur_node` (#6843)
The easiest way to provide other "multiple match" patterns (e.g. #6841) would involve similar post-processing. This PR paves the way by using post processing of the current node pattern everywhere.

This as other multiple benefits:
- deals automatically and better with generated temporary variables
- much less argument passing to the various compiling methods
- factorizes the `#{'.type' if seq_head}` pattern that was everywhere.
- makes debugging easier with simpler & nicer generated code

I had hoped to be able to minimize the number of node guards, but unions make it tricky.

For the record, here's an example of the code generated before & after.

```ruby
### Pattern:
($_ {:a :b} nil?)

### Before:
(temp1 = node0; temp1 = temp1;temp1.is_a?(RuboCop::AST::Node) &&
(capture1 = temp1.type; true) &&
(temp2 = temp1.children[0]; temp2 = temp2;(temp2 == :a) || (temp2 == :b)) &&
(temp1.children[1].nil?) &&
(temp1.children.size == 2))

### After:
node0.is_a?(RuboCop::AST::Node) &&
(capture1 = node0.type; true) &&
((temp1 = node0.children[0]) == :a || temp1 == :b) &&
node0.children[1].nil? &&
node0.children.size == 2
```

### DIFF
--- a/lib/rubocop/node_pattern.rb
+++ b/lib/rubocop/node_pattern.rb
@@ -125,6 +125,12 @@ module RuboCop
 
       SEQ_HEAD_INDEX = -1
 
+      # Placeholders while compiling, see with_..._context methods
+      CUR_PLACEHOLDER = '@@@cur'.freeze
+      CUR_NODE = "#{CUR_PLACEHOLDER} node@@@".freeze
+      CUR_ELEMENT = "#{CUR_PLACEHOLDER} element@@@".freeze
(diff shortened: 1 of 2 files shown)


bbatsov:
Fantastic work!

PR #6919, opened 2019-04-11
[Fix #6917] Bump Bundler dependency to >= 1.15.0

### LINKED ISSUE #6917: Error while parsing Gemfile
An error occurred executing RuboCop v0.67.2.
Outputs error as following:

```
/usr/local/bundle/gems/rubocop-0.67.2/exe/rubocop:16:in `[]': no implicit conversion of String into Integer (TypeError)
	from /usr/local/bundle/gems/rubocop-0.67.2/exe/rubocop:16:in `<top (required)>'
	from /usr/local/bundle/bin/rubocop:29:in `load'
	from /usr/local/bundle/bin/rubocop:29:in `<main>'
```

And the related code is following:

https://github.com/rubocop-hq/rubocop/blob/1ec76b9251edb0e3edd159a72ec7fde58e1a22f8/exe/rubocop#L12-L16

I'm using Bundler in version v1.14 and the implement of `LockFileParser#dependencies` looks different between v1.14 and v1.15. 

https://github.com/bundler/bundler/commit/9febe9a5d19d1dac8a7bd032fc2b3ab029288646#diff-3c625d3cd7d7604b3e2e3c5487a5ede6L64

So I thought it should be changed requirement version for bundler in RuboCop gemspec.

https://github.com/rubocop-hq/rubocop/blob/1ec76b9251edb0e3edd159a72ec7fde58e1a22f8/rubocop.gemspec#L44

koic:
Fixes #6917.

This PR bumps Bundler dependency to >= 1.15.0.

Bundler 2 has already been released.
This PR drops old Bundler versions (1.14 or lower) to prevent the following error.

## Bundler 1.14

An unexpected error occurs.

```console
% bundle _1.14.6_ exec rubocop
bundler: failed to load command:
rubocop (/Users/koic/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/bin/rubocop)
TypeError: no implicit conversion of String into Integer
  /Users/koic/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/rubocop-0.67.1/exe/rubocop:16:in
  `<top (required)>'
  /Users/koic/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/bin/rubocop:23:in
  `load'
  /Users/koic/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/bin/rubocop:23:in
  `<top (required)>'
```

## Bundler 1.15

Errors does not occur as expected.

```console
% bundle _1.15.0_ exec rubocop
[Warn] Performance Cops will be removed from RuboCop 0.68. Use
rubocop-performance gem instead.
       https://github.com/rubocop-hq/rubocop/tree/master/manual/migrate_performance_cops.md

(snip)
```

-----------------

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [#5977](https://github.com/rubocop-hq/rubocop/issues/5977): Remove Performance cops. ([@koic][])
 * Add auto-correction to `Naming/RescuedExceptionsVariableName`. ([@anthony-robin][])
 * [#6903](https://github.com/rubocop-hq/rubocop/issues/6903): Handle variables prefixed with `_` in `Naming/RescuedExceptionsVariableName` cop. ([@anthony-robin][])
+* [#6917](https://github.com/rubocop-hq/rubocop/issues/6917): Bump Bundler dependency to >= 1.15.0. ([@koic][])
 
 ## 0.67.2 (2019-04-05)
 
--- a/rubocop.gemspec
+++ b/rubocop.gemspec
@@ -41,7 +41,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('ruby-progressbar', '~> 1.7')
   s.add_runtime_dependency('unicode-display_width', '>= 1.4.0', '< 1.6')
 
-  s.add_development_dependency('bundler', '>= 1.3.0', '< 3.0')
+  s.add_development_dependency('bundler', '>= 1.15.0', '< 3.0')
   s.add_development_dependency('rack', '>= 2.0')
 
   s.post_install_message = File.read('manual/migrate_performance_cops.md')


bbatsov:
Thanks!

deivid-rodriguez:
Hi @koic! I don't think this PR fixes the issue.

`bundler` is not a runtime dependency of `rubocop`, so changing the `developement_dependency` won't have any effect for end users.

What you can do is activate the version your code needs inside the rubocop executable (`gem "bundler", ">= 1.15.0"` or whatever), and rescue the error and give a proper message if that fails. And in the case where `bundler` is already loaded by the time the executable is run, check for an appropriate `Bundler::VERSION`.

Also, shouldn't the warning be activated _only_ if the user is actually using performance cops?

koic:
@deivid-rodriguez Right! Thanks for your feedback.

> Also, shouldn't the warning be activated only if the user is actually using performance cops?

I opened a PR #6939 that removes a warning when running `rubocop` command. In the future, it wll be better to replace it with a mechanism that displays a warning only when the user need Performance. Thank you, again!

deivid-rodriguez:
No problem, thanks to you for your good work!

alexandergraul:
Rubocop failed for older bundlers because "Bundle dependencies" were change to be a Hash instead of an Array in: bundler/bundler@9febe9a

Fix is here: https://github.com/alexandergraul/rubocop/commit/bec5dbdd203e10b3a0ec51e561c03d99247b98e0. It may help when bringing this back only for the case  the user need Performance without having to require a newer Bundler.

PR #6921, opened 2019-04-11
[Fix #5975] Add beginning_of_first_line mode to ClosingParenthesisIndentation

### LINKED ISSUE #5975: Unexpected Layout/ClosingParenthesisIndentation behavior with non-standard indententation
I'm not sure how this doesn't fail [in the specs](https://github.com/rubocop-hq/rubocop/blob/68609e6bcd53e82043c404d6c1e4fe1e82457b63/spec/rubocop/cop/layout/closing_parenthesis_indentation_spec.rb#L31) but with 0.57.1, the basic "good" case for `Layout/ClosingParenthesisIndentation` raises an error.

This looks like it's related to changed for Issue #4136 (I see no related PR). Likely something in 68609e6 which appears to be the (large) commit related to that effort.

--------

## Expected behavior

Given the following file:

```
# frozen_string_literal: true

some_method(
    a,
    b
)
```

I would expect rubocop to accept that code. It's _exactly_ what is [described in the documentation example](https://github.com/rubocop-hq/rubocop/commit/68609e6bcd53e82043c404d6c1e4fe1e82457b63#diff-84278b38a8a718e15ad796e211b6c0c2R39).

## Actual behavior

Rubocop raises an error:

```
$ rubocop close-paren.rb 
Inspecting 1 file
C

Offenses:

close-paren.rb:6:1: C: Layout/ClosingParenthesisIndentation: Indent ) to column 2 (not 0)
)
^

```

See above.

## RuboCop version

```
$ rubocop -v
0.57.1
```

maxh:
As a follow-up to https://github.com/rubocop-hq/rubocop/pull/6826, this PR adds `beginning_of_first_line` mode to support the behavior desired by commenters here:

https://github.com/rubocop-hq/rubocop/issues/5975 "Layout/ClosingParenthesisIndentation is incorrect in 0.57.1 even on basic case"

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#5975](https://github.com/rubocop-hq/rubocop/issues/5975): [Fix #5975] Add `beginning_of_first_line` mode to `Layout/ClosingParenthesisIndentation`. ([@maxh][])
+
 ### Bug fixes
 
 * [#6914](https://github.com/rubocop-hq/rubocop/issues/6914): [Fix #6914] Fix an error for `Rails/RedundantAllowNil` when with interpolations. ([@Blue-Pix][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -394,6 +394,10 @@ Layout/ClosingParenthesisIndentation:
   Description: 'Checks the indentation of hanging closing parentheses.'
   Enabled: true
   VersionAdded: '0.49'
+  EnforcedStyle: consistent
+  SupportedStyles:
+    - consistent
+    - beginning_of_first_line
 
 Layout/CommentIndentation:
   Description: 'Indentation of comments.'
--- a/lib/rubocop/cop/layout/closing_parenthesis_indentation.rb
+++ b/lib/rubocop/cop/layout/closing_parenthesis_indentation.rb
@@ -7,10 +7,18 @@ module RuboCop
       # method calls, method definitions, and grouped expressions. A hanging
       # closing parenthesis means `)` preceded by a line break.
       #
-      # @example
+      # The default style is called 'consistent'. An alternative
+      # style is 'beginning_of_first_line'. The examples below illustrates
+      # the expected behavior.
+      #
+      # Note that the behavior in `consistent` mode is dependent on the
+      # indentation of the parameters in the method call. You can use other
+      # cops such as `FirstParameterIndentation` to adjust that.
+      #
+      # @example EnforcedStyle: consistent (default)
       #
       #   # bad
-      #   some_method(
+      #   foo = some_method(
       #     a,
       #     b
       #     )
@@ -19,7 +27,7 @@ module RuboCop
       #     a, b
       #     )
       #
-      #   some_method(a, b, c
+      #   foo = some_method(a, b, c
       #     )
       #
       #   some_method(a,
@@ -36,7 +44,7 @@ module RuboCop
       #
       #   # good: when first param is on a new line, right paren is *always*
       #   #       outdented by IndentationWidth
-      #   some_method(
+      #   foo = some_method(
       #     a,
       #     b
       #   )
@@ -67,12 +75,70 @@ module RuboCop
       #     y: 2
       #   )
       #
+      # @example EnforcedStyle: beginning_of_first_line
+      #
+      #   # bad
+      #   foo = some_method(
+      #     a,
+      #     b
+      #     )
+      #
+      #   some_method(
+      #     a, b
+      #     )
+      #
+      #   some_method(a, b, c
+      #     )
+      #
+      #   some_method(a,
+      #               b,
+      #               c
+      #     )
+      #
+      #   some_method(a,
+      #     x: 1,
+      #     y: 2
+      #     )
+      #
+      #   # good: closing paren aligns with the beginning of the first line
+      #   #       containing the expression (even when the first
+      #   #       first parameter is on the same line as the opening
+      #   #       parenthesis, or when the parameters are indented
+      #   #       in a non-standard way)
+      #   foo = some_method(
+      #     a,
+      #     b
+      #   )
+      #
+      #   some_method(
+      #     a, b
+      #   )
+      #
+      #   some_method(a, b, c
+      #   )
+      #
+      #   some_method(a,
+      #               b,
+      #               c
+      #   )
+      #
+      #   some_method(a,
+      #     x: 1,
+      #     y: 2
+      #   )
+      #
+      #   some_method(a,
+      #       x: 1,
+      #       y: 2
+      #   )
       #
       class ClosingParenthesisIndentation < Cop
         include Alignment
 
         MSG_INDENT = 'Indent `)` to column %<expected>d (not %<actual>d)'
                      .freeze
+        MSG_BEGINNING_OF_FIRST_LINE = 'Indent `)` to align with the ' \
+          'beginning of the first line containing the expression.'.freeze
         MSG_ALIGN = 'Align `)` with `(`.'.freeze
 
         def on_send(node)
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 * [#6914](https://github.com/rubocop-hq/rubocop/issues/6914): [Fix #6914] Fix an error for `Rails/RedundantAllowNil` when with interpolations. ([@Blue-Pix][])
+* [#5975](https://github.com/rubocop-hq/rubocop/issues/5975): [Fix #5975] Add `beginning_of_first_line` mode to `Layout/ClosingParenthesisIndentation`. ([@maxh][])
```
bbatsov: That's actually a new feature.

bbatsov:
I guess both styles also need examples of what happens when you have more than just a method call. E.g.

```
result = foo(bar,
                     baz)
```

Frankly, I'm not even quite sure what's the problem the new config solves at this point, but I assume it has something to do with that.

maxh:
>Frankly, I'm not even quite sure what's the problem the new config solves at this point, but I assume it has something to do with that.

The issue I was trying to solve was unexpected indentation in method definitions. I would expect this:

```
def foo(
     a,
     b
   )
...
```

To autocorrect to:
```
def foo(
  a,
  b
)
...
```

But the cop found no violations. This PR fixed the closing parenthesis.  However, it turns out this change is not needed anymore after https://github.com/rubocop-hq/rubocop/pull/6929 (Add support for method definition parameters to `Layout/FirstParameterIndentation`). That's because the default config for `Layout/ClosingParenthesisIndentation` will appropriately indent the closing paren relative to the parameters. So fixing the parameter indent fixes the closing parenthesis, if that makes sense.

So this PR is not actually needed to accomplish my goals. But the bug is open so we should either close it or fix it. My understanding of the commenter's problem is that in cases of non-standard parameter indentation they always want the closing parenthesis to be at the beginning of the first line. I updated the docs to explain/show the behavior, but maybe the OP @jeremywadsack can try this out locally and see if it addresses his concern. Otherwise, we can close this.

As an aside, discovering the interactions all these layout cops requires a lot of research. Having a recommended "strong pretty print" default configuration could be useful for people considering whether RuboCop is sufficient to address their formatting needs.

jeremywadsack:
@maxh I tested this with a very small code base and it seems to resolve the issue. It still feels weird that I have to configure this cop to do what seems like the correct thing because I changed the default parameters of another cop, but I'm sure there are configurations that have to be balanced.

zoras:
Eagerly waiting for this to get merged so that @maxh can release the `.rubocop.yml` file for all the cops stated at https://flexport.engineering/approximating-prettier-for-ruby-with-rubocop-8b863bd64dc6

bbatsov:
I was under the impression we don't need after #6929. @maxh Is this the case or not?

maxh:
Apologies for the delay! For some reason I haven't been getting emails for comments on this thread. I just opened this page intending to close the PR because we don't need it anymore and then noticed the comments. My understanding is as follows.

## Alternative fix

@jeremywadsack I believe you can accomplish the behavior you want with these new cops in the latest release:

```yml
Layout/IndentFirstArgument:
  Enabled: true
  EnforcedStyle: consistent

Layout/IndentFirstParameter:
  Enabled: true
  EnforcedStyle: consistent

Layout/AlignParameters:
  Enabled: true

Layout/AlignArguments:
  Enabled: true

Layout/ClosingParenthesisIndentation:
  Enabled: true
``` 

Can you confirm? If not, please provide an example of an unexpected result.

## Diagnosis of example provided in bug report

https://github.com/rubocop-hq/rubocop/issues/5975

```rb
some_method(
    a,
    b
)
```

We use indentation width of 2. So for us, the issue here would be that `a` and `b` are intended too far. `IndentFirstArgument` and `AlignArguments` should fix that, and then `ClosingParenthesisIndentation` will behave as expected.

I'm not sure what would be needed to set the indentation width to 4, but it seems like there is some relevant discussion on the bug https://github.com/rubocop-hq/rubocop/issues/5975. Perhaps there is a different bug that should be fixed in this cop related to indentation width. This would be a different issue than the one I was trying to result, though the behavior change here happened to fixed both issues.

## Explanation of the alternative fix

The reason we needed this PR internally was to fix the alignment of the closing parenthesis in multi-line method _definitions_, as in the example shown above. For method _calls_, the existing code worked fine for us. Why?

When I dug in, it turns out the discrepancy was because this cop's formatting of the closing parenthesis `)` is dependent on the indentation of the elements in the method call or definition, which are in turn dependent on the indentation of the first argument or parameter.

For method calls, `Layout/FirstParameterIndentation` ensured correct indentation of the first _argument_, `Layout/AlignParameter` aligned the remaining arguments with the first, and then `Layout/ClosingParenthesisIndentation` would work as expected.

For method definitions, however, `Layout/FirstParameterIndentation` did not fix the indentation of the first parameter, and the chain of dependencies broke.

This suite of changes was sufficient for our use cases:
#6982 "Add IndentFirstArgument and IndentFirstParameter"
#6981 "Extract AlignArguments out from AlignParameters"
#6983 "Rename IndentArray and IndentHash to IndentFirstArrayElement and IndentFirstHashElement"

## Suggested config for prettier-like formatting

@zoras I've now upstreamed all the cops we need for pretty printing. The layout part of our config is here: https://gist.github.com/maxh/2fc21e458cde793532d15411f061bd92

zoras:
Thanks a lot @maxh 👍 Keep up the great work! 😸

bbatsov:
So, @jeremywadsack, what will it be? ;-)

jeremywadsack:
Sorry, but this still isn't working for me with the latest `rubocop`:

```
$ rubocop --version
0.71.0
```

```
$ cat .rubocop.yml 
Layout/IndentFirstArgument:
  Enabled: true
  EnforcedStyle: consistent
  IndentationWidth: 4

Layout/IndentFirstParameter:
  Enabled: true
  EnforcedStyle: consistent

Layout/AlignParameters:
  Enabled: true

Layout/AlignArguments:
  Enabled: true

Layout/ClosingParenthesisIndentation:
  Enabled: true
```

```
$ cat 5975.rb 
# frozen_string_literal: true

some_method(
    a,
    b
)
```

```
$ rubocop 5975.rb 
Inspecting 1 file
C

Offenses:

5975.rb:6:1: C: Layout/ClosingParenthesisIndentation: Indent ) to column 2 (not 0)
)
^

```

maxh:
Closing this out per my comment here: https://github.com/rubocop-hq/rubocop/issues/5975#issuecomment-503722203

@jeremywadsack, per the linked comment maybe try setting `IndentationWidth: 4` under `IndentFirstParameter`:

```
Layout/IndentFirstParameter
  IndentationWidth: 4
```

jeremywadsack:
@maxh, sorry but that still doesn't fix it:

```
$ rubocop --version
0.71.0
```

```
$ cat 5975.rb 
# frozen_string_literal: true

some_method(
    a,
    b
)
```

```
$ cat .rubocop.yml 
Layout/IndentFirstArgument:
  Enabled: true
  EnforcedStyle: consistent
  IndentationWidth: 4

Layout/IndentFirstParameter:
  Enabled: true
  EnforcedStyle: consistent
  IndentationWidth: 4

Layout/AlignParameters:
  Enabled: true

Layout/AlignArguments:
  Enabled: true

Layout/ClosingParenthesisIndentation:
  Enabled: true
```

```
$ rubocop 5975.rb 
Inspecting 1 file
C

Offenses:

5975.rb:6:1: C: Layout/ClosingParenthesisIndentation: Indent ) to column 2 (not 0)
)
^

```

---
I guess I'm confused about the difference between these two cops:

`Layout/IndentFirstArgument`
> This cop checks the indentation of the first argument in a method call or definition.

`Layout/IndentFirstParameter`
> This cop checks the indentation of the first parameter in a method call. Parameters after the first one are checked by Layout/AlignParameters, not by this cop.

Beyond the inconsistency of "argument" vs. "parameter" why do we need two cops?

jeremywadsack:
I'm going to move that comment over to the original issue, as I think it belongs there.

PR #6927, opened 2019-04-13
Add initial autocorrection support to LineLength cop

maxh:
Per feedback on https://github.com/rubocop-hq/rubocop/pull/6883 "Add new `Layout/AutocorrectableLineLength` cop", this PR adds the autocorrect functionality from that cop into the existing `Metrics/LineLength` cop.

Notes:
- Added a `investigate_post_walk` hook to the commissioner. See comments in that file for motivation.
- Fixed typo `highligh` => `highlight` in LineLength.
- Switched from reserved word `index` to `line_index` in LineLength.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/commissioner.rb
+++ b/lib/rubocop/cop/commissioner.rb
@@ -44,6 +44,7 @@ module RuboCop
         invoke_custom_processing(@cops, processed_source)
         invoke_custom_processing(@forces, processed_source)
         walk(processed_source.ast) unless processed_source.blank?
+        invoke_custom_post_walk_processing(@cops, processed_source)
         @cops.flat_map(&:offenses)
       end
 
@@ -89,12 +90,35 @@ module RuboCop
       # If they define the #investigate method, all input parameters passed
       # to the commissioner will be passed to the cop too in order to do
       # its own processing.
+      #
+      # These custom processors are invoked before the AST traversal,
+      # so they can build initial state that is later used by callbacks
+      # during the AST traversal.
       def invoke_custom_processing(cops_or_forces, processed_source)
         cops_or_forces.each do |cop|
           next unless cop.respond_to?(:investigate)
 
           with_cop_error_handling(cop) do
             cop.investigate(processed_source)
+          end
+        end
+      end
+
+      # There are cops that require their own custom processing **after**
+      # the AST traversal. By performing the walk before invoking these
+      # custom processors, we allow these cops to build their own
+      # state during the primary AST traversal instead of performing their
+      # own AST traversals. Minimizing the number of walks is more efficient.
+      #
+      # If they define the #investigate_post_walk method, all input parameters
+      # passed to the commissioner will be passed to the cop too in order to do
+      # its own processing.
+      def invoke_custom_post_walk_processing(cops, processed_source)
+        cops.each do |cop|
+          next unless cop.respond_to?(:investigate_post_walk)
+
+          with_cop_error_handling(cop) do
+            cop.investigate_post_walk(processed_source)
           end
         end
       end
--- a/lib/rubocop/cop/metrics/line_length.rb
+++ b/lib/rubocop/cop/metrics/line_length.rb
@@ -2,6 +2,7 @@
 
 require 'uri'
 
+# rubocop:disable Metrics/ClassLength
 module RuboCop
   module Cop
     module Metrics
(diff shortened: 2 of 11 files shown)


maxh:
image
FYI the test failure appears flaky/unrelated, but I can't way to retry.

lib/rubocop/cop/metrics/line_length.rb
```diff
@@ -9,22 +9,86 @@ module Metrics
...
+      # This cop has some autocorrection capabilities.
+      # It can programmatically shorten certain long lines by
+      # inserting line breaks into expressions that can be safely
+      # split across lines. These include arrays, hashes, and
+      # method calls with argument lists.
+      #
+      # If autocorrection is enabled, the following Layout cops
+      # are recommended to further format the broken lines.
+      #
+      # ClosingParenthesisIndentation
```
bbatsov: I think you should mark those as unordered list for better output.

lib/rubocop/cop/commissioner.rb
```diff
@@ -99,6 +104,25 @@ def invoke_custom_processing(cops_or_forces, processed_source)
...
+      # There are cops that require their own custom processing **after**
+      # the AST traversal. By performing the walk before invoking these
+      # custom processors, we allow these cops to build their own
+      # state during the primary AST traversal instead of performing their
+      # own AST traversals. Minimizing the number of walks is more efficient.
+      #
+      # If they define the #investigate_after_walk method, all input parameters
+      # passed to the commissioner will be passed to the cop too in order to do
+      # its own processing.
+      def invoke_custom_processing_after_walk(cops, processed_source)
```
bbatsov: Maybe `postprocessing` is a better term here?

bbatsov:
The changes look good to me. Probably that's one auto-correct worth enabling by default.

lib/rubocop/cop/metrics/line_length.rb
maxh: Done. (I think!)

lib/rubocop/cop/commissioner.rb
maxh: After trying a few variations, I settled on `_post_walk_processing`. The name `post_processing` suggests that what's meaningful is that these hooks are invoked after the initial `processing` round, ie after `invoke_custom_processing`. But actually, what's important is that these hooks invoked after _the walk_.

Feel free to reopen if you think otherwise.

maxh:
>Probably that's one auto-correct worth enabling by default.

Done, although I'm tempted to disable it for a release or two to let early adopters try it in the wild before enabling by default.

I'm curious to see, for example, how folks respond to the autocorrection here if they do not have the other recommended formatting cops enabled. Without the other cops, especially `MultilineHashKeyLineBreaks` and friends, the autocorrection for this cop alone could be a bit underwhelming. People might prefer to manually break lines in other ways, for example:

```rb
# original
foo(bar0, bar0, bar0, bar0, bar0)

# autocorrected with this cop alone
foo(bar0,
bar0, bar0, bar0, bar0)

# autocorrected with recommended cops
foo(
  bar0,
  bar0,
  bar0,
  bar0,
  bar0,
)

# hypothetical preferred manual correction in some cases
foo(bar0, bar0, bar0,
       bar0, bar0)
```

That said, I'm pretty confident the code works. We've been running on 1.5m lines of code with 100+ engineers for several months. Up to you'all. Excited!

bbatsov:
> I'm curious to see, for example, how folks respond to the autocorrection here if they do not have the other recommended formatting cops enabled. Without the other cops, especially MultilineHashKeyLineBreaks and friends, the autocorrection for this cop alone could be a bit underwhelming. People might prefer to manually break lines in other ways, for example:

That's a good point. I think it'd be nice if the auto-correct did some simply line break when possible for the common collection literals when the pretty-printing cops are not enabled (closest entry to the length limit or something along those lines). That's not necessary in scope here, but something that we can do at some point. 

I also noticed in your blog post that expressions separated by `;` don't seem to be autocorrected on new lines, which should be simple to address.

jonas054:
I tried out the changes by setting `Metrics/LineLength:Max` to `79` and running with `--auto-correct` on RuboCop's own source code. One of the corrections I got was

```diff
diff --git a/spec/rubocop/cop/offense_spec.rb b/spec/rubocop/cop/offense_spec.rb
index 9969f7046..bf455f999 100644
--- a/spec/rubocop/cop/offense_spec.rb
+++ b/spec/rubocop/cop/offense_spec.rb
@@ -2,7 +2,8 @@
 
 RSpec.describe RuboCop::Cop::Offense do
   subject(:offense) do
-    described_class.new(:convention, location, 'message', 'CopName', :corrected)
+    described_class.new(:convention,
+                        location, 'message', 'CopName', :corrected)
   end
 
   let(:location) do
```
I think I would prefer the line break to come as late as possible.

maxh:
>I also noticed in your blog post that expressions separated by ; don't seem to be autocorrected on new lines, which should be simple to address.

+1. Added the concept of `breakable_range` to support this. The cop now breaks semi-colon separated parts before hashes, arrays, etc. Haven't tested yet but this should improve the prettier example.

code: https://github.com/rubocop-hq/rubocop/pull/6927/files#diff-fd504989ec8ee8d7f2ce117a1f72b127R150
illustrative spec: https://github.com/rubocop-hq/rubocop/pull/6927/files#diff-b206bfd15578ae3e403e777e7ebeceb1R831

>I think I would prefer the line break to come as late as possible.

Added `extract_first_element_over_column_limit` to support this.
code: https://github.com/rubocop-hq/rubocop/pull/6927/files#diff-eea62a3c55bc8afdba0d7860896f7d6aR68
illustrative spec: https://github.com/rubocop-hq/rubocop/pull/6927/files#diff-b206bfd15578ae3e403e777e7ebeceb1R537

One improvement from here would be to support breaking e.g. hash twice onto multiple lines like so:

``` rb
#                                                              |
#                                                              v
[aaaaaaaaaaaaaa, bbbbbbbbbbbbb, ccccccccccccccc, ddddddddddddd, eeeeeeeeee]

# first break (supported)
[aaaaaaaaaaaaaa, bbbbbbbbbbbbb,
ccccccccccccccc, ddddddddddddd, eeeeeeeeee]

# second break (not supported)
[aaaaaaaaaaaaaa, bbbbbbbbbbbbb,
ccccccccccccccc, ddddddddddddd,
eeeeeeeeee]
````

However, I'd rather not do that right now. The recommended alignment cops obviate the need for the second break, and it's not clear to me exactly what the preferred behavior should be without the recommended cops enabled.

I haven't extensively tested these two behavior tweaks on a real codebase, but added several new tests for them.

**Comment by bbatsov** [comment]
I'm fine with a gradual rollout of auto-correct improvements. Probably at some point some looping would be required until all lines are properly shortened. 

I'd be fine with merging this in its current form. There's also one small philosophical question - is it still a metrics cop or a layout cop? Not a big deal, but the thought did cross my mind.

**Comment by jonas054** [comment]
> Added extract_first_element_over_column_limit to support this.

Thanks a lot! I love that change. :slightly_smiling_face: 

> There's also one small philosophical question - is it still a metrics cop or a layout cop?

It has one foot in each camp. I'd keep it in the Metrics department because all other cops with a `Max` parameter that can be adjusted by `--auto-gen-config` are there. I think.

@maxh When I ran `rubocop -a` again on RuboCop sources with `Metrics/LineLength:Max:79` it made some incorrect changes that made specs fail. Can you try that and see if you get the same result?

**Comment by jonas054** [comment]
Following up on my previous comment, I think the issues I found were caused by auto-correcting the length of lines inside comments and heredocs.

A small side issue is that you can get corrections like the following:
```diff
diff --git a/lib/rubocop/cop/lint/percent_symbol_array.rb b/lib/rubocop/cop/lint/percent_symbol_array.rb
index 59762f998..bbfae08aa 100644
--- a/lib/rubocop/cop/lint/percent_symbol_array.rb
+++ b/lib/rubocop/cop/lint/percent_symbol_array.rb
@@ -41,7 +41,10 @@ module RuboCop
             node.children.each do |child|
               range = child.loc.expression
 
-              corrector.remove_trailing(range, 1) if range.source.end_with?(',')
+              if range.source.end_with?(',')
+                corrector.remove_trailing(range,
+                                          1)
+              end
               corrector.remove_leading(range, 1) if
                 range.source.start_with?(':')
             end
```
The line break between `range,` and `1` became unnecessary after `Style/IfUnlessModifier` had changed the `if` statement. I propose to fix these by introducing a new `Layout/UnnecessaryLineBreak` cop.

maxh:
@bbatsov 
>I'm fine with a gradual rollout of auto-correct improvements.
Cool, disabled by default in the latest push! Let's see how this goes.

@jonas054 
>When I ran rubocop -a again on RuboCop sources with Metrics/LineLength:Max:79 it made some incorrect changes that made specs fail. Can you try that and see if you get the same result?
Thanks for trying this. The issue was indeed that lines within comments and HEREDOCS were being split on semi-colons in the change in the previous version. I just reran on the codebase with 79 chars and the only spec failures were on specs that had been hardcoded to expect violations at 80 chars. So, seems the issue is fixed.

changes: https://github.com/rubocop-hq/rubocop/pull/6927/files#diff-fd504989ec8ee8d7f2ce117a1f72b127R153
specs: https://github.com/rubocop-hq/rubocop/pull/6927/files#diff-b206bfd15578ae3e403e777e7ebeceb1R892

>The line break between range, and 1 became unnecessary after Style/IfUnlessModifier had changed the if statement. I propose to fix these by introducing a new Layout/UnnecessaryLineBreak cop.

Good call-out. I've noticed a few things like this as well when autocorrecting our whole codebase. On individual PRs, these are easy to fix manually, but yeah it should be fixable programmatically as well for larger migrations and just ease of use in general.

One idea I had here was to build more line-breaking techniques into the LineLength cop. For example, it could detect opportunities to do `MultilineIfModifier` and prioritize those over breaking up method call arguments.

Definitely doable, but this level of complexity would be pushing the limits of the cop abstraction as I understand it. We'd need a prioritization system for different line breaking options like in the new `def breakable_range` that potentially subsumed other cops, and then an autocorrect system that chose different types of autocorrections based on the situations (ie insert a line break vs transforming a conditional). This could probably be pulled out into a mixin. One advantage of this approach over the new cop is that it would avoid thrashing by ensuring the optimal ordering of line-breaks in a centralized fashion. We could easily do both though, as we see fit.

In any case, imo we should merge this PR disabled by default and explore that as a future improvement. As-is, the PR is already changing quite a bit. What do you think?

Thanks again for the reviews!

bbatsov:
Thanks for working on this!

PR #6929, opened 2019-04-13
Add support for method definition parameters to Layout/FirstParameterIndentation

maxh:
Add `IgnoreMethodDefinitions` parameter to `Layout/FirstParameterIndentation` cop and set it to `true` by default to preserve backwards compatibility. When set to `false`, this cop will align the first parameter of method definitions (in addition to method calls).

This change fills a gap in this otherwise complementary suite of cops:

* `AlignHash` and `IndentHash`
* `AlignArray` and `IndentArray`
* `AlignParameters` (both method defs and calls) and `FirstParameterIndentation` (only method calls)

Together, these cops properly indent most multi-line expressions. The gap is that `FirstParameterIndentation` does not operate on method definitions. This PR fixes that.

Side note: if we were not constrained by the existing names, I would suggest breaking out into `AlignArguments` and `IndentArgument` (for method calls) and `AlignParameters` and `IndentParameter` (for method defs). However, working within the confines of the existing API surface, I think this PR is the cleanest way to add support for this feature.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* Add `IgnoreMethodDefinitions` option to `Layout/FirstParameterIndentation` to support indent method definition params. ([@maxh][])
+
 ### Bug fixes
 
 * [#6914](https://github.com/rubocop-hq/rubocop/issues/6914): [Fix #6914] Fix an error for `Rails/RedundantAllowNil` when with interpolations. ([@Blue-Pix][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -622,6 +622,7 @@ Layout/FirstParameterIndentation:
   VersionAdded: '0.49'
   VersionChanged: '0.56'
   EnforcedStyle: special_for_inner_method_call_in_parentheses
+  IgnoreMethodDefinitions: true
   SupportedStyles:
     # The first parameter should always be indented one step more than the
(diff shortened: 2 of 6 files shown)


bbatsov:
> Side note: if we were not constrained by the existing names, I would suggest breaking out into AlignArguments and IndentArgument (for method calls) and AlignParameters and IndentParameter (for method defs). However, working within the confines of the existing API surface, I think this PR is the cleanest way to add support for this feature.

We are not. I'd rather cause some breakage before 1.0 (hopefully a minimal one) now and fix the problem properly.

maxh:
>We are not. I'd rather cause some breakage before 1.0 (hopefully a minimal one) now and fix the problem properly.

Sounds good! Let me see if I can make that change.

maxh:
>I'd rather cause some breakage before 1.0 (hopefully a minimal one) now and fix the problem properly.

I've cleaned up the Alignment and Indentation cops in the PRs below. Please take a look!

https://github.com/rubocop-hq/rubocop/pull/6981 "Extract AlignArguments out from AlignParameters"
https://github.com/rubocop-hq/rubocop/pull/6982 "Add IndentFirstArgument and IndentFirstParameter"
https://github.com/rubocop-hq/rubocop/pull/6983 "Rename IndentArray and IndentHash to IndentFirstArrayElement and IndentFirstHashElement"

PR #1668, opened 2015-02-20
Add rails/time_zone cop

palkan:
Added `Rails::TimeZone` cop according to the [post](http://danilenko.org/2012/7/6/rails_timezones/)

PR #6930, opened 2019-04-14
[Fix #6900] Rails/TimeZone cop in strict mode

### LINKED ISSUE #6900: Rails/TimeZone EnforcedStyle: strict incorrectly autocorrects
Related issue: [Rails/TimeZone recommends `DateTime.zone.now`, which doesn't exist ](https://github.com/rubocop-hq/rubocop/issues/2151)
Related comment (that's still relevant): https://github.com/rubocop-hq/rubocop/issues/2151#issuecomment-131490820

I've found several cases where Rails/TimeZone with EnforcedStyle: 'strict', autocorrects to a code that raises syntax error.

```ruby
-          caller_reference: "#{file_name}-#{Time.current}"
+          caller_reference: "#{file_name}-#{Time.zone.current}"
```

The above code raises the following error:

```ruby
pry(main)> Time.zone.current
NoMethodError: undefined method `current' for #<ActiveSupport::TimeZone:0x00007fed573edf40>
from (pry):7:in `<main>'
```

Same goes for probably all DateTime autocorrect fixes. All of the fixes append either `.zone` or `.in_time_zone`. None of these two methods exist on `DateTime` class.

```ruby
pry(main)> DateTime.zone
NoMethodError: undefined method `zone' for DateTime:Class
from (pry):6:in `<main>'
pry(main)> DateTime.in_time_zone
NoMethodError: undefined method `in_time_zone' for DateTime:Class
from (pry):8:in `<main>'
```

I think the fix provided in v0.33 only fixed the message shown to the user, while autocorrect still produces the same, incorrect output.

Seems like the tests are wrong. This line tests for `DateTime.zone.now` (`DateTime.zone` doesn't exist):
https://github.com/rubocop-hq/rubocop/blob/master/spec/rubocop/cop/rails/time_zone_spec.rb#L77

--------

## Expected behavior

The autocorrect produces code with no syntax errors.

## RuboCop version

```
$ bundle exec rubocop -V
0.67.2 (using Parser 2.6.2.1, running on ruby 2.5.3 x86_64-darwin17)
```

vfonic:
* Split tests for DANGEROUS_METHODS and :current, which is a special case
* Use preferred `#{klass}.zone.#{method}` style, as opposed to "acceptable" `#{klass}.#{method}.in_time_zone`
* Rename TIMECLASS contsant to TIMECLASSES as it represents more than one time class
* Prefer `Time` class over `DateTime` in strict mode
* Fix `Time.current` => `Time.zone.now` instead of `Time.zone.current` which raises error

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#6900](https://github.com/rubocop-hq/rubocop/issues/6900): Fix `Rails/TimeZone` autocorrect `Time.current` to `Time.zone.now`. ([@vfonic][])
+* [#6900](https://github.com/rubocop-hq/rubocop/issues/6900): Fix `Rails/TimeZone` to prefer `Time.zone.#{method}` over other acceptable corrections. ([@vfonic][])
+
 ## 0.68.0 (2019-04-29)
 
 ### New features
--- a/lib/rubocop/cop/rails/time_zone.rb
+++ b/lib/rubocop/cop/rails/time_zone.rb
@@ -6,7 +6,7 @@ module RuboCop
       # This cop checks for the use of Time methods without zone.
       #
       # Built on top of Ruby on Rails style guide (https://github.com/rubocop-hq/rails-style-guide#time)
-      # and the article http://danilenko.org/2012/7/6/rails_timezones/ .
+      # and the article http://danilenko.org/2012/7/6/rails_timezones/
       #
       # Two styles are supported for this cop. When EnforcedStyle is 'strict'
       # then only use of Time.zone is allowed.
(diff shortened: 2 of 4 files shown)


vfonic:
The only issue this cop doesn't fix is if you're using "flexible" mode and already have a code like this: `Time.zone.now.in_time_zone`, because this cop doesn't report an issue when `zone` or `in_time_zone` methods are in method chain.

Also, I'd suggest making "strict" mode the default. "flexible" mode allows for this code:

```ruby
pry(main)> Time.zone = 'GMT'
# flexible allowed
pry(main)> Time.now.beginning_of_day.in_time_zone
Sat, 13 Apr 2019 17:00:00 GMT +00:00
# strict
pry(main)> Time.zone.now.beginning_of_day
Sun, 14 Apr 2019 00:00:00 GMT +00:00
```

The above code will be hidden under the false assumption that rubocop detected all time zone related issues.

bbatsov:
> Also, I'd suggest making "strict" mode the default. "flexible" mode allows for this code:

Frankly, at this point I don't even remember why we have two modes. @rubocop-hq/rubocop-core Can anyone refresh my memory?

jonas054:
> Frankly, at this point I don't even remember why we have two modes. @rubocop-hq/rubocop-core Can anyone refresh my memory?

I found in #1668 the two mode were present from the beginning, and that the idea seems to come from [the referenced blog post](http://danilenko.org/2012/7/6/rails_timezones/), which talks about correct, acceptable, and wrong ways to handle time zones.

vfonic:
Hey @jonas054 that's a good point! It was probably introduced as a parity with that blog post. We could make the cop simpler by always suggesting using `Time.zone`.

I think the issue here might be that many people (me included) don't know or don't check all the cops and just run `rubocop`. The default here is "flexible", which in turn suggests "Acceptable" instead of "Correct" (as per blog post) option.

I guess, even for flexible mode, we should suggest "Correct" option. Autocorrect should also correct to "Correct". The only difference would be that flexible mode accepts "Acceptable" solutions, while strict mode doesn't.

...or, as said. we can remove flexible and just enforce "Correct". 🤔 At this point, where rubocop was correcting to "Acceptable", maybe it's better to keep flexible as only solution, remove strict, but when suggesting and autocorrecting, correct to "Correct". What do you think?

vfonic:
Bump! I just rebased this on top of `master`

vfonic:
Bump! I just rebased this on top of `master` (0.68)

jonas054:
Sorry for being slow to respond. The changes look good to me, but I'd appreciate a second opinion from  @rubocop-hq/rubocop-core.

vfonic:
No worries! I understand it takes time to review these changes.

I'm just selfishly trying to push my PRs while I have time to keep them in sync with `master`. :D

bbatsov:
Thanks for tackling this! Probably it's worth reviewing the related sections of the Rails Style Guide as well.

PR #6935, opened 2019-04-15
[Fix #6931] `Layout/AccessModifierIndentation` outdent style now ignores modifiers with parameters

### LINKED ISSUE #6931: `Layout/AccessModifierIndentation` should only apply to bare access modifiers
## Expected behavior

I would expect the `Layout/AccessModifierIndentation` cop to only apply to bare access modifiers (that do not apply to a specific method like `private :foo`). I think my expectation comes from the experience of having never seen code following the recommendation this cop makes for these modifiers when using the `outdent` style. But also because otherwise, it makes it impossible to use this cop with `outdent` style in combination with `Layout/IndentationConsistency` because they contradict each other.

For example, I've seen this style in the wild

```ruby
module A
  class B
    def a
    end
    protected :a

    def b
    end

  private

    def c
    end
  end
end
```

but not this one (the one currently recommended)

```ruby
module A
  class B
    def a
    end
  protected :a

    def b
    end

  private

    def c
    end
  end
end
```

## Actual behavior

I would expect the first style to be followed, so my take would be to consider only bare access modifiers in this cop and rename it to `Layout/BareAccessModifierIndentation`. Non bare access modifiers would always be expected to have the same indentation as the method they apply to, which can be already enforced by the `Layout/IndentationConsistency` cop.

Try to use both `Layout/AccessModifierIndentation` with `outdent` style and `Layout/IndentationConsistency` over the first code sample.

## RuboCop version

```
$ [bundle exec] rubocop -V
0.67.2 (using Parser 2.5.1.2, running on ruby 2.5.1 x86_64-linux)
```

deivid-rodriguez:
This is the fix I propose for #6931.

It doesn't include the remane of `Layout/AccessModifierIndentation` to `Layout/BareAccessModifierIndentation` though. I can add that if needed.

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * [#6778](https://github.com/rubocop-hq/rubocop/issues/6778): Fix a false positive in `Style/HashSyntax` cop when a hash key is an interpolated string and EnforcedStyle is ruby19_no_mixed_keys. ([@tatsuyafw][])
 * [#6902](https://github.com/rubocop-hq/rubocop/issues/6902): Fix a bug where `Naming/RescuedExceptionsVariableName` would handle an only first rescue for multiple rescue groups. ([@tatsuyafw][])
 * [#6860](https://github.com/rubocop-hq/rubocop/issues/6860): Prevent auto-correct conflict of `Style/InverseMethods` and `Style/Not`. ([@hoshinotsuyoshi][])
+* [#6935](https://github.com/rubocop-hq/rubocop/issues/6935): `Layout/AccessModifierIndentation` should ignore access modifiers that apply to specific methods. ([@deivid-rodriguez][])
 
 ### Changes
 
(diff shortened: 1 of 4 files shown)


bbatsov:
The change looks good to me. Can you confirm it's consistent with how they do things in Rails?

bbatsov:
One more thing - you can also add some test cases of a modifier being used directly with `def`.

deivid-rodriguez:
Thanks for having a look @bbatsov!

Rails doesn't use this cop. They do have `Style/IndentationConsistency` enabled with `EnforcedStyle: Rails` style, which still passes after my changes. Is that what you wanted me to check?

Regarding the new tests, I'll add them.

Should I also rename the cop?

bbatsov:
> Should I also rename the cop?

No, I'd rather just update the cop documentation to be more explicit about its scope and what's outside of it.

deivid-rodriguez:
I tried to clarify the docs a bit, and added a changelog entry. I tagged it as a bug fix. The alternative would be to tag this as a change and change the `VersionChanged` attribute in the cop, I guess?

bbatsov:
It's fine as it is. Thanks!

deivid-rodriguez:
Great, thanks!

PR #6942, opened 2019-04-17
[Fix #6926] Allow nil values to unset config defaults

### LINKED ISSUE #6926: Cannot unset default InverseMethods in Style/InverseMethods
The default setting of `Style/InverseMethods` includes:
```yaml
Style/InverseMethods:
  InverseMethods:
    :any?: :none?
    :even?: :odd?
    :==: :!=
    :=~: :!~
    :<: :>=
    :>: :<=
```
I would like to unset the `:<` and `:>` entries, as the inverse methods are not correct when comparing classes, e.g.
```ruby
!(Integer < String)
```
is not equivalent to
```ruby
Integer >= String
```
In reading https://github.com/rubocop-hq/rubocop/issues/4289 as well as https://github.com/rubocop-hq/rubocop/blob/master/manual/configuration.md#inheritance-of-hashes-vs-other-types my understanding is that i should be able to unset them with:
```yaml
Style/InverseMethods:
  InverseMethods:
    :<: ~
    :>: ~
```
--------

## Expected behavior

I would expect the above to result in (via `rubocop --show-cops`):

```yaml
Style/InverseMethods:
  InverseMethods:
    :any?: :none?
    :even?: :odd?
    :==: :!=
    :=~: :!~
```

## Actual behavior

Instead the result is:
```yaml
Style/InverseMethods:
  InverseMethods:
    :any?: :none?
    :even?: :odd?
    :==: :!=
    :=~: :!~
    :<:
    :>:
```

This will continue to erroneously flag code such as `!(a < b)` (which, in addition, is autocorrected to `a  b`).

I have written a failing [test](https://github.com/dduugg/rubocop/commit/e43f7b87eb3ae7dbe8745eecaaff2045f72adc12) to capture the desired behavior: 
```ruby
context 'when a file inherits and overrides a hash with nil' do
  let(:file_path) { '.rubocop.yml' }

  before do
    create_file('.rubocop_parent.yml', <<-YAML.strip_indent)
      Style/InverseMethods:
        InverseMethods:
          :any?: :none?
          :even?: :odd?
          :==: :!=
          :=~: :!~
          :<: :>=
          :>: :<=
    YAML

    create_file('.rubocop.yml', <<-YAML.strip_indent)
      inherit_from: .rubocop_parent.yml

      Style/InverseMethods:
        InverseMethods:
          :<: ~
          :>: ~
          :foo: :bar
    YAML
  end

  it 'removes hash keys with nil values' do
    inverse_methods =
      configuration_from_file['Style/InverseMethods']['InverseMethods']
    expect(inverse_methods).to eq(
      '==': :!=,
      '=~': :!~,
      any?: :none?,
      even?: :odd?,
      foo: :bar
    )
  end
end
```

## RuboCop version
0.67.2

dduugg:
This PR adds the ability to unset default configuration values by using an explicit `nil` value (or `~`, in YAML). This is consistent with the [documented](https://github.com/rubocop-hq/rubocop/blob/master/manual/configuration.md#inheritance-of-hashes-vs-other-types) behavior of hash inheritance.

Care must be taken when merging hashes that this only happens when merging into the *default* config hash, which is done by adding an opt to `ConfigLoaderResolver#merge_with_default` that is only enabled when calling from `ConfigLoaderResolver#merge_with_default`. (Otherwise, merging a config hash `B` with an explicit `nil` into `A` would leave an unset value for the config. When this merged hash is merged into the default config, the default value would be unchanged.)

Resolves https://github.com/rubocop-hq/rubocop/issues/6926

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Do not annotate message with cop name in JSON output. ([@elebow][])
 * [#6914](https://github.com/rubocop-hq/rubocop/issues/6914): [Fix #6914] Fix an error for `Rails/RedundantAllowNil` when with interpolations. ([@Blue-Pix][])
 * [#6941](https://github.com/rubocop-hq/rubocop/issues/6941): Add missing absence validations to `Rails/Validation`. ([@jmanian][])
+* [#6926](https://github.com/rubocop-hq/rubocop/issues/6926): [Fix #6926] Allow nil values to unset config defaults. ([@dduugg][])
 
 ### Changes
 
--- a/lib/rubocop/config_loader_resolver.rb
+++ b/lib/rubocop/config_loader_resolver.rb
@@ -71,9 +71,8 @@ module RuboCop
         config = handle_disabled_by_default(config, default_configuration)
       end
 
-      Config.new(merge(default_configuration, config,
-                       inherit_mode: config['inherit_mode'] || {}),
-                 config_file)
+      opts = { inherit_mode: config['inherit_mode'] || {}, unset_nil: true }
+      Config.new(merge(default_configuration, config, opts), config_file)
     end
 
     # Return a recursive merge of two hashes. That is, a normal hash merge,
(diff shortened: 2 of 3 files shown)


bbatsov:
The build seems stuck. Can you rebase on `master` to trigger a new build?

dduugg:
@bbatsov thanks for your attention, done!

PR #6965, opened 2019-04-24
NodePattern: Add <> for matching children in any order.

### LINKED ISSUE #6841: Matching any child with NodePattern

marcandre:
These can be used multiple times in the same sequence as well as
nested.

Limitations: A sequence can only contain a single ellipsis (contained
or not within a <> sequence).

Other implementation changes:
- Captures are now an array (i.e. `capture0` is now `captures[0]`)
- Temporary variables like `temp1` were renamed to `node1`.
- Reading tokens until a delimiter was factorized into `tokens_until`

Example of produced code:
```ruby
# Pattern: (hash <(pair (:sym $_)) $...>)

(captures = Array.new(2)) && node0.is_a?(RuboCop::AST::Node) &&
node0.children.size >= 0 &&
node0.is_a?(RuboCop::AST::Node) && node0.hash_type? &&
(captures[1] = []) && node0.children[0..-1].each_with_object({}) { |child1, matched2|
  case
    when !matched2[0] && child1.is_a?(RuboCop::AST::Node) &&
child1.children.size == 1 &&
child1.is_a?(RuboCop::AST::Node) && child1.pair_type? &&
(node3 = child1.children[0]).is_a?(RuboCop::AST::Node) &&
  node3.children.size == 1 &&
  node3.type == :sym &&
  (captures[0] = node3.children[0]; true) then matched2[0] = true
      else captures[1] << child1
    end
}.size == 1
```

Fixes #6841

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#6841](https://github.com/rubocop-hq/rubocop/issues/6841): Node patterns can now match children in any order using `<>`. ([@marcandre][])
 * [#6928](https://github.com/rubocop-hq/rubocop/pull/6928): Add `--init` option for generate `.rubocop.yml` file in the current directory. ([@koic][])
 * Add new `Layout/HeredocArgumentClosingParenthesis` cop. ([@maxh][])
 * [#6895](https://github.com/rubocop-hq/rubocop/pull/6895): Add support for XDG config home for user-config. ([@Mange][], [@tejasbubane][])
--- a/lib/rubocop/node_pattern.rb
+++ b/lib/rubocop/node_pattern.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require 'delegate'
+require 'erb'
+
 # rubocop:disable Metrics/ClassLength, Metrics/CyclomaticComplexity
 module RuboCop
   # This class performs a pattern-matching operation on an AST node.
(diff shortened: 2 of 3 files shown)


marcandre:
Fixed a few small things and even found a strange [JRuby bug](https://github.com/jruby/jruby/issues/5710), all green now.

bbatsov:
Fantastic work, @marcandre!

pirj:
Please correct me if I'm mistaken, but it seems there's a minor mistake in the description:
```
# Pattern: (hash <(pair (:sym $_)) $...>)
```
should probably be:
```
# Pattern: (hash <(pair (sym $_)) $...>)
```

marcandre:
@pirj Both are valid and should give you similar results. Using `:sym` will generate `node.type == :sym` (note: only when it is at the beginning of a sequence!), while `sym` will generate `node.sym_type?` (in any position), and unless it is redefined by the user, `sym_type?` returns `type == :sym` :-)
HTH

marcandre:
PS: I believe you are right though that `(sym ...)` is probably more usual and better than `(:sym ...)`. I modified my description above accordingly.

pirj:
What makes this change double as awesome is that it also works for matching arguments, not only in a hash. 💯

PR #6970, opened 2019-04-26
WIP: Add Style/CollectionMethodsFilter cop

vfonic:
* This cop prefers `Enumerable#filter` over `Enumerable#find_all` and `Enumerable#select`

NOTE: I'm open for suggestions. This seems like a new approach, but maybe this approach could be used in other cops when there's a difference between what rubocop detects as a failure depending on the ruby (or rails) version.

There are potential issues with this approach:
User could specify:

```yaml
Style/CollectionMethods:
  PreferredMethods:
    find_all: 'select'
Style/CollectionMethodsFilter:
  PreferredMethods:
    select: 'find_all'
```
(maybe?)

-----------------

* [x] Added tests.


spec/rubocop/cop/style/collection_methods_spec.rb
```diff
@@ -17,14 +17,14 @@
   cop_config['PreferredMethods'].each do |method, preferred_method|
     it "registers an offense for #{method} with block" do
       inspect_source("[1, 2, 3].#{method} { |e| e + 1 }")
-      expect(cop.offenses.size).to eq(1)
```
vfonic: I removed this redundant checks, because the next expectation checks that array has exactly one element.

This improves test failure error messages as well.

Before:

```
1) RuboCop::Cop::Style::CollectionMethods registers an offense for collect with proc param
     Failure/Error: expect(cop.offenses.size).to eq(2)

       expected: 2
            got: 1

       (compared using ==)
     # ./spec/rubocop/cop/style/collection_methods_spec.rb:27:in `block (3 levels) in <top (required)>'
```

After: 
```
 1) RuboCop::Cop::Style::CollectionMethods registers an offense for collect with proc param
     Failure/Error:
       expect(cop.messages)
         .to eq([])

       expected: []
            got: ["Prefer `map` over `collect`."]

       (compared using ==)
     # ./spec/rubocop/cop/style/collection_methods_spec.rb:28:in `block (3 levels) in <top (required)>'
```

spec/rubocop/cop/style/collection_methods_spec.rb
```diff
@@ -42,8 +42,8 @@
     end
 
     it 'auto-corrects to preferred method' do
-      new_source = autocorrect_source('some.collect(&:test)')
```
vfonic: This was a bug from before where it tested `collect: 'map'` for every preferred method, instead of testing every `method: 'preferred_method'` autocorrect

vfonic:
I just rebased this on master @ 0.68

bbatsov:
I've been thinking about this and I don't like adding one new cop just to handle a subset of what CollectionMethods is supposed to handle. I think we should adjust the existing cop to be a bit smarter when it comes to those 3 aliases - e.g. we can make the mapping take one or more names to account for this special case. What do you think?

vfonic:
@bbatsov I agree that this solution looks "hacky".

One idea that came to me when I made this approach, maybe something to think for in the future:
This approach allows us to remove all the `if target_ruby_version...` kind of code.
We subclass existing cop and only override methods that should act differently for different ruby (or rails) version.
It would be great if we could make this, yet keep the same cop.

In this case that would be that for ruby <= 2.5, we'd use `RuboCop::Cop::Style::CollectionMethods` for Style/CollectionMethods cop, while for ruby >= 2.6 we'd use `RuboCop::Cop::Style::CollectionMethodsFilter` cop.

I'll just leave this comment here. :D

I'm going to update this PR with a proper fix, within the existing cop.

Thanks for taking the time to look into this!

PR #6971, opened 2019-04-26
Revert "Test with JIT compiler on latest Ruby"

bquorning:
This reverts commit b00329ebf3542cff2dc1ed3abadfded73f09f356 / PR #6456.

I don't think we have seen any value from running tests with the JIT compiler enabled – so let’s just disable it again.

-----------------

* [ ] Added tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -182,24 +182,6 @@ jobs:
     steps:
       *rubocop_steps
 
-  # With JIT compiler enabled!
-  ruby-head-spec-with-jit:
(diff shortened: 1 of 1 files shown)


koic:
I agree. 
AFAIK, MJIT may be disabled by default in Ruby 3.0.  This is because memory usage increases instead of gaining runtime speed. For example, there is a risk that it won't work with Heroku's smallest dyno.
Of course, I don't know yet what will happen in the future :-)

PR #6981, opened 2019-04-27
Extract AlignArguments out from AlignParameters

maxh:
As suggested in https://github.com/rubocop-hq/rubocop/pull/6929, this PR extracts method call argument alignment logic from method definition parameter alignment logic and cleans ups the messages a bit. Also added a few more examples.

Both of these cops are enabled by default, so this should not be a breaking change in the default configuration. However, if folks override the default behavior for `Layout/AlignParameters`, they'll need to update their configs to override for this new `Layout/AlignArguments` as well.

I suggest releasing this in the same version as https://github.com/rubocop-hq/rubocop/pull/6982 and https://github.com/rubocop-hq/rubocop/pull/6983 so that folks only have to go through one set of breaking changes.

-----------------

* [x] Added tests.


bbatsov:
Thanks!

PR #6982, opened 2019-04-27
Add IndentFirstArgument and IndentFirstParameter

maxh:
As proposed in #6929, this PR (1) renames `Layout/FirstParameterIndentation` to `Layout/IndentFirstArgument` because it operates on arguments in method calls and (2) adds a new cop with the same name `Layout/IndentFirstParameter` that operates on parameters in method definitions.

The reason to change the naming scheme from `FirstFooIndentation` to `IndentFirstFoo` is for consistency and ease of use in the docs. See also https://github.com/rubocop-hq/rubocop/pull/6983.

Because indenting the first element of method definition shares logic with array/hash literal alignment, this PR renames `ArrayHashIndentation` mixin and uses it in the cop added in (2). This PR also cleans up the use of `argument` and `parameter` throughout the relevant docs.

This PR will be a minor breaking change due to the rename. Both cops are enabled by default.

I suggest releasing this in the same version as https://github.com/rubocop-hq/rubocop/pull/6983 and https://github.com/rubocop-hq/rubocop/pull/6981 so that folks only have to go through one set of breaking changes.
 
-----------------

* [x] Added tests.


bbatsov:
Thanks!

PR #6988, opened 2019-04-28
Fix edge cases for HeredocArgumentClosingParenthesis

maxh:
This fixes a few edge cases I noticed running this on the Flexport codebase.

### 1) Autocorrect multiple calls on the same line
 
Previously, the cop was not able to autocorrect the case below because it found a violation on the outer `query.order` call instead of the inner `Arel.sql` call. This PR fixes that and adds a spec.

```rb
          query.order(Arel.sql(<<-SQL,
            foo
          SQL
                              ))
```

Now that autocorrects to this after two passes, first finding the inner violation, then the outer:

```rb
          query.order(Arel.sql(<<-SQL))
            foo
          SQL
```

### 2) Detect and autocorrect call that starts on an earlier line

Previously the cop conservatively ignored these cases, but now it fixes them.

Old:
```rb
          query.joins({
            foo: []
          }).order(Arel.sql(<<-SQL),
            bar
          SQL
                  )
```

New:
```rb
          query.joins({
            foo: []
          }).order(Arel.sql(<<-SQL))
            bar
          SQL
```

### 3) Ignore non-idiomatic cases where HEREDOC is receiver of send node

```
          description(
            <<-TEXT
            foobarbaz
          TEXT
          .foo
          )
```

Autocorrecting these cases is beyond the scope of this cop. A separate cop should fix these cases.

This cop has not been released yet, so did not update the CHANGELOG.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/heredoc_argument_closing_parenthesis.rb
+++ b/lib/rubocop/cop/layout/heredoc_argument_closing_parenthesis.rb
@@ -58,13 +58,13 @@ module RuboCop
 
         STRING_TYPES = %i[str dstr xstr].freeze
         def on_send(node)
-          heredoc = extract_heredoc_argument(node)
-          return unless heredoc
+          heredoc_arg = extract_heredoc_argument(node)
+          return unless heredoc_arg
 
-          outermost_send = outermost_send_with_parens(heredoc)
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

PR #6993, opened 2019-04-29
Allows for empty if blocks

RicardoTrindade:
On version 0.68, if you an empty if block, rubocop throws an error. This PR aims to fix this error, allowing for empty blocks with comments. Would like to get your opinion on this, do you think there's a need for a new cop that does not allow for empty if blocks, or if this fix is enough.
```
 NoMethodError:
       undefined method `type' for nil:NilClass
```
Error occured with:
0.68.0 (using Parser 2.6.3.0, running on ruby 2.3.7 x86_64-linux)
-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/safe_navigation.rb
+++ b/lib/rubocop/cop/style/safe_navigation.rb
@@ -72,7 +72,7 @@ module RuboCop
 
         # if format: (if checked_variable body nil)
         # unless format: (if checked_variable nil body)
-        def_node_matcher :modifier_if_safe_navigation_candidate?, <<-PATTERN
+        def_node_matcher :modifier_if_safe_navigation_candidate, <<-PATTERN
           {
             (if {
                   (send $_ {:nil? :!})
@@ -147,7 +147,7 @@ module RuboCop
 
         def extract_parts_from_if(node)
           variable, receiver =
-            modifier_if_safe_navigation_candidate?(node)
+            modifier_if_safe_navigation_candidate(node)
 
           checked_variable, matching_receiver, method =
             extract_common_parts(receiver, variable)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/safe_navigation.rb
```diff
@@ -151,7 +151,12 @@ def extract_parts_from_if(node)
 
           checked_variable, matching_receiver, method =
             extract_common_parts(receiver, variable)
-          matching_receiver = nil if LOGIC_JUMP_KEYWORDS.include?(receiver.type)
+
```
Darhazer: I would go first by renaming `modifier_if_safe_navigation_candidate?` to just `modifier_if_safe_navigation_candidate` as it doesn't return a boolean value. Next, if the receiver is returned, it is a Node for sure, so no need for the second check (receiver.type). So it may become `matching_receiver = nil if receiver.nil? || LOGIC_JUMP_KEYWORDS.include?(receiver.type)`. Which isLOGIC_JUMP_KEYWORDS a bit long, so it may be extracted in a method

RicardoTrindade: Alright, thanks! Made those changes, just not sure about the method extracted being called `safeguarded?`, maybe not the best of names.

lib/rubocop/cop/style/safe_navigation.rb
```diff
@@ -250,6 +252,10 @@ def add_safe_nav_to_all_methods_in_chain(corrector,
             break if ancestor == method_chain
           end
         end
+
+        def safeguarded?(receiver)
+          receiver.nil? || LOGIC_JUMP_KEYWORDS.include?(receiver.type)
```
bbatsov: Won't this be slightly better as `receiver && LOGIC_JUMP_KEYWORDS.include?(receiver.type)`? I'm not sure this method is needed, though, as I think it's name is confusing.

RicardoTrindade: Agreed, removed that method and applied your suggestion.

bbatsov:
Thanks!

PR #6994, opened 2019-04-29
Mark Naming/RescuedExceptionsVariableName as unsafe

vfonic:
I just ran `rubocop --safe-auto-correct` and it made this autocorrect change:

```ruby
-   rescue ActiveResource::Redirection => redirection
+   rescue ActiveResource::Redirection => e
      redirect_to redirection.response['Location']
    end
```

This is obviously a breaking change. I believe the easiest fix here is to mark this cop as unsafe, as, as far as I know, rubocop doesn't offer the extensive functionality for renaming a variable (`redirection`) across all of its usages.

I'm open for suggestions for better fix.

EDIT: I dug down and realized this has just recently been added by @anthony-robin (Thanks!) in this PR #6905. @anthony-robin, what do you think of marking this cop unsafe?

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1997,8 +1997,9 @@ Naming/PredicateName:
 Naming/RescuedExceptionsVariableName:
   Description: 'Use consistent rescued exceptions variables naming.'
   Enabled: true
+  Safe: false
   VersionAdded: '0.67'
-  VersionChanged: '0.68'
+  VersionChanged: '0.69'
   PreferredName: e
 
 Naming/UncommunicativeBlockParamName:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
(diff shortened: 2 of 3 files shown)


config/default.yml
```diff
@@ -1997,8 +1997,9 @@ Naming/PredicateName:
 Naming/RescuedExceptionsVariableName:
   Description: 'Use consistent rescued exceptions variables naming.'
   Enabled: true
+  Safe: false
   VersionAdded: '0.67'
-  VersionChanged: '0.68'
+  VersionChanged: '0.69'
```
vfonic: Not sure about this...

bbatsov:
The cop itself is safe (meaning it doesn't generate false positives), only it's auto-correct should be marked as unsafe.

Darhazer:
I've opened #6998 fixing the auto-correct to rename the variable inside the body as well

vfonic:
> The cop itself is safe (meaning it doesn't generate false positives), only it's auto-correct should be marked as unsafe.

@bbatsov I didn't know that's what "safe" means. Thanks!

@Darhazer awesome! :) Didn't know this is possible. I shouldn't have underestimated rubocop. :)
I'm gonna go ahead and close this PR, your PR is the one that should be merged.

PR #7007, opened 2019-04-30
[Fix #6937] Make Style/BlockDelimiters aware of safe navigation

### LINKED ISSUE #6937: Style/BlockDelimiters Conflicts with Style/MethodCalledOnDoEndBlock When Using &.
There's a circular error when `Style/BlockDelimiters` and `Style/MethodCalledOnDoEndBlock` are both configured where one wants `do/end` while the other wants curly braces.

This is only an issue when the chained method uses `&.`.  If the method is chained via the usual `.`, everything is fine.

--------

## Expected behavior

The two cops should agree.

## Actual behavior

The two cops do not agree.

My configuration for `BlockDelimiters` is as follows:

```
Style/BlockDelimiters:
  Enabled:                                    true
  EnforcedStyle:                              'braces_for_chaining'
  AllowBracesOnProceduralOneLiners:           true
```

Code snippet that fails:

```
      results.find {
        do_a_thing
      }&.fetch('count')
```

## RuboCop version

```
0.67.2 (using Parser 2.6.2.1, running on ruby 2.6.2 x86_64-darwin17)
```

Darhazer:
We have cops that check for send_node. While for some cops it's perfectly fine to work only with send nodes, others were just made before csend type nodes appeared. This PR makes `node#chained?` aware of safe navigation.

I'm not sure for the name that checks for either send or csend. Such helper method, however, will help writing code that is aware of safe navigation. I've refactored few placed that have such checks.

We have to review cops that check explicitly for `send_type?` or have `on_send` but not `on_csend`. For the majority of them handling csend won't make sense, but I guess there might be few more cops that need to be updated.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#7007](https://github.com/rubocop-hq/rubocop/pull/7007): Fix `Style/BlockDelimiters` with `braces_for_chaining` style false positive, when chaining using safe navigation. ([@Darhazer][])
+
+
 ## 0.68.1 (2019-04-30)
 
 ### Bug fixes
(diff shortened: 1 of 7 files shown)


bbatsov:
Great work!

I'm on the fence about the `call_type?` name, but I guess it will do for now.

PR #7016, opened 2019-05-03
[Fix #6969] Fix a false positive in `Style/InverseMethods`

### LINKED ISSUE #6969: Style/InverseMethods not being safe in auto-correct
## Expected behavior
running Style/InverseMethods should consider return arguments of next statements

## Actual behavior

it did change it to a result that's not the same

Simplified example:

```ruby
class TestClass
  def test_method
    [1, 2, 3, 4].select do |number|
      next(false) if number == 4

      !number.odd?
    end
  end
end

print TestClass.new.test_method
```

gets autocorrected to 
```ruby
class TestClass
  def test_method
    [1, 2, 3, 4].reject do |number|
      next(false) if number == 4

      number.odd?
    end
  end
end

print TestClass.new.test_method
```

Which is not the same result:

Console:
```
$ ruby test_class.rb      
[2]                                                                                                                                                                                     
$ rubocop -a test_class.rb
Inspecting 1 file
C

Offenses:

test_class.rb:3:5: C: [Corrected] Style/InverseMethods: Use reject instead of inverting select.
    [1, 2, 3, 4].select do |number| ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

$ ruby test_class.rb      
[2, 4]                                                                                                                                                                                  
```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.67.2 (using Parser 2.6.2.1, running on ruby 2.4.5 x86_64-darwin18)
```

dduugg:
`Style/InverseMethods` does not contemplate guard clauses or other jumps within block methods. This PR updates the cop to skip further evaluation in such cases.

It was sufficient to consider `next` nodes in testing this over a large codebase, but it also guards agains `break` and `return` nodes, out of an abundance of caution.

The `contains_jump?` implementation feels suboptimal, and I'd like to know if there's a better approach. I could neither find an existing helper to look for node types within an AST, nor find a valid `def_node_matcher` that would do the trick.

Resolves https://github.com/rubocop-hq/rubocop/issues/6969

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * [#5782](https://github.com/rubocop-hq/rubocop/issues/5782): Do not autocorrect `Lint/UnifiedInteger` if `TargetRubyVersion < 2.4`. ([@lavoiesl][])
 * [#6387](https://github.com/rubocop-hq/rubocop/issues/6387): Prevent `Lint/NumberConversion` from reporting error with `Time`/`DateTime`. ([@tejasbubane][])
 * [#6980](https://github.com/rubocop-hq/rubocop/issues/6980): Fix `Style/StringHashKeys` to allow string as keys for hash arguments to gsub methods. ([@tejasbubane][])
+* [#6969](https://github.com/rubocop-hq/rubocop/issues/6969): Fix a false positive with block methods in `Style/InverseMethods`. ([@dduugg][])
 
 ### Changes
 
@@ -33,7 +34,8 @@
 * [#6992](https://github.com/rubocop-hq/rubocop/pull/6992): Fix unknown default configuration for `Layout/IndentFirstParameter` cop. ([@drenmi][])
 * [#6972](https://github.com/rubocop-hq/rubocop/issues/6972): Fix a false positive for `Style/MixinUsage` when using inside block and `if` condition is after `include`. ([@koic][])
 * [#6738](https://github.com/rubocop-hq/rubocop/issues/6738): Prevent auto-correct conflict of `Style/Next` and `Style/SafeNavigation`. ([@hoshinotsuyoshi][])
-* [#6847](https://github.com/rubocop-hq/rubocop/pull/6847): Fix `Style/BlockDelimiters` to properly check if the node is chaned when `braces_for_chaining` is set. ([@att14][])
+* [#6847](https://github.com/rubocop-hq/rubocop/pull/6847): Fix `Style/BlockDelimiters` to properly check if the node is chained when `braces_for_chaining` is set. ([@att14][])
+
 ### Bug fixes
 
 * Replace `Time.zone.current` with `Time.zone.today` on `Rails::Date` cop message. ([@vfonic][])
--- a/lib/rubocop/cop/style/inverse_methods.rb
+++ b/lib/rubocop/cop/style/inverse_methods.rb
@@ -28,6 +28,11 @@ module RuboCop
       #   foo == bar
       #   !!('foo' =~ /^\w+$/)
       #   !(foo.class < Numeric) # Checking class hierarchy is allowed
+      #   # Blocks with guard clauses are ignored:
+      #   foo.select do |f|
+      #     next if f.zero?
+      #     f != 1
+      #   end
       class InverseMethods < Cop
         include IgnoredNode
         include RangeHelp
@@ -76,6 +81,7 @@ module RuboCop
           inverse_block?(node) do |_method_call, method, block|
             return unless inverse_blocks.key?(method)
             return if negated?(node) && negated?(node.parent)
+            return if node.each_node(:next).any?
 
             # Inverse method offenses inside of the block of an inverse method
             # offense, such as `y.reject { |key, _value| !(key =~ /c\d/) }`,
--- a/manual/cops_style.md
+++ b/manual/cops_style.md
@@ -2641,6 +2641,11 @@ foo != bar
 foo == bar
 !!('foo' =~ /^\w+$/)
 !(foo.class < Numeric) # Checking class hierarchy is allowed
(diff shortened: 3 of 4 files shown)


lib/rubocop/cop/style/inverse_methods.rb
```diff
@@ -175,6 +178,15 @@ def camel_case_constant?(node)
         def dot_range(loc)
           range_between(loc.dot.begin_pos, loc.expression.end_pos)
         end
+
+        def contains_jump?(node)
+          return false unless node.is_a?(RuboCop::AST::Node)
```
Darhazer: one-liner:
`node.each_node(:return, :break, :next).any?`
each_node starts from the receiver (`node`) and recursively visits child nodes, with optional filtering by types (which is what you are checking), and yields only nodes, so no need to check for the class.

dduugg: ty, done!

CHANGELOG.md
```diff
@@ -23,7 +24,7 @@
 * [#6992](https://github.com/rubocop-hq/rubocop/pull/6992): Fix unknown default configuration for `Layout/IndentFirstParameter` cop. ([@drenmi][])
 * [#6972](https://github.com/rubocop-hq/rubocop/issues/6972): Fix a false positive for `Style/MixinUsage` when using inside block and `if` condition is after `include`. ([@koic][])
 * [#6738](https://github.com/rubocop-hq/rubocop/issues/6738): Prevent auto-correct conflict of `Style/Next` and `Style/SafeNavigation`. ([@hoshinotsuyoshi][])
-* [#6847](https://github.com/rubocop-hq/rubocop/pull/6847): Fix `Style/BlockDelimiters` to properly check if the node is chaned when `braces_for_chaining` is set. ([@att14][])
+* [#6847](https://github.com/rubocop-hq/rubocop/pull/6847): Fix `Style/BlockDelimiters` to properly check if the node is chained when `braces_for_chaining` is set. ([@att14][])
```
dduugg: (unrelated typo fix)

dduugg:
@Drenmi thanks for the suggestion. I've withdrawn skipping blocks with `return`/`break`. The only default block methods are `select`/`reject` (and their `!` counterparts), and i've convinced myself that this cop doesn't break when paired with these other jump methods.

lib/rubocop/cop/style/inverse_methods.rb
```diff
@@ -76,6 +78,7 @@ def on_block(node)
           inverse_block?(node) do |_method_call, method, block|
             return unless inverse_blocks.key?(method)
             return if negated?(node) && negated?(node.parent)
+            return if node.each_node(:next).any?
```
rrosenblum: > It was sufficient to consider `next` nodes in testing this over a large codebase, but it also guards agains `break` and `return` nodes, out of an abundance of caution.

I don't see any code in here that would fix this issue for `break` or `return`. I think this needs to look for those node types as well.

lib/rubocop/cop/style/inverse_methods.rb
```diff
@@ -28,6 +28,8 @@ module Style
       #   foo == bar
       #   !!('foo' =~ /^\w+$/)
       #   !(foo.class < Numeric) # Checking class hierarchy is allowed
+      #   # Blocks with guard clauses are ignored:
+      #   foo.select { |f| next if f.zero?; f != 1 }
```
rrosenblum: We do not recommend using semicolons within code, and we like to have our code examples follow as many best practices as possible. Please change this example to a multi-line block.

```ruby
#   foo.select do |f| 
#     next if f.zero?
#     f != 1
#   end
```

lib/rubocop/cop/style/inverse_methods.rb
dduugg: @rrosenblum it's been reverted after further discussion/consideration. could i trouble you for an example of an issue caused by `break` or `return`?

lib/rubocop/cop/style/inverse_methods.rb
dduugg: done!

lib/rubocop/cop/style/inverse_methods.rb
rrosenblum: I think I missed something from an earlier conversation. It sounds like you had accounted for this and then reverted it. The comment that I quoted made it seem like we needed to account for `break` and `return`. These are likely edge cases because `break` and `return` don't make much sense inside of methods that are designed to have a return.

manual/cops_style.md
```diff
@@ -2658,6 +2658,8 @@ foo != bar
 foo == bar
 !!('foo' =~ /^\w+$/)
 !(foo.class < Numeric) # Checking class hierarchy is allowed
+# Blocks with guard clauses are ignored:
+foo.select { |f| next if f.zero?; f != 1 }
```
rrosenblum: You'll need to regenerate the documentation to update this.

dduugg: d'oh, done!

spec/rubocop/cop/style/inverse_methods_spec.rb
```diff
@@ -50,6 +50,20 @@
     expect_no_offenses('!!foo.reject { |e| !e }')
   end
 
+  it 'allows an inverse method in a block with next' do
+    expect_no_offenses(<<-RUBY.strip_indent)
```
koic: Could you rebase using latest master branch and change it to the following code?

```diff
-expect_no_offenses(<<-RUBY.strip_indent)
+expect_no_offenses(<<~RUBY)
```

dduugg: done!

koic:
Thank you!

PR #7026, opened 2019-05-08
[Fix #6945] Drop support for Ruby 2.2

### LINKED ISSUE #6945: Drop support for Ruby 2.2 and 2.3
Both of them reached their EOL already and according to our official support policy we don't need to support them anymore.

I don't have a strong take on this, but I'd rather remove them mostly to speed up the test suite. The final decision will also depend on how widely 2.2 and 2.3 are still used in the wild, of course. @rubocop-hq/rubocop-core Any thoughts on this?

koic:
Fixes #6945.

This is a big change. The major changes are split into commits and left in the message. Notable changes to users are:

## Drop support for Ruby 2.2

By convention, a year after Ruby became an EOL, it seems that the old Ruby version support has been dropped. I'm not sure if it's good to drop Ruby 2.3 support, but it's good time to drop Ruby 2.2 support.

## Set default `EnforcedStyle` to `squiggly` option for `Layout/IndentHeredoc` and `auto_detection` option is removed

Because we no longer use Ruby 2.2, the original default `auto_detection`  option is never used. Makes `squiggly` option enabled in Ruby 2.3 or higher a new default.

## Set default `EnforcedStyle` to `always` option for `Style/FrozenStringLiteralComment` and `when_needed` option is removed

Because we no longer use Ruby 2.2, the original default `when_needed`  option is never used. Makes `always` option enabled in Ruby 2.3 or higher a new default.

-----------------

* [x] Added tests.


bbatsov:
Seems something's wrong with JRuby.

koic:
JRuby has a squiggly heredoc feature incompatibility with MRI. I updated it to pass the tests.

bbatsov:
Fantastic work! 🙇 

> JRuby has a squiggly heredoc feature incompatibility with MRI. I updated it to pass the tests.

Incompatible in which way?

koic:
Incompatibility have arisen in squiggly heredoc that use single quotes.

```ruby
# example.rb
source = <<~'RUBY'
  def say
    puts 'hello'
  end
RUBY

puts source
```

## ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin17]

```console
% ruby example.rb
def say
  puts 'hello'
end
```

## JRuby jruby 9.2.7.0 (2.5.3) 2019-04-09 8a269e3

```console
% ruby example.rb
def say
puts 'hello'
end
```

It may be related to the following issue.
https://github.com/jruby/jruby/issues/4260

As a result, RuboCop has not replaced the following test code with squiggly heredoc.

```ruby
expect_offense(<<-'RUBY'.strip_indent)
```

PR #7036, opened 2019-05-11
[Fix #6779] Add new cop `Style/NegatedUnless`

### LINKED ISSUE #6779: Style/NegatedIf: option to only support `unless !`

Our team often uses `if !` when it would read cleaner than `unless`. `unless !` however is always a code smell for us as it can always be swapped to the clearer `if`. It would be great to have an option to toggle enforcement for either `if` or `unless` but not both:

```
e.g.

# good
if !going_to_happen?
  #...
end

# good
unless going_to_happen?
  #...
end

# bad
unless !going_to_happen?
  # ...
end

# good
if going_to_happen?
  # ...
end

# good
do_the_thing if !going_to_happen?

# good
do_the_thing unless going_to_happen?

# bad
do_the_thing unless !going_to_happen?

# good
do_the_thing if going_to_happen?
```

It would be nice to be able to choose to only support modifier `unless` (and `if !` elsewhere) but I am less concerned about that

tejasbubane:
To provide negative condition enforcement on either `if` or `unless` but not both.
Closes #6779 

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -3507,6 +3507,21 @@ Style/NegatedIf:
     - prefix
     - postfix
 
+Style/NegatedUnless:
+  Description: >-
+                 Favor if over unless for negative conditions
```
bbatsov: Breaking the line seems unnecessary here. You forgot the final `.`.

lib/rubocop/cop/style/negated_unless.rb
```diff
@@ -0,0 +1,99 @@
...
+      #
+      #   - both
+      #   - prefix
+      #   - postfix
+      #
+      # @example EnforcedStyle: both (default)
+      #   # enforces `if` for `prefix` and `postfix` conditionals
+      #
+      #   # bad
+      #
```
bbatsov: Please, remove all such blank lines. They don't look good in the rendered docs.

tejasbubane:
@bbatsov Fixed.

PR #7039, opened 2019-05-14
Use Ruby 2.6 parser for `rake documentation_syntax_check`

koic:
### Summary

This PR uses Ruby 2.6 parser for `rake documentation_syntax_check`.

### Other Information

`Lint/UselessElseWithoutRescue` example code uses a syntax that is not supported in Ruby 2.6 or higher.

```consle
% cat example.rb
begin
else
end

% ruby -vc example.rb
ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-darwin17]
example.rb:3: warning: else without rescue is useless
Syntax OKn

% ruby -vc example.rb
ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin17]
example.rb:2: else without rescue is useless
```

https://github.com/ruby/ruby/commit/140512d

Ruby 2.5 parser is used for `Lint/UselessElseWithoutRescue` example code to prevent the following error.

```console
% rake documentation_syntax_check
Files:         515
Modules:        81 (   10 undocumented)
Classes:       482 (    3 undocumented)
Constants:     733 (  726 undocumented)
Attributes:     26 (    0 undocumented)
Methods:      1160 ( 1029 undocumented)
 28.77% documented
lib/rubocop/cop/lint/useless_else_without_rescue.rb: Syntax Error in an
example. else without rescue is useless
```

In the future when Ruby 2.5 support is dropped it can be removed with `Lint/UselessElseWithoutRescue`.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/useless_else_without_rescue.rb
+++ b/lib/rubocop/cop/lint/useless_else_without_rescue.rb
@@ -5,6 +5,9 @@ module RuboCop
     module Lint
       # This cop checks for useless `else` in `begin..end` without `rescue`.
       #
+      # Note: This syntax is no longer valid on Ruby 2.6 or higher and
+      # this cop is going to be removed at some point the future.
+      #
       # @example
       #
       #   # bad
--- a/Rakefile
+++ b/Rakefile
@@ -90,6 +90,7 @@ end
 desc 'Syntax check for the documentation comments'
 task documentation_syntax_check: :yard_for_generate_documentation do
   require 'parser/ruby25'
+  require 'parser/ruby26'
 
   ok = true
   YARD::Registry.load!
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/lint/useless_else_without_rescue.rb
```diff
@@ -4,6 +4,7 @@ module RuboCop
   module Cop
     module Lint
       # This cop checks for useless `else` in `begin..end` without `rescue`.
+      # And the syntax does not support Ruby 2.6 or higher.
```
bbatsov: Maybe we should say here "Note: This syntax is no longer valid on Ruby 2.6 or higher and this cop is going to removed at some point the future."?

koic: Yes, quite right! I updated it.

lib/rubocop/cop/lint/useless_else_without_rescue.rb
```diff
@@ -5,6 +5,9 @@ module Cop
     module Lint
       # This cop checks for useless `else` in `begin..end` without `rescue`.
       #
+      # Note: This syntax is no longer valid on Ruby 2.6 or higher and
+      # this cop is going to removed at some point the future.
```
bbatsov: to be removed

koic: 💦

PR #7045, opened 2019-05-14
Fix mention of XDG env variable

Mange:
I made a mistake in #6895, which should be corrected to avoid confusion on how to use the XDG directory specification.

Thank you @FranklinYu for bringing this to my attention.


bbatsov:
👍

PR #7059, opened 2019-05-20
Add `EnforcedStyle` to `Layout/EmptyLinesAroundAccessModifier`

koic:
This PR adds `EnforcedStyle` to `Layout/EmptyLinesAroundAccessModifier` cop.
This makes it possible to select the following Rails coding convention.

> Indent and no blank line after private/protected.

https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#follow-the-coding-conventions

"Indenting" is already implemented by `Layout/IndentationConsistency` cop.
This PR implements "no blank line after private/protected" by `Layout/IndentationConsistency` cop.

The EnforcedStyle names are `normal` and `rails` which are the same as `Layout/IndentationConsistency` cop.

The default `EnforceStyle` is "`normal`", which keeps the current style so that no breaking changes occur.

I was told about this style by Rails committer @kamipo. Thanks!

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/ast/node/mixin/method_dispatch_node.rb
+++ b/lib/rubocop/ast/node/mixin/method_dispatch_node.rb
@@ -9,6 +9,7 @@ module RuboCop
       include MethodIdentifierPredicates
 
       ARITHMETIC_OPERATORS = %i[+ - * / % **].freeze
+      SPECIAL_MODIFIERS = %w[private protected].freeze
 
       # The receiving node of the method dispatch.
       #
@@ -73,6 +74,15 @@ module RuboCop
       #                   access modifier
       def non_bare_access_modifier?
         macro? && non_bare_access_modifier_declaration?
+      end
+
+      # Checks whether the dispatched method is a bare `private` or `protected`
+      # access modifier that affects all methods defined after the macro.
+      #
+      # @return [Boolean] whether the dispatched method is a bare
+      #                    `private` or `protected` access modifier
+      def special_modifier?
+        bare_access_modifier? && SPECIAL_MODIFIERS.include?(source)
       end
 
       # Checks whether the name of the dispatched method matches the argument
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#6649](https://github.com/rubocop-hq/rubocop/pull/6649): `Layout/AlignHash` supports list of options. ([@stoivo][])
 * Add `IgnoreMethodPatterns` config option to `Style/MethodCallWithArgsParentheses`. ([@tejasbubane][])
+* [#7059](https://github.com/rubocop-hq/rubocop/pull/7059): Add `EnforcedStyle` to `Layout/EmptyLinesAroundAccessModifier`. ([@koic][])
 
 ### Bug fixes
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -509,6 +509,13 @@ Layout/EmptyLinesAroundAccessModifier:
   StyleGuide: '#empty-lines-around-access-modifier'
   Enabled: true
   VersionAdded: '0.49'
+  EnforcedStyle: around
+  SupportedStyles:
+    - around
+    - only_before
+  Reference:
+    # A reference to `EnforcedStyle: only_before`.
+    - https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#follow-the-coding-conventions
 
 Layout/EmptyLinesAroundArguments:
   Description: "Keeps track of empty lines around method arguments."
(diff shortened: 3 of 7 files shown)


bbatsov:
I'm fine with the change, but we can do better with the names. In general we should avoid using names that don't carry any meaning. I think `normal` can be named `around` and `rails` can be `only_after` or `after`.

koic:
Ah, Sure! This sounds like a very nice names. I renamed to `around` and `only_after` for `Layout/EmptyLinesAroundAccessModifier` cop.

kamipo:
Does `only_after` means empty-lines `only_after` access-modifier?

koic:
Oops, I think that the name should have been `only_before`. It means a blank line "before" compared to the "around" blank lines. I updated it.

bbatsov:
Yeah, my bad!

lib/rubocop/ast/node/mixin/method_dispatch_node.rb
```diff
@@ -9,6 +9,7 @@ module MethodDispatchNode
       include MethodIdentifierPredicates
 
       ARITHMETIC_OPERATORS = %i[+ - * / % **].freeze
+      SPECIAL_MODIFIERS = %w[private protected].freeze
```
bbatsov: There's also `public`.

koic: It maintains the behavior extracted from `Layout/IndentationWidth` cop.

```diff
      class IndentationWidth < Cop # rubocop:disable Metrics/ClassLength
         (snip)
-        SPECIAL_MODIFIERS = %w[private protected].freeze
```

https://github.com/rubocop-hq/rubocop/blob/v0.69.0/lib/rubocop/cop/layout/indentation_width.rb#L55

I think that the purpose is to implement the following style. `public` is not included here.

> Indent and no blank line after private/protected.

https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#follow-the-coding-conventions

bbatsov: Hmm, I assume that's just some oversight because `public` is not used very often. I don't see any reason why someone would want to treat it differently from `private` and `protected`.

koic: Yes. I agree that `public` is not used very often.

On the other hand, I investigated about `Layout/IndentationWidth` history.
https://github.com/rubocop-hq/rubocop/pull/1759

I think that `private` and `protected` are intended to be treated specially as internal modifiers. For example, a Rails controller code created by `rails new` is:
https://github.com/rails/rails/blob/v6.0.0.rc1/railties/lib/rails/generators/rails/scaffold_controller/templates/controller.rb.tt#L53-L66

```ruby
  private
    # Use callbacks to share common setup or constraints between actions.
    def set_<%= singular_table_name %>
      @<%= singular_table_name %> = <%= orm_class.find(class_name, "params[:id]") %>
    end

    # Only allow a trusted parameter "white list" through.
    def <%= "#{singular_table_name}_params" %>
      <%- if attributes_names.empty? -%>
      params.fetch(:<%= singular_table_name %>, {})
      <%- else -%>
      params.require(:<%= singular_table_name %>).permit(<%= permitted_params %>)
      <%- end -%>
    end
```

Perhaps one improvement is to change the constant name `SPECIAL_MODIFIERS` to something like `INTERNAL_MODIFIERS`.

The following behavior changes will occur if `public` is added to constant `SPECIAL_MODIFIERS`.
Therefore I'd like to keep the current behavior first.

## An example

```ruby
% cat example.rb
class Test
  public

  def e
  end

  protected

  def f
  end

  private

  def g
  end
end
```

### Do not include `public` in `SPECIAL_MODIFIERS`

```diff
% rubocop -a
% git diff
diff --git a/example.rb b/example.rb
index 44fe815..d24b0f1 100644
--- a/example.rb
+++ b/example.rb
@@ -1,16 +1,15 @@
+# frozen_string_literal: true
+
 class Test
   public

-  def e
-  end
+  def e; end

   protected

-  def f
-  end
+    def f; end

   private

-  def g
-  end
+    def g; end
 end
```

### Include `public` in `SPECIAL_MODIFIERS`

```diff
% rubocop -a
% git diff
diff --git a/example.rb b/example.rb
index 44fe815..8760788 100644
--- a/example.rb
+++ b/example.rb
@@ -1,16 +1,15 @@
+# frozen_string_literal: true
+
 class Test
   public

-  def e
-  end
+    def e; end

   protected

-  def f
-  end
+    def f; end

   private

-  def g
-  end
+    def g; end
 end
```

bbatsov: Hmm, so basically the Rails team doesn't want to outdent stuff under public? 

I guess it might be best to just expose the modifiers as a config, so people would decide if they want to outdent stuff consistently or not. Hardcoding special cases would be a worse solution IMO.

bbatsov:
I'll merge this as is, so I can cut 0.69 and unblock your work on RuboCop Rails 2.0. We can improve it later. Thanks!

PR #7067, opened 2019-05-21
Switch to `receive` message expectation syntax

pirj:
I started off [improving `RSpec/SubjectStub` cop](https://github.com/rubocop-hq/rubocop-rspec/pull/770/) that had a special case of dealing with `expect(subject).to all receive(...)` that was originally present in RuboCop, and I've removed this special case in [this commit](https://github.com/rubocop-hq/rubocop-rspec/pull/770/commits/779c95e206c8beb539a3a970b362ff639323f430), since according to an experiment, it's not raising any single offence for repositories from [`real-world-ruby-apps`](https://github.com/jeromedalbert/real-world-ruby-apps) (11k files) and [`real-world-rails`](https://github.com/eliotsykes/real-world-rails) (79k files).

Original line when a special case was introduced:

    expect(formatter_set).to all(receive(:started).with(files))

later reworked to:

    expect(formatter_set).to all(have_received(:started).with(files))

I've replaced this with:

    expect(formatter_set[0]).to receive(:started).with(files)
    expect(formatter_set[1]).to receive(:started).with(files)

that is quite consistent with the style used in the rest of that spec file.

Also, using spies is not justified in this case.

It doesn't make much sense to use a `before` hook that is only run along with a single example, so I've simplified it as well.

However, I've faced another cop, `RSpec/MessageSpies`, that complains that `receive` is used instead of `have_received`. What confuses me is that according to [the chat from 10 Oct 2016](https://gitter.im/rubocop-rspec/Lobby):

> If a spec example has multiple stubs (allow) and you want a focused spec with just one expectation per example – and RuboCop is now telling you that you must change all those stubs into expectations. Not good.

However, [in `rubocop`](https://github.com/rubocop-hq/rubocop/blob/master/.rubocop_todo.yml#L67) this is not a problem, as multiple expectations per example are allowed:
```
RSpec/MultipleExpectations:
  Max: 25
```

So I decided to give it a shot and take a look what happens if I switch the default enforced style of `RSpec/MessageSpies` from `have_received` to `receive`, and this pull request is the result of this switch.

### Why it makes sense to merge this?

1. `rubocop-rspec`'s `SubjectStub` has a special case that deals with a single known example, here, in `rubocop`'s `formatter_set_spec.rb`.
2. Less hooks are used, the specs are more straightforward.
3. Spies have to be additionally instantiated, but now anymore.
4. Base formatter spec is now "Runner formatter invocation spec", which reflects what it tests.

### Why it wouldn't make sense to merge this?

1. Usage of `expect(...).to receive` breaks Arrange-Act-Assert pattern.

### What is redundant

I've made a couple of changes along the way not directly related to this change, specifically changed `receive { ... }` to `.and_return`. Don't remember the details, but I think it's recommended to use this syntax.

Appreciate any feedback on this.

-----------------

* [-] Added tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -93,5 +93,8 @@ Metrics/ModuleLength:
 RSpec/PredicateMatcher:
   EnforcedStyle: explicit
 
+RSpec/MessageSpies:
+  EnforcedStyle: receive
+
 RSpec/NestedGroups:
   Max: 7
--- a/spec/rubocop/cli_spec.rb
+++ b/spec/rubocop/cli_spec.rb
@@ -1187,13 +1187,12 @@ RSpec.describe RuboCop::CLI, :isolated_environment do
           Exclude:
             - ignored/**
       YAML
-      allow(File).to receive(:open).and_call_original
+      expect(File).not_to receive(:open).with(%r{/ignored/})
       expect(cli.run(%w[--format simple example])).to eq(0)
       expect($stdout.string).to eq(<<~OUTPUT)
 
         0 files inspected, no offenses detected
       OUTPUT
-      expect(File).not_to have_received(:open).with(%r{/ignored/})
     end
 
     it 'can be configured with option to disable a certain error' do
--- a/spec/rubocop/config_spec.rb
+++ b/spec/rubocop/config_spec.rb
@@ -947,23 +947,19 @@ RSpec.describe RuboCop::Config do
         }
       end
 
-      before do
-        allow(File).to receive(:file?).and_call_original
-      end
-
       it 'uses TargetRubyVersion' do
         expect(configuration.target_ruby_version).to eq ruby_version
       end
 
       it 'does not read .ruby-version' do
+        expect(File).not_to receive(:file?).with('.ruby-version')
         configuration.target_ruby_version
-        expect(File).not_to have_received(:file?).with('.ruby-version')
       end
 
       it 'does not read Gemfile.lock or gems.locked' do
+        expect(File).not_to receive(:file?).with('Gemfile')
+        expect(File).not_to receive(:file?).with('gems.locked')
         configuration.target_ruby_version
-        expect(File).not_to have_received(:file?).with('Gemfile')
-        expect(File).not_to have_received(:file?).with('gems.locked')
       end
     end
 
@@ -1029,10 +1025,9 @@ RSpec.describe RuboCop::Config do
         end
 
         it 'does not read Gemfile.lock or gems.locked' do
-          allow(File).to receive(:file?).and_call_original
+          expect(File).not_to receive(:file?).with('Gemfile')
+          expect(File).not_to receive(:file?).with('gems.locked')
           configuration.target_ruby_version
-          expect(File).not_to have_received(:file?).with('Gemfile')
-          expect(File).not_to have_received(:file?).with('gems.locked')
         end
       end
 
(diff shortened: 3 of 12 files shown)


bbatsov:
Nicely done! Thanks!

bquorning:
😢 

I find the main benefit of using `RSpec/MessageSpies` is that the specs can adhere to [the four-phase test pattern](https://thoughtbot.com/blog/four-phase-test), specifically that the “verify” step is after “exercise”.

After this PR, the `expect`ations are sometimes at the bottom of a test example, sometimes at the top, and [sometimes a bit of both](https://github.com/rubocop-hq/rubocop/blob/958cf6862fe91549bea355ba933f7c2a029b573a/spec/rubocop/config_store_spec.rb#L57-L66).

pirj:
@bquorning I had doubts about this change as well, but looking at the changes one by one I was more inclined to do this change rather than keep things as is:

1. [This](https://github.com/rubocop-hq/rubocop/pull/7067/files#diff-136dc9359b5a71d3b9242666420aefc5R1190) is an indicative example of a case when message expectations are used together with value comparison expectations. 
```diff
-     allow(File).to receive(:open).and_call_original
+     expect(File).not_to receive(:open).with(%r{/ignored/})
      expect(cli.run(%w[--format simple example])).to eq(0)
      expect($stdout.string).to eq(<<~OUTPUT)
        0 files inspected, no offenses detected
      OUTPUT
-     expect(File).not_to have_received(:open).with(%r{/ignored/})
```

There was (and still is) no clear Act/Excercise stage, it was mixed with value comparison expectation in a single statement.

The combination of `allow...to` and `expect...not_to` is a bit confusing. It's a technical construct, but it reads quite poorly.

A necessity to `and_call_original` is not clear, especially keeping in mind the `not_to` expectation.

2. [Another example](https://github.com/rubocop-hq/rubocop/pull/7067/files#diff-186eae0bf86fb115d340a976b4990206L951) when `allow.to .and_call_original` is used with `expect.not_to`.

There is no Arrange/Setup phase.

To avoid incidental state (more like behaviour than state), a block style expectation could be used, e.g. [`send_message` from saharspec](https://github.com/zverok/saharspec#send_messageobject-method-matcher).

It reads better, a hook is not used.

Also, I would disagree that setting up a spy counts towards the Setup/Arrange phase.

3. [Swapping](https://github.com/rubocop-hq/rubocop/pull/7067/files#diff-5401b417948dd5f49aee2d14a74aac12L33).

With swapping two parts of the example, e.g.
> when I load the configuration, it's being loaded from a file
vs
> configuration is loaded from a file when I load it

Both sentences read quite similarly, and I hope the examples as well.

4. [Overstubbing](https://github.com/rubocop-hq/rubocop/pull/7067/files#diff-a330b659dfeee605be88ab8826ddfe03R7)

Verified doubles are not used, incidental behaviour might strike, since there are at least two statements in Act/Excercise phase, `parse_source` and `commissioner.investigate`.

5. Scattered expectations.

Can't disagree wit this. Found two cases in this pull request, [1](https://github.com/rubocop-hq/rubocop/pull/7067/files#diff-136dc9359b5a71d3b9242666420aefc5L1196) (mentioned earlierand [2](https://github.com/rubocop-hq/rubocop/pull/7067/files#diff-b61320590de7d06cbb22897d0d1efd81L168). I can imaging an example where a number of message expectations are sent, and then the results are compared to the reference expected values.
Although, the first example had more downsides that just that.

Hiding the Act/Excercise phase in a hook is a bad practice, and this pull request changes this by making a call to `run` visible in example itself.

I really respect AAA, or 4PT, but the usages of message spies that were changed in this specific pull request did not look like a canonical example of it.

Went through the changes once again. I might have a bias for this, since my goal was to get rid of message spy without breaking `ExampleSize` and `MessageSpies`. I might have made a really bad change, appreciate if you point at the worst one. I'm all for improving it in a follow-up.

PR #7068, opened 2019-05-21
Fix Layout/AlignHash when mixed Hash styles are used

### LINKED ISSUE #7066: New Layout/AlignHash has a runtime error on mixed Hash style
## Expected behavior

Rubocop should run the Layout/AlignHash cop without an error occuring.

## Actual behavior

There occurs an error with the message "nil can't be coerced into Integer"

Minimal example:
```ruby
bar(
  foo:  'baz',
  'bar' => 'bla'
)
```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.70.0 (using Parser 2.6.3.0, running on ruby 2.6.3 x86_64-linux)
```

rmm5t:
Fixes #7066. Otherwise, this happens when mixed hash styles are used:

```ruby
      {
        'Content-Type' => 'application/json',
        Authorization: "Bearer #{@access_token}",
      }
```

```
An error occurred while Layout/AlignHash cop was inspecting /Users/rmcgeary/work/ironridge/lib/salesforce/client.rb:72:6.
nil can't be coerced into Integer
.../gems/rubocop-0.70.0/lib/rubocop/cop/mixin/hash_alignment.rb:113:in `+'
.../gems/rubocop-0.70.0/lib/rubocop/cop/mixin/hash_alignment.rb:113:in `value_delta'
.../gems/rubocop-0.70.0/lib/rubocop/cop/mixin/hash_alignment.rb:65:in `deltas'
.../gems/rubocop-0.70.0/lib/rubocop/cop/layout/align_hash.rb:239:in `block (2 levels) in check_pairs'
.../gems/rubocop-0.70.0/lib/rubocop/cop/layout/align_hash.rb:238:in `each'
.../gems/rubocop-0.70.0/lib/rubocop/cop/layout/align_hash.rb:238:in `block in check_pairs'
.../gems/rubocop-0.70.0/lib/rubocop/cop/layout/align_hash.rb:237:in `each'
.../gems/rubocop-0.70.0/lib/rubocop/cop/layout/align_hash.rb:237:in `check_pairs'
.../gems/rubocop-0.70.0/lib/rubocop/cop/layout/align_hash.rb:208:in `on_hash'
```

-----------------

* [X] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/hash_alignment.rb
+++ b/lib/rubocop/cop/mixin/hash_alignment.rb
@@ -83,6 +83,10 @@ module RuboCop
       class TableAlignment
         include ValueAlignment
 
+        def initialize
+          self.max_key_width = 0
+        end
+
         def deltas_for_first_pair(first_pair, node)
           self.max_key_width = node.keys.map { |key| key.source.length }.max
 
--- a/spec/rubocop/cop/layout/align_hash_spec.rb
+++ b/spec/rubocop/cop/layout/align_hash_spec.rb
@@ -765,6 +765,15 @@ RSpec.describe RuboCop::Cop::Layout::AlignHash, :config do
       RUBY
     end
 
+    it 'accepts aligned hash keys with mixed hash style' do
+      expect_no_offenses(<<~RUBY)
+        headers = {
+          "Content-Type" => 0,
+          Authorization: 1
+        }
+      RUBY
+    end
+
     it 'accepts an empty hash' do
       expect_no_offenses('h = {}')
     end
@@ -807,6 +816,16 @@ RSpec.describe RuboCop::Cop::Layout::AlignHash, :config do
             'b' => 1,
                   'c' => 1
                   ^^^^^^^^ Align the elements of a hash literal if they span more than one line.
+          }
+        RUBY
+      end
+
+      it 'for misaligned hash keys with mixed hash style' do
+        expect_offense(<<~RUBY)
+          headers = {
+            "Content-Type" => 0,
+             Authorization: 1
+             ^^^^^^^^^^^^^^^^ Align the elements of a hash literal if they span more than one line.
           }
         RUBY
       end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#7066](https://github.com/rubocop-hq/rubocop/issues/7066): Fix `Layout/AlignHash` when mixed Hash styles are used. ([@rmm5t][])
+
 ### Changes
 
 * [#5976](https://github.com/rubocop-hq/rubocop/issues/5976): Warn for Rails Cops. ([@koic][])


spec/rubocop/cop/layout/align_hash_spec.rb
```diff
@@ -319,6 +319,16 @@ def example
...
 
+    it 'registers an offense for misaligned hash keys with mixed hash style' do
+      expect_offense(<<~RUBY)
+        headers = {
+          "Content-Type" => 0,
+           Authorization: 1
+           ^^^^^^^^^^^^^^^^ Align the elements of a hash literal if they span more than one line.
+        }
+      RUBY
+    end
```
koic: `nil can't be coerced into Integer` does not reproduce when reverting product code.

```diff
diff --git a/lib/rubocop/cop/mixin/hash_alignment.rb b/lib/rubocop/cop/mixin/hash_alignment.rb
index 59d078c55..228bb73d6 100644
--- a/lib/rubocop/cop/mixin/hash_alignment.rb
+++ b/lib/rubocop/cop/mixin/hash_alignment.rb
@@ -102,7 +102,7 @@ module RuboCop
         end

         def hash_rocket_delta(first_pair, current_pair)
-          first_pair.loc.column + max_key_width.to_i + 1 -
+          first_pair.loc.column + max_key_width + 1 -
             current_pair.loc.operator.column
         end

@@ -111,7 +111,7 @@ module RuboCop

           correct_value_column = first_pair.key.loc.column +
                                  current_pair.delimiter(true).length +
-                                 max_key_width.to_i
+                                 max_key_width
           correct_value_column - current_pair.value.loc.column
         end
       end
```

```console
% bundle exec rspec spec/rubocop/cop/layout/align_hash_spec.rb
Run options: include {:focus=>true}

All examples were filtered out; ignoring {:focus=>true}

Randomized with seed 52306
......................................................................................

Finished in 0.19007 seconds (files took 2 seconds to load)
86 examples, 0 failures

Randomized with seed 52306
```

Test cases for the following .rubocop.yml setting need to be written.

```yaml
Layout/AlignHash:
  EnforcedHashRocketStyle:
    - key
    - table
```

The existing `let(:cop_config)` may be equivalent to that.

```ruby
     let(:cop_config) do
       {
         'EnforcedHashRocketStyle' => %w[key table],
```

spec/rubocop/cop/layout/align_hash_spec.rb
```diff
@@ -379,6 +389,15 @@ def example
...
     end
 
+    it 'accepts aligned hash keys with mixed hash style' do
+      expect_no_offenses(<<~RUBY)
+        headers = {
+          "Content-Type" => 0,
+          Authorization: 1
+        }
+      RUBY
+    end
```
koic: Probably the same as above.

spec/rubocop/cop/layout/align_hash_spec.rb
rmm5t: Yes, you are absolutely correct. I originally wrote those specs in the `%w[key table]` section when I wrote the fix, but then moved them thinking it was more of a generic problem. That was silly of me.

Fixed and force-pushed to correct it.

lib/rubocop/cop/mixin/hash_alignment.rb
```diff
@@ -102,7 +102,7 @@ def key_delta(first_pair, current_pair)
         end
 
         def hash_rocket_delta(first_pair, current_pair)
-          first_pair.loc.column + max_key_width + 1 -
+          first_pair.loc.column + max_key_width.to_i + 1 -
```
bbatsov: Why isn't this a number in the first place? Probably we can convert it earlier and this would yield slightly better code.

rmm5t: Fair question. I found this code to be a bit of spaghetti and a tad hard to follow. This value is initialized to nil as an attr_accessor, and its usage expects `deltas_for_first_pair` to have been called first, but it apparently wasn’t under the particular scenario that caused the error. 

Are you asking for an initializer that initialized this value to zero?

rmm5t: To clarify, for now, I’m just trying to keep the bug fix as small as possible considering there might be other syntax scenarios that we haven’t considered yet for the new AlignHash features.

bquorning: It’s probably related to `AlignHash` now being able to use a mix of e.g. `TableAlignment` and `KeyAlignment`, leading to – as you mention – `TableAlignment#deltas_for_first_pair` _not_ being called before accessing `max_key_width`, and thus doing calculations with `nil`…

I agree that defaulting to `0` would be a good approach here.

rmm5t: Okay, I'll add an initializer to default the value to zero. Stand by.

rmm5t: Suggested changes pushed.

bbatsov:
Thanks!

PR #7069, opened 2019-05-21
[Speed up] Skip calling `ResultCache.cleanup` when files size is 1

hoshinotsuyoshi:
## Summary

Skip calling `ResultCache.cleanup` when the inspecting file is only one. 
For performance 🚀. 

* `ResultCache.cleanup` takes time.
  * If there are 8000 cache files, it takes about 200ms.
* It is meaningful to save as much as 200ms when integrating text editors.
* When rubocop called by text editor, this `files.size` will be always 1.

If this PR is merged, we will lose the opportunity for cleaning when the cache file is increased by 1, but I believe it is not important. :)

### 3. Create 8000+ cache objects

<details><summary>How?</summary>

```bash
# First, ensure there is no caches.
$ rm -rf ~/.cache/rubocop_cache 

# Then, run rubocop 8 times with different options.
$ rubocop -P --only=Metrics/AbcSize
$ rubocop -P --only=Metrics/BlockLength
$ rubocop -P --only=Metrics/BlockNesting
$ rubocop -P --only=Metrics/ClassLength
$ rubocop -P --only=Metrics/CyclomaticComplexity
$ rubocop -P --only=Metrics/LineLength
$ rubocop -P --only=Metrics/MethodLength

# Now you get 8000+ cache files.
$ tree ~/.cache/rubocop_cache | tail -1
9 directories, 8268 files
```

</details>

### 4. Do `time rubocop -a -d Gemfile` 11times(1 rehearsal + 10 attempts)  at top directory of rubocop repository.

result:

| time(1)   | average  | min | max |
|-----|----:|-----:|----:|
| rubocop 0.70.0 (959f215)  | 1.4613 | 1.398 | 1.529 |
| This PR(4ec7a3ba) | 1.2363 | 1.199 | 1.309 | 

∴ **220ms faster !!** 🚀 

tested by:  `ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin17]`

## Is it normal to have 8000+ cache files?

I think, yes. Default limit is 20000.

See also:

* https://github.com/rubocop-hq/rubocop/blob/v0.70.0/lib/rubocop/result_cache.rb#L13-L18
* https://github.com/rubocop-hq/rubocop/blob/v0.70.0/config/default.yml#L103-L107

## Why `ResultCache.cleanup` is slow?

I think it is because ["File.file?"](https://github.com/rubocop-hq/rubocop/blob/v0.70.0/lib/rubocop/result_cache.rb#L25) is called many times.

It is a robust implementation, it works well without knowing how deep the directory hierarchy is in advance. 

If you prefer a naive implementation, there may be room for even faster.

-----------------

* [x]  Squashed related commits together.
* [ ] Added tests.

### DIFF
--- a/lib/rubocop/runner.rb
+++ b/lib/rubocop/runner.rb
@@ -72,7 +72,12 @@ module RuboCop
 
       each_inspected_file(files) { |file| inspected_files << file }
     ensure
-      ResultCache.cleanup(@config_store, @options[:debug]) if cached_run?
+      # OPTIMIZE: Calling `ResultCache.cleanup` takes time. This optimization
+      # mainly targets editors that integrates RuboCop. When RuboCop is run
+      # by an editor, it should be inspecting only one file.
+      if files.size > 1 && cached_run?
+        ResultCache.cleanup(@config_store, @options[:debug])
+      end
       formatter_set.finished(inspected_files.freeze)
       formatter_set.close_output_files
     end


lib/rubocop/runner.rb
```diff
@@ -72,7 +72,13 @@ def inspect_files(files)
 
       each_inspected_file(files) { |file| inspected_files << file }
     ensure
-      ResultCache.cleanup(@config_store, @options[:debug]) if cached_run?
+      # OPTIMIZE: Calling ResultCache.cleanup takes time, so we avoid it when
+      # the inspecting files size is small.
```
connorshea: This working is a bit better

```suggestion
      # inspecting only one file.
```

hoshinotsuyoshi: Thank you!!

hoshinotsuyoshi: Squashed.  Co-Authored-By: @connorshea . https://github.com/rubocop-hq/rubocop/pull/7069/commits/46680ddddaf18175b845ca07d87fe38009b8aacf

lib/rubocop/runner.rb
```diff
@@ -72,7 +72,13 @@ def inspect_files(files)
 
       each_inspected_file(files) { |file| inspected_files << file }
     ensure
-      ResultCache.cleanup(@config_store, @options[:debug]) if cached_run?
+      # OPTIMIZE: Calling ResultCache.cleanup takes time, so we avoid it when
+      # inspecting only one file.
+      worth_to_call_cleanup = files.size > 1
```
bbatsov: Do we even need this variable? I'd just write `if/else` instead.

lib/rubocop/runner.rb
```diff
@@ -72,7 +72,13 @@ def inspect_files(files)
 
       each_inspected_file(files) { |file| inspected_files << file }
     ensure
-      ResultCache.cleanup(@config_store, @options[:debug]) if cached_run?
+      # OPTIMIZE: Calling ResultCache.cleanup takes time, so we avoid it when
```
bbatsov: I'd expand this description to mention that the optimization targets editors forking out to RuboCop.

bbatsov:
Apart from my small remarks I'm fine with the proposed change.

hoshinotsuyoshi:
Thank you for your review! How about this? -- 0388566a4cf74005c0cc35a9cd0bafe1608e0394

bbatsov:
Looks good to me! Thanks!

hoshinotsuyoshi:
Thanks!

PR #7072, opened 2019-05-22
[Fix #6936] Fix MultilineMethodArgumentLineBreaks when bracket hash assignment on multiple lines

### LINKED ISSUE #6936: Layout/MultilineMethodArgumentLineBreaks False Positive On Multiline Assignment
This is a pretty major bug.  The `Layout/MultilineMethodArgumentLineBreaks` cop is throwing errors when an assignment spans multiple lines.  I'm assuming because it sees the assignment operator as a "method" (which it technically is but should not be treated like a method call for the purposes of this cop).

Original PR: #6826 

/cc @maxh 

--------

## Expected behavior

The following code should not throw an error.

```
class Thing
  def call
    bar['foo'] = ::Time.zone.at(
                   huh['foo'],
                 )
  end
end
```

## Actual behavior

It throws an error.

See code snippet above.

## RuboCop version

```
0.67.2 (using Parser 2.6.2.1, running on ruby 2.6.2 x86_64-darwin17)
```

maxh:
Fixes https://github.com/rubocop-hq/rubocop/issues/6936 by handling this case:

```rb
a['b'] = c(
  1,
)
```

The `[]` assignment is represented in the AST by a send node with `b` as an argument and the expression on the right as an argument. It's ok that these are on the same line.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * [#5976](https://github.com/rubocop-hq/rubocop/issues/5976): Warn for Rails Cops. ([@koic][])
 
+### Bug fixes
+
+* [#6936](https://github.com/rubocop-hq/rubocop/issues/6936): Fix `Layout/MultilineMethodArgumentLineBreaks` when bracket hash assignment on multiple lines. ([@maxh][])
+
 ## 0.70.0 (2019-05-21)
 
 ### New features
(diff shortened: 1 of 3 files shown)


mikegee:
I support merging this as is, since it definitely fixes a bug. 👍 

I'm also concerned this PR allows for similar bug. Is the following code allowed incorrectly?
```ruby
a['b',
  'c', 'd'] = e
```

maxh:
>I'm also concerned this PR allows for similar bug. Is the following code allowed incorrectly?

This PR fixes that bug as well. Added a spec to illustrate/cover that case.

bbatsov:
Thanks!

Issue #6726, opened 2019-02-04
One or more (`.+`) of X in NodePattern


baweaver:
## Is your feature request related to a problem? Please describe.

With some methods, the signature allows for multiple variations of calling it:

```ruby
> ruby-parse -e 'meth(1, 2, 3)'
(send nil :meth
  (int 1)
  (int 2)
  (int 3))

> ruby-parse -e 'meth(1, 2)'
(send nil :meth
  (int 1)
  (int 2))
```

It would be very useful to have a way to specify "one or more" child nodes that match. Doubly so in the case of splat / double-splat type method signatures.

## Describe the solution you'd like

In the case above, a potential syntax would be:

```ruby
# Idea 1
(send nil? :meth $+(int _))

# Idea 2
(send nil? :meth +$(int _))

# Idea 3
(send nil? :meth $(int _)+)
```

## Describe alternatives you've considered

There _are_ current ways to do this that involve capturing the parent node and parsing out the children after the match. The problem with this is that it doesn't allow granularity in matching in a single pass, such as I expected the method to be called with one or more ints, or maybe with `n` ints.

marcandre:
One tricky thing to consider are captures inside the `+`'ed term, e.g. `(send nil? :meth +(int $_))` to collect `[1, 2]` or `[1, 2, 3]`, especially if there are other capture, or even worse for nested `+(...)`.

One solution would be that those would still be a "single" capture but would be captured in an array. `(send nil? $_ +(int $_))` would capture `[:meth, [1, 2]]` for example. Nested `+()` would have to be arrays of arrays, etc...

marcandre:
This should be relatively easy to do now that `<...>` has been implemented, with the same limitations (only one of `...`, `< ...>` or `*/+` per sequence)

PR #7081, opened 2019-05-24
NodePattern: Use stricter criteria for children count

marcandre:
This improves the checking for some node patterns (variadic with positive minimum number of terms).

E.g.: `(array <int ...>)` produced code with condition `node0.children.size >= 0` instead of `node0.children.size >= 1`. This had no actual consequence though expect being less efficient, because the code to check for the orderless children would still work properly.

This will come in handy for future work (e.g. #6726)


bbatsov:
Thanks!

PR #7086, opened 2019-05-26
[Fix #7063] Fix unsafe autocorrect for 'Style/TernaryParentheses'

### LINKED ISSUE #7063: Style/TernaryParentheses Not safe when run with "rubocop -a --safe"
image

Following is new code after refactor

```rb
@bookings.includes(:comments, :booking_fees).each do |b|
            data << [
              b.reservation_site.present? ? "#{b.source_i18n}(#{b.reservation_site})" : b.source_i18n,
              b.airbnb? or b.airbnb2? ? b.confirm_number : b.uid,
              b.id,
              b.dtstart.to_date,
              b.dtend.to_date,
              b.house.hotel? ? "#{b.house.name}(#{b.room_unit&.room_no})" : b.house,
              b.house.tag_list.to_s,
              b.house.hotel? ? b.room_unit&.room_no : '',
              b.user,
              b.user.try(:phone),
              b.user.try(:email),
              b.user.try(:settings).try(:[], 'location'),
              b.status_i18n,
              b.checkin_status_i18n,
              b.payment_status_i18n,
              b.guestnum,
              (b.dtend.to_date - b.dtstart.to_date).to_i,
              b.confirmed_at.try(:to_date),
              b.currency,
              b.total_fee,
              b.channel_fee_amount,
              b.total_host_paid_out_before_support_fee,
              b.settings[:points].to_f,
              b.ota_collect? ? b.total_fee.to_f - b.settings[:points].to_f : 0,
              b.hotel_collect_credit? or b.hotel_collect? ? b.paid_amount.to_f : 0,
              b.hotel_collect_cash? ? b.paid_amount.to_f : 0,
              b.paid? ? 0 : b.total_fee.to_f - b.channel_fee_amount.to_f,
              b.cleaning_fee_amount,
              b.house.user ? b.total_host_paid_out_before_support_fee.to_i * (b.house.user.commission_rate || 18) / 100 : '',
              b.comments.map {|c| "#{c.user&.username}: #{c.comment}"}.join("\n")
            ]
          end
```

## RuboCop version

```
 ╰─ $ rubocop -V
0.69.0 (using Parser 2.6.3.0, running on ruby 2.6.2 x86_64-linux)
```

parkerfinch:
This takes the precedence of operators used into account when deciding
to autocorrect. The issue is that if an operator of precedence below
that of the ternary operator (e.g. "or") is used, then the existing
autocorrect can change the semantics by removing parentheses. With
this change, the autocorrect will no longer attempt to make this
correction.

I'm thinking that this is a step in the right direction, but it would probably be better to even avoid reporting that this is an issue. Also, this problem will only manifest if the `Style/AndOr` or `Style/Not` cops aren't being followed, would it make sense to only include this check if those are disabled?

I wasn't sure how to handle the other [operators with low precedence](http://phrogz.net/programmingruby/language.html#table_18.4]), it seems like some might be handled already (e.g. I _think_ `defined?` is handled in the `method_name` node matcher?), and I wasn't sure if others even needed to be handled.

All feedback/ideas very welcome, I'm still not sure I totally grok this Cop but this is a stab at a fix!

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#7063](https://github.com/rubocop-hq/rubocop/issues/7063): Fix autocorrect in `Style/TernaryParentheses` cop. ([@parkerfinch][])
+
 ### Changes
 
 * [#5976](https://github.com/rubocop-hq/rubocop/issues/5976): Remove Rails cops. ([@koic][])
--- a/lib/rubocop/cop/style/ternary_parentheses.rb
+++ b/lib/rubocop/cop/style/ternary_parentheses.rb
@@ -149,12 +149,22 @@ module RuboCop
 
         def unsafe_autocorrect?(condition)
           condition.children.any? do |child|
-            unparenthesized_method_call?(child)
+            unparenthesized_method_call?(child) ||
+              below_ternary_precedence?(child)
           end
         end
 
         def unparenthesized_method_call?(child)
           method_name(child) =~ /^[a-z]/i && !child.parenthesized?
+        end
+
+        def below_ternary_precedence?(child)
+          # Handle English "or", e.g. 'foo or bar ? a : b'
+          (child.or_type? && child.semantic_operator?) ||
+            # Handle English "and", e.g. 'foo and bar ? a : b'
+            (child.and_type? && child.semantic_operator?) ||
+            # Handle English "not", e.g. 'not foo ? a : b'
+            (child.send_type? && child.prefix_not?)
         end
 
         def_node_matcher :method_name, <<-PATTERN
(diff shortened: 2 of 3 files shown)


bbatsov:
> I'm thinking that this is a step in the right direction, but it would probably be better to even avoid reporting that this is an issue. 

Perhaps. Although this might be confusing to some users as it won't be obvious why some cases where reported and others weren't.

> Also, this problem will only manifest if the Style/AndOr or Style/Not cops aren't being followed, would it make sense to only include this check if those are disabled?

I think it's fine just to assume that anything is possible (what your PR currently does).

> I wasn't sure how to handle the other operators with low precedence, it seems like some might be handled already (e.g. I think defined? is handled in the method_name node matcher?), and I wasn't sure if others even needed to be handled.

I was thinking it'd be nice if we had something generic to handle/check precedence. //cc @Drenmi

parkerfinch:
@bbatsov thanks for your thoughts there! Think this thing is good to merge without generic precedence handling? (Or are there any other things you think I should change here?)

bbatsov:
For me the PR is ok in its current form. Just move the changelog entry, as already cut a new release since you filed the PR.

parkerfinch:
@bbatsov moved the changelog entry!
Thanks for taking a look at this, and all the work that goes into maintaining rubocop! Very excited for 1.0.

bbatsov:
You're welcome! Thanks for your contribution! 🙇

PR #7092, opened 2019-05-29
NodePattern: Add repeated patterns

### LINKED ISSUE #6726: One or more (`.+`) of X in NodePattern

marcandre:
This closes #6726 by introducing tokens for repeating patterns: `*` (0 or more), `+` (1 or more) and `?` (0 or 1).

Subject to the same features and limitations as other variadic tokens `...` and `< ...>`, i.e.:
* can be preceded/succeeded by any number of no terms
* can be nested
* but only one variadic token per sequence

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 * [#7084](https://github.com/rubocop-hq/rubocop/pull/7084): Permit to specify TargetRubyVersion 2.7. ([@koic][])
+* [#7092](https://github.com/rubocop-hq/rubocop/pull/7092): Node patterns can now use `*`, `+` and `?` for repetitions.  ([@marcandre][])
 
 ### Bug fixes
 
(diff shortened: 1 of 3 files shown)


bbatsov:
I think you used something that doesn't exist in 2.3.

marcandre:
> I think you used something that doesn't exist in 2.3.

Oops, right. Fixed.

Also added a line in the README.

bbatsov:
Thanks! 🙇 Might be a good idea at some point to document the recent improvements you've made here as well http://www.rubocop.org/en/stable/node_pattern/

marcandre:
Oh, right, will do!

Issue #5976, opened 2018-06-09
Extract Rails Cops in a separate


bbatsov:
Rails cops have to extracted into a standalone gem (named something like `rubocop-rails` or whatever).

The reason for this decision is simple. I want RuboCop to be focused just on Ruby and have no framework/library/runtime dependant pieces for the sake of being easier to focus on evolving it. Creating more extensions should also help us come up with a good official API for them.

Any volunteers to do this? I don't do much Rails anymore, so I'd rather not be involved in this project if possible. @koic I know you've been working a lot on the Rails cops, so perhaps you'd like to tackle the extraction?

Drenmi:
Unfortunately someone hogged the `rubocop-rails` name on RubyGems. I think the gem hosts the configuration used by the Rails team.

bbatsov:
Yeah, I noticed this, but someone on RubyKaigi mentioned the guy would be willing to give the name up. I'm fine with naming the extension `rubocop-ror` or whatever.

composerinteralia:
I would be happy to do this, or to help @koic do it. I wrote the couple of custom cops in rails and had a lot of fun doing it. (@koic I'm guessing we take those custom cops out of rails once the next rubocop version gets released. I would be happy to make that PR and to update rubocop-rails
as well unless you were planning to do it.)

Drenmi:
@koic Maybe you can reach out to toshimaru about the name? His e-mail is registered in RubyGems.

koic:
@bbatsov Yes. I will work on this issue.  I'm thinking about planning now.

The following are issues and solutions for realizing this plan. There are still some issues to consider 🤔 
Please let me know if you have any opinion.

## RuboCop Rails Gem's name

As discussed above in this Issue, rubocop-rails gem already exists.
https://rubygems.org/gems/rubocop-rails

@Drenmi I'm asking @toshimaru whether he can transfer the gem name of rubocop-rails to the RuboCop team.

Even if toshimaru accepts the transfer gem name ownership, it seems necessary to announce to the current rubocop-rails gems (toshimaru/rubocop-rails) users, so we will need to talk to toshimaru about it.

@toshimaru Sorry to ask this of you, would you transfer rubocop-rails gem name to the RuboCop team, is it possible? I think that for the Rails department managed by RuboCop HQ, the name RuboCop Rails (rubocop-rails) is user-friendly as a convention.

## Transition from RuboCop Core to RuboCop Rails

[...]

PR #7093, opened 2019-05-30
[#5976] Warn for `rubocop -R/--rails` option

koic:
This PR warns for `rubocop -R/--rails` option.

Here is an example using `-R/--rails` option:

```console
% cat app/models/user.rb
class User < ActiveRecord::Base
end
```

```console
% rubocop -R
`-R/--rails` option and Rails cops will be removed from RuboCop 0.72. Use the `rubocop-rails` gem instead.
https://github.com/rubocop-hq/rubocop/blob/master/manual/migrate_rails_cops.md
Inspecting 2 files
.C

Offenses:

(snip)

app/models/user.rb:1:14: C: Rails/ApplicationRecord: Models should
subclass ApplicationRecord.
class User < ActiveRecord::Base
             ^^^^^^^^^^^^^^^^^^
```

This PR provides a way to dismiss the above warning. Specify `require: rubocop-rails` in .rubocop.yml (or command option) instead of the `-R/-rails` option as indicated in the above URL of warning.

And this means that it will be used the same way as RuboCop RSpec and RuboCop Performance.

This is a migration path to remove Rails cops from the RuboCop core, Rails cops will still be disabled by default for compatibility.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 ### Changes
 
 * [#5976](https://github.com/rubocop-hq/rubocop/issues/5976): Warn for Rails Cops. ([@koic][])
+* [#5976](https://github.com/rubocop-hq/rubocop/issues/5976): Warn for `rubocop -R/--rails` option. ([@koic][])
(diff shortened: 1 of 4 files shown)


bbatsov:
Thanks!

PR #7111, opened 2019-06-05
[Fix #7108] Change autocorrection of StderrPuts to Kernel.warn

### LINKED ISSUE #7108: Style/StderrPuts: dangerous autocorrection
## Expected behavior

Style/StderrPuts should autocorrect `STDERR.puts "foo"` into `Kernel.warn "foo"` or not at all. The current behavior would call "warn", even if it is not Kernel.warn, but something else completely in the current context.  

## Actual behavior

Style/StderrPuts autocorrects `STDERR.puts "foo"` into `warn "foo"`.

rrosenblum:
This fixes #7108.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Bug fixes
 
 * [#7106](https://github.com/rubocop-hq/rubocop/issues/7106): Fix an error for `Lint/NumberConversion` when `#to_i` called on a variable on a hash. ([@koic][])
+* [#7108](https://github.com/rubocop-hq/rubocop/issues/7108): Modify the autocorrection of `Style/StderrPuts` to use `Kernel.warn` instead of `warn`. ([@rrosenblum][])
 
 ## 0.71.0 (2019-05-30)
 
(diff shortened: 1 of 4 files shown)


koic:
This cop is based on the Ruby Style Guide.
https://github.com/rubocop-hq/ruby-style-guide#warn

The Ruby Style Guide may have to be updated first. I think this change is appropriate if accepted there.

bbatsov:
Yeah, I agree with @koic. We have to keep in mind it's extremely uncommon to see `Kernel` methods be called with the module prefix, mostly because it's generally a bad practice to define methods that hide them (we might even have some cop for that or at least we planned to add one).

rrosenblum:
That makes sense. This seemed like an easy enough fix so I didn't fully think of the implication of adding `Kernel`. What are your thoughts on this being a configuration option?

bjeanes:
#7810 proposes detecting overrides of `Kernel` methods. However, I'd argue that certain `Kernel` methods are regularly overridden (e.g. `puts` and `print` are in fact convenience methods to stdout, but are a common part of the API for any `IO`-like object).

Rather, if we were able to to detect an override of `Kernel`, then this cop should check if `warn` is overridden in scope and _then_ replace with `Kernel.warn` (and otherwise continue replacing with `warn`).

As it stands right now, this cop is unsafe and can introduce stack overflows. Test case:

```rb
class Foo
  def warn(message)
    $stderr.puts "\e[33m#{message}\e[0m"
  end
end
```

PR #7113, opened 2019-06-06
Rename enforced style rails for indentation consistency

koic:
Rename `EnforcedStyle: rails` for `Layout/IndentationConsistency`

Follow up https://github.com/rubocop-hq/rubocop/pull/7059#issuecomment-493861821.

This PR renames `EnforcedStyle: rails` to `EnabledStyle: indented`  for `Layout/IndentationConsistency`.  This makes it a generic name that doesn't depend on Rails.

And this PR shows alternative for obsolete EnforcesStyle parameter name.

For example, `EnforcedStyle: rails` in config/default.yml will be replaced by `EnforcedStyle: indented`.

```diff
 # .rubocop.yml
 Layout/IndentationConsistency:
-  EnforcedStyle: rails
+  EnforcedStyle: indented
```

### Before

After `EnforcedStyle: rails` has been obsoleted, Users do not know what the parameter name to replace.

```consle
% rubocop
Error: invalid EnforcedStyle 'rails' for Layout/IndentationConsistency found in .rubocop.yml
Valid choices are: normal, indented
```

### After

Users will find that they should use `EnforcedStyle: indented` instead.

```console
% rubocop
Error: obsolete `EnforcedStyle: rails` (for Layout/IndentationConsistency)found in .rubocop.yml
`EnforcedStyle: rails` has been renamed to `EnforcedStyle: indented`
```

-----------------

* [x] Added tests.


bbatsov:
I wonder if `outdented_access_modifiers` won't be a better name for this style, as `indented` doesn't mean much for something like indentation consistency. Any thoughts @rubocop-hq/rubocop-core?

koic:
Thanks for proposing a better option name. I'd like to update it to `outdented_access_modifiers` if there is no other opinion.

bbatsov:
@koic That's fine by me.

koic:
Alternative enforced style name updated from `indented` to `outdented_access_modifiers`.

kamipo:
Probably `EnforcedStyle: rails` was intended to be used in rails/rails, but we (rails/rails) basically do not want to be outdented access modifiers.

```ruby

# not better

class Foo
  def bar
  end

private
  def baz
  end
end

# good

class Foo
  def bar
  end

  private
    def baz
    end
end
```

The naming `outdented_access_modifiers` cause misreading that rails/rails want to the above `# not better` style rather than `# good` style, it is a little unhappy to me.

bbatsov:
> Probably EnforcedStyle: rails was intended to be used in rails/rails, but we (rails/rails) basically do not want to be outdented access modifiers.

Yeah, that was the case. 

> The naming outdented_access_modifiers cause misreading that rails/rails want to the above # not better style rather than # good style, it is a little unhappy to me.

I'm puzzled. Did the style in Rails change at some point? If so I guess the name of the style and its description should be changed.

kamipo:
That is the DHH flavored style, we have not changed any style.

See an example for the style (DHH's PR): https://github.com/rails/rails/pull/35489

bbatsov:
Hmmm, I wonder why I was under the impression that Rails favours:

```
class Foo
  def bar
  end

private
  def baz
  end
end
```

That's actually why I suggested the outdented naming to @koic. I'll have to check this more closely.

bbatsov:
My bad then. Well, I'm out of a good idea for the name then. :D How does @dhh refer to that style?

koic:
It seems that my commentary on the PR description was not enough.

I named it `EnforcedStyle: indented` based on the following Rails coding conventions:

> Indent and no blank line after private/protected.

https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#follow-the-coding-conventions

This assumes the following code indicated by @kamipo.

```ruby
# good
class Foo
  def bar
  end

  private
    def baz
    end
end
```

AFAIK, the following code is generated at Rails 4.0 using `bundle exec rails g scaffold`. This is an application code generated by Rails users (In fact, this coding style has been around since ancient times).

```console
% bundle exec rails -v
Rails 4.0.0
% bundle exec rails g scaffold article author:string
```

```ruby
% cat app/controllers/articles_controller.rb
class ArticlesController < ApplicationController
  before_action :set_article, only: [:show, :edit, :update, :destroy]

  # GET /articles
  # GET /articles.json
  def index
    @articles = Article.all
  end

  (snip)

  # DELETE /articles/1
  # DELETE /articles/1.json
  def destroy
    @article.destroy
    respond_to do |format|
      format.html { redirect_to articles_url }
      format.json { head :no_content }
    end
  end

  private
    # Use callbacks to share common setup or constraints between actions.
    def set_article
      @article = Article.find(params[:id])
    end

    # Never trust parameters from the scary internet, only allow the white list through.
    def article_params
      params.require(:article).permit(:author)
    end
end
```

bbatsov:
@koic I wonder if we shouldn't revert the PR as frankly I can't come up with a better name than Rails. The problem with `indented` is that it's not clear what we are referring to. Maybe `indented_private_methods` or something like this? // @rubocop-hq/rubocop-core Help, please! :-)

We can't cut a new release before fixing this.

koic:
@bbatsov Thanks for the name suggestion. I think `indented_internal_methods` is a name that represents this style.
@kamipo I think you are a Rails committer familiar with this style. What do you think about this name? I appreciate your expert knowledge in this matter.

bbatsov:
> @bbatsov Thanks for the name suggestion. I think indented_internal_methods is a name that represents this style.

@koic I'm fine with the proposed name.

koic:
Yeah, I opened a PR https://github.com/rubocop-hq/rubocop/pull/7163 and we may be good to wait for a while to get a better name feedback.

kamipo:
👍 Thank you guys for thinking of about the naming!

PR #7116, opened 2019-06-07
Add ArrayOfRange cop

apz-stripe:
Ruby syntax is such that `[1..10]` is interpreted as an array with a single range, `(1..10)`, in it. This can lead to hard to detect bugs as in the case `[1..10].each {|e| e.do_a_thing}`. This cop would catch this case as an early warning for this class of bug.

This cop seems to catch issues in rubocop itself. I have not written an autocorrector as the "correct" change to fix unintentional bugs would likely be to change `[1..10] -> (1..10)`. However this would be a semantic change and is likely dangerous if there are intentional uses of arrays of ranges. Is there a reason for omitting the parens?

-----------------

* [x] Added tests.


koic:
The inspect representation for `[(1..10)]` is `[1..10]` and the print representation is `[1..10]`.

```console
% ruby -e 'p [(1..10)]'
[1..10]
% ruby -e 'puts [(1..10)]'
1..10
```

Therefore `[(1..10)]` looks like redundant code to me.

apz-stripe:
Thanks for the response. 

If I am understanding correctly the point you are making is that as the parens do nothing semantically (a range is just `<expression>..<expression>`), they are redundant. In this case do you mean that as ruby prints this without parens, they are unnecessary?

koic:
Yes. I also think that redundant styles tend not to be preferred.

mikegee:
@koic you make a good point. But, I wonder if we can allow `[(1..10)]` in the rare case a developer wants an array with a range as the only element?

This would be similar to the extra parens in `if (user = User.current_user); user.do_thing; end` used to mark the assignment as intended.

koic:
I don't think there is a need to warn of `[(1..10)]` redundancy. However, the notation `[1..10]` is widely used in the forms `[n..-1]` etc. such as ruby/ruby repo, rails/rails repo, bundler/bundler repo and others.  On the other hand, I don't see the use case of `[(n..- 1)]`. I think that it makes no sense to warn of this.

mikegee:
> the notation `[1..10]` is widely used in the forms `[n..-1]` etc.

I'm probably confused (typical), but I'm surprised those projects have code with array literals containing a single range. To be clear, we're not talking about giving a range argument the the `[]` method like this:

```ruby
instance_of_Array[1..10]
```

koic:
Ah this is my quote mistake. I'm sorry to confuse you 💦 Anyway it would not be preferable to be forced to write redundantly.

apz-stripe:
> To be clear, we're not talking about giving a range argument the the `[]` method like this:
> 
> ```ruby
> instance_of_Array[1..10]
> ```

This is exactly the case. I don't think it makes sense to enforce this in array access. However when using ranges and arrays in conjunction it can be not immediately clear what a give expression `[1..10]` actually is. It would be reasonable to unknowingly expect the behaviour to be equivalent to `[*1..10]` when using it like this: `[1..10].each {|e| e.do_a_thing}`. The proposed change enforces that parens are used for an array of a single range to make code more clear.

However, if this not something rubocop should worry about as it would cause redundancy that makes sense.

PR #7121, opened 2019-06-10
Allow safe-navigation in ternary conditions without parenthesis

timon:
When enforced style is `require_parenthesis_if_complex`, the following
code is considered valid:

```ruby
  foo = bar.baz? ? a : b
```
But safe navigation operator produces an offence.
This commit makes the following code also acceptable:

```ruby
  foo = bar&.baz ? a : b
```
-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/ternary_parentheses_spec.rb
+++ b/spec/rubocop/cop/style/ternary_parentheses_spec.rb
@@ -266,6 +266,10 @@ RSpec.describe RuboCop::Cop::Style::TernaryParentheses, :config do
       it_behaves_like 'code with offense',
                       'foo = bar or (baz) ? a : b',
                       'foo = bar or baz ? a : b'
+
+      it_behaves_like 'code with offense',
+                      'foo = (bar&.baz) ? a : b',
+                      'foo = bar&.baz ? a : b'
     end
 
     context 'with a complex condition' do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug fixes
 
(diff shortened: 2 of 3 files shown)


koic:
It is difficult to define simple condition and complex condition. It may be better for offense to occur because `foo&.bar` is a conditional expression similar to `foo && bar`.
But I think this PR is valid because `foo[:bar]` has no offense. It is also shown as simple condition in the test cases as a symbolic syntax.

spec/rubocop/cop/style/ternary_parentheses_spec.rb
```diff
@@ -266,6 +266,9 @@
       it_behaves_like 'code with offense',
                       'foo = bar or (baz) ? a : b',
                       'foo = bar or baz ? a : b'
+
+      it_behaves_like 'code without offense',
+                      'foo = bar&.baz ? a : b'
```
koic: It may be consistent with other test cases if a test case is written as follows.

```diff
diff --git a/spec/rubocop/cop/style/ternary_parentheses_spec.rb b/spec/rubocop/cop/style/ternary_parentheses_spec.rb
index f0438b723..9ef7e6254 100644
--- a/spec/rubocop/cop/style/ternary_parentheses_spec.rb
+++ b/spec/rubocop/cop/style/ternary_parentheses_spec.rb
@@ -267,7 +267,8 @@ RSpec.describe RuboCop::Cop::Style::TernaryParentheses, :config do
                       'foo = bar or (baz) ? a : b',
                       'foo = bar or baz ? a : b'

-      it_behaves_like 'code without offense',
+      it_behaves_like 'code with offense',
+                      'foo = (bar&.baz) ? a : b',
                       'foo = bar&.baz ? a : b'
     end
```

timon: Hi @koic, I have force-pushed the suggested change

bbatsov:
Thanks!

Issue #4193, opened 2017-03-28
Why prefer_alias is the enforced style in Style/Alias


Ana06:
In the [Ruby Style guide](https://github.com/bbatsov/ruby-style-guide#alias-method) it is said:

> Always use alias_method when aliasing methods of modules, classes, or singleton classes at runtime, as the lexical scope of alias leads to unpredictability in these cases.

Then, why `prefer_alias` is the enforced style in Style/Alias? I think the enforced style should be `prefer_alias_method`

bbatsov:
Because there it also says https://github.com/bbatsov/ruby-style-guide#alias-method-lexically :-)

Drenmi:
This comes up often enough (I mentioned it here #3619) that it deserves its own FAQ entry. 😅

Issue #3619, opened 2016-10-14
Default configuration of `Style/Alias` doesn't match the style guide recommendation


Drenmi:
Our default configuration has the following:

```yaml
# config/default.yml
Style/Alias:
  EnforcedStyle: prefer_alias
```

but the [style guide](https://github.com/bbatsov/ruby-style-guide#alias-method) recommends:

> Always use alias_method when aliasing methods of modules, classes, or singleton classes at runtime, as the lexical scope of alias leads to unpredictability in these cases.

Not sure how many people we might surprise by changing it now, though. 😅

bbatsov:
I would disagree. See the other `alias` rule in the styleguide - it's never really clear if someone is going to extend a class or not, so I think that using `alias` by default makes a lot of sense. There were a ton of lengthy discussions about this here and in the style guide's issue tracker.

Drenmi:
My bad. Only searched the document for `alias_`, and missed the previous part. Closing. 😀

PR #7129, opened 2019-06-12
Point to "Alias Method Lexically"

vfonic:
* for clearer understanding why `alias` is default over `alias_method`

### Background

After getting the rubocop `Style/Alias` error, I clicked on the link suggested in the error message:
> Style/Alias: Use alias instead of alias_method in a module body. (https://github.com/rubocop-hq/ruby-style-guide#alias-method)

The link only confused me more, as it suggests using `alias_method` (but the cop reports that `alias` should be used instead).

After the little trip down the rabbit hole, I found couple of issues of other developers asking the same question: 
[Why prefer_alias is the enforced style in Style/Alias #4193
](https://github.com/rubocop-hq/rubocop/issues/4193)
[Default configuration of `Style/Alias` doesn't match the style guide recommendation #3619
 ](https://github.com/rubocop-hq/rubocop/issues/3619)

So, I think it's better to point to the anchor above `#alias-method`, that instead suggests that `alias` should be used (as is the cop's default).

NOTE: I didn't add an entry to CHANGELOG. Let me know if I should (it IS a user-observable change)

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2126,7 +2126,7 @@ Style/AccessModifierDeclarations:
 
 Style/Alias:
   Description: 'Use alias instead of alias_method.'
-  StyleGuide: '#alias-method'
+  StyleGuide: '#alias-method-lexically'
   Enabled: true
   VersionAdded: '0.9'
   VersionChanged: '0.36'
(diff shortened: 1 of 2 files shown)


vfonic:
I just rebased this on top of `master` and rerun `bundle exec rake default` to get the `manual/cop_style.md` to point to rubystyle.guide website

vfonic:
Bump! Any thoughts on this?

I think this could remove some of the confusion.

bbatsov:
I agree. Thanks!

PR #7134, opened 2019-06-12
[Fix #7130] Skip autocorrect in `Style/FormatString` if second argument is variable

### LINKED ISSUE #7130: `Style/FormatString` cop breaks code
Rubocop auto-corrector breaks code by `Style/FormatString` cop.

Before code is

```
def hoge
  "%08x-%04x-%04x-%04x-%04x%08x" % ary
end
```

Execute auto-correct

```
> bundle exec rubocop -a
```

After code is

```
def hoge
  format("%08x-%04x-%04x-%04x-%04x%08x", ary)
end
```

## Expected behavior

No effort.

## Actual behavior

Fix to unexpected code.

## RuboCop version

```
> bundle exec rubocop -v
0.71.0
```

tejasbubane:
For `String#%` is second argument is variable, flag offense but skip autocorrect. It is not possible for rubocop to know if variable is array in which case the autocorrect leads to invalid code.

Closes #7130.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/format_string.rb
+++ b/lib/rubocop/cop/style/format_string.rb
@@ -48,6 +48,10 @@ module RuboCop
         }
         PATTERN
 
+        def_node_matcher :variable_argument?, <<-PATTERN
+          (send {str dstr} :% {send_type? lvar_type?})
+        PATTERN
+
         def on_send(node)
           formatter(node) do |selector|
             detected_style = selector == :% ? :percent : selector
@@ -70,10 +74,10 @@ module RuboCop
         end
 
         def autocorrect(node)
-          lambda do |corrector|
-            detected_method = node.method_name
+          return if variable_argument?(node)
 
-            case detected_method
+          lambda do |corrector|
+            case node.method_name
             when :%
               autocorrect_from_percent(corrector, node)
             when :format, :sprintf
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/format_string.rb
```diff
@@ -70,10 +74,10 @@ def method_name(style_name)
         end
 
         def autocorrect(node)
-          lambda do |corrector|
-            detected_method = node.method_name
+          return if variable_argument?(node)
 
-            case detected_method
+          lambda do |corrector|
+            case node.method_name
```
tejasbubane: Inlined variable to remove `MethodLength` offense for this method.

lib/rubocop/cop/style/format_string.rb
```diff
@@ -48,6 +48,10 @@ class FormatString < Cop
         }
         PATTERN
 
+        def_node_matcher :variable_argument?, <<-PATTERN
+          (send {str dstr} :% send_type?)
```
pocke: `send_type?` is not enough to avoid the problem. Because it does not cover a variable.
For example

```ruby
a = something()
puts "%f" % a
```


So I think we should use `literal?` method instead of `send_type?`.

tejasbubane: Thanks for this case. But node here is `lvar` and `literal?` does not work. I have added this test case and modified code accordingly.

bbatsov:

tejasbubane:
@bbatsov rebased.

bbatsov:
Thanks!

tejasbubane:
@bbatsov Did you accidentally close this?

bbatsov:
No, I merged it manually as something was messed up in the GitHub's UI.

PR #7140, opened 2019-06-14
Fix #7137 Add new cop: RUBY_VERSION in gemspec

### LINKED ISSUE #7137: New cop: RUBY_VERSION in gemspec

Gemspec should not contain RUBY_VERSION as a condition to switch dependencies.
For example:

```ruby
Gem::Specification.new do |s|
  if RUBY_VERSION >= '2.5'
    s.add_runtime_dependency 'gem_a'
  else
    s.add_runtime_dependency 'gem_b'
  end
end
```

It has unexpected behavior. The `RUBY_VERSION` is determined by `rake release`.
If I use Ruby 2.5 to execute `rake release`, the gem always has `gem_a` as a dependency even if a user installs the gem with Ruby 2.4 or older.

We have several ways to fix the problem.

* Remove the dependencies and use post_install_message to notice the gems.
* Add both gems to the dependency.
* If the dependencies are for development, move them to Gemfile.

Detect reference of `RUBY_VERSION` in `*.gemspec` and add an offense to it.
For example: `def_node_matcher :ruby_version?, '(const nil :RUBY_VERSION)'`

The solution is loose. It checks `RUBY_VERSION` in anywhere of gemspec.
We can implement stricter detection. For example, detect `RUBY_VERSION` only if it is placed a conditon.

But I think the looser version is better. Because I guess the cop will have many false negative if we choose the stricter version. And I guess the looser version does not have too many false positives.

## Additional context

Today I found using RUBY_VERSION in a gemspec.
https://github.com/rafaelsales/ulid/pull/17
So I think the cop is meaningful.

malyshkosergey:

* [x] Added tests.


config/default.yml
```diff
@@ -198,6 +198,11 @@ Gemspec/RequiredRubyVersion:
 
 #################### Layout ###########################
 
+Gemspec/RubyVersionGlobalsUsage:
+  Description: Checks usage of RUBY_VERSION in gemspec
+  Enabled: true
+  VersionAdded: '0.72'
```
pocke:
```suggestion
  VersionAdded: '0.72'
  Include:
    - '**/*.gemspec'
```

Can you add `Include` to execute the cop only on gemspec files?

config/default.yml
```diff
@@ -198,6 +198,11 @@ Gemspec/RequiredRubyVersion:
 
 #################### Layout ###########################
 
+Gemspec/RubyVersionGlobalsUsage:
```
pocke: Could you move the configuration to under the Gemspec section?

lib/rubocop/cop/gemspec/ruby_version_globals_usage.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Gemspec
+      # Checks that `RUBY_VERSION` constant is not used in gemspec
```
pocke: I think we need to explain why we cannot use RUBY_VERSION in gemspec.
Can you write the explanation?

config/default.yml
```diff
@@ -198,6 +198,11 @@ Gemspec/RequiredRubyVersion:
...
+  Description: Checks usage of RUBY_VERSION in gemspec
```
koic: Please add a period to EOL.

lib/rubocop/cop/gemspec/ruby_version_globals_usage.rb
```diff
@@ -0,0 +1,52 @@
...
+      #
+      # @example
+      #   # When `RUBY_VERSION` is used in gemspec
```
koic: I think that this comment can be deleted because it is redundant with the cop explanation.

malyshkosergey:
@pocke @koic thanks for valuable feedback. I've fixed your comments. Please review. Thanks!

lib/rubocop/cop/gemspec/ruby_version_globals_usage.rb
```diff
@@ -0,0 +1,55 @@
...
+      #     end
+      #   end
+      #
+      #   # good
+      #   Gem::Specification.new do |spec|
+      #     spec.add_runtime_dependency 'gem_a'
+      #   end
+      #
+      class RubyVersionGlobalsUsage < Cop
+        MSG = '`RUBY_VERSION` used in gemspec file.'
```
koic: The following may be a bit clearer as a warning.

```diff
- MSG = '`RUBY_VERSION` used in gemspec file.'
+ MSG = 'Do not use `RUBY_VERSION` in gemspec file.`
```

malyshkosergey: Yep, sounds better. Fixed.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 ### New features
+* [#7137](https://github.com/rubocop-hq/rubocop/issues/7137): Add new `Gemspec/RubyVersionGlobalsUsage` cop. ([@malyshkosergey][])
```
koic: I think that automatic merging doesn't work well because there is `### New features` in the master branch. Could you rebase on the latest master branch and squash your commits into one?

bbatsov:
I handled the conflict manually. Thanks!

Issue #7133, opened 2019-06-12
Would it make sense to mention/list rubocop extensions


vfonic:
Maybe as one of the sections in the Readme.md?

## Describe the solution you'd like

If you're new to rubocop, it's hard to know and find existing rubocop extensions. You'd have to go to `rubocop-hq` and check which repos are there, etc. We could have the most popular or all extensions right in the Readme.md to avoid confusion and people not finding extensions, especially the ones like rubocop-rails or rubocop-rspec.

What do you think?

pocke:
You can find popular custom cops from here https://docs.rubocop.org/en/stable/extensions/#custom-cops

Personally, I think README does not need to mention the custom cops because it already has a link to the official manual.
But, I think we should improve the README and official manual if people are confused by them. For example, more eye-catching link to the manual, add useful ToC  to the top page of the manual, etc.

bbatsov:
Yeah, I agree. Having everything important in the manual is what matters the most IMO. 

> But, I think we should improve the README and official manual if people are confused by them. For example, more eye-catching link to the manual, add useful ToC to the top page of the manual, etc.

I totally agree here as well.

vfonic:
// I wrote a lengthy reply and then removed it 😇 

tl;dr
Perhaps "Official manual" could be just changed to "Documentation"?

I just opened couple of repos that I could remember from the top of my head and here's the terminology they use (this is random sample):
https://github.com/facebook/create-react-app/blob/master/README.md#user-guide (words used: "User Guide" and "documentation)
https://github.com/bumps/bumps#bumps-data-fitting-and-uncertainty-estimation (words used: "documentation")
https://github.com/avajs/ava#documentation (words used: "documentation")
https://github.com/thoughtbot/factory_bot#documentation (words used: "documentation")
https://github.com/norman/friendly_id#friendlyid (words used: "documentation")

I also really like how FriendlyId does it. It's the first sentence in the README:
https://github.com/norman/friendly_id

PS what about these two extensions:
* https://github.com/rubocop-hq/rubocop-md
* https://github.com/rubocop-hq/rubocop-sequel
They don't seem to be in the list of extensions in the documentation.

bbatsov:
Fair point! I guess I meant to write official User Manual or something like this back in the day, but mistyped or whatever. :D I'm fine with rebranding to "User Manual", "documentation" or whatever else people prefer. 

> They don't seem to be in the list of extensions in the documentation.

An oversight we should address.

PR #7142, opened 2019-06-14
Add rubocop-sequel and rubocop-md extensions

vfonic:
Related to our discussion here [Would it make sense to mention/list rubocop extensions #7133](https://github.com/rubocop-hq/rubocop/issues/7133)

I also added a note about adding any missing extensions.

Not sure if it's necessary. I can remove it. Thoughts?

-----------------

* [ ] Added tests.

### DIFF
--- a/manual/extensions.md
+++ b/manual/extensions.md
@@ -34,14 +34,20 @@ See [development](development.md).
 
 * [rubocop-performance](https://github.com/rubocop-hq/rubocop-performance) -
   Performance optimization analysis
-* [rubocop-rails](https://github.com/rubocop-hq/rubocop-rails) - Rails-specific analysis
+* [rubocop-rails](https://github.com/rubocop-hq/rubocop-rails) -
+  Rails-specific analysis
 * [rubocop-rspec](https://github.com/rubocop-hq/rubocop-rspec) -
   RSpec-specific analysis
 * [rubocop-thread_safety](https://github.com/covermymeds/rubocop-thread_safety) -
   Thread-safety analysis
 * [rubocop-require_tools](https://github.com/milch/rubocop-require_tools) -
   Dynamic analysis for missing require statements
-* [rubocop-i18n](https://github.com/puppetlabs/rubocop-i18n) - i18n wrapper function analysis (gettext and rails-i18n)
+* [rubocop-i18n](https://github.com/puppetlabs/rubocop-i18n) -
+  i18n wrapper function analysis (gettext and rails-i18n)
+* [rubocop-sequel](https://github.com/rubocop-hq/rubocop-sequel) -
+  Code style checking for Sequel gem
+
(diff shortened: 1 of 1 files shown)


manual/extensions.md
```diff
@@ -34,14 +34,22 @@ See [development](development.md).
...
 * [rubocop-thread_safety](https://github.com/covermymeds/rubocop-thread_safety) -
   Thread-safety analysis
 * [rubocop-require_tools](https://github.com/milch/rubocop-require_tools) -
   Dynamic analysis for missing require statements
-* [rubocop-i18n](https://github.com/puppetlabs/rubocop-i18n) - i18n wrapper function analysis (gettext and rails-i18n)
+* [rubocop-i18n](https://github.com/puppetlabs/rubocop-i18n) -
+  i18n wrapper function analysis (gettext and rails-i18n)
+* [rubocop-sequel](https://github.com/rubocop-hq/rubocop-sequel) -
+  Code style checking for Sequel gem
+* [rubocop-md](https://github.com/rubocop-hq/rubocop-md) -
```
pocke: rubocop-md is not a custom cop plugin.
So we should add a new section for the plugin, like `Other plugins`, and place the plugin to the section.

vfonic: 🤔 Indeed. It adds rubocop functionality to code snippets in md files.

So this would be...? An extension?

Unfortunately we're using the word "extension" to represent all the plugins/add-ons/extensions. The file itself is called `extensions.md`. I think consistent terminology is important. What do you suggest is the best way going forward?

Are you still with `Other plugins`?

I'm just thinking whether we should just keep the plugin like it's now, with the other plugins.

bbatsov: Probably other extensions we should have `Additional Cops` and `Other` or something along those lines.

bbatsov: Looking at the page something like `### Other Extensions` is what we need. There's also `haml-lint` that's based on RuboCop if I remember correctly.

bbatsov: On the other hand - technically those are not plugins/extensions, but related tools. Maybe they don't belong on this page at all. I can't decide on this right now.

vfonic: 🤔 I also don't know what would be the best place to put rubocop-md (and haml-lint).

In the meantime, I've updated this PR to only add rubocop-sequel.

I think we can get that one in and maybe rubocop-md after we figure out the best place to put it.

bbatsov: Fair enough. Can you remove the unrelated line-wrapping changes that are also part of the commit currently?

vfonic: I added these intentionally, to keep consistency with the rest of the listed extensions. I think we should keep those.

Some of the listed extensions had those lines wrapped, some didn't. Now they all have. Let me know what you think?

PR #7147, opened 2019-06-16
[Fix #7144] Extend the namespaces definition for Style/Documentation cop

### LINKED ISSUE #7144: Style/Documentation accept constant visibility declaration as namespace
When declare the visibility of a constant (class and module include), the namespace object get an offense to document it.

```
module Namespace
  class Private
  end

  private_constant :Private
end
```

--------

## Expected behavior

Don't get an offense on the Namespace module. by considering constant visibility declaration as constant declaration.

## Actual behavior

Get an offense on the Namespace module.

* Create a file with the code above
* Run Rubocop on it

## RuboCop version

```
0.71.0 (using Parser 2.6.3.0, running on ruby 2.4.5 x86_64-darwin18)
```

AdrienSldy:
Extend the definition of namespaces by adding the constants visibility
statement.
No longer report an error for classes and modules that contain constants'
visibility statements.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/documentation.rb
```diff
@@ -3,10 +3,11 @@
 module RuboCop
   module Cop
     module Style
-      # This cop checks for missing top-level documentation of
-      # classes and modules. Classes with no body are exempt from the
-      # check and so are namespace modules - modules that have nothing in
-      # their bodies except classes, other modules, or constant definitions.
+      # This cop checks for missing top-level documentation of classes and
```
bbatsov: I wonder if we shouldn't add here examples of the code that's exempt, so it's clearer for users what the cop acts on and what it ignores.

bbatsov:
You're suggestions makes sense to me, but I think we should extend the documentation a bit to make it clearer what exactly are we ignoring.

lib/rubocop/cop/style/documentation.rb
```diff
@@ -31,6 +32,9 @@ class Documentation < Cop
 
         def_node_matcher :constant_definition?, '{class module casgn}'
         def_node_search :outer_module, '(const (const nil? _) _)'
+        def_node_matcher :constant_visibility_declaration?, <<-PATTERN
```
koic: Can you rebase with latest master branch and use squiggly heredoc?

```diff
- def_node_matcher :constant_visibility_declaration?, <<-PATTERN
+ def_node_matcher :constant_visibility_declaration?, <<~PATTERN
```

AdrienSldy: @koic done :)

lib/rubocop/cop/style/documentation.rb
AdrienSldy: Ok, I think you're right. I will do it soon.

AdrienSldy: @bbatsov Sorry for the time spent. What do you think of comments? That's enough for you as an explanation?

lib/rubocop/cop/style/documentation.rb
```diff
@@ -18,19 +19,52 @@ module Style
...
+      #   module Math
+      #   end
+      #
       #   # good
       #   # Description/Explanation of Person class
       #   class Person
       #     # ...
       #   end
       #
+      #   # allowed
```
bbatsov: allowed -> good

lib/rubocop/cop/style/documentation.rb
```diff
@@ -18,19 +19,52 @@ module Style
...
+      #     # Class without body
```
bbatsov: You should indent the new examples just as much as the existing ones.

bbatsov:

PR #7153, opened 2019-06-18
Add new cop `Style/FloatDivision`

tejasbubane:
StyleGuide: https://rubystyle.guide/#float-division

-----------------

* [x] Added tests.


lib/rubocop/cop/style/float_division.rb
```diff
@@ -0,0 +1,54 @@
...
+module RuboCop
+  module Cop
+    module Style
+      # This cop checks for division with integers coerced to floats.
+      # It is recommended to either always use `fdiv` or coerce left side only.
+      #
+      # @example EnforcedStyle: left_coerce (default)
+      #   # bad
+      #   a / b.to_f
+      #   a.to_f / b.to_f
```
koic: I think that there is a redundant case that `to_f` is used on both sides. On the other hand, I'm not sure if it is really good to force `to_f` to left side only.

IMO, perhaps the default style of warning only when `to_f` is used on both sides may be convenient. This is a third style.

config/default.yml
```diff
@@ -2643,6 +2643,15 @@ Style/ExpandPathArguments:
   Enabled: true
   VersionAdded: '0.53'
 
+Style/FloatDivision:
+  Description: 'Prefer either use fdiv or convert the left-side coersion for float division.'
```
koic: Please add `StyleGuide: #float-division`.

config/default.yml
```diff
@@ -2643,6 +2643,15 @@ Style/ExpandPathArguments:
...
+  Enabled: true
+  VersionAdded: '0.72'
+  EnforcedStyle: left_coerce
+  SupportedStyles:
+    - left_coerce
+    - fdiv
```
koic: I think it would be good to add `Reference: https://github.com/rubocop-hq/ruby-style-guide/issues/628` to the reference. It will help to understand how the rules were added.

tejasbubane: Fixed.

config/default.yml
tejasbubane: Added

lib/rubocop/cop/style/float_division.rb
tejasbubane: Won't this conflict with what is recommended by styleguide?

```
# bad
a / b.to_f
a.to_f / b.to_f

# good
a.fdiv(b)
a.to_f / b
```

koic: Yeah, It conflicts with the current situation. So I'd like to hear other reviewers while exploring the possibility of editing Ruby Style Guide. This is a style that has recently been merged into the style guide and is now verifiable in working code as you implement it ❤️ 

Cc @rubocop-hq/rubocop-core

bbatsov: I guess it's fine to have `right_coerce` style. We should the left version in the guide as supposedly it makes it easier to spot what you're trying to do, but obviously that's subjective.

tejasbubane:
@bbatsov @koic Added third config option `right_coerce` and kept `left_coerce` as default (as per style guide).

lib/rubocop/cop/style/float_division.rb
koic: I did not have enough explanation. I wanted to show the following option to make a bad case only when both have `to_f`.

```ruby
# bad
a.to_f / b.to_f

# good
a.to_f / b
a / b.to_f
a.fdiv(b)
```

Whether `to_f` is used on the left or right depends on the nature of `a` or `b` parameter. On the other hand, it is redundant that `to_f` is used in both. At that time, I think it is better to let users choose whether to remove the left side `to_f` or the right side `to_f`. This is the reason I recommend this option to default.

lib/rubocop/cop/style/float_division.rb
```diff
@@ -0,0 +1,72 @@
...
+        def on_send(node)
+          add_offense(node) if offense_condition?(node)
+        end
+
+        private
+
+        def offense_condition?(node)
+          style == :left_coerce && right_coerce?(node) ||
+            style == :right_coerce && left_coerce?(node) ||
+            style == :fdiv && coerce?(node)
```
bquorning: I would prefer this method to look more similar to `#message`, to make it obvious that their logic is similar e.g.
```rb
def offense_condition?(node)
  case style
  when :left_coerce
    right_coerce?(node)
  when :right_coerce
    left_coerce?(node)
  when :fdiv
    right_coerce?(node) || left_coerce?(node)
  else
    false
  end
end
```

bbatsov: I'm surprised we don't have a cop for that already. :-)

lib/rubocop/cop/style/float_division.rb
tejasbubane: @koic I have added another option `single_coerce` for this case, but did not make it default because it will be inconsistent with style guide.

lib/rubocop/cop/style/float_division.rb
tejasbubane: Fixed.

bbatsov:
After some deliberation I agree with @koic that the best default would be something that just warns if you use `to_f` on both sides. Let's also update the style guide accordingly.

bbatsov:
See also https://blog.rubystyle.guide/ruby/2019/06/21/float-division.html - unfortunately we got no feedback on it.

koic:
Thanks for your consideration. I opened a PR https://github.com/rubocop-hq/ruby-style-guide/pull/767 to rubocop-hq/ruby-style-guide repo.

lib/rubocop/cop/style/float_division.rb
koic: @tejasbubane The Ruby Style Guide has been updated.
https://rubystyle.guide/#float-division
https://github.com/rubocop-hq/ruby-style-guide/pull/767

Could you update this PR?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New features
 
 * [#7150](https://github.com/rubocop-hq/rubocop/pull/7150): Add `AllowIfModifier` option to `Style/IfInsideElse` cop. ([@koic][])
+* Add new cop `Style/FloatDivision` that checks coercion. ([@tejasbubane][])
```
koic: Please add this PR link.

tejasbubane:
@koic PR updated. Changed default style to `single_coerce`.

config/default.yml
```diff
@@ -2643,6 +2643,19 @@ Style/ExpandPathArguments:
   Enabled: true
   VersionAdded: '0.53'
 
+Style/FloatDivision:
+  Description: 'Prefer either use fdiv or convert the left-side coersion for float division.'
```
koic: Can you update this description?

tejasbubane: I missed this. Fixed now.

PR #7156, opened 2019-06-19
[Fix #6793] Clarify ruby version error

### LINKED ISSUE #6793: Unclear messaging on Ruby version check
When running an older version of rubocop against a newly released Ruby version, this error is output:
> Error: Unknown Ruby version 2.6 found in `.ruby-version`.

That's fine, but the issue is that when it appears in the normal flow of a linting run it's not clear that rubocop is what's complaining, and it often looks more like e.g. rbenv complaining that the Ruby version isn't installed on the system.

image

Would it be possible to rephrase this in such a way that it's clear that it's rubocop (or a linter in general) that doesn't know about the stated Ruby version so the remedial steps (upgrading rubocop) are clearer to the reader?

I didn't want to open a PR on this one since I didn't know how best you'd want to indicate this, but I'm happy to open one after some feedback on wording.

tejasbubane:
Showing it came from `RuboCop`.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/config.rb
+++ b/lib/rubocop/config.rb
@@ -629,13 +629,13 @@ module RuboCop
       return if KNOWN_RUBIES.include?(target_ruby_version)
 
       msg = if OBSOLETE_RUBIES.include?(target_ruby_version)
-              "Unsupported Ruby version #{target_ruby_version} found in " \
-              "#{target_ruby_source}. #{target_ruby_version}-compatible " \
+              "RuboCop found unsupported Ruby version #{target_ruby_version} " \
+              "in #{target_ruby_source}. #{target_ruby_version}-compatible " \
               'analysis was dropped after version ' \
               "#{OBSOLETE_RUBIES[target_ruby_version]}."
             else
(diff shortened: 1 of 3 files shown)


bbatsov:
I think part of the problem is still not solved - namely how to solve the error if you encounter it.

tejasbubane:
> I think part of the problem is still not solved - namely how to solve the error if you encounter it.

For newer ruby version error, we can ask users to upgrade rubocop since by the time ruby releases, we will most probably already support it.

Is there any better way for handling old obsolete version case apart from upgrading ruby itself?

bbatsov:
> For newer ruby version error, we can ask users to upgrade rubocop since by the time ruby releases, we will most probably already support it.

Indeed.

> Is there any better way for handling old obsolete version case apart from upgrading ruby itself?

They don't need to upgrade Ruby, they need to update the target ruby version. Parser has many parsing modes going all the way back to Ruby 1.9, but we typically opt to support only the currently supported Ruby releases.

PR #7122, opened 2019-06-10
[Fix #6893] Handle implicit rescue correctly in Naming/RescuedExceptionsVariableName

pocke:
Fix #6893
Fix #6907
Fix #6912
Related #6908 Thanks @anthony-robin ! I've overlooked the pull request when I was implementing this pull request.

Problem
===

This cop has many issues. The cause is wrong handling for implicit rescue.

This cop handle `rescue e` as implicit rescue. It means the cop treats the code as an assignment to `e`. But it is not an assignment. It catches `e` class's error.
The cop should handle `rescue => e` as an implicit rescue.

Note
===

This pull request will remove `@exception_name` instance variable.
An instance of a cop is allocated each file. So the instance variable will be overridden if other `rescue` clauses exist.
So we need to avoid using the instance variable.

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Added an entry to the [Changelog](https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format).
* [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.
* [x] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: https://chris.beams.io/posts/git-commit/

PR #7169, opened 2019-06-24
Fix misdetection in Naming/RescuedExceptionsVariableName

knu:
The RescuedExceptionsVariableName cop wrongly finds an offense and tries to correct a local variable name in the exception list of a rescue clause, which could never be a proper operation.

  ```ruby
  # `exception` is defined here.

  begin

  rescue exception
       # ^--- This should never be corrected.
  end
  ```

-----------------

* [x] Added tests.


pocke:
Probably it is duplicated with #7122

knu:
Looks like it.  Hope the fix will soon be delivered.

PR #7173, opened 2019-06-25
Use squiggly heredoc instead of `String#strip_indent`

pocke:
RuboCop supports Ruby 2.3 or higher, so we can use squiggly heredoc (`<<~`) :heart:

Basically, I used the following commands.

```bash
$ cd spec/
$ git grep -lE '<<-\w+\.strip_indent' | xargs sed -i -E 's/<<-(\w+)\.strip_indent/<<~\1/g'
$ git grep -lE '<<~\w+\.strip_indent' | xargs sed -i -E 's/<<~(\w+)\.strip_indent/<<~\1/g'

# Do not replace `<<-'RUBY'` to avoid JRuby bug
# $ git grep -lE "<<-'\w+'\.strip_indent" | xargs sed -i -E "s/<<-('\w+')\.strip_indent/<<~\1/g"
```

Note
---

RuboCop has `String#strip_indent` even if this pull request is applied.
I think we can replace the other `strip_indent` also, but this pull request is good for the first step of replacing `strip_indent`.
I guess we need to replace the other `strip_indent` by hand.

---

ref https://github.com/rubocop-hq/rubocop/pull/7122#discussion_r297025089

-----------------

* [ ] Added tests.

### DIFF
--- a/spec/rubocop/cli_spec.rb
+++ b/spec/rubocop/cli_spec.rb
@@ -399,21 +399,21 @@ RSpec.describe RuboCop::CLI, :isolated_environment do
       end
 
       include_examples 'UnneededCopDisableDirective not run',
-                       'individually disabled', <<-YAML.strip_indent
-        Lint/UnneededCopDisableDirective:
-          Enabled: false
-      YAML
+                       'individually disabled', <<~YAML
+                         Lint/UnneededCopDisableDirective:
+                           Enabled: false
+                       YAML
       include_examples 'UnneededCopDisableDirective not run',
-                       'individually excluded', <<-YAML.strip_indent
-        Lint/UnneededCopDisableDirective:
-          Exclude:
-            - example.rb
-      YAML
+                       'individually excluded', <<~YAML
+                         Lint/UnneededCopDisableDirective:
+                           Exclude:
+                             - example.rb
+                       YAML
       include_examples 'UnneededCopDisableDirective not run',
-                       'disabled through department', <<-YAML.strip_indent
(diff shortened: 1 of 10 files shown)


spec/rubocop/cop/layout/tab_spec.rb
```diff
@@ -22,9 +22,9 @@
   end
 
   it 'registers an offense for a line indented with mixed whitespace' do
-    expect_offense(<<-'RUBY'.strip_indent)
-       	x = 0
-       ^ Tab detected.
+    expect_offense(<<~'RUBY')
+      	 x = 0
```
pocke: I changed the test code. It was `[space][tab]x = 0`, but it is meaningless. Because the `[space]` is removed by `strip_indent`.
So it is changed to `[tab][space]x = 0` to avoid removing the space.

koic:
FYI, the reason why JRuby's CI matrix is failing may be as follows.
- https://github.com/rubocop-hq/rubocop/pull/7026#issuecomment-490667498
- https://github.com/jruby/jruby/issues/4260

koic:
After that, I got the answer that the above issue is planned to be solved in JRuby 9.2.8.0.
https://github.com/jruby/jruby/issues/4260#issuecomment-505635017

pocke:
Thank you @koic! I'll fix the problem tonight.

pocke:
I reverted the replacing `<<~'RUBY'` to avoid the error on JRuby.

bbatsov:
Thanks!

PR #7181, opened 2019-06-27
Sort analyzed file alphabetically

pocke:
The order of files that RuboCop analyzes is not fixed. Because RuboCop uses `Dir.glob` and the order of the `Dir.glob` is not fixed.

> Note that the pattern is not a regexp, it’s closer to a shell glob. See File.fnmatch for the meaning of the flags parameter. Case sensitivity depends on your system (File::FNM_CASEFOLD is ignored), as does the order in which the results are returned.
> http://ruby-doc.org/core-2.6.3/Dir.html#method-c-glob

It is not a bug, but it confused me while debugging the regression test. https://github.com/pocke/rubocop-regression-test/issues/12

I think sorted files is better than randomized. It will be easier to debug because everyone gets the same output by RuboCop.

-----------------

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changes
 
 * Allow `#**` Doxygen comment style on `Layout/LeadingCommentSpace`. ([@anthony-robin][])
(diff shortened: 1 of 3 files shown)


bbatsov:
I also like the suggested change.

pocke:
Thank you for reviewing! I'll merge it after resolving the conflict.

PR #7183, opened 2019-06-27
Allow `#**` Doxygen comment style on `Layout/LeadingCommentSpace`

anthony-robin:
This PR remove the offense for comments that starts with `**` as this style is
a valid syntax to write block comments in Ruby.

### Motivation
The `apidocjs` package only specify `=begin/=end` block to define
the documentation in ruby. However, [this comment][1] explain that the `Doxygen`
style `#**` is also compatible to write the documentation so the `Layout/LeadingCommentSpace`
should not flag any offense for it.

My [first attempt][2] was to not flag `Style/BlockComments` if the `@api` annotation
was found inside the `=begin/=end` block but this solution is more elegant.

#### Before (not clean solution)
```ruby
=begin
@api {get} /endpoint description of my endpoint
...
=end
```

#### After (elegant solution)
```ruby
#**
# @api {get} /endpoint description of my endpoint
# ...
#*
```

[1]: https://github.com/apidoc/apidoc/issues/620#issuecomment-494206190
[2]: https://github.com/rubocop-hq/rubocop/pull/6522

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* Allow `#**` Doxygen comment style on `Layout/LeadingCommentSpace`. ([@anthony-robin][])
+
 ## 0.72.0 (2019-06-25)
 
 ### New features
--- a/config/default.yml
+++ b/config/default.yml
@@ -816,6 +816,7 @@ Layout/LeadingCommentSpace:
   StyleGuide: '#hash-space'
   Enabled: true
   VersionAdded: '0.49'
+  VersionChanged: '0.73'
 
 Layout/MultilineArrayBraceLayout:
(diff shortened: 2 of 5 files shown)


koic:
Yep. I think it makes sense because it is the same as a solution to RDoc. Thanks!
db7914604665235df10c32e20785a14e3d2d51be

koic:
Oh, I missed this comment 💦 
https://github.com/rubocop-hq/rubocop/pull/6522#issuecomment-472981102

It may be better to add `AllowDoxygenCommentStyle: (true | false)` option If this cop does not want to support Doxygen style by default. This option is disabled (`false`) by default.

anthony-robin:
Oh I see, let me submit another PR to refactor it using a configurable option 👍

PR #7184, opened 2019-06-27
[WIP] [Fix #5979] Introduce cops with special status

### LINKED ISSUE #5979: Introduce new cops with a special status (instead of enabled/disabled)
We have to add a way to mark recently added cops as new and have the users decide themselves whether they want to enable them or not. 

I think that we can have `Enabled` feature a new value (e.g. `New` or `nil`) and cops marked this way will just produce a message on RuboCop start saying something like:

```
The following cops were added and you need to decided whether to enable or disable them.
```

This is going to solve the commonly reported problem that RuboCop updates are very painful for end users as usually they break their builds due to new cops being added. This would allow people go just bugfixes without having to mess much with the config. Going forward we'll take a page from ESLint and enable/disable cops upstream only in major releases. 

This ticket is somewhat related to #5978, as we'd use the version data in the message we print to the users.

Darhazer:
Consider this a proof of concept. 

The approach I took creates a couple of problems I did not foresee.

First, the need to disable the warning in 'cli spec behavior' context, as otherwise, we are polluting the output with each new cop added.

The other issue is that the newly introduced cops would not be applied to the rubocop codebase as well unless we introduce an option to override the new status (e.g. enable all non-configured cops)

Let me know what you think and I can proceed with adding rake tasks for enabling new cops and any other supporting options.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/registry.rb
+++ b/lib/rubocop/cop/registry.rb
@@ -125,10 +125,14 @@ module RuboCop
 
       def enabled?(cop, config, only_safe)
         cfg = config.for_cop(cop)
+
+        # cfg['Enabled'] might be a string `none`, which is considered disabled
+        cop_enabled = cfg.fetch('Enabled') == true
+
         if only_safe
-          cfg.fetch('Enabled') && cfg.fetch('Safe', true)
+          cop_enabled && cfg.fetch('Safe', true)
         else
-          cfg.fetch('Enabled')
+          cop_enabled
         end
       end
 
--- a/lib/rubocop/config.rb
+++ b/lib/rubocop/config.rb
@@ -14,6 +14,11 @@ module RuboCop
     include PathUtil
     include FileFinder
 
+    class << self
+      attr_accessor :test
+      alias testing? test
+    end
+
     COMMON_PARAMS = %w[Exclude Include Severity inherit_mode
                        AutoCorrect StyleGuide Details].freeze
     INTERNAL_PARAMS = %w[Description StyleGuide VersionAdded
@@ -262,6 +267,7 @@ module RuboCop
       end
       validate
       make_excludes_absolute
+      warn_on_new_cops unless self.class.testing?
       self
     end
 
@@ -507,6 +513,20 @@ module RuboCop
         warn Rainbow("Warning: unrecognized cop #{name} found in " \
                      "#{smart_loaded_path}").yellow
       end
+    end
+
+    def warn_on_new_cops
+      new_cops = keys.select do |key|
+        @hash[key]['Enabled'].to_s.casecmp('none').zero?
+      end
+
+      return if new_cops.none?
+
+      warn Rainbow('The following cops were added to RuboCop, but are not ' \
+      ' configured. Please set Enabled to either `true` or `false` in your ' \
+      '`.rubocop.yml` file.').yellow
+
+      warn Rainbow(new_cops.join("\n")).yellow
     end
 
     def validate_syntax_cop
--- a/lib/rubocop/rspec/shared_contexts.rb
+++ b/lib/rubocop/rspec/shared_contexts.rb
@@ -52,6 +52,7 @@ RSpec.shared_context 'config', :config do
       cop_name = described_class.cop_name
       hash[cop_name] = RuboCop::ConfigLoader
                        .default_configuration[cop_name]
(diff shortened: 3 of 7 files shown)


Darhazer:
Perhaps a better approach would be to have a `VersionLocked` in the user's `.rubocop.yml` and for cops introduced in later versions - to check if there is `Enabled` in the user's config, and if no - warn the user about the new cop.

This requires that the user have a VersionLocked - unless he or she does, all new cops would be effectively enabled as if the VersionLocked is set to the current version.

WDYT?

bbatsov:
> The other issue is that the newly introduced cops would not be applied to the rubocop codebase as well unless we introduce an option to override the new status (e.g. enable all non-configured cops)

I've always assumed we'd just enable those cops in RuboCop's own config file. That's not really an issue I think, as the default config and RuboCop's config are orthogonal concerns. 

> Perhaps a better approach would be to have a VersionLocked in the user's .rubocop.yml and for cops introduced in later versions - to check if there is Enabled in the user's config, and if no - warn the user about the new cop.

This was discussed early on, but I decided it's not a good option mostly because of what you said - you need to keep updating this version. This PR is one of the few essential bits before releasing RuboCop 1.0 and after we cut we won't be enabling any new cops by default until 2.0, which means that the VersionLocked will come for free. ESLint made this approach pretty popular and I think it's reasonable, that's why I'm looking for us to emulate it. 

We'll still have to refine the current behaviour and maybe add some options to suppress those messages, always enable new cops or always disable them. That's pretty simple, though.

Darhazer:
> I've always assumed we'd just enable those cops in RuboCop's own config file.
If that is the case, I'll remove the code that suppresses warnings during tests
Also, currently it warns during each config load (I just wanted to get the scope of the changes where there is a cop with a `none` status) while it should do it on the merged config. I'll update that as well

lib/rubocop/cop/registry.rb
```diff
@@ -125,10 +125,14 @@ def enabled(config, only, only_safe = false)
 
       def enabled?(cop, config, only_safe)
         cfg = config.for_cop(cop)
+
+        # cfg['Enabled'] might be a string `none`, which is considered disabled
```
bbatsov: I think `pending` or `new` might be a better name for this. Nil might be an option as well, although it's not very descriptive.

bbatsov:
Fine by me. We'll also need to add a few lines in the manual about this.

@rubocop-hq/rubocop-core That's a very important change, so I'd appreciate your input as well.

bbatsov:
Well, I'll assume that everyone's fine with my idea then. :D

pirj:
> The other issue is that the newly introduced cops would not be applied to the rubocop codebase as well unless we introduce an option to override the new status (e.g. enable all non-configured cops)

Do you mean when working on `rubocop` itself?
But there are `config/defaults.yml` where it's set to `none`/`pending`, and `.rubocop.yml`, where it can be configured. During internal investigations, the warnings will appear just like for any other repository.
Did I understand the problem correctly?

pirj:
Does this mean that all the cops will be added to `config/default.yml` with `none`/`pending`/`false` statuses only from now on?

pirj:
> First, the need to disable the warning in 'cli spec behavior' context

I'm sorry, I'm out of context with this. What does that mean exactly?

Darhazer:
> Do you mean when working on rubocop itself?

Yes. But the .rubocop.yml file solves the issue

> Does this mean that all the cops will be added to config/default.yml with none/pending/false statuses only from now on?

Yes. 

> I'm sorry, I'm out of context with this. What does that mean exactly?

It doesn't matter if RuboCop uses its own config file that enabled those cops as @bbatsov suggested

pirj:
I've moved this a bit forward:
 - moved the check to `config_loader.rb`, since otherwise if something was marked as `pending` in `default/config.yml`, the warning was emitted despite it was overridden in `.rubocop.yml`
- renamed the status from `new` to `pending`

TODO:
- when there's no project config file it should warn about pending cops (now it returns too early to warn)
- when DisabledByDefault is on - `pending` cops should be disabled, when EnabledByDefault is on - `pending` cops should be enabled (didn't check, maybe works out of the box
- write/fix specs

bbatsov:
@pirj Thanks for the update!

Darhazer:
Closing in favor of #7567

PR #7185, opened 2019-06-28
Add configurable option for Doxygen style to `Layout/LeadingCommentSpace`

anthony-robin:
Follows #7183 to take care of @koic's [comment][1].

Instead of allowing by default Doxygen comment syntax, we should use a configurable
option in the cop configuration to enable or not this behavior.

[1]: https://github.com/rubocop-hq/rubocop/pull/7183#issuecomment-506573743

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/leading_comment_space.rb
+++ b/lib/rubocop/cop/layout/leading_comment_space.rb
@@ -16,6 +16,25 @@ module RuboCop
       #
       #   # good
       #   # Some comment
+      #
+      # @example AllowDoxygenCommentStyle: false (default)
+      #
+      #   # bad
+      #
+      #   #**
+      #   # Some comment
+      #   # Another line of comment
+      #   #*
+      #
+      # @example AllowDoxygenCommentStyle: true
+      #
+      #   # good
+      #
+      #   #**
+      #   # Some comment
+      #   # Another line of comment
+      #   #*
+      #
       class LeadingCommentSpace < Cop
         include RangeHelp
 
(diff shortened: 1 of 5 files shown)


anthony-robin:
(Note: I didn't add entry to changelog as the previous PR was not already released)

lib/rubocop/cop/layout/leading_comment_space.rb
```diff
@@ -16,15 +16,25 @@ module Layout
...
       #   # Some comment
+      #
+      # @example AllowDoxygenCommentStyle: true (default false)
+      #
+      #   # good
+      #   #**
+      #   # Some comment
+      #   # Another line of comment
+      #   #*
+      #
```
koic: Could you separate the option example into `AllowDoxygenCommentStyle: false (default)` and `AllowDoxygenCommentStyle: true` as follows?

```ruby
      #
      # @example AllowDoxygenCommentStyle: false (default)
      #
      #   # bad
      #
      #   #**
      #   # Some comment
      #   # Another line of comment
      #   #*
      #
      # @example AllowDoxygenCommentStyle: true
      #
      #   # good
      #
      #   #**
      #   # Some comment
      #   # Another line of comment
      #   #*
      #
```

koic:
> (Note: I didn't add entry to changelog as the previous PR was not already released)

Could you please update the changelog history? This will probably change from `Changes` to` New features`, but I think it can be included in the next release because it is an option that does not cause breaking changes.

anthony-robin:
Thanks @koic ! I updated my code according to your review 👍

koic:
Thanks!

PR #7187, opened 2019-06-28
[WIP] Support auto_correct option with parallel

ota42y:
parallel option work these steps.
1. parallel load and save offenses to cache
2. main process collect offenses from cache

So collect result step doesn't support concurrency.

And we need real run with auto_correct option.
So parallel can't be combined with auto_correct.
https://github.com/rubocop-hq/rubocop/blob/64f6ca77af6aa14dc754f9cc5a8fbf32d4fa4f74/lib/rubocop/runner.rb#L136

I changed to support collecting errors with concurrency.
So we can use parallel option with auto_correct.

I took benchmark using rails/actioncable with default rubocop setting like this.

```bash
cd /rails
git checkout -f v5.2.3
git mv .rubocop.yml .rubocop.yml_bk

cd /rubocop
/usr/bin/time bundle exec ruby ./exe/rubocop /rails/actioncable --auto-correct --parallel
```

This table is result.
I tried execute 10 times.
`origin/master` branch mean `fffd4e6278ff47783b8055dcad00f01b67fff4ff` commit.

Test script is here (https://gist.github.com/ota42y/d5940ba9d1eb5db8789208230f1944c1)

| | user | system | total | real|
|---|---|---|---|--|
| with parallel | 0.004180 | 0.017864 | 163.198372 | 69.468754 |
| without parallel | 0.003942 | 0.016940 | 98.853728 | 101.491169 |
| origin/master | 0.003778 | 0.017028 | 97.648949 | 100.196688 |
-----------------

* [ ] Added tests.


jonas054:
@ota42y This is an interesting idea. It's nice to get output immediately in the non-autocorrect case, rather than having to wait for the cache to warm up, and the improvement in auto-correct run-time looks impressive.

But these improvements come at the price of increased complexity in the core parts of the code. The fact that we got two failures in CircleCI tells us that a bug is present somewhere. The test execution got stuck in the Runner and CLI parts. I've been able to reproduce this on my computer, but only one time. Seems to happen pretty rarely.

We should also ask ourselves if the improvements affect the most common use cases. If people use the `--auto-gen-config` feature and run `--auto-correct` only to fix newly introduced offenses, then most files will have no offenses and performance will not be improved by running in parallel. This is due to the `offenses.any?` condition in https://github.com/rubocop-hq/rubocop/blob/64f6ca77af6aa14dc754f9cc5a8fbf32d4fa4f74/lib/rubocop/runner.rb#L136

So I suggest that we don't continue with this, and I'm closing the PR. Comment if you disagree.

PR #7199, opened 2019-07-09
[Fix #5088] Fix bug in argument_in_method_call

### LINKED ISSUE #5088: Layout/MultilineMethodCallIndentation autofix lines not aligned
Steps to reproduce:
1. Create test.rb files with content
```
array.each do |cur_file|
  names << File.basename(cur_file, '.*')
             .delete('1')
             .delete('2')
             .delete('3')
end

```
2. `rubocop -a test.rb`
--------

## Expected behavior

```
array.each do |cur_file|
  names << File.basename(cur_file, '.*')
               .delete('1')
               .delete('2')
               .delete('3')
end

```

## Actual behavior

```
array.each do |cur_file|
  names << File.basename(cur_file, '.*')
           .delete('1')
           .delete('2')
               .delete('3')
end
```

Maybe expected behavior not right, but i expect all lines with `delete` to have same indent, but in result one line is differ from all others 

## RuboCop version

```
$ rubocop -V
0.51.0 (using Parser 2.4.0.2, running on ruby 2.4.2 x86_64-linux)
```

buehmann:
This fixes #5088.

When trying to determine if a method call to be aligned occurs inside the argument of another method call, the wrong `:send` node was being checked for parenthesized arguments. It
must the one whose arguments are being looked at.


koic:
This looks good. Can you squash your commits into one? It is preferable to leave a second commit message.

buehmann:
Sure, I wasn't sure how much squashing you preferred.

bbatsov:
Nicely done! Thanks!

PR #7200, opened 2019-07-09
Revert __RUBY__ to __END__ token in specs

buehmann:
I stumbled across these wrong occurrences of `__RUBY__` in the specs.

They were caused by an overzealous substitution in
8f37b7f2f0f4b01c2afa83fffdf531dfb4dce057, where heredoc delimiters such
as `END` were replaced – even inside `__END__`.


koic:
Good catch. Thanks!

Issue #6420, opened 2018-10-29
Improved syntax checking around line continuations


bquorning:
```rb
'this text is too ' \
'long'
```

`Style/LineEndConcatenation` recommends that we should use `\` instead of `+` or `<<` to concatenate strings across multiple lines. But I am still seeing wildly inconsistent formatting of these multiline strings. I would like to suggest RuboCop checking the following four aspects of formatting of multiline strings:

 1. One space between quote and `\`.
 2. Consistent indentation.
 3. Use same quotes on all lines.
 4. Keep whitespace at the end of lines.

### 1. One space between quote and `\`

```rb
# Bad
'this text is too '\
'long'

# Bad
'this text is too '  \
'long'

# Good
'this text is too ' \
'long'
```

### 2. Consistent indentation

Update: This is now handled by Layout/LineEndStringConcatenationIndentation.

Indent `IndentationWidth` or match the preceding line’s starting quote.

```rb
# IndentationWidth
puts 'this text is too ' \
  'long'

# Match quote
puts 'this text is too ' \
     'long'
```

### 3. Use same quotes on all lines

Update: This is now handled by Style/StringLiterals when setting `ConsistentQuotesInMultiline: true`.

```rb
# Bad
"this text is #{foo} " \
'long'

# Good
"this text is #{foo} " \
"long"

# Good
'this text is too ' \
'long'
```

### 4. Keep whitespace at the end of lines

In the actual text, disallow spaces at the beginning of line n+1 if line n ends with a non-space.

```rb
# Bad
'this text contains a lot of' \
'               spaces'

# Good
'this text contains a lot of               ' \
'spaces'

# Bad
'this text is too' \
' long'

# Good
'this text is too ' \
'long'
```

tejasbubane:
`3. Use same quotes on all lines.` This might create ambiguity `Style/StringLiterals` which suggests using single quotes for strings without interpolation.

```ruby
"this text is #{foo} " \
'long' # StringLiterals prefers this

"this text is #{foo} " \
"long" # But your suggesting prefers this
```

bquorning:
The idea was that this cop should take precedence over `Style/StringLiterals` on multi-line strings.

bquorning:
@bbatsov I would like to hear your opinion. Is strings continued over multiple lines within the responsibility of RuboCop, or should a separate tool (e.g. rubocop-strings) handle these niche cases?

bquorning:
[...]

PR #7202, opened 2019-07-10
[Fix #4719] Detect tabs between string literals

### LINKED ISSUE #4719: Layout/Tab ignores indentation for “string split over multiple lines”
## Expected behavior

In my test file, the last line is indented with two tabs. I expect Layout/Tab to detect these.

```
# frozen_string_literal: true

puts 'Lorem ' \
		 'Ipsum'
```

## Actual behavior

No offense is reported.

Run `rubocop` on the file above.

## RuboCop version

```
$ rubocop -V
0.49.1 (using Parser 2.4.0.0, running on ruby 2.4.1 x86_64-darwin15)
```

buehmann:
This fixes #4719 and simplifies `Layout/Tab` code.

The problem was that entire `:dstr` expressions were considered "inside a string", whereas they can be composed of multiple individual string literals:

```ruby
"foo" \
	"bar"
```

```
s(:dstr,
  s(:str, "foo"),
  s(:str, "bar"))
"foo" \
~~~~~~~... expression

s(:str, "foo")
"foo" \
    ~ end
~ begin
~~~~~ expression

s(:str, "bar")
        "bar"
            ~ end
        ~ begin
        ~~~~~ expression
```
In the AST this looks very similar to a multiline string, but the `begin` (and `end`) locations differ:

```ruby
"foo
	bar"
```

```
s(:dstr,
  s(:str, "foo\n"),
  s(:str, "\tbar"))
"foo
~ begin
~~~~... expression
        bar"
           ~ end

s(:str, "foo\n")
"foo
 ~~~... expression

s(:str, "\tbar")
        bar"
       ~~~~ expression
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [#7186](https://github.com/rubocop-hq/rubocop/issues/7186): Fix a false positive for `Style/MixinUsage` when using inside multiline block and `if` condition is after `include`. ([@koic][])
 * [#7099](https://github.com/rubocop-hq/rubocop/issues/7099): Fix an error of `Layout/RescueEnsureAlignment` on assigned blocks. ([@tatsuyafw][])
 * [#5088](https://github.com/rubocop-hq/rubocop/issues/5088): Fix an error of `Layout/MultilineMethodCallIndentation` on method chains inside an argument. ([@buehmann][])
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

bquorning:
@buehmann I would be interested in hearing your opinion on the somewhat related issue #6420 too.

Issue #1361, opened 2014-10-02
MethodLength/ClassLength cops take wrapped lines into account


jfelchner:
I was curious if there would be any way of having the Method/ClassLength cops not consider things like hashes that are split across multiple lines for readability.

Just wanted to open this up for discussion.  We like to keep our method length down pretty low (like around `15` which works great, but when it doesn't it's almost always a formatting deal like a hash or a few method parameters which have been split on multiple lines for readability.

Issue #1077, opened 2014-05-09
MethodLength should not consider Heredocs


fabiopelosin:
Any attempt of refactoring would end up in splitting the heredoc decreasing readability.

schmijos:
As of https://github.com/rubocop-hq/rubocop/pull/8159 you can configure heredocs to be counted as one line (`CountAsOne`).

Issue #494, opened 2013-09-18
"Method has too many lines" check not smart enough


jeidsath:
I really like the "Method has too many lines" check, and I think that it helps me write better code. Sometimes its warnings aren't helpful to me, and I would like to to improve the code it flags. But I'd appreciate some discussion before I submit a patch.

Should array initialization count as multiple lines for this check? Ie.,

```
[['FirstName', first_name],
 ['LastName', last_name],
 ['Email', email],
 ['OrgName', company],
 ['Company', company],
 ['Subdomain__c_lead', ai['subdomain']],
 ['TimeZone', ai['timezone']],
 ['pd_user_id__c', ai['uid']],
 ['pd_test__c', ai['test_mode']],
 ['pd_account_owner_user_id__c', ai['uid']],
 ['Created_At__c_lead', Time.now.strftime('%Y-%m-%d')],
 ['pd_test', test]]
```

Should operator separated lines count as multiple lines? Ie.,

```
get_lead_by_marketo_id(client) ||
  get_lead_by_marketo_cookie(client, account_info) ||
  get_lead_by_email(client, account_info) ||
  create_lead_by_email(account_info['email'])
```

simonc:
We're facing this issue with strong parameters.

@yltsrc your solution may work but it seems overkill to add a new class just to handle a >10LOC method which is perfectly readable, in the right context. I'm aware of the added-value of value-objects but this value is in DRYing the concept, not the code.

Having the choice between lines and statements would be a good thing IMHO ;)

jonas054:
Look at #1370. The `Metrics/AbcSize` cop was introduced in response to concerns like this. If you think that methods longer than 10 LOC aren't in themselves a problem, then you raise the `MethodLength:Max` configuration parameter a bit. If you think that 10 LOC is a good limit and just want to exempt this particular case, then you do it like this:
```ruby
  def order_params # rubocop:disable Metrics/MethodLength
    # ...
  end
```
The `order_params` method has a low score in the `AbsSize` cop.

alininja:
@jonas054 Thank you for that tip.  Is it possible to disable a rubocop check for a particular method in `.rubocop.yml` instead of within the source code?

So something like:
```
Metrics/MethodLength:
  Exclude:
    - 'Klass#method_with_large_hash'
```
Instead of:
```
def method_with_large_hash # rubocop:disable Metrics/MethodLength
...
end
```
[...]

PR #7204, opened 2019-07-10
Add new cop `Metrics/PerceivedMethodLength`

KaiqueBressi:
This cop checks if the number of statements of the method exceeds some maximum value.

There are more than one issue on rubocop talking about the `Metrics/MethodLength` not being good enough because it does count literal line breaks so methods with large array or hash literals would be catch. 

I know about `CyclomaticComplexity` but i still thinks a cop that counts statements is a great alternative to `Metrics/MethodLength` so i decided to implement it.

Issues: #494, #1077, #1361(Probably to have more issues talking about it)

I don't have too much experience with opensource contributing so if i forgot anything  please let me know, thanks :D 

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1848,6 +1848,12 @@ Metrics/PerceivedComplexity:
   VersionAdded: '0.25'
   Max: 7
 
+Metrics/PerceivedMethodLength:
+  Description: 'Avoid methods with more than 14 statements.'
+  Enabled: true
+  Max: 14
+  VersionAdded: '0.72'
+
 #################### Naming ##############################
 
 Naming/AccessorMethodName:
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -353,6 +353,7 @@ require_relative 'rubocop/cop/lint/useless_setter_call'
 require_relative 'rubocop/cop/lint/void'
 
 require_relative 'rubocop/cop/metrics/cyclomatic_complexity'
+require_relative 'rubocop/cop/metrics/perceived_method_length'
 # relies on cyclomatic_complexity
 require_relative 'rubocop/cop/metrics/utils/abc_size_calculator'
 require_relative 'rubocop/cop/metrics/abc_size'
--- a/lib/rubocop/cop/metrics/perceived_method_length.rb
+++ b/lib/rubocop/cop/metrics/perceived_method_length.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Metrics
+      # This cop checks if the number of statements of the method
+      # exceeds some maximum value.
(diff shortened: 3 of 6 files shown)


bbatsov:
Hmm, I'm on the fence about this. I get the idea, but I'm worried about cases with conditional logic, as there are things are not that clear cut. I've seen you just count things like `if` as one extra expression, which is probably reasonable. Let's see what the others (most notably @jonas054 think about your idea). 

Btw, as Ruby has mostly expressions (not statements), I'd go with this terminology.

jonas054:
> I know about CyclomaticComplexity but i still thinks a cop that counts statements is a great alternative to Metrics/MethodLength so i decided to implement it.

Yes, but do you also know about `Metrics/AbcSize`, which was added just for this reason; to detect methods that are "too large" rather than "too long" (which is what `Metrics/MethodLength` detects)? Does the new cop bring anything new to the table compared to `AbcSize`?

KaiqueBressi:
> > I know about CyclomaticComplexity but i still thinks a cop that counts statements is a great alternative to Metrics/MethodLength so i decided to implement it.
> 
> Yes, but do you also know about `Metrics/AbcSize`, which was added just for this reason; to detect methods that are "too large" rather than "too long" (which is what `Metrics/MethodLength` detects)? Does the new cop bring anything new to the table compared to `AbcSize`?

On my point of view `Metrics/AbcSize` is a way to complement `Metrics/MethodLength`, but the `Metrics/PerceivedMethodLength` is an alternative to it. 

You can have a `Metrics/AbcSize` with 17 of Max and `Metrics/MethodLength` with 7 of Max and still face the problem of having a large literal being returned by a method. Using just `Metrics/PerceivedMethodLength` with `Metrics/AbcSize` this problem wont happen. 

Most languages have linters that counts by statements, It would be really nice to have something like that on rubocop, what do you think?

jonas054:
I still don't see in what situations it would be better to run both `Metrics/PerceivedMethodLength` and `Metrics/AbcSize` instead of only `Metrics/AbcSize`. The cops seem too similar IMO. Can you give an example of inspected code where this would be useful?

KaiqueBressi:
> I still don't see in what situations it would be better to run both `Metrics/PerceivedMethodLength` and `Metrics/AbcSize` instead of only `Metrics/AbcSize`. The cops seem too similar IMO. Can you give an example of inspected code where this would be useful?

You are right, `Metrics/AbcSize` is a better metric to count how long a method is, maybe i should have opened an issue before trying to implement it haha. 

I still think `Metrics/PerceivedMethodLength` or just `MethodLength` is easier to understand than `Metrics/AbcSize`, but i can live without it. Thanks for the attention

jonas054:
OK, closing then. Thanks for the effort, though, and don't feel discouraged to contribute again.

PR #7208, opened 2019-07-15
[Fix #6653] IndentHeredoc: Do not run into next line

### LINKED ISSUE #6653: Autocorrect for `Layout/IndentHeredoc` mangles Heredoc
When `Layout/IndentHeredoc` detects a violation, its autocorrect can mess with Heredocs.

**Example file (`foo.rb`):**

```ruby
module Foo
  class Bar < Baz
    class << self
      def generate_comment(_data)
        <<~MARKDOWN
        **This is a comment**

        This is a sentence about the comment.

        - [ ] Task 1
        - [ ] Task 2:
          - Sub-task 1
          - Sub-task 2
          - Sub-task 3

        Another sentence about the comment.
        MARKDOWN
      end
    end
  end
end
```

Note that contents of the Heredoc is not at the correct indentation.

Now run the autocorrect: `rubocop -x foo.rb`

**Result:**

```ruby
module Foo
  class Bar < Baz
    class << self
      def generate_comment(_data)
        <<~MARKDOWN
          **This is a comment**
           This is a sentence about the comment.
           - [ ] Task 1
          - [ ] Task 2:
            - Sub-task 1
            - Sub-task 2
            - Sub-task 3
           Another sentence about the comment.
        MARKDOWN
      end
    end
  end
end
```

**Expected:**

```ruby
module Foo
  class Bar < Status
    class << self
      def generate_comment(_data)
        <<~MARKDOWN
          **This is a comment**

          This is a sentence about the comment.

          - [ ] Task 1
          - [ ] Task 2:
            - Sub-task 1
            - Sub-task 2
            - Sub-task 3

          Another sentence about the comment.
        MARKDOWN
      end
    end
  end
end
```

--------

## Expected behavior

I expected the contents of the Heredoc to be unchanged, except for the additional 2 spaces added to each line. 
Empty lines would still exist and be empty (admittedly, I'm not sure how empty lines should work in general).

## Actual behavior

Empty lines were removed, and some of the lines got 3 spaces, not 2 (See `Task 1`).

`rubocop -x foo.rb` on the file above.

## RuboCop version

```
rubocop -V
0.61.1 (using Parser 2.5.3.0, running on ruby 2.5.3 x86_64-darwin18)
```

buehmann:
This closes #6653 (if it were open; I was not able to open it again after the stale bot had closed it).

When finding the indentation whitespace to be replaced in a multiline
string, the regexp `/\s*/` could run into the next line (because it
matches `\n`).

I also fixed across a tiny bug in the documentation of the cop.

### DIFF
--- a/lib/rubocop/cop/layout/indent_heredoc.rb
+++ b/lib/rubocop/cop/layout/indent_heredoc.rb
@@ -37,7 +37,7 @@ module RuboCop
       #   # good
       #   # When EnforcedStyle is powerpack, bad code is auto-corrected to
       #   # the following code.
-      #   <<~RUBY
+      #   <<-RUBY.strip_indent
       #     something
       #   RUBY
       #
@@ -206,7 +206,8 @@ module RuboCop
           body = heredoc_body(node)
           body_indent_level = indent_level(body)
           correct_indent_level = base_indent_level(node) + indentation_width
-          body.gsub(/^\s{#{body_indent_level}}/, ' ' * correct_indent_level)
+          body.gsub(/^[^\S\r\n]{#{body_indent_level}}/,
+                    ' ' * correct_indent_level)
         end
 
         def indented_end(node)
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/layout/indent_heredoc.rb
```diff
@@ -37,7 +37,7 @@ module Layout
       #   # good
       #   # When EnforcedStyle is powerpack, bad code is auto-corrected to
       #   # the following code.
-      #   <<~RUBY
+      #   <<~RUBY.strip_indent
```
bbatsov: The heredoc should be normal here, not a squigly one.

buehmann: Oh, right. Fixed.

bbatsov:
Thanks!

lib/rubocop/cop/layout/indent_heredoc.rb
koic: This squiggly heredoc is a mistake that was included below. Thanks for the fixing.
https://github.com/rubocop-hq/rubocop/commit/5ffddbb92926122a4f3fc1bf8ae29d5be31d481f

PR #7214, opened 2019-07-17
Update manual with more information on --disable-uncorrectable

jonas054:
This should make it easier to discover and use `--disable-uncorrectable`.
Cc: @vergenzt


bbatsov:
Nice!

PR #7217, opened 2019-07-17
Make TrailingMethodEndStatement actually work

buehmann:
I stumbled across this bug by accident while looking at how different cops navigate within tokens. And I thought: This cannot work …

The cop was using cached tokens from the first `def` when checking and autocorrecting the next ones.

For example none of the two offenses was discovered in
```
def bar
  0
end

def foo
  1; end

def foo
  2; end
```
because the right placement of `end` in the first `def` "leaked" into the following ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #7224, opened 2019-07-20
Refactor Config class to decrease length

jonas054:
Here's a refactoring PR that deals with the big `Config` class by splitting it in three. Two classes are extracted from `Config`, one for validation and one for obsolete configuration. 

I have used the `Forwardable` mixin, which we haven't done much before in RuboCop. I'm not sure if it's been discussed before, and I was worried that it might affect performance, but I've done some measurements to reassure myself that there's no degradation.

The `target_rails_version` method remains in `Config` while `target_ruby_version` is moved to `ConfigValidation`. This seems inconsistent but it's not, actually. The target ruby version is used in validation while the target rails version is not. So there.

The three classes are fairly tightly coupled and there are calls (delegated) in two directions. I still prefer has-a over is-a (mixin) relationships, because is-a would be even tighter.

No user-observable changes, so no changelog entry.


lib/rubocop.rb
```diff
@@ -587,10 +587,12 @@
 require_relative 'rubocop/config_loader_resolver'
 require_relative 'rubocop/config_loader'
 require_relative 'rubocop/config_store'
+require_relative 'rubocop/config_validator'
 require_relative 'rubocop/target_finder'
 require_relative 'rubocop/token'
 require_relative 'rubocop/comment_config'
 require_relative 'rubocop/magic_comment'
+require_relative 'rubocop/obsolete_config'
```
bbatsov: Maybe we should name this `config_something` for consistency with the other config-related classes - e.g. `config_obsoletions` or `config_changes` or something much better? :-)

lib/rubocop/config_validator.rb
```diff
@@ -0,0 +1,238 @@
+# frozen_string_literal: true
+
+require 'pathname'
+
+module RuboCop
+  # ...
```
bbatsov: I guess you meant to write some description here. ;-)

bbatsov:
Btw, I've been thinking for a while that obsoletions should ideally be done via the config, as that would be more transparent and easier to reflect in the generated part of the docs.

lib/rubocop/config_validator.rb
jonas054: Ouch! :grimacing:

lib/rubocop.rb
jonas054: Good idea.

jonas054:
> obsoletions should ideally be done via the config

Interesting! I'll keep that in mind.

I've pushed some fixes.

spec/rubocop/config_obsoletion_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+          'Lint/RescueWithoutErrorClass' => { 'Enabled': true },
+          'Lint/InvalidCharacterLiteral' => { 'Enabled': true },
+          'Lint/SpaceBeforeFirstArg' => { 'Enabled': true },
+          # Split cops
+          'Style/MethodMissing' => { 'Enabled': true }
+        }
+      end
+
+      let(:expected_message) do
+        <<-OUTPUT.strip_indent.chomp
```
koic: Could you replace `strip_indent` with squiggly heredoc?

```diff
- <<-OUTPUT.strip_indent.chomp
+ <<~OUTPUT.chomp
```

spec/rubocop/config_obsoletion_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+          'Rails/UniqBeforePluck' => { 'EnforcedMode' => 'x' },
+          # Moved cops with obsolete parameters
+          'Lint/BlockAlignment' => { 'AlignWith' => 'end' },
+          'Lint/EndAlignment' => { 'AlignWith' => 'end' },
+          'Lint/DefEndAlignment' => { 'AlignWith' => 'end' }
+        }
+      end
+
+      let(:expected_message) do
+        <<-OUTPUT.strip_indent.chomp
```
koic:
```diff
- <<-OUTPUT.strip_indent.chomp
+ <<~OUTPUT.chomp
```

jonas054:
Thanks, @koic ! I've addressed your comments and pushed.

jonas054:
I took the liberty of self-merging. :slightly_smiling_face:

PR #7225, opened 2019-07-22
Support lower case drive letters on Windows

jonas054:
The combination of running a Git Bash terminal in Visual Studio Code has turned out to produce current directories that start with a lower case drive letter when RuboCop reads it. For example, `c:/Users/me/project`.

This is not good since it will mess up the conversion of Exclude paths to absolute paths. We need to recognize the above path as already being absolute.


koic:
Thanks!

PR #7222, opened 2019-07-19
Fix autocorrection of UnneededSplatExpasion when Array.new with a block is assigned to a variable

rrosenblum:
I was working on converting tests to use `expect_correction` and came across this bug in the auto-correction of `Lint/UnneededSplatExpansion` when `Array.new` with a block is assigned to a variable.

PR #7229, opened 2019-07-23
Use squiggly heredocs instead of strip_indent

rrosenblum:
Some occurrences of `strip_indent` were missed in the previous change. There are a few occurrences left in `spec/rubocop/cop/lint/unneeded_splat_expansion_spec.rb` that will be taken care of in #7222.

### DIFF
--- a/spec/rubocop/cli/cli_autocorrect_spec.rb
+++ b/spec/rubocop/cli/cli_autocorrect_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe RuboCop::CLI, :isolated_environment do
   end
 
   it 'plays nicely with default cops in complex ExtraSpacing scenarios' do
-    create_file('.rubocop.yml', <<-YAML.strip_indent)
+    create_file('.rubocop.yml', <<~YAML)
       # These cops change indentation and thus need disabling in order for the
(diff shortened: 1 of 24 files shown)


spec/rubocop/cli/cli_autocorrect_spec.rb
```diff
@@ -531,7 +531,7 @@ def verify_section
      'UnneededInterpolation offense unchanged' do
     # If we change string concatenation from plus to backslash, the string
     # literal that follows must remain a string literal.
-    source = <<-'RUBY'.strip_indent
+    source = <<~'RUBY'
```
koic: This is a known issue of https://github.com/jruby/jruby/issues/4260.  The JRuby build matrix does not pass until JRuby 9.2.8.0 release. Therefore, it is still pending to use heredoc for `<<~'EOS'`.

Related conversation https://github.com/rubocop-hq/rubocop/pull/7173#issuecomment-505552299

rrosenblum: I wish I had found that before making my changes. At least I don't have to diagnose the failing build. I'll go ahead an close this for now.

PR #7232, opened 2019-07-25
[Fix #7128] Make `Metrics/LineLength` aware of shebang

### LINKED ISSUE #7128: No way to disable Metrics/LineLength for shebang line

I don't think I can disable the `Metrics/LineLength` for just the shebang line.

(If it helps, the shebang line in question is `#!/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/bin/ruby --disable-gems`. The reason I'm doing this is to reference the system Ruby on macOS while bypassing the users gem environment.)

I'd like to be able to disable `Metrics/LineLength` for just the shebang line. Really I think this should be Rubocops default behavior, `Metrics/LineLength` shouldn't apply to a shebang line because it's not Ruby source code, and it can't reliably be shortened. But if it were a configuration option, I'd say having a syntax for it in [disabling cops within source code](https://www.rubocop.org/en/stable/configuration/#disabling-cops-within-source-code) would be nice.

Thanks!

koic:
Fixes #7128.

This PR makes `Metrics/LineLength` aware of shebang.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/metrics/line_length_spec.rb
+++ b/spec/rubocop/cop/metrics/line_length_spec.rb
@@ -21,6 +21,14 @@ RSpec.describe RuboCop::Cop::Metrics::LineLength, :config do
     expect_no_offenses('#' * 80)
   end
 
+  it 'accepts the first line if it is a shebang line' do
+    expect_no_offenses(<<~RUBY)
+      #!/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/bin/ruby --disable-gems
+
+      do_something
+    RUBY
+  end
+
   it 'registers an offense for long line before __END__ but not after' do
     inspect_source(['#' * 150,
                     '__END__',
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/metrics/line_length_spec.rb
```diff
@@ -21,6 +21,14 @@
     expect_no_offenses('#' * 80)
   end
 
+  it 'accepts the first line is the shebang line' do
```
jonas054: Maybe `accepts the first line if it is a shebang line` sound better.

koic: Thanks for the review. I fixed it.

bbatsov:
I'd also mention in the description of the cop that such lines are exempt from the rule, as that might not be obvious to our end users.

PR #7234, opened 2019-07-26
Guard clause clarification

jonas054:
For ages people have misunderstood the message from `Style/GuardClause`, confusing guard clauses with modifier forms of `if`/`unless`. This change adds some code to the message, making it clear what the cop wants.


bbatsov:
Nice improvement! As lately I've been thinking more and more about 1.0 I just wanted to say I'm also considering disabling this cop in the 1.0 release, as it's somewhat controversial whether it leads to cleaner code or not. When I find some time I'll open a few tickets with ideas for the default config change.

jonas054:
>  I'm also considering disabling this cop in the 1.0 release, as it's somewhat controversial whether it leads to cleaner code or not.

That makes sense to me.

PR #7238, opened 2019-07-29
Simplify Style/CommentedKeyword

buehmann:
I cleaned up this cop. All the code for heredoc detection was a leftover from a time when the cop didn't use `processed_source.comments` but looked at raw lines.

### DIFF
--- a/lib/rubocop/cop/style/commented_keyword.rb
+++ b/lib/rubocop/cop/style/commented_keyword.rb
@@ -33,26 +33,12 @@ module RuboCop
       #     y
       #   end
       class CommentedKeyword < Cop
-        include RangeHelp
-
         MSG = 'Do not place comments on the same line as the ' \
               '`%<keyword>s` keyword.'
 
         def investigate(processed_source)
-          heredoc_lines = extract_heredoc_lines(processed_source.ast)
-
           processed_source.each_comment do |comment|
-            location = comment.location
-            line_position = location.line
-            line = processed_source.lines[line_position - 1]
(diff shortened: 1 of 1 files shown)


lib/rubocop/cop/style/commented_keyword.rb
```diff
@@ -61,25 +47,19 @@ def investigate(processed_source)
...
+          line = line(comment)
+          KEYWORDS.any? { |word| line =~ /^\s*#{word}\s/ } &&
             ALLOWED_COMMENTS.none? { |c| line =~ /#\s*#{c}/ }
         end
 
-        def message(node)
-          line = node.source_line
-          keyword = /^\s*(\S+).*#/.match(line)[1]
+        def message(comment)
+          keyword = line(comment)[/(\S+).*#/, 1]
```
bbatsov: I think using an explicit `match` is a bit more readable. I've rarely seen the index syntax in the wild and it's likely going to confuse some people.

bbatsov:
Nicely done! Deleted code is the best code! 😉

lib/rubocop/cop/style/commented_keyword.rb
buehmann: Maybe we need a cop for this … 😜 Changed.

bbatsov: I wouldn't mind one. :D In general I think there are very few cases where implementing `[]` results in some more readable code, as for most people it has a very strong association with some indices.

PR #7239, opened 2019-07-29
Remove unused node matchers

buehmann:
Speaking of deleting code, here are some node matchers that are not used at all.

### DIFF
--- a/lib/rubocop/cop/lint/debugger.rb
+++ b/lib/rubocop/cop/lint/debugger.rb
@@ -56,8 +56,6 @@ module RuboCop
           (send (send {#kernel? nil?} :binding) :irb ...)
         PATTERN
 
-        def_node_matcher :pry_rescue?, '(send (const nil? :Pry) :rescue ...)'
-
         def on_send(node)
           return unless debugger_call?(node) || binding_irb?(node)
 
--- a/lib/rubocop/cop/mixin/documentation_comment.rb
+++ b/lib/rubocop/cop/mixin/documentation_comment.rb
@@ -9,8 +9,6 @@ module RuboCop
 
       private
 
-      def_node_matcher :constant_definition?, '{class module casgn}'
-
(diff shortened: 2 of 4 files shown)


bbatsov:
Nice! How did you even manage to figure this out?

buehmann:
I was interested in how often a specific node matcher was run (the reason for that escapes me right now), so I hacked some call counting into the node patterns and ran the specs. When I looked at the numbers I stumbled across some zeros:
* Many came from rubocop-rspec (whose node pattern are not used in the specs)
* 4 of them are the ones above
* 2 additional can be found in RuboCop::AST::Node (I don't know if they are considered part of some interface that cops in other gems might use?): https://github.com/rubocop-hq/rubocop/compare/master...buehmann:unused-node-matchers-in-node
* There was even a case where a node matcher was crucial for the cop's functionality but never used in the specs (because of a bug in the cop): https://github.com/rubocop-hq/rubocop/compare/master...buehmann:fix-constant-visibility

bbatsov:
Great research!

> 2 additional ones can be found in RuboCop::AST::Node (I don't know if they are considered part of some interface that cops in other gems might use?): master...buehmann:unused-node-matchers-in-node

Likely we just used them in the past. If they are not used in the 3 "official" extensions we can safely remove them. 

> There was even a case where a node matcher was crucial for the cop's functionality but never used in the specs (because of a bug in the cop): master...buehmann:fix-constant-visibility

Nice catch!

Drenmi:
Awesome! This is something that has been on my mind for a while, as I've been working on extracting things from `Node` into more specific node extensions, but I never got around to doing it. 🙏

buehmann:
> Likely we just used them in the past. If they are not used in the 3 "official" extensions we can safely remove them.

Okay! https://github.com/rubocop-hq/rubocop/pull/7240

PR #7243, opened 2019-07-30
[Fix #7236] Prevent bad auto-correction from Style/InverseMethods and Style/SymbolProc

### LINKED ISSUE #7236: auto-correct Style/InverseMethods and Style/SymbolProc together inverts meaning of code
Code like

```[:a, false, :b, false].reject { |x| !x }```

violates `Style/InverseMethods` and `Style/SymbolProc`. Fixing one or the other should be enough but when `--auto-correct`'d we get 

```[:a, false, :b, false].select(&:!)```

which has the opposite meaning. I realize Style/SymbolProc is not auto-correct-safe but here the correction _would_ be safe except for interfering with Style/InverseMethods (and this can be seen with `rubocop --safe --auto-correct`).

--------

## Expected behavior

After applying the auto-correction for Style/InverseMethods there should be no violation for Style/SymbolProc so it should not apply.

## Actual behavior

Style/SymbolProc correction was applied changing the meaning of the code.

* have a file test.rb:
```
puts([:a, false, :b, false].reject { |x| !x })
```
* `rubocop --auto-correct test.rb`
```Offenses:

test.rb:3:6: C: [Corrected] Style/InverseMethods: Use select instead of inverting reject.
puts([:a, false, :b, false].reject { |x| !x })
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
test.rb:3:36: C: [Corrected] Style/SymbolProc: Pass &:! as an argument to reject instead of a block.
puts([:a, false, :b, false].reject { |x| !x })
                                   ^^^^^^^^^^
```
* `ruby test.rb` now produces different output

## RuboCop version

0.73.0 (using Parser 2.6.3.0, running on ruby 2.6.0 x86_64-darwin18)

Drenmi:
These two cops had an interaction between their auto-corrections which produced code that wasn't semantically equivalent to the original. Example:

```ruby
[].reject { |x| !x }
```

was being corrected to:

```ruby
[].select(&:!)
```

This change fixes that by marking `Style/InverseMethods` auto-correct as incompatible with that of `Style/SymbolProc`.

-----------------

* [X] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 * [#6861](https://github.com/rubocop-hq/rubocop/issues/6861): Fix a false positive for `Layout/IndentationWidth` when using `EnforcedStyle: outdent` of `Layout/AccessModifierIndentation`. ([@koic][])
 * [#7235](https://github.com/rubocop-hq/rubocop/issues/7235): Fix an error where `Style/ConditionalAssignment` would swallow a nested `if` condition. ([@buehmann][])
 * [#7242](https://github.com/rubocop-hq/rubocop/issues/7242): Make `Style/ConstantVisibility` work on non-trivial class and module bodies. ([@buehmann][])
+* [#7236](https://github.com/rubocop-hq/rubocop/pull/7236): Mark `Style/InverseMethods` auto-correct as incompatible with `Style/SymbolProc`. ([@drenmi][])
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


bbatsov:
The changes look good to me (apart from the failing build).

Drenmi:
Hm. I was a bit confused by this. Tests and default rake task both definitely passed. 🤔 Maybe I mindlessly did some formatting change that broke the code. Will fix soon.

Drenmi:
@bbatsov I looked into the other cops that define incompatible auto-corrects. It looks like we currently don't have a way of adding tests for this. 🤔 I removed the non-functioning test from this PR, and will see if I can add some infrastructure for that in another PR.

bbatsov:

PR #7244, opened 2019-07-30
[Fix #7242] Make Style/ConstantVisibility work

### LINKED ISSUE #7242: Style/ConstantVisibility does not work on non-trivial class bodies
This is the issue I already mentioned [here](https://github.com/rubocop-hq/rubocop/pull/7239#issuecomment-515883573). I will add a PR fixing it shortly.

`Style/ConstantVisibility` does not do what it says on the cover.

Enable the cop in your .rubocop.yml
```yaml
Style/ConstantVisibility:
  Enabled: true
```
Run rubocop on the following example taken straight from the [documentation](https://github.com/rubocop-hq/rubocop/blob/f51c53c0be3c16b47407c947d85beb3db21e86ac/lib/rubocop/cop/style/constant_visibility.rb#L6-L27):
```ruby
# frozen_string_literal: true

# This cop checks that constants defined in classes and modules have
# an explicit visibility declaration. By default, Ruby makes all class-
# and module constants public, which litters the public API of the
# class or module. Explicitly declaring a visibility makes intent more
# clear, and prevents outside actors from touching private state.
#
# @example

# bad
class Foo
  BAR = 42
  BAZ = 43
end

# good
class Foo
  BAR = 42
  private_constant :BAR

  BAZ = 43
  public_constant :BAZ
end
```

## Actual behavior

```
Inspecting 1 file
.

```

## Expected behavior

```
Inspecting 1 file
C

Offenses:

constant_visibility.rb:13:3: C: Style/ConstantVisibility: Explicitly make BAR public or private using either #public_constant or #private_constant.
  BAR = 42
  ^^^^^^^^
constant_visibility.rb:14:3: C: Style/ConstantVisibility: Explicitly make BAZ public or private using either #public_constant or #private_constant.
  BAZ = 43
  ^^^^^^^^

```

## RuboCop version

master (f51c53c0be3c16b47407c947d85beb3db21e86ac)

buehmann:
This closes #7242.

The cop was not able to deal with multi-statement bodies because it did
not know how to handle a `:begin` node inside a class or module body.
Fixing that revealed another pattern-matching bug that had been masked
before because the pattern was never actually used in the specs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * [#7128](https://github.com/rubocop-hq/rubocop/issues/7128): Make `Metrics/LineLength` aware of shebang. ([@koic][])
 * [#6861](https://github.com/rubocop-hq/rubocop/issues/6861): Fix a false positive for `Layout/IndentationWidth` when using `EnforcedStyle: outdent` of `Layout/AccessModifierIndentation`. ([@koic][])
 * [#7235](https://github.com/rubocop-hq/rubocop/issues/7235): Fix an error where `Style/ConditionalAssignment` would swallow a nested `if` condition. ([@buehmann][])
+* [#7242](https://github.com/rubocop-hq/rubocop/issues/7242): Make `Style/ConstantVisibility` work on non-trivial class and module bodies. ([@buehmann][])
 
 ### Changes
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Nicely done! Thanks!

PR #7228, opened 2019-07-23
[Fix #5628] Rewrite SpaceInsideStringInterpolation

buehmann:
This fixes #5628 by rewriting `Layout/SpaceInsideStringInterpolation` in terms of the infrastructure used by similar `SpaceInside*` cops: `SurroundingSpace` and `SpaceCorrector`.

In order to to that I had to lift some limitations of  `SurroundingSpace`:
* Previously it assumed that each delimiter token was exactly 1 char wide. It can now handle wider tokens such as `#{`.
* Previously it scanned across tab characters when detecting surrounding space, but did not include them in the range to be removed. This mismatch is gone.

In style `space` there was an overlap in functionality with cop `ExtraSpace`: `Layout/SpaceInsideStringInterpolation` no longer removes excess whitespace and leaves that to `ExtraSpace`. That's why the specs needed to change.

Because of that slight change in functionality I was wondering whether to include this into the "bug fixes" or "changes" section in the change log. For now I just classified this as a bug fix.

PR #7245, opened 2019-07-30
Detect more interpolations

buehmann:
I discovered this while working on #7228. Many cops that inspect string interpolations `#{…}` only find them in one or more of the following places, but not in all of them:
* string literals (including `%W` and `%I`)
* regexp literals
* backticks/`%x`
* symbols: `:"foo#{bar}"`

This change extends those cops to cover all possibilities.


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 
 * [#5265](https://github.com/rubocop-hq/rubocop/issues/5265): Improved `Layout/ExtraSpacing` cop to handle nested consecutive assignments. ([@jfelchner][])
 * [#7215](https://github.com/rubocop-hq/rubocop/issues/7215): Make it clear what's wrong in the message from `Style/GuardClause`. ([@jonas054][])
+* Make cops detect string interpolations in more contexts: inside of backticks, regular expressions, and symbols. ([@buehmann][])
```
koic: Please add the URL of this PR at the beginning of the line.

buehmann: Done.

lib/rubocop/cop/mixin/interpolation.rb
```diff
@@ -0,0 +1,26 @@
...
+      alias on_dsym on_dstr
+      alias on_regexp on_dstr
+
+      def on_node_with_interpolations(node)
+        node.each_child_node(:begin) do |begin_node|
+          on_interpolation(begin_node)
+        end
+      end
+
+      def on_interpolation(_begin_node)
```
bbatsov: WIP?

koic: I think this is a template method that should be defined in subclasses.  It may be clearer to understand if a message is given to an argument.

e.g. `raise NotImplementedError('Method should be overridden in child classes')`

buehmann: To be implemented in classes using that mixin. This was inspired by https://github.com/rubocop-hq/rubocop/blob/16427114b8d80ff15909c68c0f2ac83a3a6271bd/lib/rubocop/cop/mixin/heredoc.rb#L17-L19

If you prefer, I can delete the method.

buehmann: I replaced the method by documentation.

bbatsov:
Thanks!

PR #7249, opened 2019-08-01
Deprecate the SafeMode module

rrosenblum:
Discussion in https://github.com/rubocop-hq/rubocop-performance/pull/72. It looks like `Performance/Count` and `Performance/Detect` were the only cops making use of the `SafeMode` feature. Rather than disabling the cops from running, we have flagged them as unsafe for auto-correction.

This will need to wait until the change is merged into RuboCop-Performance, and a new version of the gem is released.

### DIFF
--- a/lib/rubocop/cop/mixin/safe_mode.rb
+++ b/lib/rubocop/cop/mixin/safe_mode.rb
@@ -4,6 +4,9 @@ module RuboCop
   module Cop
     # Common functionality for Rails safe mode.
     module SafeMode
+      warn 'The `SafeMode` option will be removed in `RuboCop` 0.76. ' \
+        'Please update `rubocop-performance` to 1.15.0 or higher.'
+
       private
 
       def rails_safe_mode?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 * [#5265](https://github.com/rubocop-hq/rubocop/issues/5265): Improved `Layout/ExtraSpacing` cop to handle nested consecutive assignments. ([@jfelchner][])
 * [#7215](https://github.com/rubocop-hq/rubocop/issues/7215): Make it clear what's wrong in the message from `Style/GuardClause`. ([@jonas054][])
 * [#7245](https://github.com/rubocop-hq/rubocop/pull/7245): Make cops detect string interpolations in more contexts: inside of backticks, regular expressions, and symbols. ([@buehmann][])
+* Deprecate the `SafeMode` option. Users will need to upgrade `rubocop-performance` to 1.15.0+ before the `SafeMode` module is removed. ([@rrosenblum][])
 
 ## 0.73.0 (2019-07-16)
 


lib/rubocop/cop/mixin/safe_mode.rb
```diff
@@ -1,22 +0,0 @@
...
-    # Common functionality for Rails safe mode.
-    module SafeMode
-      private
-
-      def rails_safe_mode?
-        safe_mode? || rails?
-      end
-
-      def safe_mode?
-        cop_config['SafeMode']
```
koic: RuboCop Performance 1.4 or lower users will suddenly encounter errors.

```console

#!/bin/bash -eo pipefail
bundle exec rake internal_investigation
Running RuboCop...
uninitialized constant RuboCop::Cop::Performance::Count::SafeMode
/usr/local/bundle/gems/rubocop-performance-1.4.1/lib/rubocop/cop/performance/count.rb:41:in `<class:Count>'
/usr/local/bundle/gems/rubocop-performance-1.4.1/lib/rubocop/cop/performance/count.rb:40:in `<module:Performance>'
/usr/local/bundle/gems/rubocop-performance-1.4.1/lib/rubocop/cop/performance/count.rb:5:in `<module:Cop>'
/usr/local/bundle/gems/rubocop-performance-1.4.1/lib/rubocop/cop/performance/count.rb:4:in `<module:RuboCop>'
/usr/local/bundle/gems/rubocop-performance-1.4.1/lib/rubocop/cop/performance/count.rb:3:in `<top (required)>'
/usr/local/bundle/gems/rubocop-performance-1.4.1/lib/rubocop/cop/performance_cops.rb:7:in `require_relative'
/usr/local/bundle/gems/rubocop-performance-1.4.1/lib/rubocop/cop/performance_cops.rb:7:in `<top (required)>'
/usr/local/bundle/gems/rubocop-performance-1.4.1/lib/rubocop-performance.rb:11:in `require_relative'
/usr/local/bundle/gems/rubocop-performance-1.4.1/lib/rubocop-performance.rb:11:in `<top (required)>'
```

https://circleci.com/gh/rubocop-hq/rubocop/63589

I think it is user-friendly to display a warning first. The following is an example.

```diff
 def safe_mode?
+  warn '`SafeMode` option will be removed. ' \
+       'Please update rubocop-performance 1.15.0 or higher.'
+
   cop_config['SafeMode']
 end
```

Warnings are noisy, but it's better to show the solution than sudden errors.

The next step is to remove it.

rrosenblum: That sounds like a good plan to me. Do we want to specify the version of RuboCop that we will be removing the `safe_mode?` method from? That way it gives people a time frame of when they need to make the change as well.

koic: Good! The following is an example of milestones:

- RuboCop 0.75 (next release) : Warn `safe_mode?`
- RuboCop 0.76 : Remove `safe_mode?`

rrosenblum: Is there a better place for this deprecation warning to live? Adding the message here will add the warning to the terminal every time that `safe_mode?` is invoked. Since this happens inside of `on_send` in both `Performance/Detect` and `Performance/Count`, the output of RuboCop becomes unreadable rather quickly. 

I've pushed an update that includes a class variable to prevent the message from being printed multiple times. It seems like there should be a better solution to this problem.

Maybe this check could be moved to where we load the plugins into RuboCop.

lib/rubocop/cop/mixin/safe_mode.rb
```diff
@@ -1,16 +1,25 @@
...
       private
 
       def rails_safe_mode?
         safe_mode? || rails?
       end
 
       def safe_mode?
+        unless @@warned
+          @@warned = true
+          warn 'The `SafeMode` option will be removed in `RuboCop` 0.76. ' \
```
bbatsov: I guess you can just move the warning to the body of the module, as I think everything in it is scheduled for removal.

rrosenblum: Good idea! Updated.

koic:
The following warning is always displayed even if users don't use `SafeMode`.

```console
% rubocop -v
The `SafeMode` option will be removed in `RuboCop` 0.76. Please update `rubocop-performance` to 1.15.0 or higher.
0.74.0
```

But I haven't got a good solution yet.

bbatsov:
I guess we shouldn't require this file by default.

rrosenblum:
If we follow the plan in the message, this won't be an issue when we remove the file from 0.76.0.

buehmann:
@rrosenblum The message tells users to ```update `rubocop-performance` to 1.15.0```. Shouldn't that be `1.5.0` instead?

koic:
Yeah, next version of RuboCop Performance is planned for 1.5.0.

PR #7268, opened 2019-08-07
 [Fix #7262] Detect more named format sequences

### LINKED ISSUE #7262: FormatParameterMismatch fails when modifiers appear before name
## Expected behavior

When using the `Lint/FormatParameterMismatch` cop, I expect it to accept all valid variations of `format` strings, including those where modifiers appear before the named substitution.

## Actual behavior

When running the `Lint/FormatParameterMismatch` cop, it does not recognize valid `format` strings where modifiers are supplied before the named parameter.

For example, the following lines produce equivalent output in Ruby v2.6.3 (note the placement of the modifier digit `2` before and after the named substitution):
```
$ irb
2.6.3 :001 > format('%2<greeting>s', greeting: 'Hello')
 => "Hello"
2.6.3 :002 > format('%<greeting>2s', greeting: 'Hello')
 => "Hello"
```

The first one produces a `Lint/FormatParameterMismatch` warning, which is unexpected.

Run the `Lint/FormatParameterMismatch` cop against the Ruby code `format('%2<greeting>s', greeting: 'Hello')`

## RuboCop version

```
$ rubocop -V
0.74.0 (using Parser 2.6.3.0, running on ruby 2.6.3 x86_64-darwin17)
```

buehmann:
This fixes #7262 and removes some obsolete code.

### DIFF
--- a/lib/rubocop/cop/lint/format_parameter_mismatch.rb
+++ b/lib/rubocop/cop/lint/format_parameter_mismatch.rb
@@ -41,7 +41,6 @@ module RuboCop
         # @see https://ruby-doc.org/core-2.6.3/Kernel.html#method-i-format
         FIELD_REGEX =
           /(%(?:(#{FLAG}*)#{WIDTH}?#{PRECISION}?#{TYPE}|%))/.freeze
-        NAMED_FIELD_REGEX = /%\{[_a-zA-Z][_a-zA-Z]+\}/.freeze
 
         NAME = /(?:<\w+>|\{\w+\})/.freeze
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/format_parameter_mismatch.rb
```diff
@@ -22,16 +22,22 @@ class FormatParameterMismatch < Cop
         # http://rubular.com/r/CvpbxkcTzy
         MSG = "Number of arguments (%<arg_num>i) to `%<method>s` doesn't " \
               'match the number of fields (%<field_num>i).'
+
+        FLAG = /[\s#+-0\*]/.freeze
```
bbatsov: Might be a good idea to add some comments explaining what each of those are supposed to match, as it's getting a bit convoluted IMO.

buehmann: Good idea. I am adding a few pointers to the documentation I was using. I am going to squash this later.

bbatsov:
Thanks!

PR #7270, opened 2019-08-08
[Fix #7029] Make `Style/Attr` not flag offense for custom `attr` method

### LINKED ISSUE #7029: Style/Attr is broken when an "attr" method is defined
if I have a method called `attr`, e.g. `attr(*args, &block)` or something like that then it breaks this cop.

--------

## Expected behavior

maybe it could only inspect classes/modules body?

## Actual behavior

inspects methods body

source:

```ruby
class A
  def attr(*args)
    p args
  end

  def a
    attr(1)
    attr(:a, 1)
    attr(:b) { 2 }
    attr(:c) do |a|
      2
    end
  end
end
```

auto-corrected:

```ruby
class A
  def attr(*args)
    p args
  end

  def a
    attr_reader(1)
    attr_reader(:a, 1)
    attr_reader(:b) { 2 }
    attr_reader(:c) do |_a|
      2
    end
  end
end
```

(not sure how but I also had some `attr_accessor` and `attr_writer` inserted in my code, it somehow generated invalid ruby even like `attr_writer(:something`)

## RuboCop version

```
$ rubocop -V
0.68.1 (using Parser 2.6.3.0, running on ruby 2.6.3 x86_64-darwin18)
```

tejasbubane:
Closes #7029 

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#7231](https://github.com/rubocop-hq/rubocop/issues/7231): Fix the exit code to be `2` rather when `0` when the config file contains an unknown cop. ([@jethroo][])
 * [#7513](https://github.com/rubocop-hq/rubocop/issues/7513): Fix abrupt error on autocorrecting with `--disable-uncorrectable`. ([@tejasbubane][])
 * [#7537](https://github.com/rubocop-hq/rubocop/issues/7537): Fix a false positive for `Layout/SpaceAroundOperators` when using a Rational literal with `/` (e.g. `2/3r`). ([@koic][])
+* [#7029](https://github.com/rubocop-hq/rubocop/issues/7029): Make `Style/Attr` not flag offense for custom `attr` method. ([@tejasbubane][])
 
 ## 0.77.0 (2019-11-27)
 
--- a/lib/rubocop/cop/style/attr.rb
+++ b/lib/rubocop/cop/style/attr.rb
@@ -21,6 +21,10 @@ module RuboCop
 
         def on_send(node)
           return unless node.command?(:attr) && node.arguments?
+          # check only for method definitions in class/module body
+          return if node.parent &&
+                    !node.parent.class_type? &&
+                    !class_eval?(node.parent)
 
           add_offense(node, location: :selector)
         end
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/attr.rb
```diff
@@ -21,6 +21,7 @@ class Attr < Cop
 
         def on_send(node)
           return unless node.command?(:attr) && node.arguments?
+          return if node.parent && !node.parent.class_type?
```
bbatsov: So, the idea here is to check you're not calling the method in the class/module body, right? Might be a good idea to add a note explaining it.

bbatsov:
I'm OK with the proposed fix, although I really think it's generally a bad idea to define methods with such names (at they might confuse people). I guess most of the time it happens as people simply don't know they exist.

bbatsov:
Ping :-)

lib/rubocop/cop/style/attr.rb
Drenmi: I believe this will create some false negatives, for example in `#class_eval` blocks?

tejasbubane: Thanks I missed that! I just made a change to consider `class_eval` & `module_eval`.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 * [#7077](https://github.com/rubocop-hq/rubocop/issues/7077): **(Breaking)** Further standardisation of cop names. ([@scottmatthewman][])
 * [#7469](https://github.com/rubocop-hq/rubocop/pull/7469): **(Breaking)** Replace usages of the terms `Whitelist` and `Blacklist` with better alternatives. ([@koic][])
 
+### Bug fixes
+
+* [#7029](https://github.com/rubocop-hq/rubocop/issues/7029): Make `Style/Attr` not flag offense for custom `attr` method. ([@tejasbubane][])
```
koic: Please list changelog section in the order `### Bug fixes`, `### Changes`.

tejasbubane: done

tejasbubane:
The JRuby failures look unrelated to this change.

bbatsov:
@tejasbubane For some reason I can't rebase this from the GitHub UI. Can you rebase it again and ping me to merge it?

tejasbubane:
@bbatsov Rebased.

bbatsov:
Thanks!

PR #7272, opened 2019-08-08
Show warning message if passed string to 'Enabled' key in .rubocop.yml

unasuke:
## feature
It causes unexpected cop warnings because treats as truthy value if passed neither true nor false in 'Enabled' key's value like "Enabled: disable".  

## motivation
Our team occurred a problem that, unexpected  cop was in effect.
The cause was the wrong config value like an "Enabled: disabled".

-----------------

* [x] Added tests.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -17,7 +17,7 @@ Metrics/AbcSize:
 # Offense count: 47
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 174
+  Max: 182
 
 # Offense count: 219
 # Configuration parameters: CountComments, ExcludedMethods.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 * [#7274](https://github.com/rubocop-hq/rubocop/issues/7274): Add new `Lint/SendWithMixinArgument` cop. ([@koic][])
+* [#7272](https://github.com/rubocop-hq/rubocop/pull/7272): Show warning message if passed string to `Enabled`, `Safe`, `SafeAutocorrect`, and `AutoCorrect` keys in .rubocop.yml. ([@unasuke][])
 
 ### Bug fixes
 
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -199,6 +199,8 @@ module RuboCop
           raise(TypeError, "Malformed configuration in #{absolute_path}")
         end
 
+        check_cop_config_value(hash)
+
         hash
       end
 
@@ -217,6 +219,22 @@ module RuboCop
                         "`#{value}` is concealed by duplicate"
                     end
           warn Rainbow(message).yellow
+        end
+      end
+
+      def check_cop_config_value(hash, parent = nil)
+        hash.each do |key, value|
+          check_cop_config_value(value, key) if value.is_a?(Hash)
+
+          next unless %w[Enabled
+                         Safe
+                         SafeAutoCorrect
+                         AutoCorrect].include?(key) && value.is_a?(String)
+
+          abort(
+            "Property #{Rainbow(key).yellow} of cop #{Rainbow(parent).yellow}" \
+            " is supposed to be a boolean and #{Rainbow(value).yellow} is not."
+          )
         end
       end
 
--- a/spec/rubocop/config_loader_spec.rb
+++ b/spec/rubocop/config_loader_spec.rb
@@ -992,6 +992,24 @@ RSpec.describe RuboCop::ConfigLoader do
       end
     end
 
+    context 'set neither true nor false to value to Enabled' do
+      before do
+        create_file(configuration_path, <<~YAML)
+          Layout/AlignArray:
+            Enabled: disable
+        YAML
+      end
+
+      it 'gets a warning message' do
+        expect do
+          load_file
+        end.to raise_error(
+          SystemExit,
+          /supposed to be a boolean and disable is not/
+        )
+      end
+    end
+
     context 'when the file does not exist' do
       let(:configuration_path) { 'file_that_does_not_exist.yml' }
 


unasuke:
hmm,  How do I fix the "Metrics/ClassLength: Class has too many lines. [179/174]" ...?

.rubocop_todo.yml
```diff
@@ -17,7 +17,7 @@ Metrics/AbcSize:
 # Offense count: 47
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 174
+  Max: 179
```
unasuke: I can't write this code for more short. So I changed max value in `Metrics/ClassLength` in .rubocop_todo.yml

lib/rubocop/config_loader.rb
```diff
@@ -220,6 +222,19 @@ def check_duplication(yaml_code, absolute_path)
...
 
+      def check_enabled_value(hash, parent = nil)
+        hash.each do |key, value|
+          check_enabled_value(value, key) if value.is_a?(Hash)
+
+          next unless key == 'Enabled' && value.class == String
+
+          warn Rainbow(
+            "Cop `#{parent}` is enabled by value '#{value}' specified. " \
+            'It may unexpected behavior unless set ture or false.'
```
koic: `ture` is a typo of `true`.

unasuke: Fixed! thank you!

lib/rubocop/config_loader.rb
```diff
@@ -220,6 +222,19 @@ def check_duplication(yaml_code, absolute_path)
         end
       end
 
+      def check_enabled_value(hash, parent = nil)
+        hash.each do |key, value|
+          check_enabled_value(value, key) if value.is_a?(Hash)
+
+          next unless key == 'Enabled' && value.class == String
```
bbatsov: I think it'd be best to do this check for all common cop configurations that are booleans - e.g. `Safe`, `Autocorrect`, etc.

lib/rubocop/config_loader.rb
```diff
@@ -220,6 +222,19 @@ def check_duplication(yaml_code, absolute_path)
...
+
+          warn Rainbow(
+            "Cop `#{parent}` is enabled by value '#{value}' specified. " \
+            'It may unexpected behavior unless set true or false.'
```
bbatsov: I think we should probably raise an error instead and just say that property X of cop Y is supposed to be a boolean and it's not.

lib/rubocop/config_loader.rb
unasuke: Wow, It makes sense.

lib/rubocop/config_loader.rb
unasuke: OK. I change to raise an exception.

lib/rubocop/config_loader.rb
unasuke: I saw the docs https://docs.rubocop.org/en/latest/configuration/ .

I found only `Enabled` and `SafeAutoCorrect` to accept a boolean value key in available on .rubocop.yml.

lib/rubocop/config_loader.rb
unasuke: Like this?

Screenshot from 2019-08-10 20-05-44

lib/rubocop/config_loader.rb
bbatsov: There's also `AutoCorrect` and `Safe`. Guess we forgot to document them.

lib/rubocop/config_loader.rb
bbatsov: Actually instead of exception this can just be an error message with some error code. I guess the stack trace will just confuse most users. 

It will also be a good idea to colour the names of the cop and the property in the error message and to mention its current value that's causing the error.

lib/rubocop/config_loader.rb
unasuke: Ok, I'll add `AutoCorrect` and `Safe` to condition.

lib/rubocop/config_loader.rb
unasuke: How is this?

Screenshot from 2019-08-10 21-51-25

bbatsov: It's good.

lib/rubocop/config_loader.rb
```diff
@@ -220,6 +222,22 @@ def check_duplication(yaml_code, absolute_path)
         end
       end
 
+      def check_enabled_value(hash, parent = nil)
```
bbatsov: I guess we should change the name of this method to something more generic - e.g. `check/verify_cop_config_values` or something like this.

bbatsov:
You'll also have to rebase on top of the current `master`.

lib/rubocop/config_loader.rb
unasuke: Renamed it! :heart_eyes:

unasuke:
I rebased this. :ok_man:

lib/rubocop/config_loader.rb
```diff
@@ -220,6 +222,22 @@ def check_duplication(yaml_code, absolute_path)
...
       end
 
+      def check_cop_config_value(hash, parent = nil)
+        hash.each do |key, value|
+          check_cop_config_value(value, key) if value.is_a?(Hash)
+
+          next unless %w[Enabled
+                         Safe
+                         SafeAutoCorrect
+                         AutoCorrect].include?(key) && value.class == String
```
koic: I think it is better to use `is_a?(String)`  instead of `== String`. Do you have any intentions?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New features
 
 * [#7274](https://github.com/rubocop-hq/rubocop/issues/7274): Add new `Lint/SendWithMixinArgument` cop. ([@koic][])
+* [#7272](https://github.com/rubocop-hq/rubocop/pull/7272): Show warning message if passed string to 'Enabled' key in .rubocop.yml. ([@unasuke][])
```
koic: Since other than `Enabled` are supported, therefore this changelog entry may be updated.

lib/rubocop/config_loader.rb
unasuke: You're right.  It's just my habit, not deep intention. :pray:

CHANGELOG.md
unasuke: Sure. That makes sense.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New features
 
 * [#7274](https://github.com/rubocop-hq/rubocop/issues/7274): Add new `Lint/SendWithMixinArgument` cop. ([@koic][])
+* [#7272](https://github.com/rubocop-hq/rubocop/pull/7272): Show warning message if passed string to `Enabled`, `Safe`, `SafeAutocorrect` and `AutoCorrect` keys in .rubocop.yml. ([@unasuke][])
```
koic: This is a trivial comment. The following comma is forgotten.

```diff
-`Enabled`, `Safe`, `SafeAutocorrect` and `AutoCorrect`
+`Enabled`, `Safe`, `SafeAutocorrect`, and `AutoCorrect`
```

unasuke: I did't know that grammar! Thanks!

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New features
 
 * [#7274](https://github.com/rubocop-hq/rubocop/issues/7274): Add new `Lint/SendWithMixinArgument` cop. ([@koic][])
+* [#7272](https://github.com/rubocop-hq/rubocop/pull/7272): Show warning message if passed string to `Enabled`, `Safe`, `SafeAutocorrect`, and `AutoCorrect` keys in .rubocop.yml. ([@unasuke][])
```
koic: Ah, This URL seems to be https://github.com/rubocop-hq/rubocop/issues/7056.

bbatsov:
Thanks!

CHANGELOG.md
unasuke: Oh...

unasuke: Shoud I have to change the url to https://github.com/rubocop-hq/rubocop/issues/7056 ? (But already merged it... Sorry too late.)

PR #7278, opened 2019-08-11
[Fix #7275] Fix a false negative for `Style/VariableName`

### LINKED ISSUE #7275: Style/VariableName doesn't work for method arguments
The `Style/VariableName` rule isn't catching bad variable names (e.g. camelCase when snake_case is configured) for variables passed to methods as arguments.  

It also doesn't catch bad `let` variable definitions, thought I think that is understandable, but would be a great feature.

The problem then is that in a lot of specs we define the variable using `let` and then it gets passed to a method when it's used, but this rule isn't catching either of those scenarios, leaving a big blind spot. 

It does catch bad parameters in a method definition, and bad variable assignments.

Tested with Rubocop 0.74.0

## Expected behavior

When configured to use `snake_case`, the `Style/VariableName` rule should catch usage such as:
```
my_method(badVariableName, anotherBadVariableName)
```

## Actual behavior

No Rubocop violations are found.

```
$ rubocop test.rb
Inspecting 1 file
.

```

test.rb
```
# frozen_string_literal: true

let(:badVariable) { true }

def print_name(user_name)
  puts user_name
end

print_name(badVariable)
```
`rubocop test.rb` (probably need RSpec for `let`)

## RuboCop version

```
$ rubocop -V
0.74.0 (using Parser 2.6.3.0, running on ruby 2.3.7 universal.x86_64-darwin18)
```

koic:
Fixes #7275.

This PR makes `Style/VariableName` aware of argument names when invoking a method.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 * [#7252](https://github.com/rubocop-hq/rubocop/issues/7252): Prevent infinite loops by making `Layout/SpaceInsideStringInterpolation` skip over interpolations that start or end with a line break. ([@buehmann][])
 * [#7262](https://github.com/rubocop-hq/rubocop/issues/7262): `Lint/FormatParameterMismatch` did not recognize named format sequences like `%.2<name>f` where the name appears after some modifiers. ([@buehmann][])
 
+### Changes
+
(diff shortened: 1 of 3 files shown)


bbatsov:
Nice work!

PR #7279, opened 2019-08-11
[Fix #7274] Add new `Lint/SendWithMixinArgument` cop

### LINKED ISSUE #7274: Don't use `#send` etc. to include or prepend modules
Back in the day (before Ruby 2.1), `Module.include` and `Module.prepend` were private methods. So when people authored gems / patched monkeys, they wrote e.g. `Foo.send(:include, MyPatch)`. In our day and age, I think we should encourage everybody to just write `Foo.include(MyPatch)`.

Recommend calling `Foo.include(Bar)` over `Foo.send(:include, Bar)`, `Foo.public_send(:include, M)`, etc.

Recommend calling `Foo.prepend(Bar)` over `Foo.send(:prepend, Bar)`, `Foo.public_send(:prepend, M)`, etc.

None

## Additional context

None

koic:
Resolves #7274.

This PR adds new `Lint/SendWithMixinArgument` cop.

This cop checks for `send`, `public_send`, and `__send__` methods when using mix-in.

```ruby
# bad
Foo.send(:include, Bar)
Foo.send(:prepend, Bar)
Foo.send(:extend, Bar)

# bad
Foo.public_send(:include, Bar)
Foo.public_send(:prepend, Bar)
Foo.public_send(:extend, Bar)

# bad
Foo.__send__(:include, Bar)
Foo.__send__(:prepend, Bar)
Foo.__send__(:extend, Bar)

# good
Foo.include Bar
Foo.prepend Bar
Foo.extend Bar
```

`include` and `prepend` methods were private methods until Ruby 2.0, they were mixed-in via `send` method. This cop uses Ruby 2.1 or higher style that can be called by public methods.

```console
% ruby -v
ruby 2.0.0p648 (2015-12-16 revision 53162) [x86_64-darwin13.0.2]
% ruby -e 'module M end; Object.include(M)'
-e:1:in `<main>': private method `include' called for Object:Class (NoMethodError)
% ruby -e 'module M end; Object.prepend(M)'
-e:1:in `<main>': private method `prepend' called for Object:Class (NoMethodError)

% ruby -v
ruby 2.1.10p492 (2016-04-01 revision 54464) [x86_64-darwin13.0]
% ruby -e 'module M end; Object.include(M)' # no error
% ruby -e 'module M end; Object.prepend(M)' # no error
```

And `extend` method that was originally a public method is also targeted for style unification.

```console
% ruby -v
ruby 2.0.0p648 (2015-12-16 revision 53162) [x86_64-darwin13.0.2]
% ruby -e 'module M end; Object.extend(M)' # no error

% ruby -v
ruby 2.1.10p492 (2016-04-01 revision 54464) [x86_64-darwin13.0]
% ruby -e 'module M end; Object.extend(M)' # no error
```

-----------------

* [x] Added tests.


bbatsov:
The code looks great, but I think we should make this a lint cop.

koic:
Thanks for the review. I moved this cop to `Lint` department.

lib/rubocop/cop/lint/send_with_mixin_argument.rb
```diff
@@ -0,0 +1,86 @@
...
+              $(const _ _))
+        PATTERN
+
+        def on_send(node)
+          send_with_mixin_argument?(node) do |method, module_name|
+            message = message(
+              method, module_name.source, bad_location(node).source
+            )
+
+            add_offense(node, location: bad_location(node), message: message)
```
buehmann: This is in no way restricted to the mixin methods mentioned above (`%i[include prepend extend]`). Is this intended? Might result in false positives.

koic: Sure! I fixed a false positives. Thanks.

bbatsov:
Thanks!

bquorning:
Awesome 🎉

PR #7283, opened 2019-08-13
Convert some tests to use expect_correction

rrosenblum:
This covers the easy to convert tests from metrics, security, and lint. There will be more changes to come.


jonas054:
@rrosenblum The changes look good but the build is failing.

bbatsov:
Thanks!

PR #7293, opened 2019-08-18
Fix ConditionalAssignment auto-correct

jonas054:
Handle inner conditional inside else.
    
An inner conditional such as
    
```ruby
  if cond1
    a = 1
  else
    if cond2
      a = 2
    end
  end
```
    
was not correctly handled. It should not be treated as an `elsif`, because it is an `else` and there's special logic for `else` branches.


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
```
bbatsov: It seems you misplaced this section.

jonas054: Ah yes. Fixed now.

bbatsov:
Thanks!

PR #7296, opened 2019-08-19
Recognize rails/web-console debug calls in Lint/Debugger

gsamokovarov:
Rails' [Web Console](https://github.com/rails/web-console) has been around for a while now and I like debugging my applications with it. I recently shipped a `console` call in production (this is what spawns it). While the `console` call is not a security vulnerability, it results in a `NoMethodError` as Web Console is not available in production, I would have loved to have this caught by a linter. 

Do you think it has a place in `Lint/Debugger`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#7296](https://github.com/rubocop-hq/rubocop/issues/7296): Recognize `console` and `binding.console` ([rails/web-console](https://github.com/rails/web-console)) calls in `Lint/Debuggers`. ([@gsamokovarov][])
 * [#7567](https://github.com/rubocop-hq/rubocop/pull/7567): Introduce new `pending` status for new cops. ([@Darhazer][], [@pirj][])
 
 ### Bug fixes
--- a/lib/rubocop/cop/lint/debugger.rb
+++ b/lib/rubocop/cop/lint/debugger.rb
@@ -43,9 +43,9 @@ module RuboCop
         PATTERN
 
         def_node_matcher :debugger_call?, <<~PATTERN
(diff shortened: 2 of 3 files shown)


bbatsov:
My only concern is that the name of the method is a bit generic and might generate false positives in some TTY libraries. :-)

bbatsov:
@gsamokovarov ping :-)

gsamokovarov:
Bozhidar, sorry for the delays. I have recognized both `console` and `binding.console` debugging calls here and tweaked the changelog as the time passed by. 😅

bbatsov:
Thanks!

bquorning:
> My only concern is that the name of the method is a bit generic and might generate false positives in some TTY libraries. :-)

Rails itself has a `#console` method: https://api.rubyonrails.org/classes/Rails/Application.html#method-i-console

serggl:
This change introduce a false-positive in Rails railties which has a [console](https://api.rubyonrails.org/classes/Rails/Railtie.html#method-c-console) hook (executed once `rails c` starts).
So this code is now raises an offence:
```
require 'rails'

class RavenConsole < ::Rails::Railtie
  console do
    if Rails.env.production? || Rails.env.staging?
      Raven.configuration.environments << Rails.env.to_sym
      puts 'Sentry capture was disabled for this console session'
    end
  end
end
```

IMO `console` keyword capture should be removed from this cop

bbatsov:
@gsamokovarov ^^

gsamokovarov:
I'll look into this, I may leave only `binding.console` as a debugging call.

gsamokovarov:
Dropped `console` in https://github.com/rubocop-hq/rubocop/pull/7636.

PR #7302, opened 2019-08-21
Added description for RescueModifier cop explaining reasons for not to use modifier form rescue

### LINKED ISSUE #7289: RescueModifier should explain why
https://github.com/rubocop-hq/rubocop/blob/master/manual/cops_style.md#stylerescuemodifier

RescueModifier doesn't tell why the rescue modifier shouldn't be used.
But I guess this is the reason!?
https://thoughtbot.com/blog/don-t-inline-rescue-in-ruby

Please put that into the cop description!
Else people might think it's just "bad style" to use the rescue modified.
And they won't get, that this cop actually warns them about a syntax most people get wrong!

https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/style/rescue_modifier.rb
E.g.:
```
      # This cop checks for uses of rescue in its modifier form. The rescue
      # modifier is a problem, because it's behavior is very misleading.
      #
      # @example
      #   # bad
      #   some_method rescue handle_error
      #
      #   # good
      #   begin
      #     some_method
      #   rescue
      #     handle_error
      #   end
      #
      # If given an exception class as parameter, the rescue modifier will
      # return that class but not filter for it.
      #
      # @example
      #   # bad - won't work the way it looks
      #   some_method rescue SomeException
      #
      #   # bad - way it actually works out
      #   begin
      #     some_method
      #   rescue StandardError
      #     SomeException
      #   end
      #
      #   # good - what you probably really wanted
      #   begin
      #     some_method
      #   rescue SomeException
      #     nil
      #   end
      #
      # Also a newline after the rescue modifier will make it look like the next
      # line is always executed, which isn't the case.
      #
      # @example
      #   # bad
      #   some_method rescue
      #   puts 'executed on rescue'
      #
      #   # good
      #   begin
      #     some_method
      #   rescue StandardError
      #     puts 'executed on rescue'
      #   end
```
I'm not sure about your conventions how to write documentation. So I'm not sending this as a pull request.

abhaynikam:
fixes #7289 

Adds a description to the cop specifying the reason for not using modifier form rescue. The reasons are similar to the ones mentioned in the https://thoughtbot.com/blog/don-t-inline-rescue-in-ruby blog.

### DIFF
--- a/lib/rubocop/cop/style/rescue_modifier.rb
+++ b/lib/rubocop/cop/style/rescue_modifier.rb
@@ -5,14 +5,38 @@ module RuboCop
     module Style
       # This cop checks for uses of rescue in its modifier form.
       #
+      # The cop to check `rescue` in its modifier form is added for following
+      # reasons:
+      #
+      # * The syntax of modifier form `rescue` can be misleading because it
+      #   might led us to believe that `rescue` handles the given exception
+      #   but it acutally rescue all exceptions to return the given rescue
+      #   block. In this case, value returned by handle_error or
+      #   SomeException.
+      #
+      # * Modifier form `rescue` would rescue all the exceptions. It would
+      #   silently skip all exception or errors and handle the error.
+      #   Example: If `NoMethodError` is raised, modifier form rescue would
+      #   handle the exception.
+      #
       # @example
       #   # bad
       #   some_method rescue handle_error
+      #
+      #   # bad
+      #   some_method rescue SomeException
       #
       #   # good
       #   begin
       #     some_method
       #   rescue
+      #     handle_error
+      #   end
+      #
+      #   # good
+      #   begin
+      #     some_method
+      #   rescue SomeException
       #     handle_error
       #   end
       class RescueModifier < Cop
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/style/rescue_modifier.rb
```diff
@@ -5,16 +5,40 @@ module Cop
...
+      #   silently skip all exception or errors and handle the error.
+      #   Example: If `NoMethodError` is raised, modifier form rescue would
+      #   handle the exception.
+      #
       # @example
       #   # bad
       #   some_method rescue handle_error
       #
+      #   # bad
+      #   some_method rescue SomeException
```
bbatsov: We should have an extra lint cop for this down the road.

bbatsov:
Thanks!

Btw, it'd be nice to improve the matching rule in the style guide as well.

lib/rubocop/cop/style/rescue_modifier.rb
abhaynikam: I'll try work on it this week. 😊

PR #7307, opened 2019-08-23
[Fix #7287] Auto-correction of FrozenStringLiteralComment is unsafe

### LINKED ISSUE #7287: Style/FrozenStringLiteralComment should not be considered as safe for autocorrection
Hi,
I think `Style/FrozenStringLiteralComment` should not be considered as safe for autocorrection.

For example:

```
my_string = 'foo'
my_string << 'will crash with frozen_string_literal'
```

This code will crash with an error `can't modify frozen String (FrozenError)`.

Regards,
John

buehmann:
Auto-correction of `Style/FrozenStringLiteralComment` changes the state of
string objects in a program and may thus break it.

This closes #7287.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2758,7 +2758,7 @@ Style/FrozenStringLiteralComment:
                  to help transition to frozen string literals by default.
   Enabled: true
   VersionAdded: '0.36'
-  VersionChanged: '0.69'
+  VersionChanged: '0.79'
   EnforcedStyle: always
   SupportedStyles:
     # `always` will always add the frozen string literal comment to a file
@@ -2769,6 +2769,7 @@ Style/FrozenStringLiteralComment:
     # `never` will enforce that the frozen string literal comment does not
     # exist in a file.
     - never
+  Safe: false
 
 Style/GlobalVars:
   Description: 'Do not introduce global variables.'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@
 * [#7595](https://github.com/rubocop-hq/rubocop/issues/7595): Make `Style/NumericPredicate` aware of ignored methods when specifying ignored methods. ([@koic][])
 * [#7607](https://github.com/rubocop-hq/rubocop/issues/7607): Fix `Style/FrozenStringLiteralComment` infinite loop when magic comments are newline-separated. ([@pirj][])
 
+### Changes
+
+* [#7287](https://github.com/rubocop-hq/rubocop/issues/7287): `Style/FrozenStringLiteralComment` is now considered unsafe. ([@buehmann][])
+
 ## 0.78.0 (2019-12-18)
 
 ### New features
--- a/manual/cops_style.md
+++ b/manual/cops_style.md
@@ -2271,7 +2271,7 @@ EnforcedStyle | `annotated` | `annotated`, `template`, `unannotated`
 
 Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
 --- | --- | --- | --- | ---
-Enabled | Yes | Yes  | 0.36 | 0.69
+Enabled | No | Yes  | 0.36 | 0.79
 
 This cop is designed to help you transition from mutable string literals
 to frozen string literals.


koic:
IMHO, `SafeAutoCorrect: false` may not be a bad choice. For example, it causes a breaking change when a string passed from a dependent gem changes from a mutable string to a frozen string.

However, it may be better to warn users to use `String#dup` or `String#+@` for destructive operations on frozen string literals. AFAIK, matz has not given up the possibility of defaulting to a frozen string in the future. In Ruby 3.0, matz was abandant for changing it because there were not enough tools to prevent destructive changes (I heard that at RubyConf Taiwan 2019).

Therefore, I couldn't conclude and was pending this configuration.

bbatsov:
> However, it may be better to warn users to use String#dup or String#+@ for destructive operations on frozen string literals. 

I guess that should be a different cop, thought, right?

config/default.yml
```diff
@@ -2730,6 +2730,7 @@ Style/FrozenStringLiteralComment:
     # `never` will enforce that the frozen string literal comment does not
     # exist in a file.
     - never
+  SafeAutoCorrect: false
```
bbatsov: I think the cop is actually unsafe (as it doesn't actually check the source to figure out if it's ok to add the magic comment). Unsafe for a cop means that it yields false positives or the suggestions can change the meaning of the code. Unsafe for auto-correct means the cop the cop suggestions are safe, but the auto-correct can't apply those reliably.

koic:
> I guess that should be a different cop, though, right?

Yes. I think it should be a different cop's responsibility.

config/default.yml
koic: I get it! That makes sense to me. Thank for the explanation.

buehmann: > I think the cop is actually unsafe

Makes sense. Should we add `Safe: false` then as well?

Having looked into how `--safe` and `--safe-auto-correct` work I am a bit confused, though:

Is it correct that currently `rubocop --safe-auto-correct` applies "safe" autocorrections of unsafe cops (i.e. it also fixes false positives reliably)? And that you need to use `rubocop --safe-auto-correct --safe` to really not make any changes to the meaning of the code? 😕

bbatsov: > Is it correct that currently rubocop --safe-auto-correct applies "safe" autocorrections of unsafe cops (i.e. it also fixes false positives reliably)?

There are no such cases - if a cop is unsafe, its auto-correct is unsafe as well. 

> And that you need to use rubocop --safe-auto-correct --safe to really not make any changes to the meaning of the code?

Yep - basically you get the safe report, but not all of those offences would have safe fixes, so you need to specify this further. I guess that's somewhat confusing, but we couldn't come up with a better way to express it.

bbatsov: > Makes sense. Should we add Safe: false then as well?

Only it - this will affect the auto-correct as well.

buehmann: > Only it - this will affect the auto-correct as well.

Are you sure? I tried something along these lines this morning with a cop that only has `Safe: false`:
https://github.com/rubocop-hq/rubocop/blob/8d071f0c9a32f594b696cb7a3b6a0e5adabad42b/config/default.yml#L3344-L3354
```sh-session
$ cat example.rb          
# frozen_string_literal: true

# thief is not a Hash
thief.has_key?(:foo)

$ rubocop -f c example.rb 
example.rb:4:7: C: Style/PreferredHashMethods: Use Hash#key? instead of Hash#has_key?.
thief.has_key?(:foo)
      ^^^^^^^^

$ rubocop -f c --safe example.rb

$ rubocop -f c --safe-auto-correct example.rb
example.rb:4:7: C: [Corrected] Style/PreferredHashMethods: Use Hash#key? instead of Hash#has_key?.
thief.has_key?(:foo)
      ^^^^^^^^

```
Here the (wrong) autocorrect is done although the cop is safe.

bbatsov: `PreferredHashMethod` is inherently unsafe, as you can never be certain if the receiver is a hash or not. I guess it's simply not marked as unsafe.

buehmann: That's my point. The cop _is_ marked as unsafe, but its auto-correction is not marked as unsafe. You claimed above that setting `Safe: false` for a cop implies `SafeAutoCorrect: false` (if I understood you correctly). My example shows that this is not true.

So we either have a bug here or I still have a misconception how these two levels of safeness are supposed to work.

bbatsov: Or I do. 😆 Since I don't spend as much time on the code as I used to, I'm starting to forget some of my original reasoning. But in general I'm struggling to see how the cop's suggestions can be unsafe, but it corrections can be somehow safe. Safe always meant preserving the semantics on the code intact.

bbatsov:
@buehmann It'd be nice if we wrapped this PR at some point. Going back to https://github.com/rubocop-hq/rubocop/pull/7307/files#r318753133 - it seems to me that marking some cop as `unsafe` should imply that it's auto-correct is unsafe as well.

buehmann:
Thanks for the reminder. Then RuboCop does not behave as expected at the moment. I will probably create a separate issue for that and propose a fix. After that we can come back to this one.

buehmann:
@bbatsov Now that we have #7611 I changed this PR to declare the cop as `Safe: false` (only).

bbatsov:
Thanks!

config/default.yml
marcandre: > I think the cop is actually unsafe (as it doesn't actually check the source to figure out if it's ok to add the magic comment). Unsafe for a cop means that it yields false positives or the suggestions can change the meaning of the code. Unsafe for auto-correct means the cop the cop suggestions are safe, but the auto-correct can't apply those reliably.

I would propose a different definition.  Unsafe for a cop means that it yields false positives *and that's it*. (e.g. "don't use `dig` with single argument is unsafe, as it could be a custom method in a gem dependency and thus perfectly o.k. and without another alternative). The FrozenStringLiteral cop is a good example where the fact that we can't reliably autocorrect a source doesn't mean it can't or shouldn't be done. Since *any source* can modified to have a frozen string literal comment, that all sources should, and that the cop will always correctly detect if there is a frozen string literal comment, I believe this cop is Safe, only its auto-correction is unsafe. How does that sound @bbatsov ?

marcandre: ping @bbatsov

PR #7308, opened 2019-08-24
[Fix #3534] Report long modifier lines in IfUnlessModifier

### LINKED ISSUE #3534: Style/IfUnlessModifier not reporting lines longer than MaxLineLength
RuboCop auto-corrects `if/unless`-statements to use the modifier form if the line length after the correction will be less than `MaxLineLength`, but it doesn't actually work in the opposite direction, i.e. if the line is already in modifier form, but is longer than `MaxLineLength`.

---
## Expected behavior

Lines that are using the modifier form which are longer than `MaxLineLength` are reported as offenses, and auto-corrected if specified.
## Actual behavior

Lines that are using the modifier form which are longer than `MaxLineLength` are not reported as offenses, and hence, not auto-corrected.

Create a file which has line longer than `MaxLineLength` using the modifier form.
## RuboCop version

```
$ rubocop -V
0.43.0 (using Parser 2.3.1.4, running on ruby 2.3.1 x86_64-darwin15)
```

jonas054:
Make `Style/IfUnlessModifier` work in two directions. Apart from the already existing reporting of `if`/`unless` expressions on normal form that could be changed to modifier form, we add reporting and correcting of modifier form `if`/`unless` expressions that make the line too long and should be written on normal form.

It could be argued that `Metrics/LineLength` should be responsible for reporting and correcting lines that are too long, but I think it's easier to put this logic in `Style/IfUnlessModifier`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 ### Changes
 
 * [#7275](https://github.com/rubocop-hq/rubocop/issues/7275): Make `Style/VariableName` aware argument names when invoking a method. ([@koic][])
+* [#3534](https://github.com/rubocop-hq/rubocop/issues/3534): Make `Style/IfUnlessModifier` report and auto-correct modifier lines that are too long. ([@jonas054][])
 
 ## 0.74.0 (2019-07-31)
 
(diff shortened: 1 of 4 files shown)


bbatsov:
I agree. Nicely done!

Drenmi:
There seem to be cases where this cop now goes outside its jurisdiction. In particular, if I have already disabled `Metrics/LineLength` for a certain line, or block, this cop doesn't care, and registers an offense anyway. 😅

jonas054:
That's not good. I'll try to fix it.

PR #7309, opened 2019-08-24
Handle assignment to shared object in Lint/UselessSetterCall

jonas054:
When a local variable is assigned to a shared object, the referenced object is effectively no longer local, so calling a setter method on it might very well be useful and should not be reported as an offense.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * [#7290](https://github.com/rubocop-hq/rubocop/issues/7290): Handle inner conditional inside `else` in `Style/ConditionalAssignment`. ([@jonas054][])
 * [#5788](https://github.com/rubocop-hq/rubocop/issues/5788): Allow block arguments on separate lines if line would be too long in `Layout/MultilineBlockLayout`. ([@jonas054][])
 * [#7305](https://github.com/rubocop-hq/rubocop/issues/7305): Register `Style/BlockDelimiters` offense when block result is assigned to an attribute. ([@mvz][])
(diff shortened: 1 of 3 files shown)


bbatsov:
What's the high level idea to discern a non-local object from a local one? Looking at the tests - those examples could have just as easily been non-shared objects as well.

jonas054:
The idea is that setting something inside a method-local object is useless _but_ if the object can be accessed from outside the method, then the setter call might be useful and shouldn't be reported.

> those examples could have just as easily been non-shared objects as well

We can only detect that that `some_lvar` refers to an object that's accessible from outside the method. Whether the object is actually being accessed is unknown, so there could be false negatives. I'm not sure if that's what you meant, @bbatsov.

jonas054:
@bbatsov Did that answer your question? I tried to clarify, but maybe I'm missing something.

bbatsov:
> We can only detect that some_lvar refers to an object that's accessible from outside the method. 

I don't think we can actually detect this - we can only assume that it's an object that's accessible outside the method. After all you can't know if a method returns some shared object or something else, right?

jonas054:
I see what you mean now, and you're right. We're can't know if it's a shared object. But it _could_ be and we want to err on the side of caution and avoid false positives at all costs.

So I want to keep the logic as it is but maybe clarify in the code about the _possibly_ shared objects?

bbatsov:
I honestly don't know. I'm leaning towards just ignoring that problem and leaving the code as it is. I don't think that many people ran into that particular issue anyways. Perhaps we should just document the shortcomings of the cop better?

PR #7310, opened 2019-08-26
[Fix #7261] Do not insert empty lines in FrozenStringLiteralComment

### LINKED ISSUE #7261: Autocorrect reacts to EmptyLineAfterMagicComment config
## Basic Problem

When running `rubocop -a` to fix auto-correctable offenses, `# frozen_string_literal: true` and  a blank line are added to the top of every file.  It would be desirable for this newline not to be present in the case of `Layout/EmptyLineAfterMagicComment` being disabled.

## Solution

When running `rubocop -a`, do not place an empty line after `# frozen_string_literal: true` if `Layout/EmptyLineAfterMagicComment` is disabled for the given file or altogether.

buehmann:
`FrozenStringLiteralComment` should only be concerned about inserting/removing the magic comment as necessary but not about its spacing. So it should never insert empty lines.

The spacing is the responsibility of `EmptyLineAfterMagicComment`.

This closes #7261.

The change itself was pretty straightforward but I was surprised how much impact it had on other specs, especially on `cli_autocorrect_spec.rb` although the examples there mostly did not care at all about how `FrozenStringLiteralComment` works. By adding a lot of `--only`s I tried to restrict these examples to the cops they really need (if possible) such that they will not be affected by similar changes in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 
 * [#7275](https://github.com/rubocop-hq/rubocop/issues/7275): Make `Style/VariableName` aware argument names when invoking a method. ([@koic][])
 * [#3534](https://github.com/rubocop-hq/rubocop/issues/3534): Make `Style/IfUnlessModifier` report and auto-correct modifier lines that are too long. ([@jonas054][])
+* [#7261](https://github.com/rubocop-hq/rubocop/issues/7261): `Style/FrozenStringLiteralComment` no longer inserts an empty line after the comment. This is left to `Layout/EmptyLineAfterMagicComment`. ([@buehmann][])
 
 ## 0.74.0 (2019-07-31)
 
--- a/lib/rubocop/cop/style/frozen_string_literal_comment.rb
+++ b/lib/rubocop/cop/style/frozen_string_literal_comment.rb
@@ -121,42 +121,25 @@ module RuboCop
         end
 
         def insert_comment(corrector)
-          last_special_comment = last_special_comment(processed_source)
-          if last_special_comment.nil?
-            corrector.insert_before(correction_range, preceding_comment)
+          comment = last_special_comment(processed_source)
+
+          if comment
+            corrector.insert_after(line_range(comment.line), following_comment)
           else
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,10 @@
 
 * [#7275](https://github.com/rubocop-hq/rubocop/issues/7275): Make `Style/VariableName` aware argument names when invoking a method. ([@koic][])
 
+### Changes
+
+* [#7261](https://github.com/rubocop-hq/rubocop/issues/7261): `Style/FrozenStringLiteralComment` no longer inserts an empty line after the comment. This is left to `Layout/EmptyLineAfterMagicComment`. ([@buehmann][])
+
```
buehmann: Oh, I have to move this up.

lib/rubocop/cop/style/frozen_string_literal_comment.rb
```diff
@@ -122,41 +122,22 @@ def remove_comment(corrector, node)
 
         def insert_comment(corrector)
```
Drenmi: Since the value of `line` here depends on whether `last_special_comment` is `nil` or not, I think it'd be easier to follow what's going on if that was lifted into the conditional, i.e.:

```ruby
last_special_comment = last_special_comment(processed_source)

if last_special_comment.nil?
  line_range = processed_source.buffer.line_range(1)

  corrector.insert_before(line_range, preceding_comment)
else
  line_range = processed_source.buffer.line_range(last_special_comment.line)

  corrector.insert_after(line_range, following_comment)
end
```

I would personally rather have this piece of duplication and make the conditional cases clearer. WDYT?

buehmann: You're right: That piece of code can be improved. I just pushed a variant of your second suggestion.

bbatsov:
The changes look ok to me. I'll wait until you agree with @Drenmi on the final cleanup of the code.

bbatsov:
@buehmann Can you rebase?

PR #7311, opened 2019-08-27
Implement Style/UnlessWithOperator cop

ChrisBr:
Using `unless` with a conditional including an operator makes it generally harder to read because the operator also gets negated. This cop looks for *unless* with conditionals including operators.

```ruby
 # bad
 return false unless foo && bar

 # good
 return false unless foo
 return false unless bar
```


lib/rubocop/cop/style/unless_with_operator.rb
```diff
@@ -0,0 +1,28 @@
...
+      #
+      # @example
+      #   # bad
+      #   return false unless foo && bar
+      #
+      #   # good
+      #   return false unless foo
+      #   return false unless bar
+      class UnlessWithOperator < Cop
+        MSG = 'Do not use `unless` with operator. Split this in several lines.'
```
lewispb:
```suggestion
        MSG = 'Do not use `unless` with operator. Split this over several lines.'
```

lib/rubocop/cop/style/unless_with_operator.rb
```diff
@@ -0,0 +1,28 @@
...
+
+        def on_if(node)
+          return unless node.unless?
+          return unless node.condition.respond_to?(:operator)
```
lewispb: Practice what you preach 😄

spec/rubocop/cop/style/unless_with_operator_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+RSpec.describe RuboCop::Cop::Style::UnlessWithOperator do
+  subject(:cop) { described_class.new(config) }
+
+  let(:config) { RuboCop::Config.new }
+
+  it 'registers an offense when using operator in unless conditional' do
+    expect_offense(<<~RUBY)
+      return false unless false || true
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not use `unless` with operator. Split this in several lines.
```
lewispb:
```suggestion
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not use `unless` with operator. Split this over several lines.
```

lib/rubocop/cop/style/unless_with_operator.rb
davidstosik: How about moving the condition to an intent explicit method?
```rb
def on_if(node)
  add_offense(node) if unless_with_operator?(node)
end

private

def unless_with_operator?(node)
  node.unless? && node.condition.respond_to?(:operator)
end
```

Drenmi:
This cop seems a bit intense. I find it hard to come up with examples where the conditional becomes unclear.

For example:

```ruby
go_to_work unless weekend? || public_holiday?
```

or

```ruby
go_for_activity unless activity.outdoors? && weather.raining?
```

sound perfectly natural. 🤔 

The tricky ones I come across usually involve using `unless` together with negation (prefix `!`).

Perhaps you can provide some examples where splitting the conditional up would be helpful?

bbatsov:
Yep, I agree. This is quite subjective and nuanced and for the cop to be accepted we first need to agree on whether it solves a real problem.

ChrisBr:
Personally I still think `unless` with several boolean operators are hard to swallow and usually try to rewrite it to an `if` or extract to a function.

Anyway, I think it really becomes problematic when mixing up with negation. In this example with `present?` or just checking for `nil`.

```ruby
## Bad
unless client_id.present? && client_secret.present?
  raise "Missing credentials for OauthApplication #{name}."
end

unless client_id && client_secret
  raise "Missing credentials for OauthApplication #{name}."
end

## Good - flip to if
if client_id.blank? || client_secret.blank?
  raise "Missing credentials for OauthApplication #{name}."
 end

## Good - extract method
if missing_credentials?
  raise "Missing credentials for OauthApplication #{name}."
 end

def missing_credentials?
  client_id.blank? || client_secret.blank?
end
 ```

Anyway, if you think this is to subjective and nuanced, please close the PR. Otherwise, happy to improve and implement feedback. And sorry for the delay.

bbatsov:
No one joined the conversation here, so I guess it's best to just close this PR. Thanks for working on it!

PR #7313, opened 2019-08-28
[Fix #7091] Extract FormatString parser

### LINKED ISSUE #7091: Style/FormatStringToken with flags/modifiers
PoC code:

```ruby
def foo
  puts format('%s %s', 'foo', 'bar')
  puts format('%-20s %s', 'foo', 'bar')
  puts format('%-20s %-30s', 'foo', 'bar')
end
```

--------

## Expected behavior

6 warnings of `Style/FormatStringToken` to prefer annotated tokens

## Actual behavior

Only the `%s` items are detected. Any flag or modifier to the format string makes them unrecognized.

See PoC

## RuboCop version

```
0.70.0 (using Parser 2.6.3.0, running on ruby 2.3.3 i386-linux-gnu)
```

buehmann:
This extracts and unifies the format string parsing from two different cops. As a consequence, `Style/FormatStringToken` is now able to inspect more variants of format sequences.

I didn't really know where to best put the `FormatString` class, so I am open for suggestions. Maybe there is a place for such auxiliary classes that I overlooked.

### DIFF
--- a/spec/rubocop/cop/lint/format_parameter_mismatch_spec.rb
+++ b/spec/rubocop/cop/lint/format_parameter_mismatch_spec.rb
@@ -197,10 +197,6 @@ RSpec.describe RuboCop::Cop::Lint::FormatParameterMismatch do
     expect_no_offenses('format("%0*x", max_width, id)')
   end
 
-  it 'accepts an extra arg for dynamic width with other following flags' do
-    expect_no_offenses('format("%*0x", max_width, id)')
-  end
-
   it 'does not register an offense argument is the result of a message send' do
     expect_no_offenses('format("%s", "a b c".gsub(" ", "_"))')
   end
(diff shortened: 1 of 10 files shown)


spec/rubocop/cop/lint/format_parameter_mismatch_spec.rb
```diff
@@ -197,10 +197,6 @@
     expect_no_offenses('format("%0*x", max_width, id)')
   end
 
-  it 'accepts an extra arg for dynamic width with other following flags' do
-    expect_no_offenses('format("%*0x", max_width, id)')
```
buehmann: It turned out that `%*0x` is not a valid format sequence. MRI does not accept it.

bbatsov:
> I didn't really know where to best put the FormatString class, so I am open for suggestions. Maybe there is a place for such auxiliary classes that I overlooked.

All of them live under `rubocop/` now, but I agree it's better to have an `Util` folder/namespace for them. I wonder how many of the existing ones can be moved without breaking something so we can clean up a bit the file structure.

PR #7314, opened 2019-08-28
Fix the warning message about which version of rubocop-performance to upgrade to

rrosenblum:
We have received a couple of questions about this in https://github.com/rubocop-hq/rubocop/pull/7249. If there is a plan to release a minor or revision update, we may want to include this.

`rubocop-performance` hasn't created their 1.5.0 release yet. Maybe it would be easier for that version to jump to 1.15.0 🤣


bbatsov:
Thanks!

PR #7318, opened 2019-08-29
[Fix #6075] Support `IgnoredPatterns` option for `Naming/MethodName`

### LINKED ISSUE #6075: Add support for ignore pattern in MethodName cop
First of all, thanks for such a lovely tool!

Generally I think Ruby methods should be in snake case. However, SketchUp's Ruby API has several interfaces that calls methods named in camel case, e.g. the [Tool interface](http://ruby.sketchup.com/Sketchup/Tool.html) and the various Observer interfaces.

It would be useful if this cop could ignore specific, pre-defined method names.

````yaml
Naming/MethodName:
  IgnoredPatterns:
    - '^\s*onSelectionBulkChange\s.*'
    - '^\s*onSelectionCleared\s.*'
````

Using RuboCop version 0.51.0.

### PR DESCRIPTION
**Pull request by koic**
Resolves #6075.

This PR supports `IgnoredPatterns` option for `Naming/MethodName`.

```yaml
Naming/MethodName:
  IgnoredPatterns:
    - '\A\s*onSelectionBulkChange\s*'
    - '\A\s*onSelectionCleared\s*'
```

Method names matching patterns are always allowed, both `EnforcedStyle: snake_case` and `EnforcedStyle: camelCase`.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1989,6 +1989,13 @@ Naming/MethodName:
   SupportedStyles:
     - snake_case
     - camelCase
+  # Method names matching patterns are always allowed.
+  #
+  #   IgnoredPatterns:
+  #     - '\A\s*onSelectionBulkChange\s*'
+  #     - '\A\s*onSelectionCleared\s*'
+  #
+  IgnoredPatterns: []
 
 Naming/PredicateName:
   Description: 'Check the names of predicate methods.'
--- a/lib/rubocop/cop/naming/method_name.rb
+++ b/lib/rubocop/cop/naming/method_name.rb
@@ -6,6 +6,15 @@ module RuboCop
       # This cop makes sure that all methods use the configured style,
       # snake_case or camelCase, for their names.
       #
+      # This cop has `IgnoredPatterns` configuration option.
+      #
+      #   Naming/MethodName:
+      #     IgnoredPatterns:
+      #       - '\A\s*onSelectionBulkChange\s*'
+      #       - '\A\s*onSelectionCleared\s*'
+      #
+      # Method names matching patterns are always allowed.
+      #
       # @example EnforcedStyle: snake_case (default)
       #   # bad
       #   def fooBar; end
@@ -21,11 +30,13 @@ module RuboCop
       #   def fooBar; end
       class MethodName < Cop
         include ConfigurableNaming
+        include IgnoredPattern
 
         MSG = 'Use %<style>s for method names.'
 
         def on_def(node)
-          return if node.operator_method?
+          return if node.operator_method? ||
+                    matches_ignored_pattern?(node.method_name)
 
           check_name(node, node.method_name, node.loc.name)
         end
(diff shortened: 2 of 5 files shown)


lib/rubocop/cop/naming/method_name.rb
```diff
@@ -25,14 +34,25 @@ class MethodName < Cop
...
+          return if node.operator_method? ||
+                    allowed_method_name?(node.method_name)
 
           check_name(node, node.method_name, node.loc.name)
         end
         alias on_defs on_def
 
         private
 
+        def allowed_method_name?(method_name)
```
buehmann: Have you considered reusing [IgnoredMethodPatterns](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/mixin/ignored_method_patterns.rb)?

buehmann:
```sh-session
$ rubocop 
The `SafeMode` option will be removed in `RuboCop` 0.76. Please update `rubocop-performance` to 1.5.0 or higher.
Warning: Naming/MethodName does not support IgnoredPatterns parameter.

Supported parameters are:

  - Enabled
  - EnforcedStyle
  - SupportedStyles
```
I think you need to add the option to the defaults:
```patch
diff --git a/config/default.yml b/config/default.yml
index 90005cecd..5fe2ec22f 100644
--- a/config/default.yml
+++ b/config/default.yml
@@ -1989,6 +1989,7 @@ Naming/MethodName:
   SupportedStyles:
     - snake_case
     - camelCase
+  IgnoredPatterns: ~
 
 Naming/PredicateName:
   Description: 'Check the names of predicate methods.'
```

lib/rubocop/cop/naming/method_name.rb
koic: Ah, that makes sense and DRY. Thank you for the suggestion!

koic:
Sure! I updated this PR.

config/default.yml
```diff
@@ -1989,6 +1989,13 @@ Naming/MethodName:
   SupportedStyles:
     - snake_case
     - camelCase
+  # Method names matching patterns are always allowed.
+  #
+  #   IgnoredMethodPatterns:
+  #     - '\A\s*onSelectionBulkChange\s*'
+  #     - '\A\s*onSelectionCleared\s*'
+  #
+  IgnoredMethodPatterns: []
```
bbatsov: How do we typically name this? I want to make sure we are consistent with the name.

koic: Certainly, it can be made consistent with `IgnoredPatterns` using `IgnoredPattern` module. I first updated the option name added in this PR to `IgnoredPatterns`. 
After that I will open another PR to make the existing `IgnoredMethodPatterns` consistent with `IgnoredPatterns`.

koic: > After that I will open another PR to make the existing `IgnoredMethodPatterns` consistent with `IgnoredPatterns`.

I opened a PR #7313.

bbatsov:
Thanks!

PR #3361, opened 2016-08-02
Pass filename YAML loading to improve stack trace messages

rrosenblum:
Pass the filename to to YAML load. When parsing yaml that contains an error, this will add the filename of the yaml to the stack trace.

Example: 
Remove the colon from line 895 of the `enabled.yml`

Before the change
```
:~/work/rubocop (capture_filename u-1)$ bundle exec rubocop
(<unknown>): could not find expected ':' while scanning a simple key at line 895 column 1
~/.rbenv/versions/2.3.0/lib/ruby/2.3.0/psych.rb:377:in `parse'
~/.rbenv/versions/2.3.0/lib/ruby/2.3.0/psych.rb:377:in `parse_stream'
~/.rbenv/versions/2.3.0/lib/ruby/2.3.0/psych.rb:325:in `parse'
~/.rbenv/versions/2.3.0/lib/ruby/2.3.0/psych.rb:291:in `safe_load'
```

After the change
```
:~/work/rubocop (capture_filename u=)$ bundle exec rubocop
(~/work/rubocop/config/enabled.yml): could not find expected ':' while scanning a simple key at line 895 column 1
~/.rbenv/versions/2.3.0/lib/ruby/2.3.0/psych.rb:377:in `parse'
~/.rbenv/versions/2.3.0/lib/ruby/2.3.0/psych.rb:377:in `parse_stream'
~/.rbenv/versions/2.3.0/lib/ruby/2.3.0/psych.rb:325:in `parse'
~/.rbenv/versions/2.3.0/lib/ruby/2.3.0/psych.rb:291:in `safe_load'

```

PR #7322, opened 2019-08-29
Allow YAML aliases In Configuration

### LINKED ISSUE #7321: Allow YAML aliases in .rubocop.yml file

It is frustrating not to be able to use YAML aliasing in the `.rubocop.yml` configuration file.

Rather than putting

```yml
# .rubocop.yml

Rails/EnumUniqueness:
  some_key:
    - lots
    - of
    - values
Rails/FindBy:
  some_key:
    - lots
    - of
    - values
```
I would like to be able to say:

```yml
# .rubocop.yml

Rails/EnumUniqueness: &my_yaml_var # <- anchor
  some_key:
    - lots
    - of
    - values
Rails/FindBy:
  <<: *my_yaml_var # <- alias
```

I would like to be able to use [YAML anchors and aliases](https://yaml.org/spec/1.2/spec.html#id2765878) inside of my `.rubocop.yml` file, in order to avoid many lines of repetition. 

I have not delved too far into how this alternative would be implement but: perhaps `allow_yaml_aliases` could be a configurable option that a user or Rubocop can set.

## Additional context

YAML anchor/aliases are essentially variables that you can set in a YAML file. When one is set it is called an `anchor`, and subsequent references to the `anchor` are called `aliases`.

I have been delving back into the history of the `aliases` params set in the [`config_loader` file](https://github.com/rubocop-hq/rubocop/blob/8cfdc57c2ca20685dae4a6ea2872bb38f544ef71/lib/rubocop/config_loader.rb#L261), and there doesn't appear to have been any strong (vocalized) reason that aliases are not allowed. I am hoping that we can simply switch this configuration from `false` to `true`.

RaymondFallon:
# Allow YAML aliases In Configuration

[Fix #7321]

As described in detail in the above issue, disallowing YAML aliasing can force unnecessary repetition in the `.rubocop.yml` file. This PR sets `aliases: true` in the `YAML.safe_load` call, so that users can use YAML aliases in their `.rubocop.yml`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * [#7272](https://github.com/rubocop-hq/rubocop/pull/7272): Show warning message if passed string to `Enabled`, `Safe`, `SafeAutocorrect`, and `AutoCorrect` keys in .rubocop.yml. ([@unasuke][])
 * [#7295](https://github.com/rubocop-hq/rubocop/pull/7295): Make it possible to set `StyleGuideBaseURL` per department. ([@koic][])
 * [#7301](https://github.com/rubocop-hq/rubocop/pull/7301): Add check for calls to `remote_byebug` to `Lint/Debugger` cop. ([@riley-klingler][])
+* [#7321](https://github.com/rubocop-hq/rubocop/issues/7321): Allow YAML aliases in `.rubocop.yml`. ([@raymondfallon][])
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


koic:
As far as I investigated, the original setting to `aliases: false` was introduced in #3361.
`YAML.safe_load` prior to Psych 3.1.0 is a positional argument. In #3361, the 4th argument as `filename` of `YAML.safe_load` is probably important, and it seems to me that the 3rd argument `aliases` can be allowed.
So, this looks good to me if it is add this new feature entry to CHANGELOG.md.

Cc @rrosenblum

RaymondFallon:
Squashed! Thanks @koic

Drenmi:
Seems like a useful feature. Thanks @RaymondFallon! 🙏

koic:
@RaymondFallon Thanks for your first contribution to RuboCop core!

RaymondFallon:
Wow, you all move fast! My pleasure -- thank you for the great gem!

PR #7324, opened 2019-09-01
[Fix #4802] Detect all UnneededCopEnableDirective offenses

### LINKED ISSUE #4802: auto-gen-config should remove inline disable/enable
When generating the Rubocop todo file, the inspected files for a cop may change. If those files have inline enable or disable sentences, they should also be modified accordingly or we can end with something that is not the expected result.

--------

## Steps to reproduce the problem & Expected/Actual behavior

When having something like:

``` Ruby
# rubocop:disable  Style/NumericPredicate
assert $? == 0, "#{testfile} does not validate against #{f} -> #{testresult}"
# rubocop:enable Style/NumericPredicate
```

If when running `rubocop --auto-gen-config` if the number of offenses is higher than 15, and the cop get automatically disabled, I would expect that the inline `disable/enable` also get removed. If not when running `rubocop` again there would be offenses like: 

`Lint/UnneededDisable: Unnecessary disabling of Style/NumericPredicate`

Autocorrecting them would also not work, as it would only remove the disable part leaving the enable one, which is not what we pretend here.

And also if the # rubocop:enable Style/NumericPredicate``is there even if the cop is disabled, would the rest of the file get inspected? Because that would be also not expected.

## RuboCop version

```
$ rubocop -V
0.47.1 (using Parser 2.3.3.1, running on ruby 2.4.1 x86_64-linux-gnu)
```

jonas054:
After all other cops are finished we invoke the `Lint/UnneededCopDisableDirective` cop, since it depends on the results of the other cops. If it's an `--auto-correct` run, correcting `UnneededCopDisableDirective` offenses can introduce new `UnneededCopEnableDirective` offenses. For this reason we must do one more inspection after `UnneededCopDisableDirective` has corrected a file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * [#7290](https://github.com/rubocop-hq/rubocop/issues/7290): Handle inner conditional inside `else` in `Style/ConditionalAssignment`. ([@jonas054][])
 * [#5788](https://github.com/rubocop-hq/rubocop/issues/5788): Allow block arguments on separate lines if line would be too long in `Layout/MultilineBlockLayout`. ([@jonas054][])
 * [#7305](https://github.com/rubocop-hq/rubocop/issues/7305): Register `Style/BlockDelimiters` offense when block result is assigned to an attribute. ([@mvz][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #7325, opened 2019-09-01
[Fix #7299] Implement Lint/RaiseException cop

### LINKED ISSUE #7299: alert on `raise Exception` usage
similar to the "do not rescue Exception" cop it would make sense to not allow raising Exception since that will bypass test frameworks and error reporters

denys281:
Resolves https://github.com/rubocop-hq/rubocop/issues/7299
Should go with https://github.com/rubocop-hq/ruby-style-guide/pull/775

Cop checks for:
- `raise Exception`;
- `raise Exception, 'error message'`;
- `raise Exception.new('Error message')
-  Also it works for `fail` and `::Exception`

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/raise_exception.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        def_node_matcher :exception_new_with_message?, <<~PATTERN
+          (send nil? ${:raise :fail}
+            (send (const _ :Exception) :new ... ))
+        PATTERN
+
+        def on_send(node)
+          offense = exception?(node) || exception_new_with_message?(node)
+
+          add_offense(node, message: format(MSG, method: offense)) if offense
```
Drenmi: You could combine the matchers and do:

```ruby
raise_exception?(node) do |method|
  add_offense(node, message: format(MSG, method: method))
end
```

Drenmi:
Thanks @denys281! 🙏

lib/rubocop/cop/lint/raise_exception.rb
denys281: And how to combine them? I tried, but without success. You meant to write universal matcher or to combine two int one? 

or you meant  ? :) 

```
def raise_exception(node)
  exception?(node) || exception_new_with_message?(node)
end
```

config/default.yml
```diff
@@ -1505,6 +1505,11 @@ Lint/PercentSymbolArray:
   Enabled: true
   VersionAdded: '0.41'
 
+Lint/RaiseException:
+  Description: Checks for `raise` or `fail` statements which is raise `Exception` class.
```
bbatsov: which are raising

config/default.yml
```diff
@@ -1505,6 +1505,11 @@ Lint/PercentSymbolArray:
...
+  Enabled: true
+  VersionAdded: '0.75'
```
bbatsov: Is this in the style guide? If so - let's add a link to the rule.

lib/rubocop/cop/lint/raise_exception.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # This cop checks for `raise` or `fail` statements which is
```
bbatsov: which are raising

lib/rubocop/cop/lint/raise_exception.rb
```diff
@@ -0,0 +1,36 @@
...
+      # raise `Exception` class.
+      #
+      # @example
+      #   # bad
+      #   raise Exception, 'Error message here'
+      #
+      #   # good
+      #   raise StandardError, 'Error message here'
+      class RaiseException < Cop
+        MSG = 'Avoid `%<method>s` the `Exception` class. ' \
```
bbatsov: I'd just use "Use `StandardError` over `Exception`." as the message here.

config/default.yml
denys281: no, we have only this one https://github.com/rubocop-hq/ruby-style-guide#no-blind-rescues

bbatsov: Guess we can add something on the subject there then.

denys281: @bbatsov Did you mean  in style guide?

bbatsov: Yep.

denys281: @bbatsov  https://github.com/rubocop-hq/ruby-style-guide/pull/775

bbatsov:
Better late than never! :-) Thanks!

Issue #7298, opened 2019-08-20
False positive for Lint/UselessAccessModifier modules extending ActiveSupport::Concern


RST-J:
## Expected behavior
The `private` modifier in the `class_methods` block is not flagged as `Lint/UselessAccessModifier: Useless 'private' access modifier.`

## Actual behavior
The `private` modifier in the `class_methods` block is flagged as `Lint/UselessAccessModifier: Useless 'private' access modifier.`

## Steps to reproduce the problem
Run rubocop on a file with the following contents:
```ruby
# frozen_string_literal: true

module Foo
  extend ActiveSupport::Concern

  private

  def private_instance_bar
  end

  class_methods do
    private

    def private_class_bar
    end
  end
end
```

## RuboCop version
0.74.0 (using Parser 2.6.3.0, running on ruby 2.6.3 x86_64-linux)

Drenmi:
Is this `class_methods` convention common enough outside of Rails to warrant a special condition? I'm inclined to say yes, mostly based on the fact that I don't think there will be any false positives, i.e. `class_methods` implementations that don't honour visibility scopes.

RST-J:
Good question. Maybe this should go into `rubocop-rails`?
I wasn't really sure whether to open this issue over there or not. I think I brought it up here because `ActiveSupport` can be used without rails if one wants to.
As I have not much experience with Ruby off Rails, I cannot tell.

kliuless:
I ran into what seems like the same issue:
```ruby
module TestModule
  extend ActiveSupport::Concern

  included do
    private

    attr_writer :testvar1
  end

  class_methods do
    private

    attr_writer :testvar2
  end
end
```
The `private` inside `class_methods` is flagged as `Useless private access modifier`

halilim:
Another way to reproduce:

```rb
module Foo
  extend ActiveSupport::Concern

  included do
    private

    attr_reader :bar
  end

  private

  def private_instance_method
    # ...
  end
end
```

I can put the private instance methods into `included` as a workaround but then it causes a `Metrics/BlockLength` warning after a few methods.

kliuless:
Re: @halilim's example, sometimes instance methods should be outside of `included`, i.e. when you want them to be part of the inheritance chain. Putting them inside will add the method individually to each class or module that `include`s your module, preventing `super` from working.

stefankroes:
[...]

PR #7327, opened 2019-09-03
[Fix #7326] Fix a false positive for `Style/AccessModifierDeclarations`

### LINKED ISSUE #7326: `Style/AccessModifierDeclarations` wrong offense detected
## Expected behavior

No offense detected.

## Actual behavior

Offence detected.

code

```
class Hoge
  attr_reader :size
  attr_reader :density
  attr_reader :public

  def initialize(size:, density:, public:)
    @size = size
    @density = density
    @public = public
  end

  def perform(users)
    ::ActiveRecord::Associations::Preloader.new.preload(users, :avatar_photo, Attachment.all)

    users.each do |user|
      fulfill(user, user.decorate.avatar_image(size: size, density: density, public: public))

    end
  end
end
```

rubocop log

```
> bundle exec rubocop
hoge.rb:18:86: C: Style/AccessModifierDeclarations: public should be inlined in method definitions.
      fulfill(user, user.decorate.avatar_image(size: size, density: density, public: public))
                                                                                     ^^^^^^
```

rubocop setting

```
〜
Style/AccessModifierDeclarations:
  EnforcedStyle: inline
```

## RuboCop version

```
> bundle exec rubocop -V
0.74.0 (using Parser 2.6.4.0, running on ruby 2.5.1 x86_64-darwin18)
```

koic:
Fixes #7326.

This PR fixes a false positive for `Style/AccessModifierDeclarations` when access modifier name is used for hash literal value.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/access_modifier_declarations.rb
+++ b/lib/rubocop/cop/style/access_modifier_declarations.rb
@@ -63,6 +63,7 @@ module RuboCop
 
         def on_send(node)
           return unless node.access_modifier?
+          return if node.parent.pair_type?
 
           if offense?(node)
             add_offense(node, location: :selector) do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 * [#5788](https://github.com/rubocop-hq/rubocop/issues/5788): Allow block arguments on separate lines if line would be too long in `Layout/MultilineBlockLayout`. ([@jonas054][])
 * [#7305](https://github.com/rubocop-hq/rubocop/issues/7305): Register `Style/BlockDelimiters` offense when block result is assigned to an attribute. ([@mvz][])
 * [#4802](https://github.com/rubocop-hq/rubocop/issues/4802): Don't leave any `Lint/UnneededCopEnableDirective` offenses undetected/uncorrected. ([@jonas054][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -63,6 +63,7 @@ class AccessModifierDeclarations < Cop
 
         def on_send(node)
           return unless node.access_modifier?
+          return if node.parent.pair_type?
```
bbatsov: Seems to me the real problem here is that the cop is inspecting all `send` node instead of just going over whatever is in a `class/module` body. Your fix will do, but I think the cop's design cop be improved in general.

koic: Sure! I think it ’s a good improvement idea.
`block` may also need to be considered in addition to `class` and `module` (#7298). Since redesign may take some time, I'd like to proceed with another PR. What do you think?

bbatsov: > block may also need to be considered in addition to class and module

You're right.

> Since redesign may take some time, I'd like to proceed with another PR. What do you think?

Yeah, that's fine by me. We can merge this one in the mean time.

PR #7329, opened 2019-09-03
Fix new warnings from Ruby

jonas054:
Recent builds fail for ruby-head. A new warning that's been added in ruby-head says:

```
The last argument is used as the keyword parameter
```

This warning is emitted when a method parameter is declared with a double asterisk, e.g.,

```
def some_method(arg, **opts)
```

but a call does not use double asterisk, e.g.,

```
x = some_method(my_arg, my_opts)
```

### DIFF
--- a/lib/rubocop/config_loader_resolver.rb
+++ b/lib/rubocop/config_loader_resolver.rb
@@ -73,7 +73,7 @@ module RuboCop
 
       opts = { inherit_mode: config['inherit_mode'] || {},
                unset_nil: unset_nil }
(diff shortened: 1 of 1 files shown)


jonas054:
Looks like parser-2.6.4.0 has the same problem. That's why the ruby-head builds are still failing.

koic:
Yeah, this Ruby 2.7 new warning is a change by https://github.com/ruby/ruby/pull/2395. Thanks!

koic:
FYI, I opened a PR https://github.com/whitequark/parser/pull/607 to Parser gem.

PR #7335, opened 2019-09-08
Add Todo alias and make --disable-autocorrect mark Cops as Todo

desheikh:
We want to to differentiate between disables made intentionally due to false positives, or that we don't plan on fixing vs disables generated using the --disable-uncorrectable flag, which ideally should be reviewed at some point.

https://github.com/rubocop-hq/rubocop/pull/4355 discussed adding a todo directive as part of the disable-uncorrectable however it was never merged.

This is a slightly different approach which adds "todo" as a full alias to "disable" instead and either can be used interchangeably.
The --disable-uncorrectable flag will now generate comments using the todo directive instead of disable.

example:
```
# rubocop:todo Metrics/AbcSize
def my_method
...
end
# rubocop:enable Metrics/AbcSize
```

-----------------

* [x] Added tests.


desheikh:
@jonas054 as you added in the disable-uncommentable flag, do you have any feedback on this?

spec/rubocop/cop/style/double_cop_disable_directive_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe RuboCop::Cop::Style::DoubleCopDisableDirective do
   subject(:cop) { described_class.new }
 
-  it 'registers an offense when using `#bad_method`' do
+  it 'registers an offense when using `#bad_method` with disable' do
```
jonas054: It seems I was the one who named this example originally, but now it makes no sense to me. I don't know what `#bad_method` or `#good_method`means. :slightly_smiling_face:  Could you come up with something better?

spec/rubocop/cop/style/double_cop_disable_directive_spec.rb
```diff
@@ -15,6 +15,18 @@ def choose_move(who_to_move) # rubocop:disable Metrics/CyclomaticComplexity
     RUBY
   end
 
+  it 'registers an offense when using `#bad_method` with todo' do
+    expect_offense(<<~RUBY)
+      def choose_move(who_to_move) # rubocop:todo Metrics/CyclomaticComplexity # rubocop:todo Metrics/AbcSize # rubocop:todo Metrics/MethodLength
```
jonas054: How do we handle mixed `rubocop:disable` and `rubocop:todo`? Another example?

desheikh: Based on the current implementation of the cop we're essentially stripping all but the first cop, so that's what it would reduce to.

spec/rubocop/cop/style/double_cop_disable_directive_spec.rb
desheikh: Addressed :)

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@
 * [#7215](https://github.com/rubocop-hq/rubocop/issues/7215): Make it clear what's wrong in the message from `Style/GuardClause`. ([@jonas054][])
 * [#7245](https://github.com/rubocop-hq/rubocop/pull/7245): Make cops detect string interpolations in more contexts: inside of backticks, regular expressions, and symbols. ([@buehmann][])
 * Deprecate the `SafeMode` option. Users will need to upgrade `rubocop-performance` to 1.15.0+ before the `SafeMode` module is removed. ([@rrosenblum][])
+* [#7335](https://github.com/rubocop-hq/rubocop/pull/7335): Add todo as an alias to disable. `--disable-uncorrectable` will not disable Cops using `rubocop:todo` instead of `rubocop:disable`. ([@desheikh][])
```
jonas054: not -> now

jonas054:
I found something in the changelog, so please take a look. Could you also search through all the `.md` files under `manual` for `rubocop:disable` and see if you find more places where we should mention `rubocop:todo`? I think there are some, but right now I don't have time to list them here.

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@
 * [#7215](https://github.com/rubocop-hq/rubocop/issues/7215): Make it clear what's wrong in the message from `Style/GuardClause`. ([@jonas054][])
 * [#7245](https://github.com/rubocop-hq/rubocop/pull/7245): Make cops detect string interpolations in more contexts: inside of backticks, regular expressions, and symbols. ([@buehmann][])
 * Deprecate the `SafeMode` option. Users will need to upgrade `rubocop-performance` to 1.15.0+ before the `SafeMode` module is removed. ([@rrosenblum][])
+* [#7335](https://github.com/rubocop-hq/rubocop/pull/7335): Add todo as an alias to disable. `--disable-uncorrectable` will now disable cops using `rubocop:todo` instead of `rubocop:disable`. ([@desheikh][])
```
bbatsov: That should be moved to `master`.

manual/configuration.md
```diff
@@ -565,6 +565,14 @@ You can also disable *all* cops with
 # rubocop:enable all
 ```
 
+You can also disable cops using the todo alias.
```
bbatsov: I'd mention what this is an alias of exactly and what's the value of having it. Currently this section is a bit confusing.

lib/rubocop/cop/offense.rb
```diff
@@ -72,18 +72,23 @@ def initialize(severity, location, message, cop_name,
...
       def corrected
-        @status == :corrected
+        @status == :corrected || @status == :corrected_with_todo
       end
       alias corrected? corrected
 
+      def corrected_with_todo
+        @status == :corrected_with_todo
+      end
+      alias corrected_with_todo? corrected_with_todo
```
bbatsov: I don't get the purpose of this alias.

bbatsov:
> We want to to differentiate between disables made intentionally due to false positives, or that we don't plan on fixing vs disables generated using the --disable-uncorrectable flag, which ideally should be reviewed at some point.

This type of reasoning should be in the docs, otherwise no one will figure it out. :-)

Overall the code looks good, but I've left some small remarks here and there.

manual/configuration.md
desheikh: Updated the text

lib/rubocop/cop/offense.rb
desheikh: We want to output [Todo] instead of [Corrected] in the rubocop formatter output.
Having a different corrected state allows us to do that.

jonas054: The method above this one, `corrected`, is also aliased but I've done some digging in the history and found that there used to be an instance variable called `@corrected` with an `attr_reader :corrected` and then we needed the alias `corrected?` to get the nicely named predicate. Now that we have a `@status` instance variable, we could just define `corrected?` and `corrected_with_todo?` methods and no aliases.

desheikh: I like that, however perhaps that's more suited to being a separate PR?

bbatsov:
Thanks!

lib/rubocop/cop/offense.rb
bbatsov: OK, I'll merge this one, so you can do a follow-up PR.

desheikh:
🎉

PR #7336, opened 2019-09-08
[Fix #3591] Handle modifier if in Variable#reference!()

### LINKED ISSUE #3591: False positive from `Lint/UselessAssignment` with if/unless modifier
Variable assignments in `if` and `unless` modifiers can be regarded as useless if they are only used in the conditional expression. See the examples below.

---
## Expected behavior

The following examples should not be treated as useless assignments.
## Actual behavior

The following examples are in fact treated as useless assignments.

Put this in a file and run RuboCop on it:

``` ruby
# frozen_string_literal: true

a.to_f if (a = 123)
b.to_f if /(?<b>\d+)/ =~ '123'
```

You'll see this:

```
$ bin/rubocop -D x.rb 
Inspecting 1 file
W

Offenses:

x.rb:3:12: W: Lint/UselessAssignment: Useless assignment to variable - a.
a.to_f if (a = 123)
           ^
x.rb:4:11: W: Lint/UselessAssignment: Useless assignment to variable - b.
b.to_f if /(?<b>\d+)/ =~ '123'
          ^^^^^^^^^^^

```
## RuboCop version

```
$ rubocop -V
0.43.0 (using Parser 2.3.1.4, running on ruby 2.3.0 x86_64-linux)
```

jonas054:
Assignments in modifier if/unless conditions are special. The following program
exits with the error "undefined local variable or method `a' for main:Object"
when executed.

```
puts a if (a = @x)
```

A preceding assignment to `a` is needed to put it in scope. So there are no
useless assignments in

```
a = nil
puts a if (a = @x)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * [#7305](https://github.com/rubocop-hq/rubocop/issues/7305): Register `Style/BlockDelimiters` offense when block result is assigned to an attribute. ([@mvz][])
 * [#4802](https://github.com/rubocop-hq/rubocop/issues/4802): Don't leave any `Lint/UnneededCopEnableDirective` offenses undetected/uncorrected. ([@jonas054][])
 * [#7326](https://github.com/rubocop-hq/rubocop/issues/7326): Fix a false positive for `Style/AccessModifierDeclarations` when access modifier name is used for hash literal value. ([@koic][])
+* [#3591](https://github.com/rubocop-hq/rubocop/issues/3591): Handle modifier `if`/`unless` correctly in `Lint/UselessAssignment`. ([@jonas054][])
 
 ### Changes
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Nice catch! I wonder if that's intentional or it's a bug in Ruby...

jonas054:
It has to be intentional. Has to be. From [The Ruby Language FAQ](https://ruby-doc.org/docs/ruby-doc-bundle/FAQ/FAQ.html):

> 4.3 When does a local variable become accessible?

> Actually, the question may be better asked as: "at what point does Ruby work out that something is a variable?" The problem arises because the simple expression `a` could be either a variable or a call to a method with no parameters. To decide which is the case, Ruby looks for assignment statements. If at some point in the source prior to the use of `a` it sees it being assigned to, it decides to parse `a` as a variable, otherwise it treats it as a method.

Here's an even weirder example:

```ruby
a = nil if false # This line is needed.
puts a if (a = @x) # This is fine.
```
It's not about assignments being executed. It's all about being seen prior to use. And `a = @x` is not _prior_ to `puts a` so it doesn't help when dereferencing `a`.

PR #7350, opened 2019-09-15
[Fix #5212] Allow braces around block with do..end

### LINKED ISSUE #5212: BracesAroundHashParameters flagging necessary braces as redundant
In a Rails project, have a hash parameter with a proc as one of its values:

```ruby
accepts_nested_attributes_for :brands, {
  allow_destroy: true,
  reject_if: proc do |attributes|
    attributes.values_at(:name, :photo, :id).all?(&:blank?)
  end
}
```

Because of the block used to define the proc, the braces around the hash parameter are necessary. Without them, instantiating this model results in the following error:

```
>> foo = Account.first
ArgumentError: tried to create Proc object without a block
```

However, the `BracesAroundHashParameters` cop is flagging the braces as redundant:

```
app/models/account.rb:49:42: C: Redundant curly braces around a hash parameter.
  accepts_nested_attributes_for :brands, { ...
```

Is there another approach / recommended style for this scenario, or is this an issue with the cop?

---

Rubocop v0.51.0, Rails v5.1.4, Ruby 2.4.1p111

jonas054:
Users are getting false positives saying that hash braces should be removed although doing so would change semantics for a block with `do`..`end`. We fix the false positive by allowing the braces in this particular case.


bbatsov:
Thanks!

PR #7365, opened 2019-09-18
Fix #4082 Auto-correct empty class definitions

### LINKED ISSUE #4082: Auto-correct empty class definitions
The Ruby style guide [says](https://github.com/bbatsov/ruby-style-guide#single-line-classes):

> Prefer a single-line format for class definitions with no body.

```ruby
# bad
class FooError < StandardError
end

# okish
class FooError < StandardError; end

# good
FooError = Class.new(StandardError)
```

I'd like to have a cop that could auto-correct to the 'good' example.

We added one in GitLab, but just for error classes, because the 'bad' example is more reasonable when, for instance, you have an empty Rails controller that just uses the default methods provided.

Here's the cop we added: https://gitlab.com/gitlab-org/gitlab-ce/commit/8dd097a91530e4b047c4b391f21047c7d29d310d

This works for us, but I have no expectations that this would work for all cases 🙂 

What would we need to do to get this into RuboCop itself?

denys281:
Resolves: https://github.com/rubocop-hq/rubocop/issues/4082

As mentioned @koic in issue:

```

      # @example ExceptionClassOnly: true (default)
      #   # bad
      #   class FooError < StandardError
      #   end
      #
      #   # good
      #   class FooError < StandardError; end
      #
      #   class Foo < Bar; end
      #
      #   class Foo < Bar
      #   end
      #
      # @example ExceptionClassOnly: false
      #   # bad
      #   class FooError < StandardError
      #   end
      #
      #   class Foo < Bar
      #   end
      #
      #   # good
      #   class FooError < StandardError; end
      #
      #   class Foo < Bar; end
```

* [x] Added tests.


pirj:
Just for the reference, style link https://rubystyle.guide/#single-line-classes

bbatsov:
@denys281 Ping :-) I see the build is failing as well.

lib/rubocop/cop/style/empty_class.rb
```diff
@@ -0,0 +1,41 @@
...
+      #   class FooError < StandardError; end
+      #
+      #   # good
+      #   FooError = Class.new(StandardError)
+      #
+      class EmptyClass < Cop
+        MSG = 'Use a single-line format for class definitions with no body.'
+
+        def on_class(node)
+          return if !node.children.last.nil? || node.single_line?
```
pirj: `!node.children.last.nil?` => `node.children.any?`?

pirj: I guess we should also add an offence if `node.single_line?` in case the enforced style is `compact_constant`.

spec/rubocop/cop/style/empty_class_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    RUBY
+  end
+
+  it 'without offence for single line class' do
+    expect_no_offenses(<<~RUBY)
+      class FooError < StandardError; end
+    RUBY
+  end
+
+  it 'without offence for class with body' do
```
pirj: WDYT of "ignores class with body"?

denys281:
Will check! It seems, that work was unfinished :)

bbatsov:
@denys281 Great! Ping us when we should review the PR again.

Drenmi:
There is one edge case here.

Consider:

```ruby
class Foo
  # Note:
  #
  # Some extensive comments about Foo and why it
  # is not implemented, etc. How would we deal with
  # this if the cop were to implement auto-correct?
end
```

Is this empty or not empty? 🙂

pirj:
> There is one edge case here.

Move docs above class definition?

lib/rubocop/cop/style/empty_class.rb
```diff
@@ -0,0 +1,90 @@
...
+                          " < #{parent.source};"
+                        else
+                          ';'
+                        end
+
+          "class #{klass.source}#{middle_part} end"
+        end
+
+        def parent?(parent)
+          @parent ||= !parent&.source.nil?
```
pirj, line 73: Does it make sense to memoize this?

pirj: This can be just:
```ruby
def parent?(parent)
  parent&.source
end
```

denys281: You are right.

denys281:
> @denys281 Great! Ping us when we should review the PR again.

@bbatsov  ping :)

denys281:
@bbatsov  ping?:)

marcandre:
There's a lot of activity and @bbatsov has been a bit behind recently, but I can comment that the style guide has been changed in the meantime (sorry, and it's my "fault" 😅).

The best would be that this cop can be configured, either for preference of one-line, or two-line, or either, with default for two-line, and the same choice of settings for exception classes?

denys281:
Hm, well it need to be added some configuration options. Currently it is only one configuration regarding Exceptions only cases.

marcandre:
Right, but this cop can not be merged as is, since it currently enforces the contrary of what the style guide now says...

denys281:
Understood. So what was changed?

denys281:
Ok, I see it https://github.com/rubocop-hq/ruby-style-guide#single-line-classes. So is it worth to change this PR to match new style guide? In case yes, what requirements could be?

marcandre:
> So is it worth to change this PR to match new style guide? In case yes, what requirements could be?

Yes, it is worth it. Yes, it should match the new style guide, by default, but it would be nice to have a config.

As I said, ... "The best would be that this cop can be configured, either for preference of one-line, or two-line, or either, with default for two-line, and the same choice of settings for exception classes"

bbatsov:

PR #7372, opened 2019-09-19
[Fix #7320] support Naming/MethodName for attr_reader / attr_writer / attr_accessor

### LINKED ISSUE #7320: Naming/MethodName doesn't flag attr_reader / attr_writer / attr_accessor
## Expected behavior

[Naming/MethodName](https://github.com/rubocop-hq/rubocop/blob/master/manual/cops_naming.md#namingmethodname) should flag `attr_reader`, `attr_writer`, and `attr_accessor` for incorrect case.

## Actual behavior

Only methods defined with explicit `def` statements are flagged.

Given the following code:

```ruby
# Foo example
class Foo
  attr_accessor :fooBar
  attr_reader :barBaz
  attr_writer :bazQux

  def quxCorge
    'quxCorge'
  end

  def quxCorge=(_)
    raise 'not a thing'
  end
end
```

The `Naming/MethodName` cop should flag all five declarations -- `fooBar`, `barBaz`, `bazQux`, `quxCorge`, and `quxCorge=`

Instead, only the explicit `def quxCorge` and `def quxCorge=` are flagged:

```
Inspecting 1 file
C

Offenses:

/tmp/naming.rb:10:7: C: Naming/MethodName: Use snake_case for method names.
  def quxCorge
      ^^^^^^^^
/tmp/naming.rb:14:7: C: Naming/MethodName: Use snake_case for method names.
  def quxCorge=(_)
      ^^^^^^^^^

```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.74.0 (using Parser 2.6.3.0, running on ruby 2.5.1 x86_64-darwin18)
```

denys281:
Resolves #7320 

-----------------

* [x] Added tests.


bbatsov:
I still don't see any tests covering the use of `attr` directly. Apart from this the PR looks good to me.

lib/rubocop/cop/naming/method_name.rb
```diff
@@ -31,9 +31,27 @@ module Naming
       class MethodName < Cop
         include ConfigurableNaming
         include IgnoredPattern
+        include RangeHelp
 
         MSG = 'Use %<style>s for method names.'
 
+        def_node_matcher :attr?, <<~PATTERN
```
bbatsov: Btw, there's also `define_method` that should be considered. In general it might be a good idea be able to provide a list of method names that define methods via configuration so custom macros could be accounted as well.

denys281:
@bbatsov added. But.... It works only when you pass symbols, but should it also support strings ?

bbatsov:
I guess it should handle strings as well.

spec/rubocop/cop/naming/method_name_spec.rb
```diff
@@ -118,9 +179,40 @@ def on_selection_cleared(arg)
     end
   end
 
+  shared_examples 'multiple attr methods' do |enforced_style|
+    it 'registers an offense for camel case methods names in attr.' do
+      expect_offense(<<~RUBY)
+        attr :my_method, :myMethod
+             ^^^^^^^^^^^^^^^^^^^^^ Use #{enforced_style} for method names.
```
bbatsov: Shouldn't we highlight just the second symbol here?

denys281: Will check it. Do you know maybe some cops that have similar behavior? 

So it should be something like that ? 

```
attr :my_method, :myMethod, my_snake_method, : myCamelMethod
	                           ^^^^^^^^^^^^                                   ^^^^^^^^^^^^^^^^^^^ Use #{enforced_style} for method names.
```

bbatsov: Yep. See https://github.com/rubocop-hq/rubocop/pull/7385

PR #7375, opened 2019-09-23
[Fix #7366] Mark unsafe for `Style/UnneededSort`

### LINKED ISSUE #7366: Style/UnneededSort is unsafe
When sorting field value is nil, max_by returns first element of array, but sort_by...last returns the last one. Tested using ruby 2.3.3 and 2.6.3

```
2.6.3 :001 > a = [{id: 1, created_at: nil}, {id: 2, created_at: nil}]
2.6.3 :002 > a.sort_by{|x| x[:created_at]}.last[:id]
 => 2
2.6.3 :003 > a.max_by{|x| x[:created_at]}[:id]
 => 1
```

--------

```
$ cat test.rb
a = [{id: 1, created_at: nil}, {id: 2, created_at: nil}]
a.sort_by{|x| x[:created_at]}.last[:id]
```

```
$ rubocop --safe-auto-correct --only Style/UnneededSort test.rb
Inspecting 1 file
C

Offenses:

test.rb:2:3: C: [Corrected] Style/UnneededSort: Use max_by instead of sort_by...last.
a.sort_by{|x| x[:created_at]}.last[:id]
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

```
$ cat test.rb
a = [{id: 1, created_at: nil}, {id: 2, created_at: nil}]
a.max_by{|x| x[:created_at]}[:id]
```

## RuboCop version

```
0.74.0 (using Parser 2.6.4.1, running on ruby 2.3.3 x86_64-darwin18)
```

koic:
Fixes #7366.

This PR marks unsafe for `Style/UnneededSort`.

As shown in #7366, the replacement of bad and good methods has the potential to behave differently.

-----------------

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@
 * [#7091](https://github.com/rubocop-hq/rubocop/issues/7091): `Style/FormatStringToken` now detects format sequences with flags and modifiers. ([@buehmann][])
 * [#7319](https://github.com/rubocop-hq/rubocop/pull/7319): Rename `IgnoredMethodPatterns` option to `IgnoredPatterns` option for `Style/MethodCallWithArgsParentheses`. ([@koic][])
 * [#7345](https://github.com/rubocop-hq/rubocop/issues/7345): Mark unsafe for `Style/YodaCondition`. ([@koic][])
+* [#7366](https://github.com/rubocop-hq/rubocop/issues/7366): Mark unsafe for `Style/UnneededSort`. ([@koic][])
 
 ## 0.74.0 (2019-07-31)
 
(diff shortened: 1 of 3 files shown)


bbatsov:
At this rate there won't be any safe cops left. :-) In this particular case the difference in behaviour could potentially be considered a Ruby bug.

bbatsov:
Anyways, I think it's important to also expand the docs to acknowledge what exactly is the corner case. There's also the point that we got this bug report after the cop has been around for 20 releases, which makes me wonder if we need a better criteria for cop safety (e.g. some levels or whatever).

koic:
Surely. It seems that similar issues have not been feedback since the release.

Maybe `VersionAdded` can be some threshold. On the other hand, there may be other cases where it is clearly unsafe.
I don't have a specific good idea yet, but in any case it seems a good idea to set some criteria.

I should have done more verification this time for the cop 💦 Thanks for the precise advice.

bbatsov:
You're welcome!

PR #7378, opened 2019-09-25
Fix heredoc edge cases in Layout/EmptyLineAfterGuardClause

gsamokovarov:
If you'd call a method on a heredoc string the cop couldn't
recognize that a heredoc is passed. The same happened if the heredoc was
given to a method, the result of which is passed as a value in another
call.

Here are the two examples:

```ruby
def foo
  raise ArgumentError, <<~END.squish.it.good unless guard
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Add empty line after guard clause.
    A multiline message
    that will be squished.
  END

  return_value
end
```

```ruby
def foo
  raise ArgumentError, call(<<~END.squish) unless guard
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Add empty line after guard clause.
    A multiline message
    that will be squished.
  END

  return_value
end
```


koic:
@gsamokovarov Thanks for the patch and this looks good to me. Could you rebase with the latest master branch?

gsamokovarov:
Thanks, @koici, I have rebased this.

koic:
Thanks!

koic:
Oops, I overlooked the position of the changelog entry, so I fixed it with https://github.com/rubocop-hq/rubocop/commit/13b78a489c52f9b15409e56a9de80b6cfdf899ac 💦

gsamokovarov:
Thanks again!

PR #7385, opened 2019-09-28
Deprecate departmentless cop names

jonas054:
Ever since departments (namespaces) were introduced, or almost since then, it's been possible to give cop names on the command line, in configuration and in `rubocop:disable` comments without a preceding department name. RuboCop deduces what the department should be and errors out only if it doesn't find exactly one matching department.

The optional department functionality was added as a way to stay backwards compatible with configuration that was created before departments existed. But the grace period is coming to an end. We start now by issuing warnings whenever cop names without departments are encountered. Later on, in a future version, we will remove the department deduction functionality and require that departments always be given for cop names.

**Note:** Warnings about `rubocop:disable` comments are not printed when inspection results are taken from the cache, so there's a chance this warning goes unnoticed in some cases.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1854,6 +1854,14 @@ Metrics/PerceivedComplexity:
   VersionAdded: '0.25'
   Max: 7
 
+################## Migration #############################
+
+Migration/DepartmentName:
+  Description: >-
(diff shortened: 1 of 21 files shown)


bbatsov:
Great work!

PR #7399, opened 2019-10-03
[Fix #7396]: Display ABC components with the value

### LINKED ISSUE #7396: Display ABC Metric components with the value
As ABC Metric technically is a size measure (A vector with three orthogonal components), so the magnitude of the vector without its components would never provide any insights over assignments, branches, or, conditions - how would one ever determine which one is the culprit?

Current message:
```ruby
MSG = 'Assignment Branch Condition size for %s is too high. [%.4g/%.4g]'
```

I would prefer it being updated to:
```ruby
MSG = 'Assignment Branch Condition size for %s is too high. [<%d, %d, %d> %.4g/%.4g]'
```

I felt the need for this when I was writing a story about the [complexity metrics in RuboCop](https://medium.com/@avmnusng/ruby-software-complexity-metrics-part-two-calculations-a00fbab610d2).

avmnu-sng:
Providing component values gives more insights to fix the offense.

`Metrics/AbcSize: Assignment Branch Condition size for choose_move is too high. [<8, 12, 6> 15.62/15]`
definitetly provides more information than
`Metrics/AbcSize: Assignment Branch Condition size for choose_move is too high. [15.62/15]`

-----------------

* [X] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 * [#7446](https://github.com/rubocop-hq/rubocop/issues/7446): Add `merge` to list of non-mutating methods. ([@cstyles][])
 * [#7077](https://github.com/rubocop-hq/rubocop/issues/7077): Rename `Unneeded*` cops to `Redundant*` (e.g., `Style/UnneededPercentQ` becomes `Style/RedundantPercentQ`). ([@scottmatthewman][])
+* [#7396](https://github.com/rubocop-hq/rubocop/issues/7396): Display assignments, branches, and conditions values with the offense. ([@avmnu-sng][])
 
 ## 0.75.1 (2019-10-14)
 
@@ -4242,3 +4243,4 @@
 [@jfhinchcliffe]: https://github.com/jfhinchcliffe
 [@jdkaplan]: https://github.com/jdkaplan
 [@cstyles]: https://github.com/cstyles
+[@avmnu-sng]: https://github.com/avmnu-sng
--- a/lib/rubocop/cop/metrics/utils/abc_size_calculator.rb
+++ b/lib/rubocop/cop/metrics/utils/abc_size_calculator.rb
@@ -43,7 +43,10 @@ module RuboCop
               end
             end
 
-            Math.sqrt(@assignment**2 + @branch**2 + @condition**2).round(2)
+            [
+              Math.sqrt(@assignment**2 + @branch**2 + @condition**2).round(2),
+              "<#{@assignment}, #{@branch}, #{@condition}>"
+            ]
           end
 
           def evaluate_branch_nodes(node)
--- a/lib/rubocop/cop/metrics/abc_size.rb
+++ b/lib/rubocop/cop/metrics/abc_size.rb
@@ -11,7 +11,7 @@ module RuboCop
         include MethodComplexity
 
         MSG = 'Assignment Branch Condition size for %<method>s is too high. ' \
-              '[%<complexity>.4g/%<max>.4g]'
+              '[%<abc_vector>s %<complexity>.4g/%<max>.4g]'
 
         private
 
--- a/lib/rubocop/cop/mixin/method_complexity.rb
+++ b/lib/rubocop/cop/mixin/method_complexity.rb
@@ -32,13 +32,14 @@ module RuboCop
         return unless node.body
 
         max = cop_config['Max']
-        complexity = complexity(node.body)
+        complexity, abc_vector = complexity(node.body)
 
         return unless complexity > max
 
         msg = format(self.class::MSG,
                      method: method_name,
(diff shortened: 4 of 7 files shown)


bbatsov:
The suggestion is fine by me, but I wonder if the extra info won't be confusing to some users. @rubocop-hq/rubocop-core what do you think about this?

avmnu-sng:
> The suggestion is fine by me, but I wonder if the extra info won't be confusing to some users. @rubocop-hq/rubocop-core what do you think about this?

@bbatsov Might be confusing for few, but ideally, without displaying each component the ABC Size value is nothing but just a decimal value IMHO. One would never be sure about what individual components' values are unless count these manually.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 * [#7317](https://github.com/rubocop-hq/rubocop/pull/7317): Add new formatter `pacman`. ([@crojasaragonez][])
 * [#6075](https://github.com/rubocop-hq/rubocop/issues/6075): Support `IgnoredPatterns` option for `Naming/MethodName` cop. ([@koic][])
 * [#7335](https://github.com/rubocop-hq/rubocop/pull/7335): Add todo as an alias to disable. `--disable-uncorrectable` will now disable cops using `rubocop:todo` instead of `rubocop:disable`. ([@desheikh][])
+* [#7396](https://github.com/rubocop-hq/rubocop/issues/7396): Display assignments, branches, and, conditions values with the offence. ([@avmnu-sng][])
```
bbatsov: Move this to the `master` branch. I'd put it under `Changes` there.

bbatsov:
Yeah, that's a fair point. I wonder if we shouldn't put an `(A)`, `(B)` and `(C)` after each number, but I guess we can move forward without them. 

You'll also have to rebase on top of `master`.

lib/rubocop/cop/metrics/utils/abc_size_calculator.rb
```diff
@@ -44,7 +44,10 @@ def calculate
               end
             end
 
-            Math.sqrt(@assignment**2 + @branch**2 + @condition**2).round(2)
+            [
+              Math.sqrt(@assignment**2 + @branch**2 + @condition**2).round(2),
+              "<#{@assignment},#{@branch},#{@condition}>"
```
bbatsov: Might be a good idea to add a space after each `,`.

avmnu-sng: Yep, agreed. Updated.

CHANGELOG.md
avmnu-sng: Done!

avmnu-sng:
> Yeah, that's a fair point. I wonder if we shouldn't put an `(A)`, `(B)` and `(C)` after each number, but I guess we can move forward without them.
> 
> You'll also have to rebase on top of `master`.

Usually, we write vectors like `(a, b, c, d)` or `<a, b, c, d>`, But here we already use `[]`, so not sure which one is visually good, `[<a, b, c> value/max]` or `[(a, b, c) value/max]`

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 * [#7446](https://github.com/rubocop-hq/rubocop/issues/7446): Add `merge` to list of non-mutating methods. ([@cstyles][])
 * [#7077](https://github.com/rubocop-hq/rubocop/issues/7077): Rename `Unneeded*` cops to `Redundant*` (e.g., `Style/UnneededPercentQ` becomes `Style/RedundantPercentQ`). ([@scottmatthewman][])
+* [#7396](https://github.com/rubocop-hq/rubocop/issues/7396): Display assignments, branches, and, conditions values with the offence. ([@avmnu-sng][])
```
koic: There is an unnecessary comma.

```diff
-Display assignments, branches, and, conditions
+Display assignments, branches, and conditions
```

CHANGELOG.md
```diff
@@ -4242,3 +4243,4 @@
 [@jfhinchcliffe]: https://github.com/jfhinchcliffe
 [@jdkaplan]: https://github.com/jdkaplan
 [@cstyles]: https://github.com/cstyles
+[@avmnu-sng]: https://github.com/username
```
koic: It should be replaced with your account name.

koic:
After understanding `<A, B, C>`, there is no problem, but I was a little confused when I first saw it.

CHANGELOG.md
avmnu-sng: Fixed!

CHANGELOG.md
avmnu-sng: How silly! Fixed

bbatsov:
Thanks!

PR #7401, opened 2019-10-04
[Fix #7389] Fixing `-fp` exception

### LINKED ISSUE #7389: Cannot determine formatter for "p"
## Actual behavior

`Cannot determine formatter for "p"`

`rubocop -fp`

## RuboCop version

```
$ [bundle exec] rubocop -V
0.75.0
```
Caused by https://github.com/rubocop-hq/rubocop/pull/7317 as the logic of determining the formatter matches the "first" letters.¹ Since [p] is now not unique anymore this can't work
(noticed this myself when I tried to introduce my own formatter couple months ago)

What works is `rubocop -fpr` but that's not what the help tells.

So either change the help or replace the logic with an actual "pattern" match rather than a `starts with`

```
    -f, --format FORMATTER           Choose an output formatter. This option
                                     can be specified multiple times to enable
                                     multiple formatters at the same time.
                                       [pr]rogress (default)
                                       [pa]cman
```

¹ - https://github.com/rubocop-hq/rubocop/pull/6407#issuecomment-439799743

zmaupin:
When using `-fp` an error is raised because it matches two keys ('progress' and 'pacman'). Now when using `-fp` we default to progress and don't raise an exception.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#7389](https://github.com/rubocop-hq/rubocop/issues/7389): Fix error when using `-fp` so we now use `progress` when using `-fp`. ([@zmaupin][])
+
 ## 0.75.0 (2019-09-30)
 
 ### New features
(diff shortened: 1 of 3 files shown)


bbatsov:
See also https://github.com/rubocop-hq/rubocop/pull/7402 I think the solution there is a bit better.

zmaupin:
I think you're right. Thanks for looking at mine.

bbatsov:
Guess we can close this PR.

PR #7402, opened 2019-10-04
[Fix #7389] Ambiguous formatter matching

### LINKED ISSUE #7389: Cannot determine formatter for "p"

jfhinchcliffe:
Fixed an issue where passing a formatter alias which started with the
same character as another caused an error.

Specifically, passing in `-f p` caused an error as this matched
both 'pacman' and 'progress' in
Rubocop:: Formatter::FormatterSet::BUILTIN_FORMATTERS_FOR_KEYS

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ### Bug fixes
 
+* [#7389](https://github.com/rubocop-hq/rubocop/issues/7389): Fix an issue where passing a formatter might result in an error depending on what character it started with. ([@jfhinchcliffe][])
 * [#7256](https://github.com/rubocop-hq/rubocop/issues/7256): Fix an error of `Style/RedundantParentheses` on method calls where the first argument begins with a hash literal. ([@halfwhole][])
 * [#7263](https://github.com/rubocop-hq/rubocop/issues/7263): Make `Layout/SpaceInsideArrayLiteralBrackets` properly handle tab-indented arrays. ([@buehmann][])
 * [#7252](https://github.com/rubocop-hq/rubocop/issues/7252): Prevent infinite loops by making `Layout/SpaceInsideStringInterpolation` skip over interpolations that start or end with a line break. ([@buehmann][])
@@ -4208,3 +4209,4 @@
 [@raymondfallon]: https://github.com/raymondfallon
 [@crojasaragonez]: https://github.com/crojasaragonez
 [@desheikh]: https://github.com/desheikh
+[@jfhinchcliffe]: https://github.com/jfhinchcliffe
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#7389](https://github.com/rubocop-hq/rubocop/issues/7389): Fix an issue where passing a formatter might result in an error depending on what character it started with. ([@jfhinchcliffe][])
+
 ## 0.75.0 (2019-09-30)
 
 ### New features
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 
 ### Bug fixes
 
+* [#7389](https://github.com/rubocop-hq/rubocop/issues/7389): Fix an issue where passing a formatter might result in an error depending on what character it started with. ([@jfhinchcliffe][])
```
bbatsov: This should go under `master`.

lib/rubocop/formatter/formatter_set.rb
```diff
@@ -9,21 +9,21 @@ module Formatter
...
-        'html'        => HTMLFormatter,
-        'files'       => FileListFormatter,
-        'offenses'    => OffenseCountFormatter,
-        'disabled'    => DisabledLinesFormatter,
-        'worst'       => WorstOffendersFormatter,
-        'tap'         => TapFormatter,
-        'quiet'       => QuietFormatter,
-        'autogenconf' => AutoGenConfigFormatter,
-        'pacman'      => PacmanFormatter
+        '[p]rogress'    => ProgressFormatter,
```
bbatsov: I think we should sort those alphabetically.

lib/rubocop/formatter/formatter_set.rb
```diff
@@ -82,7 +82,7 @@ def formatter_class(formatter_type)
 
       def builtin_formatter_class(specified_key)
         matching_keys = BUILTIN_FORMATTERS_FOR_KEYS.keys.select do |key|
-          key.start_with?(specified_key)
+          key =~ /^\[#{specified_key}\]/ || specified_key == key.delete('[]')
```
bbatsov: Do we even need the regexp match? I guess we can always be explicit.

lib/rubocop/formatter/formatter_set.rb
```diff
@@ -9,21 +9,21 @@ module Formatter
...
+        '[j]son'        => JSONFormatter,
+        '[h]tml'        => HTMLFormatter,
+        '[fi]les'       => FileListFormatter,
+        '[o]ffenses'    => OffenseCountFormatter,
+        '[d]isabled'    => DisabledLinesFormatter,
+        '[w]orst'       => WorstOffendersFormatter,
+        '[t]ap'         => TapFormatter,
+        '[q]uiet'       => QuietFormatter,
+        '[a]utogenconf' => AutoGenConfigFormatter,
+        '[pa]cman'      => PacmanFormatter
```
buehmann: The abbreviations are repeated here: https://github.com/jfhinchcliffe/rubocop/blob/c29143c5379c14699096ddb114736f8312a64ca6/lib/rubocop/options.rb#L410-L424

Maybe we could DRY this somehow?

bbatsov:
Thanks!

PR #7405, opened 2019-10-04
support deny/allow listing specific offenses for autocorrect

d-ray:
Add the option for autocorrection to be done on only selected offenses (identified by file name, line number, column number, and cop name).  While the existing set of options allows limiting autocorrect to particular cops and/or particular files, there isn't a way to correct some offenses of a particular cop but not others within the same file.  One useful application is to autocorrect newly introduced offenses but not existing ones (think large files/legacy code), in combination with other existing tools that run rubocop and suppress messages for offenses on unchanged lines (e.g. https://github.com/prontolabs/pronto, https://github.com/appfolio/farcy/).

-----------------

* [x] Added tests.


bbatsov:
When we ignore my dislike for the names whitelist and blacklist (which these days are more of than not allowlist and denylist) I'm not sure about the usefulness of that proposal. I can imagine how doing some changes to a part of a file are likely going to break something there. I also don't really get the concept of partially fixing a file - with this approach you'll likely take forever to unify the code style in a project. Please, explain in more details how exactly do you envision this to be used.

d-ray:
I hadn't heard of allowlist and denylist as preferred alternatives to whitelist and blacklist, but happy to change the names.

On the concept of partially fixing a file, I think it's true that this change isn't particularly useful in a project where all the developers agree that unifying the code style is desirable and worth enforcing.  That is however, not the case at my current company (and I suspect we're not unique in that).  The idea of enforcing code style guidelines has come up a few times over the years, and there's always been enough resistance to the idea to prevent it from moving forward.  I'd say there's majority support for the idea that the style feedback is valuable but also majority opposition to enforcing it.  As a workaround/compromise of sorts, a couple of my coworkers built [this](https://github.com/appfolio/farcy/), which runs rubocop (and other linting tools) and comments on github pull requests with the issues that apply to changed lines.  This way people get the feedback but can act on it or ignore it as they wish.  My goal in adding partial autocorrect to rubocop is to extend farcy's functionality to include autocorrect; the current thought is that the changes would be committed to a separate branch which could then be merged into the original PR.  This may serve as a glide path towards a full embrace of style enforcement; people decide they like the ease and simplicity of autocorrect so much that we collectively agree that we should apply it everywhere.  For my money that'd be the ideal outcome, but I'm not holding my breath 😅 More generally, I think these changes make rubocop more useful to organizations like my own that place some value on style consistency but aren't willing to go all in on strict enforcement.

d-ray:
@bbatsov any further thoughts on this?

bbatsov:
Sorry about the radio silence. I've been super busy with other stuff lately and I can't catch up with all the PRs. I'll try to go over all of them this week.

marcandre:
In case this request is rejected, there is the possibility of a low level access. The recent Cop refactor allows us to dump to disk an investigation report, *including the autocorrections*. This was not possible previously as one can not serialize lambdas, but we now use `Parser::Source::TreeRewriter` that can be serialized. I've not yet taken the time to amend the rubocop cache, but I'd actually like to use this one day so that we can run `rubocop -a` and that doing `rubocop -A` right after could be near quite quick. Maybe this could provide a way for a separate tool to load the offenses too and deal with them as they see fit.

PR #7406, opened 2019-10-05
Add new Style/NonLocalControlFlowInTernary cop

Drenmi:
This cop checks for non-local control flow, i.e. `return`, `break`, `next`, `raise`, etc. used in ternary expressions.

-----------------

* [X] Added tests.


bbatsov:
This will also need an entry in the style guide.

rrosenblum:
What if control flow exists on both sides of the ternary? Should it still be flagged? `foo? ? next 1 : next 2` Or `foo? ? break : next`? Or `foo? ? raise 'Error 1' : raise 'Error 2'`?

Drenmi:
> This will also need an entry in the style guide.

👍 Will add when I have some time.

> What if control flow exists on both sides of the ternary? Should it still be flagged?

Yes. I should add a test case for that. 🙂

bbatsov:
@Drenmi ping :-)

Drenmi:
@bbatsov I have thought about it, and I think it's worth generalizing beyond just ternary expressions.

I opened a PR to The Ruby Style Guide [here](https://github.com/rubocop-hq/ruby-style-guide/pull/785).

koic:
It may not be a problem when limited to ternary operator. 
On the other hand, I commented on the Style Guide about cases that I found difficult to generalize.
https://github.com/rubocop-hq/ruby-style-guide/pull/785#issuecomment-569915592

Drenmi:
@koic @bbatsov I have rebased this and added the spec that @rrosenblum suggested. I think we can merge this. I suspect the Style Guide PR will take some time to iron out. 🙂

config/default.yml
```diff
@@ -3284,6 +3284,11 @@ Style/NilComparison:
     - predicate
     - comparison
 
+Style/NonLocalControlFlowInTernary:
+  Description: 'checks for non-local control flow in ternary expressions.'
+  Enabled: true
+  VersionAdded: '0.76'
```
koic, line 3290: It looks like the version needs to be updated.

```suggestion
  VersionAdded: '0.81'
```

lib/rubocop/cop/style/non_local_control_flow_in_ternary.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+# TODO: when finished, run `rake generate_cops_documentation` to update the docs
```
koic, line 3: Can you remove the auto-generated line?

bbatsov:

PR #7409, opened 2019-10-05
[Fix #7407] Better detect typical format string contexts

### LINKED ISSUE #7407: Style/FormatStringToken doesn't work inside a hash
## Expected behavior

Given a file foobar.rb:

```ruby
# frozen_string_literal: true

format('%{foo}', foo: 'foo')
{ bar: format('%{foo}', foo: 'foo') }
```

I expect to get two offenses:
```
foobar.rb:3:9: C: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over template tokens (like %{foo}).
format('%{foo}', foo: 'foo')
        ^^^^^^
foobar.rb:4:16: C: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over template tokens (like %{foo}).
{ bar: format('%{foo}', foo: 'foo') }
               ^^^^^^
```

## Actual behavior

I only get the first one of them. `Style/FormatStringToken` offenses from inside a hash aren't reported.

```
foobar.rb:3:9: C: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over template tokens (like %{foo}).
format('%{foo}', foo: 'foo')
        ^^^^^^
```

With a file foobar.rb from above, execute:

```
$ rubocop foobar.rb
```

## RuboCop version

```
$ rubocop -V
0.75.0 (using Parser 2.6.4.1, running on ruby 2.5.3 x86_64-linux)
```

buehmann:
This fixes #7407 by improving upon earlier attempts at avoiding false
positives: Strings that look like format strings in the `unannotated`
style (with simple format sequences such as `%a`) can as well be
something else depending on the context (e.g. date/time formats, encoded
URLs).

The documentation of the cop has long had a note about in which typical
contexts such strings are considered format strings. This is now
reflected in a node matcher in the code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#7407](https://github.com/rubocop-hq/rubocop/issues/7407): Make `Style/FormatStringToken` work inside hashes. ([@buehmann][])
+
 ## 0.75.0 (2019-09-30)
 
 ### New features
--- a/lib/rubocop/cop/style/format_string_token.rb
+++ b/lib/rubocop/cop/style/format_string_token.rb
@@ -95,11 +95,7 @@ module RuboCop
           if source_map.is_a?(Parser::Source::Map::Heredoc)
             source_map.heredoc_body
           elsif source_map.begin
-            slice_source(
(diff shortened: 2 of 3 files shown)


bbatsov:
Great work! 🙇

PR #7411, opened 2019-10-06
[Fix #7410] Add printf as a typical format string context

### LINKED ISSUE #7410: Style/FormatStringToken should inspect printf argument
While working on #7409 I noticed that `Style/FormatStringToken` recognizes simple format strings as the first argument of `sprintf` or `format` methods, but not as the first argument of `printf`:

```sh-session
$ cat printf.rb                                                   
sprintf('%08s %04s', 'foo', 'bar')
printf('%08s %04s', 'foo', 'bar')
$ rubocop printf.rb --only Style/FormatStringToken --format simple
== printf.rb ==
C:  1: 10: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over unannotated tokens (like %s).
C:  1: 15: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over unannotated tokens (like %s).

```
I would expect similar offenses in the second line.

buehmann:
The cop `Style/FormatStringToken` used to treat `sprintf`, `format`, and `%` methods as typical use cases of `printf`-style format strings, but ironically not the `printf` method itself.

This fixes #7410.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * [#7407](https://github.com/rubocop-hq/rubocop/issues/7407): Make `Style/FormatStringToken` work inside hashes. ([@buehmann][])
 
+### Changes
+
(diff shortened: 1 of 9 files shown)


bbatsov:
Nice catch!

PR #7417, opened 2019-10-08
[Fix #7273] Added cop for spaces around method calls.

### LINKED ISSUE #7273: Cop to check the space before dot
This code `"a" .class` despite of being a valid ruby code, can be improved by removing the space.

I've searched the documentation in order to find something already implementing this, but haven't found nothing, just the style cops for dot position in the method chain.

Input:
`"a" .class"`
Expected:
`"a".class"`

timothyjosefik:
This will throw a new offense anytime a space is around a dot call except
when the dot is the first thing on a line.

Note: This only works for dot calls and does not work with &. or ::

-----------------

* [x] Added tests.


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#7273](https://github.com/rubocop-hq/rubocop/issues/7273): Add new `Style/SpaceAroundMethodCall` cop. ([@timothyjosefik][])
```
buehmann: Please add a `New features` subsection; see 0.75.0 down below for an example.

lib/rubocop/cop/layout/space_around_method_call.rb
```diff
@@ -0,0 +1,72 @@
...
+          no_space_offenses(node, dot, right, AFTER_MSG)
+          return if left.line != dot.line
+
+          no_space_offenses(node, left, dot, BEFORE_MSG)
+        end
+
+        private
+
+        def dot_token(tokens)
+          tokens.reverse.find(&:dot?)
```
buehmann: This will find the wrong dot (the second one) in cases like `foo . bar(foo.bar)` and hence not detect the spaces around the first one.

lib/rubocop/cop/layout/space_around_method_call.rb
```diff
@@ -0,0 +1,72 @@
...
+        end
+
+        def previous_token(current_token)
+          index = processed_source.tokens.index(current_token)
+          index.nil? || index.zero? ? nil : processed_source.tokens[index - 1]
+        end
+
+        def next_token(current_token)
+          index = processed_source.tokens.index(current_token)
+          index.nil? || index.zero? ? nil : processed_source.tokens[index + 1]
```
buehmann: The `index.zero?` condition does not seem to make sense here. In `previous_token` it is used to not produce negative array indexes, but here?

lib/rubocop/cop/layout/space_around_method_call.rb
timothyjosefik: The reason I did `reverse` was because when doing `foo.bar.foo2 . bar2` on the first `on_send` the `node` contains `foo.bar.foo2 . bar2` on the second `on_send` the node contains just `foo.bar.foo2` so I needed to get the last token to find the space there. 

You are correct that with `foo . bar(foo.bar)` using the `reverse` will not detect the space. 

Any ideas on how to get both scenarios to work?

buehmann: You could make use of the source map of the send node in order to find the right dot: `send_node.loc.dot` will give you the location of the dot in the source which you could compare to the token positions.

Be aware that `loc.dot` can be nil because there are send nodes without dots in the source. You might want to cover this case in your specs. There is a convenience method `dot?` for send nodes, so you could do something like this:
```ruby
def on_send(node)
  return unless node.dot?
```

bbatsov:
> Note: This only works for dot calls and does not work with &. or ::

Those have to be supported by the same cop IMO.

bbatsov:

PR #7418, opened 2019-10-08
Exclude .git from internal investigation

buehmann:
In the project's `.rubocop.yml` we overwrite the default
`AllCops/Exclude` and do not exclude `.git`.

On my machine fixing this saves 1 of 3 seconds for a simple `rubocop
--list-target-files`.


koic:
Thanks!

Issue #7426, opened 2019-10-13
Frozen String Literal Behavior is Confusing


gfyoung:
According to <a href="https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FrozenStringLiteralComment">the docs</a>, my understanding was that in order for the cop to be satisfied, we had to add `# frozen_string_literal: true` to the top of the file.

However, after finding what appeared to be a false negative in one of my projects, I found #7241 and its corresponding fix in #7251 that highlighted that adding `# frozen_string_literal: false` to the top can still pass this cop.  Two points:

* I think the docs should be updated to reflect that `false` is accepted in the magic comment
* Is there a way to configure the cop so that it enforces `# frozen_string_literal: true` only?

## Rubocop Info

~~~
0.75.0 (using Parser 2.6.5.0, running on ruby 2.6.5 x86_64-linux)
~~~

koic:
Good points.

> I think the docs should be updated to reflect that `false` is accepted in the magic comment

Yeah, it will be easy to understand if it is documented.

> Is there a way to configure the cop so that it enforces `# frozen_string_literal: true` only?

AFAIK, there is no such configure option, so enhancement will be necessary.

gfyoung:
@koic : Thanks for your reply!

I imagine a docs change is something people wouldn't object to.

The enhancement would be nice.  Would be curious to know if people have any objections.

olliebennett:
I assumed the `Style/FrozenStringLiteralComment` cop already checked for `true`, to be honest.

For reference, current RuboCop [default configuration](https://github.com/rubocop-hq/rubocop/blob/master/config/default.yml#L2731) options are;

```
  EnforcedStyle: always
  SupportedStyles:
    # `always` will always add the frozen string literal comment to a file
    # regardless of the Ruby version or if `freeze` or `<<` are called on a
    # string literal. If you run code against multiple versions of Ruby, it is
    # possible that this will create errors in Ruby 2.3.0+.
    - always
    # `never` will enforce that the frozen string literal comment does not
    # exist in a file.
    - never
```

Would it make sense to add a new SupportedStyle of `alwaystrue`?

```
    # `alwaystrue` will enforce that the frozen string literal comment exists
    # and is set to `true` (i.e. enforcing that string literals _are_ frozen). 
    - alwaystrue
```

[...]

PR #7429, opened 2019-10-14
Document support for frozen_string_literal: false

gfyoung:
xref #7426

### DIFF
--- a/lib/rubocop/cop/style/frozen_string_literal_comment.rb
+++ b/lib/rubocop/cop/style/frozen_string_literal_comment.rb
@@ -11,6 +11,9 @@ module RuboCop
       # encoding comment. The frozen string literal comment is only valid in
       # Ruby 2.3+.
       #
+      # Note that the cop will ignore files where the comment exists but is set
+      # to `false` instead of `true`.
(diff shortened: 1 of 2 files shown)


gfyoung:
This is my first time contributing to the `rubocop` project, so I have a question about commit messages.  My PR is not "fixing" the issue (it's a two-parter, and this only addresses one part of it), so I wasn't sure that I needed to add a "fix" prefix to my commit message.

Let me know what makes sense here!

bbatsov:
The comment seems a bit confusing to me. Shouldn't it say that RuboCop will ignore files where the  the comment exists and is set to `false` instead?

gfyoung:
> The comment seems a bit confusing to me. Shouldn't it say that RuboCop will ignore files where the the comment exists and is set to false instead?

@bbatsov : We can write it like that as well.  I just want to make it clear that users have an alternative if they have the cop turned on.

gfyoung:
@bbatsov : I've updated the wording and included an example.  Have a look.

bbatsov:
Looks good to me. Thanks!

PR #7435, opened 2019-10-16
[Fix #7426] Add `always_true` style to Style/FrozenStringLiteralComment

### LINKED ISSUE #7426: Frozen String Literal Behavior is Confusing

parkerfinch:
This adds another style to Style/FrozenStringLiteralComment that
enforces that the comment is always enabled, that is, requiring that
string literals are always frozen. The default style, `always`,
enforces that the comment exists but does not enforce that the comment
is enabled.

@gfyoung does this line up with what you're thinking? Feel free to make any changes to this as you see fit and push up something that aligns more closely with your idea!

I had most of this put together before I saw @olliebennett's [suggestion](https://github.com/rubocop-hq/rubocop/issues/7426#issuecomment-542400659) to name this `alwaystrue` -- I'm happy to change the name here, let me know if you have thoughts!

I am an infrequent contributor, please let me know if I forgot any steps or could make this easier to take a look at!

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/frozen_string_literal_comment.rb
+++ b/lib/rubocop/cop/style/frozen_string_literal_comment.rb
@@ -51,29 +51,65 @@ module RuboCop
       #   module Baz
       #     # ...
       #   end
+      #
+      # @example EnforcedStyle: always_true
+      #   # The `always_true` style enforces that the frozen string literal
+      #   # comment is set to `true`. This is a stricter option than `always`
+      #   # and forces projects to use frozen string literals.
+      #   # bad
+      #   # frozen_string_literal: false
+      #
+      #   module Baz
+      #     # ...
+      #   end
+      #
+      #   # bad
+      #   module Baz
+      #     # ...
+      #   end
+      #
+      #   # good
+      #   # frozen_string_literal: true
+      #
+      #   module Bar
+      #     # ...
+      #   end
       class FrozenStringLiteralComment < Cop
         include ConfigurableEnforcedStyle
         include FrozenStringLiteral
         include RangeHelp
 
-        MSG = 'Missing magic comment `# frozen_string_literal: true`.'
+        MSG_MISSING_TRUE = 'Missing magic comment `# frozen_string_literal: '\
+                           'true`.'
+        MSG_MISSING = 'Missing magic "frozen_string_literal" comment.'
         MSG_UNNECESSARY = 'Unnecessary frozen string literal comment.'
+        MSG_DISABLED = 'Frozen string literal comment must be set to `true`.'
         SHEBANG = '#!'
 
         def investigate(processed_source)
           return if processed_source.tokens.empty?
 
-          if frozen_string_literal_comment_exists?
-            check_for_no_comment(processed_source)
+          case style
+          when :never
+            ensure_no_comment(processed_source)
+          when :always_true
+            ensure_enabled_comment(processed_source)
           else
-            check_for_comment(processed_source)
+            ensure_comment(processed_source)
           end
         end
 
         def autocorrect(node)
           lambda do |corrector|
-            if style == :never
+            case style
+            when :never
               remove_comment(corrector, node)
+            when :always_true
+              if frozen_string_literal_specified?
+                enable_comment(corrector)
+              else
+                insert_comment(corrector)
+              end
             else
               insert_comment(corrector)
             end
@@ -82,12 +118,27 @@ module RuboCop
 
         private
 
-        def check_for_no_comment(processed_source)
-          unnecessary_comment_offense(processed_source) if style == :never
+        def ensure_no_comment(processed_source)
+          return unless frozen_string_literal_comment_exists?
+
+          unnecessary_comment_offense(processed_source)
         end
 
-        def check_for_comment(processed_source)
-          offense(processed_source) unless style == :never
+        def ensure_comment(processed_source)
+          return if frozen_string_literal_comment_exists?
+
+          missing_offense(processed_source)
+        end
+
+        def ensure_enabled_comment(processed_source)
+          if frozen_string_literal_specified?
+            return if frozen_string_literals_enabled?
+
+            # The comment exists, but is not enabled.
+            disabled_offense(processed_source)
+          else # The comment doesn't exist at all.
+            missing_true_offense(processed_source)
+          end
         end
 
         def last_special_comment(processed_source)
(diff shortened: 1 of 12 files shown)


lib/rubocop/cop/style/frozen_string_literal_comment.rb
```diff
@@ -117,11 +165,29 @@ def unnecessary_comment_offense(processed_source)
...
+        end
+
         def remove_comment(corrector, node)
           corrector.remove(range_with_surrounding_space(range: node.pos,
                                                         side: :right))
         end
 
+        def enable_comment(corrector)
+          # require "pry"
+          # binding.pry
```
buehmann: Old debugging code

parkerfinch: Gah, thanks!

parkerfinch:
Thanks for taking a look @gfyoung, made the suggested name change!

bbatsov:
I like the idea in general. I'm thinking that `always_true` might be more in line with the current naming.

gfyoung:
> I like the idea in general. I'm thinking that always_true might be more in line with the current naming.

@bbatsov : Sounds good.  Whatever term is most intuitive!

parkerfinch:
Thanks @bbatsov , pushed that name change!

lib/rubocop/cop/style/frozen_string_literal_comment.rb
```diff
@@ -41,29 +41,62 @@ module Style
...
+      #
+      #   module Bar
+      #     # ...
+      #   end
       class FrozenStringLiteralComment < Cop
         include ConfigurableEnforcedStyle
         include FrozenStringLiteral
         include RangeHelp
 
         MSG = 'Missing magic comment `# frozen_string_literal: true`.'
```
gfyoung: Should we change this one since we don't strictly require `# frozen_string_literal: true` ?

parkerfinch: Hrrmm -- I see how the current comment could be misleading, but I'm not sure what a better solution is. Maybe if we made it clear that we weren't giving the full text of the comment, e.g. `Missing magic "frozen_string_literal" comment`? But I think there's some value in giving the exact syntax expected for the comment.

Do you have thoughts on what to replace this with?

gfyoung: > But I think there's some value in giving the exact syntax expected for the comment

If you're going to give exact syntax, then you should give all of the accepted values for clarity to the end user.  Otherwise, I would say `Missing magic "frozen_string_literal" comment`

bbatsov: Yeah, probably there should be different messages for the different enforced styles. The current message is appropriate only for the new `always_true` style.

parkerfinch:
@gfyoung @bbatsov I pushed up the message change for the current cop -- let me know if there's anything else I can do to help out here!

gfyoung:
@parkerfinch : These changes look great!  Just waiting on @bbatsov to give further feedback or merge.

parkerfinch:
@bbatsov I think this is good to go, let me know if there's anything I can do to make this easier to review!

lib/rubocop/cop/style/frozen_string_literal_comment.rb
```diff
@@ -51,29 +51,64 @@ module Style
       #   module Baz
       #     # ...
       #   end
```
koic: Please add one blank line. This is a nitpick :-)

```diff
      #   end
+     #
      # @example EnforcedStyle: always_true
```

parkerfinch: @koic Good call! If there was ever a repo in which to nitpick, this is it :)

Blank line added!

parkerfinch:
@bbatsov @koic Think this is good to merge?

gfyoung:
I would definitely love for this feature to be merged!  Thanks for all the hard work @parkerfinch!

bbatsov:

lib/rubocop/cop/style/frozen_string_literal_comment.rb
```diff
@@ -51,29 +51,65 @@ module Style
...
+      #   end
       class FrozenStringLiteralComment < Cop
         include ConfigurableEnforcedStyle
         include FrozenStringLiteral
         include RangeHelp
 
-        MSG = 'Missing magic comment `# frozen_string_literal: true`.'
+        MSG_MISSING_TRUE = 'Missing magic comment `# frozen_string_literal: '\
+                           'true`.'
+        MSG_MISSING = 'Missing magic "frozen_string_literal" comment.'
```
bbatsov: It'd be nice if all those messages employed consistent wording. Now they look a bit weird.

pirj:
Before this gets merged, I'd like to coin my 2¢.

For automatically freezing string literals I see those sane options:

1. Always enable by including `#frozen_string_literal: true` directive. Good when you know string literals you define won't be passed to third-party code that will attempt to mutate them.

2. Enable only when needed, e.g. include only if there are string literals in the file. Same use case as above. The cop would suggest to include when there are string literals and to remove when there are no string literals.
An option to this might be a hysteresis when pragma is suggested to be added when there are string literals in the file, but it's not suggested to remove the pragma when there are no string literals.

3. Disable. Nothing to be done except for disabling the cop, since freezing is disabled by default. Used when string literals might be used in third-party code attempting to mutate them.
A pretty edge case is to selectively disable, but this can be done by using `Exclude` list, or locally disabling cop for those files. Is this any worse than `frozen_string_literal: false`?

4. Selectively disable. When the interpreter is being run with a flag (`ruby --enable frozen-string-literal`).

With that, I'm trying to point out that `always_true` should probably become what `always` is now, and there's no real use case for `always` with its current behaviour.

Hope you'll find this convincing enough to change how `always` behaves, and I can handle adding `when_needed` and `only_when_needed`.

lib/rubocop/cop/style/frozen_string_literal_comment.rb
gfyoung: Addressed in https://github.com/rubocop-hq/rubocop/pull/7634

gfyoung:
Opened https://github.com/rubocop-hq/rubocop/pull/7634 to bring this over the finish line

parkerfinch:
Closing in favor of https://github.com/rubocop-hq/rubocop/pull/7634

PR #7437, opened 2019-10-16
[Fix #7434] Fix an incorrect autocorrect for `Style/MultilineWhenThen`

### LINKED ISSUE #7434: Style/MultilineWhenThen autocorrect generates invalid Ruby for certain case statements
The autocorrect for `Style/MultilineWhenThen` generates invalid ruby when correcting a case statement with the following format (`then` on the line after the `when`):

```
case foo
when :bar
  then :value
```

Definitely not a standard way to write a case statement but it is valid Ruby.

--------

## Full Example code:

```
def flavor(flave)
  case flave
  when 'chocolate'
    then %w[dark white milk].sample
  when 'fruit'
    then %w[apple banana orange].sample
  when 'null'
    then nil
  end
end
```

## Expected behavior

I would expect autocorrect to just remove the unnecessary `then` like so:

```
def flavor(flave)
  case flave
  when 'chocolate'
   %w[dark white milk].sample
  when 'fruit'
   %w[apple banana orange].sample
  when 'null'
   nil
  end
end
```

## Actual behavior

Instead, what we get is invalid ruby:

```
def flavor(flave)
  case flave
  when 'chocolate' %w[dark white milk].sample
  when 'fruit' %w[apple banana orange].sample
  when 'null' nil
  end
end
```

The sample code provided above can reproduce the problem.

## RuboCop version

```
0.75.1 (using Parser 2.6.5.0, running on ruby 2.4.2 x86_64-darwin17)
```

koic:
Fixes #7434.

This PR fixes an incorrect autocorrect for `Style/MultilineWhenThen` when the body of `when` branch starts with `then`.

```console
% cat example.rb
# frozen_string_literal: true

case foo
when bar
  then do_something
end
```

```
% ruby -c example.rb
Syntax OK
```

## Before

```console
% rubocop example.rb
Inspecting 1 file
C

Offenses:

example.rb:5:3: C: Style/MultilineWhenThen: Do not use then for
multiline when statement.
  then do_something
  ^^^^
```

It is changed to the code with syntax error as follows.

```ruby
# frozen_string_literal: true

case foo
when bar do_something
end
```

```console
% ruby -c example.rb
example.rb:4: syntax error, unexpected tIDENTIFIER, expecting do or '{'
or '('
when bar do_something
```

## After

```console
% rubocop example.rb -a
Inspecting 1 file
C

Offenses:

example.rb:5:1: C: [Corrected] Layout/IndentationWidth: Use 2 (not 1)
spaces for indentation.
 do_something
^
example.rb:5:3: C: [Corrected] Style/MultilineWhenThen: Do not use then
for multiline when statement.
  then do_something
  ^^^^

```

It will be changed to valid Ruby code as follows.

```ruby
# frozen_string_literal: true

case foo
when bar
  do_something
end
```

There is a slight difference in indentation using only this cop. That indentation is corrected using `Layout/IndentationWidth` cop.

```diff
 case foo
 when bar
- do_something
+  do_something
 end
```

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/multiline_when_then_spec.rb
+++ b/spec/rubocop/cop/style/multiline_when_then_spec.rb
@@ -84,4 +84,21 @@ RSpec.describe RuboCop::Cop::Style::MultilineWhenThen do
       end
     RUBY
   end
+
+  it 'autocorrects when the body of `when` branch starts ' \
+     'with `then`' do
+    new_source = autocorrect_source(<<~RUBY)
+      case foo
+      when bar
+        then do_something
+      end
+    RUBY
+
+    expect(new_source).to eq(<<~RUBY)
+      case foo
+      when bar
+       do_something
+      end
+    RUBY
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Changes
 
 * [#7446](https://github.com/rubocop-hq/rubocop/issues/7446): Add `merge` to list of non-mutating methods. ([@cstyles][])
+* [#7434](https://github.com/rubocop-hq/rubocop/issues/7434): Fix an incorrect autocorrect for `Style/MultilineWhenThen` when the body of `when` branch starts with `then`. ([@koic][])
 
 ## 0.75.1 (2019-10-14)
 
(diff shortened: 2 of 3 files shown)


spec/rubocop/cop/style/multiline_when_then_spec.rb
```diff
@@ -84,4 +84,21 @@
...
+      case foo
+      when bar
+        then do_something
+      end
+    RUBY
+
+    expect(new_source).to eq(<<~RUBY)
+      case foo
+      when bar
+       do_something
```
koic: 📝 There is a slight difference in indentation using only this cop. That indentation is corrected using `Layout/IndentationWidth` cop.

bbatsov:
🚀

PR #7443, opened 2019-10-18
Accept braces for hashes with kwsplat in `Style/BracesAroundHashParamerers`

gsamokovarov:
I was writing metrics code recently and wanted to splat extra data into
the last argument call of a method and rubocop wouldn't let me do that.

I run rubocop on save and my code got it autocorrected from:

```ruby
metrics_for(event, {data: values, **extra})
```

to:

```ruby
metrics_for(event, data: values, **extra)
```

The code above ☝️  has different semantics than the one I need, though,
so the braces are actually needed.

### DIFF
--- a/lib/rubocop/cop/style/braces_around_hash_parameters.rb
+++ b/lib/rubocop/cop/style/braces_around_hash_parameters.rb
@@ -109,11 +109,14 @@ module RuboCop
         # and that block uses do..end. The reason for wanting to check this is
         # that the do..end could bind to a different method invocation if the
         # hash braces were removed.
+        #
+        # Another case for true return value is when you splat extra keyword
+        # arguments into the last argument like: `f({val: 1, **extra})`.
         def braces_needed_for_semantics?(arg)
           arg.each_pair do |_key, value|
             return true if value.block_type? && !value.braces?
           end
-          false
+          arg.children.any?(&:kwsplat_type?)
         end
 
         def add_arg_offense(arg, type)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/braces_around_hash_parameters.rb
```diff
@@ -109,11 +109,14 @@ def check_context_dependent(arg, args)
         # and that block uses do..end. The reason for wanting to check this is
         # that the do..end could bind to a different method invocation if the
         # hash braces were removed.
+        #
+        # Another case for true return value is when you splat extra keyword
+        # arguments into the last argument like: `f({val: 1, **extra})`.
         def braces_needed_for_semantics?(arg)
           arg.each_pair do |_key, value|
```
gsamokovarov: `#each_pair` does not iterate over the splat, that's why I re-iterate looking for it with the `#children` call below.

buehmann:
Hi Genadi, I just looked at the ASTs for the two lines of code above and they are identical:
```console
$ ruby-parse -e 'metrics_for(event, {data: values, **extra})'
(send nil :metrics_for
  (send nil :event)
  (hash
    (pair
      (sym :data)
      (send nil :values))
    (kwsplat
      (send nil :extra))))
$ ruby-parse -e 'metrics_for(event, data: values, **extra)'
(send nil :metrics_for
  (send nil :event)
  (hash
    (pair
      (sym :data)
      (send nil :values))
    (kwsplat
      (send nil :extra))))
```
So what exactly is the difference in semantics you are concerned about?

gsamokovarov:
Holy moly! I didn't expect this for a signature like `def metric_for(event, metadata, **extra)` but trying it out, it seems to be a consistent behavior. I'll come back around to this for Ruby 3 when the kwargs changes are finalized and test it then.

I run rubocop on save and the code was transitional, it didn't execute at the end to suspect it's not how I expect it to be, just wrote it down and assume I can fix the autocorrection.

PR #7448, opened 2019-10-19
[Fix #7446] Add `merge` to `NONMUTATING_METHODS`

### LINKED ISSUE #7446: How to detect when the result of Array#sort isn't captured?

Apologies if this already exists but I couldn't find anything by searching. Is there a way to report when the result of something like `Array#sort` or `Hash#merge` isn't captured?

I'd like to see a cop that will detect when these types of methods are called but their return value isn't stored in a variable, passed as an argument, or used in some other way. In that case, I would like it if `rubocop` would suggest `Array#sort!` or `Hash#merge!` to do the operation in-place.

edit: It occurred to me after the fact that it's not always apparent that we're calling `sort` or `merge` on an `Array` or `Hash` and we wouldn't necessarily want to suggest `sort!` for other (possibly user-defined) classes where `sort` might have side effects. So I don't think this is easy to implement, at least not for a dynamically typed language like Ruby.

cstyles:
Fixes #7446

Now, if `merge` is used in a void context, the `Lint/Void` cop will
register an offense and the mutating `merge!` method will be suggested.

-----------------

* [ ] Added tests.


bbatsov:
Thanks!

PR #7449, opened 2019-10-20
Make IfUnlessModifier respect rubocop:disable comments

jonas054:
A bug was discovered in https://github.com/rubocop-hq/rubocop/pull/7308#issuecomment-544236793

When `Style/IfUnlessModifier` decides that a line on modifier form is too long and that it should be written on normal form, it must respect comments that disable the `Metrics/LineLength` cop for the line in question. If that cop is disabled, then there can be no offense.


koic:
Thanks!

PR #7450, opened 2019-10-21
Reduce ABC complexity

buehmann:
I picked a few of the `Metrics/AbcSize` offenses from the todo list and reduced the complexity of the methods.

PR #7453, opened 2019-10-23
Bring ABC size down to 16

Drenmi:
I wanted to cement the excellent work done by @buehmann in #7450 by bringing the bar down to 16. That way we can't have new offenses creeping in without notice.

-----------------

* [X] Added tests.


lib/rubocop/node_pattern.rb
```diff
@@ -429,6 +429,7 @@ def compile_ellipsis
         [0..Float::INFINITY, 'true']
       end
 
+      # rubocop:disable Metrics/AbcSize
```
Drenmi: Couldn't find any way to refactor this method that didn't make it worse. Pretty sure this is as good as it gets with the current design, but open to suggestions.

bbatsov:
🙇

lib/rubocop/node_pattern.rb
buehmann: Challenge accepted 😉 https://github.com/rubocop-hq/rubocop/pull/7457

lib/rubocop/cop/metrics/utils/abc_size_calculator.rb
```diff
@@ -55,11 +54,26 @@ def evaluate_branch_nodes(node)
...
           end
 
-          def node_has_else_branch?(node)
+          def evaluate_condition_node(node)
+            @condition += 1 if else_branch?(node)
+            @condition += 1
+          end
+
+          def else_branch?(node)
             %i[case if].include?(node.type) &&
```
Drenmi: @jonas054 I suspect we could have a subtle bug here. We check for `#else_branch?` on `CONDITION_NODES`, which comes from `CyclomaticComplexity::COUNTED_NODES`, but that list does not include `case`:

```ruby
COUNTED_NODES = %i[if while until for
                   rescue when and or].freeze
```

jaredbeck:
> Reduce ABC size of AbcSizeCalculator (5eb411b)

so meta

PR #7457, opened 2019-10-23
Reduce ABC size of NodePattern#compile_any_order

buehmann:
:wink: @Drenmi: This feels like cheating …


Drenmi:
@buehmann I accept this solution! It actually uses the `#tokens_until` method the way it was designed. 😀

bbatsov:
Nicely done!

PR #7459, opened 2019-10-23
Added new cop - Naming/AsciiConstants

pjskennedy:
This adds a new cop: `Naming/AsciiConstants`

This is an attempt to implement https://github.com/rubocop-hq/rubocop/issues/7458

Rubocop currently has [Naming/AsciiIdentifiers](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/AsciiIdentifiers) which correctly adds offenses to ruby identifiers (`tIDENTIFIER`) containing non-ascii characters. I find that the same rigor should be applied to constants (`tCONSTANT`) as the following are currently not offenses:
```rb
class Foö
end
```
```rb
module Foö
end
```
```rb
FOÖ = "foo"
```

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+### New features
+* "Add new `Naming/AsciiConstants` cop. ([@pjskennedy][])
+
 ### Bug fixes
 
 * [#7439](https://github.com/rubocop-hq/rubocop/issues/7439): Make `Style/FormatStringToken` ignore percent escapes (`%%`). ([@buehmann][])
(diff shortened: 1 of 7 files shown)


bbatsov:
The proposed cop is fine by me. @rubocop-hq/rubocop-core any objections to including it enabled by default?

koic:
Both this cop and `Naming/AsciiIdentifiers` cop conform to The Ruby Style Guide's "English Identifiers" rule. And It seems that most implementations overlap with `Naming/AsciiIdentifiers` cop. 
https://github.com/rubocop-hq/ruby-style-guide#english-identifiers

I think there is a way to enhance the `Naming/AsciiIdentifiers` cop instead of creating a new cop.
So, for example, it may be better to provide `ascii_constants: true (default)` option or `ascii_constants: false (default)` option.

bbatsov:
@koic I like this idea!

bbatsov:
@pjskennedy I'll close this PR, so we can pursue @koic's idea. Thanks for your contribution, though.

PR #7461, opened 2019-10-25
Bring ABC size down to 15 (default)

buehmann:
@Drenmi After your #7453, let's go all the way …

There are two places where I resorted to disabling `Metrics/AbcSize`: One is the well-known beast in `NodePattern`, which needs some deeper refactoring. The other one is a longer, but straight-forward case statement in `Layout/RescueEnsureAlignment#alignment_source`.

On the way, I found and fixed an off-by-one error in `Metrics/LineLength`, which could be extracted into a separate PR if you deem it necessary.

Have a nice weekend!


bbatsov:
Fantastic work! 💯 🥇 🙇 

Have a great weekend as well!

Drenmi:
I ... almost feel a bit empty inside. 😂 Luckily there's plenty of things still to improve. 😬 

Some beastly work @buehmann! 🚀

PR #7464, opened 2019-10-27
Add forward slash to LITERAL_REGEX

eugeneius:
The forward slash character has no special meaning when it appears in a regex. This will let `Performance/StartWith` and `Performance/EndWith` correct code like the following:

```ruby
path =~ %r{/\z}
```

To:

```ruby
path.end_with?('/')
```

I noticed this problem thanks to @eregon's comment here: https://github.com/rails/rails/pull/37504#discussion_r339314567

`LITERAL_REGEX` is only used in RuboCop Performance, so I couldn't add a test to demonstrate the new behaviour. Arguably the constant should be defined there instead, but removing it from RuboCop would introduce an incompatibility with all existing versions of RuboCop Performance. I opted to update the constant in-place; let me know if another approach is preferable.

-----------------

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [#7449](https://github.com/rubocop-hq/rubocop/pull/7449): Make `Style/IfUnlessModifier` respect `rubocop:disable` comments for `Metrics/LineLength`. ([@jonas054][])
 * [#7442](https://github.com/rubocop-hq/rubocop/issues/7442): Fix an incorrect autocorrect for `Style/SafeNavigation` when an object check followed by a method call with a comment at EOL. ([@koic][])
 * [#7434](https://github.com/rubocop-hq/rubocop/issues/7434): Fix an incorrect autocorrect for `Style/MultilineWhenThen` when the body of `when` branch starts with `then`. ([@koic][])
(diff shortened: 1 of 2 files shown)


bbatsov:
> LITERAL_REGEX is only used in RuboCop Performance, so I couldn't add a test to demonstrate the new behaviour. Arguably the constant should be defined there instead, but removing it from RuboCop would introduce an incompatibility with all existing versions of RuboCop Performance. I opted to update the constant in-place; let me know if another approach is preferable.

@koic I guess we should just move this constant to the `rubocop-performance` project then.

koic:
@bbatsov Yeah. That seems good to me.

@eugeneius Could you make the constant `LITERAL_REGEX` to RuboCop Performance and add  tests?
As you are concerned, combining RuboCop 0.75.1 or lower with RuboCop Performance 1.15.0 or lower will result in an incompatibility error.

>  I opted to update the constant in-place; let me know if another approach is preferable.

So, I'd like to leave this constant in RuboCop core for a while. Perhaps the constant will be removed after future release of RuboCop 0.76 and higher.
(Perhaps `TracePoint`, etc is used, it seems technically possible to display a warning about breaking change, but it is not a good way to do it 😅 )

bbatsov:
Agreed. I guess we should fix this in RuboCop for now and move the constant to `rubocop-performance` down the road.

PR #6466, opened 2018-11-10
Update terminology in http_status.rb

JuanitoFatas:
This Pull Request updates `lib/rubocop/cop/rails/http_status.rb`.

Permitted is easier to understand and better terminology. I didn't find anyone [using this constant](https://github.com/search?q=%22NumericStyleChecker%3A%3AWHITELIST_STATUS%22&type=Code), it should be safe to change.

[Original motivation](https://github.com/rails/rails/issues/33677), other community efforts examples:

- https://github.com/rails/rails/pull/33681
- https://github.com/graphiti-api/graphiti/pull/10
- https://github.com/rails/rails/pull/33718
- https://github.com/ruby/psych/pull/378
- https://github.com/ruby/ruby/pull/2008
- https://github.com/ruby/ruby/pull/2009
- https://github.com/rubocop-hq/rubocop/pull/6464
- https://github.com/rubygems/rubygems/pull/2463
- https://github.com/dtao/safe_yaml/pull/93
- https://github.com/rack/rack/pull/1314
- https://github.com/rubocop-hq/rubocop/pull/6467
- https://github.com/rspec/rspec-core/pull/2576
- https://github.com/rspec/rspec-expectations/pull/1083
- https://github.com/rspec/rspec-mocks/pull/1246
- https://github.com/rspec/rspec-support/pull/356
- https://github.com/flavorjones/loofah/pull/158
- https://github.com/pry/pry/pull/1874
- https://github.com/pry/pry/pull/1875

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [ ] Squashed related commits together.
* [ ] Added tests.
* [ ] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](../blob/master/CONTRIBUTING.md#changelog-entry-format).
* [ ] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.
* [x] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: https://chris.beams.io/posts/git-commit/

PR #7469, opened 2019-10-29
Change terminology to `Allowlist` and `Denylist`

koic:
### Summary

Follow up #6464, #6466, and #6467.

This PR changes a terminology from `Whitelist` and `Blacklist` to `Allowlist` and `Denylist`.

This change is an obvious breaking change to some cop options. So I'd like to introduce it before RuboCop 1.0 if this change is acceptable.

### Other Information

This change has also been made in Rails repository and other repos.
https://github.com/rails/rails/pull/33681

-----------------

* [ ] Added tests.


bbatsov:
I agree with the overall premise and the timing of the change, but I think we can do better with the naming - e.g. instead of naming something `Allowlist` it can be `AllowedPatterns` or something like this. I'll leave a couple of inline remarks.

config/default.yml
```diff
@@ -141,11 +141,12 @@ Bundler/GemComment:
   Description: 'Add a comment describing each gem.'
   Enabled: false
   VersionAdded: '0.59'
+  VersionChanged: '0.77'
   Include:
     - '**/*.gemfile'
     - '**/Gemfile'
     - '**/gems.rb'
-  Whitelist: []
+  Allowlist: []
```
bbatsov: I guess here we can use something like `IgnoredGems` that's more descriptive.

config/default.yml
```diff
@@ -1580,8 +1581,8 @@ Lint/SafeNavigationChain:
   Description: 'Do not chain ordinary method call after safe navigation operator.'
   Enabled: true
   VersionAdded: '0.47'
-  VersionChanged: '0.56'
-  Whitelist:
+  VersionChanged: '0.77'
+  Allowlist:
```
bbatsov: This can be `AllowedMethods`.

bbatsov: (or `AcceptedMethods`)

config/default.yml
```diff
@@ -1595,7 +1596,8 @@ Lint/SafeNavigationConsistency:
                  for all method calls on that same object.
   Enabled: true
   VersionAdded: '0.55'
-  Whitelist:
+  VersionChanged: '0.77'
+  Allowlist:
```
bbatsov: Ditto.

config/default.yml
```diff
@@ -2009,20 +2012,20 @@ Naming/PredicateName:
...
     - has_
     - have_
   # Predicate name prefixes that should be removed.
-  NamePrefixBlacklist:
+  NamePrefixDenylist:
     - is_
     - has_
     - have_
-  # Predicate names which, despite having a blacklisted prefix, or no `?`,
+  # Predicate names which, despite having a restricted list prefix, or no `?`,
```
bbatsov: `list` is redundant here

config/default.yml
```diff
@@ -2009,20 +2012,20 @@ Naming/PredicateName:
...
-  VersionChanged: '0.51'
+  VersionChanged: '0.77'
   # Predicate name prefixes.
   NamePrefix:
     - is_
     - has_
     - have_
   # Predicate name prefixes that should be removed.
-  NamePrefixBlacklist:
+  NamePrefixDenylist:
```
bbatsov: `ForbiddenPrefixes` sounds a lot better to me.

config/default.yml
```diff
@@ -2009,20 +2012,20 @@ Naming/PredicateName:
...
     - is_
     - has_
     - have_
-  # Predicate names which, despite having a blacklisted prefix, or no `?`,
+  # Predicate names which, despite having a restricted list prefix, or no `?`,
   # should still be accepted
-  NameWhitelist:
+  NameAllowlist:
```
bbatsov: `Allowed/PermittedMethods`

config/default.yml
```diff
@@ -2049,9 +2052,9 @@ Naming/UncommunicativeBlockParamName:
   # Parameter names may be equal to or greater than this value
   MinNameLength: 1
   AllowNamesEndingInNumbers: true
-  # Whitelisted names that will not register an offense
+  # Permitted names that will not register an offense
```
bbatsov: You could have used `allowed` and `forbidden` here as the names of the configs.

config/default.yml
```diff
@@ -2064,7 +2067,7 @@ Naming/UncommunicativeMethodParamName:
   # Parameter names may be equal to or greater than this value
   MinNameLength: 3
   AllowNamesEndingInNumbers: true
-  # Whitelisted names that will not register an offense
```
bbatsov: Ditto.

config/default.yml
```diff
@@ -3148,8 +3152,8 @@ Style/NestedParenthesizedCalls:
                  argument list of another parenthesized method call.
   Enabled: true
   VersionAdded: '0.36'
-  VersionChanged: '0.50'
-  Whitelist:
+  VersionChanged: '0.77'
+  Allowlist:
```
bbatsov: For this and the remaining options - something like `AllowedMethods` sounds better.

bbatsov:
@rubocop-hq/rubocop-core Please add your feedback here as well, if you have some. My general point is that we should try to do something more than just auto-replacing the existing problematic names. From my perspective the use of generic names like `whitelist/allowlist` also has an usability impact, as it's never clear a list of what we're referring to.

config/default.yml
```diff
@@ -3526,11 +3530,11 @@ Style/SafeNavigation:
                   safe navigation (`&.`).
   Enabled: true
   VersionAdded: '0.43'
-  VersionChanged: '0.56'
+  VersionChanged: '0.77'
   # Safe navigation may cause a statement to start returning `nil` in addition
   # to whatever it used to return.
   ConvertCodeThatCanStartToReturnNil: false
-  Whitelist:
+  Allowlist:
```
koic: This name will change `AllowedMethods`.

Drenmi:
> Please add your feedback here as well, if you have some. My general point is that we should try to do something more than just auto-replacing the existing problematic names. From my perspective the use of generic names like whitelist/allowlist also has an usability impact, as it's never clear a list of what we're referring to.

I have nothing against replacing `whitelist` and `blacklist`, but I agree that there are better contextual options in most cases. 👍

koic:
I brushed up this PR. I would be happy to get suggestions if you have other good names :-)

config/default.yml
```diff
@@ -1972,7 +1973,8 @@ Naming/HeredocDelimiterNaming:
   StyleGuide: '#heredoc-delimiters'
   Enabled: true
   VersionAdded: '0.50'
-  Blacklist:
+  VersionChanged: '0.77'
+  Denylist:
```
bbatsov: This should probably be `ForbiddenDelimiters/Patterns`. I'm not sure if the config supports both strings and regular expressions.

config/default.yml
```diff
@@ -2874,8 +2876,9 @@ Style/IpAddresses:
   Description: "Don't include literal IP addresses in code."
   Enabled: false
   VersionAdded: '0.58'
-  # Allow strings to be whitelisted
-  Whitelist:
+  VersionChanged: '0.77'
+  # Allow strings to be permitted
+  Allowlist:
```
bbatsov: I guess here were should have `AllowedAddresses` or something like this.

config/default.yml
```diff
@@ -3829,7 +3832,7 @@ Style/TrivialAccessors:
   # Commonly used in DSLs
   AllowDSLWriters: false
   IgnoreClassMethods: false
-  Whitelist:
+  AllowMethods:
```
bbatsov: This should be `AllowedMethod` to be consistent with our other configs.

config/default.yml
koic: Sure. Both are acceptable because they are replaced internally by regular expressions :-) I will change to `ForbiddenDelimiters`.
https://github.com/rubocop-hq/rubocop/blob/v0.76.0/lib/rubocop/cop/naming/heredoc_delimiter_naming.rb#L45

bbatsov:
Thanks!

config/default.yml
```diff
@@ -3840,7 +3842,7 @@ Style/TrivialAccessors:
   # Commonly used in DSLs
   AllowDSLWriters: false
   IgnoreClassMethods: false
-  Whitelist:
+  AllowedMethod:
```
movermeyer: Seems that this is missing the trailing `s`.
Should be `AllowedMethods`.

PR incoming.

movermeyer: Issue: https://github.com/rubocop-hq/rubocop/issues/7530
PR: https://github.com/rubocop-hq/rubocop/pull/7529

PR #7473, opened 2019-10-31
Extract commands from CLI

buehmann:
In my opinion, the CLI class is too big. So I took a stab at splitting it up into commands. What do you think?


Drenmi:
Looks a lot more manageable to me. 👍

Could you regenerate the TODO as part of this change? I imagine this fixes some of the offenses there.

Is all this code tested using `RuboCop::CLI` as an entry point? Would it make sense to split the tests up, too?

buehmann:
I just regenerated the TODO list: It does not change.

And yes, at the moment this is still tested entirely via `Rubocop::CLI`. The specs for `CLI` seem to be written in terms of `CLI#run` in an end-to-end fashion only. So more work is needed to break this down into smaller pieces.

bbatsov:
I like the proposed change as well. Great work! 🙇

PR #7494, opened 2019-11-10
[Fix #7493] Detect redundant returns at end of longer methods

### LINKED ISSUE #7493: Style/RedundantReturn does not find redundant returns in conditional constructs preceded by other code
I noticed that `Style/RedundantReturn` does not detect the redundant returns in code like
```ruby
def func
  x = foo
  if x
    return 1
  else
    return 2
  end
end
```
but only if there is nothing preceding the `if`/`case` expression at the end of a method:
```ruby
def func
  if x
    return 1
  else
    return 2
  end
end
```
```
redundant_return.rb:3:5: C: Style/RedundantReturn: Redundant return detected.
    return 1
    ^^^^^^
redundant_return.rb:5:5: C: Style/RedundantReturn: Redundant return detected.
    return 2
    ^^^^^^
```
(Example taken from https://github.com/rubocop-hq/rubocop/pull/3590 where looking into conditionals was introduced.)

I am going to add a PR fixing this shortly.

buehmann:
This fixes #7493.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#7493](https://github.com/rubocop-hq/rubocop/issues/7493): Fix `Style/RedundantReturn` to inspect conditional constructs that are preceded by other statements. ([@buehmann][])
+
 ### Changes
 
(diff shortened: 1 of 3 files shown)


koic:
Looks good to me. Can you rebase with latest master and squash your commits into one?

buehmann:
Done.

koic:
Thanks always!

PR #7496, opened 2019-11-11
Allow cops to invalidate results cache

maxh:
Internally at Flexport, we've written a few cops to enforce Rails Engine isolation. They've helped us incrementally modularize our system, and we think they might be valuable to others too. I'm working to open source them now.

For example, `GlobalModelAccessFromEngine` (https://github.com/flexport/rubocop-flexport/pull/5) forbids code within Rails Engines from directly accessing global models in the main `app/` directory. Another example is `EngineApiBoundary` (https://github.com/flexport/rubocop-flexport/pull/6).

(Note: we considered upstream these to `rubocop-rails`, and @koic suggested we create our gem instead: https://github.com/rubocop-hq/rubocop-rails/pull/152#issuecomment-554839724)

These cops read from the filesystem, taking a holistic view of our codebase while they inspect. Unfortunately, this approach does not play nice with the RuboCop results cache. As a workaround, we created a custom `lint.rb` that wraps `rubocop` and busts the cache when needed using the `--cache` param. It works ok.

But it's not ideal, especially for broader use. This PR aims to fix the issue at its source by allowing cops themselves to invalidate the cache.

## Example

Consider the following example:

(1) A filesystem contains these files:

```
app/models/my_model.rb
engines/my_engine/app/services/my_engine/my_service.rb
```

(2) We run `rubocop engines/my_engine/app/services/my_engine/my_service.rb` and find a violation -- `service.rb` contains `MyModel.find(123)`. 

(2b) During that run, a cached result is stored, keyed by: the inspected source code, the RuboCop config, command-line options, and executable version.

(3) We run the same command again. Cache is hit, same violation shown. All good.

(4) We move the model file into the engine. So now `app/models/my_model.rb` no longer exists and we have `engines/my_engine/app/models/my_engine/my_model.rb`.

(5) We run the same RuboCop command again. Because none of the cache key inputs changed, _the same violation is shown again_. This is incorrect.

(6) We run the same command again with `--cache false` and see that the violation no longer exists. This is correct.

## Implementation overview

This PR allows cops to define an `external_dependency_checksum` method that busts the cache when their external dependencies change. Cops are responsible for computing their own checksum however they deem appropriate.

Among existing cops, I believe there are no use cases for this method. For the cops we've written, there are two types of external dependencies: (1) the presence or absence of certain files and/or (2) the contents of certain files.

## Discussion

I recognize that this is perhaps an unusual use case for RuboCop. As an alternative, we could create our own repo with these cache-unfriendly cops and package them with a custom runner script that does cache busting inside the script. But that seems suboptimal. And I suspect other cops may make use of this feature in the future as well.

Feedback welcome! Thanks!

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/cop.rb
+++ b/lib/rubocop/cop/cop.rb
@@ -217,6 +217,25 @@ module RuboCop
         !relevant_file?(file)
       end
 
+      # This method should be overriden when a cop's behavior depends
+      # on state that lives outside of these locations:
+      #
+      #   (1) the file under inspection
+      #   (2) the cop's source code
+      #   (3) the config (eg a .rubocop.yml file)
+      #
+      # For example, some cops may want to look at other parts of
+      # the codebase being inspected to find violations. A cop may
+      # use the presence or absence of file `foo.rb` to determine
+      # whether a certain violation exists in `bar.rb`.
+      #
+      # Overriding this method allows the cop to indicate to RuboCop's
+      # ResultCache system when those external dependencies change,
+      # ie when the ResultCache should be invalidated.
+      def external_dependency_checksum
+        nil
+      end
+
       private
 
       def find_message(node, message)
--- a/lib/rubocop/runner.rb
+++ b/lib/rubocop/runner.rb
@@ -121,8 +121,14 @@ module RuboCop
       end
     end
 
+    def cached_result(file, team)
+      ResultCache.new(file, team, @options, @config_store)
+    end
+
     def file_offense_cache(file)
-      cache = ResultCache.new(file, @options, @config_store) if cached_run?
+      config = @config_store.for(file)
+      cache = cached_result(file, standby_team(config)) if cached_run?
+
       if cache&.valid?
         offenses = cache.load
         # If we're running --auto-correct and the cache says there are
@@ -358,6 +364,12 @@ module RuboCop
       else
         ProcessedSource.from_file(file, ruby_version)
       end
+    end
+
+    def standby_team(config)
+      @team_by_config ||= {}
+      @team_by_config[config.object_id] ||=
+        Cop::Team.new(mobilized_cop_classes(config), config, @options)
     end
   end
 end
--- a/lib/rubocop/cop/team.rb
+++ b/lib/rubocop/cop/team.rb
@@ -86,6 +86,11 @@ module RuboCop
         raise e.cause
       end
 
+      def external_dependency_checksum
+        keys = cops.map(&:external_dependency_checksum).compact
+        Digest::SHA1.hexdigest(keys.join)
+      end
+
       private
 
       def offenses(processed_source)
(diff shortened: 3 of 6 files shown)


bbatsov:
I'm fine with the proposed solution, but I'll defer to @jonas054 to evaluate the implementation.

lib/rubocop/cop/cop.rb
```diff
@@ -217,6 +217,25 @@ def excluded_file?(file)
...
       end
 
+      # This method should be overriden when a cop's behavior depends
+      # on state that lives outside of these locations:
+      #
+      #   (1) the file under inspection
+      #   (2) the cop's source code
+      #   (3) the config (eg a .rubocop.yml file)
+      #
+      # For example, some cops may want to look at at other parts of
```
jonas054: at ~at~

lib/rubocop/runner.rb
```diff
@@ -359,5 +365,12 @@ def get_processed_source(file)
         ProcessedSource.from_file(file, ruby_version)
       end
     end
+
+    def standby_team(config)
+      @team_by_config ||= {}
+      @team_by_config[config.object_id] ||= begin
```
jonas054: I think the `begin`..`end` is unnecessary here.

maxh: Thanks! Fixed in https://github.com/rubocop-hq/rubocop/commit/e7fa9a916082e463ee803f979a5ceed55791c1c5

lib/rubocop/cop/cop.rb
maxh: Thanks! Fixed in https://github.com/rubocop-hq/rubocop/commit/e7fa9a916082e463ee803f979a5ceed55791c1c5

maxh:
Here's a blog post that goes into more detail about the Rails Engine cops mentioned in the PR description:

https://flexport.engineering/isolating-rails-engines-with-rubocop-210feaba3164

When this is merged, I will be able to full upstream those cops. Thanks!

bbatsov:
Thanks!

lib/rubocop/cop/cop.rb
```diff
@@ -217,6 +217,25 @@ def excluded_file?(file)
...
+      # For example, some cops may want to look at other parts of
+      # the codebase being inspected to find violations. A cop may
+      # use the presence or absence of file `foo.rb` to determine
+      # whether a certain violation exists in `bar.rb`.
+      #
+      # Overriding this method allows the cop to indicate to RuboCop's
+      # ResultCache system when those external dependencies change,
+      # ie when the ResultCache should be invalidated.
+      def external_dependency_checksum
+        nil
```
exterm: If you run the cop on `n` files, this method will be called every time, recomputing the checksum `n` times, correct?

Maybe it would be useful for rubocop to have a "run" context that keeps state across files and is accessible from a cop.

maxh: Good question -- please see https://github.com/rubocop-hq/rubocop/pull/7543. I believe these results should be cached per team/config, so don't need to be recomputed per inspected file.

PR #7497, opened 2019-11-11
[Fix #7477] Fix semicolon line length autocorrection

### LINKED ISSUE #7477: Metrics/LineLength auto-correct breaks line in the wrong place
Hi,

I have enabled auto-correction of `Metrics/LineLength` that was introduced in 0.68. I have also enabled the other [recommended cops](https://docs.rubocop.org/en/stable/cops_metrics/#metricslinelength) that were not enabled by default.

## Expected behavior

When I run auto-correct, the code should still be valid Ruby code.

## Actual behavior

In one instance, a line break was injected in the middle of an integer literal.

I have re-created the issue in a smaller example and this is the diff:

```diff
diff --git a/rubocop_line_length.rb b/rubocop_line_length.rb
index c505caa..01762f0 100644
--- a/rubocop_line_length.rb
+++ b/rubocop_line_length.rb
@@ -1,7 +1,9 @@
 class RubocopLineLength
   def breaks_this_incorrectly
     inventory_rows_to_fix = [
-      InventoryRow.new(year: 2019, week: 30, path: 'publisher_group:123;publisher:456;slot_type:midroll', to_book: 21906),
+      InventoryRow.new(year: 2019, week: 3
+0, path: 'publisher_group:123;
+publisher:456;slot_type:midroll', to_book: 21906),
       InventoryRow.new(year: 2019, week: 31, path: 'publisher_group:123;slot_type:preroll', to_book: 1558)
     ]
   end

```

Create `.rubocop.yml`:

```
Metrics/LineLength:
  AutoCorrect: true

Layout/MultilineArrayLineBreaks:
  Enabled: true

Layout/MultilineHashKeyLineBreaks:
  Enabled: true

Layout/MultilineMethodArgumentLineBreaks:
  Enabled: true
```

Create `rubocop_line_length.rb`:

```ruby
class RubocopLineLength
  def breaks_this_incorrectly
    inventory_rows_to_fix = [
      InventoryRow.new(year: 2019, week: 30, path: 'publisher_group:123;publisher:456;slot_type:midroll', to_book: 21906),
      InventoryRow.new(year: 2019, week: 31, path: 'publisher_group:123;slot_type:preroll', to_book: 1558)
    ]
  end
end
```

Run `rubocop --only Layout,Metrics/LineLength --auto-correct rubocop_line_length.rb`

## RuboCop version

```
$ rubocop -V
0.76.0 (using Parser 2.6.5.0, running on ruby 2.6.5 x86_64-linux)
```

maxh:
Fixes https://github.com/rubocop-hq/rubocop/issues/7477. Previously the cop naively checked for the character `;` anywhere in the source code, which would match semicolons within string literals and cause correctness issues.

Now the cop instead uses the tokenizer output to find semicolons.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 * [#7469](https://github.com/rubocop-hq/rubocop/pull/7469): **(Breaking)** Replace usages of the terms `Whitelist` and `Blacklist` with better alternatives. ([@koic][])
 * [#7502](https://github.com/rubocop-hq/rubocop/pull/7502): Remove `SafeMode` module. ([@koic][])
 
+### Bug fixes
+
+* [#7477](https://github.com/rubocop-hq/rubocop/issues/7477): Fix line length autocorrect for semicolons in string literals. ([@maxh][])
+
 ## 0.76.0 (2019-10-28)
 
 ### Bug fixes
--- a/lib/rubocop/cop/metrics/line_length.rb
+++ b/lib/rubocop/cop/metrics/line_length.rb
@@ -68,6 +68,10 @@ module RuboCop
         alias on_hash on_potential_breakable_node
         alias on_send on_potential_breakable_node
 
+        def investigate(processed_source)
+          check_for_breakable_semicolons(processed_source)
+        end
+
         def investigate_post_walk(processed_source)
           processed_source.lines.each_with_index do |line, line_index|
             check_line(line, line_index)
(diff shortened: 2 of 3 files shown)


bbatsov:
I'm wondering if it's OK to auto-correct anything in a string - after all you don't really know whether something is a code snippet there or not, right?

maxh:
>I'm wondering if it's OK to auto-correct anything in a string - after all you don't really know whether something is a code snippet there or not, right?

Autocorrect should never change the internals of a string (at least for corrections of line length). The current code has bug where it changes string literals when the following are true:

- autocorrect is enabled for this cop
- a line is too long
- that line contains a string [0]
- that string contains a semicolon

Only the semicolon-related part of autocorrect has this bug, because it matches on raw characters. The rest of the line length autocorrect code operates only on AST nodes so never affects string internals. 

This PR switches the semicolon part to use lexed tokens instead of raw chars. I've confirmed in specs that string literals are not be autocorrected anymore after this change, because eg this string literal is lexed as a single token:

`"00000000000000000;0000000000000000000'000000;00000'0000;0000;000"`
```
[#<RuboCop::Token:0x00007fe3dcc62b50
    @pos=#<Parser::Source::Range (string) 0...66>,
    @text="00000000000000000;0000000000000000000'000000;00000'0000;0000;000",
    @type=:tSTRING>,
   #<RuboCop::Token:0x00007fe3dcc62b28 @pos=#<Parser::Source::Range (string) 66...67>, @text="", @type=:tNL>]
```

[0] Note that this bug doesn't affect heredocs because special support was added to protect them. that's no longer needed with token-based approach and has been removed in this PR.

maxh:
I've run this on a big codebase and confirms it behaves correctly -- good to go!

Aside: once this is fixed, I think we should consider enabling LineLength autocorrection by default since it's been stable for several months now and this is the only major issue I've seen reported.

bbatsov:
Yeah, that's a fair point. I'm fine with enabling it by default. Thanks for tackling this!

PR #7507, opened 2019-11-15
Fix IfUnlessModifier's co-operation with LineLength in edge cases

jonas054:
There's a lot of complexity in how `Metrics/LineLength` handles various exceptions regarding long lines. The best way for `Style/IfUnlessModifier` to avoid false positives on long lines is to share code with `Metrics/LineLength` and thus always have the same idea about which lines are too long.


jonas054:
Rebased and resolved conflicts. I would love some reviews. :)
Cc: @jfelchner @maxim

lib/rubocop/cop/mixin/line_length_help.rb
```diff
@@ -0,0 +1,94 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # Help methods for determining if a line is too long.
+    class LineLengthHelp
```
bbatsov: I'd use a different name here, as we haven't used the SomethingHelp pattern anywhere else. I also noticed you haven't used this class as a mixin, even though it's in the mixin dir.

jonas054: There are actually a couple of mixin modules with names ending in `Help`: `RangeHelp`, `StringHelp`, and `StringLiteralsHelp`. Only the last one is in the `mixin` directory, but I think that's because it's the only one used from multiple departments. So since I can't think of a better name than `Help`, and since there is some precedent, I propose to keep the name.

The placement of a class in `mixin` is not ideal, I can see that. There is a `utils` namespace where we could put it, but I'm not sure if that's better. Maybe I can just discard the commit that changes the module into a class? Using composition instead of inheritance doesn't make a huge difference in this case.

bbatsov: > There are actually a couple of mixin modules with names ending in Help: RangeHelp, StringHelp, and StringLiteralsHelp. Only the last one is in the mixin directory, but I think that's because it's the only one used from multiple departments. So since I can't think of a better name than Help, and since there is some precedent, I propose to keep the name.

Ah, I totally forgot about those. Fair enough. 

> The placement of a class in mixin is not ideal, I can see that. There is a utils namespace where we could put it, but I'm not sure if that's better. Maybe I can just discard the commit that changes the module into a class? Using composition instead of inheritance doesn't make a huge difference in this case.

I'll leave this up to you. I don't have any strong preferences here. If it weren't for the delegation I'd say this was an util, but now it's hard to decide. OOP is way too complex. :-)

jfelchner: I just took a look and it looks great. And for what it's worth I generally prefer composition over inheritance. ;)

jfelchner:
@jonas054 thank you for this. ❤

jonas054:
@bbatsov  I removed @koic as reviewer now that you and @jfelchner have both looked at it. I addressed your concerns about the class placed under `mixin` by making it a module. Then I forgot about this PR. :slightly_smiling_face:  Should be OK to merge now, I think.

bbatsov:
Agreed. Thanks!

PR #7514, opened 2019-11-19
Exposing correctable status on offense and in formatters

tyler-ball:
We're working on a project to run Rubocop on a list of Chef Cookbooks
and report information back to the user. One of the fields we need to
expose in addition to 'corrected' is 'correctable'.

It seemed to me, looking at the logic, that the `:unsupported` status
is the only one which indicates it _cannot_ be autocorrected.

Questions:

1. Is my logic correct here? Or would it make more sense to pass in the status of `.support_autocorrect?` from the cop to the offense?
1. If this seems an appropriate attribute to expose, would you like me to add it to the other formatters? Right now we only need the JSON formatter but if this makes sense I can add it to the other formatters.

-----------------

* [x] Added tests.


bbatsov:

> Is my logic correct here? Or would it make more sense to pass in the status of .support_autocorrect? from the cop to the offense?

I guess it doesn't make much of a difference where we take the data from. @jonas054 do you agree? 

> If this seems an appropriate attribute to expose, would you like me to add it to the other formatters? Right now we only need the JSON formatter but if this makes sense I can add it to the other formatters.

I don't think this data would be very useful in the other formatters, but if someone has use-cases in mind let's discuss those.

bbatsov:
I guess no one has any further remarks, so just rebase on top of `master` and move the changelog entry to account for the release of 0.77.

bbatsov:
@tyler-ball ping :-)

PR #7518, opened 2019-11-20
[Fix #7517] Layout/WhitespaceAroundKeyword: allow `super::const`

### LINKED ISSUE #7517: Layout/SpaceeAroundKeyword should allow `super::ConstName`
*Be clear, concise and precise in your description of the problem.
Open an issue with a descriptive title and a summary in grammatically correct,
complete sentences.*

*Use the template below when reporting bugs. Please, make sure that
you're running the latest stable RuboCop and that the problem you're reporting
hasn't been reported (and potentially fixed) already.*

*Before filing the ticket you should replace all text above the horizontal
rule with your own words.*

--------

## Expected behavior

The `Layout/SpaceAroundKeyword` cop should allow code such as `super::MyModule` and `super::MY_CONST`.

A common example of this construct could be in Rails to dynamically refer to modules such as `SomeActiveRecordClass::GeneratedAssociationMethods`

My specific use case is simply a plain-ruby method that returns a module being overridden by subclasses to return a dynamically-generated child module:

```
def storage_class
  secure? ? super::Secure : super
end
```

## Actual behavior

> Layout/SpaceAroundKeyword: Space after keyword `super` is missing.

Run rubocop over a file with the following contents:

```
# this part is just for fun; option to skip if these are already defined :-)
module RuboCop
  module Cop
    module Layout
    end
  end
end

class C
  def mod
    return RuboCop::Cop
  end
end

class D < C
  def mod
    super::Layout
  end
end
```

## RuboCop version

```
0.76.0 (using Parser 2.6.5.0, running on ruby 2.6.1 x86_64-darwin18)
```

ozydingo:
The `Layout/WhitespaceAroundKeyword` cop complains when `super::MyConst` is used. However, in Ruby, modules and classes are normal objects and can be returns by methods including by using `super`. Therefore, `super::MyConst` should be considered valid style.

This type of usage is distinct from the intention of this cop, which is to avoid missing space such as around `something = 123if test` and `while(something)`.

I considered adding a test that verifies that the cop complains when a keyword other than `super` is used in the same manner (it does). However, I believe this is a syntax error in every other case and so the behavior of this cop is perhaps best left undefined. The alternative would be to suggest that `yield ::` is the correct form of `yield::`, which is false; they're both wrong, and it's not the responsibility of this cop to know that.

-----------------

* [✓ ] Wrote [good commit messages][1].
* [✓ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [✓ ] Feature branch is up-to-date with `master` (if not - rebase it).
* [✓ ] Squashed related commits together.
* [✓ ] Added tests.
* [✓ ] Added an entry to the [Changelog](https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format).
* [✓ ] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [✓ ] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

### DIFF
--- a/lib/rubocop/cop/layout/space_around_keyword.rb
+++ b/lib/rubocop/cop/layout/space_around_keyword.rb
@@ -30,10 +30,12 @@ module RuboCop
 
         DO = 'do'
         SAFE_NAVIGATION = '&.'
+        NAMESPACE_OPERATOR = '::'
         ACCEPT_LEFT_PAREN =
           %w[break defined? next not rescue return super yield].freeze
         ACCEPT_LEFT_SQUARE_BRACKET =
           %w[super yield].freeze
+        ACCEPT_NAMESPACE_OPERATOR = 'super'
 
         def on_and(node)
           check(node, [:operator].freeze) if node.keyword?
@@ -193,6 +195,8 @@ module RuboCop
 
           return false if accepted_opening_delimiter?(range, char)
           return false if safe_navigation_call?(range, pos)
+          return false if accept_namespace_operator?(range) &&
+                          namespace_operator?(range, pos)
 
           char !~ /[\s;,#\\\)\}\]\.]/
         end
@@ -212,8 +216,16 @@ module RuboCop
           ACCEPT_LEFT_SQUARE_BRACKET.include?(range.source)
         end
 
+        def accept_namespace_operator?(range)
+          ACCEPT_NAMESPACE_OPERATOR == range.source
+        end
+
         def safe_navigation_call?(range, pos)
           range.source_buffer.source[pos, 2].start_with?(SAFE_NAVIGATION)
+        end
+
+        def namespace_operator?(range, pos)
+          range.source_buffer.source[pos, 2].start_with?(NAMESPACE_OPERATOR)
         end
 
         def preceded_by_operator?(node, _range)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#7493](https://github.com/rubocop-hq/rubocop/issues/7493): Fix `Style/RedundantReturn` to inspect conditional constructs that are preceded by other statements. ([@buehmann][])
 * [#7509](https://github.com/rubocop-hq/rubocop/issues/7509): Fix `Layout/SpaceInsideArrayLiteralBrackets` to correct empty lines. ([@ayacai115][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/layout/space_around_keyword.rb
```diff
@@ -212,10 +217,18 @@ def accept_left_square_bracket?(range)
           ACCEPT_LEFT_SQUARE_BRACKET.include?(range.source)
         end
 
+        def accept_namesapce_operator?(range)
```
buehmann: Typo: `sapce`

ozydingo: Fixed. I already squashed and force pushed it; please let me know if the preference is to use separate commits and squash after approval.

lib/rubocop/cop/layout/space_around_keyword.rb
```diff
@@ -30,10 +30,13 @@ class SpaceAroundKeyword < Cop
 
         DO = 'do'
         SAFE_NAVIGATION = '&.'
+        NAMESPACE_OPERATOR = '::'
         ACCEPT_LEFT_PAREN =
           %w[break defined? next not rescue return super yield].freeze
         ACCEPT_LEFT_SQUARE_BRACKET =
           %w[super yield].freeze
+        ACCEPT_NAMESPACE_OP =
```
koic: I think `ACCEPT_NAMESPACE_OPERATOR` is clearer than `ACCEPT_NAMESPACE_OP `.

lib/rubocop/cop/layout/space_around_keyword.rb
```diff
@@ -30,10 +30,13 @@ class SpaceAroundKeyword < Cop
...
+          %w[super].freeze
```
koic: I think it's enough to use string object instead of array object.

lib/rubocop/cop/layout/space_around_keyword.rb
ozydingo: Changed to `ACCEPT_NAMESPACE_OPERATOR`.

lib/rubocop/cop/layout/space_around_keyword.rb
ozydingo: Change to string and matching using `==` instead of `include?`

koic:
Thanks!

PR #7522, opened 2019-11-24
Fix false positive in Style/EvenOdd

buehmann:
```ruby
n % 2 == 2
```
is auto-corrected to
```
n.?
```
by `Style/EvenOdd`, which is clearly a mistake.

```
tmp-numbers.rb:1:1: C: [Corrected] Style/EvenOdd: Replace with Integer#?.
n % 2 == 2
^^^^^^^^^^
tmp-numbers.rb:1:3: E: Lint/Syntax: unexpected token tEH
(Using Ruby 2.3 parser; configure using TargetRubyVersion parameter, under AllCops)
n.?
  ^
```

I am not sure this unusual edge case warrants an entry in the changelog. Just let me know if I should add one.


koic:
I think it would be good to add the test case and the changelog entry.

buehmann:
> I think it would be good to add the test case and the changelog entry.

Okay, done!

koic:
Thanks for the perfectly update!

PR #7528, opened 2019-11-27
Add new Lint/NonDeterministicRequireOrder cop

Mangara:
### What is this PR trying to do?

`Dir[...]` and `Dir.glob(...)` make [no guarantees](https://ruby-doc.org/core/Dir.html#method-c-glob) about the order files are returned in:

> Case sensitivity depends on your system [...], as does the order in which the results are returned.

This becomes a problem when they are used for applications in which the order can matter, such as requiring files. At worst, this can lead to bugs that only happen intermittently in production and can't be reproduced in development.

This cop suggests adding a `.sort`:

```ruby
# bad
Dir["./lib/middleware/*.rb"].each do |file|
  require file
end

# good
Dir["./lib/middleware/*.rb"].sort.each do |file|
  require file
end
```

### Questions for reviewers

- Should it test for more patterns?

The current implementation very specifically looks for these three patterns: `Dir[_].each do |_|`, `Dir.glob(_).each do |_|`, and `Dir.glob(_) do |_|`.  Are there other common patterns that have the same effect? Should we allow for intermediate operations between the `Dir[...]` and `.each`?

- Should it apply regardless of whether the file is `require`d?

There are other use cases where the order matters, although this cannot in general be detected statically, and you may want to avoid the performance penalty associated with sorting. We could add a configuration option: `Always`, `OnlyForRequires`?

- Is there a better way to match the pattern?

-----------------

* [X] Added tests.


lib/rubocop/cop/lint/non_deterministic_require_order.rb
```diff
@@ -0,0 +1,70 @@
...
+
+          return unless var_name && var_is_required?(node, var_name)
+
+          add_offense(block_start)
+        end
+
+        def var_is_required?(node, var_name)
+          matcher = NodePattern.new("(send nil? :require (lvar :#{var_name}))")
+          node.each_descendant.find { |n| matcher.match(n) }
+        end
```
buehmann: There is no need to compile a new node pattern for each invocation. You can write this method like this:
```ruby
        def_node_search :var_is_required?, <<~PATTERN
          (send nil? :require (lvar %1))
        PATTERN
```
Node patterns can use arguments (`%1`) and `def_node_search` does the search in all descendants.

Mangara: Fixed, thanks! I figured there would be an easier way.

lib/rubocop/cop/lint/non_deterministic_require_order.rb
```diff
@@ -0,0 +1,69 @@
...
+        PATTERN
+
+        def on_block(node)
+          return unless node.children && node.children.length > 2
+
+          block_start = node.children[0]
+
+          return unless unsorted_dir_loop?(block_start)
+
+          var_name = loop_variable(node.children[1])
```
buehmann: Block nodes have some methods to access their parts that could make this more readable:
```ruby
return unless node.body

block_start = node.send_node
...
var_name = loop_variable(node.arguments)
```
See https://github.com/rubocop-hq/rubocop/blob/bcd21f9fe309310423cfb9cee39c979bab56768c/lib/rubocop/ast/node/block_node.rb#L16-L35

Mangara: Much nicer! It also makes me more confident that the array will look like what we expect here.

Mangara:
@buehmann What needs to happen before we can merge this?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* Add new `Lint/NonDeterministicRequireOrder` cop. ([@mangara][])
```
buehmann: Could you add a link to this PR? (See other changelog entries for examples.) Each changelog entry usually links to an issue or to a PR if there is no related issue.

lib/rubocop/cop/lint/non_deterministic_require_order.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Dir[...] and Dir.glob(...) do not make any guarantees about
```
buehmann: Please put the `Dir...` fragments in single backticks in order to format them as code.

lib/rubocop/cop/lint/non_deterministic_require_order.rb
```diff
@@ -0,0 +1,69 @@
...
+          (send (send (const nil? :Dir) {:[] :glob} ...) :each)
+        PATTERN
+
+        def_node_matcher :loop_variable, <<~PATTERN
+          (args (arg $_))
+        PATTERN
+
+        def_node_search :var_is_required?, <<~PATTERN
+          (send nil? :require (lvar %1))
+        PATTERN
```
buehmann: Could you make those methods `private`?

spec/rubocop/cop/lint/non_deterministic_require_order_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+            end
+          end
+        RUBY
+      end
+    end
+
+    context 'with unsorted glob' do
+      it 'registers an offsense' do
+        expect_offense(<<~RUBY)
+          Dir.glob(Rails.root.join(__dir__, 'test', '*.rb'), File::FNM_DOTMATCH).each do |file|
```
buehmann: Regarding your question
> Should it test for more patterns?

I think it might make sense to support this case as well from the beginning:
```ruby
Dir.glob(Rails.root.join(__dir__, 'test', '*.rb'), File::FNM_DOTMATCH) do |file|
  require file
end
```
(that is `glob` taking a block without `each`)

What do you think? (Adding this should be rather easy with a `{(send …) (send …)}` alternative in `unsorted_dir_loop?`.)

config/default.yml
```diff
@@ -1471,6 +1471,11 @@ Lint/NextWithoutAccumulator:
   Enabled: true
   VersionAdded: '0.36'
 
+Lint/NonDeterministicRequireOrder:
+  Description: 'Always sort arrays returned by Dir.glob when requiring files.'
+  Enabled: true
+  VersionAdded: '0.78'
```
koic: I think this cop should be marked unsafe because it is not compatible behaviour.

```yaml
 VersionAdded: '0.78'
+Safe: false
```

spec/rubocop/cop/lint/non_deterministic_require_order_spec.rb
Mangara: Yes! Good catch. We should definitely support that.

Mangara: Is this the best way to write the new pattern?
```
{ (send (const nil? :Dir) :glob ...) (send (send (const nil? :Dir) {:[] :glob} ...) :each) }
```

Mangara: (Only the `.glob` variant supports the direct block syntax)

Mangara: I ended up splitting it into two patterns to help with autocorrection.

lib/rubocop/cop/lint/non_deterministic_require_order.rb
```diff
@@ -0,0 +1,89 @@
...
+
+        def on_block(node)
+          return unless node.body
+          return unless unsorted_dir_loop?(node.send_node)
+
+          var_name = loop_variable(node.arguments)
+
+          return unless var_name && var_is_required?(node.body, var_name)
+
+          add_offense(node.send_node)
```
Drenmi: The generated methods also take a block, which is yielded to in the case of a match. 🙂 

```ruby
loop_variable(node.arguments) do |var_name|
  return unless var_is_required?(node.body, var_name)

  add_offense(node.send_node)
end
```

Looks like you could also potentially combine the patterns into a single matcher.

lib/rubocop/cop/lint/non_deterministic_require_order.rb
```diff
@@ -0,0 +1,89 @@
...
+        end
+
+        def autocorrect(node)
+          if unsorted_dir_block?(node)
+            lambda do |corrector|
+              corrector.replace(node.loc.expression, "#{node.source}.sort.each")
+            end
+          else
+            lambda do |corrector|
+              source = node.children[0].source
```
Drenmi: We try not to use unnamed children, but use node specific methods instead.

Since we know this is a `send` node:

```ruby
source = node.receiver.source
```

lib/rubocop/cop/lint/non_deterministic_require_order.rb
Mangara: Ah, nice! I like the block syntax.

As for the single matcher, I like that all of the building blocks are fairly simple patterns. Merging them into one would make it harder to understand, I think.

bbatsov:
Can you rebase this on top of `master`?

Mangara:
@bbatsov @koic Is this ready to merge?

koic:
The following commit is due to the impact of this cop addition.
https://github.com/rubocop-hq/rubocop/pull/7528/commits/161781d68aa47d20a8cb4e97a1961730c9b0201d

Can you squash your commits into one?

koic:
And I think @bbatsov will decide whether the next release will be new feature release or bug fix release. Please wait until the next release is a release that includes new features.

PR #3765, opened 2016-12-03
Add a validation for supported styles other than EnforcedStyle

pocke:
I added a validation for supported styles other than EnforcedStyle.

Problem
======

Currently, EnforcedStyle value is validated by SupportedStyles.

e.g.

```sh
$ cat .rubocop.yml
Style/AlignParameters:
  EnforcedStyle: aaaaaaaaaaaaaaaaaaaaaa
$ rubocop
Error: invalid EnforcedStyle 'aaaaaaaaaaaaaaaaaaaaaa' for Style/AlignParameters found in /tmp/tmp.SyHaGtdT0U/.rubocop.yml
Valid choices are: with_first_parameter, with_fixed_indentation
```

However, other styles aren't validated.

e.g.

```sh
$ cat .rubocop.yml
Style/SpaceAroundBlockParameters:
  EnforcedStyleInsidePipes: aaaaaaaaaaaaaaaaaaaaaa
$ rubocop
Inspecting 0 files

0 files inspected, no offenses detected
```

Goal
======

Validate the invalid value.

e.g.

```sh
$ cat .rubocop.yml
Style/SpaceAroundBlockParameters:
  EnforcedStyleInsidePipes: aaaaaaaaaaaaaaaaaaaaaa
$ rubocop
Error: invalid EnforcedStyleInsidePipes 'aaaaaaaaaaaaaaaaaaaaaa' for Style/SpaceAroundBlockParameters found in /tmp/tmp.SyHaGtdT0U/.rubocop.yml
Valid choices are: space, no_space
```

Changes
=====

- Add validation for styles other than `EnforcedStyle`
  - And rename some `SupportedStyles` names to match EnforcedStyle name
    - e.g. `SupportedStyles` to `SupportedStylesInsidePipes` (for `EnforcedStyleInsidePipes`)
- Add `SupportedStyles` for missing styles in config/default.yml
- Add a spec to check existence of `SupportedStyles` into `project_spec.rb`
- Add test cases to check validation

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Used the same coding conventions as the rest of the project.
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
[...]

PR #7531, opened 2019-11-27
Add support for variable alignment to Layout/RescueEnsureAlignment

### LINKED ISSUE #6918: Layout/RescueEnsureAlignment incorrect behaviour when used as assignment
`Layout/RescueEnsureAlignment` does not support assignment

--------

## Expected behavior

Should consider indentation difference from first non-whitespace character in line containing `begin`.

```rb
values = begin
  get_values_array
rescue
  []
end
```

## Actual behavior

(for actual use case, example is simplified)

```
app/models/concerns/demographics.rb:21:5: C: Layout/RescueEnsureAlignment: rescue at 21, 4 is not aligned with begin at 19, 13.
    rescue ActiveRecord::StatementInvalid => error
    ^^^^^^
```

Follow example.

## RuboCop version

```
$ bundle exec rubocop -V
0.64.0 (using Parser 2.6.0.0, running on ruby 2.4.5 x86_64-darwin18)
```

dylanahsmith:
Fixes #6918

The Layout/RescueEnsureAlignment cop was always enforcing alignment of the `rescue` keyword with the `begin` keyword, even when the `begin` block was being assigned to a variable.  For example, it would expect code like

```ruby
result = begin
           something
         rescue
           puts 'error'
         end
```

but reject code aligning `rescue` with a variable like

```ruby
result = begin
  something
rescue
  puts 'error'
end
```

even though a variable style is supported for similar cops, such as Layout/EndAlignment.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1059,6 +1059,10 @@ Layout/ParameterAlignment:
 
 Layout/RescueEnsureAlignment:
   Description: 'Align rescues and ensures correctly.'
+  EnforcedStyleAlignWith: keyword
+  SupportedStylesAlignWith:
+    - keyword
+    - variable
   Enabled: true
   VersionAdded: '0.49'
 
--- a/lib/rubocop/cop/layout/rescue_ensure_alignment.rb
+++ b/lib/rubocop/cop/layout/rescue_ensure_alignment.rb
@@ -6,22 +6,48 @@ module RuboCop
       # This cop checks whether the rescue and ensure keywords are aligned
       # properly.
       #
-      # @example
+      # Two modes are supported through the `EnforcedStyleAlignWith`
+      # configuration parameter:
+      #
+      # If it's set to `keyword` (which is the default), the `rescue`
+      # shall be aligned with the start of the begin keyword.
+      #
+      # If it's set to `variable` the `rescue` shall be aligned with the
+      # left-hand-side of the variable assignment, if there is one.
+      #
+      # @example EnforcedStyleAlignWith: keyword (default)
       #
       #   # bad
-      #   begin
+      #   result = begin
+      #              something
+      #              rescue
+      #              puts 'error'
+      #            end
+      #
+      #   # good
+      #   result = begin
+      #              something
+      #            rescue
+      #              puts 'error'
+      #            end
+      #
+      # @example EnforcedStyleAlignWith: variable
+      #
+      #   # bad
+      #   result = begin
       #     something
       #     rescue
       #     puts 'error'
       #   end
       #
       #   # good
-      #   begin
+      #   result = begin
       #     something
       #   rescue
       #     puts 'error'
       #   end
       class RescueEnsureAlignment < Cop
+        include ConfigurableEnforcedStyle
         include RangeHelp
 
         MSG = '`%<kw_loc>s` at %<kw_loc_line>d, %<kw_loc_column>d is not ' \
@@ -39,6 +65,10 @@ module RuboCop
 
         def on_ensure(node)
           check(node)
+        end
+
+        def style_parameter_name
+          'EnforcedStyleAlignWith'
         end
 
         def autocorrect(node)
@@ -122,8 +152,8 @@ module RuboCop
         def alignment_node(node)
           ancestor_node = ancestor_node(node)
 
-          return ancestor_node if ancestor_node.nil? ||
-                                  ancestor_node.kwbegin_type?
+          return ancestor_node if ancestor_node.nil?
+          return ancestor_node if style == :keyword && ancestor_node.kwbegin_type?
 
           assignment_node = assignment_node(ancestor_node)
           return assignment_node if same_line?(ancestor_node, assignment_node)
--- a/spec/rubocop/cop/layout/rescue_ensure_alignment_spec.rb
+++ b/spec/rubocop/cop/layout/rescue_ensure_alignment_spec.rb
@@ -1,62 +1,106 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::Layout::RescueEnsureAlignment, :config do
+  let(:cop_config) do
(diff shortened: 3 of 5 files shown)


spec/rubocop/cop/layout/rescue_ensure_alignment_spec.rb
```diff
@@ -3,62 +3,106 @@
...
-      end
+      RUBY
+    end
+
+    it 'registers offense for keyword alignment' do
+      expect_offense(<<~RUBY)
+        x ||= begin
+                1
+              rescue
+              ^^^^^^ `rescue` at 3, 6 is not aligned with `x` at 1, 0.
```
pariser: ❤️

lib/rubocop/cop/layout/rescue_ensure_alignment.rb
```diff
@@ -6,22 +6,48 @@ module Layout
       # This cop checks whether the rescue and ensure keywords are aligned
       # properly.
       #
-      # @example
+      # Three modes are supported through the `EnforcedStyleAlignWith`
```
pariser: Nice to make this configurable

Mangara:
```suggestion
      # Two modes are supported through the `EnforcedStyleAlignWith`
```
?

config/default.yml
```diff
@@ -995,6 +995,10 @@ Layout/ParameterAlignment:
 
 Layout/RescueEnsureAlignment:
   Description: 'Align rescues and ensures correctly.'
+  EnforcedStyleAlignWith: keyword
```
bbatsov: Why not simply `EnforcedStyle`?

bbatsov:
Apart from my small remark the suggested changes seem reasonable to me.

config/default.yml
dylanahsmith: It looked like the alignment cops were using `EnforcedStyleAlignWith`, so I was just trying to be consistent with them.  Why not simply `EnforcedStyle` for `Layout/BlockAlignment`, `Layout/DefEndAlignment`, `Layout/EndAlignment` and `Layout/RescueEnsureAlignment`?  I'm fine either way.  Did you want me to change it to `EnforcedStyle`?

ianfixes:
I came here from #6918 ... Can I request a test case be added?

The code in question is:

```ruby
  private memoize def do_something
    "Something"
  rescue NameError
    nil
  end
```

Rubocop `0.77.0` wants this:

```ruby
  private memoize def do_something
    "Something"
                  rescue NameError
                    nil
  end
```

If I remove `memoize`, Rubocop acts sensibly and allows this:

```ruby
  private def do_something
    "Something"
  rescue NameError
    nil
  end
```

Does this belong here as I came here from #6918 ... I'm looking for guidance on whether to propose a test case for this or open a separate feature request entirely.

The code in question is:

```ruby
  private memoize def do_something
    "Something"
  rescue NameError
    nil
  end
```

Rubocop `0.77.0` wants this:

```ruby
  private memoize def do_something
    "Something"
                  rescue NameError
                    nil
  end
```

If I remove `memoize`, Rubocop acts sensibly and allows this:

```ruby
  private def do_something
    "Something"
  rescue NameError
    nil
  end
```

Should I open a separate issue for supporting `EnforcedStyleAlignWith: start_of_line` configuration option for this cop, or is that a trivial addition to this PR?

config/default.yml
ianfixes: > It looked like the alignment cops were using `EnforcedStyleAlignWith`

That is my experience/expectation as well, FWIW

dylanahsmith:
It would make sense to add support for `start_of_line` as well, but perhaps that should be done in another PR to keep this one focused.

Richard-Degenne:
Any input on the status of this PR?

tdeo:
I'd love to see that merged, @bbatsov any final decision on `EnforcedStyleAlignWith` (used by `Layout/BlockAlignment`, `Layout/DefEndAlignment`, `Layout/EndAlignment`) vs `EnforcedStyle` ?

bbatsov:

ianfixes:
I was under the impression that this PR was waiting for a response by @bbatsov to the following questions:

> Did you want me to change `EnforcedStyleAlignWith` to EnforcedStyle? (https://github.com/rubocop-hq/rubocop/pull/7531#discussion_r364253406)

> @bbatsov any final decision on EnforcedStyleAlignWith (used by Layout/BlockAlignment, Layout/DefEndAlignment, Layout/EndAlignment) vs EnforcedStyle ? (https://github.com/rubocop-hq/rubocop/pull/7531#issuecomment-677566538)

config/default.yml
bbatsov: I no longer remember why we did it like this in the past, but I definitely don't like it today. Probably it was to avoid styles named `align_with_*`, but I think that's fine.

bbatsov:
Well, if that's the case - I prefer the use of `EnforcedStyle` `align_with_x`. @jonas054 If you recall why we didn't go with `EnforcedStyle` there, please do remind me.

dylanahsmith:
As mentioned in the referenced issue (https://github.com/rubocop-hq/rubocop/issues/6918#issuecomment-699469222), I think this problem is addressed by the Layout/BeginEndAlignment cop with the `EnforcedStyleAlignWith: start_of_line` configuration to achieved the desired style mentioned in the PR description.

bbatsov:
A problem that solves itself - that's my favourite kind. 😄 Thanks!

jonas054:
> Well, if that's the case - I prefer the use of EnforcedStyle align_with_x. @jonas054 If you recall why we didn't go with EnforcedStyle there, please do remind me.

No, I don't remember. I've found that we changed the name from `AlignWith` to `EnforcedStyleAlignWith` in #3765, but why we didn't go all the way to `EnforcedStyle`, I cannot say.

PR #7542, opened 2019-12-01
Move `LineLength` cop from `Metrics` to `Layout` department

koic:
### Summary

The `Layout` department was created after `Metrics` was created.

- [Create namespace Metrics and move some cops there](https://github.com/rubocop-hq/rubocop/commit/e2153e8cdaea8a79ab51977b0071c7b15cf95078)
- [Layout department](https://github.com/rubocop-hq/rubocop/pull/4278)

As far as I use `LineLength` cop, I think this cop belongs to the layout, not the logic metrics. Rather than changing the logic when `LineLength` cop becomes a offense, a solution is often a line wrap.

The horizontal length targeted by `LineLength` cop is resolved by often a line wrapping, while the vertical length targeted by `ClassLength` cop, `MethodLength` cop, and `BlockLength` cop is
resolved by a logic change.
So `ClassLength`, `MethodLength`, and `BlockLength` cop remain in
the `Metrics` department.

### Other Information

I'm wondering if the default maximum length of 80 is reasonable.
This is because there are cases where clear variable name (and etc) have to be omitted due to restrictions on `LineLength`.
However, this PR does not handle it because first of all this PR aims to move department.

-----------------

* [ ] Added tests.


bbatsov:
I agree with your reasoning, but this might be confusing to the end users. @rubocop-hq/rubocop-core what do you think about the proposed change?

> I'm wondering if the default maximum length of 80 is reasonable.

I guess we can revisit this before 1.0 when we discuss which defaults should be updated.

Drenmi:
I tend to agree with the move, actually. 👍 

---

As for the default length, I think 80 is the only length that has an actual justification. The problem is that justification is veeery outdated, dating back to IBM punch cards.

There does exist some research for optimal line length, but that is for natural language texts, and the results are inconsistent. I'm not aware of any such studies for code.

Bottom line is whatever is chosen is going to be arbitrary. 🙂

PR #7552, opened 2019-12-05
In default.yml, Exclude tmp

jaredbeck:
In certain rails projects, `tmp` can contain hundreds of thousands of files.

Excluding `tmp` in my project seems to decrease `time bin/rubocop` from 90s to 20s, though that's just an anecdote and not a scientific benchmark.

Does anyone put code in `tmp` that they actually want to lint? That'd be rare, right?

-----------------

* [N/A] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [N/A] Added tests.
* [N/A] Added an entry to the [Changelog](https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format).
* [N/A?] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

-----

Do I need to run `rake` after editing `default.yml`?

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -59,6 +59,7 @@ AllCops:
     - '**/Vagrantfile'
   Exclude:
     - 'node_modules/**/*'
+    - 'tmp/**/*'
     - 'vendor/**/*'
     - '.git/**/*'
   # Default formatter will be used if no `-f/--format` option is given.


bbatsov:
Nice catch!

PR #7562, opened 2019-12-11
Adds new cop Lint/ApprovedCallSite

wallace:
Addresses https://github.com/rubocop-hq/rubocop/issues/7561

Adds a new cop to lint call sites for a class or module. From https://github.com/rubocop-hq/rubocop/issues/7561,

> I'd like a cop that highlights when a new call is made to a constant or module.
> 
> E.g. you have `ThirdPartyAPILibrary` that you would like to ensure is only called in the appropriate context i.e., from within a wrapper class, `WrapThirdParty`. 
> 
> With a cop that will flag any occasion where `ThirdPartyAPILibrary` is the receiver a message, one can easily audit all those occasions. One could then annotate those calls made within WrapThirdParty with a local disable to show they are approved. 
> 
> If a future developer attempted to make a call to `ThirdPartyAPILibrary`, it is flagged as an offense.

-----------------

* [x] Added tests.


pocke:
I'm concerning this cop works only for a few teams, and is too abstract.

Personally I think RuboCop core only has rules that work for most people. If a cop depends on a library or project, it should be placed in a rubocop plugin.
But the cop only works teams that have the rule, which is wrapping `ThirdPartyAPILibrary` and so on.


So I recommend other ways to detect the problem.
You can create a RuboCop plugin with https://github.com/rubocop-hq/rubocop-extension-generator.
Or you can just commit `lib/rubocop/cop/lint/approved_call_site.rb` to your project directly, and require it from `.rubocop.yml`.


You can use querly otherwise.
https://github.com/soutaro/querly
It is a static analysis tool, but it's for project-specific rules.
I guess the following querly.yaml works for you.


```yaml
rules:
  - id: approved_call_site
    pattern:
      - ThirdPartyAPILibrary._
    message: Do not use ThirdPartyAPILibrary directly. Use WrapThirdParty instead.
    before:
      - 'ThirdPartyAPILibrary.some_method'
    after:
      - 'WrapThirdParty.some_method'
```



But I'd like to hear other core team members' opinions. Any thought? @rubocop-hq/rubocop-core 



Thanks for your contribution.

koic:
> I think RuboCop core only has rules that work for most people. If a cop depends on a library or project, it should be placed in a rubocop plugin.

I agree with @pocke. The following is an example of a RuboCop plug-in gem created.
https://flexport.engineering/isolating-rails-engines-with-rubocop-210feaba3164

wallace:
Thank you for taking the time to provide prompt feedback, @pocke and @koic. I greatly appreciate it.

I'll address your concerns individually to provide further context as to why I made this proposal and then conclude.

1. **Including the cop directly in the project**. One of the challenges of including the cop directly in the project is a differing testing convention. My project uses minitest. Rubocop uses RSpec. I do not want to include an untested cop in our code. I'm ignorant of the effort involved in building minitest support for testing a custom cop.
1. **Building an extension**. I did initially [build this as an extension](https://github.com/github/rubocop-github/pull/52) but my approach was too specific to the `ThirdPartyAPILibrary`. I ultimately determined that the precedent set of including such a specific information could lead to sensitive or confidential information leaking.
1. **Using querly**. Querly does not support [disabling false positives](https://github.com/soutaro/querly#too-many-false-positives).

I settled on the approach in this PR by making the approach generic enough to be useful to anyone who wanted a deeper review for usage of `ThirdPartyAPILirbary`.

Now that the approach is generic, and you do not see the broad appeal for this cop in rubocop, I will revisit [building this as an extension](https://github.com/github/rubocop-github/pull/52).

Thank you again for all your work on this excellent project.

bbatsov:
> Including the cop directly in the project. One of the challenges of including the cop directly in the project is a differing testing convention. My project uses minitest. Rubocop uses RSpec. I do not want to include an untested cop in our code. I'm ignorant of the effort involved in building minitest support for testing a custom cop.

Just take a look at `rubocop-minitest` - its test suite uses minitest. 

> I settled on the approach in this PR by making the approach generic enough to be useful to anyone who wanted a deeper review for usage of ThirdPartyAPILirbary.

In general we avoid adding anything in RuboCop's core that doesn't relate directly to Ruby's core. Thanks for the PR, but I agree with the others that this cop doesn't belong in the main RuboCop library.

PR #7565, opened 2019-12-12
EndAlignment: add support for kwbegin

jaredbeck:
`EndAlignment` currently supports `if`, `class`, `module`, `while`, etc. This PR attempts to add support for `begin`. For example, the following misalignment is generally considered bad.

```
begin
    # foo
  end
```

-----------------

* [N/A] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Added tests.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -10,6 +10,41 @@
 InternalAffairs/NodeDestructuring:
   Enabled: false
 
+# It is not intended that the following exclusions be merged. These exclusions
+# are temporary pending a maintainer decision re: the current discrepancy
+# between eg:
+#
+#     x = begin
+#       foo
+#     end
+#
+# and
+#
+#     x = if y
+#           foo
+#         end
+#
+# EndAlignment recently began to care about the former (`begin`) thus making
+# these temporary exclusions necessary.
+Layout/EndAlignment:
+  Exclude:
+    - lib/rubocop/cli.rb
+    - lib/rubocop/comment_config.rb
+    - lib/rubocop/config_validator.rb
+    - lib/rubocop/cop/generator/require_file_injector.rb
+    - lib/rubocop/cop/message_annotator.rb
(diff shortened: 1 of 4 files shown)


bbatsov:
See the related conversation here https://github.com/rubocop-hq/rubocop/pull/7286

lib/rubocop/cop/layout/end_alignment.rb
```diff
@@ -72,6 +72,15 @@ class EndAlignment < Cop
         include EndKeywordAlignment
         include RangeHelp
 
+        def on_kwbegin(node)
+          align_with = {
+            keyword: node.loc.begin,
```
jaredbeck: Is this right? The other callbacks, like `on_class` use `node.loc.keyword` but `Parser::Source::Map::Collection` doesn't respond to `keyword`. I'd really appreciate feedback here, I'm not at all confident this is correct. Thanks.

.rubocop_todo.yml
```diff
@@ -10,6 +10,41 @@
...
+#     end
+#
+# and
+#
+#     x = if y
+#           foo
+#         end
+#
+# EndAlignment recently began to care about the former (`begin`) thus making
+# these temporary exclusions necessary.
```
jaredbeck: What I'm trying to say here is that rubocop itself uses the `keyword` style for eg. `if`, but uses the `start_of_line` style for `begin`.

lib/rubocop/cop/layout/end_alignment.rb
Drenmi: @jaredbeck I've been [asking](https://github.com/whitequark/parser/issues/357) the maintainers of `parser` if the source map can be changed, but they are not interested. They suggested we patch it in RuboCop instead.

I think patching the source map is probably better, since there are other cops that are affected by this as well.

jaredbeck:
Closing in favor of https://github.com/rubocop-hq/rubocop/pull/7337

PR #7567, opened 2019-12-12
[Fix #5979] Introduce cops with special status

### LINKED ISSUE #5979: Introduce new cops with a special status (instead of enabled/disabled)

pirj:
New cops, introduced or significantly changed between major versions, are introduced with `pending` status. This is done to avoid frustration over numerous offences during updates between minor versions. RuboCop will emit a warning that new cops were introduced, and suggest to explicitly enable or disable them in the user configuration file.

Based on https://github.com/rubocop-hq/rubocop/pull/7184

Fixes #5979

-----------------

* [x] Added tests.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -10,10 +10,10 @@
 InternalAffairs/NodeDestructuring:
   Enabled: false
 
-# Offense count: 49
+# Offense count: 50
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 183
+  Max: 197
 
 # Offense count: 209
 # Configuration parameters: CountComments, ExcludedMethods.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 * [#7528](https://github.com/rubocop-hq/rubocop/pull/7528): Add new `Lint/NonDeterministicRequireOrder` cop. ([@mangara][])
 * [#7559](https://github.com/rubocop-hq/rubocop/pull/7559): Add `EnforcedStyleForExponentOperator` parameter to `Layout/SpaceAroundOperators` cop. ([@khiav223577][])
-* [#7184](https://github.com/rubocop-hq/rubocop/pull/7184): Introduce new `pending` status for new cops. ([@Darhazer][])
+* [#7567](https://github.com/rubocop-hq/rubocop/pull/7567): Introduce new `pending` status for new cops. ([@Darhazer][], [@pirj][])
 
 ### Bug fixes
 
@@ -4294,3 +4294,4 @@
 [@movermeyer]: https://github.com/movermeyer
 [@jethroo]: https://github.com/jethroo
 [@mangara]: https://github.com/mangara
+[@pirj]: https://github.com/pirj
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 * [#7528](https://github.com/rubocop-hq/rubocop/pull/7528): Add new `Lint/NonDeterministicRequireOrder` cop. ([@mangara][])
 * [#7559](https://github.com/rubocop-hq/rubocop/pull/7559): Add `EnforcedStyleForExponentOperator` parameter to `Layout/SpaceAroundOperators` cop. ([@khiav223577][])
+* [#7184](https://github.com/rubocop-hq/rubocop/pull/7184): Introduce new `pending` status for new cops. ([@Darhazer][])
 
 ### Bug fixes
 
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -91,7 +91,9 @@ module RuboCop
         else
           add_excludes_from_files(config, config_file)
         end
-        merge_with_default(config, config_file)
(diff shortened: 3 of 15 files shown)


.rubocop_todo.yml
```diff
@@ -10,10 +10,10 @@
 InternalAffairs/NodeDestructuring:
   Enabled: false
 
-# Offense count: 49
+# Offense count: 50
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 183
+  Max: 197
```
pirj: Sorry about that, but it feels that I might get into a refactoring spiral if I touch ConfigLoader.

lib/rubocop/config_loader.rb
```diff
@@ -91,7 +91,9 @@ def configuration_from_file(config_file)
         else
           add_excludes_from_files(config, config_file)
         end
-        merge_with_default(config, config_file)
+        merge_with_default(config, config_file).tap do |merged_config|
```
pirj: `merge_with_default` does not mutate first argument, but rather returns merged config, so I was between `tap` and using a temporary local variable.

bbatsov:
> What I was not 100% certain of the desired effect of EnabledByDefault/DisabledByDefault on new pending status, so I kept it aside. I've added pending spec examples for those cases though.

I guess they should override the new `pending` status, as that'd be consistent with the previous behaviour.

bbatsov:
A couple of more things:

* We should mention the new status in the manual (and the policy that new cops will be using it, until a major release when they'll be enabled/disabled by default)
* We should update the template for new cops to use the pending status

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * [#7528](https://github.com/rubocop-hq/rubocop/pull/7528): Add new `Lint/NonDeterministicRequireOrder` cop. ([@mangara][])
 * [#7559](https://github.com/rubocop-hq/rubocop/pull/7559): Add `EnforcedStyleForExponentOperator` parameter to `Layout/SpaceAroundOperators` cop. ([@khiav223577][])
+* [#7184](https://github.com/rubocop-hq/rubocop/pull/7184): Introduce new `pending` status for new cops. ([@Darhazer][])
```
Darhazer: You need to update this line (both the PR number and the author ;))

bbatsov:
@pirj Any chance you'll be wrapping this one soon? I'm wondering whether to cut a new release now or wait for this PR.

pirj:
@bbatsov Current status and a couple of questions:
1. I'm a bit confused with the default behaviour of `Enabled`/`DisabledByDefault`, since it seems to only affect RuboCop's own default config, and ignores extensions' configs. So e.g. if `rubocop-rspec`'s config has a pending cop (or, enabled, or, disabled, doesn't matter), those settings don't seem to override extension's cop enabled status.

2. docs - Intend to extend `manual/configuration.md` mentioning default `pending` status.
Not sure, if a dedicated section mentioning how we are going to introduce cops, and how users are supposed to update RuboCop should be added?

I have some time this week to address this, but if something else pops up, this might take longer, so it's better to postpone to the next release.

bbatsov:
> I'm a bit confused with the default behaviour of Enabled/DisabledByDefault, since it seems to only affect RuboCop's own default config, and ignores extensions' configs. So e.g. if rubocop-rspec's config has a pending cop (or, enabled, or, disabled, doesn't matter), those settings don't seem to override extension's cop enabled status.

Well, I think that's an orthogonal problem that has to be fixed separately. This setting dates back from a time where there were almost no extensions.

> Is there a distinction between gem's default configs, and user's configs, e.g. if we require .rubocop_base.yml that enables e.g. Lint/A, and also rubocop-rspec/config/default.yml that enables RSpec/B, and the main config has DisabledByDefault: true, shouldn't we disable RSpec/B but keep Lint/A enabled?

Yeah, I agree. 

> docs - Intend to extend manual/configuration.md mentioning default pending status.

👍 

> Not sure, if a dedicated section mentioning how we are going to introduce cops, and how users are supposed to update RuboCop should be added?

Some small section on the subject would be useful indeed.

pirj:
@bbatsov All green. Please take a look if something needs to be added.

pirj:
There's one very nasty thing that I've noticed about `:isolated_environment`. In cases there's something horribly wrong with the example, it interrupts normal RSpec cycle and reports no error. All specs appear to be green, and the only way to figure out that something is wrong is to make sure the number of examples in a spec matches those reported by RSpec.

bbatsov:
Looks good to me. Thanks for brining this across the finish line!

pirj:
Happy to move his forward 👍

PR #7582, opened 2019-12-21
Remove unnecessary use of regexp

buehmann:
This is just a line I stumbled across. There is no need to use the
regexp machinery just to match a fixed substring.


bbatsov:
Nice catch!

PR #7588, opened 2019-12-23
[Fix #7587] Add `Style/RedundantReturn`'s `AllowedMethods` option.

### LINKED ISSUE #7587: Style/RedundantReturn need white list
This issue is updated version from #7539 (with a little correction around my misunderstandings) 

With Rails controller, we sometimes see codes as below:

```ruby
  def show
    if cond?
      redirect_to posts_url
      return
    end
  end
```
Here, this `return` is not needed generally speaking, but some developers would prefer  
 to put it as it can avoid potential DoubleRenderErrors 
For example, another developer might modify as below.

```ruby
def show
  if cond?
    redirect_to posts_url
    return
  end

  # ========== snip =========

  render :some_view
end
```
Good point about this code is that we can make sure to avoid feature DoubleRenderError.
Now, Rubocop this `return` violate by Style/RedundantReturn.

I'd suggest a solution
add AllowedMethods option
```
  AllowedMethods:
    - redirect_to
```

In most cases, redirect_to and render need this option

QWYNG:
fix #7587 
-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_return.rb
+++ b/lib/rubocop/cop/style/redundant_return.rb
@@ -123,6 +123,10 @@ module RuboCop
           return if cop_config['AllowMultipleReturnValues'] &&
                     node.children.size > 1
 
+          around_methods = [method_name_before_of(node),
+                            method_name_first_child_of(node)]
+          return if (allowed_methods & around_methods).any?
+
           add_offense(node, location: :keyword)
         end
 
@@ -164,6 +168,29 @@ module RuboCop
           else
             MSG
           end
+        end
+
+        def allowed_methods
+          cop_config['AllowedMethods']&.map(&:to_sym) || []
+        end
+
+        def method_name_before_of(node)
+          parent = node.parent
+          return unless %i[begin kwbegin].include?(parent.type)
+
+          node_place_index = parent.children.find_index(node)
+          return if node_place_index.zero?
+
+          before_node = parent.children[node_place_index - 1]
+          return unless before_node.send_type?
+
+          before_node.method_name
+        end
+
+        def method_name_first_child_of(node)
+          return unless node.children.first&.send_type?
+
+          node.children.first.method_name
         end
       end
     end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/redundant_return.rb
```diff
@@ -165,6 +167,18 @@ def message(node)
...
           end
         end
+
+        def allowed_methods
+          cop_config['AllowedMethods']&.map(&:to_sym) || []
+        end
+
+        def method_name_before_of(node)
+          node_place_index = node.parent.children.find_index(node)
+          before_node = node.parent.children[node_place_index - 1]
```
buehmann: This will lead to all kinds of problems because you cannot make assumptions about what the parent of the return node will be.

For example you would allow this `return`:
```ruby
def foo
  if bar
    allowed_method
  else
    return
  end
end
```

QWYNG: Thank you.
I added `node.parent.type` check

lib/rubocop/cop/style/redundant_return.rb
```diff
@@ -165,6 +167,21 @@ def message(node)
...
+
+        def allowed_methods
+          cop_config['AllowedMethods']&.map(&:to_sym) || []
+        end
+
+        def method_name_before_of(node)
+          parent = node.parent
+          return unless %i[begin kwbegin].include?(parent.type)
+
+          node_place_index = parent.children.find_index(node)
```
buehmann: I think you need a `return unless node_place_index >= 1` or similar here. The following code won't do the right thing if the index is 0.

QWYNG: Thank you!
I added `node_place_index.zero?` gurad

QWYNG:
@buehmann 
thank you for review
I made another change that was not reviewed
I added check whitelist method name after `return`  for some case as below

```
def func
  if cond?
    return redirect_to some_path
  end
end
```

bbatsov:
As mentioned in the ticket - I'm not sure this solution is a good one, because it's too specific to that particular problem that was reported. If we want to solve one particular Rails-related problem it should be tackled in RuboCop Rails.

PR #7596, opened 2019-12-26
[Fix #7595] Make `Style/NumericPredicate` aware of ignored methods

### LINKED ISSUE #7595: Style/NumericPredicate does not honor IgnoredMethods.
`Style/NumericPredicate` does not appear to be honoring a list of `IgnoredMethods`.

```
# .rubocop.yml
Style/NumericPredicate:
  IgnoredMethods: ['zero?']

# test.rb
if foo == 0
  puts "bar"
end
```

```
$ rubocop --only Style/NumericPredicate ~/tmp/test.rb
Inspecting 1 file
C
Offenses:
test.rb:1:4: C: Style/NumericPredicate: Use foo.zero? instead of foo == 0.
if foo == 0
   ^^^^^^^^
```

Nor does it honor `IgnoredMethods` with `EnforcedStyle: comparison`.

```
# .rubocop.yml
Style/NumericPredicate:
  EnforcedStyle: comparison
  IgnoredMethods: ['zero?']

# test.rb
if foo.zero?
  puts 'bar'
end
```

```
$ rubocop --only Style/NumericPredicate ~/tmp/test.rb
Inspecting 1 file
C
Offenses:
test.rb:1:4: C: Style/NumericPredicate: Use foo == 0 instead of foo.zero?.
if foo.zero?
   ^^^^^^^^^
```

## RuboCop version

```
0.78.0 (using Parser 2.6.5.0, running on ruby 2.6.4 x86_64-darwin18)
```

koic:
Fixes #7595.

This PR makes `Style/NumericPredicate` aware of ignored methods when specifying ignored methods.

```console
% cat .rubocop.yml
# .rubocop.yml
Style/NumericPredicate:
  EnforcedStyle: comparison
  IgnoredMethods: 'zero?'

% cat example.rb
if foo.zero?
  puts 'hello'
end

% bundle exec rubocop --only Style/NumericPredicate --cache false example.rb
Inspecting 1 file
C

Offenses:

example.rb:1:4: C: Style/NumericPredicate: Use foo == 0 instead of
foo.zero?.
if foo.zero?
   ^^^^^^^^^

```

This PR resolves the issue caused by the callback node itself not being checked.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/numeric_predicate_spec.rb
+++ b/spec/rubocop/cop/style/numeric_predicate_spec.rb
@@ -192,6 +192,42 @@ RSpec.describe RuboCop::Cop::Style::NumericPredicate, :config do
       }
     end
 
+    context 'simple method call' do
+      context '`EnforcedStyle` is `predicate`' do
+        let(:cop_config) do
+          {
+            'EnforcedStyle' => 'predicate',
+            'IgnoredMethods' => %w[==]
+          }
+        end
+
+        context 'when checking if a number is zero' do
+          it_behaves_like 'code without offense', <<~RUBY
+            if number == 0
+              puts 'hello'
+            end
+          RUBY
+        end
+      end
+
+      context '`EnforcedStyle` is `comparison`' do
+        let(:cop_config) do
+          {
+            'EnforcedStyle' => 'comparison',
+            'IgnoredMethods' => %w[zero?]
+          }
+        end
+
+        context 'when checking if a number is zero' do
+          it_behaves_like 'code without offense', <<~RUBY
+            if number.zero?
+              puts 'hello'
+            end
+          RUBY
+        end
+      end
+    end
+
     context 'in argument' do
       context 'ignored method' do
         context 'when checking if a number is positive' do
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/numeric_predicate_spec.rb
```diff
@@ -188,10 +188,20 @@
       {
         'EnforcedStyle' => 'predicate',
         'AutoCorrect' => true,
-        'IgnoredMethods' => %w[where]
+        'IgnoredMethods' => %w[== where]
```
buehmann: That's not the case discussed in #7595, which has `zero?` in the ignored methods. I guess the reporter's expectation was that he could prevent the _introduction_ of this method as well by adding it to the ignored ones.

spec/rubocop/cop/style/numeric_predicate_spec.rb
```diff
@@ -188,10 +188,20 @@
...
       }
     end
 
+    context 'simple method call' do
+      context 'when checking if a number is positive' do
```
buehmann: 0 is not a positive number.

spec/rubocop/cop/style/numeric_predicate_spec.rb
koic: Thanks for the feedback. I didn't mention my thought on this PR fix in the issue 💦 I added a comment to Issue https://github.com/rubocop-hq/rubocop/issues/7595#issuecomment-569001517 and updated this test case accordingly.

buehmann: Thanks. Those thoughts were missing indeed.

bbatsov:
Looks good! Thanks!

PR #7600, opened 2019-12-28
Remove unused code

RicardoTrindade:
Used the gem zombie_scout to scout for unused methods in the lib. Found a few of them that I'm proposing to remove since they are not being called anywhere. I was also able to remove some specs that were related with the unused code.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/ast/node/mixin/conditional_node.rb
+++ b/lib/rubocop/ast/node/mixin/conditional_node.rb
@@ -14,14 +14,6 @@ module RuboCop
         loc.keyword.line == condition.source_range.line
       end
 
-      # Checks whether the condition of the node is written on more than
-      # one line.
-      #
-      # @return [Boolean] whether the condition is on more than one line
-      def multiline_condition?
-        !single_line_condition?
-      end
-
       # Returns the condition of the node. This works together with each node's
       # custom destructuring method to select the correct part of the node.
       #
--- a/lib/rubocop/ast/node/mixin/method_dispatch_node.rb
+++ b/lib/rubocop/ast/node/mixin/method_dispatch_node.rb
@@ -8,7 +8,6 @@ module RuboCop
       extend NodePattern::Macros
       include MethodIdentifierPredicates
 
-      ARITHMETIC_OPERATORS = %i[+ - * / % **].freeze
       SPECIAL_MODIFIERS = %w[private protected].freeze
 
       # The receiving node of the method dispatch.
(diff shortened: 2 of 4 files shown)


koic:

RicardoTrindade:
@koic done!

koic:
Thanks!

lib/rubocop/ast/node/mixin/conditional_node.rb
```diff
@@ -14,14 +14,6 @@ def single_line_condition?
         loc.keyword.line == condition.source_range.line
       end
 
-      # Checks whether the condition of the node is written on more than
-      # one line.
-      #
-      # @return [Boolean] whether the condition is on more than one line
-      def multiline_condition?
```
bbatsov: This is public API? Do we know its truly unused?

lib/rubocop/ast/node/mixin/method_dispatch_node.rb
```diff
@@ -153,14 +152,6 @@ def block_literal?
         parent&.block_type? && eql?(parent.send_node)
       end
 
-      # Checks whether this node is an arithmetic operation
-      #
-      # @return [Boolean] whether the dispatched method is an arithmetic
-      #                   operation
-      def arithmetic_operation?
```
bbatsov: Same here. Perhaps that's used by some of RuboCop's extensions, plus it sounds like some useful method in general.

RicardoTrindade: checked in rubocop-rails and rubocop-performance and it is not being called, not sure about the other ones

lib/rubocop/ast/node/mixin/conditional_node.rb
RicardoTrindade: checked in the whole project and rubocop-rails and performance and it is not being used. More than happy to revert if you deem it dangerous to remove like this

bbatsov: Well, technically speaking we just did 2 breaking changes. Maybe no one's really using this, but we still have to be careful with such types of changes.

koic: Ah, I only saw repos of rubocop-hq org about using these APIs. Certainly, these APIs are published interfaces, so this change is a breaking change. I was short of attention 💦

Drenmi:
I love the idea of finding unused code in cops. 😍 I think the node extensions can be best thought of as its own library (living inside RuboCop) though. 🙂

bbatsov:
Indeed.

PR #7606, opened 2019-12-30
Remove workaround code for Parser 2.7.0.0

koic:
Follow up of https://github.com/rubocop-hq/rubocop/pull/7597

https://github.com/whitequark/parser/pull/642 has been merged and Parser 2.7.0.1 has been released.
https://rubygems.org/gems/parser/versions/2.7.0.1

-----------------

* [ ] Added tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,9 +11,7 @@ gem 'bump', require: false
 # https://github.com/lsegal/yard/pull/1296
 gem 'e2mmap'
 gem 'irb', '1.0.0'
-# Workaround for Parser 2.7.0.0.
-# It specifies the upper version until Parser 2.7.0.1 release.
-gem 'parser', '>= 2.6', '< 2.7'
+gem 'parser', '>= 2.6'
 gem 'pry'
 gem 'rake', '~> 12.0'
 gem 'rspec', '~> 3.7'


Gemfile
```diff
@@ -11,9 +11,7 @@ gem 'bump', require: false
 # https://github.com/lsegal/yard/pull/1296
 gem 'e2mmap'
 gem 'irb', '1.0.0'
-# Workaround for Parser 2.7.0.0.
-# It specifies the upper version until Parser 2.7.0.1 release.
-gem 'parser', '>= 2.6', '< 2.7'
+gem 'parser', '>= 2.6'
```
buehmann: Shouldn't the whole line be removed? There is no need to repeat the dependency from the gemspec here, is there?

koic: That's right. I should have waited for a review 💦

buehmann: :slightly_smiling_face: And since this PR depends on Parser 2.7.0.1, the dependency in the gemspec should probably be updated.

koic: I think it makes sense to have Parser gem 2.7.0.1 or higher for runtime. I opened the PR #7609. Thanks!

PR #7607, opened 2019-12-30
[Fix #7586] Fix infinite loop in Style/FrozenStringLiteralComment

### LINKED ISSUE #7586: Infinite loop in Style/FrozenStringLiteralComment
I just tried applying Rubocop autocorrect rules on one of old applications and I got infinite loop in ` Style/FrozenStringLiteralComment` rule in a script. The problematic code is below. I know it's a bit incorrect usage of `coding` magic comment but the script wasn't written by me. On the other hand I consider this error as a bug in this case though. 
 
## Expected behavior

Autocorrect is without infinite loop

## Actual behavior

Infinite loop detected in the code below

```
#! bin/rails runner

# coding: utf-8

puts 'DEPRECATED!'
exit
```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.78.0 (using Parser 2.6.5.0, running on ruby 2.6.5 x86_64-darwin18)
```

pirj:
Fixes #7586

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/frozen_string_literal_comment_spec.rb
+++ b/spec/rubocop/cop/style/frozen_string_literal_comment_spec.rb
@@ -89,7 +89,7 @@ RSpec.describe RuboCop::Cop::Style::FrozenStringLiteralComment, :config do
       RUBY
     end
 
-    it 'accepts a dsabled frozen string literal below an encoding comment' do
+    it 'accepts a disabled frozen string literal below an encoding comment' do
       expect_no_offenses(<<~RUBY)
         # encoding: utf-8
         # frozen_string_literal: false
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/frozen_string_literal_comment_spec.rb
```diff
@@ -164,120 +249,21 @@
...
-          #!/usr/bin/env ruby
-        RUBY
-
-        expect(new_source).to eq(<<~RUBY)
-          #!/usr/bin/env ruby
-          # frozen_string_literal: true
-        RUBY
-      end
+    it 'registers an offence for an extra first empty line' do
+      pending 'There is a flaw that skips adding caret symbol in this case' \
```
pirj: Couldn't find a quick way around it. What's even stranger, can't find a difference in `loc` being added in case when caret is added or not.

bbatsov:
Excellent work! I guess we can handle the pending spec down the road.

PR #7612, opened 2020-01-01
[Fix #7611] An unsafe cop cannot have a safe auto-correction

### LINKED ISSUE #7611: Marking a cop as unsafe should imply that its auto-correct is unsafe as well
We found this issue when discussing https://github.com/rubocop-hq/rubocop/pull/7307.

There are two annotations that mark aspects of a cop as (un)safe:
> - Safe (true/false) - indicates whether the cop can yield false positives (by
>   design) or not.
> - SafeAutoCorrect (true/false) - indicates whether the auto-correct the cop
>   does is safe (equivalent) by design.

https://github.com/rubocop-hq/rubocop/blob/23d5e4b7eca052eaf25c63b818ed338f976162bb/manual/auto_correct.md#safe-auto-correct

There are command-line options that select cops and auto-corrections based on those flags:
```
        --safe                       Run only safe cops.
        --safe-auto-correct          Run auto-correct only when it's safe.
```

## Expected behavior

Our assumption is that a cop that is marked as not `Safe` cannot have a safe auto-correction. Otherwise auto-correction would be applied to false positives as well and the result would be unpredictable.

That means that `--safe-auto-correct` must not run auto-correction for cops with `Safe: false`.

## Actual behavior

`--safe-auto-correct` runs auto-correction for unsafe cops.

Pick a cop that only has `Safe: false`, but not `SafeAutoCorrect: false` in the default config:
https://github.com/rubocop-hq/rubocop/blob/8d071f0c9a32f594b696cb7a3b6a0e5adabad42b/config/default.yml#L3344-L3354
```sh-session
$ cat example.rb          
# frozen_string_literal: true

# thief is not a Hash
thief.has_key?(:foo)

$ rubocop -f c example.rb 
example.rb:4:7: C: Style/PreferredHashMethods: Use Hash#key? instead of Hash#has_key?.
thief.has_key?(:foo)
      ^^^^^^^^

$ rubocop -f c --safe example.rb

$ rubocop -f c --safe-auto-correct example.rb
example.rb:4:7: C: [Corrected] Style/PreferredHashMethods: Use Hash#key? instead of Hash#has_key?.
thief.has_key?(:foo)
      ^^^^^^^^

```
Here the (wrong) autocorrect is applied although the cop is supposed to be safe.

buehmann:
This fixes #7611.

* [x] Added tests.

### DIFF
--- a/lib/rubocop/config_validator.rb
+++ b/lib/rubocop/config_validator.rb
@@ -24,6 +24,7 @@ module RuboCop
 
     def validate
       check_cop_config_value(@config)
+      reject_conflicting_safe_settings
 
       # Don't validate RuboCop's own files further. Avoids infinite recursion.
       return if @config.internal?
@@ -166,6 +167,18 @@ module RuboCop
 
       msg = 'Cops cannot be both enabled by default and disabled by default'
       raise ValidationError, msg
+    end
+
+    def reject_conflicting_safe_settings
+      @config.each do |name, cop_config|
+        next unless cop_config.is_a?(Hash)
+        next unless cop_config['Safe'] == false &&
+                    cop_config['SafeAutoCorrect'] == true
+
+        msg = 'Unsafe cops cannot have a safe auto-correction ' \
+              "(section #{name} in #{smart_loaded_path})"
+        raise ValidationError, msg
+      end
     end
 
     def check_cop_config_value(hash, parent = nil)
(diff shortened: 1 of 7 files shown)


bbatsov:
LGTM. @jonas054 Any thoughts?

pirj:
Does it make sense to add config validation to make sure if `Safe` is `false`, `SafeAutoCorrect` can't be set to `true`?

bbatsov:
Makes sense to me.

buehmann:
I consider this pretty much done. I wonder if this requires a changelog entry. Strictly speaking this change is user-observable but is there any real world user that needs to know?

bbatsov:
Probably not. It might be a good idea to document this in the manual, but this can be done separately. Thanks for tackling this!

PR #7614, opened 2020-01-02
Extract TargetRuby class

buehmann:
For #7612 I need room in `ConfigValidator`, which is already too long. This extracts all the logic related to target ruby versions into a separate class.


lib/rubocop/config_validator.rb
```diff
@@ -80,20 +54,26 @@ def validate_section_presence(name)
 
     private
 
+    def target_ruby
+      @target_ruby ||= TargetRuby.new(@config)
+    end
```
Drenmi: Is there a case where it is beneficial to lazy load this? Or can we just assign it in the initializer?

lib/rubocop/target_ruby.rb
```diff
@@ -0,0 +1,153 @@
+# frozen_string_literal: true
+
+module RuboCop
+  # The kind of Ruby that code inspected by RuboCop is written in.
+  class TargetRuby
+    # 2.3 is the oldest officially supported Ruby version.
+    DEFAULT_VERSION = 2.3
```
Drenmi: It seems this is only used in a single place, and we might be better off using `KNOWN_RUBIES.first` instead, to avoid having two sources of truth. WDYT?

lib/rubocop/target_ruby.rb
```diff
@@ -0,0 +1,153 @@
...
+    class RuboCopConfig < Source
+      def name
+        "`TargetRubyVersion` parameter (in #{@config.smart_loaded_path})"
+      end
+
+      private
+
+      def find_version
+        target_ruby_version = @config.for_all_cops['TargetRubyVersion']
+        target_ruby_version&.to_f
```
Drenmi: Don't think we need the assignment here. 🙂 

```ruby
@config.for_all_cops['TargetRubyVersion']&.to_f
```

lib/rubocop/config_validator.rb
buehmann: I guess we can load this eagerly. Let me check.

lib/rubocop/target_ruby.rb
buehmann: Good idea. At least we should have `DEFAULT_VERSION = KNOWN_RUBIES.first`. I'll see if it makes sense to get rid of the constant completely.

bbatsov: I love this idea!

buehmann: I'll keep the constant for now. It is actually used in two places.

lib/rubocop/config_validator.rb
buehmann: This is now assigned in the initializer. I kept a private accessor method, though.

PR #7622, opened 2020-01-03
[Fix #7616] Fix an incorrect autocorrect for `Style/MultilineWhenThen`

### LINKED ISSUE #7616: Style/MultilineWhenThen breaks case statements with {}
## Expected behavior

The Style/MultilineWhenThen cop should remove then statements from all mulitline case statements w/o creating invalid ruby code.

## Actual behavior

If RuboCop encounters a case statement that uses then and {}s it will just remove the then statements, which results in invalid ruby:

Run `rubocop --only Style/MultilineWhenThen` on the following code:

```ruby
      case version.to_s # to_s to make sure someone didn't pass us an int
      when '9.0' then {
        'ActiveDirectory' => 'Veeam Explorer for Microsoft Active Directory',
        'SQL' => 'Veeam Explorer for Microsoft SQL Server',
        'Exchange' => 'Veeam Explorer for Microsoft Exchange',
        'SharePoint' => 'Veeam Explorer for Microsoft SharePoint',
        'Oracle' => 'Veeam Explorer for Oracle'
      }
     end
```

This will autocorrect to:

```ruby
      case version.to_s # to_s to make sure someone didn't pass us an int
      when '9.0' {
        'ActiveDirectory' => 'Veeam Explorer for Microsoft Active Directory',
        'SQL' => 'Veeam Explorer for Microsoft SQL Server',
        'Exchange' => 'Veeam Explorer for Microsoft Exchange',
        'SharePoint' => 'Veeam Explorer for Microsoft SharePoint',
        'Oracle' => 'Veeam Explorer for Oracle'
      }
     end
```

## RuboCop version

```
0.78.0
```

koic:
Fixes #7616.

This PR fixes an incorrect autocorrect for `Style/MultilineWhenThen` for when statement with then is an array or a hash.

```console
% cat example.rb
case condition
when foo then {
    key: 'value'
  }
end

% ruby -c example.rb
Syntax OK
```

It is changed to the code with syntax error as follows.

```console
% bundle exec rubocop -a --only Style/MultilineWhenThen
(snip)

% cat example.rb
case condition
when foo {
    key: 'value'
  }
end

% ruby -c example.rb
example.rb:3: syntax error, unexpected ':', expecting '}'
    key: 'value'
example.rb:4: syntax error, unexpected '}', expecting end-of-input
```

The same is true when replacing a hash with an array.
The PR will accept these cases without offense.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/multiline_when_then.rb
+++ b/lib/rubocop/cop/style/multiline_when_then.rb
@@ -35,7 +35,7 @@ module RuboCop
           return if !node.children.last.nil? && !node.multiline? && node.then?
 
           # With more than one statements after then, there's not offense
-          return if node.children.last&.begin_type?
+          return if accept_node_type?(node.body)
 
           add_offense(node, location: :begin)
         end
@@ -48,6 +48,10 @@ module RuboCop
               )
             )
           end
+        end
+
+        def accept_node_type?(node)
+          node&.begin_type? || node&.array_type? || node&.hash_type?
         end
       end
     end
--- a/spec/rubocop/cop/style/multiline_when_then_spec.rb
+++ b/spec/rubocop/cop/style/multiline_when_then_spec.rb
@@ -57,6 +57,26 @@ RSpec.describe RuboCop::Cop::Style::MultilineWhenThen do
     RUBY
   end
 
+  it 'does not register an offense for hash when statement with then' do
+    expect_no_offenses(<<~RUBY)
+      case condition
+      when foo then {
+          key: 'value'
+        }
+      end
+    RUBY
+  end
+
+  it 'does not register an offense for array when statement with then' do
+    expect_no_offenses(<<~RUBY)
+      case condition
+      when foo then [
+          'element'
+        ]
+      end
+    RUBY
+  end
+
   it 'autocorrects then in empty when' do
     new_source = autocorrect_source(<<~RUBY)
       case foo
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/multiline_when_then.rb
```diff
@@ -35,7 +35,7 @@ def on_when(node)
           return if !node.children.last.nil? && !node.multiline? && node.then?
 
           # With more than one statements after then, there's not offense
-          return if node.children.last&.begin_type?
+          return if accept_node_type?(node.children.last)
```
buehmann: How about `node.body` instead of `node.children.last`?

spec/rubocop/cop/style/multiline_when_then_spec.rb
```diff
@@ -57,6 +57,26 @@
     RUBY
   end
 
+  it 'does not register an offense for hash when statement without then' do
```
buehmann: Why `without then`?

spec/rubocop/cop/style/multiline_when_then_spec.rb
```diff
@@ -57,6 +57,26 @@
...
+    expect_no_offenses(<<~RUBY)
+      case condition
+      when foo then {
+          key: 'value'
+        }
+      end
+    RUBY
+  end
+
+  it 'does not register an offense for array when statement without then' do
```
buehmann: Same here.

lib/rubocop/cop/style/multiline_when_then.rb
koic: That is more readable. Thanks!

Drenmi:
I was expecting the fix to be:

```ruby
case condition
when foo
  {
    key: 'value'
  }
end
```

bur I think the code for that might get a bit more gnarly. 🙂

koic:
Thanks for the advice. I'd like to prioritize that broken Ruby code cannot be corrected by auto-correction. I think the next step (another PR) would be the auto-correction that matched this cop.

tas50:
Thanks for the quick fix @koic

PR #7623, opened 2020-01-03
Reduce class length of MethodCallWithArgsParentheses

buehmann:
This PR achieves

```diff
 Metrics/ClassLength:
-  Max: 197
+  Max: 183
```
by splitting the longest class `MethodCallWithArgsParentheses` into two submodules that implement the two styles supported by the cop, which are largely independent.

The change in `ShowCops` is necessary because it constructed the cops without a configuration before. But `MethodCallWithArgsParentheses` now needs to know its configuration in the initializer.


bbatsov:
Great refactoring! 🙇

PR #7626, opened 2020-01-04
[Fix #7620] Fix a false positive for `Migration/DepartmentName`

### LINKED ISSUE #7620: The Migration/DepartmentName cop doesn't handle comments with double comments
# Expected behavior

The `Migration/DepartmentName` cop shouldn't trigger on a comment like this:

```
# rubocop:disable Style/SafeNavigation # support Ruby < 2.3.0
```

I understand this might not be considered an appropriate RuboCop comment. It's not code within my codebase so I don't really control how the user here decided to write their comments. Regardless of the syntax it seems incorrect to alert the `Migration/DepartmentName` cop for this comment line and worse that the cop claims it autocorrect it every time RuboCop runs.

## Actual behavior

```
foo/libraries/bar/string_utils.rb:46:54: C: [Corrected] Migration/DepartmentName: Department name is missing.
    # rubocop:disable Style/SafeNavigation # support Ruby < 2.3.0
                                                     ^^^^
```

Run `rubocop -a --only Migration/DepartmentName` against a file with the above comment.

## RuboCop version

```
0.78.0
```

koic:
Fixes #7620.

This PR fixes the following false positive for `Migration/DepartmentName` when a disable comment contains a plain comment.

```console
% cat example.rb
# rubocop:disable Style/SafeNavigation # support Ruby < 2.3.0

% rubocop --only Migration/DepartmentName
Inspecting 1 file
C

Offenses:

example.rb:1:50: C: Migration/DepartmentName: Department name is missing.
# rubocop:disable Style/SafeNavigation # support Ruby < 2.3.0
                                                 ^^^^

```

The following is an example disable comment.

```ruby
# rubocop:disable Style/FrozenStringLiteralComment, Style/StringLiterals
```

The token used after `# rubocop: disable` are `A-z`, `/`, `,`.
And `A-z` includes `all`.

```ruby
# rubocop:disable all
```

This PR will change the processing of `Migration/DepartmentName` cop to finish when an invalid token is used.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/migration/department_name.rb
+++ b/lib/rubocop/cop/migration/department_name.rb
@@ -10,13 +10,21 @@ module RuboCop
 
         MSG = 'Department name is missing.'
 
+        # The token that makes up a disable comment.
+        # The token used after `# rubocop: disable` are `A-z`, `/`, and `,`.
+        # Also `A-z` includes `all`.
+        DISABLING_COPS_CONTENT_TOKEN = %r{[A-z/,]+}.freeze
+
         def investigate(processed_source)
           processed_source.each_comment do |comment|
             next if comment.text !~ /\A(# *rubocop:((dis|en)able|todo) +)(.*)/
 
             offset = Regexp.last_match(1).length
             Regexp.last_match(4).scan(%r{[\w/]+|\W+}) do |name|
+              break unless valid_content_token?(name.strip)
+
               check_cop_name(name, comment, offset)
+
               offset += name.length
             end
           end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/migration/department_name.rb
```diff
@@ -10,13 +10,18 @@ class DepartmentName < Cop
 
         MSG = 'Department name is missing.'
 
+        DISABLING_COPS_CONTENT_TOKEN = /[A-z,]+/.freeze
```
bbatsov: What exactly is the content token? Perhaps some comment would be useful here.

koic: Sure! I added a comment about the content token.

bbatsov:
Thanks!

PR #7634, opened 2020-01-07
Enforce frozen string literal true

gfyoung:
Bringing https://github.com/rubocop-hq/rubocop/pull/7435 over the finish line.

* Addressed merge conflicts
* Addressed error message comment from @bbatsov (https://github.com/rubocop-hq/rubocop/pull/7435#issuecomment-570307705)

cc @bbatsov @koic @pirj @parkerfinch

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/frozen_string_literal_comment_spec.rb
+++ b/spec/rubocop/cop/style/frozen_string_literal_comment_spec.rb
@@ -241,7 +241,7 @@ RSpec.describe RuboCop::Cop::Style::FrozenStringLiteralComment, :config do
       RUBY
     end
 
-    it 'registers an offence for not having a frozen string literal comment ' \
+    it 'registers an offense for not having a frozen string literal comment ' \
        'when there is only a shebang' do
       expect_offense(<<~RUBY)
         #!/usr/bin/env ruby
@@ -249,8 +249,8 @@ RSpec.describe RuboCop::Cop::Style::FrozenStringLiteralComment, :config do
       RUBY
     end
 
-    it 'registers an offence for an extra first empty line' do
-      pending 'There is a flaw that skips adding caret symbol in this case' \
+    it 'registers an offense for an extra first empty line' do
+      pending 'There is a flaw that skips adding caret symbol in this case, ' \
               'making it impossible to use `expect_offense` matcher'
 
       expect_offense(<<~RUBY)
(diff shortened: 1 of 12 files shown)


pirj:
What is the difference with #7435 in a nutshell?

https://github.com/rubocop-hq/rubocop/pull/7435#issuecomment-570307705 still stands, I believe this new implementation should replace the existing `always`, not to be added as another option.

Question is - what is more painful, to release an option that might be considered obsolete moving forward, or change the behaviour of existing option that would affect existing users with `frozen_string_literal: false`.

gfyoung:
@pirj : As described in the commit message, the difference between #7435 is that I fixed the merge conflicts and addressed @bbatsov comment about consistent error messages (https://github.com/rubocop-hq/rubocop/pull/7435#discussion_r362199898).

As for https://github.com/rubocop-hq/rubocop/pull/7435#issuecomment-570307705, that is a pretty substantive change you are proposing.  I don't necessarily disagree with the end goal, just not sure that I would want to do it so quickly.

Not sure how fast you guys move in terms of a breaking change like these, but I'll leave it to you maintainers to figure that out.

parkerfinch:
Thanks for pushing this forward @gfyoung!

spec/rubocop/cop/style/frozen_string_literal_comment_spec.rb
```diff
@@ -463,5 +463,715 @@
         puts 1
       RUBY
     end
+
+    context 'auto-correct' do
```
parkerfinch: Since the original PR here was opened, @pirj [updated](https://github.com/rubocop-hq/rubocop/commit/6fd61ce075828ce7d8d82bb8f6e531e4289ebb89#diff-092a0140f3ad2219c13d0abdade6897b) this spec file to use the `expect_correction` helper. I think we should update the new specs to use that approach as well, which will obviate the `auto-correct` context. Does that make sense?

gfyoung: Yep, that makes sense.  I'll refactor the tests to use that.

gfyoung:
@bbatsov @koic : If you can have a look.  I think all comments have been addressed.

bbatsov:
I think this will do for now. Thanks for bringing it over the finish line!

PR #7636, opened 2020-01-07
Remove `console` from `Lint/Debugger`

gsamokovarov:
It's ambiguous and we can play games and try not to recognize:

```ruby
console(with, arguments_or) do
  with_a_block
end
```

But, other use cases will come up. Web Console supports
`binding.console`, which is safe to recognize, so let's drop
the ambiguous `console` one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

bquorning:
`console do` is marked as an offense again in v1.3.0, I think because after #8929.

bbatsov:
Yep, it seems that's a regression.

PR #7637, opened 2020-01-08
7492 Add Style/TrailingCommaInBlockArgs

### LINKED ISSUE #7492: Cop Idea: Useless trailing comma in block arguments

The following two block invocations do different things:

```ruby
[[1, 2, 3], [3, 4, 5]].each { |a| p [a] }
[[1, 2, 3], [3, 4, 5]].each { |a,| p [a] }
```

The following two block invocations do exactly the same thing:

```ruby
[[1, 2, 3], [3, 4, 5]].each { |a, b| p [a, b] }
[[1, 2, 3], [3, 4, 5]].each { |a, b,| p [a, b] }
```

This is surprising, so it makes sense to enforce a consistent style there.

I'd like RuboCop to enforce the removal of the final comma in the second example, so the code reads like so:

```ruby
[[1, 2, 3], [3, 4, 5]].each { |a, b| p [a, b] }
```

The behavior is actually more similar to the one-argument case *with* a comma, so having a trailing comma could be enforced instead. However, visually, removal of the comma is more pleasing, and more in line with other cops that remove extra commas by default. I suppose the enforced style could also be a setting.

## Additional context

N/A.

pawptart:
Fixes #7492. This PR adds new cop `Style/TrailingCommaInBlockArgs` for blocks with multiple arguments. Trailing commas are only required in blocks with one argument, else a trailing comma should be omitted. 

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/useless_trailing_comma.rb
```diff
@@ -0,0 +1,113 @@
...
+
+        def on_block(node)
+          return unless useless_trailing_comma?(node)
+
+          add_offense_for_args(node, node.arguments, ARG_MSG)
+        end
+
+        def autocorrect(node)
+          lambda do |corrector|
+            return unless useless_trailing_comma?(node)
```
Drenmi: Is it redundant to check this again? We will only enter `#autocorrect` if an offense has been registered, and the offense registration is already guarded by this conditional. 🙂

lib/rubocop/cop/lint/useless_trailing_comma.rb
```diff
@@ -0,0 +1,113 @@
...
+              newlines: false
+            ).end_pos
+            range = range_between(node.loc.begin.end.begin_pos, end_pos)
+            corrector.replace(range, block_arg_string(node, node.arguments))
+          end
+        end
+
+        private
+
+        def block_arg_string(node, args)
```
Drenmi: I think `#corrected_source` would be a bit more meaningful. 🙂

lib/rubocop/cop/lint/useless_trailing_comma.rb
```diff
@@ -0,0 +1,113 @@
...
+      #
+      #  # good
+      #  add do
+      #     foo + bar
+      #  end
+      class UselessTrailingComma < Cop
+        include RangeHelp
+
+        MSG = 'Block arguments do not require a trailing comma except when ' \
+              'only a single argument is present.'
```
Drenmi: This seems unused?

lib/rubocop/cop/lint/useless_trailing_comma.rb
```diff
@@ -0,0 +1,113 @@
...
+
+        ARG_MSG = 'Useless trailing comma present in block arguments.'
```
Drenmi: If this is the message we want, we can assign this to `MSG` and not need to pass it around inside the implementation. (It is used by default by `#add_offense`.)

lib/rubocop/cop/lint/useless_trailing_comma.rb
```diff
@@ -0,0 +1,113 @@
...
+        end
+
+        def useless_trailing_comma?(node)
+          arg_count(node) > 1 && trailing_comma?(node)
+        end
+
+        def arg_count(node)
+          %i[arg optarg kwoptarg].map do |arg_type|
+            node.arguments.each_descendant(arg_type).to_a.size
+          end.reduce(:+)
```
Drenmi: No need to `map` here, as `#each_descendant` accepts multiple types:

```ruby
node.arguments.each_descendant(:arg, :optarg, :kwoptarg)
```

lib/rubocop/cop/lint/useless_trailing_comma.rb
```diff
@@ -0,0 +1,113 @@
...
+        end
+
+        def single_line_block?(node)
+          !node.source.include?("\n")
+        end
+
+        def trailing_comma?(node)
+          node.arguments.source.count(',') == arg_count(node)
```
Drenmi: Would it be more robust to check that the last non-whitespace character is `,`?

Right now, I suspect this will return `true` for e.g. an empty block:

```ruby
lambda do ||
end
```

koic:
There are the following differences:

```ruby
% ruby -wve "{foo: 1, bar: 2, baz: 3}.map {|key,| p key }"
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-darwin17]
:foo
:bar
:baz
```

```ruby
% ruby -wve "{foo: 1, bar: 2, baz: 3}.map {|key| p key }"
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-darwin17]
[:foo, 1]
[:bar, 2]
[:baz, 3]
```

So, some are my thoughts.

- I think it's `Style` department, not `Lint` department
- The good and bad cases handled by this cop are incompatible and unsafe
- The trailing comma is an idiom that I often see for me. I think disabled by default is preferable.

lib/rubocop/cop/lint/useless_trailing_comma.rb
buehmann: Counting characters in the unparsed source is inherently dangerous. You might be counting the wrong commas. For example:
```
7637.rb:5:7: W: [Corrected] Lint/UselessTrailingComma: Useless trailing comma present in block arguments.
add { |foo, bar = ','| foo + bar }
      ^^^^^^^^^^^^^^^^
```

lib/rubocop/cop/lint/useless_trailing_comma.rb
buehmann: This seems to destroy semicolons that separate block-local variables:
```
7637.rb:3:7: W: [Corrected] Lint/UselessTrailingComma: Useless trailing comma present in block arguments.
add { |foo, bar,; baz| foo + bar }
      ^^^^^^^^^^^^^^^^
====================
add { |foo, bar, baz| foo + bar }
```
I also don't like the fact that this cop changes the whitespace between the arguments, i.e. might destroy an intentional formatting (alignment, linebreaks, etc.). In my opinion, this cop should only remove a single comma and do nothing more.

lib/rubocop/cop/lint/useless_trailing_comma.rb
```diff
@@ -0,0 +1,113 @@
...
+          arg_count(node) > 1 && trailing_comma?(node)
+        end
+
+        def arg_count(node)
+          %i[arg optarg kwoptarg].map do |arg_type|
+            node.arguments.each_descendant(arg_type).to_a.size
+          end.reduce(:+)
+        end
+
+        def single_line_block?(node)
```
buehmann: Is this the same as `node.single_line?`?

lib/rubocop/cop/lint/useless_trailing_comma.rb
```diff
@@ -0,0 +1,113 @@
...
+
+        def block_arg_string(node, args)
+          arg_string = args.children.map do |arg|
+            if arg.mlhs_type?
+              "(#{block_arg_string(node, arg)})"
+            else
+              arg.source
+            end
+          end.join(', ')
+          arg_string = " |#{arg_string}|"
```
buehmann: For example, that space before `|#{arg_string}|` depends on the configuration of `Layout/SpaceInsideBlockBraces`. That is why we better not touch anything but the comma to be removed.

spec/rubocop/cop/lint/useless_trailing_comma_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+  subject(:cop) { described_class.new(config) }
+
+  let(:config) { RuboCop::Config.new }
+
+  context 'curly brace block format' do
+    it 'registers an offense when a trailing comma is not needed' do
+      expect_offense(<<~RUBY)
+        test { |a, b,| a + b }
+               ^^^^^^^ Useless trailing comma present in block arguments.
+      RUBY
```
rrosenblum: You can use `expect_correction` to keep the offense and correction testing as close together as possible. This will also avoid repeating the source of the code in multiple places.

The test `auto-corrects useless commas` can be removed entirely in favor of adding the following to this test.

```ruby
expect_correction(<<~RUBY)
  test { |a, b| a + b }
RUBY
```

spec/rubocop/cop/lint/useless_trailing_comma_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+    it 'does not auto-correct curly brace blocks without arguments' do
+      new_source = autocorrect_source(<<~RUBY)
+        test { a }
+      RUBY
+      expect(new_source).to eq(<<~RUBY)
+        test { a }
+      RUBY
+    end
+
+    it 'does not auto-correct hash literals' do
```
rrosenblum: I'm not sure that you need this test since hash literals won't register an offense.

spec/rubocop/cop/lint/useless_trailing_comma_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+      new_source = autocorrect_source(<<~RUBY)
+        test { |a,| a }
+      RUBY
+      expect(new_source).to eq(<<~RUBY)
+        test { |a,| a }
+      RUBY
+    end
+
+    it 'does not auto-correct curly brace blocks without arguments' do
+      new_source = autocorrect_source(<<~RUBY)
```
rrosenblum: This test is not necessary because the source code won't register an offense.

pawptart:
This is a lot of feedback to take in. Let me see if I can go back and make the required changes, or if not, pass along to someone who knows better than I.

lib/rubocop/cop/lint/useless_trailing_comma.rb
pawptart: No longer used.

pawptart: Reworked to touch only a single comma.

pawptart:
@koic

> * I think it's `Style` department, not `Lint` department

I didn't change this yet, but can do so if necessary.

> * The good and bad cases handled by this cop are incompatible and unsafe

This shouldn't touch trailing commas that are required (for instance a trailing comma after a single argument), so it should preserve syntax in both cases.

> * The trailing comma is an idiom that I sometimes see for me. I think disabled by default is preferable.

I disabled by default.

config/default.yml
```diff
@@ -1783,6 +1783,11 @@ Lint/UselessSetterCall:
   Enabled: true
   VersionAdded: '0.13'
 
+Lint/UselessTrailingComma:
+  Description: 'Checks for useless trailing commas in block arguments.'
+  Enabled: false
```
koic: Please add `Safe: false`.

```diff
  Enabled: false
+ Safe: false
  VersionAdded: '0.80'
```

lib/rubocop/cop/lint/useless_trailing_comma.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
```
koic: I think it's `Style` department, not `Lint` department.

bbatsov:
I think we need a better name for this cop, as right now it's too generic. Unfortunately I can't think of something that's not very long like `RedundantCommaInBlockParameters`.

bbatsov:
This can potentially get a style guide entry as well.

pawptart:
> I think we need a better name for this cop, as right now it's too generic. Unfortunately I can't think of something that's not very long like `RedundantCommaInBlockParameters`.

Just noticed as I was looking for some naming convention that a `TrailingCommaInArguments` cop also exists but handles redundant commas in method calls, so maybe something like `TrailingCommaInBlockArgs`?

bbatsov:
> Just noticed as I was looking for some naming convention that a TrailingCommaInArguments cop also exists but handles redundant commas in method calls, so maybe something like TrailingCommaInBlockArgs?

@pawptart Yeah, I think that's a good idea. 

It seems also you'll have to do some rebasing looking at the list of merge conflicts. :-)

pawptart:
> @pawptart Yeah, I think that's a good idea.
> 
> It seems also you'll have to do some rebasing looking at the list of merge conflicts. :-)

@bbatsov 

Taken care of :)

bbatsov:
Thanks!

config/default.yml
mvz: Since the one-argument case is not changed, is this still correct?

config/default.yml
```diff
@@ -3853,6 +3853,12 @@ Style/TrailingCommaInArrayLiteral:
     - consistent_comma
     - no_comma
 
+Style/TrailingCommaInBlockArgs:
+  Description: 'Checks for useless trailing commas in block arguments.'
+  Enabled: false
```
j15e, line 3858: According to the new [versioning strategy](https://docs.rubocop.org/en/latest/versioning/), I think this new cop should have been set with `Enable: pending`?

PR #7639, opened 2020-01-08
Fix  edge case in `omit_parentheses` style of `Style/MethodCallWithArgsParentheses`

gsamokovarov:
Before this change:

```ruby
foo(a) || bar(b) do
  pass
end
```

This resulted in a `SyntaxError` if corrected. The check for a call
with a logical operator did not take into account that the parent of a
send node with a block is the block node itself.

After this change, the code below is lawful:

```ruby
foo(a) || bar(b) do
  pass
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


koic:
This looks good to me. Can you rebase with the latest master branch?

gsamokovarov:
Of course, I just rebased it.

koic:
Thank you!

PR #7643, opened 2020-01-11
[Fix #7641] Remove Style/BracesAroundHashParameters cop

### LINKED ISSUE #7641: Remove (or disable) Style/BracesAroundHashParameters
Since Ruby 2.7, the existence of curly braces on the last argument has meaning.

It means, the following codes have different meanings.

```ruby
foo(x: 1) # x: 1 is a keyword argument
foo({x: 1}) # {x: 1} is a hash
```

There are the correct pairs of method definitions and method calls.
```ruby
def foo(x: 1)
end
foo(x: 1)

def foo(hash)
end
foo({x: 1})
```

For more detail https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

So I propose removing the cop. I think we should disable the cop by default at least.

--------

## Expected behavior

Remove or disable the cop.

## Actual behavior

It's available.

nothing

## RuboCop version

master

pocke:
Fix #7641

This pull request removes `Style/BracesAroundHashParameters` cop. Because keyword arguments behavior will be changed in Ruby 3, and has been warned in Ruby 2.7.
For more details, see #7641 and https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

This pull request removes the cop, and related code. Because we had hacks to avoid auto-correction generates wrong code by the cop and others.

-----------------

* [ ] Added tests.


bbatsov:
I'm guessing we'll also need to update the style guide.

spec/rubocop/cli/cli_autocorrect_spec.rb
```diff
@@ -208,196 +206,58 @@ def batch
       let(:comma_style) do
         'comma'
       end
+      let(:expected_corrected_source) do
```
pocke: I only left `Style/BracesAroundHashParameters is disabled` context.

spec/rubocop/cli/cli_autocorrect_spec.rb
```diff
@@ -1010,39 +870,6 @@ def do_something
     expect(IO.read('example.rb')).to eq(corrected)
   end
 
-  it 'corrects complicated cases conservatively' do
```
pocke: This test caes is for the removed cop, so we can remove it.

koic:
FYI, I opened #7648 to resolve ruby-head CI failure.

koic:
Can you add the `Style/BracesAroundHashParameters` cop to `RuboCop::ConfigObsoletion::REMOVED_COPS`?
https://github.com/rubocop-hq/rubocop/blob/v0.79.0/lib/rubocop/config_obsoletion.rb#L60

pocke:
> Can you add the Style/BracesAroundHashParameters cop to RuboCop::ConfigObsoletion::REMOVED_COPS?


I added the cop to the removed cops list in bf79528.
I'll squash the commits into one if you'd like.

koic:
This looks good to me. Can you rebase with the latest master branch and squash your commits into one?

pocke:
Rebased. Thanks for reviewing!

bbatsov:
Great work! Thanks!

bbatsov:
@pocke Btw, I was wondering if we actually should just revert the behaviour of the cop, so that people can migrate smoothly to Ruby 2.7+. Sorry that this came to my mind so late!

pocke:
Thanks for your comment. But I'm not sure I understand your comment correctly due to my poor English. :bowing_man:  I guess you meant you concerned removing the cop is the best way or not.



I guess we can re-implement the cop as meaningful for Ruby 2.7+. Some method can receive both of keyword arguments and a hash, for example, probably ActiveRecord's methods receive both.
In this case, we can choose one of the styles with the cop.
I'll try to implement the cop with the Ruby 3's static type analysis in https://github.com/pocke/rubocop-typed. But currently, I'm not sure the static type is helpful for the cop yet. But if we can implement it, I think it will be useful as a better replacement for Style/BracesAroundHashParameters cop.

And I guess re-implementing a useful replacement is difficult without static types. Because the replacement should be able to find a method definition from a method call, but it is difficult without types.
But perhaps we can implement it for limited cases. For example, focusing `ActiveRecord#find_by(hash_or_keyword)`, because we can guess the method call is ActiveRecord's one easily. But I'm not sure it is useful enough...

---

Please tell me if I misunderstand something :bowing_man: Thanks!

bbatsov:
You've got some great ideas, but I had something simpler in mind - I was merely thinking that probably it makes sense to have a cop that warns people that their code needs to be updated for Ruby 2.7+ - just do the inverse of what it used to do in the past (suggest using braces instead of omitting them).

Anyways, that's definitely not a big deal. Just a random thought I had after merging the PR.

PR #7645, opened 2020-01-12
Use Ruby 2.7 parser for `rake documentation_syntax_check`

koic:
## Summary

This PR uses Ruby 2.7 parser for `rake documentation_syntax_check`.
With this change, example code written using Ruby 2.7 syntax can be analyzed.

## Other infromation

The following error occurs in `Lint/CircularArgumentReference` example code when Ruby 2.7 parser is used.

```consle
% bundle exec rake documentation_syntax_check
Files:         474
Modules:        83 (   11 undocumented)
Classes:       438 (    3 undocumented)
Constants:     631 (  621 undocumented)
Attributes:     31 (    0 undocumented)
Methods:      1081 (  960 undocumented)
 29.55% documented
 lib/rubocop/cop/lint/circular_argument_reference.rb: Syntax Error in an
 example. circular argument reference pie
 lib/rubocop/cop/lint/circular_argument_reference.rb: Syntax Error in an
 example. circular argument reference dry_ingredients
```

The error is avoided by using Ruby 2.6 parser for `Lint/CircularArgumentReference` example code as a workaround.

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [ ] Added tests.
* [ ] Added an entry to the [Changelog](https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format).
* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: https://chris.beams.io/posts/git-commit/

PR #7646, opened 2020-01-12
[Fix #7549] Add `Style/ArgumentsForwarding` cop

### LINKED ISSUE #7549: Cop idea: prefer argument forwarding syntax

Since Ruby 2.7, we will be able to use `...` to forward arguments to another method.

```ruby
# 2.6
def foo(*args, &block)
  bar(*args, &block)
end

# 2.7
def foo(...)
  bar(...)
end
```

I think the `...` syntax has two advantages.

First, we can simply avoid keyword parameter warnings with `...` syntax.
For example

```ruby
# with Ruby 2.7

def bar(*args, a: 1, &block)
  p [args, a]
end

# It displays the warning.
def foo(*args, &block)
  bar(*args, &block)
end
foo 1, 2, a: 1

# It displays no warning, but it is a bit of long.
undef foo
def foo(*args, **kwargs, &block)
  bar(*args, **kwargs, &block)
end
foo 1, 2, a: 1

# No warning, and short!
undef foo
def foo(...)
  bar(...)
end

foo 1, 2, a: 1
```

Second, we can avoid missing `&block` forwarding with `...` syntax.
I guess many people forget passing `&block` to the forwarded method unexpectedly.
But we can avoid the problem with `...` syntax.

Add an offense for the following code.

```ruby
# We can replace the parameter with `...`.
# It can avoid the warning.
# It does not change the behavior at least Ruby 2.7.
# The behavior will be changed in the future.
def foo(*args, &block)
  bar(*args, &block)
end

# We can replace the parameter with `...`.
# The behaviors are completely the same in all Rubies.
# So it is cosmetic change only.
def foo(*args, **kwargs, &block)
  bar(*args, **kwargs, &block)
end

# We can replace the parameter with `...`, but it will change the behavior.
# Because `...` forwards block also.
def foo(*args)
  bar(*args)
end
```

The last example changes the method behavior actually, so probably we should add `Strict` option to disable/enable replacing the last case.

nothing

## Additional context

nothing

koic:
Closes #7549.

This PR adds `Style/ArgumentsForwarding` cop.

As shown at #7549, The following code is unsafe to be arguments forwarding.

```ruby
def foo(*args)
  bar(*args)
end
```

So by setting `AllowOnlyRestArgument` configuration option to true by default, This PR makes the cop safe by default.

-----------------

* [x] Added tests.


koic:
`rake documentation-checks` is failing but it will be resolved by #7645.

koic:
This PR is WIP because it has some false positives.

config/default.yml
```diff
@@ -2202,6 +2202,12 @@ Style/AndOr:
     - always
     - conditionals
 
+Style/ArgumentsForwarding:
+  Description: 'Use arguments forwarding.'
+  Enabled: true
```
bbatsov: Going forward new cops will be introduced with status `pending`.

bbatsov:
We'll also need a matching style guide entry for this cop.

manual/cops_style.md
```diff
@@ -164,6 +164,60 @@ EnforcedStyle | `always` | `always`, `conditionals`
...
+```
+#### Strict: true
+
+```ruby
+# bad
+# The following code can replace the arguments with `...`,
+# but it will change the behavior. Because `...` forwards block also.
+def foo(*args)
+  bar(*args)
+end
```
pocke: I'm wondering if the "Strict" option is confusing. 
I meant when `Strict: true`, the cop adds offenses if it does not change behavior. When `Strict : false`, the cop adds an offense even if it may changes the behavior.
But it looks opposite to the implementation.


```ruby
# I meant the cop does not add an offense to the following code, if `Strict: true`.
# But the implementation and documentation say the opposite.
def foo(*args)
  bar(*args)
end
```

So, I guess we can invert the meaning, change the option name, or do not change anything.

pocke:
I guess it is a good idea to add a test case that delegating block argument is not replaced with `...` by this cop. Because `foo { |...| }` is a syntax error. This cop should not add any offense to the code. 



---



Perhaps This implementation does not treat references of arguments in other places.
So can you add test case(s) for it, and update the implementation to handle it?

```ruby
# This `args` should not be replaced with `...`, but it looks replacing.
def foo(*args)
  args2 = args
  p args2

  bar(*args)
end
```

bbatsov:
@koic How are things looking here?

koic:
The progress in this PR is slow because I have other priorities for RuboCop 1.0. I will close once if things don't work for a while. Thank you for the mention.

manual/cops_style.md
koic: I updated the option name to `AllowOnlyRestArgument`.

koic:
> We'll also need a matching style guide entry for this cop.

Yeah, I opened https://github.com/rubocop-hq/ruby-style-guide/pull/845.

> So can you add test case(s) for it, and update the implementation to handle it?

I added the test cases.

bbatsov (approved):

PR #7650, opened 2020-01-14
Convert tests to use expect_correction

rrosenblum:
This converts most of the lint and layout tests to use `expect_correction`. I found that some tests can't be converted to use `expect_offense` and `expect_correction` because the offense has a zero length range. I forget the exact cops, but this typically happened with white space and extra blank lines. There are other test cases that could be converted, but they will require a full rewrite of the test file to follow the new pattern


bbatsov:
Indeed! 🙇

PR #7657, opened 2020-01-17
Add the method name to the highlight area of Layout/EmptyLineBetweenDefs

rrosenblum:
I came across this while cleaning up tests. We have made similar changes to this in the past. The extra context makes it a little easier to quickly identify which method contains the offense.

### DIFF
--- a/lib/rubocop/cop/layout/empty_line_between_defs.rb
+++ b/lib/rubocop/cop/layout/empty_line_between_defs.rb
@@ -56,7 +56,8 @@ module RuboCop
           return if nodes.all?(&:single_line?) &&
                     cop_config['AllowAdjacentOneLineDefs']
 
-          add_offense(nodes.last, location: :keyword)
+          location = nodes.last.loc.keyword.join(nodes.last.loc.name)
+          add_offense(nodes.last, location: location)
         end
 
         def autocorrect(node)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/layout/empty_line_between_defs.rb
```diff
@@ -56,7 +56,8 @@ def check_defs(nodes)
           return if nodes.all?(&:single_line?) &&
                     cop_config['AllowAdjacentOneLineDefs']
 
-          add_offense(nodes.last, location: :keyword)
+          location = nodes.last.loc.keyword.join(nodes.last.loc.name)
+          add_offense(nodes.last, location: location)
```
koic: Good improvement! It will be more eligible range for context if `:location` kwarg is not specified (default location is `:expression`).

```ruby
add_offense(nodes.last)
```

rrosenblum: I feel like including the entire method in the offense location is a little overkill in this situation. Adding the method name makes it easy to identify the offending method in the console output. Including the entire method will include the method name, but it will also include the entire contents of the method. The contents of the method will only be visible using an output that shows multi-line offenses.

`Lint/DuplicateMethods` highlights the keyword + name. If we are going to change `Layout/EmptyLineBetweenDefs` to include the entire method, should `Lint/DuplicateMethods` be changed as well?

I am fine making the change, however I'm not sure that it is necessary.

koic: I get it. I overlooked consistency with other cops. Thank you for the explanation!

koic:
Thanks!

PR #7677, opened 2020-01-28
Add Hash#each_value and Hash#each_key cop

jemmaissroff:
Added a cop to match the style guide's recommendations around `hash-each`.
This cop enforces correct usage of `Hash#each_key` and `Hash#each_value`
as opposed to `Hash#keys.each` and `Hash#values.each`.

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -2776,6 +2776,12 @@ Style/GuardClause:
   # needs to have to trigger this cop
   MinBodyLength: 1
 
+Style/HashEachMethods:
+  Description: 'Use Hash#each_key and Hash#each_value.'
+  StyleGuide: '#hash-each'
+  Enabled: true
```
bbatsov: Please, change this to `pending`. We're no longer enabling any new cops by default.

lib/rubocop/config_loader.rb
```diff
@@ -55,7 +55,7 @@ def load_file(file)
       end
 
       def add_missing_namespaces(path, hash)
-        hash.keys.each do |key|
+        hash.keys.each do |key| # rubocop:disable Style/HashEachMethods
```
bbatsov: I guess we can just update this, right?

bbatsov:
The code looks solid to me. Thanks for tackling this! 

I guess we should mark the cop as unsafe, though, as there's definitely some potential for false positives.

lib/rubocop/config_loader.rb
jemmaissroff: I actually don't think we can in this case. Due to the `hash[q] = hash.delete(key)` line below, we'll get a `     RuntimeError: can't add a new key into hash during iteration` if we try do a `hash.each_key` here.

jemmaissroff:
> The code looks solid to me. Thanks for tackling this!
> 
> I guess we should mark the cop as unsafe, though, as there's definitely some potential for false positives.

Thanks for looking, I've pushed another commit where I marked it as unsafe, and also pending!

lib/rubocop/config_loader.rb
bbatsov: Ops. my bad. Should have read the code more carefully. :-)

CHANGELOG.md
```diff
@@ -99,6 +100,7 @@
 * [#7446](https://github.com/rubocop-hq/rubocop/issues/7446): Add `merge` to list of non-mutating methods. ([@cstyles][])
 * [#7077](https://github.com/rubocop-hq/rubocop/issues/7077): **(Breaking)** Rename `Unneeded*` cops to `Redundant*` (e.g., `Style/UnneededPercentQ` becomes `Style/RedundantPercentQ`). ([@scottmatthewman][])
 * [#7396](https://github.com/rubocop-hq/rubocop/issues/7396): Display assignments, branches, and conditions values with the offense. ([@avmnu-sng][])
+* [#7434](https://github.com/rubocop-hq/rubocop/issues/7434): Fix an incorrect autocorrect for `Style/MultilineWhenThen` when the body of `when` branch starts with `then`. ([@koic][])
```
parkerfinch: Looks like this might be a rebase artifact? The same line appears above in the `Bug Fixes` section.

config/default.yml
```diff
@@ -2776,6 +2776,13 @@ Style/GuardClause:
   # needs to have to trigger this cop
   MinBodyLength: 1
 
+Style/HashEachMethods:
+  Description: 'Use Hash#each_key and Hash#each_value.'
+  StyleGuide: '#hash-each'
+  Enabled: pending
+  VersionAdded: '0.79'
```
parkerfinch: Should this be updated to `'0.80'`? Looks like `0.79` was released already.

lib/rubocop/config_loader.rb
parkerfinch: Could this whole issue be sidestepped with a
```ruby
hash.transform_keys { |key| Cop::Cop.qualified_cop_name(key, path) }
```
? Or do we need to support rubies < 2.5?

CHANGELOG.md
koic: Yeah, this change is not the difference in this PR.

config/default.yml
koic: Sure.

lib/rubocop/cop/style/hash_each_methods.rb
```diff
@@ -0,0 +1,87 @@
...
+      #
+      #   # good
+      #   hash.each_key { |k| p k }
+      #   hash.each_value { |v| p v }
+      class HashEachMethods < Cop
+        include Lint::UnusedArgument
+
+        MSG = 'Use `%<prefer>s` instead of `%<current>s`.'
+
+        def_node_matcher :kv_each, <<-PATTERN
```
koic: Can you update this entire PR with squiggly heredoc? (`<<~`)

spec/rubocop/cop/style/hash_each_methods_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      end
+
+      it 'does not register an offense for Hash#each_value' do
+        expect_no_offenses('foo.each_value { |v| p v }')
+      end
+
+      it 'auto-corrects foo#keys.each with foo#each_key' do
+        new_source = autocorrect_source('foo.keys.each { |k| p k }')
+        expect(new_source).to eq('foo.each_key { |k| p k }')
+      end
```
koic: Can you use `expect_correction` for test an automatic correction? The following address will be helpful.
https://github.com/rubocop-hq/rubocop/pull/7650

rrosenblum:
I knew this cop sounded familiar to me. We have `Performance/InefficientHashSearch` in the Performance project for changing `keys.include?` and `values.include?` to `key?` or `has_key?` and `value?` or `has_value?`. Does this cop make more sense as a Style Cop or a Performance Cop?


I also think we may have tried to implement this same cop in the past. I don't think this change is safe if you are mutating the keys or values while using `each_key` or `each_value`. 

This code does not work if you change `keys.each` to `each_key` because we remove keys
`lib/rubocop/config_loader.rb`

```ruby
      def add_missing_namespaces(path, hash)
        hash.keys.each do |key|
          q = Cop::Cop.qualified_cop_name(key, path)
          next if q == key

          hash[q] = hash.delete(key)
        end
      end
```

Darhazer:
We used to have [Performance/HashEach](https://github.com/rubocop-hq/rubocop-performance/commit/884ce6886fc5dc7cc0146b5ed3e39b77c7a08d6c). Not sure what happened to it

jemmaissroff:
> We used to have [Performance/HashEach](https://github.com/rubocop-hq/rubocop-performance/commit/884ce6886fc5dc7cc0146b5ed3e39b77c7a08d6c). Not sure what happened to it.
> 
> Update: Found [the removal of the cop](https://github.com/rubocop-hq/rubocop-performance/commit/6e58dd728225ece32aa238424dc8a441fd426c8a)

Yup, and there was [an issue filed](https://github.com/rubocop-hq/rubocop-performance/issues/66) where it was suggested to instead by a Style PR. I didn't link it as part of the commit message because it's in a different repo and so I think the link would break. Let me know if I should be doing something differently here!

bbatsov:
@jemmaissroff Just address the failing build and I think we're good here.

bbatsov:
Actually - it might be cool to also make the cop configurable, so that people could decide whether they prefer `hash.foo.each` or `hash.each_foo`. Generally we try to be inclusive of most people's preferences. :-) Anyways, that's just an idea - if you don't have time for this it's fine.

rrosenblum:
> it might be cool to also make the cop configurable

If we are classifying this as a Style Cop, then I completely agree that it would be cool for this to be configurable. If we were to classify it as a Performance Cop then I don't think the configuration makes sense.

jemmaissroff:
@bbatsov I can't quite figure out why the buillds are failing or how to reproduce locally. Any thoughts?

jemmaissroff:
@bbatsov build is fixed!

.rubocop.yml
```diff
@@ -28,6 +28,9 @@ Style/FormatStringToken:
   Exclude:
     - spec/**/*
 
+Style/HashEachMethods:
+  Enabled: false
```
koic: This setting looks unnecessary.

config/default.yml
```diff
@@ -2778,6 +2778,13 @@ Style/GuardClause:
   # needs to have to trigger this cop
   MinBodyLength: 1
 
+Style/HashEachMethods:
+  Description: 'Use Hash#each_key and Hash#each_value.'
+  StyleGuide: '#hash-each'
+  Enabled: false
```
koic: Can you update to `Enabled: pending` status?
https://github.com/rubocop-hq/rubocop/pull/7677#discussion_r371999553

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * [#7663](https://github.com/rubocop-hq/rubocop/pull/7663): Add new `Style/HashTransformKeys` and `Style/HashTransformValues` cops. ([@djudd][], [@eugeneius][])
 * [#7619](https://github.com/rubocop-hq/rubocop/issues/7619): Support autocorrect of legacy cop names for `Migration/DepartmentName`. ([@koic][])
 * [#7659](https://github.com/rubocop-hq/rubocop/pull/7659): Layout/LineLength autocorrect now breaks up long lines with blocks. ([@maxh][])
+* Add a cop for `Hash#each_key` and `Hash#each_value`. ([@jemmaissroff][])
```
koic: Can you add the PR link to this changelog entry?

.rubocop.yml
jemmaissroff: I had a build failing that prompted me to add it here. Can you tell me more about why this looks unnecessary?

koic: By default, it is better to use the setting in config/default.yml.

jemmaissroff: I think I'm getting confused around what `pending` will do. It appears that breaks builds also. Are you suggesting to leave it as `pending` and then remove it from here?

koic: Yeah, config/default.yml leaves the specification of `Enabled: pending`. OTOH, I expect this setting to be removed from .rubocop.yml.

> It appears that breaks builds

Can you show me what build error is occurring?

bbatsov:
It seems you need to rebase to `master` and run the rake command that regenerates the manual and the build will pass.

jemmaissroff:
Ah okay, that was it! Builds are now passing. Thanks for the help @bbatsov and @koic !

bbatsov:
🎉 

Thanks for working on this! 🙇

PR #7690, opened 2020-02-06
Fix apparent typo in README

ianfixes:
Generally if you want to install something using `bundler`, you _do_ require it in your `Gemfile`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ automatically fix some of the problems for you.
 $ gem install rubocop
 ```
 
-If you'd rather install RuboCop using `bundler`, require it in your `Gemfile`:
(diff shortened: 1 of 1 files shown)


bbatsov:
Read carefully the example snippet. :-) That's not a typo - you want RuboCop in your Gemfile, but you don't want to require it, as it's a stand-alone tool.

ianfixes:
Aah.  In that case it's unclear that you're saying "you _do_ put the the line in your `Gemfile` (i.e. you conceptually _require_ the dependency) but you set Ruby's `require` option to false".  I'll try to emphasize that.

bbatsov:
Thanks!

PR #7691, opened 2020-02-06
Fix documentation [doc]

marcandre:
None of these methods actually accept an array of symbols.


bbatsov:
Nice catch! Probably a leftover of some earlier implementation.

PR #7694, opened 2020-02-07
Replace Node#str_content with StrNode#value

Drenmi:
This is a small step in a quest to move overly generic methods from `Node` into the relevant node extensions. In this case `Node#str_content` is moved to `StrNode#value`.

-----------------

* [X] Added tests.

### DIFF
--- a/spec/rubocop/ast/str_node_spec.rb
+++ b/spec/rubocop/ast/str_node_spec.rb
@@ -56,4 +56,31 @@ RSpec.describe RuboCop::AST::StrNode do
       it { expect(str_node.heredoc?).to be(true) }
     end
   end
+
+  describe '#value' do
+    context 'with a normal string' do
+      let(:source) { "'foo'" }
+
+      it { expect(str_node.value).to eq('foo') }
+    end
+
+    context 'with a string with interpolation' do
+      let(:source) { '"foo #{bar} baz"' }
+
+      it { expect(str_node.value).to eq('foo  baz') }
+    end
+
+    context 'with a heredoc' do
+      let(:source) do
+        <<~RUBY
+          <<-CODE
+            foo
+            bar
+          CODE
+        RUBY
+      end
+
+      it { expect(str_node.value).to eq("  foo\n  bar\n") }
+    end
+  end
 end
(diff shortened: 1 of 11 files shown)


spec/rubocop/ast/str_node_spec.rb
```diff
@@ -56,4 +56,31 @@
...
+    context 'with a normal string' do
+      let(:source) { "'foo'" }
+
+      it { expect(str_node.value).to eq('foo') }
+    end
+
+    context 'with a string with interpolation' do
+      let(:source) { '"foo #{bar} baz"' }
+
+      it { expect(str_node.value).to eq('foo  baz') }
```
Drenmi: This may or may not be the desired behaviour, but in any case it is the current one. If we want to change this we can do so down the line. At least now the behaviour is documented. 🙂

koic:
I think it's better to add an changelog entry to the CHANGELOG.md because this change affects 3rd party cops. For example, it affects rubocop-rails and rubocop-rspec in rubocop-hq organization.

```console
% ag str_content
rubocop/lib/rubocop/cop/style/format_string_token.rb
90:          token_ranges(str_contents(str_node.loc), &block)
93:        def str_contents(source_map)

rubocop-rails/lib/rubocop/cop/rails/date.rb
137:            node.receiver.str_content.match(zone_regexp)

rubocop-rails/lib/rubocop/cop/rails/lexically_scoped_action_filter.rb
141:            [node.str_content.to_sym]
148:                v.str_content.to_sym

rubocop-rspec/lib/rubocop/cop/rspec/file_path.rb
79:          "*#{name.str_content.gsub(/\W/, '')}" unless ignore_methods?

rubocop-rspec/lib/rubocop/cop/rspec/describe_method.rb
28:          return if second_arg.str_content.start_with?('#', '.')
```

Drenmi:
Thanks for pointing that out @koic! I will have a look at those gems as well.

bbatsov:
Perhaps `str_content` can stay for now with some deprecation warning?

koic:
This change may cause an error for 3rd party gems not maintained by rubocop-hq when using `Node#str_content`. The changelog is an announcement for those users. If possible, a more user-friendly approach would be to display deprecation warning.
I think it's good to have it in the changelog first.

bbatsov:

PR #7699, opened 2020-02-08
[Fix #7680] Add new `Lint/StructNewOverride` cop

### LINKED ISSUE #7680: New cop: Detect unexpected overwriting of Struct methods

I've encountered an weird behavior on using `Struct` class:

```ruby
# good
Good = Struct.new(:foo)
Good.new.members #=> [:foo]

# bad
Bad = Struct.new(:members)
Bad.new.members #=> nil (I've expected `[:members]`)
```

This is caused by unexpectedly overwriting the [`Struct#members`](https://ruby-doc.org/core-2.7.0/Struct.html#method-i-members) method.

Add a new cop to detect unexpected overwriting such as the example above.

None.

## Additional context

`Struct` includes `Enumerable` and has many methods e.g. `length`, `count`, `min`, `max`, or `sum`, so it seems that some people might use such methods as a `Struct` member unexpectedly.

ybiquitous:
Fix #7680 

This adds a new `Lint/StructNewOverride` cop to check unexpected overrides of [`Struct`](https://ruby-doc.org/core-2.7.0/Struct.html) original methods.

I originally planned to name it "Lint/StructOverride", but I think that disallowing all the overrides is too strict, so I rename it to `Lint/StructNewOverride`. I intend that this cop limits to overrides via arguments in `Struct.new(...)`.

However, the cop name may need to be discussed more.

Furthermore, as I commented on <https://github.com/rubocop-hq/rubocop/issues/7680#issuecomment-583751130>, it may need more discussion on whether the cop should check **all** the methods (from `Object` or `Enumerable`) or not.

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/struct_new_override.rb
```diff
@@ -0,0 +1,50 @@
...
+        STRUCT_METHOD_NAMES = Struct.instance_methods
+
+        def_node_matcher :struct_new, <<~PATTERN
+          (send
+            (const ${nil? cbase} :Struct) :new ...)
+        PATTERN
+
+        def on_send(node)
+          return unless struct_new(node) do
+            node.arguments.each do |arg|
```
pocke: If the first argument is a String, it will be the class name. So the cop should ignore the first argument if it is a String, but it does not ignore it.

lib/rubocop/cop/lint/struct_new_override.rb
```diff
@@ -0,0 +1,50 @@
...
+              next unless arg.respond_to?(:value)
+
+              method_name = arg.value
```
pocke: I guess it needs `to_sym` because the method name may be a String.

```ruby
# Both of :foo and 'bar' are method names.
Struct.new(:foo, 'bar')
```

lib/rubocop/cop/lint/struct_new_override.rb
ybiquitous: Make sense. 👍 I've addressed via cb46c1a1c2204e7b0ce94ae8c6074a2b85133e94.

lib/rubocop/cop/lint/struct_new_override.rb
ybiquitous: Oh, I couldn't notice this spec of `Struct.new` to allow string member names! 😳 
I've addressed via cb46c1a1c2204e7b0ce94ae8c6074a2b85133e94. Thank you!

---

By the way, we may be able to add a new cop to check the usage of string member names. 😅 (Or, we need a type checker...?)
Because it seems that string member names can be allowed only for backward compatibility.

ybiquitous:
By following <https://github.com/rubocop-hq/rubocop/issues/7680#issuecomment-583806040>, I've mad the cop message milder via f10bc600aa3f96e962bce9b7d9d4822d2a48eb8a.

E.g.
> `:members` member overrides `Struct#members` and it may be unexpected.

lib/rubocop/cop/lint/struct_new_override.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # This cop checks unexpected overrides of the `Struct` original methods
```
bbatsov: Probably `built-in` is a better term than original here.

lib/rubocop/cop/lint/struct_new_override.rb
```diff
@@ -0,0 +1,58 @@
...
+      #   b.count #=> 1 (overriding `Enumerable#count`)
+      #
+      #   # good
+      #   Good = Struct.new(:id, :name)
+      #   g = Good.new(1, "foo")
+      #   g.members #=> [:id, :name]
+      #   g.clone #=> #<struct Good id=1, name="foo">
+      #   g.count #=> 2
+      #
+      class StructNewOverride < Cop
```
bbatsov: And then we can name this `StructBuiltInMethodOverride` which is more descriptive IMO.

bbatsov:
Apart from my small remarks - great work!

lib/rubocop/cop/lint/struct_new_override.rb
ybiquitous: Thank you. I also think that `StructBuiltInMethodOverride` is easier to understand than `StructNewOverride`. 👍 
I'll fix it.

ybiquitous: I assume that this cop detects some overrides by arguments of `Struct.new`, not normal overrides as follows:

```ruby
Foo = Struct.new(:name) do
  def to_s
    name.to_s
  end
end
```

I'm a bit concerned that `StructBuiltInMethodOverride` may be considered as **any** overrides of the built-in methods. 🤔

ybiquitous: For example, `StructBuiltInMethodOverrideByConstructor` may be more descriptive but too long. 😓 
I may think too much... Please give me any help! 🙏

bbatsov: Naming is hard! :D Let's keep the shorter name and down the road we can maybe extend this cop to cover dangerous overrides by explicit definitions as well.

ybiquitous: Make sense. Thanks! 👍

manual/cops_lint.md
```diff
@@ -1520,7 +1520,7 @@ rather than meant to be part of the resulting symbols.
 
 Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
 --- | --- | --- | --- | ---
-Enabled | Yes | No | 0.75 | -
+Pending | Yes | No | 0.81 | -
```
ybiquitous: This change is by `rake generate_cops_documentation` but why...?

bbatsov: I messed up, but @koic just fixed my mistake. Just rebase on top of `master` and this will dissapper.

ybiquitous: Thanks for your response! I've fixed it!

ybiquitous:
I've addressed all the reviews. Please review again if you have time. 🙏

bbatsov:
🚀

PR #7700, opened 2020-02-09
Optionally relax checks by SpaceBeforeComma and SpaceInsideParens after kwargs colons

amilligan:
I like these cops (Layout/SpaceBeforeComma and LayoutSpaceInsideParens), but I also like my good friend Layout/SpaceAfterColon.  They don't all get along well when it comes to keyword arguments.  For example:

```ruby
def some_kwarg_method(foo: , bar: )
end
```

The SpaceBefore/SpaceInside cops will mark the spaces after the kwargs as violations.  I like the visual consistency of having spaces after the colons.  I added the `space_after_colon` style to each of the complaining cops.  In both cases it expects no spaces *except* in the case of a kwarg with no value.

I believe I've properly updated the documentation and specs.  Each should autocorrect correctly with the new style.

For the moment, I left separate commits on the branch for easier review.  I will update the changelog once I have an ID for this PR.


amilligan:
Any thoughts or suggestions on this?  Our team has been using it for a while with no problems.

Drenmi:
Hi, @amilligan!

Thank you for your contribution. 🙂 I like the spirit of this change!

I'm a little irked by the roundabout way of looking up a node from a token, but I realize this is a result of the existing implementation of these cops.

I will have a look at this in the coming two weeks. If I can't suggest anything better by then, I think I am ready to approve this change.

amilligan:
@Drenmi I agree about the roundabout way of lookup, but I couldn't find anything better.  This is my first foray into rubocop internals, so I was flying a little blind.  If you find a better way I look forward to seeing it.

Thanks for taking a look.

amilligan:
Just rebased on the latest.  No conflicts, and all tests still pass, so I think everything is good.

Drenmi:
Thanks, @amilligan!

I had a (very brief) look at it. Couldn't find any immediate way of resolving my concern for now, so I'm going ahead and approving this, as it is still a legitimate enhancement. 🙂

CHANGELOG.md
```diff
@@ -38,6 +38,8 @@
 * Add `BracesRequiredMethods` parameter to `Style/BlockDelimiters` to require braces for specific methods such as Sorbet's `sig`. ([@maxh][])
 * [#7686](https://github.com/rubocop-hq/rubocop/pull/7686): Add new `JUnitFormatter` formatter based on `rubocop-junit-formatter` gem. ([@koic][])
 * [#7715](https://github.com/rubocop-hq/rubocop/pull/7715): Add `Steepfile` to default `Include` list. ([@ybiquitous][])
+* [#7700](https://github.com/rubocop-hq/rubocop/pull/7700): Add `space_after_colon` style to Layout/SpaceBeforeComma and Layout/SpaceInsideParens cops. ([@amilligan][])
+
```
koic: Can you remove the redundant blank line and move the entry to under the `### New features` section of the `## master (unreleased)`?

CHANGELOG.md
```diff
@@ -38,6 +38,8 @@
 * Add `BracesRequiredMethods` parameter to `Style/BlockDelimiters` to require braces for specific methods such as Sorbet's `sig`. ([@maxh][])
 * [#7686](https://github.com/rubocop-hq/rubocop/pull/7686): Add new `JUnitFormatter` formatter based on `rubocop-junit-formatter` gem. ([@koic][])
 * [#7715](https://github.com/rubocop-hq/rubocop/pull/7715): Add `Steepfile` to default `Include` list. ([@ybiquitous][])
+* [#7700](https://github.com/rubocop-hq/rubocop/pull/7700): Add `space_after_colon` style to Layout/SpaceBeforeComma and Layout/SpaceInsideParens cops. ([@amilligan][])
```
koic: It is preferable that cop names be surrounded by backticks.

```suggestion
* [#7700](https://github.com/rubocop-hq/rubocop/pull/7700): Add `space_after_colon` style to `Layout/SpaceBeforeComma` and `Layout/SpaceInsideParens` cops. ([@amilligan][])
```

config/default.yml
```diff
@@ -1111,9 +1111,16 @@ Layout/SpaceBeforeBlockBraces:
...
+  Description: >-
+                 No spaces before commas, exception optionally in the case of
+                 keyword arguments.
   Enabled: true
+  EnforcedStyle: no_space
+  SupportedStyles:
+    - no_space
+    - space_after_colon
   VersionAdded: '0.49'
+  VersionChanged: '0.80.0'
```
koic: RuboCop 0.80.0 has been released and tiny version is not required. Can you update to `0.81`?

```suggestion
  VersionChanged: '0.81'
```

config/default.yml
```diff
@@ -1209,11 +1216,12 @@ Layout/SpaceInsideParens:
   StyleGuide: '#spaces-braces'
   Enabled: true
   VersionAdded: '0.49'
-  VersionChanged: '0.55'
+  VersionChanged: '0.80.0'
```
koic:
```suggestion
  VersionChanged: '0.81'
```

CHANGELOG.md
amilligan: I've fixed up the original commit; this should be done.

CHANGELOG.md
amilligan: I've fixed up the original commit; this should be done.

config/default.yml
amilligan: I've fixed up the original commit; this should be done.

amilligan:
One quick hiccough -- I forgot to regenerate the documentation once I made the requested changes -- but I believe everything should now be good to go.

bbatsov:
Thanks for the PR! The code looks good to me overall, but I have my doubts whether we should merge those changes or not. 

I haven't seen this style in the wild, so I'm a bit reserved to adding support for something that might not be used that much. I'll have to ruminate on the subject a bit.

amilligan:
@bbatsov I can honestly say the majority of projects I've worked on for the past several years have used this style.  I this of this style as analogous to anonymous javascript functions:
```
var x = function() {
  ...
```
vs
```
var x = function () {
  ...
```

This change shouldn't affect anyone who prefers the no-space option.  Without the change, the two cops are unusable for those of us who prefer the space-after-colon option.

bbatsov:
> This change shouldn't affect anyone who prefers the no-space option. Without the change, the two cops are unusable for those of us who prefer the space-after-colon option.

I totally get this, but my concern is that we (RuboCop's team) are the ones to who end up maintaining more code in the long run. That's why we generally aim to support code styles with wide adoption, so the benefits of the extra complexity are worth it. 

So far you're the only person who asked for something like this, which for me is a strong indication that the style is not particular common, therefore the need for extra consideration. 

```
def some_kwarg_method(foo: , bar: )
```

On more fundamental note, I'm wondering how much does the extra space stand out. 

@rubocop-hq/rubocop-core Any thoughts on this? I saw that @Drenmi is OK with the idea, but I'm curious what the others think as well.

koic:
> I totally get this, but my concern is that we (RuboCop's team) are the ones to who end up maintaining more code in the long run. That's why we generally aim to support code styles with wide adoption, so the benefits of the extra complexity are worth it.

I agree with this opinion. It may be preferable to implement it as a 3rd party cop instead of supporting in the core.

I haven't seen this style until this PR.

```ruby
def some_kwarg_method(foo: , bar: )
```

Below is an example I tried. I have more than 600 repositories in my local environment, but few matched the following regular expressions. The following uses a regular expression that matches the proposed new optional style.

```console
% cd `ghq root`
% ag 'def .*:.* ,' | wc -l
      11
```

It depends on my personal environment, but it is an indicator I can show.

amilligan:
I don't think extending the attempt to convince you this is a valid style is worth anyone's time.  What are the alternatives, so we can continue using the SpaceBeforeComma and SpaceInsideParens cops?  

I saw mention of a third-party cop.  Do you suggest just copying the bulk of the cops in question into a new, private cop, and turning off the official versions?

jonas054:
> I saw mention of a third-party cop. Do you suggest just copying the bulk of the cops in question into a new, private cop, and turning off the official versions?

Yes, I'd say that makes the most sense. IMO it's a rare style that's not worth the cost in terms of added code in RuboCop itself. (It *is* possible that the style is rare because many projects use RuboCop, but even so...)

PR #7702, opened 2020-02-10
Improve documentation for Lint/UselessSetterCall

jonas054:
This closes #7309, which was an attempt to fix false positives with code. It was deemed too complex, so we're adding documentation for a, hopefully, very rare problem.

The cop configuration now marks the cop as "not safe".


bbatsov:
Thanks!

PR #7710, opened 2020-02-12
[Fix #7705] Fix problem of `Style/OneLineConditional` cop with elsif.

### LINKED ISSUE #7705: Rubocop incorrectly tries to transform one line conditional to ternary operator
Rubocop tries to correct a one line conditional with elsif clauses to a ternary operator, but the way it does it is broken and leads to a syntax error.

## Expected behavior

The actual expected behavior is a bit of a product decision that I don't dare to make. It could either leave this line unchanged or transform it to multiple ternary operators.

## Actual behavior

Rubocop messes up lines with a one line conditional that contains an elsif clause and the result has a syntax error.

Create a file with the following silly bit of code in an otherwise empty directory:

```
# frozen_string_literal: true
digit = gets.to_i % 10
nice_digit = if digit >= 5 then 5 elsif digit >= 2 then 2 else 1 end
puts nice_digit
```

Double check that it works by running it and seeing that you can enter an integer and it produces some output.

Run `rubocop -a` which will try to correct the one line conditional.

See that the line with the one line conditional is now messed up. Running the file produces a syntax error.

## RuboCop version

0.79.0 (using Parser 2.7.0.2, running on ruby 2.5.1 x86_64-linux-gnu)

Lykos:
One line conditionals with elsif branches are now transformed into multiline versions. Before this change, they were broken and autocorrect produced code that resulted in a syntax error. See Issue #7705 for more details.

### DIFF
--- a/lib/rubocop/cop/style/one_line_conditional.rb
+++ b/lib/rubocop/cop/style/one_line_conditional.rb
@@ -25,11 +25,13 @@ module RuboCop
       class OneLineConditional < Cop
         include OnNormalIfUnless
 
-        MSG = 'Favor the ternary operator (`?:`) ' \
-              'over `%<keyword>s/then/else/end` constructs.'
+        MSG_USE_TERNARY = 'Favor the ternary operator (`?:`) ' \
+                          'over `%<keyword>s/then/else/end` constructs.'
+        MSG_USE_MULTILINE = 'Use multiple lines for ' \
+                            '`if/then/elsif/then/else/end` constructs.'
 
         def on_normal_if_unless(node)
-          return unless node.single_line? && node.else_branch
+          return unless node.single_line? && node.else_branch && !node.elsif?
 
           add_offense(node)
         end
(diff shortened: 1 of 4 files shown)


Lykos:
Friendly ping?

manual/cops_naming.md
```diff
@@ -350,7 +350,7 @@ This cop can be configured with the EnforcedStyleForLeadingUnderscores
 directive. It can be configured to allow for memoized instance variables
 prefixed with an underscore. Prefixing ivars with an underscore is a
 convention that is used to implicitly indicate that an ivar should not
-be set or referenced outside of the memoization method.
+be set or referencd outside of the memoization method.
```
bbatsov: Seems you unfixed a typo. :-)

lib/rubocop/cop/style/one_line_conditional.rb
```diff
@@ -25,11 +25,13 @@ module Style
       class OneLineConditional < Cop
         include OnNormalIfUnless
 
-        MSG = 'Favor the ternary operator (`?:`) ' \
-              'over `%<keyword>s/then/else/end` constructs.'
+        MSG_USE_TERNARY = 'Favor the ternary operator (`?:`) ' \
+                          'over `%<keyword>s/then/else/end` constructs.'
+        MSG_USE_MULTILINE = 'Use multiple lines for ' \
```
bbatsov: Seems to me that this change is more than a bugfix - it actually changes the behaviour of the cop. Why would we want to treat this case differently?

bbatsov:
Sorry for the slow response. I've added one question, as I'm not sure that this is exactly a bugfix. Also - when doing functional changes you always have to update the cop documentation as well, but let's first agree on what exactly do we want to do.

Lykos:
I agree about having to update the description.

But note that this is really a bugfix even though it introduces new behavior. But it introduces new behavior *only* in the case that was previously broken. 

One line conditionals with an `elsif` were previously broken and autocorrect created a syntax error. To fix it, we have to make a choice on what should happen with them. I went with putting one line conditionals with `elsif` to multiple lines. Of course there are other options.

See the discussion on #7705. But since nobody reacted, I decided to just go ahead with the version that I considered most reasonable.

bbatsov:
> But note that this is really a bugfix even though it introduces new behavior. But it introduces new behavior only in the case that was previously broken.

Fair enough. 

> One line conditionals with an elsif were previously broken and autocorrect created a syntax error. To fix it, we have to make a choice on what should happen with them. I went with putting one line conditionals with elsif to multiple lines. Of course there are other options.

I think the fundamental problem was that we tried to make those ternary conditions instead of just make them multi-line, which is obviously a more generic solution to the problem. I'm wondering if by default we shouldn't just make everything multiline and have an additional flag to convert one line conditions to ternary ops only when feasible. I think that's going to make the cop way more consistent. 

> See the discussion on #7705. But since nobody reacted, I decided to just go ahead with the version that I considered most reasonable.

Sorry about this. Unfortunately there are many open tickets and we don't have the  time to respond to all of them in a timely fashion. Overall I think your direction makes sense, but we need to make the cop's behaviour more uniform as stated above.

bbatsov:
@Lykos ping :-)

Lykos:
I think I won't have time to do this any more, so consider this stale.

PR #7714, opened 2020-02-13
[Close #7480] Add new Style::NestedInterpolation cop

### LINKED ISSUE #7480: Not to use nested string interpolation
Hello, 

What do you think about adding a cop which can find examples of using nested interpolation (when we use string interpolation into another string interpolation). Just for me, this is a good example of the bad design and places which should be refactored.

Example,

```ruby
def say_hello(user = nil)
  puts "Hello #{ user.present? ? "dear #{user.first_name}" : "everyone" }. How are you?"
end
```

Thank you!

roman-dubrovsky:
Fix #7480 

This PR adds new `Style::NestedInterpolation` cop which checks so that developers don't use interpolation block into another in interpolation blocks. Also, this PR fixes some issues which related to new cop -  we had nested interpolation in the next two places:

- https://github.com/rubocop-hq/rubocop/blob/2ec092a013e11be5a6f7475623d70a7a44e6abcc/lib/rubocop/cop/style/parallel_assignment.rb#L253
- https://github.com/rubocop-hq/rubocop/blob/2ec092a013e11be5a6f7475623d70a7a44e6abcc/lib/rubocop/cop/style/parallel_assignment.rb#L268

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -3176,6 +3176,11 @@ Style/NegatedWhile:
   Enabled: true
   VersionAdded: '0.20'
 
+Style/NestedInterpolation:
+  Description: 'Avoid using nested interpolation.'
+  Enabled: true
```
bbatsov: New cops should be "pending" by default.

lib/rubocop/cop/style/nested_interpolation.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # This cop checks nested interpolations
```
bbatsov: I think you can expand the documentation of this cop with more rationale and examples that are following the same conventions as those in existing cops. Might be a good idea to link the cop to a style guide entry.

lib/rubocop/cop/style/nested_interpolation.rb
```diff
@@ -0,0 +1,37 @@
...
+      #
+      # @example
+      #
+      #   # bad
+      #   "Hello, #{user.blank? ? 'guest' : "dear #{user.name}"}"
+      #
+      #   # good
+      #   user_name = user.blank? ? 'guest' : "dear #{user.name}"
+      #   "Hello, #{user_name}"
+      class NestedInterpolation < Cop
```
bbatsov: I guess this should named `RedundantNestedInterpolation` to be aligned with the typical naming scheme we employ.

lib/rubocop/cop/style/nested_interpolation.rb
roman-dubrovsky: @bbatsov I've updated the description a little bit but I'm not sure that this is enough. (But the style should be the same as in existing cops). Also, added one more example.

> Might be a good idea to link the cop to a style guide entry.

Does it mean that we need to add it to the style guide? And can you guide me on how to do it?

lib/rubocop/cop/style/redundant_nested_interpolation.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # This cop checks for nested interpolation expressions
+      # which can be splitted.
```
bbatsov: can be rewritten to avoid it.

bbatsov:
On second thought - this doesn't actually seem like redundant nested interpolation, so probably `NestedInterpolation` is a better name for the cop. I mean you need it for the code to work the way you want it to. Probably the message should also say `Avoid the use of nested interpolation.` or something along those lines. 

> Does it mean that we need to add it to the style guide? And can you guide me on how to do it?

Just file a PR here https://github.com/rubocop-hq/ruby-style-guide

bbatsov:

PR #7721, opened 2020-02-16
[Fix #7708] Support EOL disable comments on comment lines

### LINKED ISSUE #7708: End-of-line disable comments can't be used to disable an offense in a comment
Per @jonas054 in #1625:
```
End-of-line disable comments can't be used to disable an offense in a comment, currently. It will work like the other kind of disable comment, the one that needs an enable. It's an edge case, but it seems like a bug to me.

@jeffkowalski Please open a new issue. If you mention me in it, I will try to fix the problem.
```
--------

## Expected behavior

Disable an offense for a single-line comment, returning the enable state afterward

## Actual behavior

Disable comment has no effect on code comment

```
$ cat test.rb 
# coding: utf-8     # rubocop:disable Style/Encoding
# frozen_string_literal: true

$ rubocop test.rb
Inspecting 1 file
W

Offenses:

test.rb:1:1: W: Lint/MissingCopEnableDirective: Re-enable Style/Encoding cop with # rubocop:enable after disabling it.
# coding: utf-8     # rubocop:disable Style/Encoding
^

```
## RuboCop version

```
$ rubocop --version
0.78.0
```

jonas054:
A hash mark `#` on a comment line doesn't start a new comment, so we have to pretend that it does in order to support disabling of cops on a single comment line.

### DIFF
--- a/lib/rubocop/comment_config.rb
+++ b/lib/rubocop/comment_config.rb
@@ -113,13 +113,18 @@ module RuboCop
     def each_mentioned_cop
       each_directive do |comment, cop_names, disabled|
         comment_line_number = comment.loc.expression.line
-        single_line = !comment_only_line?(comment_line_number)
+        single_line = !comment_only_line?(comment_line_number) ||
+                      directive_on_comment_line?(comment)
 
         cop_names.each do |cop_name|
           yield qualified_cop_name(cop_name), disabled, comment_line_number,
                 single_line
         end
       end
+    end
+
+    def directive_on_comment_line?(comment)
+      comment.text[1..-1].match(COMMENT_DIRECTIVE_REGEXP)
     end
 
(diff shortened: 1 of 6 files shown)


lib/rubocop/comment_config.rb
```diff
@@ -122,6 +123,10 @@ def each_mentioned_cop
       end
     end
 
+    def comment_on_comment_line?(comment)
```
Drenmi: What do you think about making this `#directive_on_comment_line?`, since we match the directive regexp inside? 🙂

jonas054: That's much better! I'll fix it.

Drenmi:
Great catch! 🚀

koic:
Thanks!

PR #7747, opened 2020-02-22
[Fix #7739] Allow methods raising NotImplementedError in `Lint/UnusedMethodArgument`

### LINKED ISSUE #7739: Allow unused variables when a class method only raises a NotImplementedError

While the [style guide](https://github.com/rubocop-hq/ruby-style-guide#duck-typing) prefers duck typing over inheritance, many developers prefer inheritance for the sake of communicating structure to other developers. If a method both has arguments and raises a `NotImplementedError` – a common pattern for defining method signatures – the `Lint/UnusedMethodArgument` warning is triggered.

I would like functions with arguments that only contain `raise NotImplementedError` to be ignored by `Lint/UnusedMethodArgument`. Ideally, there would be a setting (maybe `IgnoreNotImplementedMethods`) that is true by default.

This should not trigger any warnings

```
class Example
  def process_thing(thing)
    raise NotImplementedError
  end
end
```

Currently you can use the following work arounds

#### Leave a method empty, because `IgnoreEmptyMethods` is true by default
This is insufficient, because in many cases calling a method should immediately fail if it hasn't been implemented by a child class. Otherwise, bugs can appear far deeper in the execution path.

#### Prefix the variables with an underscore
This is insufficient, as it communicates that the prefixed argument(s) are optional. They may not be.

#### Don't include arguments in the signature
This is insufficient, as it fails to communicate the arguments child classes should expect for a method.

tejasbubane:
Closes #7739

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1768,9 +1768,10 @@ Lint/UnusedMethodArgument:
   StyleGuide: '#underscore-unused-vars'
   Enabled: true
   VersionAdded: '0.21'
-  VersionChanged: '0.35'
+  VersionChanged: '0.81'
   AllowUnusedKeywordArguments: false
   IgnoreEmptyMethods: true
+  IgnoreNotImplementedMethods: true
 
 Lint/UriEscapeUnescape:
   Description: >-
--- a/lib/rubocop/cop/lint/unused_method_argument.rb
+++ b/lib/rubocop/cop/lint/unused_method_argument.rb
@@ -38,8 +38,33 @@ module RuboCop
       #   def do_something(unused)
       #   end
       #
+      # @example IgnoreNotImplementedMethods: true (default)
+      #   # good
+      #   def do_something(unused)
+      #     raise NotImplementedError
+      #   end
+      #
+      #   def do_something_else(unused)
+      #     fail "TODO"
+      #   end
+      #
+      # @example IgnoreNotImplementedMethods: false
+      #   # bad
+      #   def do_something(unused)
+      #     raise NotImplementedError
+      #   end
+      #
+      #   def do_something_else(unused)
+      #     fail "TODO"
+      #   end
+      #
       class UnusedMethodArgument < Cop
         include UnusedArgument
+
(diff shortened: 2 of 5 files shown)


lib/rubocop/cop/lint/unused_method_argument.rb
```diff
@@ -38,9 +38,25 @@ module Lint
...
+      #   # bad
+      #   def do_something(unused)
+      #     raise NotImplementedError
+      #   end
+      #
       class UnusedMethodArgument < Cop
         include UnusedArgument
 
+        def_node_matcher :raise_not_implemented?, <<~PATTERN
+          (send nil? :raise (const nil? :NotImplementedError))
```
bbatsov: You should also cover `fail`.

config/default.yml
```diff
@@ -1739,6 +1739,7 @@ Lint/UnusedMethodArgument:
   VersionChanged: '0.35'
   AllowUnusedKeywordArguments: false
   IgnoreEmptyMethods: true
+  IgnoreNotImplementedMethods: true
```
bbatsov: Version changed needs to be updated.

bbatsov:
Apart from my small remarks I'm fine with the proposed changes.

config/default.yml
tejasbubane: Updated.

lib/rubocop/cop/lint/unused_method_argument.rb
tejasbubane: Handled.

bbatsov:
Thanks!

sandstrom:
@tejasbubane Awesome addition! 

What do you think about [Lint/UnreachableCode](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/lint/unreachable_code.rb) having this check too?

Basically such that it would allow:

```ruby
def my_method
  raise NotImplementedError
  
  1 + 1
end
```

PR #7751, opened 2020-02-23
Fix  Node Pattern documentation

marianosimone:
There was an unclosed code block that was messing up everything below "` for descendants"

[This is the documentation](https://docs.rubocop.org/en/latest/node_pattern/#using-node-matcher-macros)

And it looks like:

Screen Shot 2020-02-23 at 12 54 17 PM


koic:
Thanks!

PR #7761, opened 2020-02-26
[Fix #7737] Add new `Style/UnnecessaryArguments` cop

### LINKED ISSUE #7737: Remove unnecessary arguments for builtin methods.

I found something like `array.join('')` in my code recently. This can of course be replaced by `array.join`. There is also `string.split("\n")` or `string.gsub(a, '')`. I will try to collect more cases like this in this issue (help is welcome).

For a list of important builtin functions (e.g. `String#split` and `Array#join`), if an argument is passed explicitly that leads to the same behavior as leaving the argument away, suggest to remove that argument.

## Additional context

Note that I might do this myself eventually. Depends on how much time I have and how eager someone else is to take it.

tejasbubane:
Closes https://github.com/rubocop-hq/rubocop/issues/7737

-----------------

* [x] Added tests.


bbatsov:
I think that's not the best approach to building such a cop. A much better one would be to avoid hardcoding anything and build the rules for the methods and their redundant params via the configuration of the cop.

lib/rubocop/cop/style/unnecessary_arguments.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for unnecessary arguments to builtin methods
+      #
+      # @example
+      #   # bad
+      #
```
bbatsov: You misplaced the blank line here.

lib/rubocop/cop/style/unnecessary_arguments.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for unnecessary arguments to builtin methods
```
bbatsov: I'd expand on the description here (and ideally go beyond builtin methods with some form of configuration).

lib/rubocop/cop/style/unnecessary_arguments.rb
```diff
@@ -0,0 +1,42 @@
...
+      #   string.split(" ")
+      #   "first\nsecond".split(" ")
+      #
+      #   # good
+      #   array.join
+      #   [1, 2, 3].join
+      #   string.split
+      #   "first second".split
+      #
+      class UnnecessaryArguments < Cop
```
koic: For example, `RedundantSeparatorArgument` cop may be preferred for cop naming.
https://github.com/rubocop-hq/rubocop/issues/7077

config/default.yml
```diff
@@ -3898,6 +3898,12 @@ Style/UnlessElse:
   Enabled: true
   VersionAdded: '0.9'
 
+Style/UnnecessaryArguments:
+  Description: 'Check for unnecessary arguments to builtin methods.'
+  Enabled: true
```
koic: Can you specify `Enabled: pending` instead of `Enabled: true`.

lib/rubocop/cop/style/unnecessary_arguments.rb
tejasbubane: Since @bbatsov suggested extending this cop beyond builtin methods, naming this `RedundantArguments`.

config/default.yml
tejasbubane: Changed to `pending`.

lib/rubocop/cop/style/redundant_arguments.rb
```diff
@@ -0,0 +1,73 @@
...
+        end
+
+        # Parser wraps empty strings inside quotes
+        # this is a workaround to compare
+        def unwrap(string)
+          case string
+          when "''", '""'
+            ''
+          when '" "', "' '"
+            ' '
```
tejasbubane: This is a bad workaround & will fail with cases like `'    '` (string with spaces). Can anyone help me fix this?

tejasbubane:
@bbatsov Changed cop logic to be configurable for any method & argument.

bbatsov:
@tejasbubane Looks good to me overall. I'd only expand the docs to mention that people should be aware that the cop can't tell apart methods with the same name in different classes, so they should be careful with the cop.

bbatsov:
And that this is limited to methods that take a single param. To make it more generic we should be passing both a param position and a param in the config (e.g. if the redundant param is the second one), but I guess such cases are quite uncommon.

lib/rubocop/cop/style/redundant_arguments.rb
```diff
@@ -0,0 +1,64 @@
...
+      #   array.join
+      #   [1, 2, 3].join
+      #   string.split
+      #   "first second".split
+      #   A.foo
+      class RedundantArguments < Cop
+        MSG = 'Use `%<without_argument>s` instead of `%<with_argument>s`.'
+
+        def on_send(node)
+          arg = cop_config['RedundantArguments'][node.method_name.to_s]
```
bbatsov: Might be a good idea to add some validation logic for the config, so we can present the users with more meaningful error messages if they mess something up.

lib/rubocop/cop/style/redundant_arguments.rb
bbatsov: I guess first you can squeeze the extra spaces before passing them to the cases. Maybe someone else from @rubocop-hq/rubocop-core has better ideas.

lib/rubocop/cop/style/redundant_arguments.rb
```diff
@@ -0,0 +1,64 @@
...
+      #   A.foo(2)
+      #
+      #   # good
+      #   array.join
+      #   [1, 2, 3].join
+      #   string.split
+      #   "first second".split
+      #   A.foo
+      class RedundantArguments < Cop
+        MSG = 'Use `%<without_argument>s` instead of `%<with_argument>s`.'
```
bbatsov: I guess it'd be clearer to say that the arg is redundant.

lib/rubocop/cop/style/redundant_arguments.rb
bquorning: You could convert the configured values into AST nodes and compare directly against `node.arguments.first`.

I played a bit with it and got as far as this:
```rb
require 'parser/current'

def on_send(node)
  return unless redundant_argument?(node)
  # ...
end

def redundant_argument?(node)
  redundant_argument = redundant_arg_for_method(node.method_name.to_s)
  return false if redundant_argument.nil?

  node.arguments.first == redundant_argument
end

def redundant_arg_for_method(method_name)
  return nil unless cop_config['RedundantArguments'].key?(method_name)

  @mem ||= {}
  @mem[method_name] ||= begin
    arg = cop_config['RedundantArguments'].fetch(method_name)

    buffer = Parser::Source::Buffer.new('(string)', 1)
    buffer.source = arg.inspect
    builder = RuboCop::AST::Builder.new
    Parser::CurrentRuby.new(builder).parse(buffer)
  end
end
```

I am not _completely_ sure that a new buffer and builder instance is needed for each configuration, so the memoization may need to be moved around a bit.

bbatsov:
@tejasbubane ping :-)

tejasbubane:
@bbatsov Changes done.

marcandre:
Technically this cop is unsafe, as the actual default argument to join is `$OUTPUT_FIELD_SEPARATOR || ''` and that of split is `$FIELD_SEPARATOR || ' '`; if these globals are set, then `join()` and `split()` have different meaning than `join('')` and `split(' ')`

bbatsov:

tejasbubane:
@bbatsov Looks like I don't have the rights to reopen this PR. Do you mind if I open a new one?

tejasbubane:
> Technically this cop is unsafe, as the actual default argument to join is $OUTPUT_FIELD_SEPARATOR || '' and that of split is $FIELD_SEPARATOR || ' '; if these globals are set, then join() and split() have different meaning than join('') and split(' ')

@marcandre Thanks. The cop is marked unsafe. Also, these kernel variables are [deprecated since 2.7](https://github.com/ruby/ruby/pull/2920)

marcandre:
> I don't have the rights to reopen this PR. Do you mind if I open a new one?

Can't either. Says the branch was force-pushed or recreated. Not sure why github prevents it. Anyways, you can open a new one.

>  The cop is marked unsafe. 

Good, I must have forgotten to check, sorry.

PR #7762, opened 2020-02-26
[Fix #7740] Make `Style/AccessModifierDeclarations` not flag symbol method-name

### LINKED ISSUE #7740: False Positive and Incorrect Error Message for Style/AccessModifierDeclarations
RuboCop 0.80.0 is reporting offences for marking existing methods as private, even when the method is not defined within the scope it is marked private in.
Furthermore, the message produced by the offence is misleading.

--------

## Expected behavior
I expect to be able to mark methods introduced by reflection or otherwise defined outside of the current scope as protected or private.

## Actual behavior
Consider a hypothetical example using a Rails model:
```ruby
ApplicationRecord.connection.create_table(:articles) do |t|
  t.string :name
  t.string :slug
end

class Article < ApplicationRecord
  def name=(value)
    super
    self.slug = name&.parameterize
  end

  private :slug=
end
```

The rubocop offence is reported on the `private :slug=` line:
> `private` should not be inlined in method definitions. (convention:Style/AccessModifierDeclarations)

The reported offence is incorrect since:
* The line is not a method definition, hence the error message is incorrect and misleading.
* The line affects a method that is not defined by application code. It is created by Rails based on the connected database schema, and its natural access modifier is outside of application control.

Minimum case to reproduce:
```ruby
module Bar
  def bar
    nil
  end
end

class Foo
  include Bar
  private :bar
end
```
The error message is produced:
> `private` should not be inlined in method definitions. (convention:Style/AccessModifierDeclarations)

The line neither defines a method, nor does class `Foo` itself declare the method `bar`.

## RuboCop version
```
0.80.0 (using Parser 2.7.0.2, running on ruby 2.7.0 x86_64-linux)
```

tejasbubane:
Closes #7740

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2212,10 +2212,12 @@ Style/AccessModifierDeclarations:
   Description: 'Checks style of how access modifiers are used.'
   Enabled: true
   VersionAdded: '0.57'
+  VersionChanged: '0.81'
   EnforcedStyle: group
   SupportedStyles:
     - inline
     - group
+  AllowModifiersOnSymbols: true
 
 Style/Alias:
   Description: 'Use alias instead of alias_method.'
--- a/lib/rubocop/cop/style/access_modifier_declarations.rb
+++ b/lib/rubocop/cop/style/access_modifier_declarations.rb
@@ -5,11 +5,12 @@ module RuboCop
     module Style
       # Access modifiers should be declared to apply to a group of methods
       # or inline before each method, depending on configuration.
+      # EnforcedStyle config covers only method definitions.
+      # Applications of visibility methods to symbols can be controlled
+      # using AllowModifiersOnSymbols config.
       #
       # @example EnforcedStyle: group (default)
-      #
       #   # bad
-      #
       #   class Foo
       #
       #     private def bar; end
@@ -18,7 +19,6 @@ module RuboCop
       #   end
       #
       #   # good
-      #
       #   class Foo
       #
       #     private
@@ -27,10 +27,9 @@ module RuboCop
       #     def baz; end
       #
       #   end
+      #
       # @example EnforcedStyle: inline
-      #
       #   # bad
-      #
       #   class Foo
       #
       #     private
@@ -41,11 +40,26 @@ module RuboCop
       #   end
       #
       #   # good
-      #
       #   class Foo
       #
       #     private def bar; end
       #     private def baz; end
+      #
+      #   end
+      #
+      # @example AllowModifiersOnSymbols: true
+      #   # good
+      #   class Foo
+      #
+      #     private :bar, :baz
+      #
+      #   end
+      #
+      # @example AllowModifiersOnSymbols: false
+      #   # bad
+      #   class Foo
+      #
+      #     private :bar, :baz
       #
       #   end
       class AccessModifierDeclarations < Cop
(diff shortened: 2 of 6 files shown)


lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -65,6 +65,9 @@ def on_send(node)
           return unless node.access_modifier?
           return if node.parent.pair_type?
 
+          argument = node.children.last
```
bbatsov: I wonder if we can simplify this somehow. Generally I've never been fond of getting all the children in this fashion. I'd rather go with simple destructuring of the node.

lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -65,6 +65,9 @@ def on_send(node)
...
+          return if argument.is_a?(AST::Node) && argument.sym_type?
```
bbatsov: Why the explicit type check?

bbatsov:
The cop documentation should definitely be updated to reflect this. 

I think we also need some flag to enable/disable this, as if you're using modifiers in this manner, you're potentially violating grouping/non-grouping. That's why I'd prefer having more control over the behaviour.

tejasbubane:
@bbatsov Fixed to use node-matcher instead of explicit checks. 

Regarding config, do you think the following configs makes sense?
1. group (default) - forbids inline method definition but allows inline symbol
2. strict_group - forbids inline method definition & symbols
3. inline - all modifiers should be inline

Let me know if you have something else in mind.

Mihail-K:
@tejasbubane 
The act of allowing inline-symbol seems tangential to the modes of `group` and `inline`, since those refer to definition style, whereas referencing a Symbol is used for existing declarations.

Maybe it might make more sense to have it be a flag/separate option?
Allowing or forbidding the inline-Symbol behaviour could be applicable regardless of whether the preferred declaration style is inline or group.

bbatsov:
> Maybe it might make more sense to have it be a flag?

Exactly. This shouldn't be a style, it should a flag that affects both styles - something like `AllowModifiersOnSymbols` or whatever. A better name would be welcome. :-)

tejasbubane:
@bbatsov I added config option to `AllowModifiersOnSymbols`.

lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -48,6 +48,24 @@ module Style
       #     private def baz; end
       #
       #   end
+      # @example AllowModifiersOnSymbols: true
```
bbatsov: The formatting here seems off.

config/default.yml
```diff
@@ -2202,6 +2202,7 @@ Style/AccessModifierDeclarations:
   SupportedStyles:
     - inline
     - group
+  AllowModifiersOnSymbols: true
```
bbatsov: You'll have to update `VersionChanged` as well.

lib/rubocop/cop/style/access_modifier_declarations.rb
bbatsov: Hmm, I see now it's messed up in the entire file. I guess you just stuck to the existing examples.

config/default.yml
tejasbubane: Added version `0.81`

lib/rubocop/cop/style/access_modifier_declarations.rb
tejasbubane: Do you mean the empty line before `@example`?

bbatsov: That and the empty lines between `bad/good`. Just compare this to other source files and you'll see what I mean.

bbatsov:
I'd also extend the cop documentation to mention that it covers only definitions and applications of visibility methods to symbols are controlled by the new config.

lib/rubocop/cop/style/access_modifier_declarations.rb
tejasbubane: Fixed.

tejasbubane:
> I'd also extend the cop documentation to mention that it covers only definitions and applications of visibility methods to symbols are controlled by the new config.

Added.

bbatsov:
Thanks!

PR #7780, opened 2020-03-06
[Fix #6289] Naming/FileName: Add ExpectMatchingClass option

### LINKED ISSUE #6289: Naming/FileName cop's ExpectMatchingDefinition doesn't work with custom directories

In Rails, you can add directories to `config.autoload_paths`, and these directories don't have to conform to the standard Rails namespacing conventions.

For example, suppose you have the file `app/models/news/photo_story.rb`. Normally, you'd have to name the class in that file `News::PhotoStory`. But if you add `app/models/news` to `config.autoload_paths`, you can instead name the file `PhotoStory`, and autoloading will still work.

This usually isn't the best thing to do, but people do it.

If you have `ExpectMatchingDefinition: true`, Rubocop will say something like

> photo_story.rb should define a class or module called `News::PhotoStory`.

It would be nice if you could specify `autoload_paths` in the Rubocop configuration, and have Rubocop instead look for unnamespaced classes in those directories. Something like

```
Naming/FileName:
  Enabled: true
  ExpectMatchingDefinition: true
  CustomDirectories:
    - app/models/news
```

Then it would omit the `News::` prefix when looking for constants in those files. It would still report an offense if you named the class something like `PhotosStory`.

I could exclude the relevant paths, but then I'd lose out the check for those files (including the snake-case filename check).

jschneid:
https://github.com/rubocop-hq/rubocop/issues/6289 describes a limitation of the existing `ExpectMatchingDefinition: true` option on the `Naming/FileName` cop: It generates false positives when run on a project with classes which (by design) aren't namespaced with modules to match their file paths.

This PR addresses that issue by adding a new option to the `Naming/FileName` cop: `CheckDefinitionPathHieararchy`. This defaults to `true`; if set to `false`, it modifies the `ExpectMatchingDefinition` option to _only_  check whether each source file's class or module name matches the file name -- not whether the nested module hierarchy matches the subdirectory path.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#6289](https://github.com/rubocop-hq/rubocop/issues/6289): Add new `CheckDefinitionPathHierarchy` option for `Naming/FileName`. ([@jschneid][])
+
 ### Bug fixes
 
 * [#8008](https://github.com/rubocop-hq/rubocop/issues/8008): Fix an error for `Lint/SuppressedException` when empty rescue block in `def`. ([@koic][])
@@ -4542,3 +4546,4 @@
 [@laurmurclar]: https://github.com/laurmurclar
 [@jethrodaniel]: https://github.com/jethrodaniel
 [@CvX]: https://github.com/CvX
+[@jschneid]: https://github.com/jschneid
--- a/config/default.yml
+++ b/config/default.yml
@@ -2031,6 +2031,10 @@ Naming/FileName:
   # It further expects it to be nested inside modules which match the names
   # of subdirectories in its path.
   ExpectMatchingDefinition: false
+  # When `false`, changes the behavior of ExpectMatchingDefinition to match only
+  # whether each source file's class or module name matches the file name --
+  # not whether the nested module hierarchy matches the subdirectory path.
+  CheckDefinitionPathHierarchy: true
   # If non-`nil`, expect all source file names to match the following regex.
   # Only the file name itself is matched, not the entire file path.
   # Use anchors as necessary if you want to match the entire name rather than
--- a/lib/rubocop/cop/naming/file_name.rb
+++ b/lib/rubocop/cop/naming/file_name.rb
@@ -49,23 +49,34 @@ module RuboCop
 
         def for_bad_filename(file_path)
           basename = File.basename(file_path)
-          msg = if filename_good?(basename)
-                  return if matching_definition?(file_path)
 
-                  no_definition_message(basename, file_path)
-                else
-                  return if bad_filename_allowed?
+          if filename_good?(basename)
+            msg = perform_class_and_module_naming_checks(file_path, basename)
+          else
+            msg = other_message(basename) unless bad_filename_allowed?
+          end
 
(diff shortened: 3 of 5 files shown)


bbatsov:
> If RuboCop is configured to enable both the ExpectMatchingDefinition and ExpectMatchingClass options, then the ExpectMatchingDefinition option (which is more strict) takes precedence.

I think it's much better to just add an option for checking the path hierarchy instead, as it seems the new option you introduced is the same as the other sans the path checking. For me the suggested solution is confusing, but valuable - let's simplify it. :-)

jschneid:
> I think it's much better to just add an option for checking the path hierarchy instead, as it seems the new option you introduced is the same as the other sans the path checking. For me the suggested solution is confusing, but valuable - let's simplify it. :-)

Bozhidar, thanks for the comment! I'd like to fix up this PR to address your suggestion; however, I'm not entirely sure what you're suggesting -- could you please clarify?

The `ExpectMatchingDefinition` option on the `Naming/FileName` cop, as it exists today, when enabled, essentially enforces two additional checks:

1. That the file's module namespace matches its filesystem path;
2. That the name of the class or module defined in the file matches the filename.

This PR's new proposed option (`ExpectMatchingClass`), when enabled, checks only item 2 above.

Are you suggesting that we add an additional new option that would check only item 1 above? Such that we'd end up with `Naming/FileName` cop options `ExpectMatchingDefinition` (which checks 1 + 2), `ExpectMatchingClass` (which checks 2), and another option -- `ExpectMatchingPathHierarchy`, perhaps? -- which checks 1?

I'd be happy to take that on -- I just want to check my understanding of whether that's what you're suggesting, first. 🙂

jschneid:
@bbatsov, this is 100% _not_ urgent, but I wanted to bump my above question for you, just in case you hadn't seen it yet. 🙂

bbatsov:
Sorry about the slow reply. My bad, my previous reply was completely wrong. I was under the impression that the cop already was checking for some definition and I had assumed you wanted to limit the full path checking and focus only on checking for a definition. I saw this was not the case, so it seems everything's good with your PR. Just rebase to resolve the merge conflict, please.

bbatsov:
Ah, forget what I said. I'm too tired for meaningful replies. Actually seems I wasn't wrong last time around! What I meant was that instead of having `ExpectMatchingClass` and `ExpectMatchingDefinition` we can have something like `CheckDefinitionPathHieararchy` which would alter the behaviour of `ExpectMatchingDefinition`. Alternatively we can make `ExpectMatchingDefinition` accept non-boolean values, so you can specify there something like `match_hierarchy`, `match_filename`, etc. I'm just thinking that one of those approaches will result in more understandable configuration options.

jschneid:
Ok! As per your suggestion, @bbatsov, I've adjusted this PR to instead add a new `CheckDefinitionPathHieararchy`option to the `Naming/FileName` cop. Details are in the updated PR description, above.

jschneid:
I've rebased this PR to fix the typical merge conflict in the `CHANGELOG.md`. 

@bbatsov, is there anything else I can do to help with this at this point? 🙂

bbatsov:
Thanks and sorry for forgetting about this PR!

Issue #7774, opened 2020-03-03
Crash on pattern matching in ruby 2.7


artofhuman:
Crash on pattern matching
--------

## Expected behavior

No crash

## Actual behavior
```
`undefined method `type' for nil:NilClass`
```

<details>
  <summary>Stacktrace</summary>

```
undefined method `type' for nil:NilClass
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/ast/traversal.rb:57:in `block in on_in_pattern'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/ast/traversal.rb:57:in `each'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/ast/traversal.rb:57:in `on_in_pattern'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/cop/commissioner.rb:33:in `block (2 levels) in <class:Commissioner>'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/ast/traversal.rb:57:in `block in on_case_match'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/ast/traversal.rb:57:in `each'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/ast/traversal.rb:57:in `on_case_match'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/cop/commissioner.rb:33:in `block (2 levels) in <class:Commissioner>'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/ast/traversal.rb:68:in `on_lvasgn'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/cop/commissioner.rb:33:in `block (2 levels) in <class:Commissioner>'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/ast/traversal.rb:57:in `block in on_begin'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/ast/traversal.rb:57:in `each'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/ast/traversal.rb:57:in `on_begin'
/usr/local/bundle/gems/rubocop-0.80.1/lib/rubocop/cop/commissioner.rb:33:in `block (2 levels) in <class:Commissioner>'
[...]

PR #7786, opened 2020-03-09
Support Ruby 2.7's pattern match for `Layout/ElseAlignment` cop

### LINKED ISSUE #7624: Rubocop brakes on 2.7 pattern matching AST nodes while tree walking, as they children may be `nil`
Following code
```ruby
  case payload
  in key:
    method
  end
```
produces such AST (with `Parser::CurrentRuby.parse`, using `Parser::Ruby27` parser gem version: `2.7.0.1`):

```ruby
s(:case_match,
  s(:send, nil, :payload),
  s(:in_pattern,
    s(:hash_pattern,
      s(:match_var, :key)), nil, # <- here
    s(:send, nil, :method)), nil)
```

Second children of the `in_pattern` s-expression is `nil`, which leads to 
```
`undefined method `type' for nil:NilClass
```
error on line https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/ast/traversal.rb#L57

Simplest fix (https://github.com/maksar/rubocop/commit/db9f1cff43cdcbb25318bc13e0eb10ee2e3287e0) for me was adding `.compact`, but I'm not sure it is the right thing to do.

Issue is reproducible with current master (019fe5e3f0bdd0a533e51acc0c864330416cf08b)

koic:
Closes #7624 and #7774.

This PR supports pattern match for `Layout/ElseAlignment` cop.
Also this resolves the crash reported in #7624 and #7774 for using Ruby 2.7's pattern match.

-----------------

* [x] Added tests.


Rukomoynikov:
Thank you very much for you work.

PR #7794, opened 2020-03-11
[Fix #7778] Fix a false positive for `Layout/EndAlignment`

### LINKED ISSUE #7778: Layout/EndAlignment can't autocorrect condensed case statements
## Expected behavior

Given an oddly condensed case statement where the end is on the same line as the else RuboCop should be able to autocorrect the end onto its own line.

```ruby
        case other
        when IniFile then other.instance_variable_get(:@ini).keys
        when Hash then other.keys
        else raise "cannot merge contents from '#{other.class.name}'" end
```

## Actual behavior

The issue is identified, but doesn't autocorrect

```
Layout/EndAlignment: end at 201, 70 is not aligned with case at 198, 8.
        else raise "cannot merge contents from '#{other.class.name}'" end
                                                                      ^^^
```

Run RuboCop 0.80.1 against the code block above.

## RuboCop version

```
0.80.1
```

koic:
Resolves #7778.

This PR resolves a false positive for `Layout/EndAlignment` when a non-whitespace is used before the `end` keyword.

```console
% cat example.rb
if cond
else 'foo' end

% bundle exec rubocop -a --only Layout/EndAlignment
(snip)

Inspecting 1 file
W

Offenses:

example.rb:2:12: W: Layout/EndAlignment: end at 2, 11 is not aligned
with if at 1, 0.
else 'foo' end
           ^^^

```

This PR changes to not offense it.

I think the responsibility will be clear if different (maybe new) cop like `Layout/BlockEndNewline` cop handle the offense case.
So I think this role is preferably separated by `Layout/EndAlignment` cop.

-----------------

* [x] Added tests.


tas50:
Thanks @koic!

PR #7796, opened 2020-03-12
[Fix #7795] Make `Layout/EmptyLineAfterGuardClause` aware of `and return`

### LINKED ISSUE #7795: Enforce an empty line after render(:partail) and return
In the following line
`render(:partial) and return (if/unless condition)` it would be nice to have a cop to enforce an empty line, the same way we have with:
`return .... if`

koic:
Resolves #7795.

This PR makes `Layout/EmptyLineAfterGuardClause` aware of `and return`.

It was proposed based on the following code controlled by Rails controller.

```ruby
render :foo and return if condition

do_something
```

I think that it can be generalized as a control flow.

```ruby
foo(arg) and return if condition

do_something
```

So it will be newly detected by `Layout/EmptyLineAfterGuardClause` cop.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/ast/node.rb
+++ b/lib/rubocop/ast/node.rb
@@ -469,7 +469,13 @@ module RuboCop
         irange_type? || erange_type?
       end
 
-      def_node_matcher :guard_clause?, <<~PATTERN
+      def guard_clause?
+        node = and_type? || or_type? ? rhs : self
+
+        node.match_guard_clause?
+      end
+
+      def_node_matcher :match_guard_clause?, <<~PATTERN
         [${(send nil? {:raise :fail} ...) return break next} single_line?]
       PATTERN
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [#7793](https://github.com/rubocop-hq/rubocop/pull/7793): Prefer `include?` over `member?` in `Style/CollectionMethods`. ([@dmolesUC][])
 * [#7654](https://github.com/rubocop-hq/rubocop/issues/7654): Support `with_fixed_indentation` option for `Layout/ArrayAlignment` cop. ([@nikitasakov][])
 * [#7783](https://github.com/rubocop-hq/rubocop/pull/7783): Support Ruby 2.7's numbered parameter for `Style/RedundantSort`. ([@koic][])
+* [#7795](https://github.com/rubocop-hq/rubocop/issues/7795): Make `Layout/EmptyLineAfterGuardClause` aware of case where `and` or `or` is used before keyword that break control (e.g. `and return`). ([@koic][])
 
 ### Bug fixes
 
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/layout/empty_line_after_guard_clause.rb
```diff
@@ -79,7 +79,10 @@ def correct_style?(node)
         end
 
         def contains_guard_clause?(node)
-          node.if_branch&.guard_clause?
+          return unless (if_branch = node.if_branch)
+
+          if_branch.guard_clause? ||
+            if_branch.and_type? && if_branch.rhs.return_type?
```
Drenmi: Do you think this should be part of the definition of the `#guard_clause?` method?

lib/rubocop/ast/node.rb
```diff
@@ -624,6 +627,15 @@ def parent_module_name_for_block(ancestor)
         end
       end
 
+      def and_return?(node)
+        return unless node.and_type?
+
+        rhs = node.rhs
+
+        %i[return break next].include?(rhs.type) ||
+          rhs.send_type? && rhs.method?(:raise) || rhs.method?(:fail)
```
Drenmi: I think there might be an operator precedence problem here, but I'm not sure.

I assume you mean:

```ruby
rhs.send_type? && (rhs.method?(:raise) || rhs.method?(:fail))
```

but Ruby does this:

```ruby
(rhs.send_type? && (rhs.method?(:raise)) || rhs.method?(:fail))
```

lib/rubocop/cop/layout/empty_line_after_guard_clause.rb
koic: This suggestion looks good to me 😻I updated this PR that in addition to `return`, keywords that break control, such as `break`, `next`, `raise`, and `fail` can also be detected.

lib/rubocop/ast/node.rb
Drenmi: Would a node pattern like this work?

```
(and _ {return break next (send _ {:raise :fail})})
```

Drenmi: I was too slow! 😄

koic: Ah, I pushed the changes with a time lag 😅

koic:
@Drenmi Thanks for your kindness review!

bbatsov:
Shouldn't we also cover usages of `or` as well?

koic:
Sure! I updated this PR to support `or`.

bbatsov:
@koic Looks good! Feel free to merge it and the other PRs I just approved. I was a bit too lazy to resolve the merge conflicts in the changelog myself. :-)

koic:
Yeah, I will resolve conflicts and merge. Thank you for the review!

PR #7803, opened 2020-03-17
Add rake task to release to rubygems and github

ShockwaveNN:
I think it's never hurt to release this gem not only to rubygems.org, but also on [GitHub packages](https://github.com/features/packages)

I'm not sure of exact of release process (didn't find any info), but I think rake task `rake release` is used.
If so - I created new task `release_github_rubygems` which first call `rake release` and after push resulting gem to Github packages.

Please not, that person, who perform releases should execute 
```
echo ":github: Bearer GH_TOKEN" >> ~/.gem/credentials
```
before pushing gem. GH_TOKEN can be generated [here](https://github.com/settings/tokens) 

-----------------

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#7793](https://github.com/rubocop-hq/rubocop/pull/7793): Prefer `include?` over `member?` in `Style/CollectionMethods`. ([@dmolesUC][])
 * [#7654](https://github.com/rubocop-hq/rubocop/issues/7654): Support `with_fixed_indentation` option for `Layout/ArrayAlignment` cop. ([@nikitasakov][])
+* [#7803](https://github.com/rubocop-hq/rubocop/pull/7803): Add rake task `release_github_rubygems` to release gem on GitHub Packages and rubygems. ([@shockwavenn][])
 
 ### Bug fixes
 
@@ -4396,3 +4397,4 @@
 [@nikitasakov]: https://github.com/nikitasakov
 [@dmolesUC]: https://github.com/dmolesUC
 [@yuritomanek]: https://github.com/yuritomanek
+[@shockwavenn]: https://github.com/shockwavenn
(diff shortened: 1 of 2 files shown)


bbatsov:
I'd probably add a separate task to release to GitHub, plus some notes on using it, apart from the combined one. Then, again - I release using bundler, so I'm not sure I'll remember to run a rake task going forward. :D

ShockwaveNN:
> I'd probably add a separate task to release to GitHub, plus some notes on using it, apart from the combined one.

Agree, will do

> Then, again - I release using rake release, so I'm not sure I'll remember to run a new rake task going forward. :D

I think I can add releasing to github in default `rake release` task, but I don't think this is a good idea

koic:
Is this a required release method for RuboCop only? Otherwise, for example, you can propose a command to RubyGems.

And users can confirm a released gem impact by numbers such as the number of downloads on rubygem.org. I'm curious about the advantages of adding publishing using a non-central hosted GitHub package :-)

ShockwaveNN:
@koic Sorry, I'm not fully understand that you trying to say

> I'm curious about the advantages of adding publishing using a non-central hosted GitHub package

I don't know about any REAL advantages, I was just interested to try it in my projects and I thought this may be good for my favorite open-source project, but if rubocop dev team see no advantage in this - I think we can just cancel that idea

koic:
Thank you for replying. I'm also interested in the new GitHub Packages feature :-)

IMHO, For private internal hosting gems, using GitHub packages is convenient because hosting using `gem server` is not necessary. OTOH, for published available gems, it's better to release only to the centrally hosted rubygems.org. It would be simpler to focus on just one.

ShockwaveNN:
> It would be simpler to focus on just one.

So, to be clear - better close this PR?

bbatsov:
Yeah, let's close it. Thanks for the suggestion, though!

PR #7804, opened 2020-03-18
Nested methods doc fix

k3rni:
As-is, the code examples for the `Lint/NestedMethodDefinition` cop are wrong: both `class_eval` and `module_exec` are available on the class and module level respectively, and not on an instance. Only the eigenclass example, and, ironically, the "bad" example of just using `def` directly, actually work.


bbatsov:
Thanks!

PR #7808, opened 2020-03-21
[Fix #6962] Limit `Layout/ClassStructure` constant order offense to literal constants

### LINKED ISSUE #6962: Layout/ClassStructure bad autocorrection
I'm expecting RuboCop to always produce equivalent code when autocorrecting.

Layout/ClassStructure cop autocorrects badly

```ruby
class Foo
  def initialize; end

  local = 42
  EXAMPLE = local * 2
end

# gets autocorrected to:
class Foo
  EXAMPLE = local * 2
  def initialize; end

  local = 42
end
```

I believe that autocorrection should be disabled unless the constant is set to an independent literal (e.g. `CONSTANT = 42`). Even then, the way the code is currently moved around is wrong, like this HEREDOC case shows:

```ruby
class Foo
  def initialize; end

  TEXT = <<-EXAMPLE.freeze
  Hi!
  EXAMPLE
end

# autocorrected to:

class Foo
  TEXT = <<-EXAMPLE.freeze
  def initialize; end

  Hi!
  EXAMPLE
end
```

```
$ [bundle exec] rubocop -V
0.67.2 (using Parser 2.6.2.1, running on ruby 2.2.10 x86_64-darwin18)
```

tejasbubane:
Closes #6962

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/class_structure.rb
+++ b/lib/rubocop/cop/layout/class_structure.rb
@@ -147,6 +147,10 @@ module RuboCop
         MSG = '`%<category>s` is supposed to appear before ' \
               '`%<previous>s`.'
 
+        def_node_matcher :dynamic_constant?, <<~PATTERN
+          (casgn nil? _ (send ...))
+        PATTERN
+
         # Validates code style on class declaration.
         # Add offense when find a node out of expected order.
         def on_class(class_node)
@@ -168,11 +172,10 @@ module RuboCop
 
         # Autocorrect by swapping between two nodes autocorrecting them
         def autocorrect(corrector, node)
-          node_classification = classify(node)
           previous = node.left_siblings.find do |sibling|
-            classification = classify(sibling)
-            !ignore?(classification) && node_classification != classification
+            !ignore_for_autocorrect?(node, sibling)
           end
+          return unless previous
 
           current_range = source_range_with_comment(node)
           previous_range = source_range_with_comment(previous)
@@ -239,6 +242,15 @@ module RuboCop
           classification.nil? ||
             classification.to_s.end_with?('=') ||
             expected_order.index(classification).nil?
+        end
+
+        def ignore_for_autocorrect?(node, sibling)
+          classification = classify(node)
+          sibling_class = classify(sibling)
+
+          ignore?(sibling_class) ||
+            classification == sibling_class ||
+            dynamic_constant?(node)
         end
 
         def humanize_node(node)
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/layout/class_structure_spec.rb
```diff
@@ -204,6 +204,19 @@ def some_private_method
     it { expect_offense(code) }
   end
 
+  context 'constant is not a literal' do
+    specify do
+      expect_no_offenses <<~RUBY
+        class Person
+          def name; end
+
+          let foo = 5
```
bbatsov: let? :-)

lib/rubocop/cop/layout/class_structure.rb
```diff
@@ -234,10 +239,11 @@ def class_elements(class_node)
           end
         end
 
-        def ignore?(classification)
+        def ignore?(classification, node)
           classification.nil? ||
             classification.to_s.end_with?('=') ||
-            expected_order.index(classification).nil?
+            expected_order.index(classification).nil? ||
+            classification == 'constants' && dynamic_constant?(node)
```
koic: Is `classification == 'constants'` condition necessary? I think that this node type can be specified by `dynamic_constant?`.

```
def_node_matcher :dynamic_constant?, <<~PATTERN
  (casgn nil? _ (send ...))
   ^^^^^
```

tejasbubane: Right, `casgn` takes care of that.

bbatsov:
Btw, after thinking about this a bit more I'm not sure we won't to suppress the offense, as the problem we want to solve is only with the autocorrect, right? The auto-correct is not smart enough to see a constant refers to something else and changing the order messes things up, but still ideally someone has to move the constant and the related code where they belong.

lib/rubocop/cop/layout/class_structure.rb
```diff
@@ -146,6 +146,10 @@ class ClassStructure < Cop
           (send nil? { :private :protected :public })
         PATTERN
 
+        def_node_matcher :dynamic_constant?, <<~PATTERN
```
bbatsov, line 149: Probably it'd be better to just skip such nodes in the auto-correct instead of avoiding the offense for them.

tejasbubane:
@bbatsov Changed implementation to allow registering offense but no autocorrect.

tejasbubane:
@bbatsov This is ready for review.

bbatsov:

bbatsov (approved):

tejasbubane:
@bbatsov I have rebased this on top of `master`.

marcandre:
> @bbatsov I have rebased this branch on top of `master`.

Thanks!
But the Changelog entry is not at the right place. BTW, the new task `rake changelog:fix` avoids these conflicts and entries being misplaced

bbatsov:
I'll fix it manually, so it'd be part of today's release.

tejasbubane:
> I'll fix it manually, so it'd be part of today's release.

Thanks! :+1:

PR #7823, opened 2020-03-26
Support IgnoredMethod configuration in method complexity cops

Drenmi:
**Rationale:**

At work we make extensive use of functional objects to encapsulate business transactions. These objects will have a single public method `#call`, which stacks all the edge-cases as guard clauses towards the top.

It usually looks something like:

```ruby
def call
  # Edge cases.
  #
  return error(post.errors.full_messages.to_sentence) unless post.valid?
  return error("Author har been blocked from posting.") if author.blocked?
  return error("Author doesn't have an active subscription.") unless author.current_subscription.active?

  # Side effects.
  #
  create_post!
  send_notifications!

  # Success!
  #
  success(post: post)
end
```

We find this makes it very easy to understand what's going on, business transactions become composable and easy to test.

This almost always leads to `Metrics` cops violations for `#call`. However, the method is perfectly fine. With this layout, more complex business transactions aren't necessarily harder to grok than simpler ones.

Others seem to report the [same problem](https://stackoverflow.com/questions/59903382/rubocop-is-it-possible-to-add-exclusions-to-metrics-abcsize-cop).

**What is this change?**

This adds support for `IgnoredMethods` to `Metrics/AbcSize`, `Metrics/CyclomaticComplexity`, and `Metrics/PerceivedComplexity`.

**Side note:**

We should probably standardize the configuration option name before releasing 1.0. I used `IgnoredMethods` because we have an existing mixin for it, but I see other cops roll their own implementations and names.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * [#7739](https://github.com/rubocop-hq/rubocop/pull/7739): Add `IgnoreNotImplementedMethods` configuration to `Lint/UnusedMethodArgument`. ([@tejasbubane][])
 * [#7740](https://github.com/rubocop-hq/rubocop/issues/7740): Add `AllowModifiersOnSymbols` configuration to `Style/AccessModifierDeclarations`. ([@tejasbubane][])
 * [#7812](https://github.com/rubocop-hq/rubocop/pull/7812): Add auto-correction for `Lint/BooleanSymbol` cop. ([@tejasbubane][])
+* [#7823](https://github.com/rubocop-hq/rubocop/pull/7823): Add `IgnoredMethods` configuration in `Metrics/AbcSize`, `Metrics/CyclomaticComplexity`, and `Metrics/PerceivedComplexity` cops. ([@drenmi][])
 
 ### Bug fixes
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -1833,9 +1833,10 @@ Metrics/AbcSize:
     - https://en.wikipedia.org/wiki/ABC_Software_Metric
   Enabled: true
   VersionAdded: '0.27'
-  VersionChanged: '0.66'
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * [#7739](https://github.com/rubocop-hq/rubocop/pull/7739): Add `IgnoreNotImplementedMethods` configuration to `Lint/UnusedMethodArgument`. ([@tejasbubane][])
 * [#7740](https://github.com/rubocop-hq/rubocop/issues/7740): Add `AllowModifiersOnSymbols` configuration to `Style/AccessModifierDeclarations`. ([@tejasbubane][])
 * [#7812](https://github.com/rubocop-hq/rubocop/pull/7812): Add auto-correction for `Lint/BooleanSymbol` cop. ([@tejasbubane][])
+* Add `IgnoredMethods` configuration in `Metrics/AbcSize`, `Metrics/CyclomaticComplexity`, and `Metrics/PerceivedComplexity` cops. ([@drenmi][])
```
koic: This looks good to me 🌟It's a small feedback. When the link is added, it will be useful when referring to the changelog :-)

```suggestion
* [#7823](https://github.com/rubocop-hq/rubocop/pull/7823): Add `IgnoredMethods` configuration in `Metrics/AbcSize`, `Metrics/CyclomaticComplexity`, and `Metrics/PerceivedComplexity` cops. ([@drenmi][])
```

Drenmi: @koic Added! 🙇

bbatsov:
> We should probably standardize the configuration option name before releasing 1.0. I used IgnoredMethods because we have an existing mixin for it, but I see other cops roll their own implementations and names.

Agreed.

PR #7828, opened 2020-03-29
[Fix #7825] Fix crash for `Layout/MultilineMethodCallIndentation`

### LINKED ISSUE #7825: Layout/MultilineMethodCallIndentation error when parsing Dry::Monads::Try expression
Rubocop crashes when valid syntax according to https://dry-rb.org/gems/dry-monads/1.0/try/ is spread over two lines

--------

## Expected behavior

To generate a rubocop offence

## Actual behavior

```
An error occurred while Layout/MultilineMethodCallIndentation cop was inspecting ...
undefined method `dot' for #<Parser::Source::Map::Collection:0x00007fcdc8232448>
```

```
AllCops:
  TargetRubyVersion: 2.6.2
```

```
# frozen_string_literal: true

require 'dry/monads'

class ExceptionalLand
  include Dry::Monads[:try]

  def call
    Try[ZeroDivisionError] { 10 / 0 }
      .fmap { |x| x * 3 }
  end
end
```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.80.1 (using Parser 2.7.0.5, running on ruby 2.6.2 x86_64-darwin18)
```

tejasbubane:
When block passed to class constant.

Closes #7825 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/multiline_method_call_indentation.rb
+++ b/lib/rubocop/cop/layout/multiline_method_call_indentation.rb
@@ -193,7 +193,7 @@ module RuboCop
           node = node.receiver while node.receiver
           # ascend to first call which has a dot
           node = node.parent
-          node = node.parent until node.loc.dot
+          node = node.parent until node.loc.respond_to?(:dot) && node.loc.dot
 
           return if node.loc.dot.line != node.first_line
 
--- a/spec/rubocop/cop/layout/multiline_method_call_indentation_spec.rb
+++ b/spec/rubocop/cop/layout/multiline_method_call_indentation_spec.rb
@@ -304,6 +304,15 @@ RSpec.describe RuboCop::Cop::Layout::MultilineMethodCallIndentation do
         RUBY
       end
 
+      context 'target_ruby_version >= 2.5', :ruby25 do
+        it 'accepts key access to hash' do
+          expect_no_offenses(<<~RUBY)
+            hash[key] { 10 / 0 }
+              .fmap { |x| x * 3 }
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/layout/multiline_method_call_indentation.rb
```diff
@@ -193,7 +193,7 @@ def semantic_alignment_node(node)
           node = node.receiver while node.receiver
           # ascend to first call which has a dot
           node = node.parent
-          node = node.parent until node.loc.dot
+          node = node.parent until node.loc&.respond_to?(:dot) && node.loc&.dot
```
koic: Safe navigations looks unnecessary.

```suggestion
          node = node.parent until node.loc.respond_to?(:dot) && node.loc.dot
```

spec/rubocop/cop/layout/multiline_method_call_indentation_spec.rb
```diff
@@ -304,6 +304,15 @@ def foo
         RUBY
       end
 
+      context 'target_ruby_version >= 2.5', :ruby25 do
+        it 'accepts blocks passed to class constant' do
```
koic: This issue is with key access to hash rather than constant class. For example, the following can be reproduced.

```diff
-            Try[ZeroDivisionError] { 10 / 0 }
+            hash[key] { 10 / 0 }
               .fmap { |x| x * 3 }
```

Can you update the test description and changelog entry?

koic:
Thanks!

PR #7832, opened 2020-03-31
[Fix #7831] Fix a false positive for `Style/HashEachMethods`

### LINKED ISSUE #7831: False positive in Style/HashEachMethods
I have code like this:

```ruby
def validate_values
  # values is a Hash
  values.each do |key, value|
    next if value.is_a? String

    errors.add :values, :unsupported, key: key
  end
end
```
--------

## Expected behavior

No error is raised by Style/HashEachMethods.

## Actual behavior

```
 C: Style/HashEachMethods: Use each_value instead of values.each.
    values.each do |key, value|
    ^^^^^^^^^^^
```

Run RuboCop on the code snippet.

## RuboCop version

```
⟩ bundle exec rubocop -V
0.80.1 (using Parser 2.7.0.5, running on ruby 2.5.5 x86_64-darwin19)
```

koic:
Fixes #7831.

This PR fix a false positive for `Style/HashEachMethods` when there is no receiver for `keys` and `values`.

The following is an examples.

```ruby
values.each { |k, v| do_something(k, v) }
keys.each { |k, v| do_something(k, v) }
```

False negatives occur in cases such as object that inherit Hash, but that is perhaps a corner case.
I think it is more worthwhile to resolve the false positives with common naming of `values`.

-----------------

* [x] Added tests.


bbatsov:
Good change!

PR #7838, opened 2020-04-01
[Fix #6963] Check if iterating through Enumerator created with value …

### LINKED ISSUE #6963: Lint/Void cop false positive
It is the case that `#each` ignores the result of the block for `Array`, `Range` and `Enumerable` and thus creates a void context, `Enumerator#each` does not ignore the result of the block.

```
enum = [1,2,3].map
p enum.each { |_item| 42 } # => [42, 42, 42]
```

I believe that [this cop](https://github.com/rubocop-hq/rubocop/blob/v0.67.2/spec/rubocop/cop/lint/void_spec.rb#L183-L190) is thus overly strict, at least unless the receiver is an array/range literal or is named `array`, `ary`, `range`, etc.

tomurb:
…returning method

Fixes https://github.com/rubocop-hq/rubocop/issues/6963

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 * [#7860](https://github.com/rubocop-hq/rubocop/issues/7860): Change `AllowInHeredoc` option of `Layout/TrailingWhitespace` to `true` by default. ([@koic][])
 * [#7094](https://github.com/rubocop-hq/rubocop/issues/7094): Clarify alignment in `Layout/MultilineOperationIndentation`. ([@jonas054][])
 * [#7390](https://github.com/rubocop-hq/rubocop/issues/7390): **(Breaking)** Enabling a cop overrides disabling its department. ([@jonas054][])
+* [#6963](https://github.com/rubocop-hq/rubocop/issues/6963): Check if iterating through Enumerator created with value returning method. ([@tomurb][])
 
 ## 0.82.0 (2020-04-16)
 
@@ -4487,3 +4488,4 @@
 [@DracoAter]: https://github.com/DracoAter
 [@diogoosorio]: https://github.com/diogoosorio
 [@jeffcarbs]: https://github.com/jeffcarbs
+[@tomurb]: https://github.com/tomurb
--- a/lib/rubocop/ast/node.rb
+++ b/lib/rubocop/ast/node.rb
@@ -290,6 +290,18 @@ module RuboCop
 
       ## Destructuring
 
+      def_node_matcher :lvasgn_return, <<~PATTERN
+        `(lvasgn $_ (send _ $_))
+      PATTERN
+
+      def_node_matcher :enumerated_method, <<~PATTERN
+        `(send (send (send _ :array) $_) :each)
+      PATTERN
+
+      def_node_matcher :enumerated, <<~PATTERN
+        ^`(send (lvar $_) :each)
+      PATTERN
+
       def_node_matcher :receiver, <<~PATTERN
         {(send $_ ...) ({block numblock} (send $_ ...) ...)}
       PATTERN
--- a/lib/rubocop/cop/lint/void.rb
+++ b/lib/rubocop/cop/lint/void.rb
@@ -5,6 +5,10 @@ module RuboCop
     module Lint
       # This cop checks for operators, variables, literals, and nonmutating
       # methods used in void context.
+      # It correctly works for methods inherited from Enumerable. Does not work
+      # for others, except for `#defined?` method.
+      # This cop can yield some false positives in case of Enumerable-like
+      # methods.
       #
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 * [#7728](https://github.com/rubocop-hq/rubocop/issues/7728): Fix an error for `Style/OneLineConditional` when one of the branches contains a self keyword. ([@koic][])
 * [#7825](https://github.com/rubocop-hq/rubocop/issues/7825): Fix crash for `Layout/MultilineMethodCallIndentation` with key access to hash. ([@tejasbubane][])
 * [#7831](https://github.com/rubocop-hq/rubocop/issues/7831): Fix a false positive for `Style/HashEachMethods` when receiver is implicit. ([@koic][])
+* [#6963](https://github.com/rubocop-hq/rubocop/issues/6963): Check if iterating through Enumerator created with value returning method. ([@tomurb][])
```
bbatsov, line 41: That should go to the `master` section of the changelog.

lib/rubocop/ast/node.rb
```diff
@@ -356,6 +368,11 @@ def empty_source?
         {assignment? (send _recv :<< ...)}
       PATTERN
 
+      def enumerating?(method)
+        (enumerating_assigned?(method) ||
```
bbatsov, line 372: The other parens are redundant. I'm also thinking this would fit on a single line.

lib/rubocop/cop/lint/void.rb
```diff
@@ -143,6 +143,15 @@ def in_void_context?(node)
 
           return false unless parent && parent.children.last == node
 
+          value_returning_methods = %i[
```
bbatsov, line 146: Seems a bit weird to inline this array in the method. Probably it should be a constant instead.

bbatsov:
I think it'd also be a good idea to expand the cop's documentation, so it's clear what cases it can handle and what are its limitations.

bbatsov:
@tomurb I don't see any updates to the documentation.

bbatsov:
For clarity - I meant extending this part:

> This cop checks for operators, variables, literals, and nonmutating
 >  methods used in void context.

tomurb:
@bbatsov I'm not proud of it but I can't come up with anything better than
> This cop checks for operators, variables, literals, and nonmutating
methods (unless called on Enumerator instance created with mutating
method) used in void context.
Also when I think about it wasn't wrong. `#each` can't be nonmutating method if it mutates the object ;)

bbatsov:
Ah, I should explained myself better - I wanted to add some explanations about the limitations of the cops, potential false positives and to mention cases (like the enumerators) that it handles properly.

lib/rubocop/ast/node.rb
```diff
@@ -356,6 +368,10 @@ def empty_source?
         {assignment? (send _recv :<< ...)}
       PATTERN
 
+      def enumerating?(method)
```
bbatsov, line 371: @koic @Drenmi Do you think this is going to be useful in other cops or should we push it in the only cop using it?

koic: I don't know offhand 💦 

> push it in the only cop using it?

In that case, I like this choice because it's YAGNI :-)

lib/rubocop/cop/lint/void.rb
```diff
@@ -143,8 +165,23 @@ def in_void_context?(node)
...
+
+        def enumerating_chained?(parent, method)
+          method == parent.enumerated_method
+        end
+
+        def enumerating_assigned?(parent, method)
+          lvasgn_name, lvasgn_value = parent.parent&.lvasgn_return
+          lvasgn_value == method && !parent.enumerated.nil? &&
+            parent.enumerated == lvasgn_name
+        end
```
tomurb, line 184: I moved this two methods here from node.rb and removed #enumerating? 
https://github.com/rubocop-hq/rubocop/pull/7838#pullrequestreview-394320169
@bbatsov @koic

lib/rubocop/cop/lint/void.rb
```diff
@@ -5,6 +5,10 @@ module Cop
     module Lint
       # This cop checks for operators, variables, literals, and nonmutating
       # methods used in void context.
+      # It correctly works for methods inherited from Enumerable. Does not work
+      # for others, except for `#defined?` method.
+      # This cop can yield some false positives in case of Enumerable-like
+      # methods.
```
tomurb, line 11: @bbatsov Sorry it took so long but I rewrote this part too many times and I still don't feel that it's right

bbatsov:
Seem we totally forgot about this PR, given our focus on RuboCop 1.0. Sorry about that!

PR #7846, opened 2020-04-03
[Fix #7842] Add `AcceptImplicitNamespaces` to `Lint/RaiseException`

### LINKED ISSUE #7842: Lint/RaiseException not aware of namespaced "Exception" classes
Rubocop 0.81 introduced `Lint/RaiseException` which checks if we attempt to raise the top-level [`Exception`](https://ruby-doc.org/core-2.7.0/Exception.html) class.

Unfortunately, it also complains about subclasses with the name `Exception`, e.g.

```ruby
raise Gem::Exception, 'some gem error'
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# W: Lint/RaiseException: Use StandardError over Exception.
```

But the above is perfectly fine given that [`Gem::Exception`](https://ruby-doc.org/stdlib-2.7.0/libdoc/rubygems/rdoc/Gem/Exception.html) is a subclass of `RuntimeError`:

```ruby
Gem::Exception.ancestors
#=> [Gem::Exception, RuntimeError, StandardError, Exception, ...]
```

I don't think there's an easy fix because even a plain `raise Exception` could be valid from within the correct namespace, e.g.:

```ruby
module Gem
  def self.foo
    raise Exception
  end
end

Gem.foo #=> Gem::Exception (Gem::Exception)
```

The current behavior merely ensures that we don't use exceptions with the name `Exception` which seem to be of little use to me.

koic:
Fixes #7842.

This PR adds `AcceptImplicitNamespaces` option to `Lint/RaiseException`.

Users can specify and accept an implicit namespace as follows:

```yaml
Lint/RaiseException:
  AllowImplicitNamespaces:
    - 'Gem'
```

```ruby
# good
module Gem
  def self.foo
    raise Exception
  end
end

Gem.foo #=> Gem::Exception
```

`Gem` is specified by default.
https://ruby-doc.org/stdlib-2.7.0/libdoc/rubygems/rdoc/Gem/Exception.html

This way does not prevent all false positives, but it expect some cases can be solved.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1567,6 +1567,8 @@ Lint/RaiseException:
   StyleGuide: '#raise-exception'
   Enabled: pending
   VersionAdded: '0.81'
+  AllowedImplicitNamespaces:
+    - 'Gem'
 
 Lint/RandOne:
   Description: >-
--- a/lib/rubocop/cop/lint/raise_exception.rb
+++ b/lib/rubocop/cop/lint/raise_exception.rb
@@ -6,32 +6,68 @@ module RuboCop
       # This cop checks for `raise` or `fail` statements which are
       # raising `Exception` class.
       #
+      # You can specify a module name that will be an implicit namespace
+      # using `AllowedImplicitNamespaces` option. The cop cause a false positive
+      # for namespaced `Exception` when a namespace is omitted. This option can
+      # prevent the false positive by specifying a namespace to be omitted for
+      # `Exception`. Alternatively, make `Exception` a fully qualified class
+      # name with an explicit namespace.
+      #
       # @example
       #   # bad
       #   raise Exception, 'Error message here'
       #
       #   # good
       #   raise StandardError, 'Error message here'
+      #
+      # @example AllowedImplicitNamespaces: ['Gem']
+      #   # good
+      #   module Gem
+      #     def self.foo
+      #       raise Exception # This exception means `Gem::Exception`.
+      #     end
+      #   end
       class RaiseException < Cop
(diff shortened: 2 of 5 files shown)


config/default.yml
```diff
@@ -1567,6 +1567,8 @@ Lint/RaiseException:
   StyleGuide: '#raise-exception'
   Enabled: pending
   VersionAdded: '0.81'
+  AllowImplicitNamespaces:
```
bbatsov, line 1570: I think this should be named "AllowedImplicitNamespaces".

lib/rubocop/cop/lint/raise_exception.rb
```diff
@@ -5,33 +5,64 @@ module Cop
     module Lint
       # This cop checks for `raise` or `fail` statements which are
       # raising `Exception` class.
+      # You can specify a module name that will be an implicit namespace
```
bbatsov, line 8: Probably the explanation should be extended to mention the cop normally ignores namespaced "Exception", but might cause false positives if some module has its own `Exception` class. This would make it easier for people to understand the need for the extra config. That being said - I think it's a very bad idea to name base classes "Exception" and I'd rather name them "(BaseError)" or something along those lines. Hiding built-in constants is a recipe for trouble IMO.

koic: Thank you for reviewing. I extended the explanation.

bbatsov:
Looks good!

PR #7857, opened 2020-04-07
[Fix #7826] Add new `Layout/SpaceAroundMethodCall` cop

### LINKED ISSUE #7826: Cop idea: Check no spaces between `.` and the method being called

`A. new` or `Kernel.            puts` is valid ruby, but very weird.

I'd like a cop to enforce that no spaces are used between the `.` and the method being called. Not sure about naming, perhaps, `DotSpacing` inspired by the `DotPosition` cop, or maybe `SpaceAfterDot`?

None.

## Additional context

None.

saurabhmaurya15:
This MR adds a cop to for checking dot method calls to not have spaces around them.
It works for multiple method calls in a single line or in multiple lines.

closes https://github.com/rubocop-hq/rubocop/issues/7826

-----------------

* [x] Added tests.


lib/rubocop/cop/layout/space_around_method_call.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+# TODO: when finished, run `rake generate_cops_documentation` to update the docs
```
bbatsov, line 3: You should remove this line. :-)

bbatsov:
I think this cop should be a bit broader and check for all method invocation operators - @rrosenblum already mentioned safe navigation, but there's also `::`.  Probably the name should be something like "SpaceAroundMethodCallOperator" or something like this.

lib/rubocop/cop/layout/space_around_method_call.rb
```diff
@@ -0,0 +1,63 @@
...
+      #   foo . bar
+      #   foo. bar .buzz
+      #   foo
+      #     . bar
+      #     . buzz
+      #
+      #   # good
+      #   foo.bar
+      #   foo.bar.buzz
+      #   foo
```
bbatsov, line 22: You can also have chaining in a suffix form.

lib/rubocop/cop/layout/space_around_method_call.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+# TODO: when finished, run `rake generate_cops_documentation` to update the docs
+module RuboCop
+  module Cop
+    module Layout
+      # Checks dot method calls to not have spaces around them.
```
bbatsov, line 7: I'd use "method call operator" terminology here.

saurabhmaurya15: Makes sense. Changing it to `Checks method call operators to not have spaces around them.` 
Keeping it plural since it can have method chains also.

lib/rubocop/cop/layout/space_around_method_call.rb
saurabhmaurya15: Chaining in a suffix form is a bad style. There is already a cop for that so not including it in the example. 
I have added a spec for chaining in suffix form though.

saurabhmaurya15:
@bbatsov I have changed name to `SpaceAroundMethodCallOperator`. I'll add checks for safe navigation operator & `::` too.

config/default.yml
```diff
@@ -1080,6 +1080,11 @@ Layout/SpaceAroundKeyword:
   Enabled: true
   VersionAdded: '0.49'
 
+Layout/SpaceAroundMethodCallOperator:
+  Description: 'Checks method call operators to not have spaces around them.'
+  Enabled: true
```
bbatsov, line 1085: New cops should be "pending" by default.

bbatsov:
Okay, ping me when this is ready for a review again.

config/default.yml
saurabhmaurya15: Changed to pending & updated docs.

saurabhmaurya15:
@bbatsov This is ready for review.

bbatsov:
@saurabhmaurya15 I think you can improve the test suite a bit if you used shared examples there, as the tests for `.`, `::` and `.&` are mostly the same. Apart from this - it seems to me we're pretty close to finishing the cop.

bbatsov:
Thanks!

PR #7862, opened 2020-04-09
Add #wrap to corrector

marcandre:
Please expose `@source_rewriter`'s `wrap` method which can be particularly useful (equivalent to `insert_before` + `insert_after`)


bbatsov:
Nice addition! 🙇‍♂️

PR #7863, opened 2020-04-09
Have correctors accept node instead of range

marcandre:
This PR allows correctors to accept nodes in addition to ranges.

E.g: `corrector.insert_before(node, 'blah')`

Also provides better errors in case of wrong argument
Before: "undefined method `source_buffer' for 1..3:Range"
After: "Expected a Parser::Source::Range or Rubocop::AST::Node, got Range"

I don't think it would be particularly useful, but we could actually accept ranges like `1..3` if we wanted too...

This assumes #7862 is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#7863](https://github.com/rubocop-hq/rubocop/issues/7863): Corrector now accepts nodes in addition to ranges. ([@marcandre][])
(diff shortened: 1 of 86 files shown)


bbatsov:
Might be a good idea to also mention this in the docs.

marcandre:
> Might be a good idea to also mention this in the docs.

I modified slightly `manual/development.md` but couldn't find anywhere else...

I amended the PR with simplifications using this (mostly accepting node, a few using wrap, sometimes both at once...)

bbatsov:
Nicely done! 🙇‍♂️

PR #7867, opened 2020-04-11
Add support for tabs in indentation.

DracoAter:
As a stubborn tab user, I would like to have the ability to check, if all my code is indented with tabs, and also automatically correct indentation from spaces to tabs. So I:

Add support for indentation with tabs. Supports auto-correction.
The existing Layout/Tab is renamed into Layout/IndentationStyle. 

Layout/IndentationStyle now can be configured:
```yaml
# SupportedStyles: spaces, tabs
Layout/IndentationStyle:
  EnforcedStyle: spaces # default
```
`EnforcedStyle: spaces` will look for tabs in indentation and correct them into spaces.
`EnforcedStyle: tabs` will look for spaces in indentation and correct them into tabs.

Previous functionality of Layout/Tab that looked for tabs in places other than indentation is removed.
Hopefully this issue will be covered by TrailingWhitespace cop.


bbatsov:
The proposed change is fine by me. Just a couple of notes:

- I think "tabs" and "spaces" make for better style names
- We'll probably have to rename this cop, as its current name is a bit confusing after the addition of the configuration

DracoAter:
I have fixed the linting issues and changed SupportedStyles to plural (spaces, tabs).

Concerning the renaming, of the cop, that would be a major version update then. Probably it is possible to get away with minor version update, if we split it into 2 different cops. Existing Layout/Tab will search and fix the tabs outside of indentation, and a new cop (e.g. Layout/IndentationStyle will look for and fix tabs/spaces only in indentation.

If you need, I can implement any strategy you suggest.

bbatsov:
> Concerning the renaming, of the cop, that would be a major version update then.

We still haven't released 1.0, so that's not a big deal. I doubt the rename would impact many people, anyways, as it's not common to disable this cop. 

> Existing Layout/Tab will search and fix the tabs outside of indentation, and a new cop (e.g. Layout/IndentationStyle will look for and fix tabs/spaces only in indentation.

I do like this approach, though, as it's a bit weird that now the cop cares about both indentation and trailing tabs. I don't recall if it detects tabs in other contexts. I'm thinking that in essence the second cop can be something like "TrailingWhitespace" or something.

bbatsov:
Actually, we do have a trailing whitespace cop already, so unless there are other contexts in which a tab shouldn't appear I guess this check can be folded in the existing cop.

koic:
IMHO, I think the cop name can be left as `Layout/Tab` by using the option names below.

- Rename `spaces` to `soft_tabs`
- Rename `tabs` to `hard_tabs`

Of course it could be changed to a more suitable cop name or better approach.

DracoAter:
Currently it checks for tabs anywhere. Does not matter where the tab is (except inside a string), it is replaced with number of _indentation_width_ spaces, even in the middle of the line.

DracoAter:
So what is your final decision, what do I need to do now: rename, split or nothing?

bbatsov:
Let's go with the new name `IndentationStyle` or `UnitsOfIndentation` and drop the check for tabs that are not used for indentation purposes. If someone can come up with a better name - that'd be fine by me. 

I don't see much point in a cop checking for tabs at random place in the file. Hopefully the TrailingWhitespace cop already addresses trailing tabs.

bbatsov:
Don't forget to mention this breaking change in the changelog and add the cop to the list of renamed/removed cops.

config/default.yml
```diff
@@ -801,6 +801,22 @@ Layout/IndentationConsistency:
     # A reference to `EnforcedStyle: indented_internal_methods`.
     - https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#follow-the-coding-conventions
 
+Layout/IndentationStyle:
+  Description: 'Consistent indentation either with tabs only or spaces only.'
+  StyleGuide: '#spaces-indentation'
+  Enabled: true
+  VersionAdded: '0.49'
+  VersionChanged: '0.51'
```
bbatsov, line 809: You'll have to changed this to 0.82.

relnotes/v0.82.0.md
```diff
@@ -0,0 +1,9 @@
+### New features
+
+* [#7867](https://github.com/rubocop-hq/rubocop/issues/7867): Add support for tabs in indentation. ([@DracoAter][])
```
bbatsov, line 3: This file is auto-generated on release from the changelog. You shouldn't update it manually.

CHANGELOG.md
```diff
@@ -4,8 +4,13 @@
 
 ### New features
 
+* [#7867](https://github.com/rubocop-hq/rubocop/issues/7867): Add support for tabs in indentation. ([@DracoAter][])
 * [#7850](https://github.com/rubocop-hq/rubocop/issues/7850): Make it possible to enable/disable pending cops. ([@koic][])
 
+### Changes
+
+* Renamed `Layout/Tab` cop to `Layout/IndentationStyle` ([@DracoAter][])
```
bbatsov, line 18: Please add a `**(Breaking)**` prefix here.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#7867](https://github.com/rubocop-hq/rubocop/issues/7867): Add support for tabs in indentation. ([@DracoAter][])
```
koic, line 7:
```suggestion
* [#7867](https://github.com/rubocop-hq/rubocop/pull/7867): Add support for tabs in indentation. ([@DracoAter][])
```

bbatsov:
Thanks for working on this! I might not be fan of tabs, but I'm happy we'll be more accommodating to the people who are! :-)

DracoAter:
Thank you for the collaboration and quick replies.

DracoAter:
I screwed up url in changelog, could you fix it?

```
[@DracoAter]: https:/github.com/DracoAter
```
Add one more slash before github.

bbatsov:
Sure.

PR #7868, opened 2020-04-12
Improving auto correction's API

marcandre:
I am writing my first cop and I feel the current `autocorrect` API could be improved.

This PR is just a demo and not meant for merging. A few cops were refactored using the proposed API.

### When to specify corrections

My main observation is that the very best moment to do an autocorrection is when the offense is detected. Currently the autocorrection and the offense detection are (officially) decoupled. When an offense is detected, we know exactly what it is (location, type, etc.) yet we are unable to make the correction there and then.

When we are later asked to make the autocorrection, we are passed the offending "node". All the state that we might have created to analyze the offense has to be rebuilt at that point, and if we detect different style of offenses, we even have to [deduct again which offense this is](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/layout/space_around_block_parameters.rb#L44-L54) and correct accordingly.

### Current workarounds

Some cops [simply redo](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/internal_affairs/node_type_predicate.rb#L24-L32) the [detection work twice](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/style/redundant_exception.rb#L24-L44)

Some cops resort to "cheats" for the current state, like [AccessModifierIndentation](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/layout/access_modifier_indentation.rb#L51-L53) temporarily storing `@column_delta`. This works only provided that each `autocorrect` calls are made after each `add_offense`. Also, a naive approach wouldn't work reliably:

```ruby
def auto_correct(node)
  lambda do |corrector|
    corrector.insert_after(node, ' ' * @column_delta) # Would be the last value for all corrections
  end
end
```

### Overly complex de-dupping

The processing currently has two de-duping being done.

Firstly, extra offenses on the same range are ignored. I can see how this could be useful for certain cops.

Secondly, if offenses are on different ranges but the same node, the autocorrector is called only for the first one. The only builtin cop that uses this one is `SpaceInsideReferenceBrackets`: it will add one or two offenses on the opening `'['` and/or closing `']'`, and autocorrect will correct the node (detecting again which of the opening or closing brackets need to be corrected) and relies on the dedupping not to double-correct if both the opening and closing brackets were wrong. Clearly, this is an example where this de-dupping isn't needed but instead the correction should be specifiable when detected...

That dedupping is the only reason I can see why the API of `add_offense` officially asks for a `Node` and optionally a `location`. There's no good reason why an offenses would be attached to a node rather than a range. Indeed a few pass a range instead (like [EmptyLineAfterMagicComment](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/layout/empty_line_after_magic_comment.rb#L34-L36)) and cheat by repeating `location: <same location>`.

I propose that the newer API simply asks for a range (or node/comment and uses its `loc.expression`) and dedupes only on the range.

### Alternative to returning lambdas / false

Finally I personally dislike returning lambdas. Ruby makes it trivial to capture blocks which feels to me a much more natural way to proceed.

It also wasn't clear at first that `autocorrect` can also return `false` when no correction can be made. This complicates processing a bit... Before you want to autocorrect, you have to check that you *can* autocorrect.
Moreover, there is also redundancy and it permits a cop writer to return a lambda that nevertheless wouldn't make any correction.

This trouble should be avoided and deduced automatically. Technically the issue is that we would need a way to build the corrections, check if there are any, and only later merge them.

In this "demo" I use a hack. I simply run the block twice, first on a fake corrector that simply records if it is called and the usual processing will call it later on Parser's `TreeRewriter`. The plan would be not to use a hack but I would instead add the merging functionality into `TreeRewriter` (I am its author btw). Before I spend time doing it though, I'd like to see if an agreement can reached on the `add_offense` API.

### Proposed API - simple case

I would propose an improved (and fully compatible) API. 

```ruby
# Before:
def on_something(node)
  # ...
  add_offense(node)
end

def auto_correct(node)
  lambda do |corrector|
    corrector.wrap(node, '(', ')')
  end
end

# After:
include Autocorrector

def on_something(node)
  # ...
    corrector.wrap(node, '(', ')')
  end
end
```

#### Post processing

One complication is that `add_offense` already yields (unless the offense is disabled) for some rare cases where post processing is needed. I noticed that 5% of the builtin cops use this post processing vs 45% that do autocorrections; most with post processing also have autocorrection. There are many solutions to this, but the simple case is simply to use the same block:

```ruby
# Before:
def on_something(node)
  # ...
  add_offense(node) do
    # ... do only if not disabled
  end
end

def auto_correct(node)
  return false unless can_correct_this_one?(node)
  lambda do |corrector|
    corrector.wrap(node, '(', ')')
    # ...
  end
end

# After:
include Autocorrector

def on_something(node)
   # ...
    # ... do only if not disabled

    if can_correct_this_one?(node)
      corrector.wrap(node, '(', ')')
      # ...
    end
end
```

### Compatiblity

What I'm proposing should be compatible with existing cops.

The new API is opt-in for cops that decide to `include Autocorrector`. I'm proposing that the `location: ` key isn't supported in the new API (pass the range directly instead of the node in those case).

It's not clear to what extent compatibility is required, in particular with `Cop#corrections` array. I wrote a `CorrectionsProxy` so that `corrections` could still be used. Actually the rest of the code still uses it for now (`Team`, ...), this seemed like a good way to test compatiblity. I intend on cleaning things up though so that the core doesn't use `Cop#corrections`...


marcandre:
The new `parser` has the required changes to the `TreeRewriter` (couldn't resist implementing them). I'll update this PR but the API proposal remains the same.

bbatsov:
Sounds good to me! Sorry for the radio silence, but I didn't find the time to take a careful look at your proposal. Looking forward to the improved version!

marcandre:
Updated the code and tweaked the API. I removed the syntax sugar part as it conflicts with the strict rules of RuboCop (no multiple block chaining). I believe it's not actually necessary and both the corrections and the code-to-execute-only-if-enabled can be in the same block. The revised API is thus simpler.

I never spelled out the downside of my proposal: it's potentially slower. For example, if RuboCop is run without the `--autocorrect` flag, the block is still call so the calculations for the autocorrections will still be run. My guess is that the time spent to calculated the corrections should typically be negligible compared to parsing and finding the offenses, so I don't think it will make a noticeable difference.

bbatsov:
I'll try to find some time soon to review your proposal more carefully. In the mean time it'd be great if you ran some benchmarks regarding the performance impact.

marcandre:
> benchmarks regarding the performance impact.

Did you have anything in mind?

I probably should not have mentioned the theoretical performance impact, because this PR affects only the correction part, which is typically negligible to the inspection part.

I think the performance impact will be undetectable:
case
when no offenses? then no speed difference
when many offenses for cops that can't autocorrect? then no difference
when many offenses for autocorrection cops and running with auto-correction? then no speed difference
when many offenses for autocorrection cops and running with auto-correction and cops don't spend an inordinate amount of time figuring out what the correction would be? then no speed difference
else
maybe a slowdown
end

I ran the test suite before and after the changes and got similar speeds (1m12 before vs 1m7 after, no I don't think it's noticeably faster either)

marcandre:
Question remaining to be answered:

Is compatiblity with `Cop#corrections` array required? It's not clear to me why a cop would fiddle with that. Still, I wrote a `CorrectionsProxy` that should be mostly compatible

marcandre:
Actually 1.0 is coming up real soon, which is great timing. I propose to include my `CorrectionsProxy` solution with deprecation warnings, in case any code out there uses the array directly.

I will try to finalize this PR (although the proposed API will remain unchanged) and think it would make a great addition to the 1.0 release.

bbatsov:
> Is compatiblity with Cop#corrections array required? It's not clear to me why a cop would fiddle with that. Still, I wrote a CorrectionsProxy that should be mostly compatible

I don't think so. I can't even remember why we introduced this. Maybe @koic or @jonas054 do. 

I went over your proposal more carefully and I like it quite a lot overall. I guess we'll gradually be shifting the auto-corrections to the new API as times goes on. The problem with big codebases is that after some point in time it gets really hard to keep the uniform - e.g. we never migrated all the code to the NodePattern, after it was introduced, we never adopted everywhere all the node extensions and so on.

marcandre:
Well, the `corrections` array was needed to hold the corrections because they might not be applied afterall (cops can exclude other cops). In an ideal world though it would have been part of a `CopRunner` or of the `Team` class, not of the `Cop` class that is meant to be extended.

It's going to be fun de refactor the existing cops to use the new API 😄; I'll focus first on cleaning up the `Team`.

marcandre:
I've made some progress on this. What I've done is modified `cop/cop` to reflect the proposed API, so anyone looking at that file for documentation would see the "modern" API / implementation. The compatibility with the legacy behavior is secluded in two `legacy/..._support.rb` files.

I have not yet changed the `Team/Commissionner` implementation, nor most of the specs, as a check that the behavior for the internal legacy API is correctly handled (e.g. `Cop#corrections`).

When I'm satisfied with this, the last step would be to have those use the modern API and add warning to the internal legacy API (e.g. `Cop#corrections`, but not the `add_offense` with the legacy API).

There's basically one spec that fails right now that has to do with error handling. Currently, if an error occurs when a Cop analyses a file, it is handled differently than if an error occurs when a Cop autocorrects it. In the former case, it logs the error or reraises it depending on the `:raise` option. In the latter it always raises. Compare the specs for the [analysis part](https://github.com/rubocop-hq/rubocop/blob/master/spec/rubocop/cop/team_spec.rb#L163-L169) and for the [autocorrection part](https://github.com/rubocop-hq/rubocop/blob/master/spec/rubocop/cop/team_spec.rb#L196-L201). Note that both tests say "records Team#errors" but that's not what the second spec actually tests! BTW, were are not talking about `ClobberringError`s here, but any other `StandardError`.

Does anyone know the reason for this difference in behavior? Is there a potential issue in having the same behavior (i.e. raise only if the `:raise` setting is on)?

bbatsov:
Likely that's just some oversight. 

> There's basically one spec that fails right now that has to do with error handling. Currently, if an error occurs when a Cop analyses a file, it is handled differently than if an error occurs when a Cop autocorrects it. In the former case, it logs the error or reraises it depending on the :raise option. In the latter it always raises.

Probably we can simplify things, as I don't see much reason for having different approaches. 

We'll also have to document all of this in the development docs, which are pretty light on such details right now. That's all we have on implementing auto-correction https://docs.rubocop.org/en/latest/development/#auto-correct I guess everyone just learns the ropes from other cops. :-)

marcandre:
> Probably we can simplify things, as I don't see much reason for having different approaches.

Great, thanks for the answer.

> We'll also have to document all of this in the development docs, which are pretty light on such details right now. That's all we have on implementing auto-correction https://docs.rubocop.org/en/latest/development/#auto-correct

Right. I'll update and improve the docs.

> I guess everyone just learns the ropes from other cops. :-)

Or trying things out to see if it passes or not, like [this cop passing `[[location], location]` instead of a Node!](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/lint/redundant_cop_disable_directive.rb#L188)

kittiphophum18:
> I am writing my first cop and I feel the current `autocorrect` API could be improved.
> 
> This PR is just a demo and not meant for merging. A few cops were refactored using the proposed API.
> 
> ### When to specify corrections
> My main observation is that the very best moment to do an autocorrection is when the offense is detected. Currently the autocorrection and the offense detection are (officially) decoupled. When an offense is detected, we know exactly what it is (location, type, etc.) yet we are unable to make the correction there and then.
> 
> When we are later asked to make the autocorrection, we are passed the offending "node". All the state that we might have created to analyze the offense has to be rebuilt at that point, and if we detect different style of offenses, we even have to [deduct again which offense this is](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/layout/space_around_block_parameters.rb#L44-L54) and correct accordingly.
> 
> ### Current workarounds
> Some cops [simply redo](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/internal_affairs/node_type_predicate.rb#L24-L32) the [detection work twice](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/style/redundant_exception.rb#L24-L44)
> 
> Some cops resort to "cheats" for the current state, like [AccessModifierIndentation](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/layout/access_modifier_indentation.rb#L51-L53) temporarily storing `@column_delta`. This works only provided that each `autocorrect` calls are made after each `add_offense`. Also, a naive approach wouldn't work reliably:
> 
> ```ruby
> def auto_correct(node)
>   lambda do |corrector|
>     corrector.insert_after(node, ' ' * @column_delta) # Would be the last value for all corrections
>   end
> end
> ```
> 
> ### Overly complex de-dupping
> The processing currently has two de-duping being done.
> 
> Firstly, extra offenses on the same range are ignored. I can see how this could be useful for certain cops.
> 
> Secondly, if offenses are on different ranges but the same node, the autocorrector is called only for the first one. The only builtin cop that uses this one is `SpaceInsideReferenceBrackets`: it will add one or two offenses on the opening `'['` and/or closing `']'`, and autocorrect will correct the node (detecting again which of the opening or closing brackets need to be corrected) and relies on the dedupping not to double-correct if both the opening and closing brackets were wrong. Clearly, this is an example where this de-dupping isn't needed but instead the correction should be specifiable when detected...
> 
> That dedupping is the only reason I can see why the API of `add_offense` officially asks for a `Node` and optionally a `location`. There's no good reason why an offenses would be attached to a node rather than a range. Indeed a few pass a range instead (like [EmptyLineAfterMagicComment](https://github.com/rubocop-hq/rubocop/blob/9c3f904e77b3c3786eaface4fdb545764136beca/lib/rubocop/cop/layout/empty_line_after_magic_comment.rb#L34-L36)) and cheat by repeating `location: <same location>`.
> 
> I propose that the newer API simply asks for a range (or node/comment and uses its `loc.expression`) and dedupes only on the range.
> 
> ### Alternative to returning lambdas / false
> Finally I personally dislike returning lambdas. Ruby makes it trivial to capture blocks which feels to me a much more natural way to proceed.
> 
> It also wasn't clear at first that `autocorrect` can also return `false` when no correction can be made. This complicates processing a bit... Before you want to autocorrect, you have to check that you _can_ autocorrect.
> Moreover, there is also redundancy and it permits a cop writer to return a lambda that nevertheless wouldn't make any correction.
> 
> This trouble should be avoided and deduced automatically. Technically the issue is that we would need a way to build the corrections, check if there are any, and only later merge them.
> 
> In this "demo" I use a hack. I simply run the block twice, first on a fake corrector that simply records if it is called and the usual processing will call it later on Parser's `TreeRewriter`. The plan would be not to use a hack but I would instead add the merging functionality into `TreeRewriter` (I am its author btw). Before I spend time doing it though, I'd like to see if an agreement can reached on the `add_offense` API.
> 
> ### Proposed API - simple case
> I would propose an improved (and fully compatible) API.
> 
> ```ruby
> # Before:
> def on_something(node)
>   # ...
>   add_offense(node)
> end
> 
> def auto_correct(node)
>   lambda do |corrector|
>     corrector.wrap(node, '(', ')')
>   end
> end
> 
> # After:
> include Autocorrector
> 
> def on_something(node)
>   # ...
>   add_offense(node) do |corrector|
>     corrector.wrap(node, '(', ')')
>   end
> end
> ```
> 
> #### Post processing
> One complication is that `add_offense` already yields (unless the offense is disabled) for some rare cases where post processing is needed. I noticed that 5% of the builtin cops use this post processing vs 45% that do autocorrections; most with post processing also have autocorrection. There are many solutions to this, but the simple case is simply to use the same block:
> 
> ```ruby
> # Before:
> def on_something(node)
>   # ...
>   add_offense(node) do
>     # ... do only if not disabled
>   end
> end
> 
> def auto_correct(node)
>   return false unless can_correct_this_one?(node)
>   lambda do |corrector|
>     corrector.wrap(node, '(', ')')
>     # ...
>   end
> end
> 
> # After:
> include Autocorrector
> 
> def on_something(node)
>    # ...
>   add_offense(node) do |corrector|
>     # ... do only if not disabled
> 
>     if can_correct_this_one?(node)
>       corrector.wrap(node, '(', ')')
>       # ...
>     end
> end
> ```
> 
> ### Compatiblity
> What I'm proposing should be compatible with existing cops.
> 
> The new API is opt-in for cops that decide to `include Autocorrector`. I'm proposing that the `location: ` key isn't supported in the new API (pass the range directly instead of the node in those case).
> 
> It's not clear to what extent compatibility is required, in particular with `Cop#corrections` array. I wrote a `CorrectionsProxy` so that `corrections` could still be used. Actually the rest of the code still uses it for now (`Team`, ...), this seemed like a good way to test compatiblity. I intend on cleaning things up though so that the core doesn't use `Cop#corrections`...

marcandre:
@kittiphophum18 your comment appears to be a quote of my entire message (I removed it for brevity)

marcandre:
I'm very happy about the progress I've made on this PR 😀

I wrote a [Changelog](https://github.com/rubocop-hq/rubocop/blob/be9c618cd7951f233cac443482edc7e27e81ddc8/manual/cop_api_v1_changelog.md) for the changes so far.

I also updated the `development.md` to reflect the new API and improved the example.

There are a few things I still want to do:
* improve the error processing (#8016)
* better isolate what is transient (e.g. `processed_source`, probably better renamed `current_processed_source` or `investigation.processed_source`) vs what doesn't change (e.g. `config`)
* see how easy it would be for the new cops to have a clear lifecycle that would work across files (see #7968)
* final cleanup

marcandre:
I've clarified the callback sequence for cops with methods `begin_investigation` and `complete_investigation`, refactored things a bit more.

I'd still want to simplify the special processing for the exceptional cases like `Lint/Syntax`. I'm thinking a callback `on_other_file` being called instead of the walk callbacks for such cases as non Ruby files or unparseable Ruby files.

Note: I intend to squash many of the commits but am waiting when it's completely ready. I probably should make a separate PR for some changes that are not directly related [like this one](https://github.com/rubocop-hq/rubocop/pull/7868/commits/fc4ab1ae24ee42aeeef8fdd56b1d1d68eb41ec4e)

marcandre:
Refactored `Lint::Syntax`, now a "real" Cop, and pretty close of the same with `RedundantDisable` although the spec files need heavy refactoring. I'm impressed how lazy I am compared to many devs.

@bbatsov I need to convert the `manual/my_file.md` of my branch to asciidoc, what's the best way to do that?

bbatsov:
I'm using https://github.com/asciidoctor/kramdown-asciidoc

marcandre:
All green 🎉 I refactored the two cops I wanted, moved the docs.

I need to:
* rebase and fix conflicts with the `loop` setting
* check compatibility with `pronto-rubocop` and `erblint` and probably ease the upgrade path a bit; `pronto-rubocop` at least uses `Corrector` directly.
* cleanup the commits

marcandre:
Rebased, all green, registry-related refactor extracted into #8095.

Just need the easier compatibility (`Corrector.new` to call `Legacy::Corrector.new` probably, not sure if there's anything else) and a review and it's going to be ready to merge 😮

marcandre:
I checked out both `pronto-rubocop` and `erb-lint`. Very interesting.

They both try to draw a correspondence between offenses and corrections. The bad news is that there's no way to make that work seamlessly after the refactor. The good news is that I'll make a minor adjustment to `Offense` so that it stores the associated correction (if any), and that association will be trivial.

`erb_lint` does some funky thing to offset the ranges afterwards (to affect just the ruby code within the 'erb' template I imagine). I have to think a bit about if there's an easy for them to do this after the refactor.

We're getting real close. Anyone following this, is it ok to merge #8095?

docs/modules/ROOT/pages/development.adoc
```diff
@@ -267,33 +267,46 @@ end
...
+https://www.rubydoc.info/gems/parser/Parser/Source/TreeRewriter[parser's TreeRewriter]
+to which you can give instructions about what to do with the
 offensive node.
 
 Let's start with a simple spec to cover it:
 
 [source,ruby]
 ----
-it 'autocorrect `!a.empty?` to `a.any?` ' do
+  it 'corrects `!a.empty?`' do
```
bbatsov, line 280: I see this code block and the ones that follow are indented by some margin (probably because you copied them straight from some source file.

docs/modules/ROOT/pages/development.adoc
```diff
@@ -267,33 +267,46 @@ end
...
 
 [source,ruby]
 ----
-def autocorrect(node)
-  lambda do |corrector|
-    internal_expression = node.children[0].children[0].source
-    corrector.replace(node, "#{internal_expression}.any?")
-  end
-end
+        extend Autocorrector
```
bbatsov, line 297: e.g. this one is quite indented

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+2) Locate your calls to `add_offense` and make sure that you pass as the first argument either a `AST::Node`, a `::Parser::Source::Comment` or a `::Parser::Source::Range`, and no `location:` named parameter
+
+[discrete]
+==== Example:
+
+[source,ruby]
+----
+# Before
+    class MySillyCop < Cop
```
bbatsov, line 20: More code snippets with leading whitespace.

bbatsov:
I'm keeping an eye on the PR. My main concern is that right now it's not very clear what's so different between `Cop` and `Base` and that the new class doesn't have YARD docs, which seem important to me for key APIs. 

I'll try to take a closer look at the PR during the day. Hopefully other people from @rubocop-hq/rubocop-core will do the same, as once we're done with this it's time for 1.0. :-)

bbatsov:
(well, we'll probably do 0.86 as well just to be on the safe side and then cut 1.0 without any material changes to it)

docs/modules/ROOT/pages/development.adoc
marcandre: Fixed

marcandre:
> new class doesn't have YARD docs

I'm not sure what you mean by YARD docs. Do you mean method documentation? The new class was more documented than the old class, and now I've added more documentation.

I refactored some more `Team` & `Commissioner` and how they handle reports. Error handling is fixed; no errors can escape silently now.

I also made it so that `Team`s can reuse their cops. Old cops will be re-created, new cops will not, at least that's what I'm proposing. Being kept around could be opt-in. I added added a `cache` helper. This should solve #7968 

I believe I am done with all the refactoring I wanted to do 😅 

Tomorrow I want to see what an updated `erb_lint` would look like.

marcandre:
Oh, forgot I still need to check for two or three cops/mixins that need their state invalidated too

marcandre:
Changed the default to not persist, as I wrote in #7968. Renamed the callbacks to `on_new_investigation` and `on_investigation_end`. A few tweaks. I'm putting this PR as "ready for review" 🎉.

I checked `erb_lint` in more detail. It's quite impossible to maintain seamless compatibility as they fiddle with many internals, but the new API make it easier [as soon as this PR](https://github.com/whitequark/parser/pull/703) lands.

@bbatsov releasing a `0.86` with this PR would break things, for example `erb_lint`. Their gemspec specifies `~>0.75` or somesuch, so maybe releasing `1.0.0rc0` would be better when it's time?

bbatsov:
> @bbatsov releasing a 0.86 with this PR would break things, for example erb_lint. Their gemspec specifies ~>0.75 or somesuch, so maybe releasing 1.0.0rc0 would be better when it's time?

That's pretty bad, as a normal release would have given us more feedback. Perhaps we can just ping them on GitHub to update the dep restriction?

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
```diff
@@ -0,0 +1,258 @@
...
+
+=== Other error handling
+
+_Legacy:_ Clobbering errors are silently ignored. Calling `insert_before` with ranges that extend beyond the source code was silently fixed.
+
+_Current:_ Such errors are not ignored. It is still ok that a given Cop's corrections clobber another Cop's, but any given Cop should not issue corrections that clobber each other, or with invalid ranges, otherwise these will be listed in the processing errors.
+
+=== `#corrections`
+
+_Legacy:_ Corrections were held in `#corrections` an array of lambdas. A proxy was written to maintain compatibility with `+cop.corrections << ...+`, `+cop.corrections.concat ...+`, etc.
```
bquorning, line 215:
```suggestion
_Legacy:_ Corrections were held in `#corrections` as an array of lambdas. A proxy was written to maintain compatibility with `+cop.corrections << ...+`, `+cop.corrections.concat ...+`, etc.
```

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
```diff
@@ -0,0 +1,258 @@
...
+
+=== Joining forces
+
+_Legacy:_ `join_force?(force_class)` was called with every force class
+
+_Current:_ `self.joining_forces` is now used to return the force (or an array of forces) to join.
+
+== Cop persistence
+
+Cops can now be persisted between files. By default new cop instances are created for each source. See `support_multiple_source?` documentation
```
bquorning, line 233:
```suggestion
Cops can now be persisted between files. By default new cop instances are created for each source. See `support_multiple_source?` documentation.
```

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
```diff
@@ -0,0 +1,258 @@
...
+
+_Current:_ derives from `parser`'s `TreeRewriter`. No second argument to `initialize`; not needed as correctors can be merged.
+
+=== `Commissioner` & `Team`
+
+Refactored for better separation of concern, being reusable, better result reporting and better error handling.
+
+== Misc API changes
+
+* internal API clarified for Commissioner. It calls `begin_investigation` and received the results in `complete_investigation`.
```
bquorning, line 249:
```suggestion
* internal API clarified for Commissioner. It calls `begin_investigation` and receives the results in `complete_investigation`.
```

bquorning:
> I thought it was possible to use the old syntax and the new syntax in parallel for a while. Maybe I’m just missing a compabitility flag somewhere?

SORRY. I had changed the base class from `Cop` to `Base`, which caused the incompatibility. If I revert that and run the spec suite, I only get one error:
```
RuntimeError: Correction target buffer 70198717836060 name:"spec/smoke_tests/weird_rspec_spec.rb" is not current 70198718266940 name:"spec/smoke_tests/weird_rspec_spec.rb" under investigation>
```
which I need to investigate a bit further.

marcandre:
> > I thought it was possible to use the old syntax and the new syntax in parallel for a while. Maybe I’m just missing a compabitility flag somewhere?
> 
> SORRY. I had changed the base class from `Cop` to `Base`, which caused the incompatibility.

Exactly :-)

> If I revert that and run the spec suite, I only get one error:
> 
> ```
> RuntimeError: Correction target buffer 
>   70198717836060 name:"spec/smoke_tests/weird_rspec_spec.rb" is not current
>   70198718266940 name:"spec/smoke_tests/weird_rspec_spec.rb" under investigation>
> ```
> 
> which I need to investigate a bit further.
> 
> Update: I found the bug – memoization needs to be reset on `on_new_investigation` (or `investigate` in older cops).

Good. Note that clearing your memoization should not be required with cops (unless they `support_multiple_source?`, but it is likely that your test itself was calling the cop with different sources. I imagine the test wasn't using `expect_offense`, right?

docs/modules/ROOT/pages/development.adoc
```diff
@@ -267,33 +267,46 @@ end
...
+  it 'corrects `!a.empty?`' do
+    expect_offense(<<~RUBY)
+      !array.empty?
+      ^^^^^^^^^^^^^ Use `.any?` and remove the negation part.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      array.any?
+    RUBY
   expect(autocorrect_source('!a.empty?')).to eq('a.any?')
```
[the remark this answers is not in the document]
marcandre: Good catch

lib/rubocop/rspec/shared_contexts.rb
```diff
@@ -45,7 +45,7 @@
   let(:source) { 'code = {some: :ruby}' }
 
   let(:cop_class) do
-    if described_class.is_a?(Class) && described_class < RuboCop::Cop::Cop
+    if described_class.is_a?(Class) && described_class < RuboCop::Cop::Base
       described_class
     else
       RuboCop::Cop::Cop
```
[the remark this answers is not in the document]
marcandre: Good point. I added that `if` for some formatters' spec, but it is cheating. I tweaked it to raise an error if `described_class` isn't a `Cop::Base`, this way it is mandatory to specify if one is using against `Cop::Cop` or `Cop::Base`. I kept `Cop::Cop` for the 3 formatters using it currently coz I'm lazy 😅

lib/rubocop/cop/commissioner.rb
```diff
@@ -7,6 +7,35 @@ module Cop
     class Commissioner
       include RuboCop::AST::Traversal
 
+      # How a Commissioner returns the results of the investigation
+      # as a list of Cop::InvestigationReport and any errors caught
+      # during the investigation.
+      # Immutable
+      # Consider creation API private
```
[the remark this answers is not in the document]
marcandre: TBH, the only reason why I wrote this is I'd like to `keyword_argument: true` for our Structs, but that's Ruby 2.5+ only. I'd be really tempted to use my `backport` for this...

Nothing is truly private in Ruby, and for `Cop::Base` in particular since we expect people to write Cops. I tried moving some of the code into `Investigation` but gave up for now.

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
```diff
@@ -0,0 +1,258 @@
...
+end
+----
+
+=== Other API changes
+
+If you cop uses `investigate`, `investigate_post_walk`, `join_force?`, or internal classes like `Corrector`, `Commissioner`, `Team`, these have changed. See the Detailed API Changes.
+
+=== Upgrading specs
+
+It is highly recommended you use `expect_offense` / `expect_no_offense` in your specs, e.g.:
```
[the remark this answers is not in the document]
marcandre: Yes, good catch.

marcandre:
> I have given up on reviewing all the changes

Right, the diff is super noisy. It's probably easier to review the classes `Team`, `Commissioner` and `Base` in their entirety (at least the public API) if you're up for that.

bquorning:
I am running into issues with cops defining a `#message` method, which I believe would be called back after calling `add_offense` – e.g. https://github.com/rubocop-hq/rubocop-rspec/blob/8eb550e2f4a5dced6b47900b63c551695b912e73/lib/rubocop/cop/rspec/invalid_predicate_matcher.rb#L38-L40

`#message` is still being called, but with a different argument than before. In this case, I am seeing an error like 
```
     NoMethodError:
       undefined method `method_name' for #<Parser::Source::Range example_spec.rb 21...34>
```

One workaround is to avoid the callback by passing the `:message` keyword argument to `#add_offense`. If there are better ways to do it, perhaps they should be added to the cop_api_v1_changelog document?

marcandre:
> I am running into issues with cops defining a `#message` method, which I believe would be called back after calling `add_offense` – e.g. https://github.com/rubocop-hq/rubocop-rspec/blob/8eb550e2f4a5dced6b47900b63c551695b912e73/lib/rubocop/cop/rspec/invalid_predicate_matcher.rb#L38-L40
> 
> `#message` is still being called, but with a different argument than before. In this case, I am seeing an error like
> 
> ```
>      NoMethodError:
>        undefined method `method_name' for #<Parser::Source::Range example_spec.rb 21...34>
> ```
> 
> One workaround is to avoid the callback by passing the `:message` keyword argument to `#add_offense`. If there are better ways to do it, perhaps they should be added to the cop_api_v1_changelog document?

I must admit I feel that passing `message: ` is clearer, but old style `message` *is* supported with the right argument (see method `callback_argument`). I imagine that you are now calling `add_offense` with a location (which will be passed to `message`) instead of `range + location: ...` as before...

bquorning:
I am almost done changing rubocop-rspec to use this new style. For the most part, it’s been quite easy (but tedious) work. It is very easy to exceed the limits of Metrics/MethodLength and/or Metrics/AbcSize when the autocorrection code has to be put into a block, and for the most part I will probably just call out to the `#autocorrect` method, but passing the `corrector` instance along – and making the method private.

Another problem I ran into is that one cop in rubocop-rspec is using `ConfigurableNaming` which includes `ConfigurableFormatting` which is still calling `add_offense` and passing the `:location` kwarg: https://github.com/rubocop-hq/rubocop/blob/3b0552b9989db97104cfe0fc30515e9b6ae549f2/lib/rubocop/cop/mixin/configurable_formatting.rb#L13 – I am not really sure how to work around that one.

marcandre:
> Another problem I ran into is that one cop in rubocop-rspec is using `ConfigurableNaming` which includes `ConfigurableFormatting` which is still calling `add_offense` and passing the `:location` kwarg:

What happens if we change the mixin to call `add_offense(name_range, message: ...)`?

That mixin doesn't do autocorrections, so I don't think there would be most impact, but it raises a good point I hadn't considered: mixins that would want to be do corrections and be compatible with both APIs. Do we have such mixins?

bquorning:
@marcandre You can see the fruit of my work here, if you’re interested: https://github.com/rubocop-hq/rubocop-rspec/compare/master...bquorning:new-rubocop-autocorrect

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
```diff
@@ -0,0 +1,258 @@
...
+      add_offense(node, location: :selector, message: "Be positive")
+    end
+  end
+end
+
+# After
+class MySillyCop < Base
+  def on_send(node)
+    if node.method_name == :-
+      add_offense(node.loc.selector, message: "Be positive")
```
baweaver, line 32: Note for another section, but what if we had a `node.selector`?

docs/modules/ROOT/pages/development.adoc
```diff
@@ -267,33 +267,45 @@ end
 === Auto-correct
 
 The auto-correct can help humans automatically fix offenses that have been detected.
-It's necessary to define an `autocorrect` method that returns a lambda
-https://github.com/whitequark/parser/blob/master/lib/parser/rewriter.rb[rewriter]
-with the corrector where you can give instructions about what to do with the
+It's necessary to `extend Autocorrector`.
+The method `add_offense` yields a corrector object that is a thin wrapper on
```
baweaver, line 271: Nit: It might be nice to link to the corrector class here

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
marcandre: Not too sure what you mean. If you pass `node`, it's a shortcut for `node.loc.expression` (the range of the whole node). In the example, `selector` is the `:-`. Check `ruby-parse -L -e "a - b"`

baweaver:
So here's an interesting idea, what if we top-leveled common correctors?:

```ruby
include Autocorrector

def on_something(node)
  return unless we_want_to_replace?

  # Replacement
  replace(node, new_code)

  # Wrapping
  wrap(node) { |n| "wrappy_method(#{n}") }

  # Inserting before or after
  insert_before(node, '#')
end
```

Most of these are common enough usecases and the abstractions already exist in this PR to make most of them possible. If users are attempting to auto-correct this could make them a first-class citizen, and by only having them in the `Autocorrector` you'd know that they exist.

Implementation may be something like:

```ruby
# Autocorrector - naive implementations

def replace(node, new_node)
    corrector.replace(node, new_node)
  end  
end

def wrap(node, &fn)
  replace(node, yield new_node)
end

def insert_before(node, content)
  wrap(node) { |n| "#{content}#{n}" }
end
```

baweaver:
That could also allow for the creation of extremely succinct simple wrappers if you made another abstraction on top of this:

```ruby
Corrector.on(NODE_PATTERN).replace { |n, _captures| "wrappy(#{n})" }
```

...but that's a topic for another ticket.

Really liking these changes in general for what potential they unlock in making things like this possible.

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
baweaver: Aha, my mistake, disregard then

marcandre:
> That could also allow for the creation of extremely succinct simple wrappers if you made another abstraction on top of this:
> 
> ```ruby
> Corrector.on(NODE_PATTERN).replace { |n, _captures| "wrappy(#{n})" }
> ```
> 
> ...but that's a topic for another ticket.

I would like to tweak the `NodePattern` so it can generate `on_send` or whichever hooks is required, but it's not possible in general. Still, `replace` the whole thing is not the best way to go.

marcandre:
> So here's an interesting idea, what if we top-leveled common correctors?:

Oops, looks like I edited your comment instead of replying. Restored now.

Thanks for the idea, but I feel it confuses the concept of an offense and its correction (which could consist of multiple rewriting actions). There are important other arguments to `add_offense`, you'd have to add them to all the methods (there's more than replace, wrap, ...). Also wrapping and replacing code are quite different actions; the idea is to avoid replacing whole text so that there's no/less conflict between cops acting on the same code.

jonas054:
I like the direction of these changes, they seem to lead to cleaner code in the cops. But when I run auto-correction on RuboCop's own sources with this branch, it crashes.

```
marcandre__autocorrect ~/dev/rubocop$ ./exe/rubocop -a --force-default-config
Inspecting 1072 files
..C....C....CC...CCCCC.C.....C.CCCC........CC...C..C..C.C.........C...C..C.CC.C..CC...C.........CCC..C....CCC..C...C...C....C...C.....C.CCCC.....C.CC..CC.C....C.......C...C.......C..CCCC...C.CC.C......C....CC.CC.....C..C...CC.C.....C.CCCCC..C....C.....CCC..C...C..C..CC...C......C.......C..C........C..CCCW........C...C........C..C..C....C......CCC......CC..........C.C..........WC....C.....CCC.CCC.C.C....C.....C...C.C...C....C..........C......C....C.........C...C....C..C.CCCW

Offenses:

[...]

477 files inspected, 230 offenses detected, 40 offenses corrected
Infinite loop detected in /home/jonas/dev/rubocop/lib/rubocop/cop/style/redundant_self.rb.
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:289:in `check_for_infinite_loop'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:272:in `block in iterate_until_no_changes'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:271:in `loop'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:271:in `iterate_until_no_changes'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:242:in `do_inspection_loop'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:122:in `block in file_offenses'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:147:in `file_offense_cache'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:120:in `file_offenses'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:111:in `process_file'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:90:in `block in each_inspected_file'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:89:in `each'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:89:in `reduce'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:89:in `each_inspected_file'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:78:in `inspect_files'
/home/jonas/dev/rubocop/lib/rubocop/runner.rb:39:in `run'
/home/jonas/dev/rubocop/lib/rubocop/cli/command/execute_runner.rb:21:in `execute_runner'
/home/jonas/dev/rubocop/lib/rubocop/cli/command/execute_runner.rb:13:in `run'
/home/jonas/dev/rubocop/lib/rubocop/cli/command.rb:10:in `run'
/home/jonas/dev/rubocop/lib/rubocop/cli/environment.rb:17:in `run'
/home/jonas/dev/rubocop/lib/rubocop/cli.rb:65:in `run_command'
/home/jonas/dev/rubocop/lib/rubocop/cli.rb:72:in `execute_runners'
/home/jonas/dev/rubocop/lib/rubocop/cli.rb:41:in `run'
./exe/rubocop:13:in `block in <main>'
/usr/local/rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/benchmark.rb:308:in `realtime'
./exe/rubocop:12:in `<main>'

**Comment by marcandre** [comment]
> I like the direction of these changes, they seem to lead to cleaner code in the cops. But when I run auto-correction on RuboCop's own sources with this branch, it crashes.

Good catch!

I was over-enthusiastic about new correctors detecting if they were empty, and there's a [sloppy cop](https://github.com/rubocop-hq/rubocop/issues/8151) that unwitting does `corrector.replace(node, node.source)` which makes the corrector not empty (that correction could clobber with another correction, for example) but leaves the source unchanged. I've restored the previous full-source check and added a spec for this case.

File: lib/rubocop/cop/mixin/autocorrector.rb [hunk 13]
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # extend this module to signal autocorrection support
+    module Autocorrector
```
bbatsov, line 6: Btw, let's name this `AutoCorrector`, as it looks weird to me without the CamelCase. :-)

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
```diff
@@ -0,0 +1,258 @@
+= Cop API v0 to v1 API changes
```
bbatsov, line 1: Probably we can name this upgrade notes and put here everything major for the update, not just the Cop API changes.

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
```diff
@@ -0,0 +1,258 @@
...
+  def on_new_investigation
+    @data = nil
+    super  # Be nice and call super for callback
+  end
+end
+----
+
+=== Other API changes
+
+If your cop uses `investigate`, `investigate_post_walk`, `join_force?`, or internal classes like `Corrector`, `Commissioner`, `Team`, these have changed. See the Detailed API Changes.
```
bbatsov, line 110: You can add an anchor to "Detailed API Changes" here.

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
```diff
@@ -0,0 +1,258 @@
...
+    expect_no_offenses(<<~RUBY)
+      "don't".despair
+    RUBY
+  end
+end
+----
+
+In the unlikely case where you use the class `RuboCop::Cop::Corrector` directly, it has changed a bit but you can ease your transition with `RuboCop::Cop::Legacy::Corrector` that is meant to be somewhat backwards compatible. You will need to `require 'rubocop/cop/legacy/corrector'`.
+
+= Detailed API Changes
```
bbatsov, line 143: You can't have two top-level headings, so you either have to increase the nesting on split the document in two.

bbatsov:
@marcandre It seems to me everyone's out of feedback, so we can move forward with merging this. I've left a few small final remarks.

lib/rubocop/cop/mixin/autocorrector.rb
marcandre: Sure, I'm fine with it either way. Note that `auto-correct` is used for CLI options and rake task, while most uses in the sources use `autocorrect`.

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
marcandre: Absolutely.

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
marcandre: I indented one level, but anyone can feel free to split it if they feel it's better.

docs/modules/ROOT/pages/cop_api_v1_changelog.adoc
marcandre: Renamed it 👍

marcandre:
Merged! 🎉
Thanks for everyone's comments and review on this big PR 😸

tas50:
I'm very excited for some of these changes for use in Cookstyle. Thanks for all the works folks.

PR #7673, opened 2020-01-26
[Fix #7564] Drop dependency on jaro_winkler gem

buehmann:
Ruby 2.3 and later ship with the `did_you_mean` gem, which includes a
JaroWinkler implementation.

So there is no need to depend on the `jaro_winkler` gem that people have
had trouble installing lately.

Fixes #7564 and fixes #7447.

Issue #7564, opened 2019-12-12
Installation Error on MacOS


zzj0402:
RuboCop doesn't install.
--------

## Expected behavior

RuboCop installs

## Actual behavior

Console outputs:
```
ERROR:  Error installing rubocop:
	ERROR: Failed to build gem native extension.

    current directory: /Library/Ruby/Gems/2.3.0/gems/jaro_winkler-1.5.4/ext/jaro_winkler
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/bin/ruby -r ./siteconf20191212-97928-1pj6l4t.rb extconf.rb
mkmf.rb can't find header files for ruby at /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/include/ruby.h

extconf failed, exit code 1
```

## Steps to reproduce the problem

New macOS version 10.14.5, run in terminal:
`sudo gem install rubocop`

## RuboCop version

Include the output of `rubocop -V` or `bundle exec rubocop -V` if using Bundler. Here's an example:

```
$ [bundle exec] rubocop -V
0.77.0 (using Parser 2.5.1.2, running on ruby 2.5.1 x86_64-linux)
```

koic:
It seems that jaro_winkler gem installation failed. Can you open the issue to jaro_winkler?
https://github.com/tonytonyjan/jaro_winkler/issues

Thank you for the feedback.

mojavelinux:
Having a dependency on a native gem that basically requires the full suite of development tools makes this gem tremendously difficult for a lot of people and CI environments to install. Is it absolutely necessary to depend on this gem? Could it be an optional dependency with a graceful fallback?

buehmann:
@mojavelinux Thanks. Indeed we could do without the `jaro_winkler` gem, I guess. Opened a PR: https://github.com/rubocop-hq/rubocop/pull/7673

mojavelinux:
Thanks @buehmann!

Issue #7447, opened 2019-10-18
Installation fails on Ubuntu 18.04 running Ruby 2.6.3


sjayanna:
The `jaro_winkler` gem is also failing to install and so I am not sure if the issue is with that gem or with Rubocop but wanted to report it. Rubocop installs just fine if I revert back to Ruby 2.5.3.

```>> gem install rubocop
Building native extensions. This could take a while...
ERROR:  Error installing rubocop:
	ERROR: Failed to build gem native extension.

    current directory: /home/sjayanna/.rvm/gems/ruby-2.6.3/gems/jaro_winkler-1.5.3/ext/jaro_winkler
/home/sjayanna/.rvm/rubies/ruby-2.6.3/bin/ruby -I /home/sjayanna/.rvm/rubies/ruby-2.6.3/lib/ruby/site_ruby/2.6.0 -r ./siteconf20191018-14011-1xlq2d.rb extconf.rb
creating Makefile

current directory: /home/sjayanna/.rvm/gems/ruby-2.6.3/gems/jaro_winkler-1.5.3/ext/jaro_winkler
make "DESTDIR=" clean

current directory: /home/sjayanna/.rvm/gems/ruby-2.6.3/gems/jaro_winkler-1.5.3/ext/jaro_winkler
make "DESTDIR="
compiling adj_matrix.c
gcc: error: unrecognized command line option '-Wduplicated-cond'
gcc: error: unrecognized command line option '-Wmisleading-indentation'
gcc: error: unrecognized command line option '-Wrestrict'
gcc: error: unrecognized command line option '-Wimplicit-fallthrough=0'
Makefile:243: recipe for target 'adj_matrix.o' failed
make: *** [adj_matrix.o] Error 1

make failed, exit code 2

Gem files will remain installed in /home/sjayanna/.rvm/gems/ruby-2.6.3/gems/jaro_winkler-1.5.3 for inspection.
Results logged to /home/sjayanna/.rvm/gems/ruby-2.6.3/extensions/x86_64-linux/2.6.0/jaro_winkler-1.5.3/gem_make.out```

koic:
It seems that jaro_winkler gem installation failed. Could you please open the issue to jaro_winkler?
https://github.com/tonytonyjan/jaro_winkler/issues

/Cc @tonytonyjan

sjayanna:
@koic Done. https://github.com/tonytonyjan/jaro_winkler/issues/31

PR #7869, opened 2020-04-12
Drop support for Ruby 2.3

koic:
### Summary

This PR drops support for Ruby 2.3.

It is discussed in #6945.
And this suggestion would mean that RuboCop 1.0 (and pre-release) requires Ruby 2.4 or higher.

The following is a plan after this PR.
There was a build error report for a dependent jaro_winkler gem.
#5989, #6754, #7447, and #7564. And #7673 was trying to solve it.

After dropping Ruby 2.3, replace jaro_winkler with did_you_mean.
did_you_mean is written in Ruby, so the build error due to native extensions no longer occur. That change opens as anther PR.

### Other Information

The latest did_you_mean (1.4.0) supports Ruby 2.5 or higher, so it needs to be confirmed old versions it works with Ruby 2.4.

-----------------

* [x] Added tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,29 +44,6 @@ rubocop_steps: &rubocop_steps
 
 jobs:
 
-  # Ruby 2.3
-  ruby-2.3-spec:
-    docker:
-      - image: circleci/ruby:2.3
-    environment:
-      <<: *common_env
-    steps:
-      *spec_steps
-  ruby-2.3-ascii_spec:
-    docker:
-      - image: circleci/ruby:2.3
-    environment:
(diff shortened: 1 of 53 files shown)


bbatsov:
Great work! Thanks!

> After dropping Ruby 2.3, replace jaro_winkler with did_you_mean.
did_you_mean is written in Ruby, so the build error due to native extensions no longer occur. That change opens as anther PR.

I love the plan!

DirtyF:
Question: what is Rubocop policy against current maintained Ruby versions? Given 2.4 is EOL now, when will Rubocop drop 2.4 support?

PR #7872, opened 2020-04-13
Avoid infinite loop when autocorrecting with Style/IfUnlessModifier

alexdowad:
`Style/IfUnlessModifier` can autocorrect long lines which use an if/unless modifier to
`if...end` form. It can also autocorrect `if...end` form to modifier form if the body
is only a single line.

Obviously, there is the potential for an infinite autocorrect loop here. The cop tried
to avoid that by checking whether the `if...end` conditional would overflow the maximum
length of a line if converted to modifier form.

However, the check is based only on the size of the conditional code itself and not
*other* statements which may follow after `end`, like this:

    if condition
      do_this
    end; other_statement; yet_another

This (hideous) code could result from an autocorrect of something like this:

    do_this if condition; other_statement; yet_another

So the cop still can and does cause infinite loops. A simple solution is to avoid
flagging long lines which use if/unless modifiers *if* there are other statements
following on the same line. Of course, such long lines are still a problem; but we can
allow `Layout/LineLength` to flag them rather than `Style/IfUnlessModifier`.


bbatsov:
The build is currently failing.

alexdowad:
Sorry about that.

bbatsov:
Thanks!

PR #7873, opened 2020-04-13
Update docs for TrailingCommaInArguments cop

mtamhankar1:
The existing documentation for the `no_comma` option does not clearly indicate that the cop requires that there is no comma present. It only states that a comma is not required.

Also, there is a grammatical error in the existing documentation.


bbatsov:
Thanks!

PR #7883, opened 2020-04-17
[Fix #7860] Change `AllowInHeredoc` option of `Layout/TrailingWhitespace` to `true`

### LINKED ISSUE #7860: Auto correct should not automatically remove trailing whitespace in heredoc
Back in 2017 https://github.com/rubocop-hq/rubocop/issues/4517 raised the issue that trailing whitespaces should not be marked as offences in heredoc strings.

This resulted in https://github.com/rubocop-hq/rubocop/pull/5777 which added a new config option to prevent the cop from complaining.

Whilst this seems like a good fix, it still means that when running auto correct, it will remove trailing spaces automatically, which could result in a change in logic.

Take the below for example:

```
require 'digest'

my_string = <<~END
Whitespace here: 
END

sha = Digest::SHA2.new(256).hexdigest my_string
expected_sha = "74788f79010cd7a165e6df2e91013ce72903696bca4d42989de97c3150b89db2"

puts "authenticated" if expected_sha == sha

```

This will print "authenticated" before running `rubocop --auto-correct`, but will fail after.

I've had a similar issue in production where we had to verify a signed string and to test I used an example provided string, which had trailing whitespace. Running `rubocop --auto-correct` broke the tests.

--------

## Expected behavior

Running `rubocop --auto-correct` should not remove the trailing whitespace.

## Actual behavior

The whitespace is removed

See the script above

## RuboCop version

0.81.0

koic:
Fixes #7860.

This PR changes `AllowInHeredoc` option of `Layout/TrailingWhitespace` to `true` by default. The default behavior will be more safe.

-----------------

* [ ] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1295,8 +1295,8 @@ Layout/TrailingWhitespace:
   StyleGuide: '#no-trailing-whitespace'
   Enabled: true
   VersionAdded: '0.49'
-  VersionChanged: '0.55'
-  AllowInHeredoc: false
+  VersionChanged: '0.83'
+  AllowInHeredoc: true
 
 #################### Lint ##################################
 ### Warnings
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * [#7882](https://github.com/rubocop-hq/rubocop/pull/7882): Fix `Style/CaseEquality` when `AllowOnConstant` is `true` and the method receiver is implicit. ([@rafaelfranca][])
 
+### Changes
+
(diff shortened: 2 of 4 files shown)


config/default.yml
```diff
@@ -1295,8 +1295,8 @@ Layout/TrailingWhitespace:
   StyleGuide: '#no-trailing-whitespace'
   Enabled: true
   VersionAdded: '0.49'
-  VersionChanged: '0.55'
-  AllowInHeredoc: false
+  VersionChanged: '0.83'
```
koic, line 1298: This value may be `1.0`. I will change it when the next version is decided.

bbatsov:
Are changes to the cop's documentation needed as well?

koic:
Ah, I was forgetting it. I updated it!

bbatsov:
Thanks!

PR #7887, opened 2020-04-17
Ignore --parallel when passed with incompatible option

trevorrjohn:
Allow setting of parallel and combination with incompatible options. Will output a warning if combined but still execute without the `--parallel` option enabled.

https://github.com/rubocop-hq/rubocop/issues/7544

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/options.rb
+++ b/lib/rubocop/options.rb
@@ -320,9 +320,9 @@ module RuboCop
       return unless @options.key?(:parallel)
 
       if @options[:cache] == 'false'
-        raise OptionArgumentError, '-P/--parallel uses caching to speed up ' \
-                                   'execution, so combining with --cache ' \
-                                   'false is not allowed.'
+        @options.delete(:parallel)
+        warn '-P/--parallel uses caching to speed up execution, so combining '\
+             'with --cache false will ignore --parallel.'
       end
 
       validate_parallel_with_combo_option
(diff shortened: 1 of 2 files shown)


trevorrjohn:
This is a WIP, I just wanted some feedback if this is a route we could go.

jonas054:
I'm not sure. Is there a situation where changing the arguments given to `rubocop` would be hard for the user, so that automatically removing incompatible options is of real value?

trevorrjohn:
So the way I would use it, would be have my config default to run parallel and if I provided an incompatible option like `auto_fix` it could just ignore the parallel option. That being said maybe this needs to be another argument, something like `--ignore-incompatible` which would be more explicit for the user.

jonas054:
I remain unconvinced. :slightly_smiling_face: The `--ignore-incompatible` option doesn't make a huge difference. Ignoring certain options because they are incompatible with others is a bit weird IMO - not what you would expect from any tool. Reporting the error and exiting keeps the complexity down.

trevorrjohn:
I do agree it is a bit strange. But I do think for a tool usability is one of the most important features and right now it feels very rigid. Maybe it isn't rubocops job to manage this complexity though.

jonas054:
I'm closing the PR, hoping that's OK.

PR #7888, opened 2020-04-18
[Fix #7790] Set options in ConfigLoader earlier

### LINKED ISSUE #7790: --ignore-parent-exclusion does not work if --parallel is used
The switch `--ignore-parent-exclusion` seems to be ignored if `--parallel` is also present.

--------

## Expected behavior

The following commands should always inspect the same number of files:

```
rubocop --ignore-parent-exclusion
rubocop --ignore-parent-exclusion --parallel
```

## Actual behavior

Adding `--parallel` seems to disable the effect of `--ignore-parent-exclusion`.

```
.
├── .rubocop.yml
└── subdir
    ├── .rubocop.yml
    └── test.rb
```

Content of `.rubocop.yml`:
```
AllCops:
  Exclude:
    - "subdir/*"
```

Content of `subdir/.rubocop.yml`:
```
AllCops:
  Exclude:
    - "foobar"
```

This is the output of running Rubocop with different switches, from inside the `subdir` directory:

* `rubocop`: "0 file inspected"
* `rubocop --parallel`: "0 file inspected"
* `rubocop --ignore-parent-exclusion`: "1 file inspected"
* `rubocop --ignore-parent-exclusion --parallel`: "0 files inspected"

**NOTE:** to be fair, I'm not even sure why I need to use the `--ignore-parent-exclusion` flag, as I'm not specifying `inherit` in the config file inside `subdir`, so I would expect Rubocop to use that file and ignore the config file in the parent directory.

## RuboCop version

```
$ rubocop -V
0.80.1 (using Parser 2.7.0.2, running on ruby 2.5.1 x86_64-darwin18)
```

jonas054:
Transferring options to the `ConfigLoader` must be done before we validate the `--parallel` flag against configuration. Since we cache the configuration when reading it to perform validation, the cached configuration will not take all the given options into account, which leads to different behavior when running the actual inspection if the `--parallel` flag is given.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 * [#7882](https://github.com/rubocop-hq/rubocop/pull/7882): Fix `Style/CaseEquality` when `AllowOnConstant` is `true` and the method receiver is implicit. ([@rafaelfranca][])
+* [#7790](https://github.com/rubocop-hq/rubocop/issues/7790): Fix `--parallel` and `--ignore-parent-exclusion` combination. ([@jonas054][])
 
 ### Changes
 
--- a/lib/rubocop/cli.rb
+++ b/lib/rubocop/cli.rb
@@ -35,6 +35,7 @@ module RuboCop
       if @options[:init]
         run_command(:init)
       else
+        set_options_to_config_loader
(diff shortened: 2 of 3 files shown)


koic:
Thank you!

PR #7890, opened 2020-04-19
[Fix #7881] Call act_on_options earlier in CLI#run()

### LINKED ISSUE #7881: --parallel disables --force-default-config
It seems like the `--parallel`-switch breaks using `--force-default-config` when `inherited_from` uses a broken / unavailable URL. 

This is probably related to #7790

--------

## Expected behavior

In a project with a local `.rubocop.yml`, running 
```
rubocop --parallel --force-default-config .
```
should ignore the config given in `.rubocop.yml`

## Actual behavior

```
rubocop --parallel --force-default-config .
```
still uses the local `.rubocop.yml`. The same applies to using `--config /dev/null` instead of `--force-default-config` 

- Create a project with a local `.rubocop.yml` config
- Make sure the config uses `inherit_from` with an unresolvable URL. 
- Compare running rubocop with `--force-default-config` and `--force-default-config --parallel`

Alternatively use the example project at https://github.com/defsprite/rubocop-bug

## RuboCop version

```
🌀 rubocop -V
0.82.0 (using Parser 2.7.0.2, running on ruby 2.6.5 x86_64-darwin19)
```

jonas054:
This fixes another problem that's similar to the one reported in #7790. The problem is still that configuration is read and cached when we validate options vs configuration. By acting on all options first, we make sure that the first time we read configuration, it's done according to the given options. Specifically, it's the `--force-default-config` option that's now being respected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#7882](https://github.com/rubocop-hq/rubocop/pull/7882): Fix `Style/CaseEquality` when `AllowOnConstant` is `true` and the method receiver is implicit. ([@rafaelfranca][])
 * [#7790](https://github.com/rubocop-hq/rubocop/issues/7790): Fix `--parallel` and `--ignore-parent-exclusion` combination. ([@jonas054][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Nicely done!

PR #7896, opened 2020-04-20
Fix semantic versioning link

hugopeixoto:
On https://docs.rubocop.org/en/latest/versioning/, the first sentence under Release Policy reads:

> We're following http://semver.org/[SemVer] (as much as one can be following it when the major version is 0).

This PR changes the semver link to a markdown link. The other line changes are to keep the paragraph aligned to textwidth=72.

### DIFF
--- a/manual/versioning.md
+++ b/manual/versioning.md
@@ -9,12 +9,13 @@ configuration.
 
 ## Release Policy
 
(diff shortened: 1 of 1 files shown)


koic:
Thanks!

bbatsov:
Ops, seems I went AsciiDoc here! :D

PR #7897, opened 2020-04-20
Allow Generator to take version_added parameter

tejasbubane:
Other libraries like `rubocop-rspec` which rely on rubocop's generator
will be able to provide in their own version to be used in generated config. 

example: https://github.com/rubocop-hq/rubocop-rspec/pull/905

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/generator.rb
+++ b/lib/rubocop/cop/generator.rb
@@ -132,11 +132,12 @@ module RuboCop
         ).inject
       end
 
-      def inject_config(config_file_path: 'config/default.yml')
+      def inject_config(config_file_path: 'config/default.yml',
+                        version_added: bump_minor_version)
         injector =
           ConfigurationInjector.new(configuration_file_path: config_file_path,
                                     badge: badge,
-                                    version_added: bump_minor_version)
+                                    version_added: version_added)
 
         injector.inject do
           output.puts(format(CONFIGURATION_ADDED_MESSAGE,
--- a/spec/rubocop/cop/generator_spec.rb
+++ b/spec/rubocop/cop/generator_spec.rb
@@ -280,6 +280,28 @@ RSpec.describe RuboCop::Cop::Generator do
                  "#{path}.\n")
       end
     end
+
+    context 'with version provided' do
+      it 'uses the provided version' do
+        expect(File).to receive(:write).with(path, <<~YAML)
+          Style/Alias:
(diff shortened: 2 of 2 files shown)


lib/rubocop/cop/generator.rb
```diff
@@ -132,11 +132,12 @@ def inject_require(root_file_path: 'lib/rubocop.rb')
         ).inject
       end
 
-      def inject_config(config_file_path: 'config/default.yml')
+      def inject_config(config_file_path: 'config/default.yml',
+                        version_added: bump_minor_version)
```
koic, line 136: I think this API is also used in RuboCop core. It seems that the extended parameter expected to be sent by cop developer from terminal.

spec/rubocop/cop/generator_spec.rb
```diff
@@ -280,6 +280,28 @@ def on_send(node)
...
+    context 'with version provided' do
+      it 'uses the provided version' do
+        expect(File).to receive(:write).with(path, <<~YAML)
+          Style/Alias:
+            Enabled: true
+
+          Style/FakeCop:
+            Description: 'TODO: Write a description of the cop.'
+            Enabled: pending
+            VersionAdded: '1.52.1'
```
koic, line 293: Patch version is unnecessary.

lib/rubocop/cop/generator.rb
tejasbubane: `new_cop` task does not use any other parameter except cop name: https://github.com/rubocop-hq/rubocop/blob/218f476531aa596b89d0227f30530786a8899a82/tasks/new_cop.rake#L20

pirj:
To clarify: currently, if run from `rubocop-rspec`, `rake new_cop[RSpec/Abc]` generates the following in `rubocop-rspec`'s `config/default.yml`:
```yml
RSpec/Abc:
  Description: 'TODO: Write a description of the cop.'
  Enabled: pending
  VersionAdded: '0.83'
```

While RuboCop RSpec's current version is 1.38.1, and `VersionAdded` should be `1.38.2`.

tejasbubane:
Not sure why circleci is failing here.

pirj:
Seems to be a temporary failure, please re-run the build 🤞

pirj:
Good to merge?

koic:
I'll merge this one.

PR #7908, opened 2020-04-25
Add Style/RedundantRegexpEscape cop

owst:
This cop checks for redundant escapes inside Regexp literals.
```ruby
      # @example
      #   # bad
      #   %r{foo\/bar}
      #
      #   # good
      #   %r{foo/bar}
      #
      #   # good
      #   /foo\/bar/
      #
      #   # good
      #   %r/foo\/bar/
      #
      #   # bad
      #   /a\-b/
      #
      #   # good
      #   /a-b/
      #
      #   # bad
      #   /[\+\-]\d/
      #
      #   # good
      #   /[+\-]\d/
```

This would have been a good use of #7868 (which I really like the sound of @marcandre - 👍) - but instead I used the workaround he listed of "redoing the detection work" again.

-----------------

* [x] Added tests.


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New features
 
 * [#7895](https://github.com/rubocop-hq/rubocop/pull/7895): Include `.simplecov` file by default. ([@robotdana][])
+* [#7908](https://github.com/rubocop-hq/rubocop/pull/7908): Add unnecessary `/`-escapes offence to `Style/RegexpLiteral`. ([@owst][])
```
jonas054, line 8: `offenses` is the spelling we use.

owst:
Thanks for the review @jonas054 and good suggestion for the separate cop - I've made the change, which allowed me to also catch a lot of other unnecessary escapes, when you have a chance, let me know what you think 👍

lib/rubocop/cop/style/redundant_regexp_escape.rb
```diff
@@ -0,0 +1,118 @@
...
+      #   /[\+\-]\d/
+      #
+      #   # good
+      #   /[+\-]\d/
+      class RedundantRegexpEscape < Cop
+        include RangeHelp
+
+        MSG_REDUNDANT_ESCAPE = 'Redundant escape inside regexp literal'
+
+        ALLOWED_ALWAYS_ESCAPES = ' []^\\#'.chars.freeze
```
jonas054, line 37: I'm not sure why the space character is always allowed to escape. Is it when you use the `x` flag?

lib/rubocop/cop/style/redundant_regexp_escape.rb
```diff
@@ -0,0 +1,118 @@
...
+        ALLOWED_WITHIN_CHAR_CLASS_METACHAR_ESCAPES = '-'.chars.freeze
```
jonas054, line 38: You can always put a dash at the beginning or end of a character class and not have to escape it, right? It would require more logic in the cop, though.

config/default.yml
```diff
@@ -3610,6 +3610,11 @@ Style/RedundantPercentQ:
   Enabled: true
   VersionAdded: '0.76'
 
+Style/RedundantRegexpEscape:
+  Description: 'Checks for redundant escapes in Regexps.'
+  Enabled: true
+  VersionAdded: '0.85'
```
jonas054, line 3616: I'm not sure if the next version will be `0.85`, or if it will be `1.0`. @bbatsov knows.

manual/cops_style.md
```diff
@@ -5741,6 +5741,42 @@ question = '"What did you say?"'
 
 * [https://rubystyle.guide#percent-q](https://rubystyle.guide#percent-q)
 
+## Style/RedundantRegexpEscape
+
+Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
+--- | --- | --- | --- | ---
+Enabled | Yes | Yes  | 0.83 | -
```
jonas054, line 5748: Looks like the file needs to be generated again, looking at the "version added".

lib/rubocop/cop/style/redundant_regexp_escape.rb
owst: Yes, that's true - it's similar to how `^` only needs to be escaped if it is the first character of a character class. 

My two reasons for not flagging/removing such escapes is that it would involve more logic as you say, and make the resulting character classes slightly more "fragile" to changes over time - either by adding another character or by reordering the elements of the character class, e.g.:
```
[a-] => [a-b]
[-a] => [b-a]
[ab-] => [a-b]
[a^] => [^a]
```
would all change the meaning of the character class. The refactorings may not be common, but I think I fall down on allowing escapes of `^` and `-` even though they are technically redundant. What do you think? I can take a look at the implementation would be if you think it's worth flagging such cases.

manual/cops_style.md
owst: Oops, yes, good catch, I missed this after I merged master and bumped the `VersionAdded` you noted above - I'll regenerate now and can change again when @bbatsov advises about the next version number

lib/rubocop/cop/style/redundant_regexp_escape.rb
owst: Yes, exactly. I could add another explicit message that said "Unnecessary escape of space, when not using free-spacing (x) mode."?

config/default.yml
bbatsov: Seems it will be 0.85. Too much development happening recently. I want to give @marcandre time to finish the things he has in progress.

bbatsov:
Great work!

PR #7909, opened 2020-04-26
Fix a false positive for `Lint/ParenthesesAsGroupedExpression`

koic:
This PR fixes a false positive for `Lint/ParenthesesAsGroupedExpression` when using an intended grouped parentheses.

```ruby
puts (2 + 3) * 4
```

I noticed this false positive in implementing this cop's auto-correction. So, I plan to open a PR that implements auto-correction separately from this PR.

-----------------

* [x] Added tests.


bbatsov:
The fix looks good, but we should probably tweak the examples in the docs, as they look a lot like math expressions. :-)

PR #7919, opened 2020-04-29
[Fix #7855] show pending cops warnings in case passing configurationfile with -c option

### LINKED ISSUE #7855: Passing a -c configuration to rubocop hides the pending cops warnings
## Expected behavior

When `rubocop` is called by passing a configuration file (`rubocop -c .rubocop.yml`) it shows the pending cops warnings like it does when it's called without configuration file (`rubocop`).

## Actual behavior

By passing the configuration file it's not showing any warning. I know it's using the `.rubocop` file in both cases because it's not showing the frozen string warning.

```
mkdir rubocop-bug
cd rubocop-bug
echo "source 'https://rubygems.org'\ngem 'rubocop', '0.81'" > Gemfile
echo "def test\n  raise Exception, 'Error message here'\nend" > test.rb
echo "Style/FrozenStringLiteralComment:\n  Enabled: false" > .rubocop.yml
bundle
rubocop
rubocop -c .rubocop.yml
```

## Output
```
$ rubocop    
The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file:
 - Lint/RaiseException (0.81)
 - Lint/StructNewOverride (0.81)
 - Style/HashEachMethods (0.80)
 - Style/HashTransformKeys (0.80)
 - Style/HashTransformValues (0.80)
For more information: https://docs.rubocop.org/en/latest/versioning/
Inspecting 2 files
..

2 files inspected, no offenses detected

$ rubocop -c .rubocop.yml
Inspecting 2 files
..

2 files inspected, no offenses detected
```

## RuboCop version
0.81.0

## Ruby version
2.7.0

denys281:
Fix  #7855

Without `-c` option rubocop do some manipulation with config files here: 

https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/config_loader.rb#L96

But in case we pass directly config file with `-c` option all that we do with configs happening here: https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/config_store.rb#L22 so there we get missing warnings check; 

I fixed it. but specs needed, and I have no idea where to add it, any suggestions? Also should it be some refactoring here? Maybe there are other things missing? 

cc @koic @bbatsov @pirj 

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/config_store.rb
+++ b/lib/rubocop/config_store.rb
@@ -23,6 +23,8 @@ module RuboCop
       loaded_config = ConfigLoader.load_file(options_config)
       @options_config = ConfigLoader.merge_with_default(loaded_config,
                                                         options_config)
+
+      check_pending_cops(loaded_config)
     end
 
     def force_default_config!
@@ -43,6 +45,16 @@ module RuboCop
                                 print "For #{dir}: " if ConfigLoader.debug?
                                 ConfigLoader.configuration_from_file(path)
                               end
+    end
+
+    private
+
+    def check_pending_cops(loaded_config)
+      @options_config.tap do |merged_config|
+        unless ConfigLoader.possible_new_cops?(loaded_config)
+          ConfigLoader.warn_on_pending_cops(merged_config.pending_cops)
+        end
+      end
     end
   end
 end


koic:
I have an implementation in my local environment and the test code has not been completed yet. Perhaps this issue is a slightly more complicated change that includes other problems. So, the amount of change is likely to be a little too high now.

Anyway, I'll consider resuming the work I was pending. At that time, it will be opened as another new PR.
I'm sorry for the investigation and taking up your time. I should have assigned myself to the issue.

denys281:
okay. Closing PR

PR #7920, opened 2020-04-30
[Fix #595] Add ERB pre-processing for configuration files

### LINKED ISSUE #595: Exclude files in .gitignore
I think this would be a nice feature, the easiest implimention would be to run the `rubocop.yml` through erb so that you could have something like:

```
 Excludes:
   <% `git whatever`.each_line do |path| %>
    - <%= path %>
   < % end %>
```

I'd be happy to look at a PR if you think this is useful.

jonas054:
A typical use case for this new feature would be excluding files that are ignored by Git.


AlexWayfer:
Thank you!

PR #7921, opened 2020-04-30
Implement Style/SlicingWithRange cop

zverok:
Under Ruby >= 2.6, proposes to change `ary[x..-1]` to `ary[x..]`. This is what endless ranges were introduced for, and it is more obvious than a (slightly weird) agreement of "`-1` is the end".

Works only for integer `x` and only for one-argument `[]` method; correctly ignores exclude-end ranges.

Supports autocorrect.

Example of output:

```
ary[1..-1]
    ^^^^^ When slicing array till its end, prefer endless range: ary[n..] instead of ary[n..-1].
```

-----------------

* ~[ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).~
* [x] Added tests.


lib/rubocop/cop/style/endless_range.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        minimum_target_ruby_version 2.6
+
+        MSG = 'Prefer endless ranges (start..) for slicing '\
+              'till the end of array.'
+
+        def_node_matcher :range_till_minus_one?, '(irange (int _) (int -1))'
+
+        def on_send(node)
+          return unless node.method?(:[]) && node.arguments.count == 1
```
rrosenblum, line 26: I think you could handle all of these checks with a single `def_node_matcher`. It's been a little while since I wrote one, I think it would look like `(send _ :[] (irange (int _) (int -1)))`

spec/rubocop/cop/style/endless_range_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      RUBY
+    end
+
+    it 'reports no offense for array with range inside' do
+      expect_no_offenses(<<~RUBY)
+        ranges = [1...-1]
+      RUBY
+    end
+
+    it 'autocorrects' do
```
rrosenblum, line 40: You can handle auto-correction in the same method where you check for the offense by using `expect_correction`.
```ruby
expect_offense(<<~RUBY)
  ary[1..-1]
      ^^^^^ Prefer endless ranges (start..) for slicing till the end of array.
RUBY

expect_correction(<<~RUBY)
  ary[1..]
RUBY
```

lib/rubocop/cop/style/endless_range.rb
zverok: Of course I can, but I believe it is better this way for concern separation.

"Works only for method `#[]` with a singular argument" is an arbitrary decision, which can be changed in future and/or depend on settings (for ex., one may also expect `#slice` to be checked, or multi-argument `#[]` when workin with matrices). But "range ending in `-1`" is what the cop is about. So I believe it is better to leave as two separate checks.

spec/rubocop/cop/style/endless_range_spec.rb
zverok: Oh, cool, thanks. Fixed.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New features
 
+* Add [Style/EndlessRange] cop. ([@zverok][])
```
bbatsov, line 7: I guess you can link to your PR here.

bbatsov:
> UPD: Looked at it with fresh eyes, maybe name is too vague? Like SliceWithEndlessRange would be better?

That's a good point. In general we should choose more neutral names and aim to support both styles, even if I agree what you suggested is the better style.

I'd also suggest adding a guideline on the subject in the Ruby style guide.

lib/rubocop/cop/style/endless_range.rb
```diff
@@ -0,0 +1,40 @@
...
+      #   items[1..-1]
+      #
+      #   # good
+      #   items[1..]
+      class EndlessRange < Cop
+        extend TargetRubyVersion
+
+        minimum_target_ruby_version 2.6
+
+        MSG = 'Prefer endless ranges (start..) for slicing '\
```
bbatsov, line 20: I guess we can improve this message by inserting code snippets in it - e.g. prefer [new code] over [existing code]. I've noticed that people tend to understand such messages better.

manual/cops_style.md
```diff
@@ -1996,6 +1996,25 @@ at_exit { puts 'Goodbye!' }
 
 * [https://rubystyle.guide#no-END-blocks](https://rubystyle.guide#no-END-blocks)
 
+## Style/EndlessRange
```
bbatsov, line 1999: @koic I guess it won't be a bad idea to somehow incorporate the required Ruby version in this docs snippet. Might be a good idea to draw people's attention to the fact that some cops run only on certain Rubies.

koic: This idea looks good to me! So, each Ruby version supports new syntax and APIs. It can be a work independent of this PR :-)

bbatsov: Yeah, totally. I just thought of this while going over this PR and thought I should share my thoughts with you.

zverok:
@bbatsov @koic I've updated the PR according to comments (no "version in docs"/"update to styleguide" yet, tho). Can you please take a look?

lib/rubocop/cop/style/slicing_with_range.rb
```diff
@@ -0,0 +1,40 @@
...
+      #   items[1..-1]
+      #
+      #   # good
+      #   items[1..]
+      class SlicingWithRange < Cop
+        extend TargetRubyVersion
+
+        minimum_target_ruby_version 2.6
+
+        MSG = 'When slicing array till its end, prefer endless range: '\
```
bbatsov, line 20: You can just shorten this to `Prefer x over y`, as I think the rest is implied anyways. Plus, this way it's going to be more consistent with the rest of the messages used by RuboCop.

lib/rubocop/cop/style/slicing_with_range.rb
```diff
@@ -0,0 +1,40 @@
...
+              'ary[n..] instead of ary[n..-1].'
```
bbatsov, line 21: Ideally here we should use the actual array name from the code, but that's not a big deal.

lib/rubocop/cop/style/slicing_with_range.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # This cop checks that arrays are sliced with endless ranges instead of
+      # `ary[start..-1]` on Ruby 2.6+
```
bbatsov, line 7: This sentence should end with a `.`.

bbatsov:
Looks good to me overall. I've added a few tiny remarks, though. Btw, rebase on top of `master` and re-run the docs generator and you'll get the required Ruby version in the docs.

lib/rubocop/cop/style/slicing_with_range.rb
zverok: Thought about it but decided against it: realistic case may look like  `chain.of(...).lot...of.operations(...)[calculating.the.first(...).index..-1]`, and trying to reproduce the whole statement in cop's message will blur the intention.

zverok:
@bbatsov All fixed.
(About error messages — I thought a bit about their "teaching/explanatory" value, looking at Elixir's [credo](https://github.com/rrrene/credo), but that's a large topic unrelated to this PR.)

bbatsov:
Thanks, Victor! 

Yeah, the messages are a whole different topic indeed. I'm all for descriptive messages, but there's also the matter that those are used in many different context (not just in the terminal) and longer messages are problematic in some of those. That's why I also like to point people for details to the style guide, as there it's much easier to have detailed explanations.

zhuravel:
Sample lines for which this cop does not generate warnings:
```ruby
integer, fractional = number[0 ... most_right_index], number[most_right_index.next .. -1]
integer, fractional = number[0...most_right_index], number[most_right_index.next..-1]
integer, fractional, delimiter = number[0 ... i], number[i.next .. -1], number[i]

zverok:
@zhuravel thanks, I'll investigate

PR #7922, opened 2020-04-30
Add new `Layout/EmptyLineAroundAttributeAccessor` cop

koic:
This PR adds new `Layout/EmptyLineAroundAttributeAccessor` cop that checks for a newline after attribute accessor.

```ruby
# bad
attr_accessor :foo
def do_something
end

# good
attr_accessor :foo

def do_something
end

# good
attr_accessor :foo
attr_reader :bar
attr_writer :baz
attr :qux

def do_something
end
```

This PR first checks an empty line after an attribute accessor to implement the example shown in The Ruby Style Guide's rule.
https://github.com/rubocop-hq/ruby-style-guide/pull/817

In the future this cop will provide an `EnforcedStyle: around` that checks empty lines before and after an attribute accessor, and this `EnforcedStyle: only_after` option could be users selectable option.

-----------------

* [x] Added tests.


koic:
A legacy method `attr` was not included in the documentation. However, this cop's actual behavior also detects `attr`, so I updated the documentation to mention `attr` as well.

bbatsov:
I guess we're in business then! :-) Thanks!

aried3r:
Hey! I just updated to rubocop 0.83 and tried this new cop. I get the intention behind it when reading the examples, but how does the style guide suggest handling situations like these:

```Ruby
attr_accessor :validate_vat
alias validate_vat? validate_vat

# potentially more code

validates :vat_number, presence: true, if: :validate_vat?

# more code 

def do_something
  do_the_thing if validate_vat?
end
```

Personally I like to keep aliasing to add `?` close to `attr_accessor` instead of separating it.

I know I can just disable this cop, but I was wondering if that's a common pattern the cop should be able to handle or if that's very specific to our code base at work.

marcandre:
I agree with @aried3r ; ideally the cop should be able to ignore lines like the `alias`, as well as any `send` call with an argument matching any of the argument of `attr_accessor` (e.g. `alias_method :anything, :validate_vat`, `protected :validate_vat`, etc.)

koic:
I see. Perhaps customizable allow list such as `IgnoredMethods` option can be created. I think it can have some allowed method names (e.g. `alias`) by default.

PR #7923, opened 2020-05-01
[Fix #7390] Override disabled department for enabled cops

### LINKED ISSUE #7390: Override disabling a department

Hi, I am so confused about the current behavior to disable a department.

For example, assume that there are the below two files (`.rubocop.yml` and `test.rb`):

```yaml
# .rubocop.yml
Style:
  Enabled: false

Style/StringHashKeys:
  Enabled: true
```

```ruby
# test.rb
{ "foo" => 1 }
```

When I looked at the `.rubocop.yml` above at first, I misunderstood that  `Style/StringHashKeys` cop is **enabled**. 😓 

However, `Style/StringHashKeys` cop is actually **disabled** in the settings above. The manual is [here](https://github.com/rubocop-hq/rubocop/blob/v0.75.0/manual/configuration.md#enabled).

I think this behavior is very misleading. Because the settings above are equivalent to:

```yaml
Style:
  Enabled: false

Style/StringHashKeys:
  Enabled: false
```

The settings which look different are actually the same. In the settings above, I think the behavior to "disable all Style cops except for Style/StringHashKeys" should be more natural. What do you think?

---

The reproduction example via Docker:

```dockerfile
# Dockerfile
FROM rubylang/ruby

RUN mkdir /work
WORKDIR /work
RUN gem install rubocop:0.75 && \
    echo 'Style: { Enabled: false }' >> .rubocop.yml && \
    echo 'Style/StringHashKeys: { Enabled: true }' >> .rubocop.yml && \
    echo '{ "foo" => 1 }' > test.rb

CMD ["rubocop"]
```

```shell
$ docker build -t rubocop-reproduction .
...
Successfully tagged rubocop-reproduction:latest

$ docker run -it --rm rubocop-reproduction
Inspecting 1 file
.

```

In this case, I expect that a `Style/StringHashKeys` violation should be reported.

I want a behavior that `<Department>/<Cop>: { Enabled: true }` overrides `<Department>: { Enabled: true }`. For example:

```yaml
Style:
  Enabled: false  #<= Disable all Style cops

Style/StringHashKeys:
  Enabled: true   #<= Enable the cop, but other Style cops are still disabled
```

The solution above may be a breaking change, but I don't have good alternatives to avoid the breaking change. 🤷‍♂  

## Additional context

The current behavior has been added via the commit <https://github.com/rubocop-hq/rubocop/commit/37943df5b62a10864c9159dd71e441d7e2304f65>.

The commit has been released with the version 0.36.0:
https://github.com/rubocop-hq/rubocop/releases/tag/v0.36.0

jonas054:
Implement the more intuitive functionality, that if a cop sets `Enabled: true` in user configuration, it will be enabled even if its department is disabled in the same configuration file, or in any inherited file.

### DIFF
--- a/spec/rubocop/config_loader_spec.rb
+++ b/spec/rubocop/config_loader_spec.rb
@@ -440,6 +440,93 @@ RSpec.describe RuboCop::ConfigLoader do
       end
     end
 
+    context 'when a department is disabled' do
+      let(:file_path) { '.rubocop.yml' }
+
+      shared_examples 'resolves enabled/disabled for all cops' do
+        |enabled_by_default, disabled_by_default|
+        it "handles EnabledByDefault: #{enabled_by_default}, " \
+           "DisabledByDefault: #{disabled_by_default}" do
+          create_file('grandparent_rubocop.yml', <<~YAML)
+            Metrics/AbcSize:
+              Enabled: true
+
+            Metrics/PerceivedComplexity:
+              Enabled: true
+
+            Lint:
+              Enabled: false
+          YAML
+          create_file('parent_rubocop.yml', <<~YAML)
+            inherit_from: grandparent_rubocop.yml
+
+            Metrics:
+              Enabled: false
+
+            Metrics/AbcSize:
+              Enabled: false
+          YAML
+          create_file(file_path, <<~YAML)
+            inherit_from: parent_rubocop.yml
+
+            AllCops:
+              EnabledByDefault: #{enabled_by_default}
+              DisabledByDefault: #{disabled_by_default}
+
+            Style:
+              Enabled: false
+
+            Metrics/MethodLength:
+              Enabled: true
+
+            Metrics/ClassLength:
+              Enabled: false
+
+            Lint/RaiseException:
+              Enabled: true
+
+            Style/AndOr:
+              Enabled: true
+          YAML
+
+          def enabled?(cop)
+            configuration_from_file.for_cop(cop)['Enabled']
+          end
+
+          # Department disabled in parent config, cop enabled in child.
+          expect(enabled?('Metrics/MethodLength')).to be(true)
+
+          # Department disabled in parent config, cop disabled in child.
+          expect(enabled?('Metrics/ClassLength')).to be(false)
+
+          # Enabled in grandparent config, disabled in parent.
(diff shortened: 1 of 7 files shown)


bbatsov:
I love this change! Should we also elaborate on this behaviour in the docs?

jonas054:
Good idea! I'll get on it.

spec/rubocop/config_loader_spec.rb
```diff
@@ -440,6 +440,99 @@
...
+          expect(enabled?('Style/AndOr')).to be(true)
+
+          # Pending in default config, department disabled in grandparent,
+          # cop enabled in child config.
+          expect(enabled?('Lint/RaiseException')).to be(true)
+
+          # Pending in default config.
+          expect(enabled?('Layout/SpaceAroundMethodCallOperator'))
+            .to eq(enabled_by_default ||
+                   (disabled_by_default ? false : 'pending'))
```
koic, line 524: The pending cops state depends on production config/default.yml. This test seems to fail when pending cops will be enabled by default in RuboCop 1.0.

So, the test will fail if config/default.yml would be edited as follows:

```diff
diff --git a/config/default.yml b/config/default.yml
index 9c25d5ee8..38ec16d61 100644
--- a/config/default.yml
+++ b/config/default.yml
@@ -1107,7 +1107,7 @@ Layout/SpaceAroundKeyword:

 Layout/SpaceAroundMethodCallOperator:
   Description: 'Checks method call operators to not have spaces around them.'
-  Enabled: pending
+  Enabled: true
   VersionAdded: '0.82'
```

jonas054: Yes, I think we can just update the test when we change the default for that cop.

koic: I think there are two problems.

- This test case check would not be possible when there is no pending cops (For example, immediately after a major release)
- This test may break with a major release

Unfortunately I don't yet have a way to solve these 💦 Anyway, this PR feature is very user-friendly and a great implementation 👍

jonas054: This small part of the spec example is not very important. I'll remove it to save us future headache. :slightly_smiling_face:

jonas054:
I've updated the spec example as per discussion and squashed into one commit.

PR #7924, opened 2020-05-01
[Fix #7885] Consider Layout/IndentationStyle:EnforcedStyle

### LINKED ISSUE #7885: False positive of `Style/IfUnlessModifier` with tabs indentation
Hello.

I'm trying to update RuboCop to version 0.82.0 for [Flame](https://github.com/AlexWayfer/flame).

I'm using tabs in this project.

And there is a condition which does not fit in one line.

Probably related to #5785

--------

## Expected behavior

No offenses.

## Actual behavior

Offense from `Style/IfUnlessModifier`.

```ruby
					if !@controller.actions.include?(action) || find_reverse_route(action)
						next
					end
```

image

## RuboCop version

```
0.82.0 (using Parser 2.7.1.1, running on ruby 2.7.1 x86_64-linux)
```

/cc @DracoAter

jonas054:
When calculating how wide the indentation is, we must look at the `EnforcedStyle` of the `Layout/IndentationStyle` cop. If it's `spaces`, then the indentation multiplier is 1. If it's `tabs` it's a bit more complicated, but that part was already handled correctly.

The specs needed to be updated with correct configuration for all cases with `tabs` indentation.

Whether or not we should consider the `Enabled` parameter is debatable. I've chosen to not care about `Enabled`, and to use the other parameter settings in any case.

### DIFF
--- a/lib/rubocop/cop/mixin/statement_modifier.rb
+++ b/lib/rubocop/cop/mixin/statement_modifier.rb
@@ -58,10 +58,11 @@ module RuboCop
       end
 
       def indentation_multiplier
-        return 1 if config.for_cop('Layout/IndentationStyle')['Enabled']
+        indentation_style_config = config.for_cop('Layout/IndentationStyle')
+        return 1 if indentation_style_config['EnforcedStyle'] == 'spaces'
 
         default_configuration = RuboCop::ConfigLoader.default_configuration
-        config.for_cop('Layout/IndentationStyle')['IndentationWidth'] ||
+        indentation_style_config['IndentationWidth'] ||
           config.for_cop('Layout/IndentationWidth')['Width'] ||
           default_configuration
             .for_cop('Layout/IndentationStyle')['IndentationWidth'] ||
--- a/lib/rubocop/cop/mixin/statement_modifier.rb
+++ b/lib/rubocop/cop/mixin/statement_modifier.rb
@@ -4,6 +4,8 @@ module RuboCop
   module Cop
     # Common functionality for modifier cops.
     module StatementModifier
+      include LineLengthHelp
+
       private
 
       def single_line_as_modifier?(node)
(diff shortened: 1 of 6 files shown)


spec/rubocop/cop/style/if_unless_modifier_spec.rb
```diff
@@ -6,7 +6,11 @@
   subject(:cop) { described_class.new(config) }
 
   let(:config) do
-    RuboCop::Config.new('Layout/LineLength' => line_length_config)
+    RuboCop::Config.new('Layout/LineLength' => line_length_config,
+                        'Layout/IndentationStyle' => {
+                          'Enabled' => true,
```
koic, line 11: I would expect the test to fail if `'Enabled' => false` and `'EnforcedStyle' => 'spaces'`, but the test passes unexpectedly. Perhaps the changes to lib/rubocop/cop/mixin/statement_modifier.rb may not be enough.

lib/rubocop/cop/mixin/statement_modifier.rb
```diff
@@ -58,10 +58,11 @@ def max_line_length
       end
 
       def indentation_multiplier
-        return 1 if config.for_cop('Layout/IndentationStyle')['Enabled']
+        indentation_style_config = config.for_cop('Layout/IndentationStyle')
+        return 1 if indentation_style_config['EnforcedStyle'] == 'spaces'
```
koic, line 62: IMHO, looks like `indentation_style_config['Enabled']` check is needed.

spec/rubocop/cop/style/if_unless_modifier_spec.rb
jonas054: My thinking, when I made the change, was that `Style/IfUnlessModifier` needs to find out how many spaces each tab corresponds to when it finds indentation with tabs. But I see now that I haven't really followed though on this idea. 

I still believe that it should not matter to `Style/IfUnlessModifier` if 'Layout/IndentationStyle' is enabled or not when it tries to calculate the total line length of a line that is indented with tabs, but I'll try to come up with something better.

lib/rubocop/cop/mixin/statement_modifier.rb
jonas054: Just for clarity, what do you think should happen if indentation_style_config['Enabled'] is `true`?

jonas054:
@koic Thanks for the review! I've pushed more changes. Take a look and let me know if you still think we need to check `Layout/IndentationStyle:Enabled`.

lib/rubocop/cop/mixin/statement_modifier.rb
```diff
@@ -4,6 +4,8 @@ module RuboCop
   module Cop
     # Common functionality for modifier cops.
     module StatementModifier
+      include LineLengthHelp
```
jonas054, line 7: I'm not super happy about mixing in another module here, polluting the namespace, so to speak. I'm open to suggestions.

lib/rubocop/cop/mixin/statement_modifier.rb
koic: Yeah, I was concerned about changing the behavior for users who have `indentation_style_config['Enabled']` is `false`.
I think it's intuitive that if a user have disabled `Enabled`, the `EnforcedStyle` will also be disabled.

jonas054: That would mean that a user or team that want to allow indentation with spaces or with tabs, without enforcing one style, has no way of specifying what the tab length is.

I don't think we have followed that principle before, that we need to check `Enabled` before using a cop's configuration from another cop, or at least it's very rare. If I search for `config.for_cop` under `rubocop/cop/` I find many of these calls, but hardly any checking of `Enabled`.

jonas054:
Oops! I just realized that after starting to re-use `LineLengthHelp` for line length calculation, the subject line description of the PR no longer holds true. It's best if I close the PR, and we can continue the discussion after I submit a new PR.

PR #7926, opened 2020-05-03
[Fix #7885] Re-use LineLengthHelp for statement modifiers

### LINKED ISSUE #7885: False positive of `Style/IfUnlessModifier` with tabs indentation

jonas054:
_This is a new attempt at a PR that was first made in #7924. The changes are pretty much the same but the description should be more correct this time._

There's no reason to have two implementations of the same logic. They weren't exactly the same, but I think the one in `LineLengthHelp` was correct, apart from falling back to `Layout/IndentationWidth:Width` if `Layout/IndentationStyle:IndentationWidth` is `nil`, which is now added to match the description of the `Layout/IndentationStyle` parameters in `default.yml`.

Added some examples on line length with tabs indentation in `line_length_spec.rb`.

Updated `if_unless_modifier_spec` to cover the fallback on `Layout/IndentationWidth:Width` better, and removed an example that could not happen in real execution ("without any IndentationWidth config"). The spec examples now work like before, but without requiring `Layout/IndentationStyle` to be disabled.


AlexWayfer:
@jonas054, thank you, but I've found similar bug: #8041

PR #7931, opened 2020-05-04
Use did_you_mean instead of jaro_winkler

bquorning:
The `did_you_mean` gem has been a part of Ruby since v2.3, so if we can use that instead of the `jaro_winkler` gem our users will have 1 less dependency.

-----------------

* [ ] Added tests.


rubocop.gemspec
```diff
@@ -33,7 +33,7 @@ Gem::Specification.new do |s|
     'bug_tracker_uri' => 'https://github.com/rubocop-hq/rubocop/issues'
   }
 
-  s.add_runtime_dependency('jaro_winkler', '~> 1.5.1')
+  s.add_runtime_dependency('did_you_mean')
```
tas50, line 36: If did_you_mean is built in then isn't this unecesssary?

bquorning:
rubocop-rails uses `NameSimilarity`: https://github.com/rubocop-hq/rubocop-rails/blob/891c6aad8a6ac03b89a2f2bfe403554021c9dde0/lib/rubocop/cop/rails/unknown_env.rb#L18

So although this PR only changes internal API, it is actually a breaking change. Hmm, any suggestions on how to fix that?

koic:
The API may be used by third party cop that rubocop-hq org does not managed.
On the other hand, I think that it is not a major change. If the changelog mentions this change, I think it's okay. And I've looked at rubocop-rails repo, probably don't need `**(BREAK)**` mark.

rrosenblum:
This isn't the typical kind of breaking change that impacts the end user. It is a breaking change that potentially impacts other libraries that add onto RuboCop. It will be important to note that this is a breaking change for them. This could impact an end user if a RuboCop extension allows for too high of a version of RuboCop to be included. I'm not sure of the best way to handle this.

bquorning:
One possible way of handling it would be to leave the current implementation intact, with a deprecation warning.

Then add this new implementation of `#find_similar_name` under another name. Or with the same name, but adding keyword arguments?
```diff
-def find_similar_name(target_name, scope)
+def find_similar_name(target_name = nil, names = nil, name: nil, candidates: nil)
+  warn "deprecated" if target_name && names && name.nil? && candidates.nil?
```

…something along those lines.

bbatsov:
@bquorning I don't this we need to overdo this. As @koic said the impact of the change is going to be small (if any). I'd rather us removed this before the 1.0 release. Just update the changelog and we're good to go.

bquorning:
> Just update the changelog and we're good to go.

Done

bbatsov:
Thanks!

koic:
FYI, the edge RuboCop Rails has been independence from `RuboCop::NameSimilarity` API.
https://github.com/rubocop-hq/rubocop-rails/commit/be82f3b0d2a1966089738849106d99bee4d3df19

PR #7933, opened 2020-05-05
Add new Layout/PrivateConstant cop

geshwho:
This cop checks that there are no constants defined in the private or
protected blocks of a class. This does not actually make the constant
private and it will still be publicly accessible. The cop will ignore
constants that are explicitly marked with `private_constant`.

Addresses a closed issue raised in rubocop-hq/ruby-style-guide#536.

-----------------
Hi team! I recently came across some code that inspired this cop while reviewing some code. Placing constants in the private block can be misleading because it can lead the person reading/writing it to think that the constant is not publicly accessible when this is in fact not true. The issue linked above suggested using consistent classes (e.g. `Layout/ClassStructure`). This works well, but is a much broader and stricter rule than this proposition.

Let me know what you think, and if this is something that you'd consider adding to the project. Thanks!

-----------------

* [x] Added tests.


bbatsov:
I think this logic belongs to the existing `Lint/UselessAccessModifier` cop. Currently it checks for class methods, but this problem you outline is pretty much the same.

mtamhankar1:
> I think this logic belongs to the existing `Lint/UselessAccessModifier` cop. Currently it checks for class methods, but this problem you outline is pretty much the same.

:+1: I did not know about this cop. I can take a stab at extending it to support constants in addition to class methods.

lib/rubocop/cop/layout/private_constant.rb
```diff
@@ -0,0 +1,91 @@
...
+                add_offense(child)
+              end
+            end
+          end
+        end
+
+        def explicit_private_declaration(node)
+          constant_name = node.children[1]
+          node.parent.children.any? do |n|
+            NodePattern.new(
```
rrosenblum, line 83: This `NodePattern` is being redefined for each iteration of the array. I think you could improve the performance of this by moving the `NodePattern` declaration outside of the this loop.

```ruby
constant_name = node.children[1]
pattern = NodePattern.new("(send nil? :private_constant (:sym :#{constant_name}))")
node.parent.children.any? do |n|
  pattern.match(n)
end
```

lib/rubocop/cop/layout/private_constant.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Layout
+      # Placing constants in the private section does not make them private,
```
rrosenblum, line 6:
```suggestion
      # Placing constants in the `private` section does not make them private,
```

lib/rubocop/cop/layout/private_constant.rb
marcandre: It should actually be compiled *once*, using `%`...

bbatsov:

PR #7934, opened 2020-05-06
Support autocorrection for `Lint/EnsureReturn`

koic:
This PR supports autocorrection for `Lint/EnsureReturn`.

-----------------

* [x] Added tests.


marcandre:
Sadly, these autocorrections are not producing equivalent code. Removing the `return` like this can change the control flow within the `ensure` block:

```ruby
ensure
  return foo if condition
  do_something_else
# can not be corrected to
ensure
  foo if condition
  do_something_else
```

More importantly, even when there are no exceptions raised any autocorrection changes the return value (as the result of the `ensure` block is ignored)

This autocorrection could also swallow user comments.

I don't know why, but this PR also stopped flagging returns with multiple values, but the argument against `return` in `ensure` is the same.

I'm reverting this in #8388

PR #7939, opened 2020-05-07
[Fix #7929] Accept frozen_string_literal anywhere in leading comment lines

### LINKED ISSUE #7929: Bug in Style/FrozenStringLiteralComment where comment can exist anywhere in file
Looking at this history of the `Style/FrozenStringLiteralComment` cop it looks like it:
- originally only checked the first three lines of source: https://github.com/rubocop-hq/rubocop/blob/58f2f8d37b733436bbc6cfa49594828699825de2/lib/rubocop/cop/mixin/frozen_string_literal.rb#L37
- was updated to check the first three comments but still looked at the LOC to see it was within the first three: https://github.com/rubocop-hq/rubocop/blob/75eb10738e356121bcf5dc6ce54d0ba45cd1e595/lib/rubocop/cop/mixin/frozen_string_literal.rb#L42-L50
- then was updated to simply check the first three comments (to handle cases where comments were separated by newlines): https://github.com/rubocop-hq/rubocop/commit/8c792d1349e7cdf383ebde08b00c83363d52db05

`frozen_string_literal` actually must appear before any code. Here's an example that should raise an exception at runtime but doesn't:
```ruby
1 + 2
# frozen_string_literal: true

K = "foo"
K.slice!(4)
```

However, since the cop is checking the first three comments they can appear anywhere in the source. For example, according to the cop these are OK:
```ruby
class Something
end

# frozen_string_literal: true
```

```ruby
# One

class Something
  # Two
end

# frozen_string_literal: true
```

This is actually invalid but "accidentally" because the cop has logic about this needing to be one of the first three comments:
```ruby
# One

class Something
  # Two
end

# Three
# frozen_string_literal: true
```

There's actually nothing special about the first three lines/comments either. This should be valid as well:
```ruby
# something else that's important
# I love comments
# comments commenty comments
# Oh boy, more comments

# frozen_string_literal: true
q = 1 + 2
K = "foo"
K.slice!(4)
```

The reason this cop cares about "first three" is because it has special logic dealing with the `shebang` and `encoding` comments.

--------

## Expected behavior

This cop would enforce the string literal comment is somewhere in the _leading_ comments (i.e. comments appearing before any ruby code), not necessarily just the first three.

## Actual behavior

The cop checks the first three comments in source, regardless of their location in the file

See test cases above.

## RuboCop version

```
$ [bundle exec] rubocop -V
0.82.0 (using Parser 2.5.1.2, running on ruby 2.5.1 x86_64-linux)
```

jeffcarbs:
Fixes #7929 

The frozen_string_literal comment can exist anywhere in the "leading comment lines" (i.e. before the first line of ruby code). If it appears after any ruby code it is ignored which does not cause any errors but could cause the absence of an expected error which is bad.

The previous implementation was checking if this comment was one of the first three comments which had two issues:
- The comments it was checking could be anywhere, even after ruby code
- Since it was only checking three comments, it's possible that you can have an appropriate frozen_string_literal but this cop would show an error.

This commit updates the logic to look at all leading comment lines, which may be more or less than three depending on where the first non-comment token is.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/frozen_string_literal_comment_spec.rb
+++ b/spec/rubocop/cop/style/frozen_string_literal_comment_spec.rb
@@ -249,6 +249,39 @@ RSpec.describe RuboCop::Cop::Style::FrozenStringLiteralComment, :config do
       RUBY
     end
 
+    it 'accepts a frozen string literal comment after other comments' do
+      expect_no_offenses(<<~RUBY)
+        #!/usr/bin/env ruby
+        # encoding: utf-8
+        #
+        # These are more comments
+        # frozen_string_literal: false
+        puts 1
+      RUBY
+    end
+
+    it 'registers an offense for having a frozen string literal comment ' \
+       'under ruby code' do
+      expect_offense(<<~RUBY)
+        # encoding: utf-8
+        ^ Missing frozen string literal comment.
+
+        puts 1
+        # frozen_string_literal: true
+      RUBY
+
+      # Since we're only looking at the leading comments, a
+      # `frozen_string_literal` comment elsewhere in the code is invisible
+      # to this cop so the autocorrect won't remove it.
+      expect_correction(<<~RUBY)
+        # encoding: utf-8
+        # frozen_string_literal: true
+
+        puts 1
+        # frozen_string_literal: true
+      RUBY
+    end
+
     it 'registers an offense for an extra first empty line' do
       pending 'There is a flaw that skips adding caret symbol in this case, ' \
               'making it impossible to use `expect_offense` matcher'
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/frozen_string_literal_comment_spec.rb
```diff
@@ -249,6 +249,39 @@
...
+        # encoding: utf-8
+        ^ Missing frozen string literal comment.
+
+        puts 1
+        # frozen_string_literal: true
+      RUBY
+
+      # Since we're only looking at the leading comments, a `frozen_string_literal`
+      # comment elsewhere in the code is invisible to this cop so the autocorrect
+      # won't remove it.
```
jeffcarbs, line 275: I would be happy to take a tab at implementing this, but I'm hesitant since the `autocorrect` method takes a specific node and we'd have to update it to "correct" code elsewhere in the source outside of that node.

spec/rubocop/cop/style/frozen_string_literal_comment_spec.rb
```diff
@@ -249,6 +249,39 @@
...
+        # encoding: utf-8
+        ^ Missing frozen string literal comment.
+
+        puts 1
+        # frozen_string_literal: true
+      RUBY
+
+      # Since we're only looking at the leading comments, a
+      # `frozen_string_literal` comment elsewhere in the code is invisible
+      # to this cop so the autocorrect won't remove it.
```
jeffcarbs, line 275: I would be happy to take a stab at implementing this, but I'm hesitant since the `autocorrect` method takes a specific node and we'd have to update it to "correct" code elsewhere in the source outside of that node.

koic: I think this case does not need to be taken care of. Because a frozen string magic comment written other than the reading comments may be an intended content in the source code comments.

koic:
Thank you!

PR #7940, opened 2020-05-08
Add new Lint/NoReturnInBeginEndBlocks cop

jcfausto:
This cop detects the usage of explicit `return` statements inside a `begin..end` block in the context of assignments.

The issue is that a `return` inside a `begin..end` block in assignment contexts will not assign the desired value to the variable, but will return the value expected to the caller.

In case you're using memoization, the memoized variable will be always nil and the code inside the `begin..end` block will execute as many times as you access the memoized variable.

In regular assignments, the code after the `begin..end` block will never execute in case of return is executed as it'll immediately return the execution flow to the caller instead of just exiting from the `begin..end` block with a value as one might suppose.

Depending on how the caller assumptions are regarding the behaviour of the cached value, memoized, this issue might cause unexpected errors.

-----------------

* [-] Squashed related commits together.
* [x] Added tests.


lib/rubocop/cop/lint/no_return_in_begin_end_blocks.rb
```diff
@@ -0,0 +1,65 @@
...
+      #
+      #                     some_value
+      #                   else
+      #                     do_something
+      #                   end
+      #
+      class NoReturnInBeginEndBlocks < Cop
+        MSG = 'Do not `return` in `begin..end` blocks in assignment contexts.'
+
+        def on_lvasgn(node)
```
jonas054, line 47: This method is not covered by tests.

spec/rubocop/cop/lint/no_return_in_begin_end_blocks_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+RSpec.describe RuboCop::Cop::Lint::NoReturnInBeginEndBlocks do
+  subject(:cop) { described_class.new(config) }
+
+  let(:config) { RuboCop::Config.new }
+
+  it 'registers an offense when using `#bad_method`' do
+    expect_offense(<<~RUBY)
+      def bad_method
+        @bad_method ||= begin
```
jonas054, line 11: I think you should tests as many variations as possible on assignment operators, `=`, `+=` and so on, using `shared_examples`.

jcfausto:
Thanks for your review. I'll address the issues you suggested. 👍

spec/rubocop/cop/lint/no_return_in_begin_end_blocks_spec.rb
jcfausto: I've made some changes trying to cover the most common assignment operators. Could you have a look when you have some time, please? Thanks.

lib/rubocop/cop/lint/no_return_in_begin_end_blocks.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Checks for the presence of a *return* inside a *begin..end* block
+      # in assignment contexts.
+      # In this situation the, `return` will take precedence over any
```
jonas054, line 8: How about:

_In this situation, the `return` will result in an exit from the current method, possibly leading to unexpected behavior._

I guess that putting a `return` inside a `begin`..`end` is something that a Java or JavaScript programmer might do out of habit, since `return` in those languages just returns from the lambda/block/closure.

config/default.yml
```diff
@@ -1659,6 +1659,11 @@ Lint/NextWithoutAccumulator:
   Enabled: true
   VersionAdded: '0.36'
 
+Lint/NoReturnInBeginEndBlocks:
+  Description: 'Do not `return` inside `begin..end` blocks in assignment contexts.'
+  Enabled: pending
+  VersionAdded: '0.83'
```
jonas054, line 1665: This version number needs to be updated now.

jonas054 (changes requested):

spec/rubocop/cop/lint/no_return_in_begin_end_blocks_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+        return 1 if rand(1..2).odd?
+
+        2
+      end
+    RUBY
+  end
+
+  shared_examples 'accepts return inside a block' do |name, code|
+    it "accepts a return statement inside a block when using #{name}" do
+      expect_no_offenses(send(code))
```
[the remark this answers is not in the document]
marcandre: 👍  Moreover, these particular `shared_examples` are not needed at all and should be avoided altogether. I find it much harder to know what test is failing when shared examples are used.

Personally, I'm more lazy than @jonas054 and a single example of `*=` satisfies me, I wouldn't list all the `+=` and variants.

spec/rubocop/cop/lint/no_return_in_begin_end_blocks_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  [
+    ['simple assignment', 'return_inside_simple_assignment_block'],
+    ['add and assignment', 'return_inside_add_and_assignment_block'],
+    ['subtract and assignment', 'return_inside_subtract_and_assignment_block'],
+    ['multiply and assignment', 'return_inside_multiply_and_assignment_block'],
+    ['divide and assignment', 'return_inside_divide_and_assignment_block'],
+    ['exponent and assignment', 'return_inside_exponent_and_assignment_block']
+  ].each do |name, code|
+    include_examples 'accepts return inside a block', name, code
```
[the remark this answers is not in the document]
marcandre: Those `return`s are conditional so there are ok.

lib/rubocop/cop/lint/no_return_in_begin_end_blocks.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Checks for the presence of a *return* inside a *begin..end* block
```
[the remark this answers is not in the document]
jcfausto: Sure, will update that. 👍

lib/rubocop/cop/lint/no_return_in_begin_end_blocks.rb
jcfausto: Yes, I agree. I think your suggestion is nice. Apart from being more concise it's also easier to understand.

spec/rubocop/cop/lint/no_return_in_begin_end_blocks_spec.rb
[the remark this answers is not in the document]
jcfausto: Thank you both for the feedback. Yes, my initial implementation was not the best one. I really liked the way @jonas054 approached the implementation of the shared examples. As this issue will affect those operator assignments, I think it's valid to have this test for them. As they're not so many, in case of failures I believe it would be easy to spot what's going on. On a more complex scenario, perhaps I'd also avoid abusing the use of shared examples. But in this case I think it's fine :) 

I'm struggling at the moment to actually add an offense for the `=` (ivasgn) test case. Not sure how to detect the offense in this case as the only content for the node in this case only has `s(:ivasgn, :@good_method)` and nothing more.

Here's how I'm trying to detect the offense at the moment:

```ruby
def on_ivasgn(node)
  node.each_node(:kwbegin) do |kwbegin_node|
    kwbegin_node.each_node(:return) do |return_node|
      add_offense(return_node)
    end
  end
end
```

Based on this output from the AST parsed code:

```ruby
(def :bad_method_with_operator
  (args
    (arg :memoized)
    (arg :operator))
  (ivasgn :@bad_method
    (kwbegin
      (if
        (send
          (lvar :memoized) :!)
        (return
          (int 1)) nil)
      (int 2))))
```

Any advice you can share with me on how to solve this one?

Cheers.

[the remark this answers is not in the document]
jcfausto: Alright. I think I managed to solve the issue. The issue was that the cop's code was not considering multiple nodes (as in the example above) and was also returning when the node was not of a `begin_type`. 

I've updated the cop's code to consider these cases now, and also created an alias for the other operations as the check will be the same. 

With these changes, all tests are passing now for all the operators.

spec/rubocop/cop/lint/no_return_in_begin_end_blocks_spec.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Lint::NoReturnInBeginEndBlocks do
+  subject(:cop) { described_class.new(config) }
+
+  let(:config) { RuboCop::Config.new }
+
+  def return_inside_or_assignment_block
```
marcandre, line 8: Doesn't appear to be used

spec/rubocop/cop/lint/no_return_in_begin_end_blocks_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    <<-RUBY
+      @bad_method ||= begin
+        return "odd number" if rand(1..2).odd?
+
+        "even number"
+      end
+    RUBY
+  end
+
+  def no_return_inside_or_assignment_block
```
marcandre, line 18: Please don't create methods for this and instead use inline

spec/rubocop/cop/lint/no_return_in_begin_end_blocks_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        some_value #{operator} begin
+          return 1 if rand(1..2).odd?
+          ^^^^^^^^ Do not `return` in `begin..end` blocks in assignment contexts.
+          2
+        end
+      RUBY
+    end
+  end
+
+  shared_examples 'accetps a block with no return' do |operator|
```
marcandre, line 50: ^accetps^accepts

marcandre (changes requested):

marcandre:
Looks good, a few small things to tweak though

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@
 
 ### New features
 
+* Add new `Lint/NoReturnInBeginEndBlocks` cop. ([@jcfausto][])
```
jonas054, line 54: This line is in the wrong place now due to rebasing. Move it to the **master (unreleased)** heading, and also add a **New features** section there.

config/default.yml
```diff
@@ -1660,6 +1660,11 @@ Lint/NextWithoutAccumulator:
   Enabled: true
   VersionAdded: '0.36'
 
+Lint/NoReturnInBeginEndBlocks:
+  Description: 'Do not `return` inside `begin..end` blocks in assignment contexts.'
+  Enabled: pending
+  VersionAdded: '0.92'
```
jonas054, line 1666: Needs to be changed again. :slightly_smiling_face:

jonas054 (changes requested):
Found a couple more things to fix.

CHANGELOG.md
jcfausto: I see. I'll update that. 👍

spec/rubocop/cop/lint/no_return_in_begin_end_blocks_spec.rb
jcfausto: Yes, you're right. I also think this and the other method are not necessary anymore. I'll remove them from the test as the shared examples are covering all the situations now. 👍

jcfausto:
Thanks for the reviews. I applied the feedbacks and hopefully, now it's all according to the expectations :) Thanks for the rounds of review and the good feedback 👍

marcandre (approved):
LGTM 👍

jcfausto:
I'm not entirely sure how to proceed from here? Should I just wait now until the next round of merges? If there's anything else I need to do just let me know. :) Thanks.

config/default.yml
```diff
@@ -1667,6 +1667,11 @@ Lint/NextWithoutAccumulator:
   Enabled: true
   VersionAdded: '0.36'
 
+Lint/NoReturnInBeginEndBlocks:
+  Description: 'Do not `return` inside `begin..end` blocks in assignment contexts.'
+  Enabled: pending
+  VersionAdded: '0.93'
```
marcandre, line 1673: Please change to '1.1' 😅

marcandre:
Thanks for the ping @jcfausto and sorry, this fell through the cracks.

Please update the version (and the doc) and I'll merge this 😅

config/default.yml
jcfausto: Sure, no worries. I updated it to `1.1` and updated the doc as well. 👍

marcandre: Of course 1.1 has been released two days ago 🤣 

I'll merge and update myself...

Drenmi:
Interesting catch! Just fixed one of these instances at work thanks to this. 😄

PR #7943, opened 2020-05-08
Update target_ruby.rb

marcandre:
rubocop 0.81 is the last version to support 2.3 mode, not 0.82.

This is a partical revert of 44a5bee006bb075

### DIFF
--- a/lib/rubocop/target_ruby.rb
+++ b/lib/rubocop/target_ruby.rb
@@ -7,7 +7,7 @@ module RuboCop
     DEFAULT_VERSION = KNOWN_RUBIES.first
 
     OBSOLETE_RUBIES = {
-      1.9 => '0.50', 2.0 => '0.50', 2.1 => '0.58', 2.2 => '0.69', 2.3 => '0.82'
+      1.9 => '0.50', 2.0 => '0.50', 2.1 => '0.58', 2.2 => '0.69', 2.3 => '0.81'
     }.freeze
     private_constant :KNOWN_RUBIES, :OBSOLETE_RUBIES
 


lib/rubocop/target_ruby.rb
```diff
@@ -7,7 +7,7 @@ class TargetRuby
     DEFAULT_VERSION = KNOWN_RUBIES.first
 
     OBSOLETE_RUBIES = {
-      1.9 => '0.50', 2.0 => '0.50', 2.1 => '0.58', 2.2 => '0.69', 2.3 => '0.82'
+      1.9 => '0.50', 2.0 => '0.50', 2.1 => '0.58', 2.2 => '0.69', 2.3 => '0.81'
```
koic, line 10: It has been specified the RuboCop version with a Ruby version support dropped. Could you tell me why the version is wrong?

On the other hand, I noticed that the specifications are not consistent with Ruby 1.9 and Ruby 2.0. Below are the respective Ruby versions and the RuboCop versions that dropped it.

- Ruby 2.3 (0.82) https://github.com/rubocop-hq/rubocop/commit/5873894
- Ruby 2.2 (0.69) https://github.com/rubocop-hq/rubocop/commit/4910d51
- Ruby 2.1 (0.58) https://github.com/rubocop-hq/rubocop/commit/a707c23
- Ruby 2.0 (0.51) https://github.com/rubocop-hq/rubocop/commit/7532279
- Ruby 1.9 (0.42) https://github.com/rubocop-hq/rubocop/commit/9d27451

marcandre: > It has been specified the RuboCop version with a Ruby version support dropped

The name `OBSOLETE_RUBIES` is not very clear, but it indicates the last version where a Ruby version is supported.

The message says: "2.3-compatible analysis was dropped after version #{last_version}.".

The method using it is called `rubocop_version_with_support`.

marcandre: Note that this is the information that is actually helpful, i.e. "hey, I want to maintain my ruby_target so what version of RuboCop can I upgrade to then?"

marcandre: Final note: I didn't check the other Ruby versions than 2.3

koic, line 10: Thank you for the explanation. It looks like a misleading variable name. The local variable name `last_version` looks like it should be `version_removed` correctly.

bbatsov:
> Note that this is the information that is actually helpful, i.e. "hey, I want to maintain my ruby_target so what version of RuboCop can I upgrade to then?"

I think we should also add one compatibility matrix to the docs, as that'd be more useful to the end users.

PR #7950, opened 2020-05-10
Add new `Lint/DeprecatedOpenSSLConstant ` cop

bdewater:
With https://github.com/ruby/openssl/pull/366 the algorithm constants for `OpenSSL::Cipher` and `OpenSSL::Digest` are planned to be deprecated, see https://github.com/ruby/openssl/issues/304 for reasoning. This PR aims to introduce two cops to autocorrect this syntax, similar to e.g. [`Lint/BigDecimalNew`](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/lint/big_decimal_new.rb).

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -3422,6 +3422,11 @@ Style/OneLineConditional:
   VersionAdded: '0.9'
   VersionChanged: '0.38'
 
+Style/OpenSSLDigestConstant:
```
koic, line 3425: I think warnings for deprecated methods belong to the `Lint` department.

```suggestion
Lint/OpenSSLDigestConstant:
```

config/default.yml
```diff
@@ -3422,6 +3422,11 @@ Style/OneLineConditional:
...
+  Description: "Don't use algorithm constants for OpenSSL::Digest."
```
koic, line 3426: Can you enclose code with backticks? And I think that using `Reference` to indicate the information source will help users to understand :-)

```suggestion
  Description: "Don't use algorithm constants for `OpenSSL::Digest`."
  Reference: 'Please specify the information source URL here'
```

config/default.yml
```diff
@@ -3422,6 +3422,11 @@ Style/OneLineConditional:
...
+  Enabled: true
```
koic, line 3427: Can you specify `Enabled: pending` instead of `Enabled: true`?

```suggestion
  Enabled: pending
```

lib/rubocop/cop/style/open_ssl_digest_constant.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Algorithmic constants for OpenSSL::Digest are soft-deprecated since
```
koic, line 6:
```suggestion
      # Algorithmic constants for `OpenSSL::Digest` are soft-deprecated since
```

lib/rubocop/cop/style/open_ssl_digest_constant.rb
```diff
@@ -0,0 +1,91 @@
...
+      # version 2.2.0. Prefer passing a string instead to use hash functions
```
koic, line 7: It seems good to mention what version 2.2.0 the software name is :-)

lib/rubocop/cop/style/open_ssl_digest_constant.rb
```diff
@@ -0,0 +1,91 @@
...
+      # without defined constants like SHA512-256, SHA3-256 or BLAKE2s256.
+      #
+      # @example instantiation
```
koic, line 10: `@example` tag is used for `EnforcedStyle`. It is better to use comments for supplementary explanations.

```suggestion
      # @example
      # Below is an example for instantiation.
```

lib/rubocop/cop/style/open_ssl_digest_constant.rb
```diff
@@ -0,0 +1,91 @@
...
+      #
+      #   # bad
+      #   OpenSSL::Digest::SHA256.new
+      #
+      #   # good
+      #   OpenSSL::Digest.new('SHA256')
+      #
+      # @example ::Digest inherited class methods
```
koic, line 18: Ditto.

lib/rubocop/cop/style/open_ssl_digest_constant.rb
```diff
@@ -0,0 +1,91 @@
...
+          else
+            args = node.children.last.source
+            format(WITH_ARGS_MSG,
+                   method: method, algorithm: algorithm, original: node.source,
+                   args: args)
+          end
+        end
+
+        def digest_initialisation?(node)
+          node.loc.selector.source == 'new'
```
koic, line 82: Can you use `method?`?

```suggestion
          node.method?(:new)
```

bdewater:
Thanks for the review @koic!

koic:
I think `Lint/OpenSSLCipherConstant` and `Lint/OpenSSLDigestConstant`cops have the same check purpose, and they are enabled or disabled together. Also, the implementation is almost the same.
So I think it can be unified as `Lint/DeprecatedOpenSSLConstant` cop.

bbatsov:
We can even go one step further and have a `Lint/DeprecatedConstants` cop. I don't think we need a different cop for each deprecated constant. I think we already have something similar for methods.

bdewater:
Merged into one cop. A `Lint/DeprecatedConstants` cop like `Lint/DeprecatedClassMethods` is not going to work for the rewrite logic here I think 🤔

rhenium:
OpenSSL::Cipher::AES{128,192,256} have another initialization form - The cipher mode could be omitted for just these three constants.

```
OpenSSL::Cipher::AES128.new
```

needs to be auto-corrected into:

```
OpenSSL::Cipher.new("aes-128-cbc")
```

bdewater:
Thanks @rhenium! I pushed a fix.

bbatsov:
Thanks!

tas50:
Thanks for cutting a cop for this deprecation @bdewater. It makes scoping Ruby upgrades a lot more predictable.

PR #7955, opened 2020-05-12
Using rubocop ast

marcandre:
Checking to see if CI breaks after splitting off `rubocop-ast`...


bbatsov:
Well, I guess you got it right from the first try! 🥇

marcandre:
Apparently 😅

So you're 👍 to merge?

I'm fixing the workflow on `rubocop-ast` so it uses `rubocop` from source. Should I do the same for `rubocop` so it always tests against the `master` branch? Or the latest release is better?

bbatsov:
I think the latest release is better. 

> So you're 👍 to merge?

Totally!

marcandre:
I'm just thinking that I should have modified `rubocop -V` to print out the version of `rubocop-ast`.
I'll amend the CHANGELOG.
Just to be sure: it's ok that I commit directly stuff I'm pretty confident about, and continue creating for the rest?

bbatsov:
> I'm just thinking that I should have modified rubocop -V to print out the version of rubocop-ast.

Yeah, that's a good idea.

> Just to be sure: it's ok that I commit directly stuff I'm pretty confident about, and continue creating for the rest?

Sure. If you don't need input for something you can commit right away.

PR #7964, opened 2020-05-13
Fix calculation in disable_offense.

marcandre:
The column where the offence begins has nothing to do with if the line is small enough to contain `eol_comment` or not.


koic:
It makes sense and thank you for adding tests!

PR #7970, opened 2020-05-13
Improved shared context. Refactor & simplify some specs.

marcandre:
I feel there are ways to clean some specs up with default rspec `let` groups.

How does this look for a start?

Stumbling upon this when working on #7868 because some specs are really abusing the API for autocorrection.

### DIFF
--- a/lib/rubocop/rspec/shared_contexts.rb
+++ b/lib/rubocop/rspec/shared_contexts.rb
@@ -38,25 +38,65 @@ RSpec.shared_context 'isolated environment', :isolated_environment do
   end
 end
 
-# `cop_config` must be declared with #let.
-RSpec.shared_context 'config', :config do
-  let(:config) do
-    # Module#<
-    raise '`config` must be used in `describe SomeCopClass do .. end`' unless described_class < RuboCop::Cop::Cop
+# This context assumes nothing and defines `cop`, among others.
+RSpec.shared_context 'config', :config do # rubocop:disable Metrics/BlockLength
+  ### Meant to be overriden at will
 
-    hash = { 'AllCops' => { 'TargetRubyVersion' => ruby_version } }
-    hash['AllCops']['TargetRailsVersion'] = rails_version if rails_version
-    if respond_to?(:cop_config)
-      cop_name = described_class.cop_name
-      hash[cop_name] = RuboCop::ConfigLoader
-                       .default_configuration[cop_name]
-                       .merge('Enabled' => true) # in case it is 'pending'
-                       .merge(cop_config)
+  let(:source) { 'code = {some: :ruby}' }
+
+  let(:cop_class) do
+    if described_class.is_a?(Class) && described_class < RuboCop::Cop::Cop
+      described_class
+    else
+      RuboCop::Cop::Cop
     end
+  end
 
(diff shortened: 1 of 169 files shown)


marcandre:
Or maybe I could put these with the `:config` shared environment instead?

Drenmi:
I think this looks like a great improvement to the spec seen in this PR. 👍

marcandre:
I have changed my PR. I believe this shared context is very flexible and would be a good replacement for `:config`.

spec/rubocop/formatter/clang_style_formatter_spec.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-RSpec.describe RuboCop::Formatter::ClangStyleFormatter do
+RSpec.describe RuboCop::Formatter::ClangStyleFormatter, :cop do
```
bbatsov, line 3: Maybe we can have a separate context for formatters?

bbatsov:
I like the idea overall, but I'm not sure all the cops need so many things - you've picked to apply the suggest approach to two specs that are exactly typical cop specs. Perhaps we can get away with a smaller and more focused context for the actual cops.

lib/rubocop/rspec/shared_contexts.rb
```diff
@@ -60,6 +60,67 @@
   end
 end
 
+# This context assumes nothing and defines `cop`, among others.
+RSpec.shared_context 'cop', :cop do # rubocop:disable Metrics/BlockLength
+  ### Meant to be overriden at will
+
+  let(:source) { 'code = {some: :ruby}' }
+
+  let(:processed_source) { parse_source(source, 'test') }
```
bbatsov, line 69: Do cop tests often touch processed source and the source buffer?

marcandre: Right, I will move `processed_source` in the second section since it is unlikely to be overriden. It is still useful as a shortcust to access the processed source though

spec/rubocop/formatter/clang_style_formatter_spec.rb
marcandre: We could, I haven't checked if there was extra things that could be factorized.

marcandre:
Now that I'm more familiar with the `shared_context`, here's a revised PR.

It streamlines the `:config` shared context by:
1) Adding a `let(:cop)`
2) Removing the logic of `respond_to?(:cop_config)`. This could lead to cases where `let(:cop_config) {{}}` would have specs fail (see #8001).
3) Overrides `autocorrect` default setting (e.g. `JSONLoad`), not just the 'enabled' setting.

I simplified a few other specs. I also removed ~150 identical and redundant `let(:cop)`, so the PR might be a bit noisy. The main commit is the first one.

I'll open an different issue for things I dislike about our current `cop_helper` and `expect_offense`.

spec/rubocop/formatter/clang_style_formatter_spec.rb
marcandre: Didn't see much of interest

bbatsov:
Looks good! Thanks!

PR #7978, opened 2020-05-15
Add option `OnlyWhenUsingAnyOf` to the `Bundler/GemComment` cop.

ric2b:
This PR adds a boolean option `OnlyWhenUsingAnyOf` to the `Bundler/GemComment` cop. 

The goal of this option is to require gem comments only if the gem version has an explicit version specifier or other options like `required/github/etc`, so that the reason for the specifier/options is documented.

I tested it on a project I work on and it worked as expected.

-----------------

* [x] Added tests.


ric2b:
By the way, I'm also planning on doing a similar PR for parameters that change the source of a gem, such as `source:`, `github:`, etc ([documented here](https://bundler.io/v2.0/man/gemfile.5.html#SOURCE)).

I was thinking it could be a bit more general and work by receiving a list of parameter names that should activate the cop, something like: `OnlyIfUsingOptions: ['require', 'github']

Does that make sense to you?

lib/rubocop/cop/bundler/gem_comment.rb
```diff
@@ -5,7 +5,7 @@ module Cop
     module Bundler
       # Add a comment describing each gem in your Gemfile.
       #
-      # @example
+      # @example OnlyIfVersionRestricted: false (default)
```
bbatsov, line 8: I'd also expand the cop summary.

bbatsov:
You should also mention this in the default config. I'm not sure I like the idea of a boolean option, though - might be a better idea to have some styles like "always", "only gems with some extra params". I like your idea overall, but we have to be careful to make the config both easy to understand and extendable.

ric2b:
Yeah, making it extensible is a great point.

What do you think of a configuration with three parameters: 
- AllGems: Boolean
- WithVersionSpecifiers: Boolean
- WithAnyOfTheseOptions: Array

Where `AllGems: true` would ignore the other options and apply to all gems. Otherwise the rest of the options would be combined.

As an example:
```
Bundler/GemComment:
  Enabled: true
  AllGems: false
  WithVersionSpecifiers: true
  WithAnyOfTheseOptions: ['github', 'bitbucket']
```

Would only trigger for a gem that has version restrictions OR uses at least one of `github:` and `bitcucket:`.

bbatsov:
The problem with the boolean flags approach is that some of them are not compatible with one another and this makes the configuration harder. A different approach might be to just have a list of things to check - e.g.:

```
- Check
  - gems_with_version
  - gems_from_source
  - etc
```

And I guess when empty this can mean "all". Or there can be an enforced style "All" and "Subset" and the second one would consult this extra configuration option. Not sure what's the best approach, but I really don't like the boolean flags, so I'd rather us avoid it.

ric2b:
My last suggestion would make the options compatible with each other, it's just that `gems_with_version` would be a separate option because it's not a kwarg. 

But I like your suggestion, it makes the configuration more straightforward, just need to treat version specifiers as a special case 👍

bbatsov:
Fine by me.

ric2b:
@bbatsov The new version has a single new configuration: `OnlyWhenUsingAnyOf`. 

If it's empty the cop behaves the same as before, checking all gems, otherwise it only registers offenses on gems with one or more of the configured options. 

`with_version_specifiers` can also be passed as an option in to check for version specifiers.

Happy to change the names to whatever you prefer, if you don't think they're clear enough or consistent with rubocop conventions.

lib/rubocop/cop/bundler/gem_comment.rb
```diff
@@ -58,6 +92,28 @@ def ignored_gem?(node)
           ignored_gems = Array(cop_config['IgnoredGems'])
           ignored_gems.include?(node.first_argument.value)
         end
+
+        def checked_options_present?(node)
+          cop_config[CHECKED_OPTIONS_CONFIG].include?(VERSION_SPECIFIERS_OPTION) && version_specified_gem?(node) \
+          or contains_checked_options?(node)
```
mwolman, line 98:
```suggestion
          cop_config[CHECKED_OPTIONS_CONFIG].include?(VERSION_SPECIFIERS_OPTION) &&
            version_specified_gem?(node) ||
            contains_checked_options?(node)
```

lib/rubocop/cop/bundler/gem_comment.rb
```diff
@@ -58,6 +92,28 @@ def ignored_gem?(node)
...
+        end
+
+        # Besides the gem name, all other *positional* arguments to `gem` are version specifiers,
+        # as long as it has one we know there's at least one version specifier.
+        def version_specified_gem?(node)
+          # arguments[0] is the gem name
+          node.arguments[1]&.str_type? == true
```
mwolman, line 105:
```suggestion
          node.arguments[1]&.str_type?
```

spec/rubocop/cop/bundler/gem_comment_spec.rb
```diff
@@ -63,5 +64,76 @@
         GEM
       end
     end
+
+    context 'and the OnlyWhenUsingAnyOf option is set' do
```
mwolman, line 68:
```suggestion
    context 'when the "OnlyWhenUsingAnyOf" option is set' do
```

spec/rubocop/cop/bundler/gem_comment_spec.rb
```diff
@@ -63,5 +64,76 @@
...
+      before { cop_config['OnlyWhenUsingAnyOf'] = checked_options }
+
+      context 'and version specifiers are checked' do
```
mwolman, line 71:
```suggestion
      context 'when the version specifiers are checked' do
```

spec/rubocop/cop/bundler/gem_comment_spec.rb
```diff
@@ -63,5 +64,76 @@
...
+          GEM
+        end
+
+        it 'does not register an offense if an uncommented gem has only unchecked options' do
+          expect_no_offenses(<<-GEM, 'Gemfile')
+            gem 'rubocop', group: development
+          GEM
+        end
+      end
+    end
```
mwolman, line 137: What do you think about using more contexts in your tests?
https://rspec.rubystyle.guide/#use-contexts

mwolman: For example this:

```ruby
        it 'registers an offense if an uncommented gem has one of the checked options' do
          expect_offense(<<-GEM, 'Gemfile')
            gem 'rubocop', github: 'some_user/some_fork'
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Missing gem description comment.
          GEM
        end
```

should be something like this:

```ruby
      context 'when an uncommented gem has one of the checked options' do
        it 'registers an offense' do
          expect_offense(<<-GEM, 'Gemfile')
            gem 'rubocop', github: 'some_user/some_fork'
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Missing gem description comment.
          GEM
        end
      end
```

lib/rubocop/cop/bundler/gem_comment.rb
ric2b: I added parenthesis to make it more explicit, I agree that the use of `or` might be confusing with the different order of precedence.

spec/rubocop/cop/bundler/gem_comment_spec.rb
ric2b: Yeah, that looks much better :)

ric2b:
All the tests pass on my machine, the failing test seems completely unrelated to this PR, not sure what to do.

config/default.yml
```diff
@@ -159,6 +159,7 @@ Bundler/GemComment:
     - '**/Gemfile'
     - '**/gems.rb'
   IgnoredGems: []
+  OnlyWhenUsingAnyOf: []
```
bbatsov, line 162: I think this can be shortened to `OnlyFor`.

lib/rubocop/cop/bundler/gem_comment.rb
```diff
@@ -15,17 +21,45 @@ module Bundler
       #   # Helpers for the foo things.
       #   gem 'foo'
       #
+      # @example OnlyWhenUsingAnyOf: ['with_version_specifiers']
```
bbatsov, line 24: I'd name this just `version_specifiers`.

bbatsov:
I think it's a good idea to summarize in docs somewhere all the common params (https://bundler.io/man/gemfile.5.html) for its easier for people to figure out what's available to them.

bbatsov:
You should also update `VersionChanged` to 0.85.

bbatsov:
Thanks!

Issue #7981, opened 2020-05-15
Allow Rails attribute in EmptyLinesAroundAttributeAccessor


legendetm:
## Is your feature request related to a problem? Please describe.

I prefer to group all attributes together including ActiveModel::Attributes#attributes

```
  attr_accessor :foo
  attr_reader :bar
  attr_writer :baz
  attribute :time, :datetime
```

## Describe the solution you'd like

I would expect that `EmptyLinesAroundAttributeAccessor` allow to group `attribute` with other accessors

koic:
This issue similar to #7976. I've opened #7982. Rails-specific methods will be configured by user applications or RuboCop Rails gem using `IgnoredMethods` option because RuboCop core is focused on Ruby.

legendetm:
Yes I know, that RuboCop is Ruby specific and the `IgnoredMethods` option is suitable

koic:
#7981 has been merged. `IgnoredMethods` renamed to `AllowedMethods`. Thank you.

PR #7982, opened 2020-05-16
[Fix #7976] Add ignore options for `Layout/EmptyLinesAroundAttributeAccessor`

### LINKED ISSUE #7976: Allow Layout/EmptyLineAroundAttributeAccessor cop to ignore immediate keyword declarations

Say, I have the following class;
```ruby
class Foo
  attr_reader :alpha, :beta
  private :beta

  attr_accessor :gamma
  alias_method :delta=, :gamma=
end
```
`Layout/EmptyLineAroundAttributeAccessor` cop recommends that my class be refactored into:
```ruby
class Foo
  attr_reader :alpha, :beta

  private :beta

  attr_accessor :gamma

  alias_method :delta=, :gamma=
end
```
which IMHO has poorer readability compared to the original.

A config option to allow having *keywords* such as `private`, `protected`, `public`, `alias`, `alias_method`, etc to immediately follow the `attr*`.
(*N.B.: I don't want to enforce this, just have such occurrences ignored by the cop*)

Exclude the file for the cop in `.rubocop.yml`

koic:
Fixes #7976 and #7981

This PR adds `IgnoreAliasSyntax` and` IgnoredMethods` options for `Layout/EmptyLinesAroundAttributeAccessor` .

Users can configure a behavior for `alias` syntax using `IgnoreAliasSyntax` option. Ignored by default (true).

And users can specify ignored methods using `IgnoredMethods`. By default, set `alias_method`, `public`, `protected`, and `private` provided by Ruby.

`cattr_accessor`, `attribute`, other methods provided by Rails (and other gems) can be set by user applications or RuboCop Rails gem.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -468,6 +468,13 @@ Layout/EmptyLinesAroundAttributeAccessor:
   StyleGuide: '#empty-lines-around-attribute-accessor'
   Enabled: pending
   VersionAdded: '0.83'
+  VersionChanged: '0.84'
+  AllowAliasSyntax: true
+  AllowedMethods:
+    - alias_method
+    - public
+    - protected
+    - private
 
 Layout/EmptyLinesAroundBeginBody:
   Description: "Keeps track of empty lines around begin-end bodies."
--- a/lib/rubocop/cop/layout/empty_lines_around_attribute_accessor.rb
+++ b/lib/rubocop/cop/layout/empty_lines_around_attribute_accessor.rb
@@ -3,7 +3,9 @@
 module RuboCop
   module Cop
     module Layout
-      # Checks for a newline after attribute accessor.
+      # Checks for a newline after an attribute accessor or a group of them.
+      # `alias` syntax and `alias_method`, `public`, `protected`, and `private` methods are allowed by default.
+      # These are customizable with `AllowAliasSyntax` and `AllowedMethods` options.
       #
       # @example
       #   # bad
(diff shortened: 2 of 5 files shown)


config/default.yml
```diff
@@ -468,6 +468,13 @@ Layout/EmptyLinesAroundAttributeAccessor:
   StyleGuide: '#empty-lines-around-attribute-accessor'
   Enabled: pending
   VersionAdded: '0.83'
+  VersionChanged: '1.0'
+  IgnoreAliasSyntax: true
```
bbatsov, line 472: I'd rather use `Allow` and `Allowed` here.

lib/rubocop/cop/layout/empty_lines_around_attribute_accessor.rb
```diff
@@ -4,6 +4,8 @@ module RuboCop
   module Cop
     module Layout
       # Checks for a newline after attribute accessor.
+      # `alias` syntax and `alias_method`, `public`, `protected`, and `private` methods are ignored by default.
```
bbatsov, line 7: We should also be more specific in the initial sentence - it's an attribute accessor or a group of them.

bbatsov:
Good work! Probably at some point we should have something more generic for other macros that are typically grouped together.

koic:
Yeah, I update this PR.

config/default.yml
```diff
@@ -468,6 +468,13 @@ Layout/EmptyLinesAroundAttributeAccessor:
   StyleGuide: '#empty-lines-around-attribute-accessor'
   Enabled: pending
   VersionAdded: '0.83'
+  VersionChanged: '1.0'
```
bbatsov, line 471: We should also change this to 0.84.

koic: Yeah, I update it!

ashmaroli:
Thank you for the quick turnaround @koic and @bbatsov :)

PR #7983, opened 2020-05-16
Make the config loader Bundler-aware

CvX:
Before this change, when inheriting a configuration from a gem, rubocop:

1. would fail to find it, if the gem specified in the Gemfile has a `path` or `git` option (because rubygems is unaware of those gems)
2. would always load the config of the latest installed version of a gem, regardless of the version specified in the Gemfile

This commit changes that, so if executed in the context of Bundler (`defined?(Bundler)`) it will first try to get the gem path via Bundler, and fallback to the previous `Gem::Specification` path when not found.

This solves the two mentioned issues while maintaining the possibility to run rubocop in a globally installed context, i.e. without Bundler.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * [#7886](https://github.com/rubocop-hq/rubocop/issues/7886): Fix a bug in `AllowComments` logic in `Lint/SuppressedException`. ([@jonas054][])
 * [#7991](https://github.com/rubocop-hq/rubocop/issues/7991): Fix an error for `Layout/EmptyLinesAroundAttributeAccessor` when attribute method is method chained. ([@koic][])
 * [#7993](https://github.com/rubocop-hq/rubocop/issues/7993): Fix a false positive for `Migration/DepartmentName` when a disable comment contains an unexpected character for department name. ([@koic][])
+* [#7983](https://github.com/rubocop-hq/rubocop/pull/7983): Make the config loader Bundler-aware. ([@CvX][])
 
 ### Changes
 
@@ -4538,3 +4539,4 @@
 [@jeffcarbs]: https://github.com/jeffcarbs
 [@laurmurclar]: https://github.com/laurmurclar
 [@jethrodaniel]: https://github.com/jethrodaniel
+[@CvX]: https://github.com/CvX
--- a/lib/rubocop/config_loader_resolver.rb
+++ b/lib/rubocop/config_loader_resolver.rb
@@ -208,8 +208,14 @@ module RuboCop
     end
 
     def gem_config_path(gem_name, relative_config_path)
-      spec = Gem::Specification.find_by_name(gem_name)
-      File.join(spec.gem_dir, relative_config_path)
+      if defined?(Bundler)
+        gem = Bundler.load.specs[gem_name].first
+        gem_path = gem.full_gem_path if gem
+      end
+
+      gem_path ||= Gem::Specification.find_by_name(gem_name).gem_dir
+
+      File.join(gem_path, relative_config_path)
     rescue Gem::LoadError => e
       raise Gem::LoadError,
             "Unable to find gem #{gem_name}; is the gem installed? #{e}"
--- a/spec/rubocop/config_loader_spec.rb
+++ b/spec/rubocop/config_loader_spec.rb
@@ -823,31 +823,65 @@ RSpec.describe RuboCop::ConfigLoader do
         YAML
       end
 
-      it 'returns values from the gem config with local overrides' do
-        gem_class = Struct.new(:gem_dir)
-        %w[gemone gemtwo].each do |gem_name|
-          mock_spec = gem_class.new(File.join(gem_root, gem_name))
-          allow(Gem::Specification).to receive(:find_by_name)
-            .with(gem_name).and_return(mock_spec)
+      context 'and the gem is globally installed' do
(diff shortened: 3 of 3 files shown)


bbatsov:
The suggested change seems reasonable to me. @jonas054 @koic What do you think?

CvX:
I can think of a couple open questions that should be answered before merging:

1. While I added it to the changelog as a fix, should it be considered a (breaking) change? I can think of a situation where a rubocop user has a Gemfile with an outdated version a gem that's used to inherit the config from, and uses `bundle exec rubocop` to run the linting, but unknowingly has been using a different (globally installed) version of that gem. After this change, rubocop will start picking the correct version.
2. I don't know how much of a private or public API `Bundler.load.specs[gem_name].first.full_gem_path` is. Also, this API might have been different in the older versions of Bundler. Should we have an escape hatch (`rescue`?) in case it blows up? Should we test the compatibility with different versions? (1.x, 2.x, master?) What about the cases where rubocop is run in a context of some ancient (i.e. pre 1.0) versions of Bundler?

bbatsov:
Good points. 

1. I don't think many users would be impacted by this, but we should probably highlight the potential for some issues in the changelog.
2. If you've stuck on an ancient Bundler probably recent RuboCop's won't work for you either. :-) I guess if this works on 1.x and 2.x we should be fine.

koic:
This proposal looks good to me.

OTOH, I'm worried about it doesn't work with Bundler 1.2.0(Bundler's patch version is not investigated).
It may be better to handle `Bundler::GemfileError` and display a message prompting Bundler upgrade.

## Bundler 2.1.4

```
% ruby -rbundler -e "p RUBY_VERSION; p Bundler::VERSION; p Bundler.load.specs['net-telnet'].first.full_gem_path"
"2.4.10"
"2.1.4"
"/Users/koic/.rbenv/versions/2.4.10/lib/ruby/gems/2.4.0/gems/net-telnet-0.2.0"
```

## Bundler 1.3.0

```
% ruby -rbundler -e "p RUBY_VERSION; p Bundler::VERSION; p Bundler.load.specs['net-telnet'].first.full_gem_path"
"2.4.10"
"1.3.0"
"/Users/koic/.rbenv/versions/2.4.10/lib/ruby/gems/2.4.0/gems/net-telnet-0.2.0"
```

## Bundler 1.2.0

```
% ruby -rbundler -e "p RUBY_VERSION; p Bundler::VERSION; p Bundler.load.specs['net-telnet'].first.full_gem_path"
"2.4.10"
"1.2.0"
/Users/koic/.rbenv/versions/2.4.10/lib/ruby/gems/2.4.0/gems/bundler-1.2.0/lib/bundler/dsl.rb:12:in `rescue in evaluate': There was an error in your Gemfile, and Bundler cannot continue. (Bundler::GemfileError)
        from /Users/koic/.rbenv/versions/2.4.10/lib/ruby/gems/2.4.0/gems/bundler-1.2.0/lib/bundler/dsl.rb:5:in `evaluate'
        from /Users/koic/.rbenv/versions/2.4.10/lib/ruby/gems/2.4.0/gems/bundler-1.2.0/lib/bundler/definition.rb:18:in `build'
        from /Users/koic/.rbenv/versions/2.4.10/lib/ruby/gems/2.4.0/gems/bundler-1.2.0/lib/bundler.rb:144:in `definition'
        from /Users/koic/.rbenv/versions/2.4.10/lib/ruby/gems/2.4.0/gems/bundler-1.2.0/lib/bundler.rb:132:in `load'
        from -e:1:in `<main>'
```

bbatsov:
I like this suggestion!

CvX:
> OTOH, I'm worried about it doesn't work with Bundler 1.2.0

@koic It does work! 😃 The Gemfile you ran your test on apparently isn't compatible with Bundler 1.2.0. When everything aligns just right (ruby version, Bundler version, and Gemfile syntax):

```
ruby -rbundler -e "p RUBY_VERSION; p Bundler::VERSION; p Bundler.load.specs['18n'].first.full_gem_path"
"1.9.3"
"1.2.0"
"/Users/cvx/.rbenv/versions/1.9.3-p551/lib/ruby/gems/1.9.1/gems/i18n-0.9.5"
```

I don't think catching `Bundler::GemfileError` would be of any use, since it has less to do with rubocop's bundler usage and is more of a user error. If you tried to run `bundle exec rubocop` with an invalid config it would fail before reaching the config resolver.

> I don't think many users would be impacted by this, but we should probably highlight the potential for some issues in the changelog

@bbatsov Do you have any suggestion how to succinctly explain these potential edge-cases? 🙂

bbatsov:
> @bbatsov Do you have any suggestion how to succinctly explain these potential edge-cases? 🙂

Something like "Require Bundler X or newer" perhaps?

CvX:
> "Require Bundler X or newer"

This PR doesn't _require_ Bundler, but if we're talking version requirement of an _optional_ integration that it enables – I've tried the oldest version that's mentioned in [bundler's changelog](https://github.com/rubygems/rubygems/blob/8abdabeff19cf563d1b4b60edcfdd223afb849dc/bundler/CHANGELOG.md#095-february-12-2010) (0.9.5 from 2010) that snippet still worked, so I'm now even more confident that we're safe on that front. 😃

bbatsov:
No need to fret over this further then. :-)

PR #7984, opened 2020-05-16
Check commit rake task

marcandre:
I typically want to run `rubocop` and `rspec` on the last commit (or few last commits), so I thought maybe this rake task could be useful to others?

If not, I'll use it for myself and simply add something like `tasks/local` in `.gitignore`


bbatsov:
It seems useful to me, but I less we document it somewhere in dev docs I doubt someone would ever find and use it. The dev docs are pretty light right now - just one page about writing new cops that leaves a lot to be desired (I noticed its poor state only today when I saw the docs weren't rebuilding after the extraction of `rubocop-ast` - the config needed to be updated).

marcandre:
Added a mention in the docs and in the ChangeLog

ansonhoyt:
This is a neat idea. Just trying out `0.84.0` and saw this in the changelog, but I don't have this task. Nothing from `rake -T check_commit`.

How should I call it?

marcandre:
I'll amend the Changelog to state that this is available only when developing from the main repo...

If you want to use something similar for your own projects, you can inspire yourself from `https://github.com/rubocop-hq/rubocop/blob/master/tasks/check_commit.rake`, you'll need to `require 'rubocop/rake_task'` too.

ansonhoyt:
Ahh, that makes sense. Thanks for explaining!

Issue #7833, opened 2020-03-31
Revising RuboCop's defaults (survey)


bbatsov:
We're running a [short survey](https://docs.google.com/forms/d/e/1FAIpQLSe3uTB_E164hgGUaM5-DFuIVBgv7l18ub2lAU5gXmw1dw4vPQ/viewform?usp=sf_link) on RuboCop's default prior to releasing version 1.0. 

You can find more details [here](https://metaredux.com/posts/2020/03/27/adjusting-rubocop-s-defaults.html). Thanks in advance for taking the time to share your input with us!

lamont-granquist:
`I use trailing commas in collection literals (e.g. [1, 2, 3,])`

If you meant multiline you should use that word there.  I doubt anyone uses trailing commas in single line, so this question will undercount people who want trailing multiline commas.

bbatsov:
Okay, I'll update the wording there.

grncdr:
An idea regarding the overall process: what would you think about collecting `.rubocop.yml` configs and aggregating stats for which defaults are most often overridden?

grncdr:
I would maybe volunteer to write such a tool if that's of interest to the maintainers.

bbatsov:
@grncdr I think that'd be a super useful tool for us!

januszm:
I think the question about Trailing Commas in Collection Literals was poorly phrased in the survey. The way in which the multi-line variant is presented

```ruby
{ one: 1,
  two: 2, }
```

does not show the advantages of adding this comma. If it were presented like this:

```ruby
{
  one: 1,
  two: 2,
}
```

the result of the poll could vary considerably. Teams using Git / Github should appreciate that adding a trailing comma in the example above will make Git diff more readable and easier to use. This issue was raised in the comments section at https://metaredux.com/posts/2020/05/26/rubocop-defaults-survey-results.html

@bbatsov please have a second look and consider making the trailing comma `only in multi-line literals` default in 1.0 (maybe another survey?)

marcandre:
I'm sorry I missed the survey.
I ❤️ trailing commas too. I use them in all my personal projects. I'm genuinely upset at JSON's creator for not allowing them.

bbatsov:
@januszm I get your point, but I doubt this affected the results significantly. Formatting code in Google Forms is not fun. :-) There won't be any more changes to the defaults before RuboCop 1.0, as I have already committed to this to all of our users.

PR #7985, opened 2020-05-17
Add `EnforcedStyle` for `Style/DoubleNegation` cop

koic:
This PR adds `EnforcedStyle` option for `Style/DoubleNegation` cop, and `EnforcedStyle: context` by default based on #7833 survey.

```ruby
# @example EnforcedStyle: context (default)
#   # good
#   def foo?
#     !!return_value
#   end
#
# @example EnforcedStyle: always
#   # bad
#   def foo?
#     !!return_value
#   end
```

Ruby will return the last evaluated value.

This PR only supports simple contexts, such as the end of line value of method, and when using `return`.
A complex contexts should also be supported, such as when branching with `if` and others in the future.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2699,6 +2699,11 @@ Style/DoubleNegation:
   StyleGuide: '#no-bang-bang'
   Enabled: true
   VersionAdded: '0.19'
+  VersionChanged: '0.84'
+  EnforcedStyle: allowed_in_returns
+  SupportedStyles:
+    - allowed_in_returns
+    - forbidden
 
 Style/EachForSimpleLoop:
   Description: >-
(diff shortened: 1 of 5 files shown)


config/default.yml
```diff
@@ -2699,6 +2699,11 @@ Style/DoubleNegation:
   StyleGuide: '#no-bang-bang'
   Enabled: true
   VersionAdded: '0.19'
+  VersionChanged: '1.0'
```
bbatsov, line 2702: Make this 0.84, please. I'll cut one more release before 1.0 to make sure the `rubocop-ast` separation didn't cause problems.

config/default.yml
```diff
@@ -2699,6 +2699,11 @@ Style/DoubleNegation:
...
+  EnforcedStyle: context
+  SupportedStyles:
+    - context
```
bbatsov, line 2705: I think better names would be `forbidden` and `allowed_in_returns` or something along those lines.

bbatsov: Or `allowed_in_boolean_context`?

bbatsov:
The cop description should also be updated, as it currently pertains only to the `forbidden` style.

config/default.yml
koic: Thank you for suggesting a good name! I think `allowed_in_returns ` is clear and straightforward!

PR #7989, opened 2020-05-17
Set 100 as the internal Layout/LineLength Max

jonas054:
_This is an opinion piece. A plea to @rubocop-hq/rubocop-core, and anybody else who want to join the discussion, to move back a bit from the recent change of line length -- internally in RuboCop source code -- while it's still pretty easy to do so. You might not agree. I'm prepared for that, but I want to have the discussion now, rather than not say anything._

-----------------------------------

The decision to change the default `Max` value for `Layout/LineLength` was based on a survey. People seem to prefer 120 as the limit for line length, and that's fine. Let the people have the line length they prefer. But it's not necessarily a good thing to immediately adopt the same configuration setting internally in RuboCop. There's the "eat your own dog food" argument, but one might also argue that if we wanted 120 character lines in RuboCop we would have changed the internal setting earlier.

To me, the jump from 80 to 120 is too big. A bit of a shock actually. It pains me to not be able to work with a vertically split screen and read the source code without line wrapping occurring. I suggest 100 characters as a middle ground. Following the RuboCop user community in the move to longer lines, but not going all the way.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -69,6 +69,9 @@ Layout/HashAlignment:
     - key
     - table
 
+Layout/LineLength:
+  Max: 100
+
 Lint/InterpolationCheck:
   Exclude:
     - 'spec/**/*.rb'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -71,6 +71,8 @@ Layout/HashAlignment:
 
 Layout/LineLength:
   Max: 100
+  IgnoredPatterns:
+    - !ruby/regexp /\A +(it|describe|context|shared_examples|include_examples|it_behaves_like) ["']/
 
 Lint/InterpolationCheck:
   Exclude:
--- a/lib/rubocop/config.rb
+++ b/lib/rubocop/config.rb
@@ -86,7 +86,9 @@ module RuboCop
 
       excludes = for_all_cops['Exclude'] ||= []
       highest_config.for_all_cops['Exclude'].each do |path|
-        path = File.join(File.dirname(highest_config.loaded_path), path) unless path.is_a?(Regexp) || absolute?(path)
+        unless path.is_a?(Regexp) || absolute?(path)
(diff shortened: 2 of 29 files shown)


bbatsov:
That's fine by me. On my main display I fit 120 code buffers side by side (but on my ultrabook I feel your pain), but I never had issues with 80 as the line length, so 100 or 120 doesn't matter much. I did notice, however, that 120-wide diffs require a bit of scrolling in GitHub, so I guess they didn't optimize for it either. :-) Let's see what the others have to say as well.

bbatsov:
Just one more thing - in specs we might want to keep it at 120, as otherwise we either have to split some example descriptions or write more terse descriptions. The issues around those descriptions were the primary reason I felt like we needed to increase the line length. Anyways, it seems that at 100 only a handful of examples are affected, so probably not a big deal.

koic:
LineLength 120 is comfortable to me because I don't split editor vertically. And I've freed from the non-trivial pain of naming and layout by extending LineLength from 80 to 120.

However, I think core maintainers of RuboCop repository needs to be comfortable.
So, probably LineLength 100 is better than 80 and its a point of compromise. (I really like 120, but it's not important for the maintenance of this repository 😅)

I highly respect @jonas054's great maintenance activity! That is the most important to me ❤️

jonas054:
@bbatsov How about no limit for example descriptions?

```yaml
Layout/LineLength:
  Max: 100
  IgnoredPatterns:
    - !ruby/regexp /^ +(it|describe|context|shared_examples|include_examples|it_behaves_like) ["']/
```

bbatsov:
I think that's a good idea!

bbatsov:
@jonas054 Feel free to merge this after addressing the outstanding conflict.

jonas054:
Thanks @bbatsov and @koic for being so understanding. Much appreciated.

bbatsov:
@jonas054 I even added a `.dir-locals.el` for us. ;-)

PR #7990, opened 2020-05-18
Fix resolving `inherit_gem` in remote configs

CvX:
When a file inherits from a url inheriting from a gem, like so:

```yaml
# .rubocop.yml
inherit_from: http://example.com/default.yml
```

```yaml
# http://example.com/default.yml
inherit_gem:
    somegem: default.yml
```

```yaml
# somegem's default.yml
Layout/LineLength:
    Max: 48
```

Rubocop would then incorrectly try to resolve a URL like this:

`http://example.com//private/var/folders/69/bjrqv8v1225f7yqnxknm3fb80000gn/T/d20200518-94245-6dqeew/work/gems/somegem/default.yml`

With this change, local gem paths are no longer taken by mistake for relative remote paths.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/config_loader_resolver.rb
+++ b/lib/rubocop/config_loader_resolver.rb
@@ -171,10 +171,20 @@ module RuboCop
 
     def inherited_file(path, inherit_from, file)
       if remote_file?(inherit_from)
+        # A remote configuration, e.g. `inherit_from: http://example.com/rubocop.yml`.
         RemoteConfig.new(inherit_from, File.dirname(path))
+      elsif Pathname.new(inherit_from).absolute?
+        # An absolute path to a config, e.g. `inherit_from: /Users/me/rubocop.yml`.
+        # The path may come from `inherit_gem` option, where a gem name is expanded
+        # to an absolute path to that gem.
+        print 'Inheriting ' if ConfigLoader.debug?
+        inherit_from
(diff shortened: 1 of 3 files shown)


lib/rubocop/config_loader_resolver.rb
```diff
@@ -172,6 +172,9 @@ def base_configs(path, inherit_from, file)
     def inherited_file(path, inherit_from, file)
       if remote_file?(inherit_from)
         RemoteConfig.new(inherit_from, File.dirname(path))
+      elsif Pathname.new(inherit_from).absolute?
```
bbatsov, line 175: I think we should probably add some comments for each branch of the `if`, otherwise it's getting really hard to follow which case maps to what exactly.

CvX: @bbatsov I've added the comments. Let me know if you see a way to improve them further.

Btw, the 2nd and 4th branch could be merged into `Pathname.new(inherit_from).absolute? || !file.is_a?(RemoteConfig)` but I think the way it's currently written is a bit more clear. And avoids unnecessary `File.expand_path` and `File.dirname` in the case of absolute paths.

bbatsov: Agreed.

bbatsov:
Thanks!

PR #7995, opened 2020-05-19
[Fix #7993] Fix a false positive for `Migration/DepartmentName`

### LINKED ISSUE #7993: Unexpected DepartmentMissing in explanatory comments
I am not sure on which version it started (recently, something around ~80), but now I get a lot of warnings on lines looking like this:

```ruby
# rubocop:disable SomeDepartment/SomeCop -- because something, something, and something
```
Seems like `Migration/DepartmentName` cop treats _every_ comma as a sign to look for new cop... And becomes disappointed. Real example (at 0.83.0):

```
spec/factories/revision_data.rb:65:52: C: Migration/DepartmentName: Department name is missing.
    # rubocop:disable Lint/PercentStringArray -- no, silly, we need those ","!
                                                     ^^^^^
```

As there are no other warnings (like "unknown cop"), I believe the problem is just in `Migration/DepartmentName` cop.

koic:
Fixes #7993.

This PR fixes a false positive for `Migration/DepartmentName` cop when a disable comment contains an unexpected character for department name.

-----------------

* [x] Added tests.


bbatsov:
Probably we should also be more explicit in the docs that users should not add anything besides the name. Maybe even register some offence in such case.

koic:
Sure. I've mentioned the issue to manual/configuration.md. And we may be able to add a new cop for disabling comment style If RuboCop expects to make this strict.

bbatsov:
Thanks!

PR #8009, opened 2020-05-21
Amend Changelog [doc] [ci skip]

marcandre:
There's no way to push directly to `master`, even for [doc] only commits, right?


marcandre:
It's actually worse, coz `[ci skip]` only commits don't even pass the CI 😿, waits indefinitely for Circle-ci

Issue #7999, opened 2020-05-20
Lint/EmptyWhen false positive with inline comments.


marcandre:
0) Proper bug: the following will detect an offense even with `'AllowComments' => true`

```
case foo
when :bar then # do nothing
end
```

Note that the following is rightfully accepted
```
case foo
when :bar
  # do nothing
end
```

The spec file has multiple issues and is in need of love:
1) Explicit testing of `AllowComments` is minimal, should be expanded
2) Doesn't use `except_offense`/`no_offense` so spec failures are next to impossible to locate.
3) The default is `AllowComment: true`, but the `:config` spec will not load the defaults unless given there is a `cop_config` method. Add `let(:cop_config) { {} }` to the current spec and it fails (related to #7970)

Darhazer:
Should 
```
case foo
when :bar # do nothing
end
```
also be accepted?

PR #8017, opened 2020-05-23
Fix a false positive for `Lint/SuppressedException`

koic:
This PR fixes a false positive for `Lint/SuppressedException` when empty rescue with comment in `def`.

```console
% cat example.rb
def foo
  do_something
rescue
  # noop
end

% bundle exec rubocop --only Lint/SuppressedException
(snip)

Inspecting 1 file
W

Offenses:

example.rb:3:1: W: Lint/SuppressedException: Do not suppress exceptions.
rescue
^^^^^^

```

The above code should not be warned because `AllowComments: true` by defualt.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/suppressed_exception_spec.rb
+++ b/spec/rubocop/cop/lint/suppressed_exception_spec.rb
@@ -25,6 +25,29 @@ RSpec.describe RuboCop::Cop::Lint::SuppressedException, :config do
         end
       RUBY
     end
+
+    context 'when empty rescue for `def`' do
+      it 'registers an offense for empty rescue without comment' do
+        expect_offense(<<~RUBY)
+          def foo
+            do_something
+          rescue
+          ^^^^^^ Do not suppress exceptions.
+          end
+        RUBY
+      end
+
+      it 'registers an offense for empty rescue with comment' do
+        expect_offense(<<~RUBY)
+          def foo
+            do_something
+          rescue
+          ^^^^^^ Do not suppress exceptions.
+            # do nothing
+          end
+        RUBY
+      end
+    end
   end
 
   context 'with AllowComments set to true' do
@@ -41,14 +64,26 @@ RSpec.describe RuboCop::Cop::Lint::SuppressedException, :config do
       RUBY
     end
 
-    it 'registers an offense for empty rescue block in `def`' do
-      expect_offense(<<~RUBY)
-        def foo
-          do_something
-        rescue
-        ^^^^^^ Do not suppress exceptions.
-        end
-      RUBY
+    context 'when empty rescue for `def`' do
+      it 'registers an offense for empty rescue without comment' do
+        expect_offense(<<~RUBY)
+          def foo
+            do_something
+          rescue
+          ^^^^^^ Do not suppress exceptions.
+          end
+        RUBY
+      end
+
+      it 'does not register an offense for empty rescue with comment' do
+        expect_no_offenses(<<~RUBY)
+          def foo
+            do_something
+          rescue
+            # do nothing
+          end
+        RUBY
+      end
     end
 
     it 'registers an offense for empty rescue on single line with a comment after it' do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#8008](https://github.com/rubocop-hq/rubocop/issues/8008): Fix an error for `Lint/SuppressedException` when empty rescue block in `def`. ([@koic][])
 * [#8012](https://github.com/rubocop-hq/rubocop/issues/8012): Fix an incorrect autocorrect for `Lint/DeprecatedOpenSSLConstant` when deprecated OpenSSL constant is used in a block. ([@koic][])
+* [#8017](https://github.com/rubocop-hq/rubocop/pull/8017): Fix a false positive for `Lint/SuppressedException` when empty rescue with comment in `def`. ([@koic][])
 
 ## 0.84.0 (2020-05-21)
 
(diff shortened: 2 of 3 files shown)


spec/rubocop/cop/lint/suppressed_exception_spec.rb
```diff
@@ -51,6 +72,16 @@ def foo
...
 
+    it 'does not register an offense for empty rescue with comment in `def`' do
+      expect_no_offenses(<<~RUBY)
+        def foo
+          do_something
+        rescue
+          # do nothing
+        end
+      RUBY
+    end
```
mwolman, line 83:
```suggestion
    context 'with comment in `def`' do
      it 'does not register an offense for empty rescue' do
        expect_no_offenses(<<~RUBY)
          def foo
            do_something
          rescue
            # do nothing
          end
        RUBY
      end
    end
```

spec/rubocop/cop/lint/suppressed_exception_spec.rb
```diff
@@ -25,6 +25,27 @@
...
+
+    it 'registers an offense for empty rescue with comment in `def`' do
+      expect_offense(<<~RUBY)
+        def foo
+          do_something
+        rescue
+        ^^^^^^ Do not suppress exceptions.
+        end
+      RUBY
+    end
```
mwolman, line 48:
```suggestion
    context 'with comment in `def`' do
      it 'does not register an offense for empty rescue' do
      expect_offense(<<~RUBY)
        def foo
          do_something
        rescue
        ^^^^^^ Do not suppress exceptions.
        end
      RUBY
      end
    end
```

spec/rubocop/cop/lint/suppressed_exception_spec.rb
```diff
@@ -25,6 +25,27 @@
...
+    it 'registers an offense for co rescue block in `def`' do
+      expect_offense(<<~RUBY)
+        def foo
+          do_something
+        rescue
+        ^^^^^^ Do not suppress exceptions.
+          # do nothing
+        end
+      RUBY
+    end
```
mwolman, line 38:
```suggestion
    context 'without comment in `def`' do
      it 'registers an offense for co rescue block' do
        expect_offense(<<~RUBY)
          def foo
            do_something
          rescue
          ^^^^^^ Do not suppress exceptions.
            # do nothing
          end
        RUBY
      end
    end
```

marcandre:
Looking forward to this 🎉
Can you make sure that inline comments are also ok? (See #7999) E.g.:
```
def foo
  bar
rescue # ignore
end
```

spec/rubocop/cop/lint/suppressed_exception_spec.rb
```diff
@@ -25,6 +25,29 @@
...
+
+      it 'does not register an offense for empty rescue' do
+        expect_offense(<<~RUBY)
+          def foo
+            do_something
+          rescue
+          ^^^^^^ Do not suppress exceptions.
+          end
+        RUBY
+      end
```
mwolman, line 49: In this case, you don't have a comment, so this should be inside a different context with the label:
`without comment in def` instead of with

My mistake when doing the suggestion 🙏🏻 sorry

koic:
Certainly! Also the following code seems to be able to do the same.

```ruby
begin
  bar
rescue # ignore
end
```

I think that these cases can be extended separately from this PR.

spec/rubocop/cop/lint/suppressed_exception_spec.rb
koic: Ah, I was wrong 💦 Thank you for your review!

spec/rubocop/cop/lint/suppressed_exception_spec.rb
koic: > context 'with comment in `def`' do

Since it's a comment for `rescue` instead of `def`, I adjusted the context a bit :-)

spec/rubocop/cop/lint/suppressed_exception_spec.rb
```diff
@@ -25,6 +25,29 @@
...
+      it 'registers an offense for empty rescue with comment' do
+        expect_offense(<<~RUBY)
+          def foo
+            do_something
+          rescue
+          ^^^^^^ Do not suppress exceptions.
+            # do nothing
+          end
+        RUBY
+      end
```
mwolman, line 49:
```suggestion
      context 'without comments'
        it 'registers an offense' do
          expect_offense(<<~RUBY)
            def foo
              do_something
            rescue
            ^^^^^^ Do not suppress exceptions.
            end
          RUBY
        end
      end

      context 'with a comment' do
        it 'registers an offense' do
          expect_offense(<<~RUBY)
            def foo
              do_something
            rescue
            ^^^^^^ Do not suppress exceptions.
              # do nothing
            end
          RUBY
        end
      end
```

mwolman: https://rspec.rubystyle.guide/#context-descriptions

mwolman: https://rspec.rubystyle.guide/#example-descriptions here you can see that is a good practice to never end `it` with a conditional

mwolman: Same comments apply to the next block of tests 🤗

koic: Good suggestion. However, many existing tests often include a context in `it` in this repository. There is no problem in maintenance, so I leave it as it is. Thank you for your review, again 😃

mwolman: Ok. You're welcome 🤗

PR #8023, opened 2020-05-24
Lazy enlisting

pirj:
There's a problem in specs ...
I wouldn't be surprised if you stopped reading after this point, since it's usually not a worthwhile justification.
But please take a moment and bear with me.

If we define a cop for testing purposes like that:
```ruby
module Dept
  class MyCop < RuboCop::Cop::Cop
    # ...
  end
end
```
you'll end up with a constant littering the root namespace.
https://rspec.rubystyle.guide/#declare-constants
https://fili.pp.ru/leaky-constants.html

It's not a big deal if you pick unique class and namespace each time or are brave enough to deal with potential side effects.

The following is arguably better:
```ruby
klass =
  Class.new(RuboCop::Cop::Cop) do
    def external_dependency_checksum
      'something other than nil'
    end
  end
stub_const('Dept::MyCop', klass)
```

But there's a problem with this approach:

```
  1) RuboCop::Cop::Team#external_dependency_checksum when cop with different checksum joins has a different checksum for the whole team
     Failure/Error: new(*class_name.split('::').last(2))

     NoMethodError:
       undefined method `split' for nil:NilClass
     # ./lib/rubocop/cop/badge.rb:26:in `for'
     # ./lib/rubocop/cop/cop.rb:66:in `badge'
     # ./lib/rubocop/cop/registry.rb:35:in `enlist'
     # ./lib/rubocop/cop/cop.rb:62:in `inherited'
     # ./spec/rubocop/cop/team_spec.rb:372:in `initialize'
     # ./spec/rubocop/cop/team_spec.rb:372:in `new'
     # ./spec/rubocop/cop/team_spec.rb:372:in `block (4 levels) in <top (required)>'
```

Backtrace explains it all, but in a nutshell there's a pesky `def self.included` in `Cop` that enlists the newly defined cop for a department. And to get the department, it needs a name. Which is `nil` for `Class.new`.

Ha! Easy-peasy you think: define a class method `name` on that cop.
Not so easy it turns out:

```ruby
class A
  def self.inherited(base)
    fail 'Ooopsie!' unless base.name
  end
end

class B < A # ok
end

Class.new do # boom!
  def self.name
    'Multipassport!'
  end
end
```

And guess what? `self.inherited` is called **before** the block is evaluated.
`Class.new` to blame you would say!
But it's not. Here's another example for you.

```ruby
class A
  def self.inherited(base)
    fail 'Traitor!' unless base.surname
  end
end

class B < A
  def self.surname
    'Schtirlitz'
  end
end
```

```
NoMethodError: undefined method `surname' for B:Class
```

Srsly? WTH, Ruby? They have to quack, but some of them are real and have to have names and other things!

Ok, if it's not `Class.new` to blame, it's a premature `inherited` hook. What do we do?

Lazy enlist to the rescue!

This started off as https://github.com/rubocop-hq/rubocop/pull/8018#discussion_r429582159

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/registry.rb
+++ b/lib/rubocop/cop/registry.rb
@@ -23,34 +23,28 @@ module RuboCop
     # Registry that tracks all cops by their badge and department.
     class Registry
       def initialize(cops = [], options = {})
-        @registry = {}
-        @departments = {}
-        @cops_by_cop_name = Hash.new { |hash, key| hash[key] = [] }
-
+        @enlisted = []
         cops.each { |cop| enlist(cop) }
         @options = options
       end
 
       def enlist(cop)
-        @registry[cop.badge] = cop
-        @departments[cop.department] ||= []
-        @departments[cop.department] << cop
-        @cops_by_cop_name[cop.cop_name] << cop
+        @enlisted << cop
       end
 
       # @return [Array<Symbol>] list of departments for current cops.
       def departments
-        @departments.keys
+        staff.keys
       end
 
       # @return [Registry] Cops for that specific department.
       def with_department(department)
-        with(@departments.fetch(department, []))
+        with(staff.fetch(department, []))
       end
 
       # @return [Registry] Cops not for a specific department.
       def without_department(department)
-        without_department = @departments.dup
+        without_department = staff.dup
         without_department.delete(department)
 
         with(without_department.values.flatten)
@@ -120,21 +114,21 @@ module RuboCop
 
       def unqualified_cop_names
         @unqualified_cop_names ||=
-          Set.new(@cops_by_cop_name.keys.map { |qn| File.basename(qn) }) <<
+          Set.new(cops_by_cop_name.keys.map { |qn| File.basename(qn) }) <<
           'RedundantCopDisableDirective'
       end
 
       # @return [Hash{String => Array<Class>}]
       def to_h
-        @cops_by_cop_name
+        cops_by_cop_name
       end
 
       def cops
-        @registry.values
+        registry.values
       end
 
       def length
-        @registry.size
+        registry.size
       end
 
       def enabled(config, only, only_safe = false)
@@ -172,7 +166,7 @@ module RuboCop
       end
 
       def sort!
-        @registry = Hash[@registry.sort_by { |badge, _| badge.cop_name }]
+        @registry = Hash[registry.sort_by { |badge, _| badge.cop_name }]
 
         self
       end
@@ -188,17 +182,45 @@ module RuboCop
       # @param [String] cop_name
       # @return [Class, nil]
       def find_by_cop_name(cop_name)
-        @cops_by_cop_name[cop_name].first
+        cops_by_cop_name[cop_name].first
+      end
+
+      protected
+
+      def cops_by_cop_name
+        @cops_by_cop_name || call_for_serve_and_protect!.cops_by_cop_name
+      end
+
+      def registry
+        @registry || call_for_serve_and_protect!.registry
+      end
+
+      def staff
+        @staff || call_for_serve_and_protect!.staff
       end
 
       private
+
+      def call_for_serve_and_protect!
+        @registry = {}
+        @staff = Hash.new { |hash, key| hash[key] = [] }
+        @cops_by_cop_name = Hash.new { |hash, key| hash[key] = [] }
+
+        @enlisted.each do |cop|
+          @registry[cop.badge] = cop
+          @staff[cop.department] << cop
+          @cops_by_cop_name[cop.cop_name] << cop
+        end
+
+        self
+      end
 
       def with(cops)
         self.class.new(cops)
       end
 
       def qualify_badge(badge)
-        @departments
+        staff
           .map { |department, _| badge.with_department(department) }
           .select { |potential_badge| registered?(potential_badge) }
       end
@@ -214,7 +236,7 @@ module RuboCop
       end
 
       def registered?(badge)
-        @registry.key?(badge)
+        registry.key?(badge)
       end
     end
   end
--- a/lib/rubocop/rspec/support.rb
+++ b/lib/rubocop/rspec/support.rb
@@ -4,10 +4,13 @@
 
 require_relative 'cop_helper'
 require_relative 'host_environment_simulation_helper'
-require_relative 'shared_contexts'
 require_relative 'expect_offense'
+require_relative 'shared_contexts'
+require_relative 'stub_cop'
 
 RSpec.configure do |config|
   config.include CopHelper
+  config.include RuboCop::RSpec::ExpectOffense
+  config.include RuboCop::RSpec::StubCop
   config.include HostEnvironmentSimulatorHelper
 end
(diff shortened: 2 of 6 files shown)


pirj:
This innocent seemingly self-contained change knocks down specs all around RuboCop.
I would kindly appreciate any help and guidance on how to fix them and will address any cosmetic review notes. However, due to different priorities, I'm not up for a deeper investigation or multi-class refactoring at this moment.

`ruby-x.y-rubocop` build job failures seem unrelated to my changes.

lib/rubocop/cop/registry.rb
```diff
@@ -23,34 +23,28 @@ def initialize(name, origin, badges)
...
       # @return [Array<Symbol>] list of departments for current cops.
       def departments
-        @departments.keys
+        staff.keys
       end
 
       # @return [Registry] Cops for that specific department.
       def with_department(department)
-        with(@departments.fetch(department, []))
+        with(staff.fetch(department, []))
```
pirj, line 42: Since `staff` is a `Hash.new { |hash, key| hash[key] = [] }` it might make sense to get rid of `fetch`.

lib/rubocop/cop/registry.rb
```diff
@@ -23,34 +23,28 @@ def initialize(name, origin, badges)
...
       end
 
       # @return [Registry] Cops not for a specific department.
       def without_department(department)
-        without_department = @departments.dup
+        without_department = staff.dup
         without_department.delete(department)
```
pirj, line 48: `@departments.reject { |dept, _| dept = department }` ?

lib/rubocop/rspec/support.rb
```diff
@@ -4,10 +4,13 @@
...
 require_relative 'host_environment_simulation_helper'
-require_relative 'shared_contexts'
 require_relative 'expect_offense'
+require_relative 'shared_contexts'
+require_relative 'stub_cop'
 
 RSpec.configure do |config|
   config.include CopHelper
+  config.include RuboCop::RSpec::ExpectOffense
+  config.include RuboCop::RSpec::StubCop
```
pirj, line 14: It's quite odd to add all the helpers to all of the examples.
I'd go with:
```ruby
  config.define_derived_metadata(file_path: %r{/spec/rubocop/cop/}) do |meta|
    meta[:cop_spec] = true
  end

  config.include CopHelper, :cop_spec
end
```

marcandre:
I was wondering why putting those cops in the 'Test' department isn't good enough? Note that `Cop::Cop.all` excludes those.

pirj:
I could have said that some extensions had to use ugly crutches to work this `self.inherited` around, but it seems that it's [not completely true](https://github.com/rubocop-hq/rubocop-rspec/pull/915/files#diff-89ff371da5dcfc94e318a5285e095123L13).
There's still this though:
```ruby
class RuboCop::Cop::RSpec::Cop
  ...
  def self.inherited(subclass)
    RuboCop::Cop::Cop.inherited(subclass)
  end
end
```

for a reason that will become clear from this example:
```ruby
class BaseBase
  @registry = []
  class << self
    attr_reader :registry
  end

  def self.inherited(base)
    registry << base
  end
end

class BaseExt < BaseBase

end

class Cop < BaseExt
end

# NoMethodError: undefined method `<<' for nil:NilClass
# from (pry):8:in `inherited'

BaseBase.registry # => [BaseExt]

class BaseExt
  def self.inherited(base)
    superclass.inherited(base)
  end
end

class AnotherCop < BaseExt
end

BaseBase.registry # => [BaseExt, AnotherCop]
```

marcandre:
I'm proposing to move the registry in #7868, one reason being that it should not be stored in an attribute of a class that people inherit from... I believe it would resolve this issue.

pirj:
Sure it will 👍 I like the proposed change to autocorrect syntax as well.

However, this specific change is only to improve specs (also across extensions) and doesn't contradict with your larger-scale refactoring.

marcandre:
For specs, in my PR there's an easy way:

```
around do |test|
  described_class.with_temporary_global { test.run }
end
```

This runs the tests while reseting the global registry afterwards. Couldn't think of a better name though 😅

One thing I like about the idea of lazy enlisting is that it would allow an easy way to have a `not_enlistable` or similar method that could remove a cop class from the list, for easy abstract base classes.

I would favor a slightly different implementation though, so that it's possible to enlist cops even after calling `cops` or something. If I'm not mistaken, after `call_for_serve_and_protect!` enlisting wouldn't work at all.

marcandre:
Haven't tested it, but something like https://github.com/marcandre/rubocop/commit/0bb18cb22ff918ccd9599301fc5df4768b0e29b3

marcandre:
Merged #8095, closing this.

pirj:
Nice!
But did #8095 solve the issue with leaky constants in specs? https://github.com/rubocop-hq/rubocop/pull/8023/files#diff-23dfc9af273ca5d2af6c37e1aef3c0c5L10

pirj:
I still see
```
      NoMethodError:
        undefined method `split' for nil:NilClass
 ```
if I:
```diff
diff --git a/spec/rubocop/cop/registry_spec.rb b/spec/rubocop/cop/registry_spec.rb
index 0524230fa..0cdaf5678 100644
--- a/spec/rubocop/cop/registry_spec.rb
+++ b/spec/rubocop/cop/registry_spec.rb
@@ -4,27 +4,6 @@ RSpec.describe RuboCop::Cop::Registry do
   subject(:registry) { described_class.new(cops, options) }
 
   let(:cops) do
-    stub_const('RuboCop::Cop::Test', Module.new)
-    stub_const('RuboCop::Cop::RSpec', Module.new)
-
-    # rubocop:disable RSpec/LeakyConstantDeclaration
-    module RuboCop
-      module Cop
-        module Test
-          # Create another cop with a different namespace
-          class FirstArrayElementIndentation < Cop
-          end
-        end
-
-        module RSpec
-          # Define a dummy rspec cop which has special namespace inflection
-          class Foo < Cop
-          end
-        end
-      end
-    end
-    # rubocop:enable RSpec/LeakyConstantDeclaration
-
     [
       RuboCop::Cop::Lint::BooleanSymbol,
       RuboCop::Cop::Lint::DuplicateMethods,
@@ -37,6 +16,11 @@ RSpec.describe RuboCop::Cop::Registry do
 
   let(:options) { {} }
 
+  before do
+    stub_const('RuboCop::Cop::Test::FirstArrayElementIndentation', Class.new(RuboCop::Cop::Cop))
+    stub_const('RuboCop::Cop::RSpec::Foo', Class.new(RuboCop::Cop::Cop))
+  end
+
   # `RuboCop::Cop::Cop` mutates its `registry` when inherited from.
   # This can introduce nondeterministic failures in other parts of the
   # specs if this mutation occurs before code that depends on this global cop
```

pirj:
Please disregard my previous comment, I forgot to update `master`.
All cool. I'll open a PR to clean up specs for which #8095 allows now 👍

marcandre:
Right, it doesn't remove the leaky constants, but it makes it easy to not affect the global Registry.

And yeah, that error of a Cop class not having a name just yet is solved with the lazy enlisting.

PR #8024, opened 2020-05-24
Use Regexp/String method #match? instead of #match

andrykonchin:
Minor performance optimization.

Replaced calls of `#match` method with `#match?` for `String`/`Regexp` classes in boolean context when matching details like captured substrings are ignored e.g.

```ruby
REGEXES.any? { |regex| literal.match(regex) }
```
becomes

```ruby
REGEXES.any? { |regex| literal.match?(regex) }
```

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/comment_config.rb
+++ b/lib/rubocop/comment_config.rb
@@ -124,7 +124,7 @@ module RuboCop
     end
 
     def directive_on_comment_line?(comment)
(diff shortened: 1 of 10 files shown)


koic:
Thank you! I remember that `Performance/Regexp` cop couldn't detect it yet :-)

Issue #7968, opened 2020-05-13
Allow cops to share context between files


ajorgensen:
## Is your feature request related to a problem? Please describe.

tl;dr It would be great to have a way to have a cop run across all processed sources to compare between files.

On a project I work on we make use of rails engines to separate different parts of our codebase into smaller components. As such we have a use case where we need to make sure there are no duplicate controller names in any of the rails engines so for example two separate rails engines should not be allowed to create the duplicate class name FooController.

Initially this was solved with a class level cache which would store all of the class names that had been seen and add an offense in there were any duplicates. This was a bit of a hack but since Rubocop creates a new cop instance per processed source this was the only way to share information between processed source files. When we added rubocop-daemon this became a problem because the cache now persists between runs due to the long running process. This leads to false positives because the cache will likely already contain all of the class names that we saw on the last run.

I know this isn't necessarily Rubocop's issue since its really the long running process via rubocop-daemon that creates the problem but I think that the issue can only really be solved within Rubocop itself.

## Describe the solution you'd like

There are three solutions I can think of. The easiest might be to provide a run identifier that can uniquely identify each different rubocop run. This would allow the user to check this ID during cop execution and determine if they want to clear the cache wherever it is stored. This might be the easiest and is essentially how I solved it for our cop (i'll explain my current solution further down).

A second solution would be to have a `context` object that is created at the beginning of the run and allow cops to store and share context between each of the cops. In my case, this would allow my cop to store the different class names which can be checked in the next instance of the cop for the next processed file. This would probably be the most flexible and fit into the current model.

Thirdly would be to change the current contract of a cop to allow it to be long lived. This would be to create all of the cop instances up front and then use those instance for each processed file. This would also likely require the addition of a few hooks that can be used by cops `before_run`, `after_run`, etc to allow the cop to manually clean up any state that it has.

## Describe alternatives you've considered

I have currently worked around the issue essentially implementing idea #1. In the cop we still have the class level cache but we also keep track of the object_id of the Rubocop::Runner instance. 

```
def _run_id
  ObjectSpace.each_object(RuboCop::Runner).to_a.last&.object_id || 1
end
```

This instance is long lived for the life of the run and is recreated in rubocop-daemon for each run. This gives us a unique identifier that can be checked against for each processed file. If the object_id has changed then we know its a new run and can clear the cache.

## Additional context

None

marcandre:
I imagine very few cops need this so it's not clear a builtin solution is advisable.

[...]

PR #8030, opened 2020-05-25
Better separation of concerns between Team and Commissioner

marcandre:
Excluding non-relevant cops is a duty of `Team`, not `Commissioner`.

Before this PR:
* a second call to `Commissioner#investigate` could be missing a cop that was excluded previously
* a force might be added by Team for a single cop that is later judged to be non relevant by a `Commissioner` but the later can't exclude the force now
* cops judged not relevant by the `Commissioner` were still looped through (e.g. in `autocorrect_all_cops`)

PR also includes some code cleanup. This is a prelude to #7968

### DIFF
--- a/lib/rubocop/cop/commissioner.rb
+++ b/lib/rubocop/cop/commissioner.rb
@@ -36,7 +36,6 @@ module RuboCop
 
       def investigate(processed_source)
         reset_errors
-        remove_irrelevant_cops(processed_source.file_path)
         reset_callbacks
         prepare(processed_source)
         invoke_custom_processing(@cops, processed_source)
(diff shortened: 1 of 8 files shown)


marcandre:
Argh, this is all good but the spec_helper needs to be fixed. It also shows it was duplicating the finding the forces. It should create a Team, not a Commissionner, but then this shows that `Team` doesn't have the right interface: it should accept cops, not cop classes. On hold.

marcandre:
I refactored `Team.new` to accept an array of cops instead of an array of cop classes. `Team` doesn't actually need to create the cops itself and is more useful to accept cops. Previous functionality was moved to `Team.mobilize`.

I feel that's the cleanest API, but it would be easy for `Team.new` to accept an array of cops *or* an array of cop classes, thus maintaining compatibility. Is there any code out there that might call `Team.new` and should we care, especially for a v1? /cc @bbatsov

bbatsov:
> I feel that's the cleanest API, but it would be easy for Team.new to accept an array of cops or an array of cop classes, thus maintaining compatibility.

Compatibility with what? :-) I think this class is not used outside the main gem. I did a quick grep in some extensions and I don't see any usages.

bbatsov:
Btw, this might also be a good opportunity to actually document the responsibilities of the class - I see currently they are "FIXME". :D

marcandre:
Cool. I'll add a short doc and a Changelog entry

marcandre:
> Compatibility with what? :-) I think this class is not used outside the main gem. I did a quick grep in some extensions and I don't see any usages.

Now we know that the answer is, at least, `pronto-rubocop` and `erb_lint` 😅.

bbatsov:
So we do. :-)

PR #8039, opened 2020-05-26
Fix false positives for Lint/ParenthesesAsGroupedExpression 

CamilleDrapier:
This PR fixes the following false positive for `Lint/ParenthesesAsGroupedExpression`

- when an expression is followed by an operator

```ruby
func (x) || y
```

- when an expression is followed by a chain of functions

```ruby
func (x).func.func.func.func.func
```

This PR adds some reproduction tests.

-----------------

I was initially thinking of changing the auto-correct to produce things such as `func((x) || y)` and `func((x).func.func.func.func.func)`, but since we probably do not want `eq (x).to_i` to become suddenly `eq((x).to_i)` (https://github.com/rubocop-hq/rubocop/pull/7909), I went with ignoring these pasterns. This is because the current auto-correct behavior breaks the code for these pasterns. Please do tell me if we would prefer going in another direction 🙏 

I'm not very comfortable with these language nodes, so please treat this PR as an humble attempt 🙏 

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Bug fixes
 
+* [#8039](https://github.com/rubocop-hq/rubocop/pull/8039): Fix false positives for `Lint/ParenthesesAsGroupedExpression` in when using operators or chain functions. ([@CamilleDrapier][])
 * [#8196](https://github.com/rubocop-hq/rubocop/issues/8196): Fix a false positive for `Style/RedundantFetchBlock` when using with `Rails.cache`. ([@fatkodima][])
 * [#8195](https://github.com/rubocop-hq/rubocop/issues/8195): Fix an error for `Style/RedundantFetchBlock` when using `#fetch` with empty block. ([@koic][])
 * [#8193](https://github.com/rubocop-hq/rubocop/issues/8193): Fix a false positive for `Style/RedundantRegexpCharacterClass` when using `[\b]`. ([@owst][])
@@ -4633,3 +4634,4 @@
 [@mauro-oto]: https://github.com/mauro-oto
 [@fatkodima]: https://github.com/fatkodima
 [@karlwithak]: https://github.com/karlwithak
+[@CamilleDrapier]: https://github.com/CamilleDrapier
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/parentheses_as_grouped_expression.rb
```diff
@@ -48,17 +48,22 @@ def valid_context?(node)
...
-        def grouped_parentheses?(node)
+        def chained_calls?(node)
           first_argument = node.first_argument
+          first_argument.send_type? && (node.children.last&.children&.count || 0) > 1
+        end
 
-          first_argument.send_type? && first_argument.receiver&.begin_type?
+        def logical_operator?(node)
+          first_argument = node.first_argument
+          (first_argument.and_type? || first_argument.or_type?) && first_argument.logical_operator?
```
koic, line 66: Can you use `operator_keyword?`?

```suggestion
          first_argument.operator_keyword? && first_argument.logical_operator?
```

And the test passes even if `&& first_argument.logical_operator?` will be removed. Is this condition necessary?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Bug fixes
 
+* [#8039](https://github.com/rubocop-hq/rubocop/issues/8039): Fix false positives for `Lint/ParenthesesAsGroupedExpression` in when using operators or chain functions. ([@CamilleDrapier][])
```
koic, line 7:
```suggestion
* [#8039](https://github.com/rubocop-hq/rubocop/pull/8039): Fix false positives for `Lint/ParenthesesAsGroupedExpression` in when using operators or chain functions. ([@CamilleDrapier][])
```

lib/rubocop/cop/lint/parentheses_as_grouped_expression.rb
CamilleDrapier: Thank you for suggesting this, it seems much simpler 👍 🙇

koic:
Thanks!

Issue #8022, opened 2020-05-24
Quick performance review


marcandre:
I had fun checking what takes time when inspecting RuboCop's own files.

Once I excluded the worst offender (see #8021), here's the top 20:

```
     584  (    3.7%)  RuboCop::Cop::Layout::SpaceInsideArrayLiteralBrackets#on_array
     425  (    2.7%)  RuboCop::Cop::Style::RedundantSelf#on_block
     390  (    2.5%)  RuboCop::Cop::Layout::SpaceBeforeFirstArg#on_send
     356  (    2.3%)  RuboCop::RSpec::TopLevelDescribe#on_send
     289  (    1.8%)  RuboCop::Cop::RSpec::SubjectStub#on_block
     255  (    1.6%)  RuboCop::Cop::Layout::IndentationConsistency#on_begin
     249  (    1.6%)  RuboCop::Cop::Layout::SpaceInsideReferenceBrackets#on_send
     246  (    1.6%)  RuboCop::Cop::Layout::FirstArgumentIndentation#on_send
     236  (    1.5%)  RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#on_block
     206  (    1.3%)  RuboCop::Cop::Layout::DefEndAlignment#on_send
     184  (    1.2%)  RuboCop::Cop::Layout::SpaceInsideHashLiteralBraces#on_hash
     183  (    1.2%)  RuboCop::Cop::Style::TrailingMethodEndStatement#on_def
     176  (    1.1%)  RuboCop::Cop::Performance::FixedSize#on_send
     160  (    1.0%)  RuboCop::Cop::StringHelp#on_str
     154  (    1.0%)  RuboCop::Cop::RSpec::ReturnFromStub#on_block
     153  (    1.0%)  RuboCop::Cop::RSpec::InstanceVariable#on_block
     152  (    1.0%)  RuboCop::Cop::Interpolation#on_dstr
     143  (    0.9%)  RuboCop::Cop::RSpec::LetSetup#on_block
     138  (    0.9%)  RuboCop::Cop::RSpec::MultipleSubjects#on_block
```

The fastest cops take 0.0%, so there's potential for improvement.

First comes `RuboCop::Cop::Layout::SpaceInsideArrayLiteralBrackets`. It uses `tokens` which should probably be optimized (e.g. using `bsearch_index`), but I'm not sure it should even be looking at tokens at all.

[...]

PR #8048, opened 2020-05-27
Minor performance tweaks

rrosenblum:
This is somewhat related to https://github.com/rubocop-hq/rubocop/issues/8022. I decided to briefly look over some of the potential performance issues. I don't think anything in here is going to radically improve performance. It mostly seemed like low hanging fruit to avoid an extra allocation.

### DIFF
--- a/lib/rubocop/cop/layout/multiline_array_line_breaks.rb
+++ b/lib/rubocop/cop/layout/multiline_array_line_breaks.rb
@@ -27,7 +27,7 @@ module RuboCop
           'on a separate line.'
 
         def on_array(node)
-          check_line_breaks(node, node.children)
+          check_line_breaks(node.children)
         end
 
         def autocorrect(node)
--- a/lib/rubocop/cop/layout/multiline_hash_key_line_breaks.rb
+++ b/lib/rubocop/cop/layout/multiline_hash_key_line_breaks.rb
@@ -32,7 +32,7 @@ module RuboCop
           # Style/MultilineMethodArgumentLineBreaks handles those.
           return unless starts_with_curly_brace?(node)
 
-          check_line_breaks(node, node.children) if node.loc.begin
+          check_line_breaks(node.children) if node.loc.begin
         end
 
         def autocorrect(node)
--- a/lib/rubocop/cop/layout/multiline_method_argument_line_breaks.rb
+++ b/lib/rubocop/cop/layout/multiline_method_argument_line_breaks.rb
@@ -40,7 +40,7 @@ module RuboCop
             args = args.concat(args.pop.children) if last_arg.hash_type? && !last_arg.braces?
           end
 
-          check_line_breaks(node, args)
+          check_line_breaks(args)
         end
 
         def autocorrect(node)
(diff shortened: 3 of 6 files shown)


marcandre:
Sorry I haven't had time to plugin some measuring tools. Without that, it's difficult to see how good a change is. Do you know the [first rule of optimisation](https://wiki.c2.com/?FirstRuleOfOptimization)? 😆

In your PR, for example, I bet you a beer that your changes of `children.each` to `each_child_node` will make things slower.

Of your porposed changes, I like the one to `cop/variable_force.rb` because it's a good use of `each_children_node`. I'm afraid I'm less convinced that the others are actual improvements (speed or legibility).

If you want to help with #8022 it would be awesome. I'll try to get a basic tool for measurement, but there's already a benchmarking task you can use, since you know which cops to attack. Those cops should get at least an order of magnitude faster. Many of them seem to rely on analyzing the tokens for whitespace and all...

rrosenblum:
I have not tested these changes with any kind of profiler, so it is possible that they don't make any kind of impact. I've made changes similar to that in `cop/variable_force.rb` to help with performance in the past. Same with converting chained looping operators (`map`, `select`, `reject`, `each`) to use `each_with_object`. As the JIT compiler is improved, I'm not sure that those kinds of changes will be necessary in the future.

What is your reasoning for believing changing `children.each` to `each_child_node` will slow things down?

In the past, I have relied on [`memory_profiler`](https://github.com/SamSaffron/memory_profiler) to help with fixing performance issues

marcandre:
> What is your reasoning for believing changing `children.each` to `each_child_node` will slow things down?

First, `each_child_node` does something additional which is to check that a child is a node before yielding it. That testing is not done with `children.each` so it's not quite possible to do more and yet be faster.

But even assuming there was no additional checking, it's not possible for `each_child_node` to be faster because it has to, whatever implementation it has, call `children` itself (or access `@children`) and must call `each` (or some other iterator which would be slower). So it's at best exactly the same, probably a tiny bit slower (say one more method call).

That's the very best case scenario, which would requires the implementation to use `&block` [and Ruby 2.5+](https://bugs.ruby-lang.org/issues/14045); if it calls `yield` it will be slower as builtin methods (like `Array#each` here) yield faster than Ruby methods. If it checks if it needs to return an enumerator if a block is given (as it should and it does), that's more processing, etc.

You just can't beat an attribute call (super optimized) + builtin `Array#each` (super optimized).

Even if you avoided creating an intermediate array (which you can't here), it is typically not faster. It's a bit frustrating, but yielding from Ruby code is actually a bit slow. When optimizing Parser's `TreeRewriter` I measured that it was faster to return a temporary array of things and act on them then yielding them and avoid creating the intermediate array (few thousand elements IIRC, test in Ruby 2.7) 🤷‍♂️.

As for any talk about optimization: this is theoretical, as unless someone changes the `O` of an algorithm (say [like `O(n)` vs `O(1)`](https://github.com/rubocop-hq/rubocop-ast/issues/22)), or someone measures that there's actually a bottleneck at some point, it's typically not worth trying to optimize it.

> The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times
-- Knuth

rrosenblum:
Thanks for providing the feedback. There were some aspects of the change that I had not considered before. You are absolutely correct about `node.children.each` being better than `node.each_child_node`. 

There are a couple of small changes in here that I think still could be good. Overall, they are rather small. I still need to actually test the performance diff.

PR #8050, opened 2020-05-28
New option --display-only-failed for junit format

burnettk:
Speeds up test report processing for large codebases and helps address
the sorts of concerns raised at
https://github.com/mikian/rubocop-junit-formatter/issues/18.

In one case for me, this reduced the number of `testcase` xml elements from 66,000 to 1, which made a material difference in terms of jenkins test report load times.

Refactors file_finished method in junit_formatter.rb to reduce
complexity.

Fixes buggy error message when --disable-uncorrectable is run without
--auto-correct, which I noticed because I added similar error messaging
when you try to run --display-only-failed without --format junit.

Tests, CHANGELOG.md, and help message updates to accompany new option

### DIFF
--- a/legacy-docs/formatters.md
+++ b/legacy-docs/formatters.md
@@ -242,7 +242,7 @@ The JSON structure is like the following example:
 **Machine-parsable**
 
 The `junit` style formatter provides the JUnit formatting.
-This formatter is based on [rubocop-junit-formatter gem](https://github.com/mikian/rubocop-junit-formatter).
+This formatter is based on the [rubocop-junit-formatter gem](https://github.com/mikian/rubocop-junit-formatter).
 
 ```sh
 $ rubocop --format junit
@@ -267,6 +267,18 @@ $ rubocop --format junit
   </testsuite>
 </testsuites>
 ```
+
+The `junit` style formatter is very useful for continuous integration systems
+such as Jenkins, most of which support junit formatting when parsing test
+results. A typical invocation in this type of scenario might look like:
+
+```sh
+$ rubocop --format junit --out test-reports/junit.xml
+```
+
+Since there is one XML node for each cop for each file, the size of the resulting
+XML can get quite large. If it is too large for you, you can restrict the output
+to just failures by adding the `--display-only-failed` option.
 
 ## Offense Count Formatter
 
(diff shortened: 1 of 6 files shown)


spec/rubocop/options_spec.rb
```diff
@@ -224,6 +226,18 @@ def abs(path)
       end
     end
 
+    describe '--display-only-failed' do
+      it 'fails if given without --format junit' do
+        expect { options.parse %w[--display-only-failed] }
+          .to raise_error(RuboCop::OptionArgumentError)
+      end
+
+      it 'works if given with --format json' do
```
bbatsov, line 235: I guess you meant to write junit here.

spec/rubocop/options_spec.rb
```diff
@@ -92,6 +92,8 @@ def abs(path)
                                                This option applies to the previously
                                                specified --format, or the default format
                                                if no format is specified.
+                  --display-only-failed        Only output offense messages. Omit passing
+                                               cops. Only valid for --format junit
```
bbatsov, line 96: You forgot the final `.`

bbatsov:
I'm fine with the proposed change. I think it'd be a good idea to also mention it here https://docs.rubocop.org/rubocop/formatters.html#junit-style-formatter

burnettk:
Corrected the two specific issues and updated the relevant documentation. I love this tool, @bbatsov, thank you!

legacy-docs/formatters.md
```diff
@@ -242,7 +242,7 @@ The JSON structure is like the following example:
 **Machine-parsable**
 
 The `junit` style formatter provides the JUnit formatting.
-This formatter is based on [rubocop-junit-formatter gem](https://github.com/mikian/rubocop-junit-formatter).
+This formatter is based on the [rubocop-junit-formatter gem](https://github.com/mikian/rubocop-junit-formatter).
```
bbatsov, line 245: You didn't make the same edit to the actual docs. There's no need to touch the markdown files - they will be removed as soon as we setup permanent redirects between the old docs site that used them (powered by RTD) and the new one (docs.rubocop.org).

bbatsov:
Thanks for tackling this! I'll do the small doc change I mentioned myself.

PR #8054, opened 2020-05-28
[Fix #7971] Fix --disable-uncorrectable not updating uncorrected code

### LINKED ISSUE #7971: Lint/UnusedMethodArgument does not correct or mark keyword args as "todo"
When using the `--safe-auto-correct --disable-uncorrectable` flags,  `Lint/UnusedMethodArgument` has some inconsistent behaviour. It auto-corrects methods with standard arguments, but not keyword arguments. It doesn't add `rubocop:todo` comments to methods with keyword arguments either.

--------

## Expected behavior

Auto-correct and disabling uncorrectable offenses works for methods with keyword arguments

_example.rb_
```ruby
def ordinary_method(some_arg)
  puts "Ignoring args"
end

def method_with_keyword_arg(some_keyword_arg:)
  puts "Ignoring args"
end
```

```
$ bundle exec 'rubocop --safe-auto-correct --disable-uncorrectable --only Lint/UnusedMethodArgument example.rb'
Inspecting 1 file
W

Offenses:

example.rb:1:21: W: [Corrected] Lint/UnusedMethodArgument: Unused method argument - some_arg. If it's necessary, use _ or _some_arg as an argument name to indicate that it won't be used. You can also write as ordinary_method(*) if you want the method to accept any arguments but don't care about them.
def ordinary_method(some_arg)
                    ^^^^^^^^
example.rb:5:29: W: Lint/UnusedMethodArgument: Unused method argument - some_keyword_arg. You can also write as method_with_keyword_arg(*) if you want the method to accept any arguments but don't care about them.
def method_with_keyword_arg(some_keyword_arg:)
                            ^^^^^^^^^^^^^^^^

```

## RuboCop version
```
$ bundle exec 'rubocop -V'
0.83.0 (using Parser 2.7.1.2, running on ruby 2.6.5 x86_64-darwin19)
```

rrosenblum:
This fixes #7971. 

I still need to decide how to handle the failing metrics in `cop.rb`. I know @marcandre mentioned that #7868 will make this easier/change the way this would work. I haven't had a chance to look over that PR.

There are a few more things that I still want to look into with `--disable-uncorrectable`. I'm curious how well it works on lines that already contain a different `rubocop:todo` or `rubocop:disable` comment.


marcandre:
I wasn't qute correct. My PR helps in detecting non-corrections, but you're right that they should be then disabled. I'm :+1: to merge (but you need to resolve/disable the "metrics/" offenses 🙄)

bbatsov:
Thanks!

PR #8056, opened 2020-05-29
Ruby 2.3 cleanup

bquorning:
Since we dropped support for Ruby 2.3 in 58738943eb3f7a41266b4546a829fd0a5a13703d, there’s some cleanup to be done:

- Documentation should not refer to Ruby 2.3.
- Code comments should not refer to Ruby 2.3.
- `Layout/HeredocIndentation` cop should not allow `active_support`, `powerpack` and `unindent` styles, only the `squiggly` style.

Those three issues are fixed by this PR.

Please note that after removing functionality from `Layout/HeredocIndentation` I couldn’t resist cleaning up the remaining code and specs. Those changes _may_ be deemed large enough for a separate pull request. Please let me know.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/layout/heredoc_indentation_spec.rb
+++ b/spec/rubocop/cop/layout/heredoc_indentation_spec.rb
@@ -8,37 +8,6 @@ RSpec.describe RuboCop::Cop::Layout::HeredocIndentation, :config do
     }
   end
 
-  shared_examples 'offense' do |name, code, correction = nil|
-    it "registers an offense for #{name}" do
-      inspect_source(code)
-      expect(cop.offenses.size).to eq(1)
-    end
-
-    it "autocorrects for #{name}" do
-      corrected = autocorrect_source_with_loop(code)
-      expect(corrected).to eq(correction)
-    end
-  end
-
-  shared_examples 'accept' do |name, code|
-    it "accepts for #{name}" do
-      expect_no_offenses(code)
-    end
-  end
-
-  shared_examples 'warning' do |message|
-    it 'warns' do
-      correct = lambda do
-        autocorrect_source(<<~RUBY)
-          <<-RUBY2
-          foo
-          RUBY2
-        RUBY
-      end
-      expect(&correct).to raise_error(RuboCop::Warning, message)
-    end
-  end
-
   shared_examples 'all heredoc type' do |quote|
     context "quoted by #{quote}" do
       it 'does not register an offense when not indented but with ' \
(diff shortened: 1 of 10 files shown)


bbatsov:
The changes seems good to me overall, but I'm confused why did you modify the legacy markdown docs instead of the new docs.

bbatsov:
Btw, I didn't delete the markdown site only to avoid breaking links to it. At some point we'll have to redirect somehow from the RTD to the new pages.

bquorning:
> The changes seems good to me overall, but I'm confused why did you modify the legacy markdown docs instead of the new docs.

Ahh, sorry. I ran `rake` before realising I had to rebase on master. I’ll have force-pushed with the real doc changes.

So the entire `manual` folder is going away? If it’s not supposed to be kept updated, shouldn’t it be deleted?

bbatsov:
The problem is when you Google for RuboCop docs now you get the following:

image

If we just delete the entire folder without setting up some redirects that's going to break the top search results. I guess that's not that big of a deal, but we should probably avoid it.

bquorning:
> we should probably avoid it.

Good point.

spec/rubocop/cop/layout/heredoc_indentation_spec.rb
```diff
@@ -111,142 +21,170 @@ def foo
...
+          <<#{quote}RUBY2#{quote}
+          foo
+          ^^^ Use 2 spaces for indentation in a heredoc by using `<<~` instead of `<<`.
           RUBY2
         RUBY
+
+        expect_offense(annotated_source)
+
+        corrected = looped_autocorrect_from_annotated_source(annotated_source)
+        expect(corrected).to eq(<<~CORRECTION)
```
bquorning, line 134: It would actually be pretty neat if you could pass a `loop: true` 2nd argument to `expect_correction`. I noticed a few other places in the specs that rely on multiple passes when testing autocorrection.

bbatsov: You can open a separate ticket to track this idea.

sergey-alekseev:
>I'm confused why did you modify the legacy markdown docs instead of the new docs

Actually it was confusing  
1. getting the warning (see the comment footer),
2. going to https://rubocop.readthedocs.io/en/stable/cops_layout/#layoutheredocindentation (by googling or remembering that site),
3. finding out there is still `EnforcedStyle`

In my view having both websites should be modified until  
>At some point we'll have to redirect somehow from the RTD to the new pages.

Now I see there is https://docs.rubocop.org/rubocop/0.85/cops_layout.html#layoutheredocindentation, but I found it only after going to the changelog, searching for 'Layout/HeredocIndentation' and finding this PR. There is no docs.rubocop.org among my search results:  
Screen Shot 2020-06-03 at 10 54 33

---

Initially I encountered the following warning and had to found this PR. Leaving this to be indexed by search bots and potentially help other people:  
```
Warning: Layout/HeredocIndentation does not support EnforcedStyle parameter.

Supported parameters are:

  - Enabled
```

PR #8058, opened 2020-05-29
[fix #7902] Unsafe conversion methods as symbols

### LINKED ISSUE #7902: Lint/NumberConversion doesn't handle `to_i` in symbol form
Running `rubocop --only "Lint/NumberConverion" --auto-correct` does not handle the following cases:

```
"1,2,3,foo,5,6,7,8".split(',').map(&:to_i)

"foo".try(:to_i) # when using Rails, returns a 0
```

--------

## Expected behavior

At the minimum, **it should report a violation**.

For auto-correction, I'm not completely sure if this will work for all scenarios, but it could look like the following:
```
"1,2,3,foo,5,6,7,8".split(',').map { |i| Integer(i) } 
```
and
```
"foo".try{ |n| Integer(n) }
```

## Actual behavior
The above violations of using `to_i` are not reported.

Just use any of the above examples in a file and run rubocop over them.

## RuboCop version
```
$ rubocop -V
0.78.0 (using Parser 2.6.5.0, running on jruby 2.5.3 java)
```

petar-lazarov:
Passing `&:to_i`, `&:to_f` or `&:to_c` to `map` and `collect` methods
should be considered offense by NumberConversion cop.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [#8017](https://github.com/rubocop-hq/rubocop/pull/8017): Fix a false positive for `Lint/SuppressedException` when empty rescue with comment in `def`. ([@koic][])
 * [#7990](https://github.com/rubocop-hq/rubocop/issues/7990): Fix resolving `inherit_gem` in remote configs. ([@CvX][])
 * [#8035](https://github.com/rubocop-hq/rubocop/issues/8035): Fix a false positive for `Lint/DeprecatedOpenSSLConstant` when using double quoted string argument. ([@koic][])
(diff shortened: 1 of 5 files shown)


petar-lazarov:
:wave: Hello,

the original issue mentions `try(:to_i)` as violation of `Lint/NumberConversion`, but I'm not sure if this should be handled, because the `try` method is specific for Rails.

What do you think? Should I add handling for `try(:to_method)` to the cop?

Also I was thinking about handling all `&:to_i`, `&:to_f` or `&:to_c` independent of what method is called (e.g. `something.try(&:to_i)` to be violation as well) but I am not sure if this is OK for all the cases.

bbatsov:
I think it might be better to have some list of methods handled by the cop via configuration rather than hardcoding just `map` and `collect`. I agree this probably covers most cases, but making this more generic is always better.

bbatsov:

Issue #8003, opened 2020-05-21
cop_helpers should be deprecated


marcandre:
As a followup to #7970, I feel that all of the methods in `cop_helper` should be deprecated. Here is why:

`inspect_source` and `autocorrect_source` seem too low level and don't lead to the right patterns. Among other things, they don't use `let`s for anything. So some things get duplicated (e.g. building the `processed_source`). No instance is shared by default, so many specs are calling `inspect_source` and then `autocorrect_source`, which is a really bad pattern: the same `cop` is called twice, with different `source_buffer` (although the same source code), with settings that has been hacked (via `instance_variable_get` 😱), something that won't even work currently if the `cop` is constructed with `auto_correct` set to `true`.

Example: [this spec](https://github.com/rubocop-hq/rubocop/blob/master/spec/rubocop/cop/style/ternary_parentheses_spec.rb) checks 78 cases (59 with offenses, 19 without).  The `inspect_source` is called 118 times for nothing. Each case with offense also calls `autocorrect_source` twice.

 I feel these (at least) should be officially deprecated (at least in the doc) in `1.0`, in favor of `expect_offense`/`expect_no_offense` or better.

`inspect_source_file` is used in one spec, I don't know if there's a good reason for it at all.

Maybe `parse_source` could remain as is, although it is still very low level and I'd still favor `processed_source` (in shared context).

`autocorrect_source_with_loop` => I checked the one cop spec that uses it. Unless I'm mistaken, this isn't really about the cop but simply that some autocorrected cases may lead to another autocorrectable offense. Would be best handled with two `expect_offense` with correction or similar.

bbatsov:
No objections from me. Those are all remnants from the early days of RuboCop and we certainly have better APIs at our disposal today.

bquorning:
> `autocorrect_source_with_loop` => I checked the one cop spec that uses it. Unless I'm mistaken, this isn't really about the cop but simply that some autocorrected cases may lead to another autocorrectable offense. Would be best handled with two `expect_offense` with correction or similar.

I am actually working on adding looping to the `expect_correction` method.

marcandre:
> I am actually working on adding (optional) looping to the `expect_correction` method.

Sounds great!

rrosenblum:
I originally posted this info in #8127. I think it makes more sense here since it is a more technical discussion.

[...]

PR #8062, opened 2020-05-29
expect_correction with loop

bquorning:
In some places, we spec what happens after `one` auto-correct iteration, and in some places we spec what happens after _many_ auto-correct iterations.

As mentioned in #8003, there is no longer any reason to use helpers like `autocorrect_source_with_loop`. This PR removes that method and all our uses of it.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/rspec/expect_offense.rb
+++ b/lib/rubocop/rspec/expect_offense.rb
@@ -96,17 +96,35 @@ module RuboCop
 
         expect(actual_annotations.to_s).to eq(expected_annotations.to_s)
       end
-      # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
 
-      def expect_correction(correction)
+      def expect_correction(correction, loop: false)
         raise '`expect_correction` must follow `expect_offense`' unless @processed_source
 
-        corrector =
-          RuboCop::Cop::Corrector.new(@processed_source.buffer, cop.corrections)
-        new_source = corrector.rewrite
+        iteration = 0
+        new_source = loop do
+          iteration += 1
+
+          corrector =
+            RuboCop::Cop::Corrector.new(@processed_source.buffer, cop.corrections)
+          corrected_source = corrector.rewrite
+
+          break corrected_source unless loop
+          break corrected_source if cop.corrections.empty?
+
+          if iteration > RuboCop::Runner::MAX_ITERATIONS
+            raise RuboCop::Runner::InfiniteCorrectionLoop.new(@processed_source.path, [])
+          end
+
+          # Prepare for next loop
+          cop.instance_variable_set(:@corrections, [])
+          @processed_source = parse_source(corrected_source,
+                                           @processed_source.path)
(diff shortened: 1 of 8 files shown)


lib/rubocop/rspec/expect_offense.rb
```diff
@@ -98,12 +98,30 @@ def expect_offense(source, file = nil)
...
+
+          break corrected_source if !loop
+          break corrected_source if cop.corrections.empty?
+
+          if iteration > RuboCop::Runner::MAX_ITERATIONS
+            raise RuboCop::Runner::InfiniteCorrectionLoop.new(@processed_source.path, [])
+          end
+
+          # Prepare for next loop
+          cop.instance_variable_set(:@corrections, [])
```
bquorning, line 120: This bit is ugly! Is there another way to accomplish the same?

bquorning:
@pirj Maybe you want to take a look at this PR?

lib/rubocop/rspec/expect_offense.rb
marcandre: It also won't work after #7868 lands...

marcandre:
API-wise, I like that it's very convenient; I wonder if it would it better to also spec the intermediate steps?

E.g.
```ruby
expect_offense(<<~RUBY)
  # ... code with offense
RUBY
expect_correction_offense(<<~RUBY)
  # ... corrected code with offense
RUBY
expect_correction(<<~RUBY)
  # ... final corrected code without offense
RUBY
```

One advantage I see is that it makes it clear how the final correction is had. Also it specs what happens if someone entered the intermediate code for some reason. Maybe that's too strict or verbose though?

lib/rubocop/rspec/expect_offense.rb
```diff
@@ -96,17 +96,35 @@ def expect_offense(source, file = nil)
 
         expect(actual_annotations.to_s).to eq(expected_annotations.to_s)
       end
-      # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
 
-      def expect_correction(correction)
+      def expect_correction(correction, loop: false)
```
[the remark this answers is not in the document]
pirj: `rspec` execution time didn't change with `loop: true` (I've actually removed this parameter and removed `loop: true` around the code.

I've faced an `RuboCop::Runner::InfiniteCorrectionLoop` and some other spec failures. Is this due to disagreeing cops?

marcandre:
I still feel this raises questions:
- should we be more specific with spec'ing the intermediate steps?
- or should we spec that a step leads to the expected correction (as @pirj suggests)?
- or should we not always test only the final step?
- why are there infinite loop corrections in our test base?

bbatsov:
I think it's best to discuss them in a dedicated ticket. This PR improves the status quo, which is why I decided to merge it, but I agree that there's plenty of room for improvement.

marcandre:
>This PR improves the status quo

No doubt about that :-)

I opened #8075 and #8076

PR #8065, opened 2020-05-29
Update vim integrations

andyw8:
vim-rubocop hasn't been updated since 2015. I think for most users, picking one of the general linting plugins makes more sense, so let's put those first?


koic:
I agree with this. Thanks!

PR #8069, opened 2020-05-30
Spec refactoring

marcandre:
This PR adds a new functionality: `expect_offense` can do some text replacements (see `style/redundant_exception_spec.rb`)

It refactors slightly other specs. This is a prelude to #7868


bbatsov:
Great work! Probably we should document the specs API somewhere in the dev docs as well.

Issue #2716, opened 2016-01-25
Extended limit with source comments


iGEL:
First of all, thanks for RuboCop.

When introducing rubocop into legacy code, you usually can't fix all offences right away. So the usual way I use is to introduce source comments to make RuboCop pass. But complex methods are now just whitelisted, which might lead thoughtless contributors to unintentionally make bad stuff even worse. For offences, that are limited by a numerical value like `Metrics/AbcSize`, it would be really great if you could fix a new maximum.

I'm thinking about something like this:

``` ruby
# rubocop:disable Metrics/AbcSize(Max: 17.10)
```

While a bigger value clearly should be a non autofixable offense, I'm not so sure about what should happen about lower values. While it should be autofixable, I'm not sure whether it should be an offense. What do you guys think?

If you like the idea, I would like to try to implement it.

jonas054:
I have nothing against the idea. It might be a little tricky to implement, as it has to play nicely with `--auto-gen-config` for example.

When it comes to syntax of the comment, I think I would prefer something like
```ruby
# rubocop:set Metrics/AbcSize: Max: 17.10
# [...]
# rubocop:reset Metrics/AbcSize: Max
```

What about other properties? I guess this means that all configuration must be per line rather than per file. Sounds like a lot of work.

alexdowad:
> Sounds like a lot of work.

A lot of work. A **LOT** of work.

rrosenblum:
Sounds like a really cool idea. I also have wanted a bit more control over some of the Metrics cops when working with legacy code. If someone chooses to try and tackle it, it may take a while to implement. I think something like this is nice to have on our radar.

bbatsov:
It's a good idea. I'd just mention there should be a way to alter a few attributes at once and probably `reset` should just restore the default config so it shouldn't take params. I think that a layered `reset` would make this feature way more complex and would be largely redundant in practice.

bbatsov:
Seems that we won't be moving forward with this, so I'll close it.

PR #8071, opened 2020-05-31
Introduce rubocop:set/rubocop:reset directives

zverok:
> A lot of work. A **LOT** of work. (@alexdowad at #2716)

So, here we go! _This PR should be considered rather of a draft-y quality, to discuss the whole idea and its implementation. But I consider it _mostly_ complete._

This PR introduces `rubocop:set Cop{Setting: value}` and `rubocop:reset Cop|all` directives.

Design/implementation decisions:

**0.** Directive syntax is `rubocop:set Metrics/AbcSize{Max: 20}`. The `{Max: 20}` part is expected to be a valid YAML. 
**1.** Directives are not nested:
```ruby
# rubocop:set Metrics/AbcSize{Max: 20}
# ...here it is 20...
# rubocop:set Metrics/AbcSize{Max: 30}
# ...here it is 30
# rubocop:reset Metrics/AbcSize
# ...here it is default (fully reset), not back to 20
```
I believe it is less confusing, but open to discussion. One exception is single-line directives:
```ruby
# rubocop:set Metrics/AbcSize{Max: 20}
# ...here it is 20...
def very_really_complicated # rubocop:set Metrics/AbcSize{Max: 50}
  # 50 for this method only
end
# ...but here it is 20 again, not reset to default
```
**2.** Directives work only for `#on_<nodetype>` style cops, not `#investigate`-style cops. That's because in `#investigate`, there is no place where `Commissioner` or base `Cop` is hooked to change something line-by-line (`rubocop:disable` works this around by allowing cop to _analyze_, but preventing it to _report_ on disabled lines). This is a hard limitation, I believe the feature would be useful nevertheless, but it should be communicated clearly (may be with additional feature with auto-documentation, which checks for `def investigate` and specifies whether cop supports inline settings, depending on it).
**3.** I took the liberty to rewrite `CommentConfig` completely. I've introduced a notion of `Directive` class, which should open a way for, probably, implementing new ones. 

On PR quality:
1. Currently, the code of `CommentConfig` is written in the style that allows to see algorithms clearly (large-ish methods, and usage of `rubocop:set` to enlarge thresholds, just to make a point). When/if we agree on the final implementation, I am ready to change the style (though, see PS about it)
2. The tests are a bit shallow now (again, awaiting discussion of architectural details)

PS: On methods complexity: TBH, it is my current belief that Rubocop's default thresholds for complexity metrics suit nice for business logic (where if you have three branches of `if`, they probably represent Nameable Cases, and should be extracted to methods or even classes), but much less for algorithms. To show one case, here is method `CommentConfig#extra_enabled_comments` that provides list of redundant `rubocop:enable` directives
<details>

```ruby
def extra_enabled_comments # rubocop:set Metrics/AbcSize{Max: 26}
  disabled = all_cop_names.to_h { |name| [name, false] }

  directives
    .select(&:disable_enable?).reject(&:single_line?)
    .chunk { |d| d.enable? && d.all? }
    .each_with_object([]) do |(enable_all, dirs), redundant|
      # if this group of directives is exploded from
      # `# rubocop:enable all` comment, it should be handled
      # as one statement
      if enable_all
        redundant << [dirs.first.comment, 'all'] if disabled.values.none?
        disabled.transform_values! { false }
      else
        dirs.each do |dir|
          redundant << [dir.comment, dir.cop] if dir.enable? && !disabled[dir.cop]
          disabled[dir.cop] = dir.disable?
        end
      end
    end
end
```

</details>
I'd argue that this is the best representation of the algorithm. The method itself has one local variable, and one large iteration (with several smaller "preparatory" Enumerable methods), with one `if` inside it; it depends on 2 other, very generic methods, and only for setting initial data. It fits on half a screen even on small laptop, and makes it easy to follow "what'll be in what variable" at every step. But it violates both method length (14 max, method is 20), and AbcSize (15 max, method is 26) cops.

Alternatively, [the current implementation](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/comment_config.rb#L44) is split into multiple small methods throughout the class, works with _two_ mutable collections (array and hash) passed from method to method and changed in each of them; and as a person who tried to understand what and how it does, I can witness it was not easy to follow through all the micro-methods.

I am not sure what is my point here, and it is probably not a discussion for this PR, but still :)

PPS: I'll comment on some non-obvious changes inline.

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/missing_cop_enable_directive.rb
```diff
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-# rubocop:disable Lint/RedundantCopDisableDirective
```
zverok, line 3: I believe it was a bug in old `CommentConfig`: It treated code like this:
```ruby
# Example of usage:
#  ```ruby
#  # rubocop:disable SomeCop
```
...as a real directive. I fixed it, but making sure that `rubocop:<directive_name>` is the first thing in comment.
Therefore some of disable/enable throughout codebase, made to workaround this problem, became irrelevant and were removed.

spec/rubocop/comment_config_spec.rb
```diff
@@ -50,7 +50,7 @@
         '# rubocop:disable Layout/BlockAlignment some comment why',
         '# rubocop:enable Style/Send, Layout/BlockAlignment but why?',
         '# rubocop:enable Lint/RandOne foo bar!',            # 43
-        '# rubocop:disable Lint/EmptyInterpolation',
```
zverok, line 53: It was a bug in test.
Example description for this cop reads "supports enabling/disabling cops without a prefix", but cop was WITH a prefix :)

spec/rubocop/comment_config_spec.rb
```diff
@@ -61,7 +61,7 @@
     end
 
     def disabled_lines_of_cop(cop)
-      (1..source.size).each_with_object([]) do |line_number, disabled_lines|
+      (1..source.lines.count).each_with_object([]) do |line_number, disabled_lines|
```
zverok, line 64: Was bug in the test. `source.size` is character count of source, not line count.

zverok:
About CI fails:
1. I left _some_ of Rubocop self-investigation fails till final refactoring.
2. Two of failing tests: I am not sure they test the right thing! For example, one tests that if the cop has wrong namespace `# rubocop:disable Style/LineLength`, then (besides other errors), we'll have a "Redundant enable" warning about subsequent `# rubocop: enable Layout/LineLength`, which fails for my implementation, and I believe it is better this way (even if the disable was in the wrong namespace, we "got what you meant", so subsequent enable shouldn't be warned about).

I marked PR as draft, but it means "I want to discuss it", not "I'll be working on it for some more time till it would be ready for discussion".

bbatsov:
It's really great to see someone tackling this! I'll review the PR in the next couple of days, as lately it has been pretty busy on my end. 

Have you measured the performance impact of the changes? (especially in the absence of annotations). I'm assuming it's not insignificant if you're hooking into the node callbacks.

spec/rubocop/comment_config_spec.rb
bbatsov: Nice catch!

spec/rubocop/comment_config_spec.rb
bbatsov: 👍

lib/rubocop/cop/commissioner.rb
```diff
@@ -119,6 +121,18 @@ def with_cop_error_handling(cop, node = nil)
         err = ErrorWithAnalyzedFileLocation.new(cause: e, node: node, cop: cop)
         @errors << err
       end
+
+      def with_inline_config(cop, node = nil)
```
bbatsov, line 125: That's the code that concerns me in terms of performance. We have to be really careful with any overhead added to the node callbacks.

zverok:
@bbatsov
> Have you measured the performance impact of the changes?

Not yet. I wanted to hear an opinion on the whole idea first, but I definitely plan to profile/benchmark/optimize before calling it a day.
As for _viability_ of the approach, I checked it empirically (running update rubocop against its own codebase), and it hasn't become _unbearably slow_ (or even _noticeably slower_) on my machine (which is pretty rough metric, I agree!). That's why I believe it is possible to proceed this way.

zverok:
@bbatsov So, I profiled the situation a bit, and it appears that new method indeed becomes a bottleneck, in a quite bad way. The thing is, "whether we have config for this cop" is of course called once-per `on_` method, and it leads to numbers like ~2mln calls on 254 not-very-large files being analyzed. At this rate, performance penalty strucks on the most innocent methods. For example, `ProcessedSource#comment_config` spends ~2 seconds for the simple `@comment_config ||= CommentConfig.new(self)` -- it calls real (and trivial) `CommentConfig.new` just once per file (e.g. 224 times), but turns out that at least on my machine's Ruby 2.6 that simple method call + `||=` is heavy enough to provide a noticeable slowdown. The same with `Cop#cop_name` (again, 2mln of calls of simple memoization `||=` manifests itself into 1 second of slowdown).

I don't believe it is acceptable, so I am postponing this PRs further development and will try to think of another approach (not sure what it might be... maybe pre-dissecting ProcessedSource into chunks with different configs once on reading). Sorry for taking your time.

bbatsov:
@zverok Do you plan to revisit this any time soon or can we close the PR for now?

zverok:
@bbatsov This one definitely might be closed. I plan the second attack to the problem for ~December, but will try a different approach.
I'll hopefully also (sooner) will extract small PR with found bugs in tests.

PR #8073, opened 2020-06-01
Avoid false-positives in Style/RedundantRegexpEscape

owst:
In free-space mode we should ignore any comments

-----------------

* [x] Added tests.


lib/rubocop/cop/mixin/regexp_literal_help.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # Common functionality for handling Regexp literals.
+    module RegexpLiteralHelp
```
bbatsov, line 6: I'll merge this as is, because I'm planning to cut a release today, but this has to be moved to a note extension in `rubocop-ast`.

Issue #8076, opened 2020-06-01
Infinite corrections in some of our cops?


marcandre:
If we change all the default of `expect_correction(loop: false)` to `loop: true` there seem to be  some infinite loop corrections in our test base... Are these cops conflicting in their corrections? are these potential issues?

bquorning:
I have been looking a bit in to this, and there seems to be

- issues with my changes in #8062, and
- actual bugs in a few cops.

I am working on a PR and/or some issues to open. I hope I will have them ready by tomorrow.

Issue #8075, opened 2020-06-01
expect_correction & successive transformations


marcandre:
As a followup to #8062, when testing for a cop that relies on successive autocorrections:

* should we be more specific with spec'ing the intermediate steps?
* or should we spec that a step leads to the expected correction (as @pirj suggests)?
* or should we not always test only using the final stable step?

bquorning:
I have found only two specs (and both in the same file) that separately tests the result after _one_ autocorrect and the result after looped autocorrects: https://github.com/rubocop-hq/rubocop/blob/774fe63486f6aa947dfe2e4f1d122c4fa48db080/spec/rubocop/cop/style/conditional_assignment_assign_to_condition_spec.rb#L46-L99 and https://github.com/rubocop-hq/rubocop/blob/774fe63486f6aa947dfe2e4f1d122c4fa48db080/spec/rubocop/cop/style/conditional_assignment_assign_to_condition_spec.rb#L2197-L2243

I have of course also found some tests that produce different results when looping autocorrects vs. running just one autocorrect. Some of these are quite silly, e.g. https://github.com/rubocop-hq/rubocop/blob/774fe63486f6aa947dfe2e4f1d122c4fa48db080/spec/rubocop/cop/layout/space_inside_array_literal_brackets_spec.rb#L522-L535 which says “Do not use space inside array brackets” and then promptly adds an offending space at the _other_ end of the array.

PR #8090, opened 2020-06-02
Always loop when testing autocorrection

bquorning:
Related to #8062, #8075 and #8076, this PR changes `#expect_correction` to always do looped autocorrection.

For most cops and specs, this won't change anything. The offense is corrected in one go, and the second iteration doesn't change anything.

For a few specs, something will change. A few specs have to explicitly pass `loop: false` to test the result of the first round of autocorrect, and another few specs need to pass it to avoid exceptions such as `InfiniteCorrectionLoop`. And that's the real value of using looped autocorrection in the specs – we will find more errors before releasing them to our users.

(Also in this PR: Fixing two bugs in `#expect_correction`, discovered when enabling the looping)

-----------------

* [x] Added tests.
* [-] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [-] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

### DIFF
--- a/lib/rubocop/rspec/expect_offense.rb
+++ b/lib/rubocop/rspec/expect_offense.rb
@@ -115,7 +115,7 @@ module RuboCop
         expect(actual_annotations.to_s).to eq(expected_annotations.to_s)
       end
 
-      def expect_correction(correction, loop: false)
+      def expect_correction(correction, loop: true)
         raise '`expect_correction` must follow `expect_offense`' unless @processed_source
 
         iteration = 0
--- a/lib/rubocop/rspec/expect_offense.rb
+++ b/lib/rubocop/rspec/expect_offense.rb
@@ -128,6 +128,7 @@ module RuboCop
 
           break corrected_source unless loop
           break corrected_source if cop.corrections.empty?
+          break corrected_source if corrected_source == @processed_source.buffer.source
 
           if iteration > RuboCop::Runner::MAX_ITERATIONS
             raise RuboCop::Runner::InfiniteCorrectionLoop.new(@processed_source.path, [])
@@ -135,6 +136,8 @@ module RuboCop
 
           # Prepare for next loop
           cop.instance_variable_set(:@corrections, [])
+          # Cache invalidation. This is bad!
+          cop.instance_variable_set(:@token_table, nil)
           @processed_source = parse_source(corrected_source,
                                            @processed_source.path)
           _investigate(cop, @processed_source)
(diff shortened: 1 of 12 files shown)


lib/rubocop/rspec/expect_offense.rb
```diff
@@ -128,13 +128,16 @@ def expect_correction(correction, loop: false)
...
+          break corrected_source if corrected_source == @processed_source.buffer.source
 
           if iteration > RuboCop::Runner::MAX_ITERATIONS
             raise RuboCop::Runner::InfiniteCorrectionLoop.new(@processed_source.path, [])
           end
 
           # Prepare for next loop
           cop.instance_variable_set(:@corrections, [])
+          # Cache invalidation. This is bad!
+          cop.instance_variable_set(:@token_table, nil)
```
marcandre, line 140: The mixin `SurroundingSpace` mixin should define `investigate` (`on_before_walk` after my PR lands) to clear its instance variables. If you prefer you can leave things like this since I'll have to tweak it in one case or the other...

lib/rubocop/rspec/expect_offense.rb
```diff
@@ -128,13 +128,16 @@ def expect_correction(correction, loop: false)
 
           break corrected_source unless loop
           break corrected_source if cop.corrections.empty?
+          break corrected_source if corrected_source == @processed_source.buffer.source
```
pirj, line 131: 👍

lib/rubocop/rspec/expect_offense.rb
bquorning: If `SurroundingSpace` implements `investigate`, we might want to also implement `investigate` in `Cop` to reset the `@corrections` array. But then all existing `investigate` methods should also call `super`…

Knowing that you have some changes coming, I’d prefer to just leave the changes as they are, even with the “This is bad!” comment. WDYT?

marcandre: > we might want to also implement investigate in Cop to reset the @corrections array

My refactor indeed does that automatically, but not in an official callback that could be mistakenly overriden. I will add in the code documentation that it is recommended to call super anyways though.

> I’d prefer to just leave the changes as they are

Sure, as I wrote either way is fine by me.

marcandre:
Cool.

I just rebased my refactor branch and found [some interesting consequences](https://github.com/rubocop-hq/rubocop/pull/7868/commits/37057350eb34e8cd73a3e755eac4920a05105935)

PR #8093, opened 2020-06-03
[Fix #8083] Fix an error for `Lint/MixedRegexpCaptureTypes`

### LINKED ISSUE #8083: Lint/MixedRegexpCaptureTypes and "premature end of pattern"
New cops, relying on [Regexp parser](https://github.com/ammar/regexp_parser) might fell on something that is _valid_ Ruby regexp. My example:

```ruby
response.body.scan(/data = ({"words":.+}}}[^}]*})/m)
```
Here we are doing a pretty nasty thing (in specs): scanning HTML-with-JS rendered view, find one JSON object there (and then parse and check this object). But this is still valid Ruby, and it "just works".

When `Lint/MixedRegexpCaptureTypes` tries to scan it, it falls with the "Premature end of pattern" exception in the scanner.

The deal is, formally, in Ruby regexps `{}` is for quantifiers (for ex., `x{15}`), but if it is _not_ a valid quantifier:
* Ruby falls back to treat `{}` as just text
* regexp-parser decides to fail: https://github.com/ammar/regexp_parser/issues/15 "It raises exceptions now, keep it like this. But document the fact that regexp_parser does not support each MRI quirk."

regexp-parser is kinda "theoretically right", but it leads to what is perceived as a "bug" by Rubocop users. It seems like one of the solutions below might be reasonable:
* catch such errors in cop, and report them specially (something like "non-standard-conforming Regexp")
* persuade regexp-parser's to follow all Ruby quirks, either by default, or maybe in special "compatibility" mode, maybe even reporting non-standard stuff as warnings

```
$ [bundle exec] rubocop -V
0.85.0 (using Parser 2.7.1.3, rubocop-ast 0.0.3, running on ruby 2.6.3 x86_64-linux)
```

koic:
Fixes #8083.

This PR fixes an error for `Lint/MixedRegexpCaptureTypes` cop when using a regular expression that cannot be processed by regexp_parser gem.
https://github.com/ammar/regexp_parser

Probably false negatives will be made, however false negatives will be reduced when a processing pattern of regexp_parser increases.

So this PR relies on regexp_parser for error resolution.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## 0.85.0 (2020-06-01)
 
+### Bug fixes
+
+* [#8083](https://github.com/rubocop-hq/rubocop/issues/8083): Fix an error for `Lint/MixedRegexpCaptureTypes` cop when using a regular expression that cannot be processed by regexp_parser gem. ([@koic][])
+
 ### New features
 
 * [#6289](https://github.com/rubocop-hq/rubocop/issues/6289): Add new `CheckDefinitionPathHierarchy` option for `Naming/FileName`. ([@jschneid][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Looks good! Thanks!

PR #8095, opened 2020-06-03
Registry improvements

marcandre:
This PR:
* moves registry-related globals to Registry (while maintaining compatibility)
* makes `Registry` dupable
* makes `Registry` lazy enough to allow delisting (useful for base classes, see #8023)
* adds `with_temporary_global` for easy testing that won't affect the global registry

Prelude to autocorrect refactor #7868

### DIFF
--- a/lib/rubocop/cop/cop.rb
+++ b/lib/rubocop/cop/cop.rb
@@ -23,7 +23,7 @@ module RuboCop
     #       # Do custom processing
     #     end
     #   end
-    class Cop
+    class Cop # rubocop:disable Metrics/ClassLength
       extend RuboCop::AST::Sexp
       extend NodePattern::Macros
       include RuboCop::AST::Sexp
@@ -46,6 +46,10 @@ module RuboCop
 
       def self.inherited(subclass)
         Registry.global.enlist(subclass)
+      end
+
+      def self.exclude_from_registry
+        Registry.global.dismiss(self)
       end
 
       def self.badge
--- a/lib/rubocop/cop/cop.rb
+++ b/lib/rubocop/cop/cop.rb
@@ -44,22 +44,8 @@ module RuboCop
       attr_reader :config, :offenses, :corrections
       attr_accessor :processed_source # TODO: Bad design.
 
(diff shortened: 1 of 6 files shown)


lib/rubocop/cop/cop.rb
```diff
@@ -246,6 +236,23 @@ def external_dependency_checksum
         nil
       end
 
+      ### Deprecated registry access
+
+      # Deprecated. Use Registry.global
```
bbatsov, line 241: Might be a good idea to put some YARD annotations on the deprecated methods.

lib/rubocop/cop/registry.rb
```diff
@@ -24,34 +24,40 @@ def initialize(name, origin, badges)
...
 
       def enlist(cop)
-        @registry[cop.badge] = cop
-        @departments[cop.department] ||= []
-        @departments[cop.department] << cop
-        @cops_by_cop_name[cop.cop_name] << cop
+        @enrollment_queue << cop
+      end
+
+      def delist(cop)
```
bbatsov, line 42: I think "dismiss" or something along those lines would fit better the overall terminology. :-) Not a native speaker, so I'm not sure if there are better opposites of "enlist" that we can consider.

bbatsov:
Overall I'm fine with the suggested changes. Might be a good idea to use more YARD comments for the API methods, as it seems this class is almost undocumented today, which isn't great for public API.

lib/rubocop/cop/registry.rb
marcandre: Dismiss is good (not a native speaker either 😸)

lib/rubocop/cop/cop.rb
marcandre: Thanks, didn't know there was a special tag

marcandre:
Anyone has a clue why `config_loader_spec.rb:1237` and `config_loader_spec.rb:1226` are suddenly failing?

The spec is badly written and the error message is uninformative... A higher level matcher that checks STDERR and the exist status would be much more helpful.

PR #8105, opened 2020-06-06
[WIP] Minor optimizations

andrykonchin:
Addresses the issue https://github.com/rubocop-hq/rubocop/issues/8022

**NOTE** I didn't analyze the optimizations deeply so some of them can be occasional and will be dropped. Some code can be too quick and straightforward so will be polished and rewritten later.

Profiled Rubocop with stackprof and ran against following Ruby projects:
- Rails
- Chef
- Puppet
- aws-ruby-sdk
- Rubocop

Rubocop was run with disabled cache and ignored `rubocop.yml` config of selected Ruby projects

```shell
exe/rubocop --force-default-config --cache false --out rubocop.out <DIR>
```

### Current result

Despite the largest bottlenecks were optimized the general effect is not so big and total time of `rubocop` command  is decreased only by 5-10%.

-----------------

* [ ] Added tests.


lib/rubocop/cop/mixin/classish_length.rb
```diff
@@ -23,14 +23,38 @@ def code_length(node)
...
+          #puts line_range
+          line_numbers.concat(line_range.to_a)
+          #puts "line numbers #{line_numbers}"
         end
 
-        line_numbers.to_a
+        line_numbers.uniq
+      end
+
+      def _each_descendant(node, *types, &block)
```
marcandre, line 45: Are this and the next method used at all?

lib/rubocop/cop/cop.rb
```diff
@@ -142,10 +148,15 @@ def find_location(node, loc)
         loc.is_a?(Symbol) ? node.loc.public_send(loc) : loc
       end
 
+      # TODO: remove - it's unused
```
marcandre, line 151: `Cop::Cop` is being heavily refactored, all these changes will conflict, but thanks for reminding me to optimize `duplicate_location?` as [`Source::Range` can now be used as hash keys / set elements](https://github.com/whitequark/parser/pull/675)

lib/rubocop/cop/layout/line_length.rb
```diff
@@ -213,7 +213,7 @@ def excess_range(uri_range, line, line_index)
         end
 
         def max
-          cop_config['Max']
+          @max ||= cop_config['Max']
```
marcandre, line 216: I highly doubt this makes much of a difference, as `cop_config` is already cached (as it should) and the rest is just a hash lookup with a frozen string literal which is highly optimized.

lib/rubocop/cop/mixin/tokens_optimized.rb
```diff
@@ -0,0 +1,17 @@
+module RuboCop
```
marcandre, line 1: Right, building a tokens index should make a good difference. I'm curious if we really need to use tokens though. If we do, the best would be to share the cache with a `Force`, even though they haven't really be designed for this.

bbatsov:
> Despite the largest bottlenecks were optimized the general effect is not so big and total time of rubocop command is decreased only by 5-10%.

5-10% are not little. :-)

andrykonchin:
@marcandre Thank you for your review. Will come back to this PR soon.

lib/rubocop/cop/cop.rb
andrykonchin: @marcandre Would you mind my opening a separate PR with this optimization if it's a straightforward solution?

PR #8126, opened 2020-06-09
[Fix #8124] Fix a false positive for `Lint/FormatParameterMismatch`

### LINKED ISSUE #8124: Escaped % being flagged as invalid when matched with other format with latest 0.85.1
PR https://github.com/rubocop-hq/rubocop/pull/8042 Introduced a change that flagged mismatched format as an offense, but I believe it doesn't take into account the escaped %. Something like `%% %{foo}s` should be valid as it has a named component, but the first part is just escaping a % sign.

## Expected behavior

No offense when pairing a named format with an escaped %. Other combinations with an escaped % may need to be checked.

## Actual behavior

Offense flagged with `Lint/FormatParameterMismatch: Format string is invalid because formatting sequence types (numbered, named or unnumbered) are mixed`

Something like `format('%%%<hex>02X', hex: 10)` -> %0A should be valid. This format string does not throw errors when run.

## RuboCop version

```
$ rubocop -V
0.85.1 (using Parser 2.7.1.3, rubocop-ast 0.0.3, running on ruby 2.6.4 x86_64-darwin18)
```

koic:
Fixes #8124.

This PR fixes a false positive for `Lint/FormatParameterMismatch` using named parameters with escaped `%`.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Bug fixes
 
 * [#8115](https://github.com/rubocop-hq/rubocop/issues/8115): Fix false negative for `Lint::FormatParameterMismatch` when argument contains formatting. ([@andrykonchin][])
+* [#8124](https://github.com/rubocop-hq/rubocop/issues/8124): Fix a false positive for `Lint/FormatParameterMismatch` when using named parameters with escaped `%`. ([@koic][])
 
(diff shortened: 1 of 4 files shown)


bbatsov:
Thanks!

PR #8136, opened 2020-06-10
[Fix #8098]: Fix false-positive in Style/RedundantRegexpCharacterClass

### LINKED ISSUE #8098: Style/RedundantRegexpCharacterClass flags single characters in interpolations
## Expected behavior

The Style/RedundantRegexpCharacterClass should not flag or fix single characters in brackets in interpolations. Interpolations with bracket notation are typically used to access objects and should not be treated as single-element classes.

## Actual behavior

The Style/RedundantRegexpCharacterClass flags and fixes single characters in brackets in interpolations. Interpolations with bracket notation is viewed as a single-element class and has the brackets removed. This results in the Lint/Syntax cop flagging the change due to unrecognized characters in the interpolation. See below.

Take this trivial script:

```ruby
string = "one.two.three"
regexp = /#{string.split('.')[0]}/

puts regexp.match('one')
```

The expected behavior is that the regexp should resolve to `/one/`. However, running the Style/RedundantRegexpCharacterClass will flag `[0]` as a single-element class even though it is valid notation for accessing the first element of an array. Attempting to auto-correct with this cop results in the following change to the file:

```ruby
string = "one.two.three"
regexp = /#{string.split('.')0}/

puts regexp.match('one')
```

The interpolation `string.split('.')0` has invalid syntax and results in the Lint/Syntax cop flagging the change due to an unexpected integer:

```shell
E: Lint/Syntax: unexpected token tINTEGER
(Using Ruby 2.5 parser; configure using TargetRubyVersion parameter, under AllCops)
      regex = /#{string.split('.')0}/
```

## RuboCop version

```shell
$ rubocop -V
0.85.0 (using Parser 2.7.1.3, rubocop-ast 0.0.3, running on ruby 2.6.3 x86_64-darwin19)
```

owst:
Also correct the handling of interpolations and comments - we need to
preserve their width in the checked pattern source without their
contents triggering the cop in question, so that we can correctly
highlight following offenses based on the offset in the pattern_source.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * [#8115](https://github.com/rubocop-hq/rubocop/issues/8115): Fix false negative for `Lint::FormatParameterMismatch` when argument contains formatting. ([@andrykonchin][])
 * [#8124](https://github.com/rubocop-hq/rubocop/issues/8124): Fix a false positive for `Lint/FormatParameterMismatch` when using named parameters with escaped `%`. ([@koic][])
+* [#8098](https://github.com/rubocop-hq/rubocop/issues/8098): Fix a false positive for `Style/RedundantRegexpCharacterClass` when using interpolations. ([@owst][])
 
 ## 0.85.1 (2020-06-07)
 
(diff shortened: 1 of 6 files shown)


bbatsov:
Thanks!

PR #8143, opened 2020-06-12
DidYouMean::SpellChecker may not be available

### LINKED ISSUE #7979: uninitialized constant DidYouMean::SpellChecker
TLDR: In c92a591, the name similarity module started using `DidYouMean::SpellChecker` but this gem is not explicitly listed as a dependency.

--------

## Expected behavior

Rubocop does not crash. We get an offense such as:
`Unnecessary disabling of Lint/SuperCoolCop (did you mean Layout/SuperCoolCop?)`

## Actual behavior

We see the following error when running Rubocop:
```
Uninitialized constant DidYouMean::SpellChecker
--
  | /var/www/vendor/bundle/ruby/2.6.0/gems/rubocop-0.83.0/lib/rubocop/name_similarity.rb:18:in `find_similar_names'
  | /var/www/vendor/bundle/ruby/2.6.0/gems/rubocop-0.83.0/lib/rubocop/name_similarity.rb:9:in `find_similar_name'
```

Add a magic comment to disable a cop that does not actually exist. E.g. `# rubocop:disable Lint/SuperCoolCop`

## RuboCop version

```
$ bundle exec rubocop -V
0.83.0 (using Parser 2.7.1.2, running on ruby 2.6.6 x86_64-darwin19)
```

bquorning:
Fixes #7979 

In #7979 it was reported that `DidYouMean::SpellChecker` may not always be available. We can work around that by doing a feature check before using the `SpellChecker` class.

I did not add any tests.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/name_similarity.rb
+++ b/lib/rubocop/name_similarity.rb
@@ -12,6 +12,12 @@ module RuboCop
     end
 
     def find_similar_names(target_name, names)
+      # DidYouMean::SpellChecker is not available in all versions of Ruby, and
+      # even on versions where it *is* available (>= 2.3), it is not always
+      # required correctly. So we do a feature check first.
+      # See: https://github.com/rubocop-hq/rubocop/issues/7979
+      return [] unless defined?(DidYouMean::SpellChecker)
+
       names = names.dup
       names.delete(target_name)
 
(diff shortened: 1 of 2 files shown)


lib/rubocop/name_similarity.rb
```diff
@@ -12,6 +12,8 @@ def find_similar_name(target_name, names)
     end
 
     def find_similar_names(target_name, names)
+      return [] unless defined?(DidYouMean::SpellChecker)
```
koic, line 15: It would be helpful to link a reference (#8143) to the reason for the workaround :-)

bquorning: Good point. I will add a commit.

bbatsov:
Thanks!

PR #8146, opened 2020-06-12
Use UTC in Rubocop todo file generation

mauro-oto:
Most of our team is distributed, so this is what the diff for our todo file
usually looks like:

```
- # on 2020-06-05 11:48:43 +1000 using RuboCop version 0.80.1.
+ # on 2020-06-12 12:23:47 -0300 using RuboCop version 0.80.1.
```

I think it'd be nice to just use UTC and avoid changing timestamp and timezones
so much depending on who generates the file.

I'm aware we can disable the timestamp generation, but I think it's nice to
have. Should this be a similar on/off config, or do you agree it's a saner
default?

Thank you!


bbatsov:
The suggested change is fine by me, although I guess it deserves a changelog entry.

marcandre:
> The suggested change is fine by me, although I guess it deserves a changelog entry.

What's the official criteria for deserving a changelog entry? I thought it was only user-facing changes, which I think might be overly restrictive, but this PR is not user-facing and I'd venture to say that even devs don't need to know; there's nothing that needs to be done differently than before.

bbatsov:
It's user-facing in a way that next time a person regens the TODO they'll likely get a different TZ. Probably many people won't notice it, but I guess some will. Not a big deal for this PR.

marcandre:
> It's user-facing in a way that next time a person regens the TODO they'll likely get a different TZ. Probably many people won't notice it, but I guess some will. Not a big deal for this PR.

Indeed, brain fart, sorry.

mauro-oto:
@bbatsov sure, I wasn't sure if it needed one either, I just created a changelog entry for it, thanks!

marcandre:
Thanks @mauro-oto 👍

PR #8147, opened 2020-06-12
Add new `Style/RedundantFetchBlock` cop

fatkodima:
Transferred from https://github.com/rubocop-hq/rubocop-performance/pull/126

Related reference - https://github.com/JuanitoFatas/fast-ruby#hashfetch-with-argument-vs-hashfetch--block-code

When block for `fetch` just returns `Numeric`, `Symbol`, `Rational`, `Complex`, `true`, `false`, `nil`, `String` or constant it is shorter (and faster) to pass an argument instead of using a block.

```ruby
# bad
hash.fetch(:key) { 5 }
hash.fetch(:key) { true }
hash.fetch(:key) { false }
hash.fetch(:key) { nil }
array.fetch(5) { :value }
ENV.fetch(:key) { VALUE }

# good
hash.fetch(:key, 5)
hash.fetch(:key, true)
hash.fetch(:key, false)
hash.fetch(:key, nil)
array.fetch(5, :value)
ENV.fetch(:key, VALUE)
```

When hash *contains* a key, performance is the same for argument vs block syntaxes, but when hash is *missing* a key - argument version is faster.

### Benchmark

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  gem 'benchmark-ips'
end

hash_with_key = { key: 1 }
hash_without_key = { not_a_key: 1 }

def fast(hash)
  hash.fetch(:key, 1)
end

def slow(hash)
  hash.fetch(:key) { 1 }
end

Benchmark.ips do |x|
  x.report('fast hash_with_key') { fast(hash_with_key) }
  x.report('slow hash_with_key') { slow(hash_with_key) }
  x.compare!
end

Benchmark.ips do |x|
  x.report('fast hash_without_key') { fast(hash_without_key) }
  x.report('slow hash_without_key') { slow(hash_without_key) }
  x.compare!
end
```

### Results

```
  fast hash_with_key     1.043M i/100ms
  slow hash_with_key   996.735k i/100ms
  fast hash_with_key     10.292M (± 2.9%) i/s -     52.149M in   5.072099s
  slow hash_with_key      9.940M (± 1.5%) i/s -     49.837M in   5.015103s

Comparison:
  fast hash_with_key: 10291722.7 i/s
  slow hash_with_key:  9939687.7 i/s - same-ish: difference falls within error

fast hash_without_key
                         1.017M i/100ms
slow hash_without_key
                       659.264k i/100ms
fast hash_without_key
                         10.344M (± 0.5%) i/s -     51.892M in   5.016881s
slow hash_without_key
                          6.663M (± 1.6%) i/s -     33.622M in   5.047510s

Comparison:
fast hash_without_key: 10343873.0 i/s
slow hash_without_key:  6663049.0 i/s - 1.55x  (± 0.00) slower
```


config/default.yml
```diff
@@ -3596,6 +3596,18 @@ Style/RedundantException:
   VersionAdded: '0.14'
   VersionChanged: '0.29'
 
+Style/RedundantFetchBlock:
+  Description: >-
+                  Use `fetch(key, value)` instead of `fetch(key) { value }`
+                  when value has Numeric, Rational, Complex, Symbol or String type, `false`, `true`, `nil` or is a constant.
+  Reference: 'https://github.com/JuanitoFatas/fast-ruby#hashfetch-with-argument-vs-hashfetch--block-code'
+  Enabled: true
```
bbatsov, line 3604: This should say "pending". You should also mark the cop as unsafe, as there's a change someone has another method named `fetch` in their code.

lib/rubocop/cop/style/redundant_fetch_block.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # This cop identifies places where `fetch(key) { value }`
+      # can be replaced by `fetch(key, value)`.
+      #
+      # In such cases `fetch(key, value)` method is faster
```
bbatsov, line 9: You should add some examples of the "SafeForConstants" here as well.

fatkodima:
Updated with suggestions.

config/default.yml
```diff
@@ -3597,6 +3597,19 @@ Style/RedundantException:
...
+                  Use `fetch(key, value)` instead of `fetch(key) { value }`
+                  when value has Numeric, Rational, Complex, Symbol or String type, `false`, `true`, `nil` or is a constant.
+  Reference: 'https://github.com/JuanitoFatas/fast-ruby#hashfetch-with-argument-vs-hashfetch--block-code'
+  Enabled: 'pending'
+  Safe: false
+  VersionAdded: '0.86'
+  SafeForConstants: false
+  # If enabled, this cop will autocorrect usages of
+  # `fetch` being called with block returning a constant.
+  # This can be dangerous since constants will not be defined at that moment.
```
koic, line 3611: This is a nitpick, Can you update to the following change?

```suggestion
  # If enabled, this cop will autocorrect usages of
  # `fetch` being called with block returning a constant.
  # This can be dangerous since constants will not be defined at that moment.
  SafeForConstants: false
  VersionAdded: '0.86'
```

fatkodima:
@koic Updated.

bbatsov:
Thanks!

Drenmi:
Since this is only concerned with the performance of one case, and not the style of `#fetch` in general, it should probably have gone to `rubocop-performance`. 😅

PR #8153, opened 2020-06-14
Use `match?` instead of `=~`

koic:
This PR uses `match?` instead of `=~`.

In future this change would be better if it could be detected by the extended `Performance/RegexpMatch`. So far, it is detected and corrected by handcraft.

-----------------

* [x] Added tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -21,7 +21,7 @@ Dir['tasks/**/*.rake'].each { |t| load t }
 desc 'Run RuboCop over itself'
 RuboCop::RakeTask.new(:internal_investigation).tap do |task|
   if RUBY_ENGINE == 'ruby' &&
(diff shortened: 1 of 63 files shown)


andrykonchin:
Looks pretty good but there is a potential issue with such replacing because it isn't safe.

The following case works well without error:

```ruby
nil =~ // # => nil
```

But calling `match?` will lead to `NoMethodError` exception.

That's why `match?` is calling on regexp instead of string here:
- https://github.com/rubocop-hq/rubocop/pull/8153/files#diff-5532de1b82bbcce6397d5fbaec6e6ddf
- https://github.com/rubocop-hq/rubocop/pull/8153/files#diff-938b46da0d108c71380ac9427b83dde0

koic:
True, it's safer to put regexp on the LHS and preferable to prevent potential errors. I've updated to safe changes this PR.

bbatsov:
Nice!

PR #8155, opened 2020-06-14
Add helper constants for nonmutating method names

fatkodima:
For context https://github.com/rubocop-hq/rubocop-performance/pull/140/files#r439732580

Bullet number 4
> and those ENUMERABLE_METHOD_NAMES, NONMUTATING_ARRAY_METHODS and NONMUTATING_HASH_METHODS into rubocop-asts AST::CollectionNode module?

But it is not really belongs to `rubocop-ast`, I was wrong. So opened pr here.


marcandre:
>  it is not really belongs to `rubocop-ast`, I was wrong.

Why? IMO all criteria that are independent of a particular cop, that can be factually said to be accurate outside of a context, should be in `rubocop-ast`, these included.

fatkodima:
Ok. So is this file a good place for those constants https://github.com/rubocop-hq/rubocop-ast/blob/master/lib/rubocop/ast/node/mixin/method_identifier_predicates.rb ?
And if so, should I also add pedicate methods like `nonmutable_array_method?`, etc there?

marcandre:
I think so

fatkodima:
Opened on `rubocop-ast` https://github.com/rubocop-hq/rubocop-ast/pull/38

PR #8159, opened 2020-06-16
Add new `CountAsOne` option for code length related `Metric` cops

### LINKED ISSUE #7724: Metrics/MethodLength should ignore/configure configuration lists
Rubocop's `Metrics/MethodLength` prefers

```ruby
  def build_filter_chain
    [ByCategory.new(@category_id.presence),
     ByMarketing.new(@promotion.presence, @novelty.presence),
     ByPlainScopes.new(featured: @featured, random: @random, slugged: @slugged, unslugged: @unslugged,
                       with_volume_discounts: @with_volume_discounts, with_images: @with_images),
     ByBrand.new(@brand_id.presence),
     BySupplier.new(@supplier_id.presence),
     ByCustomerHistory.new(customer, @historical),
     ByHavingShop.new(@shop),
     BySearch.new(@q, customer)]
  end
```

over

```ruby
  def build_filter_chain
    [
      ByCategory.new(@category_id.presence),
      ByMarketing.new(@promotion.presence, @novelty.presence),
      ByPlainScopes.new(
        featured: @featured,
        random: @random,
        slugged: @slugged,
        unslugged: @unslugged,
        with_volume_discounts: @with_volume_discounts,
        with_images: @with_images
      ),
      ByBrand.new(@brand_id.presence),
      BySupplier.new(@supplier_id.presence),
      ByCustomerHistory.new(customer, @historical),
      ByHavingShop.new(@shop),
      BySearch.new(@q, customer)
    ]
  end
```

I find the second one much more readable and would like if there was a possibility to opt-out from `Metrics/MethodLength` for configuration lists. This means, I'd count an array as 1 line. 

Does this sound sane? I don't propose a new cop, but a new configuration option like:
`CountAsOne` defaulting to `[]`. Possible values could be `Array`, `Hash`, `Heredoc`.

We would do the same thing for these if wanted: https://github.com/rubocop-hq/rubocop/issues/5394, https://github.com/rubocop-hq/rubocop/issues/1077, https://github.com/rubocop-hq/rubocop/issues/2686 and many more.

fatkodima:
Closes #7724 

This new setting allows to specify which ruby construction (`array`, `hash` and `heredoc` are currently implemented) we want to treat as one line.

The algorithm from the high level:
1. we calculate the total length for the whole code of specified node
2. then, for each the most top level descendant of foldable type, we returning back its code length to the total length
3. and so after steps 1 and 2 we get a resulted length


Drenmi:
Great start!

I have been thinking about a similar setting myself, but the point I kept getting stuck at is the fact that arrays and hashes can have arbitrary amounts of code in them.

In the test cases for this configuration option, all the arrays and hashes are assumed to be shallow collections of primitive values, which is the case where this makes sense. But I must assume this introduces some false negatives? 🤔

fatkodima:
> In the test cases for this configuration option, all the arrays and hashes are assumed to be shallow collections of primitive values

No, the code works for arbitrary complex arrays/hashes. For example, the multiline array in the original issue would be folded into 1 line. 
Looks like I missed to add test cases for more complex examples.

> But I must assume this introduces some false negatives? 🤔

Am I answered also this question or should you provide an example than, if not?

docs/modules/ROOT/pages/cops_metrics.adoc
```diff
@@ -45,13 +45,12 @@ and https://en.wikipedia.org/wiki/ABC_Software_Metric.
...
 | 0.44
-| 0.66
+| 0.86
 |===
 
-This cop checks if the length of a block exceeds some maximum value.
-Comment lines can optionally be ignored.
-The maximum allowed length is configurable.
-The cop can be configured to ignore blocks passed to certain methods.
+You can set construction types you want to fold with `CountAsOne`.
```
bbatsov, line 51: Probably "literals" would make more sense here than "construction types".

lib/rubocop/cop/metrics/utils/code_length_calculator.rb
```diff
@@ -0,0 +1,129 @@
...
+            node.each_child_node do |child|
+              if types.include?(child.type)
+                yield child
+              else
+                each_top_level_descendant(child, *types, &block)
+              end
+            end
+          end
+
+          def classish_node?(node)
```
bbatsov, line 96: classlike :-)

bbatsov:
Yeah, it'd be nice if there are some examples in the cop's description about using this new config.

fatkodima:
Updated with suggestions.

bbatsov:
Thanks!

colinlieberman:
@fatkodima - I know this is 8 months old now, but I found it today, updated rubocop to the latest version, and am very, very happy. Thank you for this

PR #8164, opened 2020-06-18
Support autocorrection for `Lint/InterpolationCheck`

koic:
This PR supports autocorrection for `Lint/InterpolationCheck` and marks it unsafe because autocorrected code is incompatible.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/interpolation_check.rb
+++ b/lib/rubocop/cop/lint/interpolation_check.rb
@@ -30,6 +30,19 @@ module RuboCop
           add_offense(node)
         end
 
+        def autocorrect(node)
+          lambda do |corrector|
+            starting_token, ending_token = if node.source.include?('"')
+                                             ['%{', '}']
+                                           else
+                                             ['"', '"']
+                                           end
+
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/lint/interpolation_check.rb
```diff
@@ -30,6 +30,13 @@ def on_str(node)
           add_offense(node)
         end
 
+        def autocorrect(node)
+          lambda do |corrector|
+            corrector.replace(node.loc.begin, '"')
+            corrector.replace(node.loc.end, '"')
```
robotdana, line 36: i think this will fail when correcting a string like `'"#{}"'`. it may also change the output of a string like `%{\{#{}}` (where what once was a string that equaled `"{"` it's now equal to `"\{"`. These are guesses without running it though

koic: Sure! I updated to use `%{...}` if a offending string contains double quoted strings.

bbatsov:
Thanks!

PR #8175, opened 2020-06-20
Add AllowedCompactTypes option for Style/RaiseArgs

pdobb:
## Style/RaiseArgs -- Add `AllowedCompactTypes` Option
```yaml
Style/RaiseArgs
  AllowedCompactTypes: [] # default
```

Add custom exception names that should be ignored when
Style/RaiseArgs investigates code with `EnforcedStyle: exploded`.
This is useful for certain exception types where we really do want an
instance of an exception to be raised. e.g. in Rails:

```ruby
raise ActiveRecord::RecordInvalid.new(my_model)
```

Or to raise a custom error type that wraps an object.

To allow the above to pass even while `EnforcedStyle: exploded` is set
while still not allowing other Exception types:

```yaml
Style/RaiseArgs
  AllowedCompactTypes:
    - 'ActiveRecord::RecordInvalid'
    - 'MyCustomWrappedError'
```

---

Image 2020-06-27 16-26-48

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -3912,11 +3912,12 @@ Style/RaiseArgs:
   StyleGuide: '#exception-class-messages'
   Enabled: true
   VersionAdded: '0.14'
-  VersionChanged: '0.40'
+  VersionChanged: '1.2'
   EnforcedStyle: exploded
   SupportedStyles:
     - compact # raise Exception.new(msg)
     - exploded # raise Exception, msg
+  AllowedCompactTypes: []
 
 Style/RandomWithOffset:
   Description: >-
--- a/spec/rubocop/cop/style/raise_args_spec.rb
+++ b/spec/rubocop/cop/style/raise_args_spec.rb
@@ -255,6 +255,36 @@ RSpec.describe RuboCop::Cop::Style::RaiseArgs, :config do
       end
     end
 
+    context 'when exception type is in AllowedCompactTypes' do
+      before do
+        stub_const('Ex1', StandardError)
+        stub_const('Ex2', StandardError)
+      end
+
+      let(:cop_config) do
+        {
+          'EnforcedStyle' => 'exploded',
+          'AllowedCompactTypes' => ['Ex1']
+        }
+      end
+
+      it 'accepts exception constructor with no arguments' do
+        expect_no_offenses('raise Ex1.new')
+      end
+
+      context 'with one argument' do
+        it 'accepts exception constructor' do
+          expect_no_offenses('raise Ex1.new(msg)')
+        end
+      end
+
+      context 'with more than one argument' do
+        it 'accepts exception constructor' do
+          expect_no_offenses('raise Ex1.new(arg1, arg2)')
+        end
+      end
+    end
+
     it 'accepts exception constructor with more than 1 argument' do
       expect_no_offenses('raise MyCustomError.new(a1, a2, a3)')
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#8983](https://github.com/rubocop-hq/rubocop/pull/8983): Support auto-correction for `Naming/HeredocDelimiterCase`. ([@koic][])
 * [#8004](https://github.com/rubocop-hq/rubocop/issues/8004): Add new `GitHubActionsFormatter` formatter. ([@lautis][])
+* [#8175](https://github.com/rubocop-hq/rubocop/pull/8175): Add new `AllowedCompactTypes` option for `Style/RaiseArgs`. ([@pdobb][])
 
 ### Bug fixes
 
@@ -5058,3 +5059,4 @@
 [@matthieugendreau]: https://github.com/matthieugendreau
 [@miry]: https://github.com/miry
 [@lautis]: https://github.com/lautis
+[@pdobb]: https://github.com/pdobb
--- a/docs/modules/ROOT/pages/cops_style.adoc
+++ b/docs/modules/ROOT/pages/cops_style.adoc
@@ -7686,7 +7686,7 @@ p = proc { |n| puts n }
 | Yes
 | Yes
 | 0.14
-| 0.40
+| 1.2
 |===
 
 This cop checks the args passed to `fail` and `raise`. For exploded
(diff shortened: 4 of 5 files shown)


bbatsov:
I understand the reasoning behind the proposed change, but I'm a bit wary of adding an option that affects only one of the styles. Let's see what the others think about the change as well.

config/default.yml
```diff
@@ -3569,11 +3569,12 @@ Style/RaiseArgs:
   StyleGuide: '#exception-class-messages'
   Enabled: true
   VersionAdded: '0.14'
-  VersionChanged: '0.40'
+  VersionChanged: '0.87'
   EnforcedStyle: exploded
   SupportedStyles:
     - compact # raise Exception.new(msg)
     - exploded # raise Exception, msg
+  AllowedNonExplodedTypes: []
```
bbatsov, line 3577: Probably we should stick to the "compact" terminology here, if we decide to go forward with this change.

pdobb: I can get behind that, thanks. I was trying to convey that it only affects the "Exploded" style. But I also agree with your comment that it's awkward that this option only affects one of the styles.

spec/rubocop/cop/style/raise_args_spec.rb
```diff
@@ -237,6 +237,60 @@ def do_something
...
+    context 'when exception type is in AllowedNonExplodedTypes' do
+      before do
+        stub_const('Ex1', StandardError)
+        stub_const('Ex2', StandardError)
+      end
+
+      let(:cop_config) do
+        {
+          'EnforcedStyle' => 'exploded',
+          'AllowedNonExplodedTypes' => [%w[Ex1 Ex2], ['Ex1']].sample
```
jonas054, line 249: I don't think we have any other specs where some of the set-up is random. We usually do shared examples and test both (all) cases every time.

pdobb: My intention was to show that all of the options are equivalent, without wasting the processing time. But, I understand that this isn't common so I'm happy to switch to 2 different tests.

pdobb:
@jonas054 Updated as a fixup commit for easy diff verification. Will squash if accepted.

pirj:
> This is useful for certain exception types where we really do want an instance of an exception to be raised. e.g. in Rails:

```ruby
raise ActiveRecord::RecordInvalid.new(my_model)
```

Sorry, doesn't it do the same as:
```ruby
raise ActiveRecord::RecordInvalid, my_model
```

It would be quite cumbersome to list all the exceptions that accept arguments in `.rubocop.yml`.

pdobb:
@pirj Yes, I understand / you're right that one can spell out `raise ActiveRecord::RecordInvalid, my_model`. For my team:
1. Certain custom exception types with specialized initializers just feel wrong to "explode", while others do not
2. Our preference is not to have to rely on that 3rd attribute *sometimes* being a parameter that is passed to the exception's initializer and *sometimes* just a message. We want to be explicit.

I just used ActiveRecord::RecordInvalid as an example because it is a well known exception type that takes an argument. But basically I'm trying to get enforceability of the "exploded" style while still letting my team tell the story they want to tell with certain custom exception types.

pirj:
> Our preference is not to have to rely on that 3rd attribute sometimes being a parameter that is passed to the exception's initializer and sometimes just a message. We want to be explicit.

As far as I recollect, the third argument passed to `raise` won't be delegated to exception initializer anyway.

So:
```ruby

# good
raise E
raise E, "message"
raise E, model

# Ruby error!
raise E, model, "message"
# ArgumentError: wrong number of arguments (given 1, expected 2)
```

And I guess it's not RuboCop's business to deal with raise offences for what Ruby considers to be incorrect code.

Please correct me if I'm mistaken, but this example for `compact` style is wrong:
```ruby
      #   raise RuntimeError, arg1, arg2, arg3
```

pdobb:
> As far as I recollect, the third argument passed to `raise` won't be delegated to exception initializer anyway.

@pirj Yes. I said the wrong thing earlier (now corrected). I meant to say the 2nd argument.

```ruby
# Using the exploded style:

# good
raise E, "message"

# doesn't make my team happy
raise E, model

# would prefer to add an exception on the exploded style for E so that this is valid:
raise E.new(model)
```

For example:
```ruby
# If we use the default exploded style
raise E, model

# ... and then we want to add a message, we can't just do this because it won't work (ruby think `"message"` is the caller / stack trace)
raise E, model, "message"
# so we have to add code churn (change code that isn't directly related to the change we are making)
raise E.new(model), "message"

**Comment by pirj** [comment]
Agree that for
```ruby
raise MyWrappedError.new(obj), 'message'
```
there's no other way around.

So we have those cases:
```ruby
# all good
raise E, "message"

# still good, most probably it's a message
raise E, SOME_FAILURE_TEXT

# ambiguous - message or object? can lead to churn
raise E, var_with_failure_text
raise E, method_with_failure_text(error_arg_for_interpolation)

raise E.new(object), "message"
raise E.new(object), SOME_FAILURE_TEXT
raise E.new(object), method_with_failure_text
```

I wouldn't be so worried about code churn when it comes to readability, and the reader has to figure out if it's a message or an object passed to the exception.

I'm thinking about a more generic solution to this so an exclusion list becomes unnecessary. What are your thoughts on this?

pdobb:
@pirj That sounds great. It does seem basically deterministic. At least enough so that non-deterministic exceptions can be disabled on a case by case basis.

pirj:
Please don't get me wrong, @pdobb , I've started this discussion because even though this will fit your team's style quite well, it may turn out to be used exclusively by your team. This option is a code that needs to be supported and with team style evolution this feature may become orphaned.

This cop needs love and it seems it can be improved - for everyone. Will the cop be able to fit your team style and be beneficial for a wider audience?

I don't mind that addition (FYI: I'm just a bystander here, not affiliated with RuboCop core), and it's approved by the two long-time maintainers. Anyway, wondering how you see this pull request's destiny?

pirj:
@pdobb Ping

pdobb:
@pirj Sorry, life happened. I think you made great points but I don't think I have the time to learn how to implement the dynamic solution right now or any time soon. The change I made in this PR is sufficient for my team's needs still. If it's not accepted I'm OK with just continuing to disable this cop as needed in my projects.

bbatsov:
@pdobb Rebase this and I'll merge it. Seems no one has a better solution or the time to work on one.

pdobb:
@bbatsov all set, thanks

bbatsov:
Thanks!

PR #8186, opened 2020-06-22
[Fix #8185] Fix a false positive for `Style/YodaCondition`

### LINKED ISSUE #8185: YodaCondition false positive on string interpolation
Yoda condition cop is fired in a situation where the left side is an interpolated string and right hand side is a constant. It works fine if string interpolation is removed.

--------

## Expected behavior

I expect YodaCondition not to be reported when comparing an interpolated string (left side) to a constant (right side).

## Actual behavior

YodaCondition gets triggered.

The following triggers the cop: `"/#{foo}" == CONSTANT` and I don't think it should as the right hand side is constant and left hand side is not.

## RuboCop version

0.85.1

koic:
Fixes #8185.

This PR fixes a false positive for `Style/YodaCondition` when interpolation is used on the left side.

String interpolation (and Regexp interpolation) contains variables, I think it's acceptable to use it on the left hand side.

-----------------

* [x] Added tests.


bbatsov:
Looks good!

PR #8187, opened 2020-06-22
More black list and white list updates

kaoru:
**Update multiline layout specs to use DENYLIST**

Use DENYLIST rather than BLACKLIST in example code being used to test the multiline cops.

**Use DENYLISTED in BinaryOperatorParameterName cop**

Prefer DENYLIST to BLACKLIST in code as well as in cop config names.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/naming/binary_operator_parameter_name.rb
+++ b/lib/rubocop/cop/naming/binary_operator_parameter_name.rb
@@ -18,7 +18,7 @@ module RuboCop
               'name its argument `other`.'
 
         OP_LIKE_METHODS = %i[eql? equal?].freeze
-        BLACKLISTED = %i[+@ -@ [] []= << === `].freeze
+        EXCLUDED = %i[+@ -@ [] []= << === `].freeze
 
         def_node_matcher :op_method_candidate?, <<~PATTERN
           (def [#op_method? $_] (args $(arg [!:other !:_other])) _)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/naming/binary_operator_parameter_name.rb
```diff
@@ -18,7 +18,7 @@ class BinaryOperatorParameterName < Cop
               'name its argument `other`.'
 
         OP_LIKE_METHODS = %i[eql? equal?].freeze
-        BLACKLISTED = %i[+@ -@ [] []= << === `].freeze
+        DENYLISTED = %i[+@ -@ [] []= << === `].freeze
```
bbatsov, line 21: Instead of inventing new words I'd go with "EXCLUDED" here. 😉

spec/rubocop/cop/layout/multiline_method_call_indentation_spec.rb
```diff
@@ -881,7 +881,7 @@ def a
         expect_no_offenses(<<~RUBY)
           return #{keyword} receiver.nil? &&
               !args.empty? &&
-              BLACKLIST.include?(method_name)
+              DENYLIST.include?(method_name)
```
bbatsov, line 884: I can't see where's the actual constant definition that has to be renamed.

bbatsov: Ah, that's an example. My bad! Use something like "FORBIDDEN_METHODS" or whatever, please.

kaoru:
I've made those changes you suggested @bbatsov 😄

bbatsov:
Thanks!

PR #8192, opened 2020-06-23
[Fixes #8156] CLI options change: auto-correct now safe by default.

### LINKED ISSUE #8156: v1 default: Include Safe: false by default
@bbatsov wrote in #5980 about `--safe-auto-correct`
> Let's implement the metadata and the options firsts, changing the defaults is easy afterwards. 😄

 Now would be a good time to ask ourselves if the current default is the right one. safe or unsafe by default?

`rubocop -a` does all autocorrections, including unsafe ones. One has to add `--safe-auto-correct` to exclude unsafe ones.

Should this be the default, or should it be the reverse?

marcandre:
I'm proposing slightly different options than in #8156:

`-a` / `--auto-correct` are now safe only, replacing `--safe-auto-correct`
`-A` / `--auto-correct-all` apply safe and unsafe corrections

Help reads:
```
...
-a, --auto-correct               Auto-correct offenses (only when it's safe).
-A, --auto-correct-all           Auto-correct offenses (safe and unsafe)
...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changes
 
 * [#7868](https://github.com/rubocop-hq/rubocop/pull/7868): **(Breaking)** Extensive refactoring of internal classes `Team`, `Commissioner`, `Corrector`. `Cop::Cop#corrections` not completely compatible. See Upgrade Notes. ([@marcandre][])
+* [#8156](https://github.com/rubocop-hq/rubocop/issues/8156): **(Breaking)** `rubocop -a / --autocorrect` no longer run unsafe corrections; `rubocop -A / --autocorrect-all` run both safe and unsafe corrections. Options `--safe-autocorrect` is deprecated. ([@marcandre][])
 
 ### Bug fixes
 
--- a/docs/modules/ROOT/pages/usage/auto_correct.adoc
+++ b/docs/modules/ROOT/pages/usage/auto_correct.adoc
@@ -4,7 +4,9 @@ In auto-correct mode, RuboCop will try to automatically fix offenses:
 
 [source,sh]
 ----
-$ rubocop -a
+$ rubocop -A
+# or
+$ rubocop --auto-correct-all
 ----
 
 There are a couple of things to keep in mind about auto-correct:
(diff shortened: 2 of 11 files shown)


koic:
I think this is a practical change, and `-a` and `-A` are easy to understand 👍 
Could you add a changelog entry to mention to users that `-a` option's default behavior is a breaking change?

> If this is a breaking change, mark it with `**(Breaking)**`.

bbatsov:
Might be a good idea to keep the old option around with some deprecation message, or at least show a message when people try to invoke it.

marcandre:
I've added back "--safe-autocorrect" with a warning "--safe-autocorrect is deprecated; use --autocorrect".
Changelog entry added too.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Changes
 
 * [#7868](https://github.com/rubocop-hq/rubocop/pull/7868): **(Breaking)** Extensive refactoring of internal classes `Team`, `Commissioner`, `Corrector`. `Cop::Cop#corrections` not completely compatible. See Upgrade Notes. ([@marcandre][])
+* [#8156](https://github.com/rubocop-hq/rubocop/pull/8156): **(Breaking)** `rubocop -a / --autocorrect` no longer run unsafe corrections; `rubocop -A / --autocorrect-all` run both safe and unsafe corrections. Options `--safe-autocorrect` is deprecated. ([@marcandre][])
```
koic, line 12: This is a small feedback :-)

```suggestion
* [#8156](https://github.com/rubocop-hq/rubocop/issues/8156): **(Breaking)** `rubocop -a / --autocorrect` no longer run unsafe corrections; `rubocop -A / --autocorrect-all` run both safe and unsafe corrections. Options `--safe-autocorrect` is deprecated. ([@marcandre][])
```

bbatsov:
You should also update https://docs.rubocop.org/rubocop/0.86/usage/auto_correct.html

marcandre:
> You should also update https://docs.rubocop.org/rubocop/0.86/usage/auto_correct.html

Oops, right. Done.

bbatsov:
Thanks!

PR #8203, opened 2020-06-23
[Fix #8189] Consider spaces for Layout/MultilineBlockLayout

### LINKED ISSUE #8189: Layout/MultilineBlockLayout triggering if single line would result in exactly 121 characters
This scope is giving a Layout/MultilineBlockLayout cop warning:

```
scope :xxxxxxxxxxxxxxx, lambda {
  |yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy, zzzzzzzzzzzz, wwwwwwwwwwwwwwwwwwwwwwwwww|
}
```

If I add a single character to any of the variables then the warning goes away, this is likely due to (I'm guessing) the cop not triggering if the single line version would be too long (over 120 characters).

The problem is, that the single line version is already 121 characters.

It seems when doing the calculations for how long a single line would be, it's forgetting that I would need to add a space after the `{`. This is confirmed when I add a trailing space, then the warning goes away (however another warning appears)

--------

## Expected behavior

This should not give any warnings:
```
scope :xxxxxxxxxxxxxxx, lambda {
  |yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy, zzzzzzzzzzzz, wwwwwwwwwwwwwwwwwwwwwwwwww|
}
```

This should give a warning:
```
scope :xxxxxxxxxxxxxxx, lambda {
  |yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy, zzzzzzzzzzzz, wwwwwwwwwwwwwwwwwwwwwwwwwww|
}
```
(difference being one extra `w`)

## Actual behavior

Both examples gave a warning.

Simply paste the example scope into a script and run rubocop.

Try this as well to confirm that it gives a line length warning which is why I had to break it up into 2 lines:

```
scope :xxxxxxxxxxxxxxx, lambda { |yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy, zzzzzzzzzzzz, wwwwwwwwwwwwwwwwwwwwwwwwww|
}
```

## RuboCop version

```
$ rubocop -V
0.86.0 (using Parser 2.7.1.0, rubocop-ast 0.0.3, running on ruby 2.7.0 x86_64-linux)
```

KNejad:
Before this commit when deciding whether to trigger a Layout/MultilineBlockLayout it would not consider any extra spaces required to join the lines. This commit fixes that issue.

For example the below would trigger a LayoutMultilineBlockLayout warning before but not anymore:

```
some_result = lambda do
  |so_many, parameters, it_will, be_too_long, for_one_line, line_length, has_increased, add_3_more|
  do_something
end
```

~I had to disable the AbcSize metric for the method I was extending, as much as I tried I couldn't get a nice method with an Abc < 15. The best I got was 15.07 :unamused:. I could get it < 15 but IMHO it looked worse and seemed to go against the spirit of the cop.~ (Thanks to @jonas054 I fixed this issue)

**Note:** I also updated the max_line_length to default to the new 120 value if nothing is defined

Resolves https://github.com/rubocop-hq/rubocop/issues/8189

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/multiline_block_layout.rb
+++ b/lib/rubocop/cop/layout/multiline_block_layout.rb
@@ -95,11 +95,22 @@ module RuboCop
         end
 
         def line_break_necessary_in_args?(node)
-          needed_length = node.source_range.column +
-                          node.source.lines.first.length +
-                          block_arg_string(node, node.arguments).length +
-                          PIPE_SIZE
-          needed_length > max_line_length
+          needed_length_for_args(node) > max_line_length
+        end
+
+        def needed_length_for_args(node)
+          node.source_range.column +
+            characters_needed_for_space_and_pipes(node) +
+            node.source.lines.first.chomp.length +
+            block_arg_string(node, node.arguments).length
+        end
+
+        def characters_needed_for_space_and_pipes(node)
+          if node.source.lines.first.end_with?("|\n")
+            PIPE_SIZE
+          else
+            1 + PIPE_SIZE * 2
+          end
         end
 
         def add_offense_for_expression(node, expr, msg)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 
 ### Bug fixes
 
+* [#8189](https://github.com/rubocop-hq/rubocop/issues/8189): Fix an error for `Layout/MultilineBlockLayout` where spaces for a new line where not considered. ([@knejad][])
 * [#8252](https://github.com/rubocop-hq/rubocop/issues/8252): Fix a command line option name from `--safe-autocorrect` to `--safe-auto-correct`, which is compatible with RuboCop 0.86 and lower. ([@koic][])
 * [#8259](https://github.com/rubocop-hq/rubocop/issues/8259): Fix false positives for `Style/BisectedAttrAccessor` when accessors have different access modifiers. ([@fatkodima][])
 * [#8253](https://github.com/rubocop-hq/rubocop/issues/8253): Fix false positives for `Style/AccessorGrouping` when accessors have different access modifiers. ([@fatkodima][])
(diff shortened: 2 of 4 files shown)


lib/rubocop/cop/layout/multiline_block_layout.rb
```diff
@@ -94,13 +94,24 @@ def args_on_beginning_line?(node)
             node.loc.begin.line == node.arguments.loc.last_line
         end
 
+        # rubocop:disable Metrics/AbcSize
```
jonas054, line 97: Did you consider making `needed_length` a method to get the `AbcSize` down? E.g.,

```ruby
needed_length_for_args(node) > max_line_length
```

KNejad:
Thanks for the review @jonas054 I made the changes you suggested.

I'm really not sure what I did but I somehow screwed everything up with git and the PR got closed so I had to reverse my work and do a bit of git magic to get it all working again. The PR is ready now so please merge it if you are happy with everything :+1:

marcandre:
Looks like a good PR.

Still needs to be squashed...

KNejad:
@marcandre done! Out of curiosity why can't you just use the GitHub "squash and merge" functionality to prevent the manual work? Can that cause issues?

marcandre:
> @marcandre done! Out of curiosity why can't you just use the GitHub "squash and merge" functionality to prevent the manual work? Can that cause issues?

You are right, we could and maybe I should not have asked 😅. No issues other than potentially fiddling with the commit message.

bbatsov:
Thanks, everyone! :bowing_man: 

> @marcandre done! Out of curiosity why can't you just use the GitHub "squash and merge" functionality to prevent the manual work? Can that cause issues?

This was godsend to me, as before this I often had to wait a lot for people or manually checkout branches and squash myself. These days I only ask people to do squashing themselves if I feel that a PR shouldn't be reduced to a single commit - e.g. the raw variant has 10 and we can probably have 5-6 meaningful commits there.

PR #8204, opened 2020-06-23
Tweaks to Metrics/PerceivedComplexity

marcandre:
`Metrics/PerceivedComplexity` currently ignores `else` branches in `case` statement which seems like a clear bug, considering the weight if gives to `else` in if statements:

```ruby
if foo
  bar
else
  baz
end
# counts for more than the equivalent:
case
when foo
  bar
else
  baz
end
```

For the rest, there's no clear definition of what `PerceivedComplexity` is  supposed to do, but it sounds like it's the same thing as `CyclomaticComplexity` with a twist for `case` and `if` with `else`.

This PR fixes the `else` issue and makes the same changes that #8149 made for CyclomaticComplexity (`||=`, iterating methods, etc.) Default bumped also from 7 to 8. 3 new todos all of which are already CyclomaticComplexity todos.

If we're ok with these fixes, it would be best for v1 as it is breaking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 * [#8338](https://github.com/rubocop-hq/rubocop/pull/8338): **potentially breaking**. Config#for_department now returns only the config specified for that department; the 'Enabled' attribute is no longer calculated. ([@marcandre][])
 * [#8037](https://github.com/rubocop-hq/rubocop/pull/8037): **(Breaking)** Cop `Metrics/AbcSize` now counts ||=, &&=, multiple assignments, for, yield, iterating blocks. `&.` now count as conditions too (unless repeated on the same variable). Default bumped from 15 to 17. Consider using `rubocop -a --disable-uncorrectable` to ease transition. ([@marcandre][])
 * [#8276](https://github.com/rubocop-hq/rubocop/issues/8276): Cop `Metrics/CyclomaticComplexity` not longer counts `&.` when repeated on the same variable. ([@marcandre][])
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 ### Changes
 
 * [#7868](https://github.com/rubocop-hq/rubocop/pull/7868): **(Breaking)** Extensive refactoring of internal classes `Team`, `Commissioner`, `Corrector`. `Cop::Cop#corrections` not completely compatible. See Upgrade Notes. ([@marcandre][])
+* [#8149](https://github.com/rubocop-hq/rubocop/pull/8149): **(Breaking)** Cop `Metrics/CyclomaticComplexity` now counts `&.`, `||=`, `&&=` and blocks known to iterate. Default bumped from 6 to 7. Consider using `rubocop -a --disable-uncorrectable` to ease transition. ([@marcandre][])
+
```
koic, line 13: Can you update the issue link and mentioned default value?

```suggestion
* [#8149](https://github.com/rubocop-hq/rubocop/issues/8149): **(Breaking)** Cop `Metrics/CyclomaticComplexity` now counts `&.`, `||=`, `&&=` and blocks known to iterate. Default bumped from 7 to 8. Consider using `rubocop -a --disable-uncorrectable` to ease transition. ([@marcandre][])
```

marcandre: Oops, copy-pasted and forgot to update. Fixed

marcandre:
@jonas054 what do you think of this?

bbatsov:
@marcandre Go ahead and merge this at your discretion.

PR #8206, opened 2020-06-24
[Fix #8205] Avoid false positive in Style/RedundantRegexpCharacterClass

### LINKED ISSUE #8205: False positive for Style/RedundantRegexpCharacterClass
When using an escaped square bracket in a character class, rubocop suggests incorrect fix.

--------

## Expected behavior

`[(\[]TM[\])]` passes the RedundantRegexpCharacterClass check.

## Actual behavior

```
C: Style/RedundantRegexpCharacterClass: Redundant single-element character class, [\] can be replaced with \.
```

Run the check on a line with:
```
/\b ?[(\[]TM[\])]/i
```

## RuboCop version

```
0.86.0 (using Parser 2.7.1.4, rubocop-ast 0.0.3, running on ruby 2.6.6 x86_64-darwin18)
```

owst:
If the first element of a character class was an escaped-] our pattern
was treating it as a single element classs containing a single backslash
(which is invalid, and incorrect).

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/redundant_regexp_character_class_spec.rb
+++ b/spec/rubocop/cop/style/redundant_regexp_character_class_spec.rb
@@ -43,6 +43,19 @@ RSpec.describe RuboCop::Cop::Style::RedundantRegexpCharacterClass do
     end
   end
 
+  context 'with a character class containing an escaped [' do
+    it 'registers an offense and corrects' do
+      expect_offense(<<~'RUBY')
+        foo = /[\[]/
+               ^^^^ Redundant single-element character class, `[\[]` can be replaced with `\[`.
+      RUBY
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/redundant_regexp_character_class_spec.rb
```diff
@@ -43,6 +43,19 @@
     end
   end
 
+  context 'with a character class containing an escaped [' do
```
owst, line 46: This test wasn't strictly related to this PR, but I noticed we were missing it despite having the equivalent one for an escaped `]`

koic:
Thank you for fixing!

PR #8210, opened 2020-06-24
Fix a false positive for `Style/RedundantFetchBlock` when using with `Rails.cache`

### LINKED ISSUE #8196: False positive for `Style/RedundantFetchBlock` cop with `Rails.cache.fetch`
I work in a Rails project in which we have a few instances of
```ruby
Rails.cache.fetch('something') { 'value_for_cache_miss' }
```

The `Style/RedundantFetchBlock` cop shouldn't trigger in this case since
```ruby
Rails.cache.fetch('something', 'value_for_cache_miss')
```
isn't the same. It's documented [here](https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html#method-i-fetch). Anything following the first parameter is considered to be an option. The value for a cache miss has to be passed as a block.

--------

## Expected behavior

I expect RuboCop to not register offenses for `Rails.cache.fetch` as explained above.

## Actual behavior

An offense for the `Style/RedundantFetchBlock` cop is registered. This is a false positive.

With `Rails.cache.fetch('something') { "value_for_cache_miss" }`

## RuboCop version

```
0.86.0 (using Parser 2.7.1.4, rubocop-ast 0.0.3, running on ruby 2.5.7 x86_64-linux-gnu)
```

fatkodima:
Closes #8196

### DIFF
--- a/lib/rubocop/cop/style/redundant_fetch_block.rb
+++ b/lib/rubocop/cop/style/redundant_fetch_block.rb
@@ -46,8 +46,7 @@ module RuboCop
 
         def on_block(node)
           redundant_fetch_block_candidate?(node) do |send, body|
-            return if body&.const_type? && !check_for_constant?
-            return if body&.str_type? && !check_for_string?
+            return if should_not_check?(send, body)
 
             range = fetch_range(send, node)
             good = build_good_method(send, body)
@@ -81,6 +80,16 @@ module RuboCop
         def const_type?(node)
           node&.const_type?
         end
+
+        def should_not_check?(send, body)
+          (body&.const_type? && !check_for_constant?) ||
+            (body&.str_type? && !check_for_string?) ||
+            rails_cache?(send.receiver)
+        end
+
+        def_node_matcher :rails_cache?, <<~PATTERN
+          (send (const _ :Rails) :cache)
+        PATTERN
 
         def fetch_range(send, node)
           range_between(send.loc.selector.begin_pos, node.loc.end.end_pos)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/redundant_fetch_block.rb
```diff
@@ -82,6 +81,16 @@ def const_type?(node)
           node&.const_type?
         end
 
+        def should_not_check?(send, body)
+          (body&.const_type? && !check_for_constant?) ||
+            (body&.str_type? && !check_for_string?) ||
+            rails_cache?(send.receiver)
```
koic, line 87: I was late for the discussion of #8196.

I agree with the following @bbatsov's opinion.

> Fixing this is going to be tricky, due to the dynamic nature of Ruby - we can never be sure about the type of a receiver. The best thing that comes to mind is to have some configuration about allowed receivers or whatever, as I don't want to hardcode some special handling of Rails.cache. //cc @fatkodima

Ideally RuboCop core should be Rails unaware because RuboCop core focuses on Ruby. It is better to create `IgnoredReceivers` option and configure it with RuboCop Rails (or users).

marcandre: I think we all agree as to what is the best. I see this PR as a temporary improvement, before a `IgnoredReceivers` option.

koic: Yeah, I get it! I'll merge this PR.

koic:
Thanks!

PR #8211, opened 2020-06-24
Detect class_variable_set in Style/ClassVars

biinari:
`class_variable_set(:@@name, 2)` has the same effect as `@@name = 2`

This variation can be within the class definition:

```ruby
class A
  class_variable_set(:@@name, 2)
end
```

Or it could be called elsewhere, with the class as receiver:

```ruby
class A; end

class B
  def change(name, value)
    A.class_variable_set(":@@#{name}", value)
  end
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/docs/modules/ROOT/pages/cops_style.adoc
+++ b/docs/modules/ROOT/pages/cops_style.adoc
@@ -951,6 +951,15 @@ class A
   @@test = 10
 end
 
+class A
+  def self.test(name, value)
+    class_variable_set("@@#{name}", value)
+  end
+end
+
+class A; end
+A.class_variable_set(:@@test, 10)
+
 # good
 class A
   @test = 10
@@ -959,6 +968,12 @@ end
 class A
   def test
     @@test # you can access class variable without offense
+  end
+end
+
+class A
+  def self.test(name)
+    class_variable_get("@@#{name}") # you can access without offense
   end
 end
 ----
--- a/lib/rubocop/cop/style/class_vars.rb
+++ b/lib/rubocop/cop/style/class_vars.rb
@@ -19,6 +19,15 @@ module RuboCop
       #     @@test = 10
       #   end
       #
+      #   class A
+      #     def self.test(name, value)
+      #       class_variable_set("@@#{name}", value)
+      #     end
+      #   end
+      #
+      #   class A; end
+      #   A.class_variable_set(:@@test, 10)
+      #
       #   # good
       #   class A
       #     @test = 10
@@ -30,12 +39,24 @@ module RuboCop
       #     end
       #   end
       #
+      #   class A
+      #     def self.test(name)
+      #       class_variable_get("@@#{name}") # you can access without offense
+      #     end
+      #   end
+      #
       class ClassVars < Cop
         MSG = 'Replace class var %<class_var>s with a class ' \
               'instance var.'
 
         def on_cvasgn(node)
           add_offense(node, location: :name)
+        end
+
+        def on_send(node)
+          return unless node.method?(:class_variable_set)
+
+          add_offense(node.first_argument)
         end
 
         def message(node)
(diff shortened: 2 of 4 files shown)


lib/rubocop/cop/style/class_vars.rb
```diff
@@ -38,10 +38,23 @@ def on_cvasgn(node)
           add_offense(node, location: :name)
         end
 
+        def on_send(node)
+          return unless class_var_set?(node)
+
+          _, _, name, = *node
+          add_offense(name)
+        end
```
koic, line 46: It seems to be simple as follows.

```diff
diff --git a/lib/rubocop/cop/style/class_vars.rb b/lib/rubocop/cop/style/class_vars.rb
index e4772213e..e50bc5506 100644
--- a/lib/rubocop/cop/style/class_vars.rb
+++ b/lib/rubocop/cop/style/class_vars.rb
@@ -39,22 +39,15 @@ module RuboCop
         end

         def on_send(node)
-          return unless class_var_set?(node)
+          return unless node.method?(:class_variable_set)

-          _, _, name, = *node
-          add_offense(name)
+          add_offense(node.arguments.first)
         end

         def message(node)
           class_var, = *node
           format(MSG, class_var: class_var)
         end
-
-        private
-
-        def_node_matcher :class_var_set?, <<~PATTERN
-          (send _ :class_variable_set _ _)
-        PATTERN
       end
     end
   end
```

koic:
Can you add this bad case to the example code?

lib/rubocop/cop/style/class_vars.rb
biinari: Nice. Still getting to grips with the different ways of matching nodes. Thanks

docs/modules/ROOT/pages/cops_style.adoc
```diff
@@ -951,6 +951,12 @@ class A
   @@test = 10
 end
 
+class A
+  class_variable_set(:@@test, 10)
+end
+
+A.class_variable_set(":@@#{name}", 10)
```
koic, line 958: I think this example will occur a `NameError`.

```suggestion
A.class_variable_set("@@#{name}", 10)
```

docs/modules/ROOT/pages/cops_style.adoc
```diff
@@ -961,6 +967,12 @@ class A
     @@test # you can access class variable without offense
   end
 end
+
+class A
+  def test(name)
+    class_variable_get(":@@#{name}") # you can access without offense
```
koic, line 973: I think this example will occur a `NoMethodError`.

```ruby
class Foo
  def do_something
    class_variable_get('@@var') # you can access without offense
  end
end

Foo.new.do_something
```

```console
% ruby example.rb
        1: from example.rb:7:in `<main>'
example.rb:3:in `do_something': undefined method `class_variable_get' for #<Foo:0x00007feb088407d8> (NoMethodError)
```

biinari: Sorry, I should have actually run my own examples before committing. Updated to use class methods and for each bad/good example to be complete.

docs/modules/ROOT/pages/cops_style.adoc
biinari: Good spot. Sometimes comments are the hardest place to get code right first time... :)

biinari:
Thanks for the review @koic. Rebased from master now. Looks like the Changelog has somehow ended up with two `Changes` sections in the unreleased part. Would you like me to merge those together?

koic:
Yeah, "Changes" section has been merged on the master branch.
https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md#master-unreleased

biinari:
Ah thanks. Now I see why it didn't get thrown up as a conflict. That changelog entry should be right now.

koic:
Thanks 👍

PR #8212, opened 2020-06-25
Fix spec from symlink: remove bundle env

biinari:
I have checked out rubocop to a directory under a symlink. When I try to
run the specs from the symlink instead of the original directory, odd
behaviour ensues.

Ruby resolves the directory for `'rubocop/version'` to different
locations. When ruby first starts, the symlink location is loaded. After
`require 'rubocop'`, ruby loads the original directory.

Presumably, this has to do with the gemspec requiring `rubocop/version`
and require_relative being used in `lib/rubocop.rb` but that is common
practice.

To fix this for the specs, I call upon bundler's
`with_unbundled_env` (which used to be `with_clean_env`). Then for ruby
commands in `:isolated_environment`, use `bundle exec` with this project's
`Gemfile`.

-----------------

* [ ] Added tests.


biinari:
My original thoughts were to apply a fix that is only on the spec side as it's clearly not an issue that troubles the normal use of rubocop. I'm closing this as I can quite easily run from the original directory instead of from a symlink.

PR #8214, opened 2020-06-26
Fix Style/RedundantParentheses with hash literal as first argument to yield

karlwithak:
Fixes https://github.com/rubocop-hq/rubocop/issues/8208

Do not consider the parentheses around a yielded hash literal as redundant.


koic:
Thanks!

PR #8220, opened 2020-06-27
Add cop to enforce passing a pattern to Enumerable predicate methods all?/any?/none?/one?

sebastian-palma:
This PR adds a Style/PatternArgument cop that enforces passing the argument of the `===` method call in the block body when using any of the current Enumerable predicate methods all?/any?/none?/one?. As shown [here](https://bugs.ruby-lang.org/issues/11286).

E.g.:

```ruby
[1, 2, 3].any? { |e| Set[1, 2, 3] === e }
[1, 2, 3].one? { |e| Range.new(1, 10) === e }
```

with:

```ruby
[1, 2, 3].any?(Set[1, 2, 3])
[1, 2, 3].one?(Range.new(1, 10))
```

It includes its autocorrect, which takes into consideration beginless and endless range - introduced in Ruby 2.6 and 2.7 correspondingly.

This cop should only be run on Ruby 2.5 or greater.


bbatsov:
Should this cop cover additional methods like `include?`?

sebastian-palma:
That might be an option @bbatsov, also for `member?` and any other aliases for `===`. Do we consider adding it?

bbatsov:
Seems to me this would make the cop more useful, as I doubt many people are using `===` directly these days. Didn't we also had some cop that would flag `Set[1, 2, 3] === e`? I have some memory there was something discouraging the use of `===` in general.

sebastian-palma:
Nice! I'm going to check which others classes implement `===` and add them to the cop. I just ran this new cop against the master branch of Rails and there were a few uses of `===`.

For that we have the `Style/CaseEquality`, and because `===` isn't widely use we could see some other uses, like `=~` or `match?` when using any of these 4 enumerable predicate methods.

I've seen sometimes `match?` used to check if the receiver matches a regular expression, but in that simple context the benefits of using `match?` aren't really necessary, we could maybe use instead `=~` which result can be evaluated against its truthy value, and in the same case reduce that to `regex === string`, where we can take the regex and pass it as the pattern.

Although limited, we could advise passing the method param as the any?/all?/none?/one? pattern when using `=~`, `match?` and/or `match`.

sebastian-palma:
Hey @bbatsov, just added some changes to include Set, Range, and Regexp support. Would you mind taking a look?

lib/rubocop/cop/style/pattern_argument.rb
```diff
@@ -0,0 +1,122 @@
...
+        end
+
+        def regexp?(body, receiver)
+          obj, = *receiver
+          _, obj_class = *obj
+          (receiver.regexp_type? || obj_class == :Regexp) &&
+            %i[=== =~ match match?].include?(body.method_name)
+        end
+
+        def set?(body, receiver)
```
marcandre, line 97: Have you considered using NodePattern for these kind of checks? It would probably be more readable.

Also it looks like `set?` is missing some logic, as `:Set` does not appear in the source code...

sebastian-palma: Thanks @marcandre, the truth is I just followed some other cops while working on this one. I updated it to use the NodePattern and I got rid of all those checks, amazing.

I'm not sure if there's a better way to get [this](https://github.com/rubocop-hq/rubocop/pull/8220/files#diff-c0aa68a59ced803500d8a79422f5ee7fR48) done.

lib/rubocop/cop/style/pattern_argument.rb
```diff
@@ -0,0 +1,74 @@
...
+      #   # good
+      #   %w[foo bar foobar].any?(/oo/)
+      #
+      #   [1, 2, 3].any?(1..10)
+      #
+      #   [1, 2, 3].any?(Set[1, 2, 3])
+      #
+      #   [1, 2, 3].none?(Integer)
+      #
+      class PatternArgument < Cop
```
marcandre, line 31: Could you inherit from `Base` instead? See [the docs](https://docs.rubocop.org/rubocop/0.87/v1_upgrade_notes.html)

lib/rubocop/cop/style/pattern_argument.rb
```diff
@@ -0,0 +1,74 @@
...
+            add_offense(node,
+                        location: method.loc.selector,
+                        message: format(MSG, method: method.method_name, pattern: pattern.source))
+          end
+        end
+
+        def autocorrect(node)
+          lambda do |corrector|
+            _, replacement = pattern_argument_candidate?(node).map(&:source)
+            replacement = "(#{replacement})" unless replacement.match?(/^\(.+\)$/)
```
marcandre, line 48: Tricky. I was thinking of checking for `begin_type`, but this looks pretty safe like this.

lib/rubocop/cop/style/pattern_argument.rb
```diff
@@ -0,0 +1,74 @@
...
+        end
+
+        private
+
+        MSG = 'Pass `%<pattern>s` as an argument to `%<method>s` instead of a block.'
+
+        def_node_matcher :pattern_argument_candidate?, <<~PATTERN
+          (block $({send csend} _ {:all? :any? :none? :one?}) (:args _) {
+            (send ${(begin ({irange erange} _ _))
+                    (send (const nil? :Range) :new _ _)
```
marcandre, line 60: `const {nil? cbase}` is better. We'll soon have `global_const?(:Range)` which will be even better.

lib/rubocop/cop/style/pattern_argument.rb
```diff
@@ -0,0 +1,74 @@
...
+            corrector.replace(block_range_with_space(node), replacement)
+          end
+        end
+
+        private
+
+        MSG = 'Pass `%<pattern>s` as an argument to `%<method>s` instead of a block.'
+
+        def_node_matcher :pattern_argument_candidate?, <<~PATTERN
+          (block $({send csend} _ {:all? :any? :none? :one?}) (:args _) {
```
marcandre, line 58: I like it.

In case you don't know, you could create an individual matcher for all possibilites and then have your candidate matcher being "(... {#range? #set? #regexp?})". Not sure if it would be better or not here, just letting you know.

spec/rubocop/cop/style/pattern_argument_spec.rb
```diff
@@ -0,0 +1,263 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Style::PatternArgument, :config do
+  context 'when using Ruby 2.5 or newer', :ruby25 do
+    %w[all? any? none? one?].map { |method| [method, '^' * method.length] }.each do |method, arrows|
```
marcandre, line 5: No need for arrows... Use `^{method}` and pass `method: method` to expect_offense 😄

lib/rubocop/cop/style/pattern_argument.rb
sebastian-palma: I did that, also updated the tests to use `expect_correction`, even better. Just don't know if the autocorrect is fine now [right inside `on_block`](https://github.com/rubocop-hq/rubocop/pull/8220/files#diff-c0aa68a59ced803500d8a79422f5ee7fR44).

lib/rubocop/cop/style/pattern_argument.rb
sebastian-palma: I moved that to use `RuboCop::AST::NodePattern.match`, since ranges can be endless, beginless, floats, integers, etc. and anything enclosed in parentheses would match that query. I thought the best tool is to use NodePattern. And to avoid destructuring, first we'd have to check the replacement is `begin_type?`, then check the element inside is `range_type?`, isn't it?

lib/rubocop/cop/style/pattern_argument.rb
sebastian-palma: I got it working with something like this:

```ruby
def_node_matcher :literal_range?, '(begin ({irange erange} _ _))'
def_node_matcher :new_range?, '(send (const {nil? cbase} :Range) :new _ _)'
def_node_matcher :set?, '(send (const {nil? cbase} :Set) {:[] :new} ...)'
def_node_matcher :literal_regexp?, '(:regexp (:str _) _)'
def_node_matcher :new_regexp?, '(send (const {nil? cbase} :Regexp) :new _)'

def_node_matcher :pattern_argument_candidate?, <<~PATTERN
  (block $({send csend} _ {:all? :any? :none? :one?}) (:args _) {
    (send ${#literal_range? #new_range? #set?} {:=== :include? :member?} ...)
    (send ${#literal_regexp? #new_regexp?} {:=== :=~ :match :match?} ...)
    (send $(const nil? _) :=== _)
    (send lvar :is_a? $(const nil? _))})
PATTERN
```

For some reason I can't combine _literal_ and _new_ node_matchers into a single one, I got "_Couldn't compile due to unbalanced pattern. Pattern: #literal_range? #new_range?_"

marcandre: You probably forgot the `{}`. You must write `def_node_matcher :range?, '{#literal_range? #new_range?}'`

lib/rubocop/cop/style/pattern_argument.rb
marcandre: Yes, that's perfect. Except you already have the `method` and `pattern`, so you should avoid calling it again in `replacement`...

sebastian-palma: Nice, I was wondering there was something strange there 🎉

lib/rubocop/cop/style/pattern_argument.rb
sebastian-palma: Thanks! Just split them, I think that way it's easier to read.

lib/rubocop/cop/style/pattern_argument.rb
```diff
@@ -0,0 +1,83 @@
...
+            ) do |corrector|
+              corrector.replace(block_range_with_space(node), replacement(pattern))
+            end
+          end
+        end
+
+        private
+
+        MSG = 'Pass `%<pattern>s` as an argument to `%<method>s` instead of a block.'
+        RANGE_PATTERN = '(begin ({irange erange} {nil? _} {nil? _}))'
```
marcandre, line 52: `{nil? _}` == `_`

lib/rubocop/cop/style/pattern_argument.rb
```diff
@@ -0,0 +1,83 @@
...
+          (block $({send csend} _ {:all? :any? :none? :one?}) (:args _) {
+            (send ${#range? #set?} {:=== :include? :member?} ...)
+            (send ${#regexp?} {:=== :=~ :match :match?} ...)
+            (send $(const nil? _) :=== _)
+            (send lvar :is_a? $(const nil? _))})
+        PATTERN
+
+        def replacement(pattern)
+          source = pattern.source
+          return source if RuboCop::AST::NodePattern.new(RANGE_PATTERN).match(pattern)
```
marcandre, line 71: You should never create `NodePattern`s on the fly, very ineficient. We should have an internal cop for this actually, I'll open an issue.

sebastian-palma: What do you suggest? Checking the source by using a regular expression? Or using the `begin_type?` and `range_type?` predicates?

marcandre: I'm not sure I understand. You can use `def_node_matcher` (which does the compilation once instead of everytime like here), or `begin_type?`, or your previous string test...

sebastian-palma: Ended up doing `pattern.begin_type? && [*pattern].first.range_type?` as I'm not sure what other things could just return true for `pattern.begin_type?`

sebastian-palma:
> We're getting closer.
> You need to be much stricter in your matcher though.
> Right now, there are many false positives: `any? {|x| Set[...].include?(foo(x)}` or `any? {|x, y| Set[foo(y), bar(y)].include?(x)}`or `any? {|_, x| Set[...].include?(x)}` for example.
> You'll have to be strict about the `args` signature, as well about the `lvar` and the absence of that `lvar` anywhere else.
> 
> Reminder: NodePattern supports unification and passing parameters (including unified values)

Nice @marcandre, I modified the node matcher to check the object used for block method is a local variable (lvar - lvar_type?, can they work interchangeably?), that make the tests pass, but I think the best is to check the argument given for the block object is the same as the yielded value from the block:

```ruby
[1, 2, 3].any? { |x| Set[1, 2, 3].include?(foo(x)) } # foo(x) isn't lvar_type? but send_type?; x != foo(x)
[1, 2, 3].any? { |x| Set[1, 2, 3].include?(x) }         # pass by checking x is lvar;                    x == x
```

With the examples you mention, yielding more than one variable in the block, that's covered by the `(:args _)` right after checking the methods used on the receiver `(block $({send csend} _ {:all? :any? :none? :one?}) (:args _) ...)`.

As per your reminder, I'm not sure I got it, I'm pretty new to the NodePattern. Thanks for checking these changes and the help ;)

bbatsov:

PR #8235, opened 2020-07-03
Add new `Style/RedundantAssignment` cop

### LINKED ISSUE #7373: Detect useless assignment and return
Given this code, I'd like RuboCop to detect that the assignment is useless and autocorrect the code to remove the assignment.

```ruby
# bad
def foo(input)
  output = input.sort
  output
end

# good
def foo(input)
  input.sort
end
```

I'm not sure if this is a new cop or an addition to the existing `Link/UselessAssignment` cop.

fatkodima:
Closes #7373 

Can confirm this is a popular pattern - I have seen those many times. Also I have found 3 offenses in `rails` codebase and 1 in `rubocop`s own.
 
This cop finds the pattern recursively, not just as last 2 expressions in method's body.


marcandre:
Cool 😄 

I didn't test it but I fear your code will erroneously flag the following, right?

```ruby
def foo
  bar = something

  bar
ensure
  something_else(bar)
end
```

fatkodima:
@marcandre Oh, wow, great catch! 💪 

Changed handling of ensure nodes and added test case for this.

bbatsov:
I like the proposed cop. I'm only wondering if we should shorten its name and drop `BeforeReturn`, as I can't imagine another case of redundant assignment - I guess anything else goes to the unused assignment category.

fatkodima:
@bbatsov Agreed, I can not imagine too. Stripped that `BeforeReturn` from the cop name.

bbatsov:
Thanks!

Issue #7776, opened 2020-03-05
Rubocop generates invalid code


adfoster-r7:
When formatting the following code, rubocop will generate invalid ruby:

```ruby
class Bar
  def initialize(*args)
  end
end

class Foo < Bar
  def initialize
    super(foo(bar,
      'ExpiresAt' => {'Hours' => 60 * 60 * 24} # 24h
    ))
  end
end
```

## Expected behavior

Not to generate invalid ruby code

## Actual behavior

Invalid ruby code is created due to the comment handling of `# 24 h`

```ruby
# frozen_string_literal: true
class Bar
  def initialize(*args); end
end

class Foo < Bar
  def initialize
    super(foo(bar,
              'ExpiresAt' => { 'Hours' => 60 * 60 * 24 }) # 24h)
  end
end
```

## Steps to reproduce the problem

1. Create new ruby file
2. add original code snippet
3. bundle exec rubocop -a new_file.rb
4. Run again, verify invalid ruby code has been generated

## RuboCop version

```
$ [bundle exec] rubocop -V
0.80.0 (using Parser 2.7.0.2, running on ruby 2.6.5 x86_64-darwin18)
```

bmorrall:
Had a quick look at this, running autocorrect with only the `Layout/MultilineMethodCallBraceLayout` Cop produces this error.

MattBeaty:
I'd like to help out on this project. Mind if I take a crack at this issue?

PR #8237, opened 2020-07-04
[Fix-7777-7776]: Generate valid code if comment exists before closing brace

### LINKED ISSUE #7777: Rubocop places commas into comments, generating invalid ruby code
When formatting the following code, rubocop will generate invalid ruby:

```ruby
class Bar
  def initialize(*args)
  end
end

class Foo < Bar
  def initialize
    super(
      'firstKey' => [ 'value a', #description1
                      'value b' #description2
      ],
      'secondKey' => 'Bar'
    )
  end
end
```

## Expected behavior

Not to generate invalid ruby code

## Actual behavior

Invalid ruby code is created due to incorrectly handling the comment and placing a comma on the comment line:

```ruby
class Bar
  def initialize(*args)
  end
end

class Foo < Bar
  def initialize
    super(
      'firstKey' => ['value a', # description1
                      'value b'] # description2,
      'secondKey' => 'Bar'
    )
  end
end
```

1. Create new ruby file
2. add original code snippet
3. bundle exec rubocop  -x --safe-auto-correct new_file.rb
4. Run again, verify invalid ruby code has been generated

## RuboCop version

```
$ [bundle exec] rubocop -V
0.80.0 (using Parser 2.7.0.2, running on ruby 2.6.5 x86_64-darwin18)
```

shekhar-patil:
closes #7777 and #7776 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/correctors/multiline_literal_brace_corrector.rb
+++ b/lib/rubocop/cop/correctors/multiline_literal_brace_corrector.rb
@@ -42,8 +42,34 @@ module RuboCop
 
         corrector.insert_before(
           last_element_range_with_trailing_comma(node).end,
-          node.loc.end.source
+          content_if_comment_present(corrector, node)
         )
+      end
+
+      def content_if_comment_present(corrector, node)
+        range = range_with_surrounding_space(
+          range: children(node).last.source_range,
+          side: :right
+        ).end.resize(1)
+        if range.source == '#'
+          select_content_to_be_inserted_after_last_element(corrector, node)
+        else
+          node.loc.end.source
+        end
+      end
+
+      def select_content_to_be_inserted_after_last_element(corrector, node)
+        range = range_between(
+          node.loc.end.begin_pos,
+          range_by_whole_lines(node.loc.expression).end.end_pos
+        )
+
+        remove_trailing_content_of_comment(corrector, range)
+        range.source
+      end
+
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/correctors/multiline_literal_brace_corrector.rb
```diff
@@ -42,8 +42,30 @@ def correct_next_line_brace(corrector)
...
+        end
+      end
+
+      def select_content_to_be_inserted_after_last_element(corrector, node)
+        range = range_between(
+          node.loc.end.begin_pos,
+          range_by_whole_lines(node.loc.expression).end.end_pos
         )
+
+        corrector.remove(range)
```
jonas054, line 67: If I remove this line, all tests still pass. I suspect that this is because the call to `remove` is not needed, rather that being related to poor test coverage. But I don't see why it's not needed. :confused: 

Anyway, it's better if the method only selects a certain source range, like its name says, and not does anything else (single responsibility principle).

shekhar-patil: @jonas054 , Thanks for reviewing, Yes, The mentioned line is needed. I wrote the wrong test case which I have corrected now. Please review now.

bbatsov:
Thanks!

PR #8238, opened 2020-07-04
Add new `AsciiConstants` option for `Naming/AsciiIdentifiers`.

### LINKED ISSUE #7458: Request for Naming/AsciiConstants
Rubocop currently has [Naming/AsciiIdentifiers](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/AsciiIdentifiers) which correctly adds offenses to ruby identifiers (`tIDENTIFIER`) containing non-ascii characters. I find that the same rigor should be applied to constants (`tCONSTANT`) as the following are currently not offenses:
```rb
class Foö
end
```
```rb
module Foö
end
```
```rb
FOÖ = "foo"
```

#### Possible Solution:
A new cop: `Naming/AsciiConstants` which adds offenses to code where non-ascii characters are used in constant declarations (See examples above)

#### Alternative: 
A configuration option for `Naming/AsciiIdentifiers` to incorporate constants. Offenses would exist for the examples above. However it does not appear that ruby considers constants as identifiers so this option could be misleading.

fatkodima:
Closes #7458 

Implementation as suggested in https://github.com/rubocop-hq/rubocop/pull/7459#issuecomment-546032027

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 * [#7868](https://github.com/rubocop-hq/rubocop/pull/7868): `Cop::Base` is the new recommended base class for cops. ([@marcandre][])
+* [#7458](https://github.com/rubocop-hq/rubocop/issues/7458): Add new `AsciiConstants` option for `Naming/AsciiIdentifiers`. ([@fatkodima][])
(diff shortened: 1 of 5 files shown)


koic:
Thanks!

Issue #7433, opened 2019-10-15
Fix inherit_from issue with global configuration


bkuhlmann:
## Expected behavior

Would expect Rubocop to honor the project specific directive for `inherit_from` instead of the global configuration of the entire system. For example, here's my setup:

- [$HOME/.config/rubocop/config.yml](https://github.com/bkuhlmann/dotfiles/blob/master/home_files/.config/rubocop/config.yml.tt) - This is part of my dotfiles which provides default behavior for any new Ruby project that hasn't been customized yet.
- [$HOME/OSS/benchmarks/.rubocop.yml](https://github.com/bkuhlmann/benchmarks/blob/master/.rubocop.yml) - This is my Benchmarks project. Notice it *doesn't* include the Rubocop Performance gem but the global configuration does.

## Actual behavior

When running Rubocop within the Benchmarks project, I get the following stack dump:

``` bash
cannot load such file -- rubocop-performance
/Users/bkuhlmann/.rubies/ruby-2.6.5/lib/ruby/site_ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
/Users/bkuhlmann/.rubies/ruby-2.6.5/lib/ruby/site_ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
/Users/bkuhlmann/.gem/ruby/2.6.5/gems/rubocop-0.75.1/lib/rubocop/config_loader_resolver.rb:15:in `block in resolve_requires'
/Users/bkuhlmann/.gem/ruby/2.6.5/gems/rubocop-0.75.1/lib/rubocop/config_loader_resolver.rb:11:in `each'
/Users/bkuhlmann/.gem/ruby/2.6.5/gems/rubocop-0.75.1/lib/rubocop/config_loader_resolver.rb:11:in `resolve_requires'
/Users/bkuhlmann/.gem/ruby/2.6.5/gems/rubocop-0.75.1/lib/rubocop/config_loader.rb:45:in `load_file'
/Users/bkuhlmann/.gem/ruby/2.6.5/gems/rubocop-0.75.1/lib/rubocop/config_loader_resolver.rb:142:in `block in base_configs'
/Users/bkuhlmann/.gem/ruby/2.6.5/gems/rubocop-0.75.1/lib/rubocop/config_loader_resolver.rb:141:in `map'
[...]

PR #8239, opened 2020-07-05
Don't load exclusions from personal files when passing a config with a custom name

deivid-rodriguez:
If passing a configuration with a custom name, like `rubocop --config .my_custom_project_config.yml`, that configuration should act as the project configuration and we want personal configuration files disregarded.

That's actually the only case where `find_files_upwards` can return an empty list, because when given an existing `.rubocop.yml` as the starting point, it will at least return that file, and the fallback to personal config files will not be applied.

Well, there's another case where `find_files_upwards` could return an empty list, which is when giving it a non existent file as the starting point. However, I don't think that ever happens in real life, and in tests it was only happening once, and due to a typo I believe.

So, we can completely remove the explicit fallback to look for exclusions in personal configuration files.

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 * [#8252](https://github.com/rubocop-hq/rubocop/issues/8252): Fix a command line option name from `--safe-autocorrect` to `--safe-auto-correct`, which is compatible with RuboCop 0.86 and lower. ([@koic][])
+* [#8239](https://github.com/rubocop-hq/rubocop/pull/8239): Don't load `.rubocop.yml` from personal folders to check for exclusions if given a custom configuration file. ([@deivid-rodriguez][])
 
 ## 0.87.0 (2020-07-06)
 
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -51,6 +51,10 @@ files:
 * `~/.config/rubocop/config.yml`
 * https://github.com/rubocop-hq/rubocop/blob/master/config/default.yml[RuboCop's default configuration]
 
+All the previous logic does not apply if a specific configuration file is passed
+on the command line through the `--config` flag. In that case, the resolved
+configuration file will be the one passed to the CLI.
+
 == Inheritance
 
 All configuration inherits from https://github.com/rubocop-hq/rubocop/blob/master/config/default.yml[RuboCop's default configuration] (See
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -119,7 +119,6 @@ module RuboCop
 
       def add_excludes_from_files(config, config_file)
         found_files = find_files_upwards(DOTFILE, config_file)
-        found_files = [find_user_dotfile, find_user_xdg_config].compact if found_files.empty?
 
         return if found_files.empty?
         return if PathUtil.relative_path(found_files.last) ==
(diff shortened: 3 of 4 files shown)


deivid-rodriguez:
This is a follow up to #8176. @jonas054 what do you think?

bbatsov:
> If passing a configuration with a custom name, like rubocop --config-file .my_custom_project_config.yml, that configuration should act as the project configuration and we want personal configuration files disregarded.

What do you call personal configuration? Some examples might help understand better the problem you're trying so solve.

deivid-rodriguez:
Hi @bbatsov, sorry for not being clear.

When I talk about personal configuration files, I mean configuration files in either `$HOME` or `$XDG_CONFIG` folders.

Currently these files are not very useful IMO because they get loaded in situations where you don't want them loaded. See for example https://github.com/rubocop-hq/rubocop/issues/7433. In my opinion, they should get only loaded as a fallback when the project has no specific rubocop configuration, or no explicit configuration is being passed to the CLI.

I'm submitting a series of PRs to fix these issues.

* In #8176 I fix the issue when there's an explicit project configuration and neither $HOME, or $XDG_CONFIG are parent folders of the current project.
* In this PR I fix the issue when there's an explicit configuration given on the command line.
* In a followup PR, I'd like to fix the problem in general but I'm trying small changes first that can be easily justified.

A practical example of this specific fix would be when trying to add some initial rubocop configuration to a project. You might want to first try different "template" configurations with different settings, and the `--config-file` option can help there. In that case, I don't think you don't want your personal `~/.rubocop.yml` to influence the configuration you're testing.

deivid-rodriguez:
By the way, if we can find a better terminology, I'm happy to change this and the already merged changelog entries. Maybe "user configuration" is better since it matches more closely the terminology used by the code. I've also seen "global configuration" used.

bbatsov:
> Currently these files are not very useful IMO because they get loaded in situations where you don't want them loaded. See for example #7433. In my opinion, they should get only loaded as a fallback when the project has no specific rubocop configuration, or no explicit configuration is being passed to the CLI.

Agreed. I had almost forgotten about them at this point.

deivid-rodriguez:
> I had almost forgotten about them at this point.

Me too! I tried using them at some point but eventually gave up, because they kept getting in the middle. Then I saw #7433 and realized other people were trying to use them too, so I decided to give it another try.

deivid-rodriguez:
I rebased this PR and moved the changelog entry in response to the new release :smiley:

bbatsov:
I think you should also update the docs to reflect this.

deivid-rodriguez:
@bbatsov I added a note about the `--config` flag which I believe was undocumented. (Also changed my previous comments in this PR since they were mentioning `--config-file` instead, which doesn't exist :sweat_smile:).

Regarding the behaviour of looking up the highest config file to look for exclusions, there's this: https://github.com/rubocop-hq/rubocop/blob/master/docs/modules/ROOT/pages/configuration.adoc#include-and-exclude-are-relative-to-their-directory.

In my opinion, the problem with that docs is that they don't explain _why_ this is done differently from everything else. I'm currently investigating the _why_, since I think there's no good reason for this and all these code could eventually go. But until then I'm not sure I can improve those docs. From the current wording I don't think it's implied that user configuration files should be searched, so I'd say this is just a bug fix that doesn't need documentation.

bbatsov:
Fair enough! Thanks for tackling this!

deivid-rodriguez:
No problem! I'll follow up with more stuff!

PR #8241, opened 2020-07-05
Add new `Style/AccessorGrouping` cop

### LINKED ISSUE #3983: Add a cop that checks grouping of accessor methods
As per [this issue](https://github.com/bbatsov/rubocop/issues/3936#issuecomment-273704948), there should be a cop that checks the grouping of `attr_reader`, `attr_writer` and `attr_accessor`, i.e. `Style/AccessorGrouping`.

All references to accessor methods in the style guide uses a grouped style, e.g.:

```
attr_reader :foo, :bar
```

and I think this should be the default. There is no explicit section on this however, so it should probably be added as part of work on this cop.

**Note:** accessor methods are affected by visibility scopes (`public`, `protected`, `private`) and the cop needs to take this into account.

fatkodima:
Closes #3983

```ruby
# EnforcedStyle: grouped (default)

# bad
class Foo
  attr_reader :bar
  attr_reader :baz
end

# good
class Foo
  attr_reader :bar, :baz
end


# EnforcedStyle: separated

# bad
class Foo
  attr_reader :bar, :baz
end

# good
class Foo
  attr_reader :bar
  attr_reader :baz
end
```


lib/rubocop/cop/style/accessor_grouping.rb
```diff
@@ -0,0 +1,135 @@
...
+      #   # good
+      #   class Foo
+      #     attr_reader :bar
+      #     attr_reader :baz
+      #   end
+      #
+      class AccessorGrouping < Cop
+        include ConfigurableEnforcedStyle
+
+        ACCESSOR_METHODS = %i[attr_reader attr_writer attr_accessor].freeze
```
bbatsov, line 37: + `attr` :-)

lib/rubocop/cop/style/accessor_grouping.rb
```diff
@@ -0,0 +1,135 @@
...
+          style == :grouped
+        end
+
+        def separated_style?
+          style == :separated
+        end
+
+        def sibling_accessors(send_node)
+          send_node.parent
+                   .each_child_node(:send).select do |sibling|
```
bbatsov, line 91: Seems to me you didn't need to break this line to stay within the 100 line-length limit.

lib/rubocop/cop/style/accessor_grouping.rb
```diff
@@ -0,0 +1,135 @@
...
+            sibling.macro? && sibling.method?(send_node.method_name)
+          end
+        end
+
+        def message(send_node)
+          suffix = grouped_style? ? 'a single statement' : 'separate statements'
```
bbatsov, line 97: Technically speaking, they are expressions, not statements. :-)

lib/rubocop/cop/style/accessor_grouping.rb
```diff
@@ -0,0 +1,135 @@
...
+      #   class Foo
+      #     attr_reader :bar
+      #     attr_reader :baz
+      #   end
+      #
+      class AccessorGrouping < Cop
+        include ConfigurableEnforcedStyle
+
+        ACCESSOR_METHODS = %i[attr_reader attr_writer attr_accessor].freeze
+        MSG = 'Put `%<accessor>s` accessors in %<suffix>s.'
```
bbatsov, line 38: I think for the sake of clarity it'd be best to use different messages - e.g. `Group together all ...` and `Use one attribute per ...`. Something along those lines.

fatkodima:
Updated.

ghiculescu:
FYI, this cop is not compatible with sorbet. See [this example](https://sorbet.run/#%23%20typed%3A%20true%0A%0A%23%20given%20this%20code%0Aclass%20Original%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20returns(String)%20%7D%0A%20%20attr_reader%20%3Afoo%0A%20%20%0A%20%20sig%20%7B%20returns(Integer)%20%7D%0A%20%20attr_reader%20%3Abar%0Aend%0A%0AT.assert_type!(Original.new.foo%2C%20String)%0AT.assert_type!(Original.new.bar%2C%20Integer)%0A%0A%23%20rubocop%20-a%20will%20generate%20this%0Aclass%20AutoCorrected%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20returns(String)%20%7D%0A%20%20attr_reader%20%3Afoo%2C%20%3Abar%0A%0A%20%20sig%20%7B%20returns(Integer)%20%7D%0Aend%0A%0AT.assert_type!(AutoCorrected.new.foo%2C%20String)%0AT.assert_type!(AutoCorrected.new.bar%2C%20Integer))

This code:

```ruby
class Original
  extend T::Sig

  sig { returns(String) }
  attr_reader :foo
  
  sig { returns(Integer) }
  attr_reader :bar
end
```

Is auto corrected to this:

```
class AutoCorrected
  extend T::Sig

  sig { returns(String) }
  attr_reader :foo, :bar

  sig { returns(Integer) }
end
```

marcandre:
@ghiculescu I'm not super familiar with Sorbet yet, I imagine the "separated" style is fine. Is it mandatory, actually? I'm too lazy to look it up myself: if two readers are strings, can you write `sig { returns(String) }; attr_reader :foo, :bar`?

ghiculescu:
Yes, that works. The problems are

1) that doesn't work if they return different types
2) we are autocorrecting to invalid syntax

To be fair, I'm not sure if it's rubocop's responsibility to deal with that. i'm just commenting so other sorbet users are aware of this.

marcandre:
I believe it warrants at least a note in the doc: #8828

I believe the cop already considers comments, but maybe our notion of "comment" should be expanded for `Sorbet`.

PR #8244, opened 2020-07-05
Add new `Style/BisectedAttrAccessor` cop

fatkodima:
This cop checks where `attr_reader :foo` and `attr_writer :foo` can be combined into `attr_accessor :foo`.

I was surprised when I haven't found a similar check.

There were a couple of offenses in my project, that stayed after not so careful refactoring. So I would like to have a cop with suggestions.


lib/rubocop/cop/style/bisected_attr_accessor.rb
```diff
@@ -0,0 +1,121 @@
...
+      #     attr_writer :bar
+      #   end
+      #
+      #   # good
+      #   class Foo
+      #     attr_accessor :bar
+      #   end
+      #
+      class BisectedAttrAccessor < Cop
+        MSG = 'Combine both accessors into `attr_accessor`.'
```
bbatsov, line 24: I'd mention the attribute name in the message for clarity.

bbatsov:
Great work! I'm guessing the cop should also cover the somewhat deprecated `attr`, which I think was an alias for `attr_reader`.

fatkodima:
Updated with suggestions.

bbatsov:
Thanks!

PR #8261, opened 2020-07-07
Fix false positives for `Style/AccessorGrouping` when accessors have different access modifiers

### LINKED ISSUE #8253: Style/AccessorGrouping doesn't take access control into account.
I have a class that has an `attr_reader` call at the top (default public access), and then again underneath `private` - essentially:

```ruby
class Foo
  attr_reader :alpha, :beta

  private

  attr_reader :gamma, :delta
end
```

I would expect this to be fine - I only want the accessors underneath private to be available internally to the class instances - but the new `Style/AccessorGrouping` cop says they should _all_ be grouped together. So, I'm guessing it's not picking up on the different access control levels? This feels like a bug to me, unless I'm missing something.

```
$ rubocop -V
0.87.0 (using Parser 2.7.1.4, rubocop-ast 0.1.0, running on ruby 2.6.6 x86_64-darwin19)

$ rubocop --only Style/AccessorGrouping
Inspecting 1 file
C

Offenses:

foo.rb:2:3: C: Style/AccessorGrouping: Group together all attr_reader attributes.
  attr_reader :alpha, :beta
  ^^^^^^^^^^^^^^^^^^^^^^^^^
foo.rb:6:3: C: Style/AccessorGrouping: Group together all attr_reader attributes.
  attr_reader :gamma, :delta
  ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

fatkodima:
Closes #8253


bbatsov:
Great work!

pat:
Thank you for the quick fix :)

PR #8263, opened 2020-07-07
Fix false positives for `Style/BisectedAttrAccessor` when accessors have different access modifiers

### LINKED ISSUE #8259: Style/BisectedAttrAccessor doesn't respect public/private

Often times I only want to make the `attr_reader` public, but still define the `attr_writer` privately for internal use. However, with this cop enabled (especially with autocorrect turned on) this totally changes the access rules and therefore the API of the class.

I have no idea how easy this would be to implement, but it would be great if the cop could only "fire" (is that the correct term?) if the reader and writer have the same access level, i.e. both public, private, or protected.

fatkodima:
Closes #8259


bbatsov:
Thanks!

ujh:
Thanks for the quick fix! 💚

PR #8266, opened 2020-07-07
Add new `Style/CaseLikeIf` cop

### LINKED ISSUE #7736: Suggest case/when if an if/elsif/else can be easily translated

I only learned about the power of case statements quite recently, so there are lots of opportunities in my code that could be written more nicely with a cop that recognizes these patterns. I am sure there are others with the same situation.

If there is an if/elsif/else construct with at least one elsif and all conditions can be translated appropriately, suggest using a case/when construct instead.

## Additional context

E.g. something like this can be translated:

```
if a == 1
  # do something
elsif a == 2 || a == 3
   # do something
end
```

Or something like this:

```
if a.is_a?(Integer)
  handle_integer(a)
elsif a.is_a?(Float)
  handle_float(a)
end
```

Note that I might even implement this myself eventually, I'll just add the issue so it's tracked.

fatkodima:
This cop tries to convert `if-elsif`s to `case-when`s.

Algorithm, in brief:

```ruby
case target
when conditions1
  # ...statements...
when conditions2
  # ...statements...
else
  # ...statements...
end
```

1. Find possible target from `if` condition
2. if found, then for each branch collect conditions and check if they are all depend on target and can be converted to `case`
3. if all branches are convertible to `case`s, convert them

Closes #7736


bbatsov:
Great work! I wonder if we can do better with the name, though. I'm thinking of `CaseLikeIf` (case-like if) or `SingleValueDispatchIf`, but maybe someone from @rubocop-hq/rubocop-core will have better ideas.

koic:
> `CaseLikeIf` (case-like if)

It sounds very comfortable to me :-)

config/default.yml
```diff
@@ -3011,6 +3011,12 @@ Style/IdenticalConditionalBranches:
   Enabled: true
   VersionAdded: '0.36'
 
+Style/IfElsifToCase:
+  Description: 'This cop identifies places where `if-elsif` constructions can be replaced with `case-when`.'
+  StyleGuide: 'https://rubystyle.guide/#case-vs-if-else'
```
koic, line 3016:
```suggestion
  StyleGuide: '#case-vs-if-else'
```

bbatsov:

fatkodima:
Rebased.

PR #8271, opened 2020-07-08
Add new `Style/HashArrayLastItem` cop

### LINKED ISSUE #4286: rule suggestion hash inside array
When a ruby hash is the last item in an array, we can use the same shorthand syntax as for method calls. It would be nice to have a linter to enforce one style or another:

```ruby
# braces
my_array = ['foo', 'bar', { baz: 'butz' }]

# no_braces
my_array = ['foo', 'bar', baz: 'butz']
```

fatkodima:
Closes #4286


bbatsov:
I think we should also mention this in the style guide with a suggestion to use braces for the sake of clarity.

config/default.yml
```diff
@@ -2961,6 +2961,17 @@ Style/GuardClause:
   # needs to have to trigger this cop
   MinBodyLength: 1
 
+Style/HashArrayLastItem:
```
bbatsov, line 2964: Perhaps `HashAsLastArrayItem`? Another hard to name cop.

koic:
Can you open a PR as the rule to the style guide?

fatkodima:
Changed name to `HashAsLastArrayItem`, opened https://github.com/rubocop-hq/ruby-style-guide/pull/826 and added link to that.

marcandre:
Looks good, but will you start using `Cop::Base` instead of `Cop::Cop`? In this particular case, I think that instead of having an `if such_and_such_a_style` three times, it would be more readable if there was a single such `if`

fatkodima:
> I think that instead of having an if such_and_such_a_style three times, it would be more readable if there was a single such if

Didn't get it. Can you point to code?

marcandre:
Here's what I had in mind. What do you think of this style?

```ruby
class HashAsLastArrayItem < Base
  include ConfigurableEnforcedStyle

  def on_hash(node)
    return unless node.parent&.array_type?

    if braces_style?
      check_braces(node)
    else
      check_no_brances(node)
    end
  end

  private

  def check_braces(node)
    return unless node.braces

    add_offense(node, message: 'Wrap hash in `{` and `}`.') do |corrector|
      corrector.wrap(node, '{', '}')
    end
  end

  def check_no_braces(node)
    return if node.braces

    add_offense(node, message: 'Omit the braces around the hash.') do |corrector|
      corrector.remove(node.loc.begin)
      corrector.remove(node.loc.end)
    end
  end

  def braces_style?
    style == :braces
  end
end
```

fatkodima:
Ah, got it. Looks nicer 👍 
Updated with your suggestions.

koic:
Thanks!

PR #8245, opened 2020-07-06
Detect top-level constant usage like `::Const`

biinari:
[Fix #8215]

Detect top-level constant usage (in offenses or allowed forms) like `::Const` in the following cops:

* Gemspec/RubyVersionGlobalsUsage - `::RUBY_VERSION`
* Lint/NestedMethodDefinition - `::Class.new`, `::Module.new`, `::Struct.new`
* Lint/NonDeterministicRequireOrder - `::Dir`
* Lint/RandOne - `::Kernel.rand(1)`
* Lint/RedundantSplatExpansion - `::Array.new`
* Lint/UnusedMethodArgument - `::NotImplementedError`
* Lint/UselessAccessModifier - `::Class.new`, `::Module.new`, `::Struct.new`
* Metrics/ClassLength - `::Class.new`
* Metrics/ModuleLength - `::Module.new`
* Style/DateTime - Allow using `::Date` constants for historic date
* Style/Dir - `::File`
* Style/EmptyLiteral - `::Array.new`, `::Hash.new`, `::String.new`
* Style/ExpandPathArguments - `::File.expand_path`, `::Pathname.new`
* Style/MutableConstant - `::ENV`, `::Struct`
* Style/Proc - `::Proc.new`
* Style/RandomWithOffset - `::Kernel.rand`, `::Random.rand`
* Style/RedundantException - `::RuntimeError`
* Style/RedundantFreeze - `::ENV`
* Style/RescueStandardError - `::StandardError`
* Style/SignalException - `::Kernel.raise`, `::Kernel.fail`
* Style/StderrPuts - `::STDERR`
* Style/StructInheritance - `::Struct`
* Style/SymbolProc - `::Proc.new`
* Style/ZeroLengthPredicate - `::File.stat`, `::Tempfile.new`, `::StringIO.new`

Also in `Mixin/EnforceSuperclass` used by [rubocop-rails ApplicationRecord](https://github.com/rubocop-hq/rubocop-rails/blob/master/lib/rubocop/cop/rails/application_record.rb) and maybe other gems:

* Detect `::Class.new` and `Class.new` with or without a block.
* Added a spec using a stubbed cop class to include the mixin.

If any of these changes are unwanted or desired to be split out, I'm happy to do so. That is why I have not yet squashed my commits.

I have not added any examples to documentation as they would not aid the reader's understanding of what the cops are looking for.

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
[...]

PR #8279, opened 2020-07-08
Detect require block passed as argument in Lint/NonDeterministicRequireOrder

biinari:
Detect `&method(:require)` passed as a block argument. For example:

```ruby 
Dir[Rails.root.join('spec/support/**/*.rb')].each(&method(:require))

Dir.glob(
  Rails.root.join('spec', 'support', '**', '*.rb'),
  &method(:require)
)
```

These are examples from our code base where the non-deterministic require order was not picked up.

### Notes

I originally tried adding `on_send` handler for this but that had problems:
* Reaching the correct level of node for the different forms (`Dir.glob(...)` vs `Dir[...].each`).
* Detecting whether the block is passed as an argument.
* The cop would then have to look at much more code.

Some pending specs for this leaked in via #8245 without description, sorry about that.

-----------------

* [x] Added tests.


bbatsov:
Thanks!

PR #8280, opened 2020-07-09
Add new `Style/HashLikeCase` cop

### LINKED ISSUE #8247: New cop to suggest replacing case/when with hash lookup
When `case/when` represents a simple 1:1 mapping, suggest to replace it with a hash lookup.

```ruby
# bad
case country
when "europe"
  "http://eu.example.com"
when "america"
  "http://us.example.com"
end

# good
SITES = {
  "europe"  => "http://eu.example.com",
  "america" => "http://us.example.com"
}
SITES[country]
``` 

I'm propose to name it `Style/CaseHashMapping` (is there a better name?). Maybe it would be wise to add a `Min` config option, specifying for how many branches should be to consider an offense.

@bbatsov Interested in your thoughts on this.

fatkodima:
Closes #8247


lib/rubocop/cop/style/format_string_token.rb
```diff
@@ -75,11 +75,11 @@ def message(detected_style)
 
         # rubocop:disable Style/FormatStringToken
         def message_text(style)
-          case style
-          when :annotated then 'annotated tokens (like `%<foo>s`)'
-          when :template then 'template tokens (like `%{foo}`)'
-          when :unannotated then 'unannotated tokens (like `%s`)'
-          end
+          {
```
bbatsov, line 78: Any particular reason you didn't move this to a constant?

bbatsov:
The code looks good overall. You might want to add some tests with an `else` branch.

lib/rubocop/cop/style/format_string_token.rb
fatkodima: Those are just pieces of the message, not particularly useful for the whole class, so I decided it is cleaner to left it in a method instead of extracting to the top-level constant.
Can extract into a constant, if asked.

fatkodima:
Added a test for `else` branch.

lib/rubocop/cop/style/format_string_token.rb
bbatsov: Got it.

bbatsov:
Thanks!

config/default.yml
```diff
@@ -2980,6 +2980,16 @@ Style/HashEachMethods:
   VersionAdded: '0.80'
   Safe: false
 
+Style/HashLikeCase:
+  Description: >-
+                  Checks for places where `case-when` represents a simple 1:1
+                  mapping and can be replaced with a hash lookup.
+  Enabled: 'pending'
```
koic, line 2987: Will this cop be enabled by default at RuboCop 1.0? I'm not sure about creating a constant of hash instance is a better style than `case` syntax if there is no performance advantage. I think it may be considered disabled by default.

bbatsov: I think it most cases using the hash results in code that's simpler, as you eliminate branching completely. Obviously a simple `case` is not a big problem, but I feel that in most cases using a hash results in cleaner code.

lib/rubocop/cop/style/format_string_token.rb
marcandre: I'll repeat that I feel it's bad style to keep on recreating the same constant hash over and over. Wast of CPU cycles and when reading you have to check the whole hash wondering "Hum, is there anything dynamic in here"?

Drenmi:
Technically a `case` with an `else` branch can be replaced with a hash with a default value, but it's not nearly as neat as a hash literal, so I think this is a good cut-off. 🙂

PR #8292, opened 2020-07-09
Fix `Style/AccessorGrouping` to not register offense for accessor with comment

### LINKED ISSUE #8289: Style/AccessorGrouping cop autocorrect mess-up code with comments

1. Create `test.rb` file with content:
```ruby
class Foo
  # [String] Some bar value return
  attr_reader :bar

  # @return [String] value of foo_bar
  attr_reader :foo_bar
end

```

2. run `rubocop -a` with enabled `Style/AccessorGrouping` cop

## Expected behavior

Do not mess up my comments and `yard` documentation

## Actual behavior

Result file is:
```ruby
class Foo
  # [String] Some bar value return
  attr_reader :bar, :foo_bar

  # @return [String] value of foo_bar
end

```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.87.1 (using Parser 2.7.1.4, rubocop-ast 0.1.0, running on ruby 2.7.1 x86_64-linux)
```

tejasbubane:
Closes #8289 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/accessor_grouping.rb
+++ b/lib/rubocop/cop/style/accessor_grouping.rb
@@ -58,6 +58,10 @@ module RuboCop
 
         private
 
+        def previous_line_comment?(node)
+          comment_line?(processed_source[node.first_line - 2])
+        end
+
         def class_send_elements(class_node)
           class_def = class_node.body
 
@@ -75,6 +79,8 @@ module RuboCop
         end
 
         def check(send_node)
+          return if previous_line_comment?(send_node)
+
           if grouped_style? && sibling_accessors(send_node).size > 1
             add_offense(send_node)
           elsif separated_style? && send_node.arguments.size > 1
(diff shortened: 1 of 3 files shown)


bbatsov:
@fatkodima Please, take a look as well.

lib/rubocop/cop/style/accessor_grouping.rb
```diff
@@ -41,6 +41,8 @@ class AccessorGrouping < Cop
         ACCESSOR_METHODS = %i[attr_reader attr_writer attr_accessor attr].freeze
 
         def on_class(node)
+          return if grouped_style? && !accessors_without_comments(node)
```
fatkodima, line 44: This check seems not needed.

tejasbubane: Right. I was thinking it might be needed when there are multiple separate accessors and only one without comment. But that case also seems to be handled by `check`. Removed this.

koic:
Thanks!

PR #8293, opened 2020-07-09
Add new `Lint/DuplicateElsifCondition` cop

fatkodima:
This cop checks for duplicated conditions in `elsif`s. This is almost definitely a programmer mistake due to copy-paste, refactoring, etc.

```ruby
# bad
if x == 1
  do_something
elsif x == 1
  do_something_else
end

# good
if x == 1
  do_something
elsif x == 2
  do_something_else
end
```

This cop is in addition to existing `Lint/DuplicateCaseCondition`.


bbatsov:
Seems to me this should be a Lint cop.

bbatsov:
Seems it's a Style cop only in the PR description and the commit message. :-)

fatkodima:
Haha, yes, it is a bad copy-paste 😄

bbatsov:
Thanks!

PR #8295, opened 2020-07-10
Add new `Style/ArrayCoercion` cop

fatkodima:
https://rubystyle.guide/#array-coercion


bbatsov:
Great work! It'd be good to make this cop configurable down the road and leave it to the users to choose whatever style of coercion they prefer, though, as some people might prefer `[*var]` despite the advantages of `Array()`.

PR #8301, opened 2020-07-10
[Fix #8299] Fix an incorrect auto-correct for `Style/RedundantCondition`

### LINKED ISSUE #8299: Style/RedundantCondition can generate syntax errors when safe-auto-correcting
In certain situations, Style/RedundantCondition can generate a syntax errors.

--------

## Expected behavior

I'd expect rubocop to either not auto correct (since it was supposed to be a safe auto-correction), or to not generate invalid code.

## Actual behavior

rubocop generated code with invalid syntax.

 I had a code like this:

```
def func
  a = 1

  if a
    return a
  else
    raise "Something went wrong"
  end
end
```

Style/RedundantReturn corrected it to:
```
def func
  a = 1

  if a
    a
  else
    raise "Something went wrong"
  end
end
```

Which then Style/RedundantCondition auto corrected to:
```
def func
  a = 1

  a || raise "Something went wrong"
end
```

## RuboCop version

```
❯ rubocop -V
0.85.1 (using Parser 2.7.1.4, rubocop-ast 0.1.0, running on ruby 2.5.8 x86_64-linux)
```

koic:
Fixes #8299

This PR fixes an incorrect auto-correct for `Style/RedundantCondition` when using `raise` without argument parentheses in `else`.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#8324](https://github.com/rubocop-hq/rubocop/issues/8324): Fix crash for `Layout/SpaceAroundMethodCallOperator` when using `Proc#call` shorthand syntax. ([@fatkodima][])
 * [#8323](https://github.com/rubocop-hq/rubocop/issues/8323): Fix a false positive for `Style/HashAsLastArrayItem` when hash is not a last array item. ([@fatkodima][])
+* [#8299](https://github.com/rubocop-hq/rubocop/issues/8299): Fix an incorrect auto-correct for `Style/RedundantCondition` when using `raise`, `rescue`, or `and` without argument parentheses in `else`. ([@koic][])
 
 ## 0.88.0 (2020-07-13)
 
--- a/lib/rubocop/cop/style/redundant_condition.rb
+++ b/lib/rubocop/cop/style/redundant_condition.rb
@@ -90,10 +90,10 @@ module RuboCop
         end
 
         def else_source(else_branch)
-          if else_branch.basic_conditional? &&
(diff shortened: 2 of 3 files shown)


marcandre:
This is a good fix, but issues remain for all operators with lower precedence than `||`...

```
if a
  a
else
  b rescue c
# or
  b ? c : d
# or
  b .. c
# or
  b and c
# or
  b if c # or unless / white / until
end
```

koic:
Thank you for your review.

> This is a good fix, but issues remain for all operators with lower precedence than ||...

Some cases were already covered, but `rescue` and `and` were not. I updated this PR to cover them.

Also, the failure of Ruby 2.8.0-dev is probably unrelated to this change. I will investigate (or open an issue) separately from the PR.

marcandre:
> Also, the failure of Ruby 2.8.0-dev is probably unrelated to this change. I will investigate (or open an issue) separately from the PR.

These failures seem to occur reliably. It's quite confusing when we want to merge other things if the build is broken. Could we please fix this (or else temporarily revert) please

koic:
This build error is due to a breaking change in the edge Ruby. RuboCop's master branch is broken independent of this PR.
(I had confirmed it locally so I should have commented it.)

I have asked the following address.
https://bugs.ruby-lang.org/issues/17017#note-6

marcandre:
Ah, right, sorry.

I even told jeremyevans to watch out for Infinity. I expect it will be fixed quickly 🤞

koic:
Yeah, I got the answer and opened the PR #8343.

PR #8314, opened 2020-07-12
Don't load configuration files outside of the current project unless there's no project specific configuration

deivid-rodriguez:
In previous PRs, #8176 and #8239, I fixed particular cases of this problem, but it was still not fixed in general. So, if I have my personal configuration at `~/.config/rubocop/config.yml` and my project is at `~/Code/my_project`, then my personal configuration won't be loaded thanks to those PRs. However, if I have my personal configuration at `~/.rubocop.yml`, then `rubocop` will still load exclusions from there because `$HOME` is an ancestor of the current project folder, and that's how `find_last_file_upwards` currently works.

To fix this problem in general, I made sure `find_last_file_upwards` (and any configuration search in general) doesn't go past the current project's root. To do that, I needed to define what the root of a project is. While there were other options, like presence of a `.git` folder, for example, I decided go with "the highest folder up in the directory hierarchy containing a Gemfile or gems.rb" file.

This is the same definition `bundler` uses and I believe it makes sense, specially because the original problems I want to solve here manifest were personal configuration files try to load rubocop plugins outside of the current bundle, and thus making rubocop crash. See https://github.com/rubocop-hq/rubocop/issues/7433. By using Gemfile presence as a criteria, we prevent exactly this kind of problem.

As a note, the reason the test suite never catched this issue is that it intentionally defines a not very common folder structure where the current project folder _is not_ inside `$HOME`. I think that _maybe_ the structure was defined this way in order to workaround this problem.

Finally, although this should work for most ruby projects out there, projects without a `Gemfile` don't fit into this criteria. I believe we can leave it as it is tough, and wait for issues from such projects before adding more fallbacks to this "root of a project" definition.

I'd be very happy to hear your thoughts, thanks for reading!

* [x] Added tests.

### DIFF
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -24,7 +24,7 @@ module RuboCop
 
       attr_accessor :debug, :ignore_parent_exclusion,
                     :disable_pending_cops, :enable_pending_cops
-      attr_writer :default_configuration
+      attr_writer :default_configuration, :project_root
 
       alias debug? debug
       alias ignore_parent_exclusion? ignore_parent_exclusion
@@ -164,7 +164,7 @@ module RuboCop
       private
 
       def find_project_dotfile(target_dir)
-        find_file_upwards(DOTFILE, target_dir)
+        find_file_upwards(DOTFILE, target_dir, project_root)
       end
 
       def find_project_root
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -118,7 +118,7 @@ module RuboCop
       end
 
       def add_excludes_from_files(config, config_file)
-        exclusion_file = find_last_file_upwards(DOTFILE, config_file)
+        exclusion_file = find_last_file_upwards(DOTFILE, config_file, project_root)
 
         return unless exclusion_file
         return if PathUtil.relative_path(exclusion_file) == PathUtil.relative_path(config_file)
@@ -132,6 +132,12 @@ module RuboCop
                                      print 'Default ' if debug?
                                      load_file(DEFAULT_FILE)
                                    end
+      end
+
+      # Returns the path rubocop inferred as the root of the project. No file
+      # searches will go past this directory.
+      def project_root
+        @project_root ||= find_project_root
       end
 
       def warn_on_pending_cops(pending_cops)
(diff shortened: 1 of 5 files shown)


lib/rubocop/config_loader.rb
```diff
@@ -24,7 +24,7 @@ class << self
 
       attr_accessor :debug, :ignore_parent_exclusion,
                     :disable_pending_cops, :enable_pending_cops
-      attr_writer :default_configuration
+      attr_writer :default_configuration, :project_root
```
deivid-rodriguez, line 27: This is only for test code to be able to reset state, the project_root doesn't really change at runtime. I can alternatively revert this change and use `instance_variable_set` inside the test code to reset it.

lib/rubocop/config_loader.rb
```diff
@@ -134,6 +134,12 @@ def default_configuration
                                    end
       end
 
+      # Returns the path rubocop inferred as the root of the project. No file
+      # searches will go past this directory.
+      def project_root
+        @project_root ||= find_project_root
+      end
+
```
deivid-rodriguez, line 142: I intentionally left this method in the public section and documented it since I believe it could be useful for extensions. At least it's useful for [a cop I'm helping with in the `rubocop-packaging` extension](https://github.com/utkarsh2102/rubocop-packaging/pull/4).

bbatsov:
Nicely done!

> Finally, although this should work for most ruby projects out there, projects without a Gemfile don't fit into this criteria. I believe we can leave it as it is tough, and wait for issues from such projects before adding more fallbacks to this "root of a project" definition.

Probably `.gemspec` should be used as a root marker as well, but that's not a big deal.

Might be a good idea to add a few lines in the docs about not adding exclusions to the global config file, or potentially adding some validation and warnings when this is done.

deivid-rodriguez:
Thanks @bbatsov! :purple_heart:

> Probably .gemspec should be used as a root marker as well, but that's not a big deal.

I can prepare a followup PR to add this :+1:.

> Might be a good idea to add a few lines in the docs about not adding exclusions to the global config file, or potentially adding some validation and warnings when this is done.

I think there might be some situations where exclusions in the global configuration file make sense. Like if you have a lot of projects with a `tempdir` folder (or something else not included in `rubocop` default exclusions), some of them configured to use `rubocop` and some where `rubocop` is not used. For those where rubocop is not used, you might want to add this exclusion to your global (personal) configuration, so that they are picked up there.

PR #8322, opened 2020-07-13
Support autocorrect for `Style/CaseEquality` cop

fatkodima:


tejasbubane:
@fatkodima Since rubocop 0.88 is released, you should update `VersionChanged` & move changelog under `master`.

fatkodima:
Updated.

lib/rubocop/cop/style/case_equality.rb
```diff
@@ -29,13 +29,20 @@ module Style
       #   (1..100).include?(7)
       #   some_string =~ /something/
       #
-      class CaseEquality < Cop
+      class CaseEquality < Base
```
tejasbubane, line 32: I think this should not be changed:

```suggestion
      class CaseEquality < Cop
```

koic: The auto-correction has been added by the new API, I think it is okay to extend `Base`. Rather, changing the inheritance to `Cop` will fail tests.

tejasbubane: @koic I might be a little confused between Base and Cop. I see most cops inherit from `Cop`, do not `extend AutoCorrector` and have separate method `autocorrect`. Which of these is the preferred/new approach?

marcandre: Inheriting from  `Base` is preferred/new approach. Here's the doc: https://docs.rubocop.org/rubocop/v1_upgrade_notes.html

lib/rubocop/cop/style/case_equality.rb
```diff
@@ -47,6 +54,18 @@ def const?(node)
...
+
+        def replacement(lhs, rhs)
+          case lhs.type
+          when :regexp
+            "#{rhs.source} =~ #{lhs.source}"
+          when :begin
+            child = lhs.children.first
+            "#{lhs.source}.include?(#{rhs.source})" if child&.range_type?
+          when :const
+            "#{rhs.source}.is_a?(#{lhs.source})"
```
knu, line 66: Not all constants are class/module instances.  Regexps are often bound to a constant, to be used with `===`.

I think a better bet is to replace the constant only if it is CamelCased.

lib/rubocop/cop/style/case_equality.rb
```diff
@@ -47,6 +54,18 @@ def const?(node)
             true
           end
         end
+
+        def replacement(lhs, rhs)
+          case lhs.type
+          when :regexp
+            "#{rhs.source} =~ #{lhs.source}"
```
knu, line 61: `/regexp/ === object` and `object =~ /regexp/` are not interchangeable.  The former returns true or false whereas the latter returns an index or nil.

lib/rubocop/cop/style/case_equality.rb
koic: Thank you for your feedback. I've opened the PR #8526.

lib/rubocop/cop/style/case_equality.rb
koic: Exactly! I've opened the PR #8527.

PR #8331, opened 2020-07-13
Support autocorrect for `Lint/EmptyWhen` cop

fatkodima:


spec/rubocop/cop/lint/empty_when_spec.rb
```diff
@@ -47,7 +63,13 @@
         else 3
         end
       RUBY
-      expect_no_corrections
+
+      expect_correction(<<~RUBY)
+        case foo
```
marcandre, line 68: This is not equivalent at all...
I'm not sure that even when equivalent it's such a good idea to do so.

fatkodima: 😅 Correct, my bad. Updated code and tests.

When there is an empty `when` clause it is probably that developer made some refactoring and forgot to remove that branch. So it is a good idea to suggest and when it is equivalent, it is not a (at least 'very') bad idea to autocorrect for him, like other `Empty...` cops do.

marcandre:
I'm hesitant about the fact that this auto-corrects by removing comments. Are there any other auto-correction that removes comments? What is the comment is "# TODO: Super important to implement this, I'll do it next week but don't want to forget this"?
It's hard for me to say because I think it's a mistake to run this cop with `AllowComment` set to `false`, so I can't quite imagine what those users think.

fatkodima:
@marcandre After second thought, yes, you are correct. User can create a `case` without `else` and with an empty `when`. Then this cop will autocorrect by removing that `case`. In future developer can add `else` and after that that deleted before `when` will be handled by `else`, which might not be correct. I will close this.

Not directly related to this pr, but I haven't seen a cop for `if/elsif` empty bodies, like `EmptyWhen`. Wdyt, does it make sense to implement one?

marcandre:
> Not directly related to this pr, but I haven't seen a cop for `if/elsif` empty bodies, like `EmptyWhen`. Wdyt, does it make sense to implement one?

Indeed, there's something to be said about an empty body for an `if/elsif` without a comment.

PR #8337, opened 2020-07-14
Support `defined?`-based memoization for `Naming/MemoizedInstanceVariableName` cop

### LINKED ISSUE #5717: Trigger `MemoizedInstanceVariableName` cop for `return @foo if defined? @foo`
Thanks for #5445!

`@foo ||= value` is good, but not always: https://karolgalanciak.com/blog/2017/05/28/ruby-memoization-%7C%7C-equals-vs-defined-syntax/

So, I'm suggesting to add pattern `return @foo if defined? @foo` (and `defined?(@foo)`) for this cop.

/cc @satyap 

--------

## Expected behavior

`MemoizedInstanceVariableName` triggers for:

```ruby
def foo
  return @bar if defined? @bar
  @bar = false
end
```

## Actual behavior

`MemoizedInstanceVariableName` doesn't trigger for `defined?` with mismatched variable name.

Code is above.

## RuboCop version

```
$ rubocop -V
0.54.0 (using Parser 2.5.0.4, running on ruby 2.5.0 x86_64-linux)
```

fatkodima:
Closes #5717 

As for me, code is a bit hairy, but splitting it further probably won't make it better.

[Rendered without space changes](https://github.com/rubocop-hq/rubocop/commit/4635fcab3dbf8a0ee0f9c1f1eab9f14c83bdebcb?w=1)

### DIFF
--- a/changelog/change_memoized_instance_variable_cop.md
+++ b/changelog/change_memoized_instance_variable_cop.md
@@ -1,0 +1,1 @@
+* [#5717](https://github.com/rubocop-hq/rubocop/issues/5717): Support `defined?`-based memoization for `Naming/MemoizedInstanceVariableName` cop. ([@fatkodima][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2313,7 +2313,7 @@ Naming/MemoizedInstanceVariableName:
     Memoized method name should match memo instance variable name.
   Enabled: true
   VersionAdded: '0.53'
-  VersionChanged: '0.58'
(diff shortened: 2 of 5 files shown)


bbatsov:

fatkodima:
Rebased.

[Rendered without space changes](https://github.com/rubocop-hq/rubocop/pull/8337/files?w=1) (half the diff size).

bbatsov:
Thanks!

PR #8338, opened 2020-07-14
Config#for_department: be more accurate

marcandre:
I noticed that `for_department` is quite error prone. This fixes it:

* Don't rely on the department name being an unknown cop. E.g. `for_department('Foo')` would return the config for the cop "Foo/Foo" if it exists, instead of the department's.

* Don't provide 'Enabled' at the department level unless it is specified in the config.

* Accept symbols (like that returned by `Badge#department`) too


bbatsov:
Might be a good idea to add a couple of unit tests specifying the expected behaviour, otherwise i can imagine regressions down the road.

marcandre:
Tests added, merged, thanks for the review

PR #8339, opened 2020-07-14
Add Config#for_badge

marcandre:
This optimization saves about 5MB of unneeded intermediate hashes...


bbatsov:
The changes look good apart from the failing build.

PR #8340, opened 2020-07-14
[Fix #8335] Improve class detection in `Style/RedundantRegexpEscape`

### LINKED ISSUE #8335: Style/RedundantRegexpEscape can generate incorrect fixes for hyphens in character classes
`RedundantRegexpEscape` is meant to detect and remove unneeded escapes for regular expressions. In certain cases an escaped hyphen in the middle of a character class can be automatically unescaped such that its meaning changes from a literal hyphen to creating a range match.

--------

## Expected behavior

A hyphen in the middle of a character class should only be unescaped if its resulting meaning stays the same.

## Actual behavior

In at least some scenarios the hyphen can be converted to a range meaning. See below for a concrete example.

Example regexp: `%r{[:\.\-/]*[^:\.\-/]+[:\.\-/]*}`
This regexp could be safely reduced to `%r{[:./-]*[^:./-]+[:./-]*}` (moving the hyphen to the end of each class to avoid creating a range). Or one could leave the hyphen in place but leave it escaped (only un-escaping the period).
Test string `"america:pnw.oregon-portland/kenton"`
Using Ruby `scan`, results are: `["america:", "pnw.", "oregon-", "portland/", "kenton"]`

Under this version of Rubocop the regexp above is incorrectly simplified to `%r{[:\.-/]*[^:.\-/]+[:\.-/]*}`
Ruby `scan` with this change instead yields: `["america:", "pnw.", "oregon", "portland/", "kenton"]`

## RuboCop version

```
$ bundle exec rubocop -V
0.88.0 (using Parser 2.7.1.4, rubocop-ast 0.1.0, running on ruby 2.6.6 x86_64-darwin19)
```

owst:
The previous code wouldn't detect character classes that had ":" as
their first element (due to a poor attempt at avoiding issues due to
POSIX bracket expressions). The handling of nested character classes was
also improved to avoid false-negatives.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [#8324](https://github.com/rubocop-hq/rubocop/issues/8324): Fix crash for `Layout/SpaceAroundMethodCallOperator` when using `Proc#call` shorthand syntax. ([@fatkodima][])
 * [#8323](https://github.com/rubocop-hq/rubocop/issues/8323): Fix a false positive for `Style/HashAsLastArrayItem` when hash is not a last array item. ([@fatkodima][])
 * [#8299](https://github.com/rubocop-hq/rubocop/issues/8299): Fix an incorrect auto-correct for `Style/RedundantCondition` when using `raise`, `rescue`, or `and` without argument parentheses in `else`. ([@koic][])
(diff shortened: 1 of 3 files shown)


marcandre:
Thanks for the quick fix! 👍

(Merging, I'm assuming failures are due to broken build...)

PR #8341, opened 2020-07-14
Add new `Lint/EmptyConditionalBody` cop

fatkodima:
This cop checks for empty bodies in `if`/`elsif`/`unless` branches.
This is in addition to existent `Lint/EmptyWhen` cop.


lib/rubocop/cop/lint/uri_regexp.rb
```diff
@@ -43,10 +43,8 @@ def on_send(node)
 
         def autocorrect(node)
           lambda do |corrector|
-            if (captured_values = uri_regexp_with_argument?(node))
-            else
-              captured_values = uri_regexp_without_argument?(node)
-            end
+            captured_values =
```
rrosenblum, line 46: I'm always happy when a new cop catches weird code within our own source code. Really helps to confirm that the cop is useful.

bbatsov:

PR #8351, opened 2020-07-16
[Fix #8347] Fix an incorrect auto-correct for `Style/HashSyntax`

### LINKED ISSUE #8347: HashSyntax plus HashAlignment breaks the page
When mixing `Style/HashSyntax` with a setting of `hash_rockets` and the default `Layout/HashAlignment` cop, auto-correction can break the page and render a syntax error.

## Expected behavior

Original code:
```ruby
some_method(a: 'abc', b: 'abc',
        c: 'abc', d: 'abc'
        )
```

Expected output:
```ruby
some_method(:a => 'abc', :b => 'abc',
            :c => 'abc', :d => 'abc'
        )
  end
```

## Actual behavior

Actual output:

```ruby
some_method(:a => 'abc', :b => 'abc',
        :    c => 'abc', :d => 'abc'
        )
```

Running `rubocop -a` on the above code will replicate the problem.

## RuboCop version

```
0.88.0 (using Parser 2.7.1.4, rubocop-ast 0.1.0, running on ruby 2.7.0 x86_64-darwin16)
```

koic:
Fixes #8347.

This PR fixes an incorrect auto-correct for `EnforcedStyle: hash_rockets` of `Style/HashSyntax` with `Layout/HashAlignment`.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cli/cli_autocorrect_spec.rb
+++ b/spec/rubocop/cli/cli_autocorrect_spec.rb
@@ -174,6 +174,28 @@ RSpec.describe RuboCop::CLI, :isolated_environment do
     RUBY
   end
 
+  it 'corrects `EnforcedStyle: hash_rockets` of `Style/HashSyntax` with `Layout/HashAlignment`' do
+    create_file('.rubocop.yml', <<~YAML)
+      Style/HashSyntax:
+        EnforcedStyle: hash_rockets
+    YAML
+    source = <<~RUBY
+      some_method(a: 'abc', b: 'abc',
+              c: 'abc', d: 'abc'
+              )
+    RUBY
+    create_file('example.rb', source)
+    expect(cli.run([
+                     '--auto-correct-all',
+                     '--only', 'Style/HashSyntax,Style/HashAlignment'
+                   ])).to eq(0)
+    expect(IO.read('example.rb')).to eq(<<~RUBY)
+      some_method(:a => 'abc', :b => 'abc',
+                  :c => 'abc', :d => 'abc'
+              )
+    RUBY
+  end
+
   describe 'trailing comma cops' do
     let(:source) do
       <<~RUBY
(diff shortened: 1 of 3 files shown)


spec/rubocop/cli/cli_autocorrect_spec.rb
```diff
@@ -174,6 +174,28 @@ def batch
...
+                     '--auto-correct-all',
+                     '--only', 'Style/HashSyntax,Style/HashAlignment'
+                   ])).to eq(0)
+    expect(IO.read('example.rb')).to eq(<<~RUBY)
+      some_method(:a => 'abc', :b => 'abc',
+                  :c => 'abc', :d => 'abc'
+              )
+    RUBY
+  end
+
```
dorner, line 198: Amazing - I just spent about four hours trying to get this to work via way overcomplicating the `expect_correction` method (basically by adding a `let(:run_first)` variable). This is a million times cleaner. :) Going to miss having a PR to Rubocop but I guess I'll have other opportunities!

marcandre: I'm curious: what was the difficulty with `expect_correction`? This test with the CLI is much slower...

dorner: The issue is that you can only run one cop at a time. There's no easy way to pass it multiple cops to run one after another. This bug only happens when one cop makes a change and the next one makes a second change - the two insertions end up in an invalid state.

marcandre: I was under the impression that `let(:other_cops) { { 'Other/Cop' => ... } }` was the way to go, but indeed looks like it's not the case. I opened #8352 as it should be easily doable to specify additional cops.

bbatsov:
Thanks!

PR #8353, opened 2020-07-16
[Fixes #8352] Allow multiple cops in shared_context.

### LINKED ISSUE #8352: shared_context should allow for multiple cops
There's no easy way to test interactions between multiple cops' autocorrections currently.

dorner:
Fixes #8352.

This allows additional cops to be run before the current one, allowing us to test interactions between cops more effectively.

cc: @marcandre 

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/commissioner.rb
+++ b/lib/rubocop/cop/commissioner.rb
@@ -25,6 +25,11 @@ module RuboCop
           @correctors ||= cop_reports.map(&:corrector)
         end
 
+        # @return [Corrector]
+        def merged_correctors
+          correctors.compact.inject(:merge!)
+        end
+
         def offenses
           @offenses ||= offenses_per_cop.flatten(1)
         end
--- a/lib/rubocop/cop/commissioner.rb
+++ b/lib/rubocop/cop/commissioner.rb
@@ -27,7 +27,7 @@ module RuboCop
 
         # @return [Corrector]
         def merged_correctors
-          correctors.inject(Corrector.new(processed_source), :merge!)
+          correctors.compact.inject(Corrector.new(processed_source), :merge!)
         end
 
         def offenses
--- a/lib/rubocop/cop/commissioner.rb
+++ b/lib/rubocop/cop/commissioner.rb
@@ -27,7 +27,7 @@ module RuboCop
 
         # @return [Corrector]
         def merged_correctors
-          correctors.compact.inject(:merge!)
+          correctors.inject(Corrector.new(processed_source), :merge!)
         end
 
         def offenses
--- a/lib/rubocop/rspec/cop_helper.rb
+++ b/lib/rubocop/rspec/cop_helper.rb
@@ -50,7 +50,7 @@ module CopHelper
     cop_array = Array(cops)
     team = RuboCop::Cop::Team.new(cop_array, nil, raise_error: true)
     report = team.investigate(processed_source)
-    @last_corrector = report.merged_correctors || RuboCop::Cop::Corrector.new(processed_source)
+    @last_corrector = report.merged_correctors
     report.offenses
   end
 end
--- a/lib/rubocop/rspec/cop_helper.rb
+++ b/lib/rubocop/rspec/cop_helper.rb
@@ -46,11 +46,20 @@ module CopHelper
     @last_corrector.rewrite
   end
 
-  def _investigate(cop, processed_source)
-    team = RuboCop::Cop::Team.new([cop], nil, raise_error: true)
+  def _investigate(cops, processed_source)
+    cop_array = Array(cops)
+    team = RuboCop::Cop::Team.new(cop_array, nil, raise_error: true)
     report = team.investigate(processed_source)
-    @last_corrector = report.correctors.first || RuboCop::Cop::Corrector.new(processed_source)
+    @last_corrector = _corrector(report, processed_source)
     report.offenses
+  end
+
+  def _corrector(report, processed_source)
+    corrector = report.correctors.first || RuboCop::Cop::Corrector.new(processed_source)
+    if report.correctors&.compact&.any?
+      report.correctors.compact[1..-1].each { |corr| corrector.merge!(corr) }
+    end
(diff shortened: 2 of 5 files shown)


dorner:
Just pushing this immediately for now to get feedback on the approach.

marcandre:
Cool!

First question is: order should not matter, so it shouldn't be "run first", simply "also run". We already have `other_cops` for extra config. One thing I was curious is: if we add the cops of these config to the running cops, will this create issue with the current test suite? Because if not, that would probably be a good starting point. Maybe `let(:other_cop_classes) { other_cops.each_key{ |cop_name| # get the cop }`, which could be overwritten. What do you think?

lib/rubocop/rspec/cop_helper.rb
```diff
@@ -46,12 +46,21 @@ def autocorrect_source(source, file = nil)
     @last_corrector.rewrite
   end
 
-  def _investigate(cop, processed_source)
-    team = RuboCop::Cop::Team.new([cop], nil, raise_error: true)
+  def _investigate(cops, processed_source)
+    cop_array = Array(cops)
+    team = RuboCop::Cop::Team.new(cop_array, nil, raise_error: true)
     report = team.investigate(processed_source)
-    @last_corrector = report.correctors.first || RuboCop::Cop::Corrector.new(processed_source)
```
marcandre, line 52: I'm not sure why there's a `first` there. All correctors should be merged, I think `correctors.compact.inject(:merge!)` should work, and that method should be added to `InvestigationReport`

dorner:
Thanks for giving it a look!

You'd think order doesn't matter, but unfortunately that's not true. In the original issue I was looking at fixing, running the cops in the opposite order actually did not surface the bug. That's because if you insert spaces and then a colon, everything is fine; inserting a colon and then spaces causes the break.

The approach of using `other_cops` is a good one - didn't think of doing it backwards that way :) Will give that a go! I'll rerun the suite and see if it causes any failures.

lib/rubocop/rspec/cop_helper.rb
dorner: Not sure I 100% understand how you'd like this to look, but I'll give it a go and you'll let me know if I did it wrong. 😛

dorner:
@marcandre bad news - this causes 47 failures in the current test suite. :( Right now the config is basically set so that the cop under test can *check* settings, but the *output* expects only that one cop to be run (so the other cops are catching issues that this cop doesn't care about).

lib/rubocop/cop/commissioner.rb
```diff
@@ -25,6 +25,11 @@ def correctors
           @correctors ||= cop_reports.map(&:corrector)
         end
 
+        # @return [Corrector]
+        def merged_correctors
+          correctors.compact.inject(:merge!)
```
marcandre, line 30: My bad, should be `inject(Corrector.new(processed_source), :merge!)` otherwise we modify the first corrector, or return `nil` if there are no correctors. Other looks good

lib/rubocop/rspec/cop_helper.rb
```diff
@@ -50,17 +50,9 @@ def _investigate(cops, processed_source)
     cop_array = Array(cops)
     team = RuboCop::Cop::Team.new(cop_array, nil, raise_error: true)
     report = team.investigate(processed_source)
-    @last_corrector = _corrector(report, processed_source)
+    @last_corrector = report.merged_correctors || RuboCop::Cop::Corrector.new(processed_source)
```
marcandre, line 53: The `|| ...` won't be necessary :-)

lib/rubocop/cop/commissioner.rb
dorner: Fixed!

marcandre:
Thanks for checking about `other_cops`.

The ordering thing bothers me tbh.

I'm wondering if another approach wouldn't be better, something like...

```ruby
expect_offense(<<~RUBY, cop_class: FirstCopClass)
 # ...
expect_correction(<<~RUBY)
  # intermediate stage
expect_offense(<<~RUBY)
  # ...
expect_correction(<<~RUBY)
  # final stage
```

I realize that you can't override `cop_class` or similar at the `expect_offense` level but that's something I was thinking would be cool to have.

dorner:
That's definitely a neat idea. More specifically, instead of relying on `cop_class`, rewriting it to allow you to pass in a cop to the things you're expecting. I can look at that approach as well.

marcandre:
What I had in mind is:
```ruby
  let(:cop) do
    cop_class.new(config, cop_options).tap do |cop|
      cop.send :begin_investigation, processed_source
    end
  end
# become
  def cop(cop_class: cop_class, cop_options: cop_options, config: config, source: source, processed_source: processed_source(source))
    # ...
  end
```

TBH, the fact that `cop` depends does the `begin_investigation` is kind of a problem and should not be necessary, so this complicates things...

dorner:
Hmm... I'm experimenting with the approach you mentioned above (using intermediate steps) and the problem is that the cops need to be part of the same rewriter in order to expose the bug. If you get the source out of the corrector and then pass it into the next cop, the bug won't happen. :( You really need to pass in both cops at the same time.

marcandre:
Oh wow, ok. So I guess the full CLI is a good regression test 🤷‍♂️

dorner:
In this case, yep. The other method would be using the `run_first_cops` method which is currently in this PR. Not super clean but at least possible without running through the entire test suite. I'm OK either way.

bbatsov:
@marcandre How are we proceeding here? I'm not very interested in this change, so I'll defer the decision on how to proceed to you.

marcandre:
It's on my (long) todo list :-)

marcandre:
I re-read this. The code is fine, I'm just not 100% we need it. The last bad interaction between cops we've had that I can recall is the one I created with the `#{ '  ' }` autocorrection, and this would not have helped, and wasn't necessary for testing.

So I'll close this for now and we can resurrect it if the need arises again.

PR #8365, opened 2020-07-19
Allow cops to restrict callbacks `on_send` to specific method names only

marcandre:
Many of the recent optimizations for Cops seem to be to modify `on_send` to check early for the `node.method_name` against a specific list of names.

This PR makes this pattern builtin so it is easy to do (just define `RESTRICT_ON_SEND = %i[foo bar baz]`) *and* makes it even more efficient.

Let's imagine 100 cops defining `on_send` that could have such a restricted method list. On a run over 10k `SendNode`, if these cops are optimized with the first thing in `on_send` being `return if SOME_LIST.include?(node.method_name)`, there would still be 100*10k calls to `on_send`, and 100*10k calls to `Set#include?` (assuming that `SOME_LIST` is a `Set`, as it should). With this PR, if the Cops are optimized with the right constant, then there would be only 10k hash lookups.

cc/ @fatkodima

### DIFF
--- a/lib/rubocop/cop/commissioner.rb
+++ b/lib/rubocop/cop/commissioner.rb
@@ -7,6 +7,9 @@ module RuboCop
     class Commissioner
       include RuboCop::AST::Traversal
 
+      RESTRICTED_CALLBACKS = %i[on_send on_csend].freeze
+      private_constant :RESTRICTED_CALLBACKS
+
       # How a Commissioner returns the results of the investigation
       # as a list of Cop::InvestigationReport and any errors caught
       # during the investigation.
@@ -43,6 +46,7 @@ module RuboCop
         @forces = forces
         @options = options
         @callbacks = Hash.new { |h, k| h[k] = cops_callbacks_for(k) }
+        @restricted_map = {}
 
         reset
       end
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/commissioner.rb
```diff
@@ -7,6 +7,9 @@ module Cop
     class Commissioner
       include RuboCop::AST::Traversal
 
+      RESTRICTED_NODE_TYPES = %i[on_send on_csend].freeze
```
fatkodima, line 10: Great work 👍 

This constant probably should be named differently, something like `RESTRICTED_CALLBACK_NAMES`.

marcandre: Indeed 😄

marcandre:
ping @bbatsov

marcandre:
ping @bbatsov Let me know if my ping frequency is too high too 😅

bbatsov:
The frequency is fine, but I couldn't catch up with all the open PRs. I like the proposed change, but I think we need to document it better if we expect someone to actually use it.

marcandre:
Cool. I'll document this and then merge. Thanks!

tas50:
@marcandre I reworked the 200+ cops in Cookstyle to use this new functionality where we could. Since we're autocorrecting a DSL there's a lot of times we're just looking for a deprecated method call that needs to be replaced with a new one. Lots and lots of simple on_sends. My testing shows this sped things up about 16% against a large set of cookbooks. Nice work.

marcandre:
16% is awesome, especially given the minimal trouble it is to specify it! 🎉
Thank @fatkodima who started using this technique manually 😄

PR #8369, opened 2020-07-19
Refactor documentation generator and make it usable for related projects

bquorning:
I have long wanted to turn cops_documentation.rake into a class instead having all the code inside a Rake task.

It has also annoyed me that we have _almost_ identical code existing in [rubocop](https://github.com/rubocop-hq/rubocop/blob/2d3f95c55d05faab3ad659ff9a56f408e53a2bf3/tasks/cops_documentation.rake#L12), [rubocop-rspec](https://github.com/rubocop-hq/rubocop-rspec/blob/0cad53a9ba5a96161be78a315262d9a8da959782/tasks/cops_documentation.rake#L13), [rubocop-rails](https://github.com/rubocop-hq/rubocop-rails/blob/2f09406560fbece1adced14edd68f7d8f9e3254e/tasks/cops_documentation.rake#L13), and [rubocop-performance](https://github.com/rubocop-hq/rubocop-performance/blob/8e6fc14ad8825bbc3a27620b9f1fb68524e542a9/tasks/cops_documentation.rake#L13).

This PR does two things (sorry):

1. Extracts the job of verifying if cop documentation is up to date into a separate Rake task called `confirm_cops_documentation`.
2. Extracts the logic for `generate_cops_documentation` into a class, which is initialized with an array of Departments for which you want to generate documentation. This means that this class can be re-used in rubocop-rspec, rubocop-rails, rubocop-performance etc.

I really, really want to refactor this extracted class (and add tests), ~but I’ve restrained myself to make this PR contain as few changes as possible~ and the last 4 commits is the least I could do for now.

Please check if the `docs/` changes look ok in these sub-projects:

- https://github.com/rubocop-hq/rubocop-rspec/compare/docs-without-own-rake-task
- https://github.com/rubocop-hq/rubocop-rails/compare/docs-without-own-rake-task
- https://github.com/rubocop-hq/rubocop-performance/compare/docs-without-own-rake-task
- https://github.com/rubocop-hq/rubocop-minitest/compare/docs-without-own-rake-task

Related issues:

- https://github.com/rubocop-hq/rubocop-rspec/issues/972
- https://github.com/rubocop-hq/rubocop-rspec/issues/922

-----------------

* [ ] Added tests.
* [-] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.


tasks/cops_documentation.rake
```diff
@@ -10,291 +11,21 @@ end
...
+  deps = %w[Bundler Gemspec Layout Lint Metrics Migration Naming Security Style]
+  CopsDocumentationGenerator.new(departments: deps).call
+end
 
-    YARD::Registry.load!
-    cops.departments.sort!.each do |department|
-      print_cops_of_department(cops, department, config)
-    end
+desc 'Confirm that documentation is up to date'
+task confirm_cops_documentation: :generate_cops_documentation do
```
bbatsov, line 19: I'd probably use "verify" or "check" here.

bquorning: Changed to `verify_cops_documentation`

lib/rubocop/cops_documentation_generator.rb
```diff
@@ -0,0 +1,290 @@
...
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  def properties(cop)
+    header = [
+      'Enabled by default', 'Safe', 'Supports autocorrection', 'VersionAdded',
+      'VersionChanged'
+    ]
+    autocorrect = if cop.support_autocorrect?
+                    "Yes#{' (Unsafe)' unless cop.new(config).safe_autocorrect?}"
```
pirj, line 76: `cop.new(config).safe_autocorrect?}`
Can this be fetched from the config instead? `cop_config.fetch('SafeAutoCorrect', true)`?

lib/rubocop/cops_documentation_generator.rb
```diff
@@ -0,0 +1,290 @@
...
+    file_name = "#{Dir.pwd}/docs/modules/ROOT/pages/cops_#{department.downcase}.adoc"
+    File.open(file_name, 'w') do |file|
+      puts "* generated #{file_name}"
+      file.write(content.strip + "\n")
+    end
+  end
+
+  def print_cop_with_doc(cop)
+    cop_config = config.for_cop(cop)
+    non_display_keys = %w[
```
pirj, line 230: A good candidate for a const.

lib/rubocop/cops_documentation_generator.rb
```diff
@@ -0,0 +1,290 @@
...
+  #
+  def initialize(departments: [])
+    @departments = departments.map(&:to_sym).sort!
+    all_cops = RuboCop::Cop::Cop.registry
+    qualified_cops = all_cops.select { |cop| qualified_cop?(cop) }
+    @cops = RuboCop::Cop::Registry.new(qualified_cops)
+    @config = RuboCop::ConfigLoader.default_configuration
+  end
+
+  def call
```
pirj, line 19: `generate!`?

lib/rubocop/cops_documentation_generator.rb
bquorning: It can: https://github.com/rubocop-hq/rubocop/blob/343f62e4555be0470326f47af219689e21c61a37/lib/rubocop/cop/autocorrect_logic.rb#L23-L26

lib/rubocop/cops_documentation_generator.rb
bquorning: In a later refactoring maybe?

lib/rubocop/cops_documentation_generator.rb
bquorning: I always thought that for classes that do one thing, and the classname describes what it does, `#call` is the appropriate method name. It could be `#run` or `#main`, but since `#call` is what Ruby uses for Lambdas/Procs, I tend to use it for these purposes.

lib/rubocop/cops_documentation_generator.rb
pirj: Now I see, makes sense 👍
The need to instantiate the cop triggered me, but since there's no other way around I guess it's totally fine.

bquorning: I mean, I think we could just call `cop_config.fetch('Safe', true) && cop_config.fetch('SafeAutoCorrect', true)`, but we’d be duplicating the logic and producing wrong results if the implementation of `#safe_autocorrect?` changes in the future. Would it be possible to move the logic into the `cop_config` object?

marcandre: > Would it be possible to move the logic into the `cop_config` object?

I'd second that.

bquorning: `cop_config` is a Hash, but it _might_ be possible to add the method by changing `RuboCop::Config#for_cop` to e.g.
```rb
@for_cop[cop.respond_to?(:cop_name) ? cop.cop_name : cop].tap do |config|
  def config.safe_autocorrect?
    fetch('Safe', true) && fetch('SafeAutoCorrect', true)
  end
end
```

Let’s consider this in a separate PR please. I think this PR is large enough as is.

marcandre: Oh, my bad, forget it then...

pirj: 😱

bbatsov:
> Please check if the docs/ changes look ok in these sub-projects:

They look fine to me. We should have done this a long time ago!

bquorning:
I’ve rebased, and it turned out I could basically remove the commit b23deeaad243339beae5a807973fe06fc8a65546 which did
```diff
 def call
-  cops   = RuboCop::Cop::Cop.registry
+  all_cops       = RuboCop::Cop::Cop.registry
+  qualified_cops = all_cops.select { |cop| qualified_cop?(cop) }
+  cops           = RuboCop::Cop::Registry.new(qualified_cops)
   config = RuboCop::ConfigLoader.default_configuration
 end
```
Since we only loop over the passed-in `departments`, it doesn’t matter if there are too many cops in the registry. Much simpler 🎉

pirj:
👏 
I expect to see a followup in RuboCop RSpec!

bquorning:
Definitely, after the next release of RuboCop.

Issue #7876, opened 2020-04-14
RequiredRubyVersion


marcandre:
The `Gemspec/RequiredRubyVersion` cop checks that `spec.required_ruby_version=` (if present) corresponds to `TargetRubyVersion`.

That's good.

I'd like a more basic check: that `required_ruby_version` is there at all. I believe that it should be an accepted community standard to have it; there is no valid reason not to have a `required_ruby_version` specified that I can think of.

I opened an [rfc about that](https://github.com/rubygems/rfcs/pull/26), but it would probably take time before this is accepted / implemented / released and mostly updated... A Rubocop might make things faster.

In case there is a `.travis.yml` file, it would be nice to check it correponds too.

1) How does a cop making `required_ruby_version` mandatory sound?

2) If that's a go, should it be a different cop?

3) What about `.travis.yml` check?

bbatsov:
I like the idea. I can't come up with a good reason why a gem shouldn't specify some required Ruby version.

marcandre:
2. should it be a different cop?

3. What about .travis.yml check?

fatkodima:
Should this gets closed?

PR #8370, opened 2020-07-19
Enhance `Gemspec/RequiredRubyVersion` cop with check that `required_ruby_version` is specified

fatkodima:
Closes? #7876

It seems to me that no new cop is needed and the existing cop should be extended, but I have no strong opinion on this.

cc/ @marcandre


marcandre:
Great PR imo, I'll let @bbatsov have final say

PR #8376, opened 2020-07-20
Add new `Lint/MissingSuper` cop

fatkodima:
For context - https://github.com/rubocop-hq/ruby-style-guide/issues/809

1. I'm not sure if this should go to `Lint` or `Style` department.
2. There is already a `Style/MethodMissingSuper` cop, checking only `method_missing` callback. Should that cop be deprecated in favor of this new cop?


lib/rubocop/cop/base.rb
```diff
@@ -137,6 +137,7 @@ def external_dependency_checksum
       end
 
       def self.inherited(subclass)
+        super
```
marcandre, line 140: 😊

lib/rubocop/cop/lint/missing_super.rb
```diff
@@ -0,0 +1,90 @@
...
+          !(node.method?(:initialize) || callback_method?(node.method_name)) ||
+            contains_super?(node)
+        end
+
+        def callback_method?(method_name)
+          CALLBACK_NAMES.include?(method_name)
+        end
+
+        def contains_super?(node)
+          node.each_descendant.any? { |c| c.super_type? || c.zsuper_type? }
```
marcandre, line 76: May be preferable: `node.each_descendant(:super, :zsuper).any?`

marcandre:
Great cop & PR <3

fatkodima:
Thanks 😊 
I'm also in need of answers to my questions to finish it.

marcandre:
`Style/MethodMissingSuper`: I'm surprised it doesn't check `respond_to_missing?` also, and that if you have `respond_to` you should also have a `respond_to_missing?` method.

fatkodima:
> and that if you have respond_to you should also have a respond_to_missing? method.

Looks like a `Style/MissingRespondToMissing` is for this.

bbatsov:
> There is already a Style/MethodMissingSuper cop, checking only method_missing callback. Should that cop be deprecated in favor of this new cop?

That sounds reasonable. 

> I'm not sure if this should go to Lint or Style department.

Lint. I'm surprised we've put the other two cops until Style, but I guess that's historical - in the very beginning Style was the only department. :-) 

> Looks like a Style/MissingRespondToMissing is for this.

E.g. this should probably go under Lint.

lib/rubocop/cop/lint/missing_super.rb
```diff
@@ -0,0 +1,90 @@
...
+      #       do_something
+      #     end
+      #   end
+      #
+      class MissingSuper < Base
+        CONSTRUCTOR_MSG = 'Call `super` to initialize state of the parent class.'
+        CALLBACK_MSG    = 'Call `super` to invoke callback defined in the parent class.'
+
+        STATELESS_CLASSES = %w[BasicObject Object].freeze
+        CALLBACK_NAMES = %i[inherited included extended prepended].to_set.freeze
```
bbatsov, line 45: Probably we can have a better name here, but I don't have a solid proposal - e.g. `ClassLifecycleCallbacks` or something along those lines.

bbatsov:
I'm wondering if we should expose the ability to add more stateless classes and callbacks, but probably that's an overkill at that point. Good work!

fatkodima:
> >There is already a Style/MethodMissingSuper cop, checking only method_missing callback. Should that cop be deprecated in favor of this new cop?

>That sounds reasonable.

Implemented this.

> but probably that's an overkill at that point. 

Agreed.

lib/rubocop/cop/lint/missing_super.rb
```diff
@@ -0,0 +1,100 @@
...
+
+          add_offense(node, message: CALLBACK_MSG)
+        end
+
+        private
+
+        def should_not_check?(node)
+          !(node.method?(:initialize) || callback_method_def?(node)) ||
+            contains_super?(node)
+        end
```
Drenmi, line 71: I was a bit surprised that this predicate is inverted. It also means it can't short circuit, but must evaluate everything. Also the name makes it sound like it's a pre-check, but it's also the actual check itself. Compare, e.g.:

```ruby
def offender?(node)
  (node.method?(:initialize) || callback_method_def?(node)) && !contains_super?(node)
end
```

fatkodima: Yes, thanks, much better.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New features
 
 * [#8322](https://github.com/rubocop-hq/rubocop/pull/8322): Support autocorrect for `Style/CaseEquality` cop. ([@fatkodima][])
+* [#8376](https://github.com/rubocop-hq/rubocop/pull/8376): Add new `Lint/MissingSuper` cop. ([@fatkodima][])
```
rrosenblum, line 8: There should probably be a note in here that this replaces the `Style/MethodMissingSuper` cop.

lib/rubocop/cop/lint/missing_super.rb
```diff
@@ -0,0 +1,99 @@
...
+             CLASS_LIFECYCLE_CALLBACKS.include?(method_name)
+
+            node.each_ancestor(:class, :sclass, :module).first
+          elsif MODULE_LIFECYCLE_CALLBACKS.include?(method_name)
+            node.each_ancestor(:module).first
+          end
+        end
+
+        def contains_super?(node)
+          node.each_descendant(:super, :zsuper).any?
```
rrosenblum, line 85: `any` is not the same thing as `!.empty?`. It will work the same way in most cases, however this will fail if the array contains `nil`.

```ruby
[1] pry(main)> [nil].any?
=> false
[2] pry(main)> ![nil].empty?
=> true
```

spec/rubocop/cop/lint/missing_super_spec.rb
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Lint::MissingSuper do
```
rrosenblum, line 3: I think this is missing some test cases. I don't see any tests that cover `method_missing`

lib/rubocop/cop/lint/missing_super.rb
fatkodima: Correct, but `each_descendant(:super, :zsuper)` won't return array with nils - either empty or with `Node` instances.

ruanltbg:
Hi @fatkodima && @marcandre.
This cop is complaining about: 
```
  def self.included(base)
    base.before(:each) { Warden.test_mode! }
    base.after(:each) { Warden.test_reset! }
  end
```
I don't see any reason for me to add a `super` anywhere in this code. Maybe `included` should be excluded from this cop? What are your thoughts on this?

Full code:

```
module RequestSpecHelper
  include Warden::Test::Helpers

  def self.included(base)
    base.before(:each) { Warden.test_mode! }
    base.after(:each) { Warden.test_reset! }
  end

  def sign_in(resource)
    login_as(resource, scope: warden_scope(resource))
  end

  def sign_out(resource)
    logout(warden_scope(resource))
  end

  private

  def warden_scope(resource)
    resource.class.name.underscore.to_sym
  end
end
```

The result from rubocop:
```
Offenses:

spec/support/request_spec_helper.rb:4:3: W: Lint/MissingSuper: Call super to invoke callback defined in the parent class.
  def self.included(base) ...
  ^^^^^^^^^^^^^^^^^^^^^^^

187 files inspected, 1 offense detected
```

marcandre:
Upon reflection, I have to agree that the case of `included` and `extended` callbacks is weaker than I thought, because including a method doesn't mean one inherits from its `self.included` method.

Unless there's objection, I'd remove them.

fatkodima:
I'd also add missing `method_added/_removed/_undefined`, which I forgot to add.

marcandre:
@fatkodima oh, good catch. Want to take care of this? You seem to write up PRs faster than I can review them 😅

fatkodima:
😄 
If you have time, please do. I'm working on another one right now 😄

marcandre:
Sure thing: #8480

haines:
Would it be possible to make the list of stateless classes configurable? We have a lot of offenses for children of a couple of base classes where we know it's unnecessary to call `super` in `initialize`.

silva96:
I have a similar problem like @haines my services classes inherit but don't need to call supper

https://github.com/rubocop-hq/ruby-style-guide/issues/809#issuecomment-673096138

I would like to exclude those classes or any that inherit from ApplicationService.

lib/rubocop/cop/base.rb
pirj: @fatkodima Marc-André is smiling for a reason.
There's no `super` for a class with no parent.

spec/rubocop/cop/lint/missing_super_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+          end
+        end
+      RUBY
+    end
+
+    it 'does not register an offense when `method_missing?` contains `super` call' do
+      expect_no_offenses(<<~RUBY)
+        class Foo
+          def method_missing(*args)
+            super
```
pirj, line 112: Is this calling `method_missing` on `Object`?

spec/rubocop/cop/lint/missing_super_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+          end
+        end
+      RUBY
+    end
+
+    it 'registers an offense when class callback without `super` call' do
+      expect_offense(<<~RUBY)
+        class Foo
+          def self.inherited(base)
+          ^^^^^^^^^^^^^^^^^^^^^^^^ Call `super` to invoke callback defined in the parent class.
```
pirj, line 70: What would `super` refer to? `Foo` has `Object` as an implicit parent.
Why `super` is needed in this case?

lib/rubocop/cop/base.rb
marcandre: Actually I still like adding it, and I was surprise I didn't in the first place.

PR #8377, opened 2020-07-20
Added new `Lint/TopLevelReturnWithArgument` cop

### LINKED ISSUE #7425: New cop: Detect return with argument on the top-level

`return` has been allowed on the top-level since Ruby 2.4.

```ruby
foo

return # Since Ruby 2.4.

bar # It is never executed.
```

But the argument of `return` is ignored always.

```ruby
return 1 # 1 is always ignored.
```

It will be detected since Ruby 2.7. https://github.com/ruby/ruby/commit/aa97410b0a85cb4ceb956ab943b5eee92a128411
But RuboCop doesn't.

Add a new cop to detect the problem. Maybe it will be named `Lint/TopLevelReturnWithArgument`.

nothing

## Additional context

nothing

IamRaviTejaG:
Fixes #7425 

This cop detects top level return statements with argument(s).

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/top_level_return_with_argument.rb
```diff
@@ -0,0 +1,63 @@
...
+      #
+      # @example
+      #
+      #   # Detected since Ruby 2.7
+      #   return 1 # 1 is always ignored.
+      class TopLevelReturnWithArgument < Cop
+        MSG = 'Top level return with argument detected.'
+
+        def on_return(return_node)
+          add_offense(return_node) if ancestors_valid?(return_node) && !return_node.arguments.empty?
```
marcandre, line 46: `return_node.arguments.empty?` can also be written `return_node.arguments?`

lib/rubocop/cop/lint/top_level_return_with_argument.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+
+        private
+
+        def ancestors_valid?(return_node)
+          return true if return_node.parent.nil?
+
+          if return_node.parent.instance_of?(AST::Node) && return_node.parent.parent.nil?
```
marcandre, line 54: Won't `return_node.parent.instance_of?(AST::Node)` always be true?

lib/rubocop/cop/lint/top_level_return_with_argument.rb
```diff
@@ -0,0 +1,63 @@
...
+            return true
+          end
+
+          false
```
marcandre, line 58: I'm surprised there's no cop against `if condition; return true; end; return false` => `return condition`

lib/rubocop/cop/lint/top_level_return_with_argument.rb
IamRaviTejaG: `return_node.parent.is_a?(AST::Node)` will always return true, but `return_node.parent.instance_of?(AST::Node)` would be true only if that object is strictly of that exact class, and not the subclass.

lib/rubocop/cop/lint/top_level_return_with_argument.rb
IamRaviTejaG: I did write it in one line, but then it exceeded the 100 characters limit, so had to change it to this.

IamRaviTejaG: @marcandre There actually is a cop that suggests this, one suggests for single line if and the other suggests to remove redundant false.

lib/rubocop/cop/lint/top_level_return_with_argument.rb
marcandre: Oh, right, my bad. Then I would say you need to tweak this test; you should not rely on the class of `Node`, only on their `type`. Here anyways I believe it is not the best anyways; you're excluding `if` for example.

lib/rubocop/cop/lint/top_level_return_with_argument.rb
marcandre: In this case here, you can remove the `if`, the `return true\n  end\n  false` completely...

lib/rubocop/cop/lint/top_level_return_with_argument.rb
IamRaviTejaG: But to find out the type in Ruby, we do `is_a?(ClassName)` or `instance_of?(ClassName)`. If not, then I guess I quite did not get what you suggested.

lib/rubocop/cop/lint/top_level_return_with_argument.rb
IamRaviTejaG: The if condition checks for the types and returns things appropriately, if we remove that, the cop will report false positives.

lib/rubocop/cop/lint/top_level_return_with_argument.rb
IamRaviTejaG: Aah yes!

lib/rubocop/cop/lint/top_level_return_with_argument.rb
IamRaviTejaG: Updated this! Thanks for the idea. 😄

IamRaviTejaG:
@marcandre Updated the PR. Now `return 42 if a == b` are being caught.

marcandre:
> @marcandre Updated the PR. Now `return 42 if a == b` are being caught.

RIght, good, but it won't catch `if a == b; warn 'hey'; return 42; end`...

I was trying to hint at a more general logic. Unless I'm missing something, you can check all ancestors for either a `block`, `def` or `defs`; if you don't find any then you are still at the top level scope.

IamRaviTejaG:
> 
> 
> > @marcandre Updated the PR. Now `return 42 if a == b` are being caught.
> 
> RIght, good, but it won't catch `if a == b; warn 'hey'; return 42; end`...
> 
> I was trying to hint at a more general logic. Unless I'm missing something, you can check all ancestors for either a `block`, `def` or `defs`; if you don't find any then you are still at the top level scope.

Now this checks for the ancestors to not belong to the `AST::BlockNode` or `AST::DefNode` class.

marcandre:
> Now this checks for the ancestors to not belong to the `AST::BlockNode` or `AST::DefNode` class.

Much better. Please do not rely on classes ever, as this is subject to change. You should rely on the result of `Node#type` or `Node#block_type? / def_type? / defs_type?` In this case, `each_ancestor` already accepts a list of types...

IamRaviTejaG:
> 
> 
> > Now this checks for the ancestors to not belong to the `AST::BlockNode` or `AST::DefNode` class.
> 
> Much better. Please do not rely on classes ever, as this is subject to change. You should rely on the result of `Node#type` or `Node#block_type? / def_type? / defs_type?` In this case, `each_ancestor` already accepts a list of types...

Aah, this suggestion and the resulting changes made the code so compact! Can't thank you enough for this.

lib/rubocop/cop/lint/top_level_return_with_argument.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        def on_return(return_node)
+          add_offense(return_node) if ancestors_valid?(return_node) && return_node.arguments?
+        end
+
+        private
+
+        def ancestors_valid?(return_node)
+          prohibited_ancestors = return_node.each_ancestor(:block, :def, :defs)
+          return false if prohibited_ancestors.any?
```
marcandre, line 29: Personal preference, I'd simply write `prohibited_ancestors.none?`, no return, no if...

lib/rubocop/cop/lint/top_level_return_with_argument.rb
```diff
@@ -0,0 +1,36 @@
...
+      #
+      # @example
+      #
+      #   # Detected since Ruby 2.7
+      #   return 1 # 1 is always ignored.
+      class TopLevelReturnWithArgument < Cop
+        MSG = 'Top level return with argument detected.'
+
+        def on_return(return_node)
+          add_offense(return_node) if ancestors_valid?(return_node) && return_node.arguments?
```
marcandre, line 22: I'd switch the conditions' order, as `arguments?` is very quick while `ancestors_valid?` might be slower

marcandre:
> Aah, this suggestion and the resulting changes made the code so compact! Can't thank you enough for this.

Cool, I'm happy you're happy about the resulting code. I have two last remarks for styling/performance...

lib/rubocop/cop/lint/top_level_return_with_argument.rb
IamRaviTejaG: Yes, this also makes a lot of sense given Ruby's short circuit evaluation.

IamRaviTejaG:
> Cool, I'm happy you're happy about the resulting code. I have two last remarks for styling/performance...
> 
> Also you'll have to rebase your branch

Made performance & styling related changes, and also rebased my branch. 😄

docs/modules/ROOT/pages/cops_lint.adoc
```diff
@@ -3358,6 +3358,34 @@ def to_json(*_args)
...
+| -
+|===
+
+This cop checks for top level return with arguments. If there is a
+top-level return statement with an argument, then the argument is
+always ignored. This is detected automatically since Ruby 2.7.
+
+This cop works by validating the ancestors of the return node. A
+top-level return node's ancestors will never belong to `AST::BlockNode`
+or `AST::DefNode` class.
```
marcandre, line 3379: I'd remove this comment; implementation detail is too technical and not useful to cop users.

spec/rubocop/cop/lint/top_level_return_with_argument_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+  end
+
+  context 'Code segment with block level returns other than the top-level return' do
+    it 'Expects no offense from the return without arguments' do
+      expect_no_offenses(<<~RUBY)
+        foo
+
+        [1, 2, 3, 4, 5].each { |n| return n }
+
+        return # Should raise a `top level return with argument detected` offense
```
marcandre, line 39: Change comment...

spec/rubocop/cop/lint/top_level_return_with_argument_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  context 'Code segment with method-level return statements' do
+    it 'Expects no offense from the method-level return statement' do
+      expect_no_offenses(<<~RUBY)
+        def method
+          return 'Hello World'
+        end
+      RUBY
+    end
+
```
marcandre, line 67: Remove this spec, it is redundant with the next spec.

spec/rubocop/cop/lint/top_level_return_with_argument_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+        bar
+      RUBY
+    end
+
+    it 'Expects an offense from the return with arguments' do
+      expect_offense(<<~RUBY)
+        if a == b; warn 'hey'; return 42; end
+                               ^^^^^^^^^ Top level return with argument detected.
+      RUBY
+    end
```
marcandre, line 130: Seems redundant with previous spec

spec/rubocop/cop/lint/top_level_return_with_argument_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    it 'Expects no offense from the return without arguments' do
+      expect_no_offenses(<<~RUBY)
+        if a == b; warn 'hey'; return; end
+      RUBY
+    end
+
```
marcandre, line 137: Seems redundant with next spec

marcandre:
Implementation is perfect now 🎉. I think many specs can be removed and some doc can be modified.

lib/rubocop/cop/lint/top_level_return_with_argument.rb
```diff
@@ -0,0 +1,34 @@
...
+  module Cop
+    module Lint
+      # This cop checks for top level return with arguments. If there is a
+      # top-level return statement with an argument, then the argument is
+      # always ignored. This is detected automatically since Ruby 2.7.
+      #
+      # This cop works by validating the ancestors of the return node. A
+      # top-level return node's ancestors will never belong to `AST::BlockNode`
+      # or `AST::DefNode` class.
+      #
```
marcandre, line 13: Please update and put inside the class so it doesn't show in the doc

IamRaviTejaG: Updated this and added it inside class.

spec/rubocop/cop/lint/top_level_return_with_argument_spec.rb
IamRaviTejaG: Yes, removed this.

spec/rubocop/cop/lint/top_level_return_with_argument_spec.rb
IamRaviTejaG: Yes, removed this spec.

spec/rubocop/cop/lint/top_level_return_with_argument_spec.rb
IamRaviTejaG: Done.

spec/rubocop/cop/lint/top_level_return_with_argument_spec.rb
IamRaviTejaG: Removed.

docs/modules/ROOT/pages/cops_lint.adoc
IamRaviTejaG: Makes sense to me, removed this. Added this inside the class instead.

IamRaviTejaG:
Hey @bbatsov, can you please review this PR. 😄

marcandre:
It's ok, I probably should have simply merged it, looks really good. Looks like there are conflicts again, could you rebase and I'll merge it?

IamRaviTejaG:
@marcandre One test failed erroneously, and I cannot seem to re-run it, can you please re-run it? Once that passes this should be good to merge.

marcandre:
Thanks @IamRaviTejaG 🎉

PR #8378, opened 2020-07-21
Add new `Style/StringConcatenation` cop

### LINKED ISSUE #5067: [New Cop] Style/PreferInterpolation
I'd like to request a new cop, similar to eslint's [prefer-template](https://eslint.org/docs/rules/prefer-template) rule. It would check for places where string interpolation could be used in place of concatenation.

Here's the "bad/good" example:

```ruby
# bad
a = "abc"
b = a + "def"

# good
a = "abc"
b = "#{a} def"
```

fatkodima:
https://rubystyle.guide/#string-interpolation

```ruby
# bad
email_with_name = user.name + ' <' + user.email + '>'

# good
email_with_name = "#{user.name} <#{user.email}>"
```

Closes #5067 

1. Another good name for this cop, I think, is `StringConcatenation`.
2. Made it `Safe: false`, because there can be [very rare cases](https://github.com/rubocop-hq/rubocop/issues/5067#issuecomment-345468292) where it can give false positives.
3. It may be also a good idea to add some config that indicates after which amount of terms to add an offense.


lib/rubocop/cop/correctors/line_break_corrector.rb
```diff
@@ -39,7 +39,7 @@ def move_comment(eol_comment:, node:, corrector:)
 
           text = eol_comment.loc.expression.source
           corrector.insert_before(node,
-                                  text + "\n" + (' ' * node.loc.keyword.column))
+                                  "#{text}\n#{(' ' * node.loc.keyword.column)}")
```
marcandre, line 42: Would be nice if the autocorrection dropped those `()`

spec/rubocop/cop/style/prefer_interpolation_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  it 'registers an offense and corrects for string concatenation as part of other expression' do
+    expect_offense(<<~RUBY)
+      users = (user.name + ' ' + user.email) * 5
+               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer string interpolation instead of string concatenation.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      users = ("\#{user.name} \#{user.email}") * 5
```
marcandre, line 24: Would be nice to drop these outter `()` too, but maybe a different cop should do this?

spec/rubocop/cop/style/prefer_interpolation_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    RUBY
+  end
+
+  it 'correctly handles strings with special characters' do
+    new_source = autocorrect_source(
```
marcandre, line 29: Why not use `expect_correction`?

marcandre:
Great PR (as usual 😆)

config/default.yml
```diff
@@ -3624,6 +3624,13 @@ Style/PerlBackrefs:
   Enabled: true
   VersionAdded: '0.13'
 
+Style/PreferInterpolation:
```
bbatsov, line 3627: In general it's best to have more neural cop names, as most cops tend to be configurable in the end.

bbatsov:
> Another good name for this cop, I think, is StringConcatenation.

I think I like this name better. Might also be a good idea to have the ability to autocorrect using `format`, but that's not a big deal.

fatkodima:
Changed name to `Style/StringConcatenation`.

> Would be nice to drop these outer () too, but maybe a different cop should do this?

Yes, to not complicate the actual implementation and also this is needed in many other cops, so I think it should be better handled by a different cop (named like `Style/RedundantParentheses`, for example). I will try to create one in the future, if no one gets ahead of me.

marcandre:
> Changed name to `Style/StringConcatenation`.
> 
> > Would be nice to drop these outer () too, but maybe a different cop should do this?
> 
> Yes, to not complicate the actual implementation and also this is needed in many other cops, so I think it should be better handled by a different cop (named like `Style/RedundantParentheses`, for example). I will try to create one in the future, if no one gets ahead of me.

Agreed. I created #8379 to be sure we don't forget :+1:

PR #8382, opened 2020-07-22
Remove redundant `require 'set'`

koic:
Follow https://github.com/rubocop-hq/rubocop/commit/8798bc8.

~~RuboCop AST 0.1.0 requires `Set`.~~
~~https://github.com/rubocop-hq/rubocop-ast/commit/8e595c3.~~

~~RuboCop requires RuboCop AST 0.1.0 or higher. Therefore this PR removes redundant `require 'set'`.~~

This PR makes RuboCop to explicitly require `set` only once at lib/rubocop.rb
https://github.com/rubocop-hq/rubocop/blob/v0.88.0/lib/rubocop.rb#L6

It focuses on the obviously redundant code in the rubocop repository.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/target_finder.rb
+++ b/lib/rubocop/target_finder.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'set'
-
 module RuboCop
   # This class finds target files to inspect by scanning the directory tree
   # and picking ruby files.
--- a/lib/rubocop/cop/layout/empty_lines.rb
+++ b/lib/rubocop/cop/layout/empty_lines.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'set'
-
 module RuboCop
   module Cop
     module Layout
--- a/lib/rubocop/cop/layout/extra_spacing.rb
+++ b/lib/rubocop/cop/layout/extra_spacing.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'set'
-
 module RuboCop
   module Cop
     module Layout
--- a/lib/rubocop/cop/layout/indentation_style.rb
+++ b/lib/rubocop/cop/layout/indentation_style.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'set'
-
 module RuboCop
   module Cop
     module Layout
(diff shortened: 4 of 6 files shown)


lib/rubocop/target_finder.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'set'
```
bquorning, line 3: This file explicitly uses a `Set`, so I think the require should stay.

marcandre:
> I think files that explicitly calls e.g. `Set.new` should still keep the require.

I'm curious why. I hope I didn't start a big argument (especially since there's really not much at stake). My opinion if that [sets should be builtin](https://bugs.ruby-lang.org/issues/16989) and most gems should currently require it, once, where the files are loaded (like we do in `rubocop` & `rubocop-ast`). We don't require other external dependencies where we use them either (e.g. `parser`)...

bquorning:
First, I agree that Set should be built-in.

> gems should currently require it, once, where the files are loaded (like we do in rubocop & rubocop-ast).

I read the PR description as if RuboCop should not require Set, and just depend on rubocop-ast requiring it. That feels wrong.

Re. requiring external dependencies in each file where they’re used: It makes testing easier/faster and it makes code easier to reuse in smaller chunks (think e.g. requiring activesupport/core_ext/string ) if you aren’t depending on always loading up the entire library and all it’s dependencies.

marcandre:
> I read the PR description as if RuboCop should not require Set, and just depend on rubocop-ast requiring it. That feels wrong.

That's my fault, I mentioned `rubocop-ast` but in this case `rubocop` also requires it explicitly. Note that `parser` is only required once in `rubocop-ast`.

> requiring external dependencies in each file where they’re used: It makes testing easier/faster  and it makes code easier to reuse in smaller chunks

You are right for reuse in smaller chunks, but we don't (yet?) support it.

I don't feel strongly about it.

marcandre:
> First, I agree that Set should be built-in.

If you have the time, please comment positively on the `Set` issues (and negatively only here 😆). There was basically zero feedback by anybody 🤷‍♂️

koic:
Thank you for the argument. I've updated this PR so that RuboCop explicitly requires `set` only once in the following lib/rubocop.rb.
https://github.com/rubocop-hq/rubocop/blob/v0.88.0/lib/rubocop.rb#L6

It refocuses on the obviously redundant code in the rubocop repository.

PR #8383, opened 2020-07-22
Support autocorrect for `Lint/Loop` cop

fatkodima:


marcandre:
I think you might be loosing comments in this case:
```ruby
begin
        something
end while test && # don't remove me
          test2
```

fatkodima:
I hope it is impossible to find the code like that 😄 
This is much possible:
```ruby
begin
  something
end while test # don't remove me
```

And it will be correctly handled.

bbatsov:
I wonder why we made this a lint cop, as I definitely don't remember it today.  Seems we could have written better documentation. 😆

marcandre:
Am I weird?

I prefer

```ruby
loop do
  # ...
  break if condition
end
```

But I'd rather have
```ruby
begin
   # ...
end while condition
# than
loop do
   # ...

  break unless condition
end
```

The answer is probably "yeah". I'm glad it's rare that there's a reason to loop

bbatsov:
You're weird for sure, but then again - I've never had to use `loop` in my entire Ruby career. 😆

fatkodima:
>  I've never had to use loop in my entire Ruby career. 😆

I have used `loop`, but I would say I have never used `for` loop (but I do not see it that bad as everybody talks), do not remember if I ever really used `protected` keyword with its intended purpose and never found the case where I really needed to made the choice between `proc` and `lambda` because of their differences.

bbatsov:

fatkodima:
Rebased.

PR #8387, opened 2020-07-23
Add new `Lint/SelfAssignment` cop

### LINKED ISSUE #8291: New `Lint/SelfAssignment` cop
Self assignments have no effect, so probably those are a copy-parse error, missed `self` for properties or due to incomplete refactoring.

```ruby
class User
  def fullname=(value)
    name, surname = value.split
    name = name # forgot to use `self`
    self.surname = surname
  end
end

x = x
[a, b] = [a, b]
```

There is already a `Lint/UselessAssignment` cop which checks for unused variables after assignment. That cop is a little difficult, as I can say, to extend with this and created with a little different purpose. So I'm proposing creating a new `Lint/SelfAssignment` cop to handle described use cases. Wdyt?

Inspired by https://eslint.org/docs/rules/no-self-assign and https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html#sa-self-assignment-of-field-sa-field-self-assignment   and http://errorprone.info/bugpattern/SelfAssignment

fatkodima:
Closes #8291


bbatsov:
Nice!

PR #8389, opened 2020-07-23
Add new `Lint/DuplicateRescueException` cop

fatkodima:
This cop checks that there are no repeated exceptions used in 'rescue' expressions.

It is an addition to existing similar cops - [Lint/DuplicateCaseCondition](https://docs.rubocop.org/rubocop/0.88/cops_lint.html#lintduplicatecasecondition) and [Lint/DuplicateElsifCondition]( https://docs.rubocop.org/rubocop/0.88/cops_lint.html#lintduplicateelsifcondition)


lib/rubocop/cop/lint/duplicate_rescue_exception.rb
```diff
@@ -0,0 +1,66 @@
...
+        include RescueNode
+
+        MSG = 'Duplicate `rescue` exception detected.'
+
+        def on_rescue(node)
+          return if rescue_modifier?(node)
+
+          _body, *rescues, _else = *node
+
+          rescues.each_with_object([]) do |rescue_node, previous|
```
marcandre, line 38: I'd use a `Set` and `Set#add?` should simplify the code even further...

PR #8390, opened 2020-07-23
Add new `Style/SoleNestedConditional` cop

fatkodima:
This cop checks for places where nested conditionals can be merged into outer ones, like:
```ruby
# bad
if condition_a
  if condition_b
    do_something
  end
end

# good
if condition_a && condition_b
  do_something
end
```

This, imo, makes code more readable and, for sure, reduces nestings.

Inspired by [Style/IfInsideElse](https://docs.rubocop.org/rubocop/0.88/cops_style.html#styleifinsideelse) cop.
That cop also has (disabled, btw, in `rubocop`s `.rubocop.yml` file) `AllowIfModifier` setting. I haven't added that as I haven't get the point of its usefulness and how this 
```ruby
if condition_a
  do_something if condition_b
end
```

can be better than
```ruby
if condition_a && condition_b
  do_something
end
```

Probably, the cop name and wordings can be improved.


marcandre:
I can see how this:
```ruby
if important_criteria
  do_something(:important, 'with a long argument') unless unlikely_minor_cond?
end

# could be better than
if important_criteria && !unlikely_minor_cond?
  do_something(:important, 'with a long argument')
end
```

fatkodima:
> Also, did you handle the cases if (var = calculate) manually? That's cheating 🤣. I think that's if a criteria is an assignment than no offense should be raised, no?

Not sure. Personally, I'm happy with offense raised in this case 😄  I extracted that cases as a separate assignment before conditional, but would also just append nested conditions to that assignment. I looks more simpler/clear to me as a separate line.

fatkodima:
Added `AllowModifier` option and test with comment.

PR #8395, opened 2020-07-24
Use `Cop::Base` API for `Lint` department

koic:
Follow #7868.

This PR uses `Cop::Base` API for `Lint` department's cops.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/ambiguous_regexp_literal.rb
+++ b/lib/rubocop/cop/lint/ambiguous_regexp_literal.rb
@@ -21,20 +21,19 @@ module RuboCop
       #
       #   # With parentheses, there's no ambiguity.
       #   do_something(/pattern/i)
-      class AmbiguousRegexpLiteral < Cop
+      class AmbiguousRegexpLiteral < Base
         include ParserDiagnostic
+        extend AutoCorrector
 
         MSG = 'Ambiguous regexp literal. Parenthesize the method arguments ' \
               "if it's surely a regexp literal, or add a whitespace to the " \
               'right of the `/` if it should be a division.'
 
-        def autocorrect(node)
-          lambda do |corrector|
-            add_parentheses(node, corrector)
-          end
-        end
-
         private
+
+        def autocorrect(corrector, node)
+          add_parentheses(node, corrector)
+        end
 
         def relevant_diagnostic?(diagnostic)
           diagnostic.reason == :ambiguous_literal
--- a/lib/rubocop/cop/lint/non_deterministic_require_order.rb
+++ b/lib/rubocop/cop/lint/non_deterministic_require_order.rb
@@ -51,7 +51,9 @@ module RuboCop
       #   # good
       #   Dir.glob(Rails.root.join('test', '*.rb')).sort.each(&method(:require))
       #
-      class NonDeterministicRequireOrder < Cop
+      class NonDeterministicRequireOrder < Base
+        extend AutoCorrector
+
         MSG = 'Sort files before requiring them.'
 
         def on_block(node)
@@ -61,7 +63,9 @@ module RuboCop
           loop_variable(node.arguments) do |var_name|
             return unless var_is_required?(node.body, var_name)
 
-            add_offense(node.send_node)
+            add_offense(node.send_node) do |corrector|
+              correct_block(corrector, node.send_node)
+            end
           end
         end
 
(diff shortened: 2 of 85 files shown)


lib/rubocop/cop/lint/ambiguous_regexp_literal.rb
```diff
@@ -21,21 +21,20 @@ module Lint
...
 
-        def autocorrect(node)
-          lambda do |corrector|
-            add_parentheses(node, corrector)
-          end
-        end
-
         private
 
+        def autocorrect(corrector, node)
```
marcandre, line 34: I don't understand how this is supposed to work tbh... Who is calling `autocorrect`?

lib/rubocop/cop/lint/ambiguous_regexp_literal.rb
```diff
@@ -21,21 +21,20 @@ module Lint
       #
       #   # With parentheses, there's no ambiguity.
       #   do_something(/pattern/i)
-      class AmbiguousRegexpLiteral < Cop
+      class AmbiguousRegexpLiteral < Base
         include ParserDiagnostic
```
marcandre, line 25: My 2 cents is that this module doesn't do enough to justify its existence and makes things more complicated than if it didn't exist.

marcandre:
Wow, impressive PR 💪 🙇‍♂️
I'd love to review more but won't be able till next monday.

lib/rubocop/cop/lint/non_deterministic_require_order.rb
```diff
@@ -61,45 +63,43 @@ def on_block(node)
...
           end
         end
 
         def on_block_pass(node)
           return unless method_require?(node)
           return unless unsorted_dir_pass?(node.parent)
 
-          add_offense(node.parent)
+          add_offense(node.parent) do |corrector|
+            autocorrect(corrector, node.parent)
```
marcandre, line 77: One of the goal of `Base` is to avoid calling a single entry point like `autocorrect` here and redoing an `if` there. Can't we call `correct_block_pass` directly here?

koic: This did not have to be a single entry point. I updated it.

lib/rubocop/cop/lint/script_permission.rb
```diff
@@ -30,29 +30,32 @@ module Lint
...
-          lambda do |_corrector|
-            FileUtils.chmod('+x', node.loc.expression.source_buffer.name)
+          add_offense(comment, message: message) do
+            autcocorrect(comment)
           end
         end
 
         private
 
+        def autcocorrect(comment)
```
marcandre, line 55: Oh! Interesting! 😅
This is a problem, because now the block is called even if autocorrection won't happen...
So this would always do the `chmod`.
I had not considered this case 😆, and there's no test for that either I think.
If you add `if autocorrect_requested?` you should be pretty much ok.

lib/rubocop/cop/lint/uri_regexp.rb
```diff
@@ -43,31 +45,31 @@ def on_send(node)
...
-              "#{top_level}URI::DEFAULT_PARSER.make_regexp#{argument}"
-            )
+          add_offense(node.loc.selector, message: format) do |corrector|
+            autocorrect(corrector, node)
           end
         end
 
-        private
+        def autocorrect(corrector, node)
+          if (captured_values = uri_regexp_with_argument?(node))
```
marcandre, line 59: I'm surprised we don't have a cop to change this to 
```
captured_values = uri_regexp_with_argument?(node) ||
                  uri_regexp_without_argument?(node)
```

lib/rubocop/cop/lint/script_permission.rb
koic: > (Also, there's a typo)

Oops! I missed this typo 💦 

> This is a problem, because now the block is called even if autocorrection won't happen...

I did print debugging. The block execution seems to be running only when offensed. Is `if autocorrect_requested?` required?

lib/rubocop/cop/lint/uri_regexp.rb
koic: Sure! I've also improved other redundant logic in uri_regexp.rb. Thank you for your review ❤️

koic:
> I found just one extra thing in addition to my comment about `ParserDiagnostic`

Yeah, I would like to refactor it in another PR due to this update changes the module structure.

lib/rubocop/cop/lint/script_permission.rb
marcandre: I added a spec in #8441
I think that if you rebase now, you will get a failure. Sorry, the `expect_offense` method need improvement...

koic: The build failed as expected with #8441. I added `if autocorrect_requested?` and the build has been passed. Thank you for adding the reproduction test.

marcandre: Great, thanks

marcandre:
Cool!

Great work 💪

> > I found just one extra thing in addition to my comment about `ParserDiagnostic`
> 
> Yeah, I would like to refactor it in another PR due to this update changes the module structure.

Looking forward to it, thanks!

koic:
I've opened #8442 :-)

Issue #7755, opened 2020-02-25
Cop idea: Out of range referencing for Regexp


pocke:
## Is your feature request related to a problem? Please describe.

When a reference for Regepx is out of range, it always returns nil.

```ruby
/(foo)bar/ =~ 'foobar'
# ok
p $1 # => foo
# bad - $2 always returns nil.
p $2 # => nil
```

## Describe the solution you'd like

Add a cop to detect the problem.

## Describe alternatives you've considered

Nothing

## Additional context

We will be able to the cop with regexp_parser gem. ref: #7746

PR #8407, opened 2020-07-27
Add new Lint/OutOfRangeRefInRegexp cop #7755

sonalinavlakhe:
This cop looks for out of range referencing for Regexp, as while capturing groups out of range reference always returns nil.

      /(foo)bar/ =~ 'foobar'

      #   bad - always returns nil
      #   puts $2 # => nil

      #   good
      #   puts $1 # => foo
-----------------
Issue: [https://github.com/rubocop-hq/rubocop/issues/7755](https://github.com/rubocop-hq/rubocop/issues/7755)

* [x] Added tests.


lib/rubocop/cop/lint/out_of_range_ref_in_regexp.rb
```diff
@@ -0,0 +1,71 @@
...
+      #   puts $1 # => foo
+      #
+      class OutOfRangeRefInRegexp < Cop
+        MSG = 'Do not use out of range reference for the Regexp.'
+
+        def investigate(processed_source)
+          ast = processed_source.ast
+          valid_ref = cop_config['Count']
+          ast.each_node do |node|
+            if node.regexp_type?
```
marcandre, line 25: Could you use `on_regexp` and `on_nth_ref`? It's the axiomatic way to do it and is more performant.

lib/rubocop/cop/lint/out_of_range_ref_in_regexp.rb
```diff
@@ -0,0 +1,71 @@
...
+      # @example
+      #   /(foo)bar/ =~ 'foobar'\
+
+      #   # bad - always returns nil
+      #   puts $2 # => nil
+
+      #   # good
+      #   puts $1 # => foo
+      #
+      class OutOfRangeRefInRegexp < Cop
```
marcandre, line 18: Please inherit from `Cop::Base`, see https://docs.rubocop.org/rubocop/v1_upgrade_notes.html

spec/rubocop/cop/lint/out_of_range_ref_in_regexp_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+  # RuboCop does not know a value of variables that it will contain in the regexp literal.
+  # For example, `/(?<foo>#{var}*)` is interpreted as `/(?<foo>*)`.
+  # So it does not offense when variables are used in regexp literals.
+  context 'when containing a non-regexp literal' do
+    it 'does not register an offence when containing a lvar' do
+      expect_no_offenses(<<~'RUBY')
+        var = '(\d+)'
+        /(?<foo>#{var}*)/
+      RUBY
```
marcandre, line 78: I think your test (and following) should use `$4` or similar, otherwise they are not testing anything...

lib/rubocop/cop/lint/out_of_range_ref_in_regexp.rb
```diff
@@ -0,0 +1,71 @@
...
+          nil
+        end
+
+        def regexp_captures(tree)
+          named_capture = numbered_capture = 0
+          tree.each_expression do |e|
+            named_capture += 1 if e.instance_of?(Regexp::Expression::Group::Named)
+            numbered_capture += 1 if e.instance_of?(Regexp::Expression::Group::Capture)
+          end
+          return named_capture if numbered_capture.zero?
```
marcandre, line 62: I'm not sure this is correct. You should have a test with 1 named capture and 2 numbered captures, for example...

lib/rubocop/cop/lint/out_of_range_ref_in_regexp.rb
```diff
@@ -0,0 +1,71 @@
...
+      #
+      class OutOfRangeRefInRegexp < Cop
+        MSG = 'Do not use out of range reference for the Regexp.'
+
+        def investigate(processed_source)
+          ast = processed_source.ast
+          valid_ref = cop_config['Count']
+          ast.each_node do |node|
+            if node.regexp_type?
+              break if contain_non_literal?(node)
```
marcandre, line 26: I think you meant `next` instead of `break`. A test like the following would fail, right?
```
def foo
  /#{dynamic}/ =~ foo
end

def bar
  /(not)(dynamic)/ =~ bar
  $3 # << should raise error
end
```

Extracting this to a method might help make this more readable.

lib/rubocop/cop/lint/out_of_range_ref_in_regexp.rb
sonalinavlakhe: @marcandre, the example you have shared is mixed captures Regex, and in that case, we have to only consider named capture count as numbered captures will be ignored. so in the end, I am returning named capture count.

Just FYI, I have implemented the spec example for mix names and numbered captures.

marcandre: Oh, right, I got confused by the algorithm. I think `named_capture.non_zero? || numbered_capture` would be easier, or `named_capture > 0 ? named_capture : numbered_capture` if you prefer would be much easier to read and equivalent, right?

docs/modules/ROOT/pages/cops_lint.adoc
```diff
@@ -2187,6 +2187,21 @@ p [''.frozen?, ''.encoding] #=> [true, #<Encoding:US-ASCII>]
 p [''.frozen?, ''.encoding] #=> [true, #<Encoding:US-ASCII>]
 ----
 
+== Lint/OutOfRangeRefInRegexp
+
+|===
+| Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
+
+| Pending
+| Yes
```
marcandre, line 2196: Need to mark as unsafe, as there will be false negatives, e.g.:
```ruby
/(a)(b)/ =~ str
/(a)/ =~ str if foo
return $2 unless foo # will raise offense even though code is correct, right?
```

lib/rubocop/cop/lint/out_of_range_ref_in_regexp.rb
```diff
@@ -0,0 +1,69 @@
...
+        def on_nth_ref(node)
+          backref, = *node
+          return if @valid_ref.nil?
+
+          add_offense(node) if backref > @valid_ref
+        end
+
+        private
+
+        def contain_non_literal?(node)
```
marcandre, line 45: This should be simplified. Is there any other case than `children.size == 1` and `children.first.str_type?` that we handle?

sonalinavlakhe: @marcandre , Thank you for giving it a look and guiding me :) 
 I am not understanding this comment can you elaborate more with an example.
Just FYI, the reference for this one is -  https://github.com/rubocop-hq/rubocop/pull/8074

marcandre: If you replace the `def` with
```ruby
def contain_non_literal?(node)
  children.size == 1 && children.first.str_type?
end
```

Does this fail for some examples?

sonalinavlakhe: I think this will work but we have to check with the condition - node.children.size != 2 
```
def contain_non_literal?(node)
   node.children.size != 2 || !node.children.first.str_type?
end
```

because regex node will always have the last child of node_type  regopt

marcandre: Oh, right, indeed 👍

docs/modules/ROOT/pages/installation.adoc
```diff
@@ -1,4 +1,4 @@
-= Installation
+  = Installation
```
marcandre, line 1: You probably didn't mean to change this

config/default.yml
```diff
@@ -1620,6 +1620,12 @@ Lint/OrderedMagicComments:
   Enabled: true
   VersionAdded: '0.53'
 
+Lint/OutOfRangeRefInRegexp:
+  Description: 'Checks for out of range reference for Regep because it always returns nil.'
```
marcandre, line 1624: Regep => Regexp

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -0,0 +1,60 @@
...
+      #   puts $2 # => nil
+
+      #   # good
+      #   puts $1 # => foo
+      #
+      class OutOfRangeRegexpRef < Base
+        MSG = 'Do not use out of range reference for the Regexp.'
+
+        def on_regexp(node)
+          @valid_ref = cop_config['Count']
```
marcandre, line 22: I'm not sure I understand `cop_config['Count']`, maybe you simply meant `nil`?

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        def contain_non_literal?(node)
+          node.children.size != 2 || !node.children.first.str_type?
+        end
+
+        def regexp_captures(tree)
+          named_capture = numbered_capture = 0
+          tree.each_expression do |e|
+            named_capture += 1 if e.instance_of?(Regexp::Expression::Group::Named)
+            numbered_capture += 1 if e.instance_of?(Regexp::Expression::Group::Capture)
```
marcandre, line 53: Is it possible to rely on `e.type` instead of the class?

sonalinavlakhe: yeah, we can do it but we have to check with the token instead.

```
tree.each_expression do |e|
  named_capture += 1 if e.token == :named
  numbered_capture += 1 if e.token == :capture
end
````

**Comment by sonalinavlakhe** [comment]
Never mind @marcandre, for me, it's a huge learning opportunity 😄 

What I understood from the below comment 
" you should add a test if $4 is encountered before any regexp. You'll need to define on_new_investigation too" 
that you must be referring below example
 ```
$4
 /(foo)(bar)/ =~ "foobar"
```
 Do I need to initialize @valid_ref to nil in the on_new_investigation method? if so then  I am not able to understand the use case for the on_new_investigation method because we are not creating any offense if we got the reference before any regexp.
Please suggest.

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
marcandre: Neither seem great to me, but that might be the gem's fault. Could you use `e.capturing?`, and then `e.respond_to?(:name)`, say?

marcandre:
> Do I need to initialize @valid_ref to nil in the on_new_investigation method?

Yes

> if so then I am not able to understand the use case for the on_new_investigation method because we are not creating any offense if we got the reference before any regexp.

First, shouldn't we create an offense then?
Second: in warning mode you would get a 'uninitialized instance variable' I think. Also running the cop on a second source would still hold the result form the previous one.

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
sonalinavlakhe: I can check with `e.capturing?` but again we cant check further for `e.respond_to?(:name)`.

sonalinavlakhe:
> First, shouldn't we create an offense then?

Please suggest we should consider this case to raise offense.

> Second:` in warning mode you would get a 'uninitialized instance variable' I think. Also running the cop on a second source would still hold the result form the previous one.

I got this one. Thanks 👍

marcandre:
> Please suggest we should consider this case to raise offense.

Sorry, I don't understand... If I'm not mistaken, `def foo; $4; end` should raise an offense, in that it can not be non nil, so maybe we should initialize `@valid_ref` to 0, no?

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
marcandre: Sorry, why?

sonalinavlakhe:
> Sorry, I don't understand... If I'm not mistaken, def foo; $4; end should raise an offense, in that it can not be non nil, so maybe we should initialize @valid_ref to 0, no?

Yes, you are right. only for regular expression with non-literals, I have to make @valid_ref nil, so will have a check to not raise an offense.

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
sonalinavlakhe: @marcandre, `e.cpaturing?` will raise an error 
`undefined method  capturing?  for #<Regexp::Expression::CharacterType::Word:0x00007fb466af9cf0>`
 for all non-group type expressions so better we to use `e.type?(:group)` then  `e.respond_to?(:name)`

marcandre: `e.type?(:group) && e.respond_to?(:name)` 👍, I think it's much better than relying on the class, even if longer...

spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+    end
+
+    it 'does not register an offence when containing a constant' do
+      expect_no_offenses(<<~'RUBY')
+        CONST = "12"
+        /(?<foo>#{CONST}*)/ =~ "12"
+        puts $1
+        puts $3
+      RUBY
+    end
```
marcandre, line 128: All these tests can be removed, we only need to test one "dynamic" case.

spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+      /(?<foo>FOO)(BAR)/ =~ "FOOBAR"
+      puts $1
+    RUBY
+  end
+
+  # See https://github.com/rubocop-hq/rubocop/issues/8083
+  it 'does not register offense when using a Regexp cannot be processed by regexp_parser gem' do
+    expect_no_offenses(<<~'RUBY')
+      /data = ({"words":.+}}}[^}]*})/m
+    RUBY
```
marcandre, line 67: Might as well add `$3` or something. Hopefully the Regexp parser will handle this one day and that test will fail, but right now it would always pass.

spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
sonalinavlakhe: hi @marcandre but actually these specs are self-explanatory and devs who are reading these will get to understand the code better is what I think, please give an example of how this can be improved better with `dynamic` test case.

marcandre: What I meant to say is that the previous test (with a variable) is sufficient; a good implementation (like we have now) can not pass the variable test and fail the instance variable, the method call, etc... or vice versa. They all test the same idea:is the regexp dynamic or not. It's not about what is inside the `#{...}`, just that there is a `#{...}` or not. Otherwise to be "complete", we'd have to add all possibilities, for example method call with block, or with rescue, or literal, or ...

sonalinavlakhe:
@marcandre, Please review the updated changes. let me know all good or we can still improve 😄

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # This cops looks for out of range referencing for Regexp, as while capturing groups out of
+      # out of range reference always returns nil.
+
```
marcandre, line 8: I just noticed that the doc file is incomplete, because you have a missing `#` here

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -0,0 +1,65 @@
...
+      # @example
+      #   /(foo)bar/ =~ 'foobar'
+
```
marcandre, line 11: and here...

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -0,0 +1,65 @@
...
+      #   # bad - always returns nil
+      #   puts $2 # => nil
+
```
marcandre, line 14: and here. Might be a cop idea! 😆

sonalinavlakhe: Cool, hope I fixed everything now 🤣 🤞

spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
marcandre: There's still no reason to raise an offense here once parsing is fixed, right?

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # This cops looks for out of range referencing for Regexp, as while capturing groups out of
```
marcandre, line 6: I'm not a native speaker, but the "as while" sounds strange to me. How about?

"This cops looks for references of Regexp captures that are out of range and thus always returns nil."

spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
sonalinavlakhe: I have added $1 in last commit to fix this. please refer [this](https://github.com/rubocop-hq/rubocop/pull/8407/commits/f97bd95cdde545bad681cdf63c8e19fc5c5fa1f9#diff-e36629e056bdba9159de0a2347dc3807R74)

sonalinavlakhe:
> There's still no reason to raise an offense here once parsing is fixed, right? That's why I wrote to use $3 or something.

I have added $1 in the last commit to fix this. please refer [this](https://github.com/rubocop-hq/rubocop/pull/8407/commits/f97bd95cdde545bad681cdf63c8e19fc5c5fa1f9#diff-e36629e056bdba9159de0a2347dc3807R74)

marcandre:
> > There's still no reason to raise an offense here once parsing is fixed, right? That's why I wrote to use $3 or something.
> 
> I have added $1 in the last commit to fix this.

Right, but the "unparseable" regexp has one captured group, so `$1` is not out-of-range...

sonalinavlakhe:
> Right, but the "unparseable" regexp has one captured group, so $1 is not out-of-range...

Sorry, I missed that one, when I tried to use other than $1 i.e $2 or $3 this test is failing. I checked more on this and conclude that regex_parser now able to parse this type of expression. 
I checked with a series of examples  - [https://github.com/rubocop-hq/rubocop/issues/8083](https://github.com/rubocop-hq/rubocop/issues/8083) and [https://github.com/ammar/regexp_parser/issues/15](https://github.com/ammar/regexp_parser/issues/15). regex_parser parsing those without throwing an exception

```
irb(main):001:0> require 'regexp_parser'
=> true
irb(main):002:0> Regexp::Parser.parse('data = ({"words":.+}}}[^}]*})')
=> #<Regexp::Expression::Root:0x00007fa07a9c5c08 @type=:expression, @token=:root, @text="", @ts=0, @level=nil, @set_level=nil, @conditional_level=nil, @nesting_level=0, @quantifier=nil, @options={}, @expressions=[#<Regexp::Expression::Literal:0x00007fa07b857bb8 @type=:literal, @token=:literal, @text="data = ", @ts=0, @level=0, @set_level=0, @conditional_level=0, @nesting_level=1, @quantifier=nil, @options={}>, #<Regexp::Expression::Group::Capture:0x00007fa07b857b90 @type=:group, @token=:capture, @text="(", @ts=7, @level=0, @set_level=0, @conditional_level=0, @nesting_level=1, @quantifier=nil, @options={}, @expressions=[#<Regexp::Expression::Literal:0x00007fa07b8579d8 @type=:literal, @token=:literal, @text="{\"words\":", @ts=8, @level=1, @set_level=0, @conditional_level=0, @nesting_level=2, @quantifier=nil, @options={}>, #<Regexp::Expression::CharacterType::Any:0x00007fa07b8579b0 @type=:meta, @token=:dot, @text=".", @ts=17, @level=1, @set_level=0, @conditional_level=0, @nesting_level=2, @quantifier=#<Regexp::Expression::Quantifier:0x00007fa07b857938 @token=:one_or_more, @text="+", @mode=:greedy, @min=1, @max=-1>, @options={}>, #<Regexp::Expression::Literal:0x00007fa07b8578e8 @type=:literal, @token=:literal, @text="}}}", @ts=19, @level=1, @set_level=0, @conditional_level=0, @nesting_level=2, @quantifier=nil, @options={}>, #<Regexp::Expression::CharacterSet:0x00007fa07b8578c0 @negative=true, @closed=true, @type=:set, @token=:character, @text="[", @ts=22, @level=1, @set_level=0, @conditional_level=0, @nesting_level=2, @quantifier=#<Regexp::Expression::Quantifier:0x00007fa07b857640 @token=:zero_or_more, @text="*", @mode=:greedy, @min=0, @max=-1>, @options={}, @expressions=[#<Regexp::Expression::Literal:0x00007fa07b8577d0 @type=:literal, @token=:literal, @text="}", @ts=24, @level=1, @set_level=1, @conditional_level=0, @nesting_level=3, @quantifier=nil, @options={}>]>, #<Regexp::Expression::Literal:0x00007fa07b8575f0 @type=:literal, @token=:literal, @text="}", @ts=27, @level=1, @set_level=0, @conditional_level=0, @nesting_level=2, @quantifier=nil, @options={}>], @number=1, @number_at_level=1>]> 
```
Please suggest should we remove the rescue block then? 
```
rescue Regexp::Scanner::ScannerError
  return
end

```

as Errors like this ([https://github.com/ammar/regexp_parser/blob/master/spec/scanner/errors_spec.rb)]() are handled by Lint/Syntax cop.
#app/controllers/accounts_controller.rb
```
/\p{foobar}/
$1
```
```
Inspecting 1 file
E

Offenses:

app/controllers/accounts_controller.rb:7:4: E: Lint/Syntax: invalid character property name {foobar}: /\p{foobar}/
(Using Ruby 2.5 parser; configure using TargetRubyVersion parameter, under AllCops)
   /\p{foobar}/
   ^^^^^^^^^^^^

```

marcandre:
That's good news :-)

Right, I see in their Changelog the issue has been changed in v1.7.1, so we could bump our gemspec requirement to that version and remove the `rescue` (and the test). If ever there's such an exception, it won't be the end of the world (just a notice there was an issue when processing the file) and a bug report can be filed with the gem.

It also means that the other `rescue` in the code could be removed and that the associated specs are incorrect (as they should fail with the regexp parser 1.7.1)

sonalinavlakhe:
> That's good news :-)
> 
> Right, I see in their Changelog the issue has been changed in v1.7.1, so we could bump our gemspec requirement to that version and remove the `rescue` (and the test). If ever there's such an exception, it won't be the end of the world (just a notice there was an issue when processing the file) and a bug report can be filed with the gem.
> 
> It also means that the other `rescue` in the code could be removed and that the associated specs are incorrect (as they should fail with the regexp parser 1.7.1)

@marcandre  Well, I got this ok I have to remove the `rescue` block from code and the test from the out_of_range_regexp_ref_spec.rb. 
Any other action item needed to fix this?

marcandre:
> Any other action item needed to fix this?

No, rest can be done separately.

You'll have to rebase to fix the Changelog conflict though

sonalinavlakhe:
@marcandre, I have done rebase and all the changes.

marcandre:
🎉 Thank you @sonalinavlakhe for this PR and your quick turn around to my multiple review comments 👍

Congratulations on your first Cop 🍾 😄

PR #8411, opened 2020-07-28
Add new `Layout/EmptyLineAfterMultilineCondition` cop

### LINKED ISSUE #8384: New cop: Empty line after conditions spanning multiple lines
Inspired by https://github.com/airbnb/ruby#multiline-if-newline

It is better to differentiate between conditions and body in this case
```ruby
if @reservation_alteration.checkin == @reservation.start_date &&
   @reservation_alteration.checkout == (@reservation.start_date + @reservation.nights)

  redirect_to_alteration @reservation_alteration
end
```

rather than in this
```ruby
if @reservation_alteration.checkin == @reservation.start_date &&
   @reservation_alteration.checkout == (@reservation.start_date + @reservation.nights)
  redirect_to_alteration @reservation_alteration
end
```

This can be extended to other syntax constructs like `unless`, `while`, etc.

fatkodima:
Closes #8384 

This cop will also benefit from https://github.com/rubocop-hq/rubocop/pull/8404#pullrequestreview-455385942  (first paragraph).
@marcandre Friendly ping - have you started work on that? 😃


koic:
To be honest, I've never encountered this style. I think this cop should provide an alternative enforced style option and the default enforced style shouldn't insert a blank line.

fatkodima:
Made this cop disabled.

bbatsov:
I'd add a note to AirBnB's guide as a reference and mention this is disable, because that style is not very common in practice.

fatkodima:
Added reference and note.

bbatsov:
Thanks!

PR #8412, opened 2020-07-29
Add new `Style/OptionalBooleanParameter` cop

fatkodima:
https://rubystyle.guide/#boolean-keyword-arguments

```ruby
# bad
def some_method(bar = false)
  puts bar
end

# good
def some_method(bar: false)
  puts bar
end
```


config/default.yml
```diff
@@ -2545,6 +2545,12 @@ Style/BlockDelimiters:
   # the `line_count_based` style.
   BracesRequiredMethods: []
 
+Style/BooleanKeywordArgument:
+  Description: 'Use keyword arguments when defining method with boolean argument.'
+  StyleGuide: '#boolean-keyword-arguments'
+  Enabled: pending
```
koic, line 2551: I think that this cop is unsafe, just like `Style/OptionalArguments` cop.

marcandre: I'm curious why `Style/OptionalArguments` is "unsafe", or this one?

For this one, my only concern is if there is a call to `super`, or `super(that_argument)`, then maybe there should be no offense since this method is just copying the interface it inherits. Maybe there should not be an offense then; the base method will be flagged probably. What other cases are unsafe?

lib/rubocop/cop/registry.rb
```diff
@@ -100,7 +100,7 @@ def contains_cop_matching?(names)
       # @note Emits a warning if the provided name has an incorrect namespace
       #
       # @return [String] Qualified cop name
-      def qualified_cop_name(name, path, shall_warn = true)
+      def qualified_cop_name(name, path, shall_warn: true)
```
marcandre, line 103: I like this, but maybe call it simply `warn`?

lib/rubocop/cop/style/boolean_keyword_argument.rb
```diff
@@ -0,0 +1,42 @@
...
+      class BooleanKeywordArgument < Base
+        MSG = 'Use keyword arguments when defining method with boolean argument.'
+        BOOLEAN_VALUES = %w[true false].freeze
+
+        def on_def(node)
+          node.arguments.each do |arg|
+            next unless arg.optarg_type?
+
+            _name, value = *arg
+            add_offense(arg) if BOOLEAN_VALUES.include?(value.source)
```
marcandre, line 35: I'd use the `type` instead of the `source` to check; source must allocate and copy a string...

config/default.yml
koic: I think it means a breaking change to users if the API signature is a published interface.
https://martinfowler.com/bliki/PublishedInterface.html

fatkodima: Made it unsafe.

bbatsov:
Seems we've made a mistake with the heading in the style guide - I think the cop should be named `OptionalBooleanParameter".

fatkodima:
Updated with new name.

bbatsov:
Great cop!

PR #8417, opened 2020-07-30
Add new `Style/GlobalStdStream` cop

fatkodima:
This cop enforces the use of `$stdout/$stderr/$stdin` instead of `STDOUT/STDERR/STDIN`.

https://rubystyle.guide/#global-stdout

```ruby
# bad
STDOUT.puts('hello')

hash = { out: STDOUT, key: value }

def m(out = STDOUT)
  out.puts('hello')
end

# good
$stdout.puts('hello')

hash = { out: $stdout, key: value }

def m(out = $stdout)
  out.puts('hello')
end
```


bbatsov:
Great work! :bowing_man:

PR #8430, opened 2020-07-31
Add new `Lint/UnreachableLoop` cop

fatkodima:
Inspired by https://eslint.org/docs/rules/no-unreachable-loop and by https://github.com/rubocop-hq/rubocop/blob/10cedb8eeedb49464ea9c72666f3252eefb670d5/lib/rubocop/cop/lint/safe_navigation_chain.rb#L54-L61

This cop checks for loops that will have at most one iteration.

A loop that can never reach the second iteration is a possible error in the code.
In rare cases where only one iteration (or at most one iteration) is intended behavior, the code should be refactored to use `if` conditionals.

```ruby
# bad
while node
  do_something(node)
  node = node.parent
  break
end

# good
while node
  do_something(node)
  node = node.parent
end

# bad
def verify_list(head)
  item = head
  begin
    if verify(item)
      return true
    else
      return false
    end
  end while(item)
end

# good
def verify_list(head)
  item = head
  begin
    if verify(item)
      item = item.next
    else
      return false
    end
  end while(item)

  true
end

# bad
def find_something(items)
  items.each do |item|
    if something?(item)
      return item
    else
      raise NotFoundError
    end
  end
end

# good
def find_something(items)
  items.each do |item|
    if something?(item)
      return item
    end
  end
  raise NotFoundError
end
```


bbatsov:
Fantastic work! Дякую! :bowing_man:

fatkodima:
Будь ласка :) Радий зробити свій внесок.

marcandre:
Good stuff! 🙇‍♂️ That `FIX ME`: 🤣

Unless I'm mistaken, the following case would raise an offense when it shouldn't, right?

```ruby
while something do
  raise 'oops' rescue nil
end
```

Probably best to exclude cases with a `rescue`. Even if you do, this cop is unsafe. But if you do, this cop's autocorrection would actually be safe, something we don't handle well (cop with known false positives but no known false negatives).

fatkodima:
> Unless I'm mistaken, the following case would raise an offense when it shouldn't, right?

No, this won't raise an offense. 

> Probably best to exclude cases with a rescue or catch

Yes, those are actually excluded (not checked) https://github.com/rubocop-hq/rubocop/blob/e1c057d898cadfb2fcf061cd9dd7546be71bb15b/lib/rubocop/cop/lint/unreachable_loop.rb#L127-L141

marcandre:
> No, this won't raise an offense.

Oh, good, my mistake 👍

PR #8432, opened 2020-08-01
Add new `Lint/FloatComparison` cop

fatkodima:
This is a known fact that comparing floats is bad, so this cop checks for the presence of precise comparison of floating point numbers.

```ruby
a = (0.3 * 3) + 0.1 # 0.9999999999999999
b = 1.0

a == b # false 
```

```ruby
# bad
x == 0.1
x != 0.1

# good
(x - 0.1).abs < Float::EPSILON

# good
tolerance = 0.0001
(x - 0.1).abs < tolerance
```

Reference: https://www.jetbrains.com/help/resharper/CompareOfFloatsByEqualityOperator.html


docs/modules/ROOT/pages/cops_lint.adoc
```diff
@@ -1229,6 +1229,44 @@ end
...
+=== Examples
+
+[source,ruby]
+----
+# bad
+x == 0.1
+x != 0.1
+
+# good
+(x - 0.1).abs < Float::EPSILON
```
bbatsov, line 1260: There's also BigDecimal that we can recommend. It seems to me that using it makes the code much simpler to understand.

bbatsov:
Great cop idea! Can you also submit a PR for this to the Ruby style guide?

lib/rubocop/cop/lint/float_comparison.rb
```diff
@@ -0,0 +1,91 @@
...
+      #
+      #   # good
+      #   tolerance = 0.0001
+      #   (x - 0.1).abs < tolerance
+      #
+      #   # Or some other epsilon based type of comparison:
+      #   # https://www.embeddeduse.com/2019/08/26/qt-compare-two-floats/
+      #
+      class FloatComparison < Base
+        MSG = 'Do not use precise comparison of floats. '\
```
bbatsov, line 29: I think we can simplify the message to something like "Avoid (in)equality comparisons of floats as they are unreliable." or something along those lines.

lib/rubocop/cop/lint/float_comparison.rb
```diff
@@ -0,0 +1,91 @@
...
+              'Use some sort of epsilon based comparison.'
+
+        EQUALITY_METHODS = %i[== != eql?].freeze
```
marcandre, line 32: I would add `equal?` (which is worse than these tbh as it is implementation dependent)

fatkodima: From [documentation](https://ruby-doc.org/core-2.7.1/BasicObject.html#method-i-equal-3F)

> Unlike #==, the equal? method should never be overridden by subclasses as it is used to determine object identity (that is, a.equal?(b) if and only if a is the same object as b)

So seems like this is not actually needed? As it does not compare values...

marcandre: An optimization on 64 bit systems stores most floats (but not all) are immediate values and thus `equal?`:

```ruby
x=4.2
(x*1).equal?(x) # => true on 64 bit platforms, false of 32 platforms (and JRuby / TruffleRuby?)
```

fatkodima: Can you, please, explain a bit, why:
```ruby
-0.0.equal?(-0.0) # => false
0.0.equal?(-0.0) # => false
0.0 == -0.0 # => true
```
?

marcandre: The `equal?` thing is technical...
`x.equal? y` is same as `x.object_id == y.object_id`.
Now Ruby store objects in 32 / 64 bits (depending on platform) and uses the first bits to say what's in the object, either a pointer to memory (most cases) or an integer value (if it fits in the remaining bits), or a float (if it fits too), or special values `nil`, `false`, `true`, symbols. That's kind of what `object_id` returns...
So `(1 << 20).object_id` will always be the same because it fits, and `(1 << 70).object_id` always it doesn't fit and requires a pointer to allocate memory.
It so happens that `-0.0` does not fit and requires a memory allocation, but `0.0` fits (on 64-bit platform).

`-0.0` is quite odd. `0.0 == -0.0 # => true` is [mandated by the IEEE standard](https://en.wikipedia.org/wiki/Signed_zero#Comparisons) but I don't understand why. It's not even clear [what should `-0.0.negative?` return?](https://bugs.ruby-lang.org/issues/17098)

fatkodima: That part about how ruby stores objects is quite well described in "Ruby under microscope" book. Thanks for refresher!

fatkodima:
Updated with all suggestions.

bbatsov:
Thanks!

dvandersluis:
Should this flag comparisons to 0 or 0.0? I don't believe there are precision issues with zero (`Float#zero?` is true for 0.0).

PR #8433, opened 2020-08-01
Add new `Lint/BinaryOperatorIdenticalOperands` cop

fatkodima:
Inspired by https://www.viva64.com/en/w/v501/

This cop supersedes (with other binary operators) and deprecates `Lint/UselessComparison`. That existent cop should already be marked as unsafe due to reasons - https://github.com/rubocop-hq/rubocop/compare/master...fatkodima:binary_operator_identical_operands?expand=1#diff-e9973b98907964eb1cc98fd20b55fe1cR8-R10

So I marked this cop as unsafe.


lib/rubocop/cop/lint/binary_operator_identical_operands.rb
```diff
@@ -0,0 +1,43 @@
...
+      #   if a.x != 0 && a.x != 0
+      #     do_something
+      #   end
+      #
+      #   def childs?
+      #     left_child || left_child
+      #   end
+      #
+      class BinaryOperatorIdenticalOperands < Base
+        MSG = 'Binary operator has identical operands.'
```
bbatsov, line 25: I'd mention in the message the name of the binary operator (and maybe even the operand in question).

lib/rubocop/cop/lint/binary_operator_identical_operands.rb
```diff
@@ -0,0 +1,43 @@
...
+      #
+      #   if a.x != 0 && a.x != 0
+      #     do_something
+      #   end
+      #
+      #   def childs?
+      #     left_child || left_child
+      #   end
+      #
+      class BinaryOperatorIdenticalOperands < Base
```
bbatsov, line 24: Let's name this BinaryOperatorWithIdenticalOperands.

lib/rubocop/cop/lint/binary_operator_identical_operands.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # This cop checks for places where binary operator has identical operands.
```
bbatsov, line 6: MIght be a good idea to mention here all the operands this covers.

lib/rubocop/cop/lint/binary_operator_identical_operands.rb
fatkodima: Added mention about operator. I think, mentioning the operand is not necessary, as it will be obvious from code. Example: `foo == foo`.

bbatsov:
Thanks!

PR #8440, opened 2020-08-03
Tweak recommendation for cops' instance variables [doc]

marcandre:
How does this sound @pirj ?


pirj:
It does, yes, but I use my both hands to vote for the removal of this cop reuse feature completely 🙌 
😆

Issue #8278, opened 2020-07-08
Should we mark some modules as private before v1?


marcandre:
When reviewing #8274 I realized that it might be a good idea to mark some internal modules and classes as "@api private" before v1. Or is that too late or futile?

deivid-rodriguez:
I believe it's good idea. If a whole review of all modules and classes is deemed like it would delay the release too much, at least mark those modules where privateness is straightforward. It might flexibilize future refactorings :)

marcandre:
@bbatsov should I work into this?

bbatsov:
Yeah, that's a good idea.

pirj:
Makes total sense: if not done otherwise, we'll have to refrain from refactoring and keep all the internals intact to avoid breakages in custom extensions using this internals directly. At least until we hit 2.0, in ~year from now.

I was thinking of adding a suffix to all methods and running well-known extensions against this. All the breakages will remain in the public API, all the rest can be made `private` and/or marked as such.

Do you want me to tackle this?

pirj:
Actually, @jonatas, it seems to be a great way to do with `ffast`. Unfortunately, I don't have much experience with using it. Would you give it a shot?

marcandre:
I pushed my branch, I made some progress but didn't quite finish...

marcandre:
> I was thinking of adding a suffix to all methods

I'm 👎  on that. It makes the codebase ugly. For anyone already using our "private" methods, that's instant breakage when it might never have to break.

pirj:
> It makes the codebase ugly.

Sorry, I didn't express myself clearly. We do this temporarily, to separate those that are never used by well-known extensions.

If we only mark what is an internal implementation detail with `@api private` and don't make it `private`, we postpone the problem. When we refactor, it will break extensions somewhere between 1.0 and 2.0. I would personally prefer it to break earlier to avoid `gem 'rubocop', '~> 1.0', '< 1.3'` dance.

marcandre:
> We do this temporarily

Ah! Sure, for known "clients" that could be worthwhile to discover issues.

jonatas:
Ouch! sorry for being late here @pirj, the idea is very simple through the `Fastfile`.

I didn't test but it would be something like:

```ruby
Fast.shortcut(:mark_api_private) do
  ruby_files_from(ARGV.last) do |file|
    replace_all('class', file) do |node|
       insert_before(node.loc.expression, "# @api private\n")
    end
  end
end
```
[...]

PR #8448, opened 2020-08-04
Fix line continuations in spec examples

biinari:
When describing the code that should be detected / auto-corrected, some specs do not escape line continuations properly.

For example:

```ruby
<<~RUBY
  something \
    continues
RUBY
```

results in the string `"something   continues\n"` instead of the desired `"something \\\n  continues\n"`

Correct specs escape the backslash or use single quoted heredocs to keep the line continuation in the source under test, like:

```ruby
# single quoted heredoc (looks more like the source would)
<<~'RUBY'
  something \
    continues
RUBY

# or escape the backslash (allows interpolation and other escapes where needed)
<<~RUBY
  something \\
    continues
RUBY
```

In correcting the specs, it became apparent that `Style/NestedParenthesizedCalls` cop did not behave as intended. `RangeHelp` mixin has been adapted to optionally include line continuations (backslash newline sequence).

Feedback is welcome on whether this behaviour on replacing whitespace with parentheses matches with your expectations.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/range_help.rb
+++ b/lib/rubocop/cop/mixin/range_help.rb
@@ -46,7 +46,8 @@ module RuboCop
       end
 
       def range_with_surrounding_space(range:, side: :both,
-                                       newlines: true, whitespace: false)
+                                       newlines: true, whitespace: false,
+                                       continuations: false)
         buffer = @processed_source.buffer
         src = buffer.source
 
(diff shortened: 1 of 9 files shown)


lib/rubocop/cop/mixin/range_help.rb
```diff
@@ -101,17 +105,27 @@ def directions(side)
         end
       end
 
-      def final_pos(src, pos, increment, newlines, whitespace)
+      # rubocop:disable Metrics/ParameterLists
+      def final_pos(src, pos, increment, continuations, newlines, whitespace)
```
marcandre, line 109: I imagine this method is not meant to be used directly? I really need to finish #8278 ...

biinari: Indeed, there's a [comment](https://github.com/rubocop-hq/rubocop/pull/8448/files#diff-fe7a38dd2048339cac7b172ed7cf2668R85) above that says these helpers are not supposed to be used directly. I agree it would be better to mark them individually as `@api private`.

bbatsov:
Nice catch! Thanks!

PR #8450, opened 2020-08-04
Optimize getting node tokens

fatkodima:
Ran on 30k files.

### Before
```
(1) 74659  (    6.8%)  RuboCop::Cop::Layout::SpaceInsideReferenceBrackets#on_send
(2) 39346  (    3.6%)  RuboCop::Cop::Layout::SpaceInsideArrayLiteralBrackets#on_array
    38575  (    3.5%)  RuboCop::Cop::Style::RedundantSelf#on_block
    33935  (    3.1%)  RuboCop::Cop::Layout::SpaceBeforeFirstArg#on_send
    26757  (    2.4%)  RuboCop::Cop::Layout::FirstArgumentIndentation#on_send
    26271  (    2.4%)  RuboCop::Cop::StringHelp#on_str
    24590  (    2.2%)  RuboCop::Cop::Layout::DefEndAlignment#on_send
    23826  (    2.2%)  RuboCop::Cop::Layout::IndentationConsistency#on_begin
(3) 17870  (    1.6%)  RuboCop::Cop::Layout::SpaceInsideHashLiteralBraces#on_hash
    16582  (    1.5%)  RuboCop::Cop::MultilineExpressionIndentation#on_send
    15623  (    1.4%)  RuboCop::Cop::Interpolation#on_dstr
......
```

### After
```
    45373  (    3.7%)  RuboCop::Cop::Style::RedundantSelf#on_block
    40617  (    3.3%)  RuboCop::Cop::Layout::SpaceBeforeFirstArg#on_send
    36688  (    3.0%)  RuboCop::Cop::Layout::FirstArgumentIndentation#on_send
    31966  (    2.6%)  RuboCop::Cop::StringHelp#on_str
    31289  (    2.5%)  RuboCop::Cop::Layout::DefEndAlignment#on_send
    29682  (    2.4%)  RuboCop::Cop::Layout::IndentationConsistency#on_begin
(1) 25490  (    2.1%)  RuboCop::Cop::Layout::SpaceInsideReferenceBrackets#on_send
    23245  (    1.9%)  RuboCop::Cop::MultilineExpressionIndentation#on_send
    18734  (    1.5%)  RuboCop::Cop::Style::AccessModifierDeclarations#on_send
    17297  (    1.4%)  RuboCop::Cop::Layout::ArgumentAlignment#on_send
    16772  (    1.4%)  RuboCop::Cop::Style::FormatStringToken#on_str
    15592  (    1.3%)  RuboCop::Cop::Style::NumericPredicate#on_send
    14655  (    1.2%)  RuboCop::Cop::Layout::LineLength#on_potential_breakable_node
    14547  (    1.2%)  RuboCop::Cop::Layout::DotPosition#on_send
    13943  (    1.1%)  RuboCop::Cop::Layout::EmptyLinesAroundAccessModifier#on_send
    13836  (    1.1%)  RuboCop::Cop::Layout::SpaceAroundOperators#on_send
    13180  (    1.1%)  RuboCop::Cop::Style::TrailingCommaInArguments#on_send
    12772  (    1.0%)  RuboCop::Cop::Style::ConditionalAssignment#on_send
    12746  (    1.0%)  RuboCop::Cop::Metrics::BlockLength#on_block
    11655  (    0.9%)  RuboCop::Cop::Style::ZeroLengthPredicate#on_send
    11447  (    0.9%)  RuboCop::Cop::Style::MethodCallWithoutArgsParentheses#on_send
    11365  (    0.9%)  RuboCop::Cop::CheckAssignment#on_send
    11055  (    0.9%)  RuboCop::Cop::Layout::IndentationWidth#on_send
    10931  (    0.9%)  RuboCop::Cop::MethodComplexity#on_def
    10492  (    0.9%)  RuboCop::Cop::MethodComplexity#on_def
    10311  (    0.8%)  RuboCop::Cop::Style::InverseMethods#on_send
    10222  (    0.8%)  RuboCop::Cop::StringHelp#on_str
    10211  (    0.8%)  RuboCop::Cop::Layout::ClosingParenthesisIndentation#on_send
    10131  (    0.8%)  RuboCop::Cop::Interpolation#on_dstr
    9837  (    0.8%)  RuboCop::Cop::Style::SignalException#on_send
    9732  (    0.8%)  RuboCop::Cop::Layout::BlockAlignment#on_block
    9620  (    0.8%)  RuboCop::Cop::Style::RedundantSort#on_send
(2) 9319  (    0.8%)  RuboCop::Cop::Layout::SpaceInsideArrayLiteralBrackets#on_array
    8877  (    0.7%)  RuboCop::Cop::Style::EmptyLiteral#on_send
    8427  (    0.7%)  RuboCop::Cop::Style::RedundantException#on_send
    8300  (    0.7%)  RuboCop::Cop::Style::ExpandPathArguments#on_send
    8288  (    0.7%)  RuboCop::Cop::Lint::SafeNavigationChain#on_send
    7771  (    0.6%)  RuboCop::Cop::CheckAssignment#on_send
    7580  (    0.6%)  RuboCop::Cop::Style::EvalWithLocation#on_send
    7485  (    0.6%)  RuboCop::Cop::Style::NonNilCheck#on_send
```

Improvement: `9-10%`

### DIFF
--- a/lib/rubocop/cop/tokens_util.rb
+++ b/lib/rubocop/cop/tokens_util.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+
+module RuboCop
+  # Common methods and behaviors for dealing with tokens.
+  module TokensUtil
+    module_function
+
+    # rubocop:disable Metrics/AbcSize
+    def tokens(node)
+      @tokens ||= {}
+      return @tokens[node.object_id] if @tokens[node.object_id]
+
+      @tokens[node.object_id] =
+        # The tokens list is always sorted by token position,
+        # except for cases when heredoc is passed as a method argument.
+        # In this case tokens are interleaved by heredoc contents' tokens.
+        # We can try a fast (binary) search, assuming the mentioned cases are rare,
+        # and fallback to linear search if failed.
+        if (tokens = fast_tokens(node))
+          tokens
+        else
+          begin_pos = node.source_range.begin_pos
+          end_pos   = node.source_range.end_pos
+
+          processed_source.tokens.select do |token|
+            token.end_pos <= end_pos && token.begin_pos >= begin_pos
+          end
+        end
+    end
+    # rubocop:enable Metrics/AbcSize
+
+    def index_of_first_token(node)
+      index = fast_index_of_first_token(node)
+      return index if index
+
+      begin_pos = node.source_range.begin_pos
+      processed_source.tokens.index { |token| token.begin_pos == begin_pos }
+    end
+
+    def index_of_last_token(node)
+      index = fast_index_of_last_token(node)
+      return index if index
+
+      end_pos = node.source_range.end_pos
+      processed_source.tokens.index { |token| token.end_pos == end_pos }
+    end
+
+    private
+
+    def fast_index_of_first_token(node)
+      begin_pos = node.source_range.begin_pos
+      tokens = processed_source.tokens
+
+      index = tokens.bsearch_index { |token| token.begin_pos >= begin_pos }
+      index if index && tokens[index].begin_pos == begin_pos
+    end
+
+    def fast_index_of_last_token(node)
+      end_pos = node.source_range.end_pos
+      tokens = processed_source.tokens
+
+      index = tokens.bsearch_index { |token| token.end_pos >= end_pos }
+      index if index && tokens[index].end_pos == end_pos
+    end
+
+    def fast_tokens(node)
+      begin_index = index_of_first_token(node)
+      end_index   = index_of_last_token(node)
+
+      tokens = processed_source.tokens[begin_index..end_index]
+      tokens if sorted_tokens?(tokens)
+    end
+
+    def sorted_tokens?(tokens)
+      prev_begin_pos = -1
+      tokens.each do |token|
+        return false if token.begin_pos < prev_begin_pos
+
+        prev_begin_pos = token.begin_pos
+      end
+      true
+    end
+  end
+end
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -23,6 +23,7 @@ require_relative 'rubocop/string_interpreter'
 require_relative 'rubocop/error'
 require_relative 'rubocop/warning'
 
+require_relative 'rubocop/cop/tokens_util'
 require_relative 'rubocop/cop/util'
 require_relative 'rubocop/cop/offense'
 require_relative 'rubocop/cop/message_annotator'
--- a/lib/rubocop/cop/mixin/surrounding_space.rb
+++ b/lib/rubocop/cop/mixin/surrounding_space.rb
@@ -30,31 +30,6 @@ module RuboCop
         Parser::Source::Range.new(buffer, begin_pos, end_pos)
       end
 
-      def index_of_first_token(node)
-        range = node.source_range
-        token_table[range.line][range.column]
-      end
-
-      def index_of_last_token(node)
-        range = node.source_range
-        table_row = token_table[range.last_line]
-        (0...range.last_column).reverse_each do |c|
-          ix = table_row[c]
-          return ix if ix
-        end
-      end
-
-      def token_table
-        @token_table ||= begin
-          table = {}
-          processed_source.tokens.each_with_index do |t, ix|
-            table[t.line] ||= {}
-            table[t.line][t.column] = ix
-          end
-          table
-        end
-      end
-
       def on_new_investigation
         @token_table = nil
         super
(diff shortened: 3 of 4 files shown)


marcandre:
Awesome!
It's unfortunate that tokens aren't always sorted, I didn't know this 🙇‍♂️

I'm ok to merge this, but one easy thing we really should be doing is putting the caching / searching as methods of `ProcessedSource`. This would have two advantages: methods not being accessible to cops (the fact that your mixin's methods are private makes no difference) and the cache doesn't have to be rebuilt for different cops (although I imagine they might not access the same nodes?). WDYT?

Also: did you try sorting the tokens by positions?

lib/rubocop/cop/tokens_util.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+
+module RuboCop
+  # Common methods and behaviors for dealing with tokens.
+  module TokensUtil
+    module_function
+
+    # rubocop:disable Metrics/AbcSize
+    def tokens(node)
+      @tokens ||= {}
```
marcandre, line 10: I realize you just copy-pasted, but `{}.compare_by_identity` would be nicer to use.

fatkodima:
> and the cache doesn't have to be rebuilt for different cops (although I imagine they might not access the same nodes?)

Yes, looks like the cache not needed in this case. I have also removed such a cache https://github.com/rubocop-hq/rubocop/pull/8450/files#diff-938b46da0d108c71380ac9427b83dde0L47-L56 and didn't notice any difference. In my previous PRs, where I tried to reduce memory, I've seen that method allocated enough memory (can't remember how much), so a win here also.

> WDYT?

Yes, will move those methods to `ProcessedSource`. 

> Also: did you try sorting the tokens by positions?

This will always work in `O(nlogn)`, while original works in `O(n)` and the proposed approach almost always will work in `O(logn + #node_tokens)`. Or I'm missing something?

marcandre:
> This will always work in `O(nlogn)`

Correct

> while original works in `O(n)`

Yes, each search is `O(n)`, but the number of searches is also roughly linear with the number of tokens, so overall it was `O(n^2)`, and now with bsearch it's `O(n log(n))` overall, so maybe a sort (assuming it's once per `ProcessedSource` and not per cop) is similar speed (and simpler), or it could be slower or faster, I don't know.

Probably better would simply be to index the tokens by start and end position and do a straight lookups, since we will should always find tokens that start and end exactly where we are looking to, right? Building the indices would be `O(n)` and each lookup would be `O(1)`.

fatkodima:
> but the number of searches is also roughly linear with the number of tokens

Why? I would say, it is a constant. Most of the time, we are not performing searches for every token, but just for some of them, like in `Lint/SpaceInsideArrayLiteralBrackets` for example, for the whole file we may be searching only for just `[` and `]`.

marcandre:
> > but the number of searches is also roughly linear with the number of tokens
> 
> Why? I would say, it is a constant. Most of the time, we are not performing searches for every token, but just for some of them, like in `Lint/SpaceInsideArrayLiteralBrackets` for example, for the whole file we may be searching only for just `[` and `]`.

Right, but how many `[` ... `]` literals will you find in a Ruby file of `n` bytes, on average? I imagine that the best approximation is linear with `n`. The size in bytes, the number of tokens, the number of `def` nodes, of comments, of lines of codes, of string literals, of array literals, etc., all of these should be roughly linear to one another, even if they are not directly related...

What do you think of my idea of indexing the begin and end position of tokens?

fatkodima:
> What do you think of my idea of indexing the begin and end position of tokens?

I'm still have doubt that this will make things faster. Will implement all 3 approaches (original, sorting and indexing) and report results here.

marcandre:
> I'm still have doubt that this will make things faster. Will implement all 3 approaches (original, sorting and indexing) and report results here.

That would be great, but I wouldn't want you to feel forced; as stated previously, I'm 👍 to merge this as is.

fatkodima:
> but I wouldn't want you to feel forced

No problems 😄 

Ok, I have tested 3 implementations and was a bit surprised as I get almost identical results for them. I rechecked, I was testing 3 different versions, not the same one, so everything ok here.

With sorting version, I get the most concise code and it does not allocate extra space, like in indexing version, so I would implement that and submit a PR into `rubocop-ast` tomorrow.

marcandre:
> Ok, I have tested 3 implementations and was a bit surprised as I get almost identical results for them. I rechecked, I was testing 3 different versions, not the same one, so everything ok here.

Cool! Glad you checked 👍

bbatsov:
Good work!

marcandre:
> Good work!

🤣 we'll be picking a different implementation, but it won't be difficult to revert this

bbatsov:
Yeah, I saw this, but I'm thinking of cutting a release later today, so that's going to benefit our users in the mean time.

PR #8455, opened 2020-08-04
Register an offense for yield without arguments in Style/ExplicitBlockArgument

eugeneius:
I don't see a reason not to handle this case, and it's far more common in our codebase than the case where yield is called with arguments.

As with https://github.com/rubocop-hq/rubocop/pull/8453, I didn't add a changelog entry because this cop hasn't been released yet.

-----------------

* [x] Added tests.


koic:
Yeah, this will make the behavior consistent for the style.

marcandre:
It's an unsafe example though since `Integer#times` yields a number

PR #8463, opened 2020-08-05
Set valid ref range when regexp is matched in Lint/OutOfRangeRegexpRef

eugeneius:
Last match global variables are set when the regexp is matched, not when it's defined. This meant that the previous implementation gave incorrect results when those things happened separately, e.g. when a regexp was assigned to a local variable or a constant and then matched elsewhere.

This change means that the cop only handles matching on regexp literals, but that was the only case where it worked correctly before anyway. It should be possible to improve the cop to handle more complex cases with `VariableForce`, but that can be done as a separate change.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


marcandre:
Agreed, this cop is too eager. I should have thought of that.

Would be nice to add the case `'string'.gsub(/.../) { ... }` and `case str; when /..../ then ....`
Do you feel like handling that?

eugeneius:
Sure, I'm taking a look now.

PR #8464, opened 2020-08-05
Handle more matching methods in Lint/OutOfRangeRegexpRef

eugeneius:
Followup to @marcandre's comment here: https://github.com/rubocop-hq/rubocop/pull/8463#issuecomment-669550324

`when`, `grep`, `gsub`, `gsub!` `sub`, `sub!`, `[]`, `slice`, `slice!` `scan`, `index`, `rindex`, `partition`, `rpartition`, `start_with?`, and `end_with?` can all be used to match regexps and set last match global variables, so we should update the range of valid references when we encounter them.

These were all of the relevant methods I could find from a quick grep of [ruby/spec](https://github.com/ruby/spec), although there could well be more.

**Update:** I read through the [String class' docs](https://ruby-doc.org/core-2.7.1/String.html) and added several more methods.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
+++ b/lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
@@ -21,7 +21,10 @@ module RuboCop
       class OutOfRangeRegexpRef < Base
         MSG = 'Do not use out of range reference for the Regexp.'
 
-        REGEXP_CAPTURE_METHODS = %i[=~ === match].to_set.freeze
+        REGEXP_RECEIVER_METHODS = %i[=~ === match].to_set.freeze
+        REGEXP_ARGUMENT_METHODS = %i[=~ match grep gsub gsub! sub sub! [] slice slice! index rindex
+                                     scan partition rpartition start_with? end_with?].to_set.freeze
+        REGEXP_CAPTURE_METHODS = (REGEXP_RECEIVER_METHODS + REGEXP_ARGUMENT_METHODS).freeze
 
         def on_new_investigation
           @valid_ref = 0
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -38,11 +41,23 @@ def on_send(node)
...
 
+        def on_when(node)
+          @valid_ref = nil
+
+          regexp_conditions = node.conditions.select(&:regexp_type?)
+          return if regexp_conditions.empty?
+
+          @valid_ref = regexp_conditions.map do |condition|
+            check_regexp(condition)
+          end.compact.min
```
marcandre, line 58: Should be `max`, right?

spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
```diff
@@ -133,4 +133,146 @@
...
+      end
+    RUBY
+  end
+
+  it 'uses the minimum number of captures for when clauses with multiple conditions' do
+    expect_offense(<<~RUBY)
+      case "foobarbaz"
+      when /(foo)(bar)/, /(bar)baz/
+        $2
+        ^^ Do not use out of range reference for the Regexp.
```
marcandre, line 161: No offense here, $2 might be `'bar'`, right?

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -38,11 +41,23 @@ def on_send(node)
...
+            && REGEXP_ARGUMENT_METHODS.include?(node.method_name)
             check_regexp(node.first_argument)
           end
         end
 
+        def on_when(node)
+          @valid_ref = nil
+
+          regexp_conditions = node.conditions.select(&:regexp_type?)
+          return if regexp_conditions.empty?
```
marcandre, line 54: You could remove this and `@valid_ref = nil`, as `[].max # => nil`. Might not be as clear though, your choice

spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
eugeneius: My reasoning was that the backref needs to be valid for _all_ branches, but now I think you're right that getting nil here in some cases could be an intentional idiom, and my interpretation is too strict.

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
eugeneius: This was a case of red, green, no refactor. :) I'll ✂️

marcandre:
Looks great, thanks for the PR! 👍

PR #8474, opened 2020-08-06
Add new `Lint/DuplicateRequire` cop

fatkodima:
This cop checks for duplicate `require`s and `require_relative`s.

I ran it on 30k files and it detected 13 offenses, so looks useful.

This cop does not detect cases like `require 'foo'` vs `require './foo'` and the like. Looks like those will be very rare cases and not sure if it is worth the complexity.


bbatsov:
One thing to keep in mind is that some requires might be conditional (e.g. in some `if`, instead at the top-level), but those cases are rare and we probably shouldn't worry about them too much.

lib/rubocop/cop/lint/duplicate_require.rb
```diff
@@ -0,0 +1,52 @@
...
+          left_siblings_of(node).each do |sibling|
+            next unless require_call?(node) && duplicate_require?(node, sibling)
+
+            add_offense(node, message: format(MSG, method: node.method_name))
+          end
+        end
+
+        private
+
+        def left_siblings_of(node)
```
fatkodima, line 38: There are a lot of places where such `siblings`, `left_siblings` are used. 
@marcandre Wdyt about adding such helpers to `rubocop-ast`?

marcandre: Agreed. rubocop-hq/rubocop-ast#93

lib/rubocop/cop/lint/duplicate_require.rb
```diff
@@ -0,0 +1,52 @@
...
+        REQUIRE_METHODS = %i[require require_relative].freeze
+
+        def_node_matcher :require_call?, <<~PATTERN
+          (send {nil? (const _ :Kernel)} {:#{REQUIRE_METHODS.join(' :')}} _)
+        PATTERN
+
+        def on_send(node)
+          return unless REQUIRE_METHODS.include?(node.method_name)
+
+          left_siblings_of(node).each do |sibling|
```
marcandre, line 29: This is `O(n^2)`, (`n` == number of requires), would it be hard to refactor as `O(n)` as it should be using a set?

fatkodima: >  as it should be using a set?

Can you provide a hint on how can I rewrite to use this?

marcandre: Sure. I haven't tested, but I think that something like this should work:
```ruby
def on_new_investigation
  # Holds the known required files for a given parent node (used as key)
  @required = Hash.new{|h, k| h[k]=Set[] }.compare_by_identity
end

def on_send(node)
  return unless REQUIRE_METHODS.include?(node.method_name)

  next if @required[node.parent].add?(node.first_argument)
  add_offense(node, message: format(MSG, method: node.method_name))
end
```

fatkodima: Updated with suggestion.

marcandre:
Good stuff, thanks! 🎉

PR #8477, opened 2020-08-06
[Fix #8475] Fix a false positive for Style/HashAsLastArrayItem when t…

### LINKED ISSUE #8475: Style/HashAsLastArrayItem copy with style "no_braces" can result in invalid code
## Expected behavior

Using Style/HashAsLastArrayItem with EnforcedStyle: no_braces should yield valid Ruby code when there are duplicate Hashes in the array.  This is similar to https://github.com/rubocop-hq/rubocop/issues/8323, but has not been fully resolved when there are identical hashes in the array. 

## Actual behavior

The cop is not safe and results in invalid Ruby code.

```yaml
# .rubocop.yml
AllCops:
  NewCops: enable

Style/HashAsLastArrayItem:
  EnforcedStyle: no_braces
```

```ruby
# hash_as_last_array_item.rb before
# frozen-string-literal: true

# Note that the hashes are identical
a = [
  {
    foo: :bar,
  },
  {
    foo: :bar
  }
]

pp a
```

```console
$ ruby hash_as_last_array_item.rb
[{:foo=>:bar}, {:foo=>:bar}]
```

```console
$ rubocop -a hash_as_last_array_item.rb
Inspecting 1 file
E

Offenses:

hash_as_last_array_item.rb:5:3: C: [Corrected] Style/HashAsLastArrayItem: Omit the braces around the hash.
  { ...
  ^
hash_as_last_array_item.rb:6:14: C: [Corrected] Style/TrailingCommaInHashLiteral: Avoid comma after the last item of a hash.
    foo: :bar,
             ^
hash_as_last_array_item.rb:7:3: E: Lint/Syntax: unexpected token tCOMMA
  ,
  ^
hash_as_last_array_item.rb:8:3: C: [Corrected] Style/HashAsLastArrayItem: Omit the braces around the hash.
  { ...
  ^

1 file inspected, 4 offenses detected, 3 offenses corrected
```

```ruby
# hash_as_last_array_item.rb after
# frozen-string-literal: true

a = [

    foo: :bar
  ,

    foo: :bar

]

pp a
```

```console
$ ruby hash_as_last_array_item.rb
hash_as_last_array_item.rb:7: syntax error, unexpected ',', expecting ']'
hash_as_last_array_item.rb:11: syntax error, unexpected ']', expecting end-of-input
```

## RuboCop version

```
$ rubocop -v
0.89.0
```

wcmonty:
Closes #8475

The issue is that we were previously comparing values to the last node.  If a node equals the last node it will be marked with an offense, when it should not be.

### DIFF
--- a/lib/rubocop/cop/style/hash_as_last_array_item.rb
+++ b/lib/rubocop/cop/style/hash_as_last_array_item.rb
@@ -40,7 +40,7 @@ module RuboCop
           parent = node.parent
           return false unless parent
 
-          parent.array_type? && parent.values.last == node
+          parent.array_type? && parent.children.last.equal?(node)
         end
 
         def check_braces(node)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#8478](https://github.com/rubocop-hq/rubocop/issues/8478): Relax `Lint/BinaryOperatorWithIdenticalOperands` for mathematical operations. ([@marcandre][])
 * [#8480](https://github.com/rubocop-hq/rubocop/issues/8480): Tweak callback list of `Lint/MissingSuper`. ([@marcandre][])
 * [#8481](https://github.com/rubocop-hq/rubocop/pull/8481): Fix autocorrect for elements with newlines in `Style/SymbolArray` and `Style/WordArray`. ([@biinari][])
+* [#8475](https://github.com/rubocop-hq/rubocop/issues/8475): Fix a false positive for `Style/HashAsLastArrayItem` when there are duplicate hashes in the array. ([@wcmonty][])
 
 ## 0.89.0 (2020-08-05)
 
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/hash_as_last_array_item.rb
```diff
@@ -40,7 +40,7 @@ def last_array_item?(node)
           parent = node.parent
           return false unless parent
 
-          parent.array_type? && parent.values.last == node
+          parent.array_type? && parent.children.last.object_id == node.object_id
```
koic, line 43: I think using `equal?` can be a bit simpler :-)

```suggestion
          parent.array_type? && parent.children.last.equal?(node)
```

wcmonty: Agreed your suggestion is simpler.  I went ahead and committed.  Happy to squash if necessary.

koic:
This looks good to me. Can you add changelog entry and squash your commits into one?
The following is the changelog entry format.

wcmonty:
@koic This is rebased/squashed with the changelog entry added.  One set of specs is failing (https://github.com/rubocop-hq/rubocop/pull/8477/checks?check_run_id=959581086), but it doesn't look related to this PR.

koic:
It was a false alarm of CI. I re-ran it again and passed it. Thank you!

PR #8480, opened 2020-08-06
Tweak Lint/MissingSuper [See #8376]

marcandre:


ruanltbg:
Thank you for the hard/fast work.

fatkodima:
Seems like there are another 3 ones https://ruby-doc.org/core-2.7.1/BasicObject.html#method-i-singleton_method_added
Sorry, forgot about them also.

marcandre:
> Seems like there are another 3 ones https://ruby-doc.org/core-2.7.1/BasicObject.html#method-i-singleton_method_added

I never realize those existed. I can't think of a reason for their existence either, as `def singleton_class.method_added` should have the same effect. I added them to the list 👍

bbatsov:
Indeed!

PR #8483, opened 2020-08-07
Add new `Style/StaticClass` cop

fatkodima:
Implementation of this - https://rubystyle.guide/#modules-vs-classes


lib/rubocop/cop/correctors/empty_line_corrector.rb
```diff
@@ -2,24 +2,24 @@
 
 module RuboCop
   module Cop
-    # This class does empty line auto-correction
-    class EmptyLineCorrector
-      class << self
```
koic, line 7: I think `class << self` is an acceptable case.

fatkodima: But what about if that contains only class method definitions or constants?

koic: Hmm, In any case, I’m wondering what the merit is to change class to module.

fatkodima: For me, module adds to explicitness that this is more a class of utility methods and therefore it does not make sense to create instances of this class. 
Maybe an author of this section of styleguide had more reasons?

koic: Yeah, I think it's agree if it's designed as a module :-) However modifying an existing implementation designed as a class can have unintended potential breaking changes. So, especially when it's published API. I’m worried that this enforced style seems too strong to be applied by default.

fatkodima: Yes, I marked it as unsafe and briefly mentioned in the description of the cop why.
So are you suggesting to make it disabled? If yes, then should it be also disabled for `rubocop`s own codebase?

koic: > So are you suggesting to make it disabled? If yes, then should it be also disabled for rubocops own codebase?

Yes. I'm worried that style preferences will affect an existing design. I'm not sure that it will be preferred to be applied in existing development by default. And it seems to be no reason to enable it on RuboCop’s codebase only.
OTOH, it may be possible to create some condition for it to be enabled by default in the future. However, it seems that the current behavior is a stronger warning than the explanation in the style guide.

spec/rubocop/cop/style/static_class_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  it 'does not register an offense when class has a parent' do
+    expect_no_offenses(<<~RUBY)
+      class C < B
+        def self.class_method; end
+      end
+    RUBY
+  end
+
+  it 'does not register an offense when class includes/extends a module' do
```
marcandre, line 86: Is it the same with `extend`? Why no offense then?

marcandre:
I wish we wouldn't recommens `module_function`... I hope I can [change the style-guide on that one](https://github.com/rubocop-hq/ruby-style-guide/issues/556#issuecomment-653703624)

bbatsov:
I think in this particular case that's fine, as such modules are not meant to be used in the inheritance chain anyways.

spec/rubocop/cop/style/static_class_spec.rb
fatkodima: @marcandre 
Did you mean that in the case
```ruby
class C
  extend M
end
```
we should generate an offense?

marcandre: Yes, that is what I meant. Presumably, the `extend` only adds class methods, so `C` is solely comprised of class methods and this cop should raise an offense, no?

fatkodima: Agreed. Updated with the new code and test case to reflect that.

bbatsov:
Looks good! Thanks!

PR #8486, opened 2020-08-07
Add new `Style/CombinableLoops` cop

fatkodima:
This cop checks for places where multiple consecutive loops over the same data can be combined into a single loop. 
It is very likely that combining them will make the code more efficient and more concise.

```ruby
  # bad
  def method
    items.each do |item|
      do_something(item)
    end

    items.each do |item|
      do_something_else(item)
    end
  end

  # good
  def method
    items.each do |item|
      do_something(item)
      do_something_else(item)
    end
  end
```

Inspired by https://rules.sonarsource.com/java/RSPEC-3047

I ran it on ~30k files and it detected 133 offenses. So seems useful.


spec/rubocop/cop/style/combinable_loops_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Style::CombinableLoops do
+  subject(:cop) { described_class.new }
+
+  it 'registers an offense when looping over the same data as previous loop' do
```
marcandre, line 6: There are no spec for other methods than `each`...

spec/rubocop/cop/lint/void_spec.rb
```diff
@@ -15,18 +15,14 @@
         a %{op} b
       RUBY
     end
-  end
```
marcandre, line 18: I love it when you find things in this codebase 👍

lib/rubocop/cop/style/combinable_loops.rb
```diff
@@ -0,0 +1,62 @@
...
+
+          sibling = left_sibling_of(node)
+          add_offense(node) if same_collection_looping?(node, sibling)
+        end
+
+        private
+
+        def collection_looping_method?(node)
+          method_name = node.send_node.method_name.to_s
+          method_name.start_with?('each') || method_name.end_with?('_each')
```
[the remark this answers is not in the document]
fatkodima: on ruby >= 2.7

config/default.yml
```diff
@@ -2687,6 +2687,13 @@ Style/ColonMethodDefinition:
   Enabled: true
   VersionAdded: '0.52'
 
+Style/CombinableLoops:
+  Description: >-
+                  Checks for places where multiple consecutive loops over the same data
+                  can be combined into a single loop.
+  Enabled: pending
+  VersionAdded: '0.90'
```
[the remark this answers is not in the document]
fatkodima: Can you provide an example? I can think of only when `each`like method is not doing just iteration or modifies a body while iterating, so the next loop iterates over modified collection.

lib/rubocop/cop/style/combinable_loops.rb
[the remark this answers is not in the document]
marcandre: `Regexp#match?` is Ruby 2.4+

config/default.yml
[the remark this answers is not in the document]
marcandre: The case you are referring to, or in general any loop that modifies a state that the second loop depend on:
```ruby
avoid = Set[]
enum.each do |i|
  avoid << i.foo if i.bar
end
enum.each do |i|
  return i.baz if avoid.include?(i.qux)
end
```

Also might be worth checking for potential arguments, e.g. the `matrix` lib accepts an argument to `each`

fatkodima:
Updated with suggestions.

lib/rubocop/cop/style/combinable_loops.rb
```diff
@@ -0,0 +1,65 @@
...
+
+          sibling = left_sibling_of(node)
+          add_offense(node) if same_collection_looping?(node, sibling)
+        end
+
+        private
+
+        def collection_looping_method?(node)
+          method_name = node.send_node.method_name
+          method_name.match?(/^each/) || method_name.match?(/_each$/)
```
marcandre, line 47: the `|| method_name.match?(/_each$/)` is still not covered by the specs, right?

fatkodima: Added test for `reverse_each`

lib/rubocop/cop/style/combinable_loops.rb
```diff
@@ -0,0 +1,65 @@
...
+
+module RuboCop
+  module Cop
+    module Style
+      # This cop checks for places where multiple consecutive loops over the same data
+      # can be combined into a single loop. It is very likely that combining them
+      # will make the code more efficient and more concise.
+      #
+      # It is marked as unsafe, because the first loop can modify
+      # a state that the second loop depends on those two aren't combinable.
```
marcandre, line 11: I would tweak slightly => "It is marked as unsafe, because the first loop might modify a state that the second loop depends on; these two aren't combinable."

bbatsov:
I'm guessing we should make this work with `for` as well, right? As for the name - I can't really think of a better name right now. I'm thinking, however, it'd be nice to have a similar check for conditionals - e.g. consecutive `if`s with the same condition.

fatkodima:
> I'm guessing we should make this work with for as well, right?

Yes, added support for it.

> I'm thinking, however, it'd be nice to have a similar check for conditionals - e.g. consecutive ifs with the same condition.

I guess it will be better implemented as a new cop?

bbatsov:
> I guess it will be better implemented as a new cop?

Yeah, definitely.

PR #8490, opened 2020-08-08
[Fix #8044] Change cop department name calculation

### LINKED ISSUE #8044: Cops defined in gems can conflict even when they are in different namespaces
Raising this issue as discussed here: https://github.com/rubocop-hq/rubocop-rspec/pull/834#issuecomment-570497415

rubocop-rails provides `RuboCop::Cop::Rails::HttpStatus`, and rubocop-rspec provides `RuboCop::Cop::RSpec::Rails::HttpStatus`.

Both cops get the same badge, `Rails/HttpStatus`, and this means that only one cop works (from whichever gem is required last).

Related: https://github.com/rubocop-hq/rubocop-rspec/issues/611

First reported: https://github.com/rubocop-hq/rubocop/issues/5738#issuecomment-388838142

--------

## Expected behavior

Both cops report offenses.

## Actual behavior

Only one cop reports offenses (from whichever gem is required last).

https://gist.github.com/haines/369ac2298ce6fe042865def931e8448b

## RuboCop version

```
0.84.0 (using Parser 2.7.1.3, rubocop-ast 0.0.3, running on ruby 2.7.1 x86_64-darwin19)
```

dmytro-savochkin:
Fixes #8044.

This is a pretty deep change. As I described in the issue page (https://github.com/rubocop-hq/rubocop/issues/8044#issuecomment-667700656) I can't see a way to fix this problem without changing the way cop departments are computed. So I'm proposing this fix.

After the fix the conflicting cops `RuboCop::Cop::Rails::HttpStatus` and `RuboCop::Cop::RSpec::Rails::HttpStatus` will belong to `Rails` and `RSpec/Rails` departments, respectively. Previously they both belong to `Rails` department, thus the issue.

The biggest problem with this solution that all the dependent gems that have deep nested cops would need to update their `config/default.yml` files to reflect the changes. For example, for `rubocop-rspec` the following seven cops would need to be updated:
`Capybara/CurrentPathExpectation` -> `RSpec/Capybara/CurrentPathExpectation`
`Capybara/FeatureMethods` -> `RSpec/Capybara/FeatureMethods` 
`Capybara/VisibilityMatcher` -> `RSpec/Capybara/VisibilityMatcher`
`FactoryBot/AttributeDefinedStatically` -> `RSpec/FactoryBot/AttributeDefinedStatically`
`FactoryBot/CreateList` -> `RSpec/FactoryBot/CreateList`
`FactoryBot/FactoryClassName` -> `RSpec/FactoryBot/FactoryClassName`
`Rails/HttpStatus` -> `RSpec/Rails/HttpStatus`

The good news is that `rubocop-rspec` and `rubocop-i18n` are probably the only gems that would need to be changed.

This change should not affect any cops that have 4 or less levels of nesting. For example, some cop named `Rubocop::Cop::Foo::Bar` now has department `Foo` and this will still be the case after this change. Cop `Rubocop::Cop::Foo::Bar::Baz`, which now has department `Bar`, will have department `Foo/Bar` after the change however.  

Due to the magnitude of this in the discussion in the comments we agreed to move forward with this when rubocop hits version 1.0.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/badge.rb
+++ b/lib/rubocop/cop/badge.rb
@@ -10,37 +10,22 @@ module RuboCop
     # allow for badge references in source files that omit the department for
     # RuboCop to infer.
     class Badge
-      # Error raised when a badge parse fails.
-      class InvalidBadge < Error
-        MSG = 'Invalid badge %<badge>p. ' \
-              'Expected `Department/CopName` or `CopName`.'
-
-        def initialize(token)
-          super(format(MSG, badge: token))
-        end
-      end
-
       attr_reader :department, :cop_name
 
       def self.for(class_name)
-        new(*class_name.split('::').last(2))
+        parts = class_name.split('::')
+        name_deep_enough = parts.length >= 4
+        new(name_deep_enough ? parts[2..-1] : parts.last(2))
       end
 
       def self.parse(identifier)
-        parts = identifier.split('/', 2)
-
-        raise InvalidBadge, identifier if parts.size > 2
-
-        if parts.one?
-          new(nil, *parts)
-        else
-          new(*parts)
-        end
+        new(identifier.split('/'))
       end
 
-      def initialize(department, cop_name)
-        @department = department.to_sym if department
-        @cop_name   = cop_name
+      def initialize(class_name_parts)
+        department_parts = class_name_parts[0...-1]
+        @department = (department_parts.join('/').to_sym unless department_parts.empty?)
+        @cop_name = class_name_parts.last
       end
 
       def ==(other)
@@ -66,7 +51,7 @@ module RuboCop
       end
 
       def with_department(department)
-        self.class.new(department, cop_name)
+        self.class.new([department.to_s.split('/'), cop_name].flatten)
       end
     end
   end
--- a/spec/rubocop/cop/badge_spec.rb
+++ b/spec/rubocop/cop/badge_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::Badge do
-  subject(:badge) { described_class.new('Test', 'ModuleMustBeAClassCop') }
+  subject(:badge) { described_class.new(%w[Test ModuleMustBeAClassCop]) }
 
   it 'exposes department name' do
     expect(badge.department).to be(:Test)
@@ -11,34 +11,59 @@ RSpec.describe RuboCop::Cop::Badge do
     expect(badge.cop_name).to eql('ModuleMustBeAClassCop')
   end
 
-  describe '.parse' do
-    it 'parses Department/CopName syntax' do
-      expect(described_class.parse('Foo/Bar'))
-        .to eq(described_class.new('Foo', 'Bar'))
+  describe '.new' do
+    shared_examples 'assignment of department and name' do |class_name_parts, department, name|
+      it 'assigns department' do
+        expect(described_class.new(class_name_parts).department).to eq(department)
+      end
+
+      it 'assigns name' do
+        expect(described_class.new(class_name_parts).cop_name).to eq(name)
+      end
     end
 
-    it 'parses unqualified badge references' do
-      expect(described_class.parse('Bar'))
-        .to eql(described_class.new(nil, 'Bar'))
+    include_examples 'assignment of department and name', %w[Foo], nil, 'Foo'
+    include_examples 'assignment of department and name', %w[Foo Bar], :Foo, 'Bar'
+    include_examples 'assignment of department and name', %w[Foo Bar Baz], :'Foo/Bar', 'Baz'
+    include_examples 'assignment of department and name', %w[Foo Bar Baz Qux], :'Foo/Bar/Baz', 'Qux'
+  end
+
+  describe '.parse' do
+    shared_examples 'cop identifier parsing' do |identifier, class_name_parts|
+      it 'parses identifier' do
+        expect(described_class.parse(identifier)).to eq(described_class.new(class_name_parts))
+      end
     end
+
+    include_examples 'cop identifier parsing', 'Bar', %w[Bar]
+    include_examples 'cop identifier parsing', 'Foo/Bar', %w[Foo Bar]
+    include_examples 'cop identifier parsing', 'Foo/Bar/Baz', %w[Foo Bar Baz]
+    include_examples 'cop identifier parsing', 'Foo/Bar/Baz/Qux', %w[Foo Bar Baz Qux]
   end
 
   describe '.for' do
-    it 'parses cop class name' do
-      expect(described_class.for('RuboCop::Cop::Foo::Bar'))
-        .to eq(described_class.new('Foo', 'Bar'))
+    shared_examples 'cop class name parsing' do |class_name, class_name_parts|
+      it 'parses cop class name' do
+        expect(described_class.for(class_name)).to eq(described_class.new(class_name_parts))
+      end
     end
+
+    include_examples 'cop class name parsing', 'Foo', %w[Foo]
+    include_examples 'cop class name parsing', 'Foo::Bar', %w[Foo Bar]
+    include_examples 'cop class name parsing', 'RuboCop::Cop::Foo', %w[Cop Foo]
+    include_examples 'cop class name parsing', 'RuboCop::Cop::Foo::Bar', %w[Foo Bar]
+    include_examples 'cop class name parsing', 'RuboCop::Cop::Foo::Bar::Baz', %w[Foo Bar Baz]
   end
 
   it 'compares by value' do
-    badge1 = described_class.new('Foo', 'Bar')
-    badge2 = described_class.new('Foo', 'Bar')
+    badge1 = described_class.new(%w[Foo Bar])
+    badge2 = described_class.new(%w[Foo Bar])
 
     expect(Set.new([badge1, badge2]).one?).to be(true)
   end
 
   it 'can be converted to a string with the Department/CopName format' do
-    expect(described_class.new('Foo', 'Bar').to_s).to eql('Foo/Bar')
+    expect(described_class.new(%w[Foo Bar]).to_s).to eql('Foo/Bar')
   end
 
   describe '#qualified?' do
@@ -48,6 +73,10 @@ RSpec.describe RuboCop::Cop::Badge do
 
     it 'says `Department/CopName` is qualified' do
       expect(described_class.parse('Department/Bar').qualified?).to be(true)
+    end
+
+    it 'says `Deep/Department/CopName` is qualified' do
+      expect(described_class.parse('Deep/Department/Bar').qualified?).to be(true)
     end
   end
 end
--- a/spec/rubocop/cop/registry_spec.rb
+++ b/spec/rubocop/cop/registry_spec.rb
@@ -172,6 +172,21 @@ RSpec.describe RuboCop::Cop::Registry do
     it 'exposes a list of cops' do
       expect(registry.cops).to eql(cops)
     end
+
+    context 'with cops having the same inner-most module' do
+      let(:cops) do
+        [RuboCop::Cop::Foo::Bar, RuboCop::Cop::Baz::Foo::Bar]
+      end
+
+      before do
+        stub_const('RuboCop::Cop::Foo::Bar', Class.new(RuboCop::Cop::Base))
+        stub_const('RuboCop::Cop::Baz::Foo::Bar', Class.new(RuboCop::Cop::Base))
+      end
+
+      it 'exposes both cops' do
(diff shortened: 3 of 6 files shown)


bbatsov:
I like the proposed change. In the past I simply didn't envision that there would be more than one level of namespacing, but this clearly has happened regardless. :-) @rubocop-hq/rubocop-core any objections to changing this?

marcandre:
Having a department like `"RSpec/Capybara"` seems good and necessary.

Could we look at a an opt-in way instead? For example, `rubocop-rspec` seems to be the one with the most need for this; all these cops inherit from `RSpec::Cop` so a good and easy solution could be to have a hook for computing the department and for `RSpec::Cop` to refine it. Either a `namespace` class method that it would define as `"RSpec"`, or a `department_name` class method that could look like `"RSpec/#{super}"` maybe?

Also, we already have some algorithm to lookup Cops equivalents, it would be nice to tweak them so that they can understand that "Capybara/FeatureMethods" => "RSpec/Capybara/FeatureMethods"

bquorning:
cc @Darhazer, @pirj 

> The good news is `rubocop-rspec` is probably the only one among _widely_ used gems that would need to be changed.

Another good thing is that we plan to do a major release (v2.0) shortly after RuboCop v1 is released. Which would be a good opportunity to release a breaking change like this. Of course, that necessitates waiting with merging this change until right before the 1.0 release. Or, making an opt-in solution like suggested by @marcandre and rubocop-rspec can then just drop the fallback in our 2.0 release.

dmytro-savochkin:
> Having a department like `"RSpec/Capybara"` seems good and necessary.
> 
> Could we look at a an opt-in way instead? For example, `rubocop-rspec` seems to be the one with the most need for this; all these cops inherit from `RSpec::Cop` so a good and easy solution could be to have a hook for computing the department and for `RSpec::Cop` to refine it. Either a `namespace` class method that it would define as `"RSpec"`, or a `department_name` class method that could look like `"RSpec/#{super}"` maybe?
> 
> Also, we already have some algorithm to lookup Cops equivalents, it would be nice to tweak them so that they can understand that "Capybara/FeatureMethods" => "RSpec/Capybara/FeatureMethods"

@marcandre, those are nice points!

However, if I understood you correctly, I wouldn't want to make it an opt-in solution. Because in such a case it would fix the immediate problem but still leave room for problems in the future. What would happen if two different gems would choose the same department name for their new cop? In such an optional approach it's a matter of time when some confusion happens again. On the contrary, in the proposed solution the only conflict that could happen is when there are two cops with exactly the same namespaces and names, which is probably pretty easy to detect early.

As to the second point, yes, that might probably help a lot! I'll try to look into this later.

pirj:
Nice, thanks a lot for dealing with this long-standing issue!

Currently, extensions (checked the infamous `rubocop-rspec`) will cease to work with:
```
An error occurred while RSpec/FactoryBot/CreateList cop was inspecting /Users/pirj/source/rubocop-rspec/spec/project/changelog_spec.rb:3:0
```
Does it make sense to add an error message for config loader to notify that e.g. `FactoryBot/CreateList` that is defined in extension's `config/default.yml` (or user's `.rubocop.yml` even) does not exist?

dmytro-savochkin:
Okay, so I tried to do something to address the raised concerns.

At first, I wasn't able to write some general code that would catch all the 'undeepened' names and convert them to 'deepened' ones. It really seemed to be pretty hard.
But later I thought that maybe we can just find all the cases and work with them? So I went to rubygems and manually checked all the gems that seem to be related in some way to rubocop (https://rubygems.org/search?utf8=%E2%9C%93&query=rubocop). And I found another one that has too deep namespace: `rubocop-i18n`. So, it's actually only two gems! After that I tried to fix this somehow and the only thing that seemed to work was to inject some sort of 'fixer' inside `Config` class. I wanted to use `ConfigValidator` or `ConfigObsoletion` but rubocop-rspec seems to be injecting its data into `Config` directly in its `RuboCop::RSpec::Inject` class. Maybe this can be changed if taking a little more time but this still illustrates the idea.

So, this new `ConfigObsoletionExternal` class I added iterates over all the known problematic config entries and changes them silently to the corrected ones. It doesn't do anything with `Rails/HttpStatus` entry though because it's pretty hard to understand where it belongs to (to `Rails` or `RSpec/Rails`?). We can add some sort of warning messages in the debug mode later if this approach looks okay.

So, what do you guys think? 
And by the way, when do you plan to release the 1.0 version? Because it is now starting to look more and more complicated and only to solve the problem of handling different versions together. If the release is pretty soon, say in this year, maybe it would be easier to just wait and merge it without any messy validation/fixing code?

pirj:
> maybe it would be easier to just wait and merge it without any messy validation/fixing code?

Agree. Release this change with 1.0 (but not earlier), and we'll make sure `rubocop-rspec` will follow. We already have preventive measures in the upcoming release to depend of `rubocop` < 1.0. Wish we did that earlier.

lib/rubocop/cop/badge.rb
```diff
@@ -10,37 +10,23 @@ module Cop
...
-          super(format(MSG, badge: token))
-        end
-      end
-
       attr_reader :department, :cop_name
 
       def self.for(class_name)
-        new(*class_name.split('::').last(2))
+        parts = class_name.split('::')
+        parts_without_first_two = parts[2..-1]
```
bquorning, line 17: Could we come up with a better name for this variable?

Actually, I wonder if what we are _really_ checking is whether the class name is a subclass of `RuboCop::Cop`/`RuboCop::Base`? In that case, I’d suggest a more explicit namespace check:
```rb
parts = class_name.split('::')
if class_name.start_with?('RuboCop::Cop', 'RuboCop::Base')
  new(parts[2..-1])
else
  new(parts.last(2))
end
```

spec/rubocop/cop/badge_spec.rb
```diff
@@ -11,34 +11,59 @@
...
-        .to eq(described_class.new('Foo', 'Bar'))
+    shared_examples 'cop class name parsing' do |class_name, class_name_parts|
+      it 'parses cop class name' do
+        expect(described_class.for(class_name)).to eq(described_class.new(class_name_parts))
+      end
     end
+
+    include_examples 'cop class name parsing', 'Foo', %w[Foo]
+    include_examples 'cop class name parsing', 'Foo::Bar', %w[Foo Bar]
+    include_examples 'cop class name parsing', 'RuboCop::Cop::Foo', %w[Cop Foo]
```
bquorning, line 53: I think this line is wrong. We don’t want the class name `"RuboCop::Cop::Foo"` to be recognized as a cop `Foo` with department `Cop`. Since the string starts with `RuboCop::Cop`, `Foo` must clearly be the department name, and the cop’s name is missing. I would expect` Badge.for` to raise an error.

spec/rubocop/cop/registry_spec.rb
```diff
@@ -172,6 +172,24 @@
...
+      let(:cops) do
+        [
+          RuboCop::Cop::Foo::Bar,
+          RuboCop::Cop::Baz::Foo::Bar
+        ]
+      end
+
+      before do
+        stub_const('RuboCop::Cop::Baz::Foo::Bar', Class.new(RuboCop::Cop::Base))
+        stub_const('RuboCop::Cop::Foo::Bar', Class.new(RuboCop::Cop::Base))
```
bquorning, line 186: Absolute nitpick, but could you please switch these two lines, to be in the same order as lines 179–180?

spec/rubocop/cop/registry_spec.rb
```diff
@@ -172,6 +172,24 @@
...
+      end
+
+      it 'exposes both cops' do
+        expect(registry.cops).to eql(cops)
```
bquorning, line 190: Actually, it would be fine to just inline the value of `cops` here, in the spec example.

lib/rubocop/cop/badge.rb
pirj: Thank could also be `MyCoolCompany::Cop::Mailers::ExplicitRecipient`.

bquorning: How about e.g. `MyCoolCompanyCops::Mailers::ExplicitRecipient`? It seems a bit arbitrary to decide that the first *two* namespaces are not part of the department name.

Even worse, how about `MyCoolCompany::Cop::Style::Alias`? Should we would allow people’s internal cops to conflict with our cops.

pirj: Also, as an experiment, I've moved one of the `rubocop-rspec` cops up the hierarchy, and removed its department altogether. Everything went quite smoothly.
I'm not aware if the department is mandatory for cops, but if it worked previously, I believe it should continue to work.

spec/rubocop/cop/badge_spec.rb
pirj: Or, the department should be detected as none, and the cop name as `Foo`.

lib/rubocop/cop/badge.rb
dmytro-savochkin: > Could we come up with a better name for this variable?

Actually we can get rid of this variable. I'm gonna update the PR.

> Actually, I wonder if what we are really checking is whether the class name is a subclass of
> RuboCop::Cop/RuboCop::Base? In that case, I’d suggest a more explicit namespace check:

Yes, if we are to change the code to the one proposed by @bquorning above, then we would still possibly have the same problem for cops `MyCoolCompany::Cop::Mailers::ExplicitRecipient` and `MyCoolCompany::Cop::Rails::Mailers::ExplicitRecipient`.

> It seems a bit arbitrary to decide that the first two namespaces are not part of the department name.

I do understand that it is arbitrary to decide the first two namespaces are not part of the department name but I don't see a better solution right now, and this particular solution I'm proposing fixes the problem for all the current cases I think.

> Even worse, how about MyCoolCompany::Cop::Style::Alias? Should we would allow people’s internal cops to 
> conflict with our cops.

That's a problem. But maybe a different one? Once and for all this can be fixed only if you put every part of this cop class name into its department/class name, something like `MyCoolCompany/Cop/Style/Alias`. It would require everyone to change every existing cop. I am not happy proposing such a huge change :) Can this be fixed another way?

spec/rubocop/cop/badge_spec.rb
dmytro-savochkin: Well, I think this is exactly how it was before. Previously method `for` was `new(*class_name.split('::').last(2))` so it would parse `RuboCop::Cop::Foo` as `Cop/Foo` (`Cop` department).

If I am to add something like this to `for` method
```
if class_name.start_with?('RuboCop::Cop', 'RuboCop::Base') && parts.length == 3
  raise
end
```
it would break a lot of specs. For example, we have spec/rubocop/cop/cop_spec.rb and in that file we have the following:
```
RSpec.describe RuboCop::Cop::Cop, :config do
  context 'with no submodule' do
    it('has right name') { expect(cop_class.cop_name).to eq('Cop/Cop') }
    it('has right department') { expect(cop_class.department).to eq(:Cop) }
  end
end
```

So... it seems it was intentional it doesn't raise an error? 

So I can add whatever thing you like here but maybe let's confirm it with the other core developers first since it would be a big change?

bbatsov:
@dmytro-savochkin Can you please rebase on the current `master`? It seems we're finally ready for 1.0!

spec/rubocop/cop/badge_spec.rb
bbatsov: I guess this was necessary during the period in which we had to support cops without a department (originally there were none and all cops were in the same namespace so to speak). I agree this should be changed.

dmytro-savochkin:
@bbatsov sure, rebased.

bbatsov:
Thanks!

ThomasKoppensteiner:
Puppetlabs archived the https://github.com/puppetlabs/rubocop-i18n repo. This means https://github.com/puppetlabs/rubocop-i18n/issues/45 will not be resolved.

How do we move on?
Does anyone have plans to fork & maintain the repo?

@dmytro-savochkin or @mvz maybe?

mvz:
I've thought about it since https://github.com/puppetlabs/rubocop-i18n/pull/46 wasn't getting merged.

Ideally, such a fork would live under the rubocop-hq organization. If the current owners are willing to grant access, maintainership of the gem could just be taken over. An alternative is to split the gem into a RailsI18n part and a GetText part, with new gem names. This may be a cleaner design since it would remove the need to activate one or the other i18n method.

Is there some better place to discuss this further?

bbatsov:
@mvz You can open a new issue about this. I'd be happy to host the old gem or an alternative solution under `rubocop-hq` and will add whoever wants to maintain it to the org.

mvz:
@bbatsov done!

PR #8491, opened 2020-08-08
Refactor if_unless_modifier_spec

mockdeep:
Related to rubocop-hq/rubocop-rspec#862

This adds a `spaces` helper method and simplifies the `other_cops` `let`
blocks. There's still more that could be done to make these tests a
little easier to maintain, but thought this was a decent first step.
It's also probably easier to discuss in discrete steps.

### DIFF
--- a/spec/rubocop/cop/style/if_unless_modifier_spec.rb
+++ b/spec/rubocop/cop/style/if_unless_modifier_spec.rb
@@ -76,7 +76,6 @@ RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
       end
 
       describe 'IgnoreCopDirectives' do
-        let(:comment) { '# rubocop:disable Style/For' }
         let(:body) { "puts '#{spaces(57)}'" }
 
         context 'and the long line is allowed because IgnoreCopDirectives is ' \
@@ -85,7 +84,7 @@ RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
             expect("#{body} if condition".length).to eq(77) # That's 79 including indentation.
             expect_no_offenses(<<~RUBY)
               def f
-                #{body} if condition #{comment}
+                #{body} if condition # rubocop:disable Style/For
               end
             RUBY
           end
@@ -104,7 +103,7 @@ RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
           it 'registers an offense' do
             expect_offense(<<~RUBY, body: body)
               def f
-                %{body} if condition #{comment}
+                %{body} if condition # rubocop:disable Style/For
                 _{body} ^^ Modifier form of `if` makes the line too long.
               end
             RUBY
@@ -113,7 +112,7 @@ RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
               def f
                 if condition
                   #{body}
-                end #{comment}
+                end # rubocop:disable Style/For
               end
             RUBY
           end
@@ -158,10 +157,10 @@ RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
   end
 
   context 'multiline if that fits on one line' do
-    let(:condition) { 'a' * 38 }
-    let(:body) { 'b' * 38 }
-
     it 'registers an offense' do
+      condition = 'a' * 38
+      body = 'b' * 38
+
       # This if statement fits exactly on one line if written as a
       # modifier.
       expect("#{body} if #{condition}".length).to eq(80)
@@ -313,9 +312,9 @@ RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
     let(:body) { 'b' * 36 }
 
     context 'when a multiline if fits on one line' do
-      let(:conditional) { "/#{'a' * 36}/" }
-
       it 'registers an offense' do
+        conditional = "/#{'a' * 36}/"
+
         expect("  #{body} if #{conditional}".length).to eq(80)
 
         expect_offense(<<-RUBY.strip_margin('|'))
@@ -330,9 +329,9 @@ RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
     end
 
     context "when a multiline if doesn't fit on one line" do
-      let(:conditional) { "/#{'a' * 37}/" }
-
       it 'accepts' do
+        conditional = "/#{'a' * 37}/"
+
         expect("  #{body} if #{conditional}".length).to eq(81)
 
         expect_no_offenses(<<-RUBY.strip_margin('|'))
@@ -525,9 +524,9 @@ RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
       let(:body) { 'bbb' }
 
       context 'it fits on one line' do
-        let(:condition) { 'aaa' }
-
         it 'registers an offense' do
+          condition = 'aaa'
+
           # This if statement fits exactly on one line if written as a
           # modifier.
           expect("#{body} if #{condition}".length).to eq(10)
@@ -546,11 +545,10 @@ RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
       end
 
       context "it doesn't fit on one line" do
-        let(:condition) { 'aaaa' }
-
         it "doesn't register an offense" do
-          # This if statement fits exactly on one line if written as a
-          # modifier.
+          condition = 'aaaa'
+
+          # This if statement fits exactly on one line if written as a modifier.
           expect("#{body} if #{condition}".length).to eq(11)
 
           expect_no_offenses(<<~RUBY)
--- a/spec/rubocop/cop/style/if_unless_modifier_spec.rb
+++ b/spec/rubocop/cop/style/if_unless_modifier_spec.rb
@@ -4,26 +4,24 @@
 # Note: most of these tests probably belong in the shared context "condition modifier cop"
 ######
 RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
-  let(:ignore_cop_directives) { true }
-  let(:allow_uri) { true }
-  let(:line_length_config) do
+  def spaces(count)
+    ' ' * count
+  end
+
+  let(:other_cops) do
     {
-      'Enabled' => true,
-      'Max' => 80,
-      'AllowURI' => allow_uri,
-      'IgnoreCopDirectives' => ignore_cop_directives,
-      'URISchemes' => %w[http https]
+      'Layout/LineLength' => {
+        'Max' => 80, 'IgnoreCopDirectives' => true, 'AllowURI' => true
+      }
     }
   end
-  let(:other_cops) { { 'Layout/LineLength' => line_length_config } }
 
   extra = ' Another good alternative is the usage of control flow `&&`/`||`.'
   it_behaves_like 'condition modifier cop', :if, extra
   it_behaves_like 'condition modifier cop', :unless, extra
 
   context 'modifier if that does not fit on one line' do
-    let(:spaces) { ' ' * 59 }
-    let(:body) { "puts '#{spaces}'" }
+    let(:body) { "puts '#{spaces(59)}'" }
     let(:source) { "#{body} if condition" }
     let(:long_url) do
       'https://some.example.com/with/a/rather?long&and=very&complicated=path'
@@ -44,7 +42,7 @@ RSpec.describe RuboCop::Cop::Style::IfUnlessModifier, :config do
           def f
             # Comment 1
             if condition
-              puts '#{spaces}'
+              puts '#{spaces(59)}'
             end # Comment 2
           end
         RUBY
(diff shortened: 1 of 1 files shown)


mockdeep:
Part of the problem I'm seeing is that so much of the Rubocop testing infrastructure depends implicitly on various `let` definitions, which mandates more `let` definitions. One good example is the `source` calls. It would be great to be able to pass an argument to it as a method:

```rb
def source(body_length)
  <<~RUBY
    x = if a
      #{'b' * body_length}
    end
  RUBY
end
```

But because `shared_contexts.rb` implicitly relies on `source` to not require any arguments, it has to pull in its dependencies less explicitly via other `let` blocks.

mockdeep:
Added a second commit with some additional changes.

spec/rubocop/cop/style/if_unless_modifier_spec.rb
```diff
@@ -44,7 +42,7 @@ def f
           def f
             # Comment 1
             if condition
-              puts '#{spaces}'
+              puts '#{spaces(59)}'
```
pirj, line 45: I've changed this to `#{body}` in my PR, since in the other examples and more importantly, the `expect_offense` statement above it's used like that.

spec/rubocop/cop/style/if_unless_modifier_spec.rb
```diff
@@ -583,7 +584,7 @@ def f
   end
 
   context 'when Layout/LineLength is disabled' do
-    let(:line_length_config) { { 'Enabled' => false } }
+    let(:other_cops) { {} }
```
pirj, line 587: How does this work? Isn't `LineLength` enabled by default in `config/default.yml`? If we override the `other_cops` definition in the beginning of this spec, I would expect RuboCop to use that default, and that contradicts to context title.

marcandre:
FWIW, I prefer this to #8447.

It still does not seem like an improvement to me, on the contrary.

It seems clear that these changes there not to actually make the code better, not because there's an actual issue with any of these `let`s, but simply to defeat a misguided metric.

I'll take just one example despite it being one of the most innocuous:

```ruby
context "when a multiline if doesn't fit on one line" do
  let(:conditional) { "/#{'a' * 37}/" }  # <<< Changing this line...
  it 'accepts' do
    conditional = "/#{'a' * 37}/"        # <<< ... into this line
    expect("  #{body} if #{conditional}".length).to eq(81)
    expect_no_offenses # ...
    # ...
```

I believe this makes no meaningful improvement. The most important metric for me is ease of comprehension. It's basically the same idea, saying that `conditional` is, for this spec `"/aaaa.../"`. Not sure why that string btw, `'a' * 39` is simple better and clearer.

As for *practical* differences, there are a few, all against the local var.

I'll remark the first `expect`: this line does not belong in the `it 'accepts'` block but as a separate `before` block of the `context`; it's not actually part of spec, but a sanity check on the setup:

```ruby
context "when a multiline if doesn't fit on one line" do
  let(:conditional) { 'a' * 39 }
  before { expect("  #{body} if #{conditional}".length).to eq(81) } # sanity check
  it 'accepts' do
    expect_no_offenses # ...
```

I believe that *that* is the most expressive version of the code. `context` => setup (and sanity check) for that context => spec.

Further refactoring / testing might also require adding a `let(:replacement) { "  #{body} if #{conditional}" }` that would be reused in a `it 'registers an offense'`, etc. There are actually quite a number of such tests here and there, a refactoring of that could be a net gain.

Adding any spec for the same conditional: do you duplicate the local var? Do you put it outside of the block? Why?

I remain strongly convinced that counting the number of `let` is a misguided metric. It pains me to think of Rubyists confronted to such an offense (say because their manager believes in metrics / rubocop) and having to waste their time into going around that offense when nothing is wrong with any of their `let`.

"When a measure becomes a target, it ceases to be a good measure." -- Goodhart's law

In this case, it's not even a good measure to start with. Just the fact that this metric does not count `def`s (or local variables) should be a hint of that.

```ruby
- let(:conditional) { 'a' * 39 }   # Oops, too many `let`
+ def conditional; 'a' * 39; end  # yay, no offense!
```

spec/rubocop/cop/style/if_unless_modifier_spec.rb
mockdeep: Oh, interesting. I was under a different understanding, but you made me check my assumptions. Turns I was (mostly) correct! All cops are `'Enabled' => true` by default in the tests, including the disabled ones in `config/default.yml`. But they have **no** configurations:

```irb
> config.for_cop('Layout/LineLength')
 => {"Enabled"=>true}
> config.for_cop('Bundler/GemComment')
 => {"Enabled"=>true}
```

So without a `'Max'` setting, `Layout/LineLength` is effectively disabled.

mockdeep:
```rb
context "when a multiline if doesn't fit on one line" do
  let(:conditional) { 'a' * 39 }
  before { expect("  #{body} if #{conditional}".length).to eq(81) } # sanity check
  it 'accepts' do
    expect_no_offenses # ...
```

It's hard to argue in a way that doesn't dip into the realm of subjectivity, but I think this creates a lot of visual noise. If there's only one example, I'd actually go so far as to eliminate the `context`:

```rb
it 'accepts when a multiline if does not fit on one line' do
  conditional = 'a' * 39
  expect("  #{body} if #{conditional}".length).to eq(81)

  expect_no_offenses...
end
```

I agree that the sanity check deserves to be somehow separated, but I think in terms of simplicity I would take this tradeoff for the sake of less complexity in the test setup. It also makes me think maybe there's a method that should be extracted, like `body = make_body(length: 81)` or the like. Some of this is made more difficult by the fact that the existing test helpers depend implicitly on things like `source`.

> Adding any spec for the same conditional: do you duplicate the local var? Do you put it outside of the block? Why?

Copying from [my other comment](https://github.com/rubocop-hq/rubocop/pull/8447#discussion_r467538132): Personally, I'd probably actually copy paste it into each test. I think some duplication can be advantageous across tests, since it makes it a lot easier to write varying test cases, and it makes each test more self-contained. That, to me, is easier to maintain and understand without scrolling up and down the file (and other files) to trace through all of the implicit dependencies. The test itself becomes more "plain ruby" without a lot of indirection and uncertainty about what happens when.

With scattered `let` blocks all over the test file, I have to mentally piece together the existing semi-global state that I'm working with in my test, and what ways I need to modify it to fit my needs. I often need to introduce a new `context` just to override some `let` somewhere else.

```rb
- let(:conditional) { 'a' * 39 }   # Oops, too many `let`
+ def conditional; 'a' * 39; end  # yay, no offense!
```

While I agree in principle that simply replacing `let` with `def` doesn't solve the core issue in question, I think it opens some doors. To that extent, I'm actually happy with a sweep to replace all `let`s with `def`s. When we need to vary the results from one test to the next, the solutions are very different. With a `let` we have to add a chunk of additional structure to each test. Probably a new `context` and new `let`, or maybe do some in-line modification inside the test block. For methods we can accept a parameter. For this case, we do the same "letter * number" logic all over the test file, which to me makes the test less expressive. I might instead have a method `conditional(length:)`, or even `text(length:)`, since we do the same thing for several other cases.

Using methods in these cases also prompts me to think more deeply about the test structure as a whole. I don't add a new method as lightly as I add a new `let`, so I'm going to think twice about what is the best structure for the file. And if I'm writing a helper like `text(length:)`, maybe it's a thing that would apply to more test file, so could be made available everywhere.

marcandre:
> It's hard to argue in a way that doesn't dip into the realm of subjectivity

I'm glad we both agree on this. As a tough reviewer, I wouldn't ask a submitter of either of these two variants to change it.

The problem is that I now might have my computer completely disregards this subjectivity and tell me "no, one way is baaaaad, only the other is good", when I'm trying to actually fix real issues. It's the kind of thing that makes me want to curse, even though I very rarely loose my cool in real life.

mockdeep:
A couple more thoughts on this. Yes, it's trivial to satisfy the rule without actually addressing the underlying issue it is flagging. In this way, it's similar to other rules like `ClassLength`, pretty much all of the `Metrics` cops, and probably many others. You can just grab a bag of methods and toss it into a module and `include` it back into your class and `ClassLength` will be satisfied. These sorts of rules aren't any good unless you use them in good faith.

Thinking about things from a higher level, here is a non exhaustive list of valuable qualities in a test file:

* It needs to verify that the code actually works, of course.
* It should be easy to maintain and extend.
* It should provide a good source of reference for how the code works.

I don't think that this test file as its written does a great job on any of these. You could argue that it satisfies the first, but it's hard to be sure when there is so much indirection in the test file. Likewise, it's hard to extend and maintain, since any change I want to make has to account for all of the pre-existing structures and state in the file. And similarly, it's hard for it to be a useful source of reference when you need to mentally knit together an understanding from nested context descriptions and logic distributed across the file.

This isn't specific to the use of `let` blocks, but more generally about the way that we tend to write tests in RSpec. I think we try to apply the same ideas of abstraction that are good for our application logic, and I think that actually makes the tests worse. DRY in particular is a principle that really gets abused in tests.

marcandre:
I pretty much agree with all of the above.

What is less clear to me is that the number of `let` is in direct opposition of your stated goals. What I mean is that indeed, if there are 10 independent `let` in a spec suite, overridden here and there, that can be quite hellish. But there are valid cases, imo, where a `let` doesn't cost much if at all, like a temporary intermediate variable. Maybe the number of *overridden* `let`s should be considered?

```ruby
let(:source) { '2 + 2' } # overridden in multiple places, counts as +1
let(:processed_source) { ... } # never overridden, counts as +0
let(:ast) { processed_source.ast } # never overridden, counts as +0
subject(:node) { ast } # overridden in some places, counts as +1
```

Is the code really more complex because the `processed_source` intermediate step was named?

Here's another way to think about this. `ClassLength` offense asks the question: can you break the functionality into separate independent and coherent modules? That's a valid question and the end result, if done in good faith, will be better.
`AbcSize`/`...Complexity` offense asks the question: can you break your method into smaller chunks that make sense. Again, net gain.
This `let` metric, what is the right question? As it stands, it asks "can you have less named things (and this is only considering one particular way of naming things)". It's like an `AbcSize` metric, but just for `A`, and for the whole class, and only counting a specific set of `A`. I typically see no way to obtain noticeably better result, and if two very knowledgeable devs like yourself and @pirj can't achieve it (at least in my opinion)...

Thanks for bearing with me up to now 😅

mockdeep:
So a couple of clarifications in the way that the rule behaves. For starters, it does not count *all* of the `let` blocks in the test file. So if you have 6 parallel contexts each with a single `let`, it will not report an issue. It reports an issue when a single context has access to more than 5 `let` blocks, which sounds to me like a reasonable limit. It does not account for overrides, but to me overrides are an anti-pattern best to be avoided. In my experience, they cause a lot of confusion in the structure of tests.

Also, the rule by default does not count `subject` blocks against you, so I don't think your example would be likely to cause any complaints. Likewise, this test file actually might not have as many complaints as you might expect looking at it. In concert with the `NestedGroups` cop, though, it could help guide a better structure for the file.

> Here's another way to think about this. `ClassLength` offense asks the question: can you break the functionality into separate independent and coherent modules?

> This `let` metric, what is the right question? As it stands, it asks "can you have less named things (and this is only considering one particular way of naming things)".

The way this is framed casts a negative light on the new `let` rule, but you can frame `ClassLength` in the same way: "can you have less lines of code in this file?". It isn't really a measure of code quality, just a signifier that maybe some additional thinking should be done. You can satisfy the rule by pulling an arbitrary bag of methods into a module. It just sub-divides the mess without really solving the underlying architectural problem. In the same way, if you don't engage with the `let` cop in good faith, you can just use method definitions instead and satisfy the rule. The question you ask yourself for this rule should be something more like "can I refactor my test to require less indirection?". Maybe the solution is moving some logic to factories or external helpers, maybe it is to collapse `let` code together when it is only referred to from another `let`, maybe it is to move it directly into the test, maybe it is to define a method with a parameter that allows you to more easily create variations.

Ultimately, none of these rules can enforce good quality. The best we can hope for is that they prompt us to think more about the way we structure our code. Tests in particular often feel a bit like the wild west, varying greatly in structure and style.

marcandre:
>  In the same way, if you don't engage with the let cop in good faith, you can just use method definitions instead and satisfy the rule.

Sorry if I wasn't clear, my entire last remark was about someone engaging the rules in good faith.

> "can I refactor my test to require less indirection"

This is the case only if a `let` is "far" from it's point of use, or is redefined somewhere "far". The number of `let` doesn't seem like a good proxy for this.

> but to me overrides are an anti-pattern best to be avoided

We'll have to agree that we *strongly* disagree on this one. I wish that disagreement was enough for this Cop to be disabled by default. I won't force anyone to use multiple `let`, but I only wish no-one forcing me to avoid using them, and that includes when hacking on `rubocop` itself.

bbatsov:
So, what will it be? I've been going over the list of PRs today and it's time to either merge this or close it. I'm leaning towards closing this suggestion as I don't think it's a clear improvement over the existing code, but I'll leave the final decision to @marcandre.

marcandre:
I remain unconvinced about the proposed refactors, and the rule in general, so closing this.

PR #8492, opened 2020-08-08
remove `let(:source)` from `shared_contexts.rb`

mockdeep:
Poking around in the tests, I noticed that we were parsing this source
for every test, even in cases where we never use it, which is most of
them. This removes the default `let(:source)` and stops automatically
calling `begin_investigation` on the cop. Instead it calls it in places
where it is necessary.

I benchmarked it locally, and while not a huge win, it consistently cut
about 2-3 seconds off my build times, from ~52 seconds to ~50 seconds.

-----------------

* [ ] Added tests.


bbatsov:
I think that's a nice improvement. Thanks!

PR #8501, opened 2020-08-09
[Fixes #8278] Tweak `@api private`.

### LINKED ISSUE #8278: Should we mark some modules as private before v1?

marcandre:
Many internal Modules and Classes marked as private.
Severity marked as public


marcandre:
Unfinished though...

marcandre:
I split the PR in two commits. The second one adds `@api private` to all cops and formatters. I though of writing a cop but was lazy and did it with search and replace. I hope I didn't miss too many things with the first one.

I don't know if anyone wants to review this? or maybe we should merge and someone can review the result?

lib/rubocop/cli/command/auto_genenerate_config.rb
```diff
@@ -4,6 +4,9 @@ module RuboCop
   class CLI
     module Command
       # Generate a configuration file acting as a TODO list.
+      # @api private
```
bbatsov, line 7: I guess the duplicate is an oversight, right?

bbatsov:
> The second one adds @api private to all cops and formatters.

Isn't this an overkill? It seems to me that such things are private by definition. But I guess it's still needed to generate the yard docs correctly.

lib/rubocop/cli/command/execute_runner.rb
```diff
@@ -4,6 +4,9 @@ module RuboCop
   class CLI
     module Command
       # Run all the selected cops and report the result.
+      # @api private
+      #
+      # @api private
```
bbatsov, line 9: Same here.

lib/rubocop/cli/command/auto_genenerate_config.rb
marcandre: Oops, search and replace error...

marcandre:
> > The second one adds @api private to all cops and formatters.
> 
> Isn't this an overkill? It seems to me that such things are private by definition. But I guess it's still needed to generate the yard docs correctly.

I don't know what's best tbh. Note that we share constants from some Cops to other Cops (e.g. `AbcSizeCalculator` uses `CyclomaticComplexity::COUNTED_NODES`), so it's not that clear. Do I need to repeat that some of these constants are Sets, others Arrays, and that changing the type is a potential source of breakage unless something is done about that?

I'm completely neutral about the 2nd commit, and of course even the multiple `@api private` of the first commit won't necessarily change all that much, but shouldn't hurt much either...

Let me know how we should proceed.

marcandre:
Added a small 3rd commit for some constants that can be made private. I'm just starting to realize that normal constants should probably be private by default...

bbatsov:
Yeah, I was also thinking that constants should probably be private for real. It's common sense, but I've rarely seen Rubyists do this, as public by default has always been the mantra in our community. There are some merge conflicts that need to be addressed and you can merge this. I don't think any of the changes can hurt.

PR #8510, opened 2020-08-10
Parsed Regexp utilities

marcandre:
We are starting to have more Cops handling `Regexp`. This PR
* simplifies the access to the regexp parsed tree with `RegexpNode#parsed_tree`
* caches to avoid reparsing the same node multiple times
* adds `RegexpNode#each_capture` which simplifies processing.

This could be helpful for #8504 too.

I'm hesitating between adding it to `rubocop-ast` or extending it as in this PR.


bbatsov:
Won't we end up duplicating functionality from https://github.com/ammar/regexp_parser ?

marcandre:
Right, I thought of making a PR for the gem itself; there are a few things that are annoying with the current API, etc.

In any case, `RegexpNode#each_capture` seems like a nice addition; maybe in the future it can be refactored to delegate the call to the gem?

bbatsov:
> In any case, RegexpNode#each_capture seems like a nice addition

I agree. 

> maybe in the future it can be refactored to delegate the call to the gem?

Yeah, that sounds reasonable.

PR #8511, opened 2020-08-10
Fix a false positive for `Style/CaseLikeIf` when conditional contains comparison with a class

### LINKED ISSUE #8508: Bad autocorrect for Style/CaseLikeIf
In rubocop 0.89.0

If you have code like this:

```ruby
if klass == ClassA
  do_something
elsif klass == ClassB
  do_something_else
end
```

It incorrectly gets autocorrected to:

```ruby
 case klass
 when ClassA
   do_something
 when ClassB
   do_something_else
 end
```

@marcandre warned about this here https://github.com/rubocop-hq/rubocop/issues/7736#issuecomment-634915216 but I think this advice went unheeded.

fatkodima:
Closes #8508

### DIFF
--- a/lib/rubocop/cop/style/case_like_if.rb
+++ b/lib/rubocop/cop/style/case_like_if.rb
@@ -159,9 +159,10 @@ module RuboCop
           case node.method_name
           when :is_a?
             node.arguments.first if node.receiver == target
-          when :==, :eql?, :equal?, :=~, :match, :match?
-            lhs, _method, rhs = *node
-            condition_from_binary_op(lhs, rhs, target)
+          when :==, :eql?, :equal?
+            condition_from_equality_node(node, target)
+          when :=~, :match, :match?
+            condition_from_match_node(node, target)
           when :===
             lhs, _method, rhs = *node
             lhs if rhs == target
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/style/case_like_if.rb
```diff
@@ -202,6 +214,16 @@ def const_reference?(node)
             name == name.upcase
         end
 
+        def class_reference?(node)
+          return false unless node.const_type?
+
+          name = node.children[1].to_s
```
marcandre, line 220: How about using `node.children[1].match? /[[:lower:]]/` instead?

fatkodima: The original code checks that const name is not all uppercase, so the code could be written as `!name.match?(/^[[:upper:]]+$/)`, but this does not handle numbers in names and so, the original code looks simpler to me.

fatkodima:
Marked it as unsafe.

lib/rubocop/cop/style/case_like_if.rb
marcandre: > checks that const name is not all uppercase

Right. I'm sorry though, I don't see how "not all uppercase" != "has a lowercase", could you provide an example?

fatkodima: You're right, sorry. I mistakenly matched that regexp only with the first letter 😄

fatkodima: Updated.

marcandre:
I'll merge after the merge conflict is resolved... 😢

marcandre:
Thanks!

PR #8513, opened 2020-08-10
Lint/ShadowingOuterLocalVariable: clarify the ruby warning [doc]

chocolateboy:
Make it clear that ruby [no longer warns](https://docs.ruby-lang.org/en/2.6.0/NEWS.html#label-Language+changes) about variable shadowing:

> The “shadowing outer local variable” warning is removed. [Feature [#12490](https://bugs.ruby-lang.org/issues/12490)]


marcandre:
Thank you!

PR #8516, opened 2020-08-10
Add new `Style/RedundantSelfAssignment` cop

### LINKED ISSUE #8451: Cop idea: redudant self assignment
Since methods like `Array#concat` returns `self`:

```ruby
# bad
args = args.concat(ary)
self.foo = foo.concat(ary)

# good
args.concat(foo) # or
args += foo
foo.concat(ary) # or
self.foo += ary
```

Noticed [in our code](https://github.com/rubocop-hq/rubocop/blob/4943d5005b44c61973910b77adbb5fa42209bbfd/lib/rubocop/cop/mixin/check_line_breakable.rb#L174)

fatkodima:
Closes #8451


marcandre:
Thanks for this cop ♥️

docs/modules/ROOT/pages/cops_style.adoc
```diff
@@ -7899,6 +7899,46 @@ end
...
+
+[source,ruby]
+----
+# bad
+args = args.concat(ary)
+hash = hash.merge!(other)
+
+# good
+args.concat(foo)
+args += foo
```
bbatsov, line 7931: I think you forgot the bad example for this.

fatkodima: This and the previous line are both good examples for `args = args.concat(ary)` (couple of lines above)

bbatsov: Ah, my bad!

PR #8517, opened 2020-08-10
Handle to_h with block in Style/HashTransformKeys and Style/HashTransformValues

eugeneius:
Followup to https://github.com/rubocop-hq/rubocop/pull/7663.

Since Ruby 2.6, `Hash#to_h` accepts a block which is called with each pair and must return a key and value to include in the new hash: https://rubyreferences.github.io/rubychanges/2.6.html#to_h-with-a-block

If either the key or the value are always returned unchanged, using `Hash#transform_keys` or `Hash#transform_values` is faster and clearer.

This is similar to the existing logic that detects `map { ... }.to_h`.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/hash_transform_values_spec.rb
+++ b/spec/rubocop/cop/style/hash_transform_values_spec.rb
@@ -133,4 +133,25 @@ RSpec.describe RuboCop::Cop::Style::HashTransformValues, :config do
       {a: 1, b: 2}.transform_values {|v| foo(v)}.to_h {|k, v| [v, k]}
     RUBY
   end
+
+  context 'when using Ruby 2.6 or newer', :ruby26 do
+    it 'flags _.to_h{...} when transform_values could be used' do
+      expect_offense(<<~RUBY)
+        x.to_h {|k, v| [k, foo(v)]}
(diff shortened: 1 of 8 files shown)


spec/rubocop/cop/style/hash_transform_values_spec.rb
```diff
@@ -81,6 +81,17 @@
     RUBY
   end
 
+  it 'flags _.to_h {...} when transform_values could be used' do
```
koic, line 84: I think it should necessary to detect it only for Ruby 2.6 or higher. It should not detected in Ruby 2.5 or lower.

```ruby
% ruby -v
ruby 2.5.8p224 (2020-03-31 revision 67882) [x86_64-darwin17]

% ruby -e "p({a: 1, b: 2, c: 3}.to_h { |k, v| [k, -v] })"
{:a=>1, :b=>2, :c=>3}

% ruby -e "p({a: 1, b: 2, c: 3}.transform_values { |v| -v })"
{:a=>-1, :b=>-2, :c=>-3}
```

eugeneius: Fixed, thanks @koic! 🙇

koic:
Good catch! Thanks @eugeneius!

PR #8523, opened 2020-08-11
[Fix #8499] Fix Style/IfUnlessModifier to prevent offense when first-line comment and code after end block

### LINKED ISSUE #8499: Wrong correction in Style/IfUnlessModifier with first-line comment and code after the end block on the same line
I think `Style/IfUnlessModifier` doesn't correctly handle the case when there is a first-line comment (right after the `if` keyword) AND some code right after the `end` keyword. If you correct this code with `rubocop -a` it would put the comment right after the `end` keyword, before the code that was there previously. This basically makes the code invalid after correction.

This probably can be better seen in the example below.

I think in such a case we should stop auto-correction because it's pretty non-deterministic: where would that comment should go? Did it belong to the if-end condition? In this case it should go right after it, before the next code. But that's impossible in a single-line modifier form, unless we use these `/* comments */` but we probably wouldn't want to do this, so better not to auto-correct.

I can provide a PR if it seems right.

--------

We have file `test.rb` with the following code:

```
[
  1, if foo # this is due to ...
       2
     end, 3
]
```

We run `rubocop test.rb --only Style/IfUnlessModifier -a`.

## Actual behavior

```
[
  1, (2 if foo) # this is due to ..., 3
]
```

## Expected behavior

Everything stays the same and no offense is given (I think it's the right way):
```
[
  1, if foo # this is due to ...
       2
     end, 3
]
```
or somehow it's fixed to
```
[
  1, (2 if foo), 3 # this is due to ...
]
```
but in this case the comment is in the wrong place, or it's fixed to
```
[
  1, (2 if foo) /* # this is due to ... */, 3 
]
```
but we probably don't want to use this syntax for comments.

## RuboCop version

0.89.0 (using Parser 2.7.1.4, rubocop-ast 0.2.0, running on ruby 2.5.1 x86_64-darwin18)

dmytro-savochkin:
Fixes #8499.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#8499](https://github.com/rubocop-hq/rubocop/issues/8499): Fix `Style/IfUnlessModifier` and `Style/WhileUntilModifier` to prevent an offense if there are both first-line comment and code after `end` block. ([@dsavochkin][])
+
 ## 1.2.0 (2020-11-05)
 
 ### New features
--- a/docs/modules/ROOT/pages/cops_style.adoc
+++ b/docs/modules/ROOT/pages/cops_style.adoc
@@ -4485,14 +4485,18 @@ unless qux.empty?
   Foo.do_something
 end
 
-do_something_in_a_method_with_a_long_name(arg) if long_condition
+do_something_with_a_long_name(arg) if long_condition_that_prevents_code_fit_on_single_line
 
 # good
 do_stuff(bar) if condition
 Foo.do_something unless qux.empty?
 
-if long_condition
-  do_something_in_a_method_with_a_long_name(arg)
+if long_condition_that_prevents_code_fit_on_single_line
+  do_something_with_a_long_name(arg)
+end
+
+if short_condition # a long comment that makes it too long if it were just a single line
+  do_something
 end
 ----
 
@@ -11239,6 +11243,17 @@ end
 
 # good
 x += 1 until x > 10
+----
+
+[source,ruby]
+----
+# bad
+x += 100 while x < 500 # a long comment that makes code too long if it were a single line
+
+# good
+while x < 500 # a long comment that makes code too long if it were a single line
+  x += 100
+end
 ----
 
 === References
--- a/lib/rubocop/cop/mixin/statement_modifier.rb
+++ b/lib/rubocop/cop/mixin/statement_modifier.rb
@@ -19,7 +19,8 @@ module RuboCop
       def non_eligible_node?(node)
         node.modifier_form? ||
           node.nonempty_line_count > 3 ||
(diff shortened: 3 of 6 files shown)


marcandre:
I'd love to approve this PR, but there are many issues with `IfUnlessModifier` and `WhileUntilModifier`; ideally resolving this issue would be part of a bigger refactor. If you are not willing to tackle that, than the minimum acceptable is to add a corresponding test to `WhileUntilModifier`; for example the `first_line_comment` method in the module you are tweaking is redefined for `IfUnless` but not `WhileUntil`... I wish I had to time to look at all this myself.

dmytro-savochkin:
> I'd love to approve this PR, but there are many issues with `IfUnlessModifier` and `WhileUntilModifier`; ideally resolving this issue would be part of a bigger refactor. If you are not willing to tackle that, than the minimum acceptable is to add a corresponding test to `WhileUntilModifier`; for example the `first_line_comment` method in the module you are tweaking is redefined for `IfUnless` but not `WhileUntil`... I wish I had to time to look at all this myself.

Well, actually I was thinking about handling `WhileUntilModifier` too. But later, not in this ticket.

And to be honest, I don't really think `IfUnlessModifier` and `WhileUntilModifier` are that much similar. They differ in the fact that `if/unless` constructs return something meaningful while `while/until` always return nil. Because of that people use `if` in assignment, in arrays, in hashes, etc, while no one usually uses `while` like that
```
[
  1, while foo # this is due to ...
       do_something
     end, 3
]
```

because what for? It would just return nil, no matter what happens inside the loop.

But still, I agree that something must be done with `WhileUntilModifier`, but definitely not in this PR. That's why there is a separate issue for that problem: https://github.com/rubocop-hq/rubocop/issues/8281.

marcandre:
> It would just return nil, no matter what happens inside the loop.

That is a good point, but not completely accurate... `while true; break 42; end` returns 42.


Sorry if I wasn't clear: this PR is probably fine as is, but without a corresponding test in `WhileUntilModifier`, it's difficult for me to see the effect of the change on the module it shares, especially since they don't share the implementation of `first_line_comment` (which seems like a code smell).

dmytro-savochkin:
> That is a good point, but not completely accurate... `while true; break 42; end` returns 42.
> 
> Sorry if I wasn't clear: this PR is probably fine as is, but without a corresponding test in `WhileUntilModifier`, it's difficult for me to see the effect of the change on the module it shares, especially since they don't share the implementation of `first_line_comment` (which seems like a code smell).

Ah, got it now. Updating the code.

dmytro-savochkin:
@marcandre 
Is there any chance this PR can be approved? :)

bbatsov:
@dmytro-savochkin I'm fine with the proposed change. It'd be nice if you updated the cop examples, so it's clear to users we're taking the comments into account.

dmytro-savochkin:
@bbatsov, updated the examples.

bbatsov:
Thanks!

PR #8526, opened 2020-08-12
Fix a false positive for `Style/CaseEquality` cop

koic:
Follow https://github.com/rubocop-hq/rubocop/pull/8322#discussion_r468443896

Fix a false positive for `Style/CaseEquality` cop when the receiver is not a camel cased constant.

The following constant should not be replaced with `is_a?`.
What kind of object is actually assigned depends on the constant. e.g.:

```ruby
REGEXP_CONSTANT === something #=> does not register an offense.
```

This PR will only consider camel-cased constant as class name. e.g.:

```ruby
Array === something #=> auto-correct to `something.is_a?(Array)`
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/case_equality.rb
+++ b/lib/rubocop/cop/style/case_equality.rb
@@ -39,6 +39,8 @@ module RuboCop
 
         def on_send(node)
           case_equality?(node) do |lhs, rhs|
+            return if lhs.const_type? && !lhs.module_name?
+
             add_offense(node.loc.selector) do |corrector|
               replacement = replacement(lhs, rhs)
               corrector.replace(node, replacement) if replacement
--- a/rubocop.gemspec
+++ b/rubocop.gemspec
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('rainbow', '>= 2.2.2', '< 4.0')
   s.add_runtime_dependency('regexp_parser', '>= 1.7')
   s.add_runtime_dependency('rexml')
-  s.add_runtime_dependency('rubocop-ast', '>= 0.3.0', '< 1.0')
+  s.add_runtime_dependency('rubocop-ast', '>= 0.4.0', '< 1.0')
   s.add_runtime_dependency('ruby-progressbar', '~> 1.7')
   s.add_runtime_dependency('unicode-display_width', '>= 1.4.0', '< 2.0')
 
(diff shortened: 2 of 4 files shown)


lib/rubocop/cop/style/case_equality.rb
```diff
@@ -33,11 +33,14 @@ class CaseEquality < Base
         extend AutoCorrector
 
         MSG = 'Avoid the use of the case equality operator `===`.'
+        CAMEL_CASE = /[A-Z]+[a-z]+/.freeze
```
knu, line 36: I think this pattern should be more specific, otherwise `SomeNamespace::REGXP` would pass when it shouldn't.  I'd suggest something like `\A(?=.*[a-z])(?:[A-Z][a-z]*|::)+\z` instead, but I'm not sure  if that's good enough.

koic: This regular expression is based on the following: 
https://github.com/rubocop-hq/rubocop/blob/v0.89.1/lib/rubocop/cop/style/inverse_methods.rb#L45

But the suggested regexp seems more better. I updated this PR. Thank you.

knu: Considering that the last component of a constant name is all that matters, this pattern could work better: `(\A|::)(?=.*[a-z])([A-Z][a-z0-9]*)+\z`. (I also took digits into account)

In any case, class names like `URI::HTTP` and `Addressable::URI` cannot be handled nicely unless you keep a known class name list.

marcandre: Just do `lhs.children.last.match? /[[:lower:]]/`. The last child is the name of the constant without its scope.

marcandre: Maybe I should introduce a `ConstNode` and provide `class_const?`

marcandre: How about this... rubocop-hq/rubocop-ast#99

koic: @marcandre Thank you for creating the new API. I'd like to replace it with `class_name?` when new RuboCop AST is released.

marcandre: @koic I added it as `module_name?` and aliased it to `class_name?` 👍

koic: I've updated this PR with `module_name?`. This PR requires the next RuboCop AST release (0.4.0?) due to its dependency on the unreleased API.

rubocop.gemspec
```diff
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('rainbow', '>= 2.2.2', '< 4.0')
   s.add_runtime_dependency('regexp_parser', '>= 1.7')
   s.add_runtime_dependency('rexml')
-  s.add_runtime_dependency('rubocop-ast', '>= 0.3.0', '< 1.0')
+  s.add_runtime_dependency('rubocop-ast', '>= 0.4.0', '< 1.0')
```
koic, line 41: RuboCop AST 0.4.0 has been released and I updated this one.
https://rubygems.org/gems/rubocop-ast/versions/0.4.0

PR #8527, opened 2020-08-12
Prevent an incorrect auto-correction for `Style/CaseEquality` cop

koic:
Follow https://github.com/rubocop-hq/rubocop/pull/8322#discussion_r468672970

This PR fixes an incorrect auto-correction for `Style/CaseEquality` cop when comparing with `===` against a regular expression receiver.

This PR auto-correct `===` with `match?` to keep compatibility of returning a boolean value.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/case_equality.rb
+++ b/lib/rubocop/cop/style/case_equality.rb
@@ -14,7 +14,7 @@ module RuboCop
       #   # good
       #   something.is_a?(Array)
       #   (1..100).include?(7)
-      #   some_string =~ /something/
+      #   /something/.match?(some_string)
       #
       # @example AllowOnConstant
       #   # Style/CaseEquality:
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/style/case_equality.rb
```diff
@@ -58,7 +58,7 @@ def const?(node)
         def replacement(lhs, rhs)
           case lhs.type
           when :regexp
-            "#{rhs.source} =~ #{lhs.source}"
+            "#{lhs.source}.match?(#{rhs.source})"
```
knu, line 61: Before doing this, you need to check if the subsequent code does not refer to the match data (`Regexp.last_match`, `$~`, `$1`, etc.).

koic: Sure! I have a idea. `Performance/Regexp` cop detects `Regexp.last_match`, and etc. It is expected that `a === b` will be auto-corrected to `a.match?(b)` in future.
Reelated issue: https://github.com/rubocop-hq/rubocop-performance/issues/152

It currently doesn't support unsafe auto-correction, so it may be better to delegate it to `Perfomance/Regexp` cop.

PR #8529, opened 2020-08-12
Mark `Lint/FrozenStringLiteralComment` as `Safe`, but with unsafe auto-correction.

marcandre:
@bbatsov [wrote](https://github.com/rubocop-hq/rubocop/pull/7307#discussion_r463971537):
> I think the cop is actually unsafe (as it doesn't actually check the source to figure out if it's ok to add the magic comment). Unsafe for a cop means that it yields false positives or the suggestions can change the meaning of the code. Unsafe for auto-correct means the cop the cop suggestions are safe, but the auto-correct can't apply those reliably.

I would propose a different definition.  Unsafe for a cop means that it yields false positives *and that's it*. (e.g. "don't use `dig` with single argument" is unsafe, as it could be a custom method in a gem dependency and thus perfectly o.k. and without another alternative). The `Lint/FrozenStringLiteral` cop is a good example where the fact that we can't reliably auto-correct a source doesn't mean it can't or shouldn't be done. Since *any source* can modified to have a frozen string literal comment, that all sources *should* be modified, and that the cop will always correctly detect if there is a frozen string literal comment, I believe this cop is safe, only its auto-correction is potentially unsafe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * [#8362](https://github.com/rubocop-hq/rubocop/issues/8362): Add numbers of correctable offenses to summary. ([@nguyenquangminh0711][])
 * [#8513](https://github.com/rubocop-hq/rubocop/pull/8513): Clarify the ruby warning mentioned in the `Lint/ShadowingOuterLocalVariable` documentation. ([@chocolateboy][])
 * [#8517](https://github.com/rubocop-hq/rubocop/pull/8517): Make `Style/HashTransformKeys` and `Style/HashTransformValues` aware of `to_h` with block. ([@eugeneius][])
+* [#8529](https://github.com/rubocop-hq/rubocop/pull/8529): Mark `Lint/FrozenStringLiteralComment` as `Safe`, but with unsafe auto-correction. ([@marcandre][])
 
 ## 0.89.1 (2020-08-10)
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -3033,7 +3033,7 @@ Style/FrozenStringLiteralComment:
     # `never` will enforce that the frozen string literal comment does not
     # exist in a file.
     - never
-  Safe: false
+  SafeAutoCorrect: false
 
 Style/GlobalStdStream:
   Description: 'Enforces the use of `$stdout/$stderr/$stdin` instead of `STDOUT/STDERR/STDIN`.'
--- a/docs/modules/ROOT/pages/cops_style.adoc
+++ b/docs/modules/ROOT/pages/cops_style.adoc
@@ -3259,7 +3259,7 @@ format('%s', 'Hello')
 | Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
 
 | Enabled
-| No
+| Yes
 | Yes (Unsafe)
 | 0.36
 | 0.79


koic:
First of all thank you for opening a good question. In conclusion, I think this cop will be `Safe: false`.

As far as I understand, the following could be `Safe: false`. For example...

- Not safe if false positives is not avoidable because the type of instance assigned to variable is unknown
- Not safe if (potentially) incompatible behaviours are proposed

I think that `Lint/FrozenStringLiteralComment` cop is unsafe (`Safe: false`) in the latter respect.

On the other hand, it is about unsafe auto-correction (e.g.: `Safe: true` and `SafeAutoCorrect: false`) .
For example, replacing `obj =~ /\Afoo/` with `obj.start_with?('foo')` is unsafe. Because nil error occurs if `obj` is `nil`. So, I think it will be set to `SafeAutoCorrect: false` if it is unsafe to "auto"-correction.

If users want to safely correct will have to manually handcraft. In this case, users may change using safe navigation operator if needed.

So, `Lint/FrozenStringLiteralComment` cop will behave (potentially) incompatibly with proposed corrrection, either manually or automatically. So I understand that this case is `Safe: false`.

Anyway, I'm also interested in @bbatsov's opinion :-) Thank you!

marcandre:
> So, `Lint/FrozenStringLiteralComment` cop will behave (potentially) incompatibly with proposed corrrection, either manually 
> If users want to safely correct will have to manually handcraft. 

*Manually* one could prefix all needed string literals in the source with a `+` and get 100% compatible behavior.

```ruby
# before

str = 'hello'
str << ' world'

# after

# frozen_string_literal: true
str = +'hello'
str << ' world'
```

That is my whole point...

koic:
Thank you for understanding the explanation! Surely both are `SafeAutoCorrect: false`.  So, I get the definition of `SafeAutoCorrect: false` is done.

On the other hand, I thought again about `Safe`.

- Does `Safe: false` have false positives? => I think this is definitely yes.
- Does `Safe: false` have incompatible behaviour changes? => I'm not sure about the current focus.

It's difficult to say which is user-friendly definition of `Safe` 😅

bbatsov:
Btw, after our original conversation I've documented the safety definitions here https://docs.rubocop.org/rubocop/0.89/usage/auto_correct.html#safe-auto-correct I wouldn't make any changes that are not aligned with those (unless we also update the definition of safety). 

> I believe this cop is safe, only its auto-correction is potentially unsafe.

I agree.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * [#8362](https://github.com/rubocop-hq/rubocop/issues/8362): Add numbers of correctable offenses to summary. ([@nguyenquangminh0711][])
 * [#8513](https://github.com/rubocop-hq/rubocop/pull/8513): Clarify the ruby warning mentioned in the `Lint/ShadowingOuterLocalVariable` documentation. ([@chocolateboy][])
 * [#8517](https://github.com/rubocop-hq/rubocop/pull/8517): Make `Style/HashTransformKeys` and `Style/HashTransformValues` aware of `to_h` with block. ([@eugeneius][])
+* [#x](https://github.com/rubocop-hq/rubocop/pulls/x): Mark `Lint/FrozenStringLiteralComment` as `Safe`, but with unsafe auto-correction. ([@marcandre][])
```
bbatsov, line 10: I think you forgot to update the links. :-)

koic:
> I've documented the safety definitions here

Ah... I overlooked this nice document. I understand the safety definition, thank you!

andyw8:
Note: This PR title appears be incorrect, it's `Style/FrozenStringLiteralComment`, not `Lint/FrozenStringLiteralComment`

marcandre:
Good catch @andyw8, Changelog & PR title fixed, thanks 👍

PR #8537, opened 2020-08-14
Allow a trailing comment as a description comment for `Bundler/GemComment`

pocke:
`Bundler/GemComment` cop isn't aware of trailing comments.

```ruby
# Gemfile


# It is allowed

# It's a cool gem
gem 'rubocop'


# But it is not allowed

gem 'rubocop' # It's a cool gem
```

This pull request fixes this problem.

The cop will allow both of the `gem 'rubocop'` with this patch.

# Cause

In `gem 'rubocop' # It's a cool gem`, the comment isn't associated with the `send` node. Actually it is associated with the `str` node, which is `'rubocop'`.
So we need to check comments for descendants. 

-----------------

* [x] Added tests.


koic:
Good catch 👍

PR #8538, opened 2020-08-14
Proposal: Introduce new cop, Style/NoMetaprogramming, to discourage the use of meta programming

kellysutton:
This Pull Request defines a new cop, `Style/NoMetaprogramming`, that discourages the use of metaprogramming. 

As our engineering team has grown, we've seen some success with using Rubocop to nudge folks away from styles that we might no longer want to encourage within our application. Specifically, we've been removing metaprogramming from our business logic. We've found that this makes this easier to change the application and easier for newcomers to Ruby to understand most code.

I have set this to `default: false`, since I think it might be a bit controversial. Ruby's metaprogramming is what makes it so great, but we've found that it's best used in moderation.

I figure'd we'd upstream some of our work to give back to Rubocop. If this cop is too much, I don't mind closing the PR as well. Thanks for reviewing!

-----------------

* [x] Added tests.


lib/rubocop/cop/style/no_metaprogramming.rb
```diff
@@ -0,0 +1,101 @@
...
+        end
+
+        def on_def(node)
+          using_method_missing?(node) do
+            add_offense(node, message: METHOD_MISSING_MSG)
+          end
+        end
+
+        def on_send(node)
+          using_define_method?(node) do
```
marcandre, line 82: Implementation wise, you could have a hash {method: 'error'} and this would be simpler and faster

marcandre:
I ❤️  metaprogramming, thanks for proposing it to be disabled. I would consider also `Safe: false`.
You're missing a few callbacks (see `Lint/MissingSuper` for inspiration)

lib/rubocop/cop/style/no_metaprogramming.rb
kellysutton: Good idea. I've integrated that into subsequent commits.

kellysutton:
@marcandre 

> I ❤️ metaprogramming, thanks for proposing it to be disabled.

Indeed, that's what makes Ruby so great. We use Rubocop as a tool for teaching newer engineers to Ruby what is okay vs. what might need closer consideration. Over the years, we've found metaprogramming to be one of those things that needs a double-check. But without metaprogramming, we wouldn't have Rails, RSpec, Rubocop, etc. :)

lib/rubocop/cop/style/no_metaprogramming.rb
```diff
@@ -0,0 +1,73 @@
...
+      # This cop specifically will warn on usage of `.included`, `.inherited`,
+      # `#method_missing`, `.define_method`, `#instance_eval`, `.class_eval`,
+      # and `.define_singleton_method`.
+      #
+      class NoMetaprogramming < Base
+        def_node_matcher :using_method_missing?, <<~PATTERN
+          (def :method_missing ...)
+        PATTERN
+
+        METHOD_MISSING_MSG = 'Please do not use method_missing. Instead, explicitly define the'\
```
bbatsov, line 20: I'd be nice if all your messages followed this format, as I see in some of them it's inverted. Let's avoid this inconsistency. Also - please put backticks around all method names referred in those messages. (e.g. `method_missing`).

lib/rubocop/cop/style/no_metaprogramming.rb
```diff
@@ -0,0 +1,73 @@
...
+
+module RuboCop
+  module Cop
+    module Style
+      # Discourages the use of metaprogramming. It's designed
+      # to be enabled in select folders within your application. It was
+      # originally developed with the idea of "Business logic should minimize
+      # the amount of metaprogramming."
+      #
+      # This cop specifically will warn on usage of `.included`, `.inherited`,
```
bbatsov, line 11: Looking at the actual code it seems to disable a few more things which should be mentioned here. The cop also needs some good/bad code examples.

bbatsov:
I'd name the cop something more neural - e.g. just Metaprogramming. Might be nice to have a certain degree of configuration as to what's allowed and not allowed, as I don't think all metaprogramming methods introduce a lot of complexity/mental overhead, but as the cop's disabled by default I'd be fine to accept it in its current state. You'll also have to rebase on top of master, as RuboCop 0.89 was released a while ago.

bbatsov:

PR #8540, opened 2020-08-14
Tweak definition of 'Safe: false'

marcandre:
Also provides example of two ways of being unsafe

Followup to #8529


koic:
Thank you for following up on the documentation updates, and I got the opportunity to improve understanding to `Safe` and `SafeAutoCorrect` 🙏

PR #8543, opened 2020-08-15
[Fix #8512] Enable autocorrect for `Style/NumericPredicate`

### LINKED ISSUE #8512: Style/NumericPredicate
It would be nice if that cop could auto-correct, e.g. `return if i.nil? || i == 0` into `return if i.nil? || i.zero?`

tejasbubane:
Autocorrect was implemented but somehow not enabled.
The cop and autocorrect are still left unsafe.

Closes #8512 

-----------------

* [ ] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -3593,7 +3593,6 @@ Style/NumericPredicate:
   # object. Switching these methods has to be done with knowledge of the types
   # of the variables which rubocop doesn't have.
   SafeAutoCorrect: false
-  AutoCorrect: false
   Enabled: true
   VersionAdded: '0.42'
   VersionChanged: '0.59'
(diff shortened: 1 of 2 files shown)


tejasbubane:
I see some other cops (eg. `Style/ModuleFunction`) whose autocorrect is disabled. Now that `rubocop -a` no longer runs unsafe autocorrects, remove `autocorrect: false` from default configs of all such cops?

Also adding `autocorrect: false` makes the docs generator see `autocorrect` as a configurable attribute.

tejasbubane:
CI failure seems to be unrelated.

bbatsov:
> I see some other cops (eg. Style/ModuleFunction) whose autocorrect is disabled. Now that rubocop -a no longer runs unsafe autocorrects, can we remove autocorrect: false from default configs of all such cops?

Yeah, that's a historical artefact. I don't think it makes sense to have any auto-correct explicitly disabled going forward.

PR #8552, opened 2020-08-16
Add new `Lint/UselessMethodDefinition` cop

### LINKED ISSUE #8472: New cop: `Lint/UselessMethodDefinition`
There are at least 2 cases:
1. empty constructors - this may be just overriding parent's constructor, but this is bad anyway 
2. methods just calling `super`

fatkodima:
Closes #8472

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1942,6 +1942,13 @@ Lint/UselessElseWithoutRescue:
   Enabled: true
   VersionAdded: '0.17'
 
+Lint/UselessMethodDefinition:
+  Description: 'Checks for useless method definitions.'
+  Enabled: pending
+  VersionAdded: '0.90'
+  Safe: false
+  AllowComments: true
+
 Lint/UselessSetterCall:
   Description: 'Checks for useless setter call to a local variable.'
   Enabled: true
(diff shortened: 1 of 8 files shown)


config/default.yml
```diff
@@ -1942,6 +1942,13 @@ Lint/UselessElseWithoutRescue:
   Enabled: true
   VersionAdded: '0.17'
 
+Lint/UselessMethod:
```
bbatsov, line 1945: I think "UselessMethodDefinition" might be a better name for this.

lib/rubocop/cop/lint/useless_method.rb
```diff
@@ -0,0 +1,77 @@
...
+      # @example AllowComments: false
+      #   # bad
+      #   def initialize
+      #     # Comment.
+      #   end
+      #
+      class UselessMethod < Base
+        extend AutoCorrector
+
+        MSG = 'Useless method detected.'
```
bbatsov, line 46: method definition

fatkodima:
Updated with new name.

bbatsov:
Thanks!

PR #8563, opened 2020-08-18
Add new `Style/KeywordArgumentsOrder` cop

fatkodima:
Related https://github.com/rubocop-hq/ruby-style-guide/pull/771

This cop enforces that optional keyword arguments are placed at the end of the arguments list.

```ruby
# bad
def some_method(first: false, second:, third: 10)
  # body omitted
end

# good
def some_method(second:, first: false, third: 10)
  # body omitted
end
```


lib/rubocop/cop/style/keyword_arguments_order.rb
```diff
@@ -0,0 +1,58 @@
...
+      #   def some_method(first: false, second:, third: 10)
+      #     # body omitted
+      #   end
+      #
+      #   # good
+      #   def some_method(second:, first: false, third: 10)
+      #     # body omitted
+      #   end
+      #
+      class KeywordArgumentsOrder < Base
```
bbatsov, line 24: In this context it's parameters, not arguments - params in the definition, args in the call sites. The messages have to updated accordingly as well.

PR #8573, opened 2020-08-23
[Fix #7705] Fix `Style/OneLineConditional` to handle if/then/elsif/then/else and add AlwaysCorrectToMultiline config option

### LINKED ISSUE #7705: Rubocop incorrectly tries to transform one line conditional to ternary operator

dmytro-savochkin:
Fixes #7705.

This is continuation of work started in https://github.com/rubocop-hq/rubocop/pull/7710 by https://github.com/Lykos. I basically used the code proposed in that PR, fixed some minor things, added more specs, and also added a new config parameter called `AlwaysCorrectToMultiline` (I think this addresses the suggestion raised in https://github.com/rubocop-hq/rubocop/pull/7710#issuecomment-593073187). 

-----------------

* [x] Added tests.


bbatsov:
Thanks for driving this to the finish line!

PR #8578, opened 2020-08-24
Add `:restore_registry` context and `stub_cop_class` helper class

marcandre:
Makes it trivial define a new cop and/or insure that the global cop registry doesn't get polluted.


koic (approved):
Cool! ❤️

bbatsov:
I think that was already merged with #8579.

PR #8579, opened 2020-08-24
Add `Cop.documentation_url`

marcandre:
Add `Cop.documentation_url`. This returns a full URL for builtin cops, `nil` for others.

Should be easy for other cops (e.g. `rubocop/rspec`) to override and return their own URL.


marcandre:
Includes #8578

tas50:
I've needed this for a while for cookstyle. Thanks @marcandre

bbatsov:
Looks good!

PR #8594, opened 2020-08-26
[Fixes #8321] Enable auto correction for `Layout/{Def}EndAlignment`, `Lint/EmptyEnsure`

### LINKED ISSUE #8321: Layout/EndAlignment should autocorrect
```
         def rspec_pattern
           @rspec_pattern ||= if rspec_pattern_config?
                               Regexp.union(rspec_pattern_config.map(&Regexp.public_method(:new)))
                             else
                               DEFAULT_PATTERN_RE
           end
         end
```

I get "Layout/EndAlignment: end at 52, 10 is not aligned with if at 48, 29."

The cop should autocorrect it.

marcandre:


bbatsov (approved):
I'm still on the fence about the value of Lint auto-corrections (mostly because often they are nuanced and the obvious correction might hide the real problem - e.g. did you forget to put some code in the ensure block, etc), but as we have some and people find them useful, as I guess it's fine to enable all of them.

PR #8595, opened 2020-08-26
Enable autocorrection for `Style/ClassAndModuleChildren`

marcandre:
We have 3 cops remaining with `AutoCorrect: false`:

* `Layout/LineLength`: probably too crude to be enabled by default
* `Security/JSONLoad`: too crude too, and quite minor
* `Style/ClassAndModuleChildren`: correctly marked as unsafe auto-correction, I think it should be enabled by default, hence this PR.

Follows #8594


bbatsov:
> Layout/LineLength: probably too crude to be enabled by default

On the other hand - I have it on good authority many people don't even know this auto-correction exists and dream of it. I'd vote to enable it. Maybe this will force us to improve it. :-)

marcandre:
> > Layout/LineLength: probably too crude to be enabled by default
> 
> On the other hand - I have it on good authority many people don't even know this auto-correction exists and dream of it. I'd vote to enable it. Maybe this will force us to improve it. :-)

`LineLength` is definitely the offense that I raise the most in my code, and I *almost never* want it to be fixed automatically; I typically will extract some sub-expression and store it in a local and no improved autocorrector could do that (and name the variable properly). The only auto-correction I wouldn't mind happening is an inline block being expanded to a multi line one.

Is there a way to have a `.rubocop.local.yml` for local overrides (for RuboCop)? We'd need a `inherit_from_if_exist` that doesn't raises if the config doesn't exist? This way I wouldn't be impacted; otherwise I will probably be forced to use `-a` instead of `-A`

bbatsov:
> LineLength is definitely the offense that I raise the most in my code, and I almost never want it to be fixed automatically; I typically will extract some sub-expression and store it in a local and no improved autocorrector could do that (and name the variable properly). The only auto-correction I wouldn't mind happening is an inline block being expanded to a multi line one.

I understand this perspective, but on the other hand without reflowing long lines automatically you can't really use RuboCop as a formatter. 

> Is there a way to have a .rubocop.local.yml for local overrides (for RuboCop)? We'd need a inherit_from_if_exist that doesn't raises if the config doesn't exist? This way I wouldn't be impacted; otherwise I will probably be forced to use -a instead of -A

What would be the purpose of this local config? Overriding project settings or user settings?

PR #8609, opened 2020-08-28
Revert "Mark Cops and Formatters as private"

koic:
This reverts commit 90d5333f4c150bfe33d07805a12879b7fa4d6201.

I was late to participate in the discussion below.
https://github.com/rubocop-hq/rubocop/pull/8501

Cop names are specified by users to .rubocop.yml, `--only` option, and etc. IMHO, the names and behaviors cannot be
changed without impacting users.

Users should refrain from depend on it as an implementation API, but I don't think it's hidden from users.

Cc @rubocop-hq/rubocop-core

-----------------

* [ ] Added tests.


marcandre:
>  the names and behaviors cannot be changed without impacting users.

Right, and we have a whole documentation section for this.

I feel the implementation of this is private, though, and would like to avoid others inheriting our Cops and overriding methods and thinking "that's not going to break". I think we should reserve the right to change pretty much everything about the cops, including their namespace.

That being said, I'm not strongly opposed to reverting the commit 😄

bbatsov:
Let's revert this for now and ruminate on the subject more down the road.

PR #8638, opened 2020-09-03
[Fix #8572] Fix a false positive for `Style/RedundantParentheses`

### LINKED ISSUE #8572: Bug: Invalid autocorrection interaction between Lint/ParenthesesAsGroupedExpression & Style/RedundantParentheses
In Standard @sethherr noticed an undesired interaction between the safe autocorrections of `Lint/ParenthesesAsGroupedExpression` and `Style/RedundantParentheses` (both enabled-by-default): https://github.com/testdouble/standard/issues/192

--------

## Expected behavior

Here's the starting code example:

```ruby
do_something (foo)
```

Here's what happens when I run only `Lint/ParenthesesAsGroupedExpression` and `Style/RedundantParentheses`:

```
$ rubocop --only "Lint/ParenthesesAsGroupedExpression,Style/RedundantParentheses" foo.rb 
Inspecting 1 file
W

Offenses:

foo.rb:1:13: W: Lint/ParenthesesAsGroupedExpression: (...) interpreted as grouped expression.
do_something (foo)
            ^
foo.rb:1:14: C: Style/RedundantParentheses: Don't use parentheses around a method call.
do_something (foo)
             ^^^^^

```

My expectation is the autocorrect would ultimately result in either `do_something foo` or `do_something(foo)`

## Actual behavior

When running the same command with autocorrect, this is what we get:

```
$ be rubocop --only "Lint/ParenthesesAsGroupedExpression,Style/RedundantParentheses" -a foo.rb
Inspecting 1 file
W

Offenses:

foo.rb:1:13: W: [Corrected] Lint/ParenthesesAsGroupedExpression: (...) interpreted as grouped expression.
do_something (foo)
            ^
foo.rb:1:14: C: [Corrected] Style/RedundantParentheses: Don't use parentheses around a method call.
do_something (foo)
             ^^^^^

```

And now the result is definitely wrong:

```
$ cat foo.rb 
do_somethingfoo

```

I hope the instructions above are a clear enough minimal reproduction

## RuboCop version

```
$ rubocop -V
0.89.1 (using Parser 2.7.1.4, rubocop-ast 0.3.0, running on ruby 2.7.1 x86_64-darwin19)
```

koic:
Fixes #8572.

This PR fixes a false positive for `Style/RedundantParentheses` when like method argument parentheses.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_parentheses.rb
+++ b/lib/rubocop/cop/style/redundant_parentheses.rb
@@ -26,9 +26,7 @@ module RuboCop
                          '^^(block (send _ _ equal?(%0) ...) ...)'
 
         def on_begin(node)
-          return if !parentheses?(node) || parens_allowed?(node)
-          return if node.parent && (node.parent.while_post_type? ||
-                                    node.parent.until_post_type?)
+          return if !parentheses?(node) || parens_allowed?(node) || ignore_syntax?(node)
 
           check(node)
         end
@@ -40,6 +38,13 @@ module RuboCop
             first_arg_begins_with_hash_literal?(node) ||
             rescue?(node) ||
             allowed_expression?(node)
+        end
+
+        def ignore_syntax?(node)
+          return false unless (parent = node.parent)
+
+          parent.while_post_type? || parent.until_post_type? ||
+            like_method_argument_parentheses?(parent)
         end
 
         def allowed_expression?(node)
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/style/redundant_parentheses.rb
```diff
@@ -27,8 +27,8 @@ class RedundantParentheses < Base
 
         def on_begin(node)
           return if !parentheses?(node) || parens_allowed?(node)
-          return if node.parent && (node.parent.while_post_type? ||
-                                    node.parent.until_post_type?)
+          return if node.parent && (node.parent.while_post_type? || node.parent.until_post_type? ||
```
bbatsov, line 30: Probably that big condition can become a helper method as well.

koic: I updated this PR, but I'm not sure due to there may be a better approach (or name) 💦

bbatsov:
Looks good!

PR #8639, opened 2020-09-03
Feature: add support for RUBOCOP_CACHE_ROOT env variable to control cache path

sascha-wolf:
This PR introduces the `--cache-root` command line option and support for the `RUBOCOP_CACHE_ROOT` env variable to control the cache root directory.

It takes precedence over everything else (even the config file) which makes it perfectly suited for CI usage where it's very helpful to control the cache path (without interference from the config file) to consistently be able to save and restore rubocop's cache files.

## Possible further steps

- [x] also add a cli option `--cache-root`
- [ ] print a warning when there is both: `RUBOCOP_CACHE_ROOT` and `CacheRootDirectory` in the config file

## TODO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* New option `--cache-root` and support for the `RUBOCOP_CACHE_ROOT` environment variable. Both can be used to override the `AllCops: CacheRootDirectory` config, especially in a CI setting. ([@sascha-wolf][]) 
+
 ### Bug fixes
 
 * [#8627](https://github.com/rubocop-hq/rubocop/issues/8627): Fix a false positive for `Lint/DuplicateRequire` when same feature argument but different require method. ([@koic][])
@@ -4829,3 +4833,4 @@
 [@Lykos]: https://github.com/Lykos
 [@jaimerave]: https://github.com/jaimerave
 [@Skipants]: https://github.com/Skipants
+[@sascha-wolf]: https://github.com/sascha-wolf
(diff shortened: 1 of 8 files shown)


bbatsov:
I'm fined with the proposed change, although the rationale for it in the docs should be expanded upon. I also like the idea a having a cli option for this, which IMO should have the highest possible priority (and also introduces less cognitive overhead).

sascha-wolf:
@bbatsov I've added a `--cache-root` CLI option in 96e215d87ff369c40ffe95874e420fb4bc66c771, dccc65ba19fd2eba293983e88b718e6579efc8bf, and 293c93687b0997ba1685efaca7a3baeb097734cc.

>  although the rationale for it in the docs should be expanded upon

What exactly do you mean with this? Do you mean that the docs should also mention the CI use-case?

sascha-wolf:
On another note, do you think it would be beneficial to print a warning, if either `--cache-root` or `RUBOCOP_CACHE_ROOT` take precedence while other configuration exists?

sascha-wolf:
@bbatsov I've expanded the docs to include the `--cache-root` command line option and list the options to control the cache root in order of precedence (bfcea678da6d90b5df7eea4b4ebe7f11e6b9d168).

Is this what you had in mind?

bbatsov:
> On another note, do you think it would be beneficial to print a warning, if either --cache-root or RUBOCOP_CACHE_ROOT take precedence while other configuration exists?

Might be useful, but I don't think that will add that much value, as in some cases the users would have done this intentionally and the warning would just be annoying to them (e.g. different setup for local development and CI with the same config file).

bbatsov:
Thanks!

PR #8643, opened 2020-09-03
Optimize cache invalidation when running RuboCop as a gem [See #8633]

marcandre:


lib/rubocop.rb
```diff
@@ -627,3 +628,7 @@
 require_relative 'rubocop/remote_config'
 require_relative 'rubocop/target_ruby'
 require_relative 'rubocop/yaml_duplication_checker'
+
+unless File.exist?("#{__dir__}/../Gemfile") # Check if we are a gem
```
dvandersluis, line 632: Would `File.exist?(Bundler.default_gemfile)` be better here, since the gemfile does not have to be actually named `Gemfile` these days?

dvandersluis: I guess since you're just looking for rubocop's Gemfile itself it's ok to hardcode it.

marcandre: I didn't know that. I'm not sure if `Bundler` is always required, so I'll leave it as is. The potential impact would be for someone who is developping on `RuboCop`, and somehow went out of their way to not use our `Gemfile` and deleting it to boot...
Thanks for bringing it up though.

bbatsov: Btw, isn't it better to check for the gemspec instead?

marcandre: No problem; #8645 should merge soon.

PR #8644, opened 2020-09-03
Use `RESTRICT_ON_SEND` for the rest of cops

### LINKED ISSUE #8606: Cops using `on_send` can be optimized
Now that #8365 is merged, most cops using `on_send` can now be optimized by defining a constant `RESTRICT_ON_SEND` with the list of methods the cop needs to watch for.

Cops that don't depend on a particular list of methods (say because they only look at the arguments) should be left as is.

If you want to help, just comment on which cops you want to look at first (e.g. `Lint/A-E`)

fatkodima:
Closes #8606 

/cc @marcandre

### DIFF
--- a/lib/rubocop/cop/lint/multiple_comparison.rb
+++ b/lib/rubocop/cop/lint/multiple_comparison.rb
@@ -25,9 +25,11 @@ module RuboCop
         extend AutoCorrector
 
         MSG = 'Use the `&&` operator to compare multiple values.'
+        COMPARISON_METHODS = %i[< > <= >=].freeze
+        RESTRICT_ON_SEND = COMPARISON_METHODS
 
         def_node_matcher :multiple_compare?, <<~PATTERN
-          (send (send _ {:< :> :<= :>=} $_) {:< :> :<= :>=} _)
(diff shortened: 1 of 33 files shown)


lib/rubocop/cop/lint/multiple_comparison.rb
```diff
@@ -25,9 +25,11 @@ class MultipleComparison < Base
         extend AutoCorrector
 
         MSG = 'Use the `&&` operator to compare multiple values.'
+        COMPARISON_METHODS = %i[< > <= >=].freeze
+        RESTRICT_ON_SEND = COMPARISON_METHODS
 
         def_node_matcher :multiple_compare?, <<~PATTERN
-          (send (send _ {:< :> :<= :>=} $_) {:< :> :<= :>=} _)
+          (send (send _ {:< :> :<= :>=} $_) {:#{COMPARISON_METHODS.join(' :')}} _)
```
marcandre, line 32: FYI, if `COMPARISON_METHODS` was a set, you could use `%COMPARISON_METHODS` here... but I'm also working on optimizing `{:list :of :literals}` into a set, so no need to change this. For the future though, might be more readable :-)

marcandre (approved):
Good job!

PR #8648, opened 2020-09-04
[Fix #8630] Fix some hash transformation false positives

### LINKED ISSUE #8630: Style/HashTransformKeys with to_h with a block is too greedy now
This code (and any like this, using `Array#to_h` or `Enumerable#to_h`) is a false positive now:
```ruby
%i[a b c].zip([1, 2, 3]).to_h { |name, val| [name, val**2] }
```
↓
```
C: Style/HashTransformValues: Prefer transform_values over to_h {...}.
%i[a b c].zip([1, 2, 3]).to_h { |name, val| [name, val**2] }
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
(After #8517 probably)

eugeneius:
We can't know for certain when the receiver is a hash, which is why these cops are marked as unsafe. However there are some false positives that we can detect.

Array literals and `each_with_index` were already handled; here I'm adding `with_index` and `zip`.

-----------------

* [x] Added tests.


koic:
Thanks!

PR #8649, opened 2020-09-04
[Fix #8642] Fix a false negative for `Style/SpaceInsideHashLiteralBraces` when a correct empty hash precedes the incorrect hash

### LINKED ISSUE #8642: Layout/SpaceInsideHashLiteralBraces does not find all violations in some situations and has to be run at least twice
## Expected behavior

I expect the cop to find all violations:
```bash
❯ rubocop demo.rb
Inspecting 1 file
C

Offenses:

demo.rb:8:3: C: Layout/SpaceInsideHashLiteralBraces: Space inside { missing.
  {key: 1}
  ^
demo.rb:8:10: C: Layout/SpaceInsideHashLiteralBraces: Space inside } missing.
  {key: 1}
         ^

1 file inspected, 2 offenses detected, 2 offenses auto-correctable
```

## Actual behavior

It only finds one violation:
```bash
❯ rubocop demo.rb
Inspecting 1 file
C

Offenses:
demo.rb:8:3: C: Layout/SpaceInsideHashLiteralBraces: Space inside { missing.
  {key: 1}
  ^

```

Create the file `demo.rb`:
```ruby
# frozen_string_literal: true

def abc
  {}
end

def xyz
  {key: 1}
end
```

Run rubocop:
```bash
❯ rubocop demo.rb
```

I don't know why, but the method body of `abc` is relevant. If you change it from `{}` to for example `'a string'`, then the cop works as expected.

## RuboCop version

```bash
❯ rubocop -V
0.90.0 (using Parser 2.7.1.4, rubocop-ast 0.3.0, running on ruby 2.7.1 x86_64-darwin19)
```

dvandersluis:
Fixes #8642.

I'm not 100% sure I understand what `ConfigurableEnforcedStyle` is doing (it'd be great if it could have more documentation added!), but the problem here is that the detected style was flipping between `space` and `nospace` depending on which hash was being evaluated. The fix is to just pass the defined `style` to `ambiguous_or_unexpected_style_detected` rather than trying to calculate it.

---

Consider the following code (which is also added as a test):
```ruby
{}
{key: 1}
```

1. When the first empty hash is evaluated by the cop, the memoized detected style (in `ConfigurableEnforcedStyle`) is `nil`. Because the spacing for the empty has is correct, `correct_style_detected` is called, which sets [`detected_as_strings` and `updated_list`](https://github.com/rubocop-hq/rubocop/blob/d3f6b097a8fc65a64d050b5844a1a4fe2505d503/lib/rubocop/cop/mixin/configurable_enforced_style.rb#L30-L35) to `['space']` (since that is the `EnforcedStyle` value).
2. When the first brace of the second hash is evaluated by the `check` method, it [detects an offense and runs `incorrect_style_detected`](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/layout/space_inside_hash_literal_braces.rb#L106-L108). At this point, a space is required [`no_space` is passed to `ambiguous_or_unexpected_style_detected`](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/layout/space_inside_hash_literal_braces.rb#L132-L133). Because of this, `updated_list` is set to the intersection of `[space]` and `[no_space]`, which is `[]`, and [`no_acceptable_style!`](https://github.com/rubocop-hq/rubocop/blob/d3f6b097a8fc65a64d050b5844a1a4fe2505d503/lib/rubocop/cop/mixin/configurable_enforced_style.rb#L37) is called. The offense gets registered (and corrected).
3. When further hash braces are encountered, `no_acceptable_style?` returns true, and [`style_detected` short-circuits](https://github.com/rubocop-hq/rubocop/blob/d3f6b097a8fc65a64d050b5844a1a4fe2505d503/lib/rubocop/cop/mixin/configurable_enforced_style.rb#L24). This causes `incorrect_style_detected` to return before adding an offense, and **all further hash spacing issues are ignored in the file**.

My change is to not override style but pass in the style from config on both `correct_style_detected` and `incorrect_style_detected`, which prevents `no_acceptable_style?` from short-circuiting further offense detection. None of the other code or existing tests had to be changed.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * [#8671](https://github.com/rubocop-hq/rubocop/issues/8671): Fix an error for `Style/ExplicitBlockArgument` when using safe navigation method call. ([@koic][])
 * [#8682](https://github.com/rubocop-hq/rubocop/pull/8682): Fix a positive for `Style/HashTransformKeys` and `Style/HashTransformValues` when the `each_with_object` hash is used in the transformed key or value. ([@eugeneius][])
 * [#8688](https://github.com/rubocop-hq/rubocop/issues/8688): Mark `Style/GlobalStdStream` as unsafe autocorrection. ([@marcandre][])
+* [#8642](https://github.com/rubocop-hq/rubocop/issues/8642): Fix a false negative for `Style/SpaceInsideHashLiteralBraces` when a correct empty hash precedes the incorrect hash. ([@dvandersluis][])
 
 ### Changes
 
--- a/lib/rubocop/cop/layout/space_inside_hash_literal_braces.rb
+++ b/lib/rubocop/cop/layout/space_inside_hash_literal_braces.rb
@@ -126,11 +126,10 @@ module RuboCop
 
         def incorrect_style_detected(token1, token2,
                                      expect_space, is_empty_braces)
+          return unless ambiguous_or_unexpected_style_detected(style, token1.text == token2.text)
+
           brace = (token1.text == '{' ? token1 : token2).pos
           range = expect_space ? brace : space_range(brace)
-
-          style = expect_space ? :no_space : :space
-          return unless ambiguous_or_unexpected_style_detected(style, token1.text == token2.text)
 
           add_offense(range, message: message(brace, is_empty_braces, expect_space)) do |corrector|
             autocorrect(corrector, range)
--- a/spec/rubocop/cop/layout/space_inside_hash_literal_braces_spec.rb
+++ b/spec/rubocop/cop/layout/space_inside_hash_literal_braces_spec.rb
@@ -216,4 +216,21 @@ RSpec.describe RuboCop::Cop::Layout::SpaceInsideHashLiteralBraces, :config do
       expect_no_offenses('{ key: "{" }')
     end
   end
+
+  context 'offending hash following empty hash' do
+    # regression test; see GH issue 8642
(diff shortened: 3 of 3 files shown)


dvandersluis:
This was a regression in 0.90 btw, it worked fine in 0.89.

bbatsov:
Did you manage to identify which commit introduced the regression?

dvandersluis:
@bbatsov I ran `git bisect` on it, and 6c05f69f4c1919b40fe6e331bab97833cf87bd0e is the first bad commit.

dvandersluis:
@bbatsov I think it's because of https://github.com/rubocop-hq/rubocop/commit/6c05f69f4c1919b40fe6e331bab97833cf87bd0e#diff-434a2611de47b29849cbf58aeb8094aeL141-R134

Before `ambiguous_or_unexpected_style_detected` was only called inside the `add_offense` block, but now it's called every time `incorrect_style_detected` is called.

koic (approved):
This regression is due to my mistake 💦 Thank you for the fixing. Can you rebase with the latest master branch?

dvandersluis:
@koic of course! done.

koic:
@dvandersluis Thanks!

getaaron:
@koic @dvandersluis thank you for fixing this so quickly! would it be possible to cut a new release 0.91 with this fix kinda soon? it's fairly frustrating

PR #8657, opened 2020-09-06
Fix a false positive for `Lint/UnreachableLoop` when conditional branch includes continue statement preceding break statement

### LINKED ISSUE #8607: Lint/UnreachableLoop false positive when next appears in a branch
In a loop where an `if` branch has both outcomes ending in `break` or `rescue`, but one branch has a `next` guard clause in it, Rubocop still complains about Lint/UnreachableLoop. Consider this:

```ruby
[1, 2].each do |x|
  p x
  if x == 1
    next if x == 1
    break
  else
    break
  end
end
```

This runs twice.

--------

## Expected behavior

Should not complain about Lint/UnreachableLoop

## Actual behavior

The loop runs twice, but Rubocop complains: "Lint/UnreachableLoop: This loop will have at most one iteration"

Run sample code

## RuboCop version

```
0.89.1 (using Parser 2.7.1.4, rubocop-ast 0.3.0, running on ruby 2.4.5 x86_64-darwin19)
```

Love all the work you're doing 👍

fatkodima:
Closes #8607

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Bug fixes
 
 * [#8627](https://github.com/rubocop-hq/rubocop/issues/8627): Fix a false positive for `Lint/DuplicateRequire` when same feature argument but different require method. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #8658, opened 2020-09-07
Fix a false positive for `Style/RedundantSelfAssignment` when calling coercion methods

fatkodima:
Those two returns self, not modifies its receiver in place. So that was my mistake when implementing this cop.


marcandre:
Thanks! (they don't always return `self` either...)

PR #8661, opened 2020-09-07
Fix an incorrect auto-correct for `Style/MultilineTernaryOperator`

koic:
This PR fixes the following incorrect auto-correct for `Style/MultilineTernaryOperator` when returning a multiline ternary operator expression.

```console
% cat example.rb
return distance_in_minutes == 0 ?
       locale.t(:less_than_x_minutes, count: 1) :
       locale.t(:x_minutes, count: distance_in_minutes)

% bundle exec rubocop -a --only Style/MultilineTernaryOperator
(snip)

Inspecting 2 files
C.

Offenses:

example.rb:1:8: C: [Corrected] Style/MultilineTernaryOperator: Avoid
multi-line ternary operators, use if or unless instead.
return distance_in_minutes == 0 ? ...
       ^^^^^^^^^^^^^^^^^^^^^^^^^^

2 files inspected, 1 offense detected, 1 offense corrected

% cat example.rb
return if distance_in_minutes == 0
  locale.t(:less_than_x_minutes, count: 1)
else
  locale.t(:x_minutes, count: distance_in_minutes)
end

% ruby -c example.rb
example.rb:3: syntax error, unexpected `else', expecting end-of-input
```

I found it with the following code.
https://github.com/rails/rails/blob/v6.0.3.2/actionview/lib/action_view/helpers/date_helper.rb#L109-L111

-----------------

* [x] Added tests.

### DIFF
--- a/docs/modules/ROOT/pages/cops_style.adoc
+++ b/docs/modules/ROOT/pages/cops_style.adoc
@@ -5757,6 +5757,9 @@ end
 
 This cop checks for multi-line ternary op expressions.
 
+NOTE: `return if ... else ... end` is syntax error. If `return` is used before
+multiline ternary operator expression, it cannot be auto-corrected.
+
 === Examples
 
 [source,ruby]
--- a/lib/rubocop/cop/style/multiline_ternary_operator.rb
+++ b/lib/rubocop/cop/style/multiline_ternary_operator.rb
@@ -5,6 +5,9 @@ module RuboCop
     module Style
       # This cop checks for multi-line ternary op expressions.
       #
+      # NOTE: `return if ... else ... end` is syntax error. If `return` is used before
+      # multiline ternary operator expression, it cannot be auto-corrected.
+      #
       # @example
       #   # bad
       #   a = cond ?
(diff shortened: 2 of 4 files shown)


docs/modules/ROOT/pages/cops_style.adoc
```diff
@@ -5777,6 +5777,9 @@ a = if cond
 else
   c
 end
+return cond ?
```
bbatsov, line 5780: I guess we shouldn't auto-correct this, but I think we should still flag this, as it looks quite weird.

koic: Ah, I get it and updated this PR!

bbatsov (approved):
You can add in the docs some explanation why won't auto-correct a ternary op in `return`.

lib/rubocop/cop/style/multiline_ternary_operator.rb
```diff
@@ -29,9 +29,11 @@ class MultilineTernaryOperator < Base
               'use `if` or `unless` instead.'
 
         def on_if(node)
-          return unless node.ternary? && node.multiline?
+          return unless offense?(node)
 
           add_offense(node) do |corrector|
+            next unless offense?(node) && !node.parent.return_type?
```
bbatsov, line 35: You can add a comment here explaining why this is needed for future reference.

koic:
Thank you for your advice. I added a note to the doc.

PR #8676, opened 2020-09-09
Fix an offense creation for `Lint/EmptyFile`

### LINKED ISSUE #8669: Lint/EmptyFile throws error when html formatter is used
`Lint/EmptyFile` (added in 0,90.0) throws an error when the `html` formatter is used. 

### Steps to Reproduce

1. Enable the cop in a `.rubocop.yml` 

    ```yaml
    Lint/EmptyFile:
      Enabled: true
    ```
1. Ensure the project has an empty `.rb` file.

1. Run 

    ```bash
    $ rubocop --format html --out results.html
    ```

    The following error occurs:

    ```
    index 0 outside of array bounds: 0...0
    /home/dev/.rvm/gems/ruby-2.6.6/gems/parser-2.7.1.4/lib/parser/source/buffer.rb:270:in `fetch'
    ...
    ```

1. Run

    ```bash 
    $ rubocop
    ```

    It works as expected.

### Environment

```
$ [bundle exec] rubocop -V
0.90.0 (using Parser 2.7.1.4, rubocop-ast 0.3.0, running on ruby 2.6.6 x86_64-linux)
```

fatkodima:
Closes #8669

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * [#8627](https://github.com/rubocop-hq/rubocop/issues/8627): Fix a false positive for `Lint/DuplicateRequire` when same feature argument but different require method. ([@koic][])
 * [#8674](https://github.com/rubocop-hq/rubocop/issues/8674): Fix an error for `Layout/EmptyLineAfterMultilineCondition` when conditional is at the top level. ([@fatkodima][])
 * [#8658](https://github.com/rubocop-hq/rubocop/pull/8658): Fix a false positive for `Style/RedundantSelfAssignment` when calling coercion methods. ([@fatkodima][])
+* [#8669](https://github.com/rubocop-hq/rubocop/issues/8669): Fix an offense creation for `Lint/EmptyFile`. ([@fatkodima][])
 * [#8607](https://github.com/rubocop-hq/rubocop/issues/8607): Fix a false positive for `Lint/UnreachableLoop` when conditional branch includes continue statement preceding break statement. ([@fatkodima][])
 * [#8572](https://github.com/rubocop-hq/rubocop/issues/8572): Fix a false positive for `Style/RedundantParentheses` when parentheses are used like method argument parentheses. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Can you explain why those changes are needed? Personally, I don't quite get it.

marcandre:
I thought the HTML formatter was the culprit (or maybe even my rewriter code)

fatkodima:
The original code
```ruby
range = source_range(processed_source.buffer, 1, 0)		
add_offense(range)
```

adds an offense to non existent (empty) source range - 1st line, 0-th column. 
When generating html report (I didn't tested, but this can be a problem for other formatters), this https://github.com/rubocop-hq/rubocop/blob/9bc7908b1ad171b33ede24908acced724db6e8d4/lib/rubocop/formatter/html_formatter.rb#L118-L126

119 and 124 lines fails here 

https://github.com/whitequark/parser/blob/c25585d8d91f3470d9cdfd6f3b6a553ad9a809fb/lib/parser/source/buffer.rb#L269-L271

because of `fetch` on empty `source_lines`.

`add_global_offense` adds a global offense, roughly speaking, not attached to any specific location and overrides `source_line` to empty string, so this eliminates the problem.

bbatsov:
Got it. Might be nice to integrate those global offenses with `expect_offence`, though, as I'm certain people will wonder why are the tests using it.

fatkodima:
> Might be nice to integrate those global offenses with expect_offence

`syntax_spec.rb` uses the same style, as this test file, so also needs an update. 
Wdyt about introducing it in a separate PR?

marcandre:
I think:

```ruby
source_range(processed_source.buffer, 1, 0)
# should be
source_range(processed_source.buffer, 1, 0, 0)	
```

FWIW, this helper is terrible. There are too many of these multiple argument *public* methods... The cop in charge of that should really have different settings for private (I don't care much, really) and public (should be max 1 imo).

If you prefer to go the `add_global_offense` route (seems a bit more complicated), I think the odd spec is fine. `add_global_offense` was really meant for non-parseable files tbh.

fatkodima:
> source_range(processed_source.buffer, 1, 0)
# should be
source_range(processed_source.buffer, 1, 0, 0)	

Just tested, the same error as in the original issue.

bbatsov:
Well, I guess the suggested fix will do for now. Thanks!

PR #8686, opened 2020-09-10
Put changelog item for #8629 in correct version

dvandersluis:
Somehow this changelog item ended up in 0.90 instead of master.


koic:
Thanks!

PR #8687, opened 2020-09-10
[Fix #8663] Fix issues with Style/ClassMethodsDefinitions

### LINKED ISSUE #8663: Style/ClassMethodsDefinitions is wrongly creating an instance method
I use autocorrect on a Style/ClassMethodsDefinitions issue and it wrongly restructures the code

Version rubocop (0.90.0)

--------

## Expected behavior

```
class Repository
  attr_reader :causation_id, :correlation_id

  def self.with_causation_id(causation_id)
    new(causation_id: causation_id, correlation_id: causation_id)
  end

  def self.with_event(event)
    new(causation_id: event.id, correlation_id: event.correlation_id)
  end

  def initialize(causation_id:, correlation_id:)
```

## Actual behavior

**Major issues:**
- ` def with_causation_id(causation_id)` is not remove
- `  class << self  end` empty block stay in the code

**Minor Issues:**
- an extra empty line is created
- indentation is wrong

```
class Repository
  attr_reader :causation_id, :correlation_id

  
    def self.with_causation_id(causation_id)
      new(causation_id: causation_id, correlation_id: causation_id)
    end

  
    def self.with_event(event)
      new(causation_id: event.id, correlation_id: event.correlation_id)
    end

    def with_causation_id(causation_id)
      new(causation_id: causation_id, correlation_id: causation_id)
    end

  class << self  end

  def initialize(causation_id:, correlation_id:)
  ....
```

Starting code:

```
class Repository
  attr_reader :causation_id, :correlation_id

  class << self
    def with_event(event)
      new(causation_id: event.id, correlation_id: event.correlation_id)
    end

    def with_causation_id(causation_id)
      new(causation_id: causation_id, correlation_id: causation_id)
    end
  end

  def initialize ....
```

## RuboCop version

```
$ rubocop -V
0.90.0 (using Parser 2.7.1.4, rubocop-ast 0.3.0, running on ruby 2.6.6 x86_64-darwin18)
```

dvandersluis:
I found quite a few issues with autocorrection in `Style/ClassMethodsDefinitions`, that are now handled, with fuller test coverage.

* More than one method within a `class << self` block is now handled correctly (previously it would duplicate methods)
* Empty `self << class` blocks after autocorrection are now removed.
* Extra whitespace around newly written methods is removed.
* New methods are indented at the same level `class << self` was.

Fixes #8663.

---

Previously each def within `class << self` was assigned a separate offense, but this was causing a number of issues that when I tried to fix I was getting a clobbering error (for instance, `source_range_with_comment` in `CommentsHelp` was returning the entire body of the `sclass`, rather than just a single def node + comments, which was leading to duplicated methods).

I could not find a way to add offenses on each `def` but only autocorrect once, so I decided to move the offense to the `sclass` itself. I hope this is an okay tradeoff; please let me know if there's a better way.

I am not 100% satisfied with the amount of steps necessary in the corrector but I could not find a simpler solution that satisfied all the things I was trying to do (correct indentation, remove extra lines, and remove the empty `class << self`). Again if there is a suggestion for a better corrector, please let me know!

-----------------

* [x] Added tests.


lib/rubocop/cop/mixin/comments_help.rb
```diff
@@ -7,15 +7,9 @@ module CommentsHelp
...
-            start_node = find_visibility_start(node) || node
-            end_node = find_visibility_end(node) || node
-            [begin_pos_with_comment(start_node),
-             end_position_for(end_node) + 1]
-          else
-            [begin_pos_with_comment(node), end_position_for(node)]
-          end
+        begin_pos = begin_pos_with_comment(node)
+        end_pos = end_position_for(node)
+        end_pos += 1 if node.def_type?
```
dvandersluis, line 12: As mentioned, this method was incorrectly getting a range for not just the `def`, but all `def`s within the same visibility group. I've fixed it to just return the `def` and its preceding comments.

bbatsov:
> I could not find a way to add offenses on each def but only autocorrect once, so I decided to move the offense to the sclass itself. I hope this is an okay tradeoff; please let me know if there's a better way.

That's my only concern about the PR, but I feel that the tradeoff makes sense. @koic @marcandre Any thoughts from you?

marcandre:
> > I could not find a way to add offenses on each def but only autocorrect once, so I decided to move the offense to the sclass itself. I hope this is an okay tradeoff; please let me know if there's a better way.
> 
> That's my only concern about the PR, but I feel that the tradeoff makes sense. @koic @marcandre Any thoughts from you?

That's a reasonable way to do it. OTOH, it should be reasonably easy to add offenses to the `def` nodes:

```ruby
def on_new_investigation
  @already_corrected_classes = Set[]
end

# ...
  add_offense(def_node) {
    class_node = class_for_def(def_node)
    next unless @already_corrected_classes.add?(class_node)

    # actual auto-correction
  end
```

dvandersluis:
@marcandre I wanted to collect the `def`s so I could move them out of the `class << self` and spread them out properly which I couldn’t figure out when moving them one at a time (and I was getting clobbering errors too). 

If you want me to try again I will but if it is a reasonable solution as-is, that’s ok by me 😅

marcandre:
Oh, sorry, I didn't understand completely.

I'm surprised you got clobbering errors, as I imagine your corrections were a bunch of `remove` of the original `def`s and a bunch of `insert` of the modified def after the `sclass`' end. The corrector should have no issue inserting texts at the same location a bunch of times.

The main difference for the user is where to put `rubocop:disable`. It doesn't sound too bad that the decision has to be per `sclass` and not per `def`.

In short: I'm 👍 either way. If you prefer the multiple offense route and run into difficulties with clobbering, I can have a look.

dvandersluis:
@marcandre I think it was because I was trying to keep the defs in the same order as they originally were in and also retain the new lines between them. Could be I was doing something wrong though. 

Looks like @bbatsov merged it — thanks! — but if you want me to take another stab at it let me know.

dvandersluis:
Oh and the other problem was removing the empty `sclass` when the offenses/autocorrecting was per `def`.

marcandre:
> @marcandre I think it was because I was trying to keep the defs in the same order as they originally were in and also retain the new lines between them. Could be I was doing something wrong though.

Right, I imagine it can be a bit tricky.

> Oh and the other problem was removing the empty `sclass` when the offenses/autocorrecting was per `def`.

"swallowed deletions" are ignored so should not have been an issue to delete it on the last offense.

> Looks like @bbatsov merged it — thanks! — but if you want me to take another stab at it let me know.

We're all glad it works, that's the most important! Thanks for this 👍

bbatsov:
> Looks like @bbatsov merged it — thanks! — but if you want me to take another stab at it let me know.

Up to you. I merged it mostly because I plan a new release later today and even with some quirks that PR improves what we had before. It's never late for more improvements. :-) Thanks for tackling this!

PR #8694, opened 2020-09-10
[Fix #8673] Fix the JSON parse error when specifying `--format=json` and `--stdin`

### LINKED ISSUE #8673: Using the `--fix` option with `--stdin` appends file content to the json result
When I run `rubocop` using `--fix` and `--stdin` flags, I can see that the content of the file is appended to the JSON result.
This prevent tools (like RubyMine) to parse the response (https://youtrack.jetbrains.com/issue/RUBY-27006)

## Expected behavior

I would expect to have only the JSON result.

## Actual behavior

Example:
```
❯ rubocop --format json --force-exclusion --fix --stdin application_controller.rb < app/controllers/application_controller.rb 
{"metadata":{"rubocop_version":"0.90.0","ruby_engine":"ruby","ruby_version":"2.7.0","ruby_patchlevel":"0","ruby_platform":"x86_64-darwin19"},"files":[{"path":"application_controller.rb","offenses":[]}],"summary":{"offense_count":0,"target_file_count":1,"inspected_file_count":1}}====================
class ApplicationController < ActionController::API
end
```

Just run the following in a Rails project:

`rubocop --format json --force-exclusion --fix --stdin application_controller.rb < app/controllers/application_controller.rb `

## RuboCop version

```
❯ bundle exec rubocop -V              
0.90.0 (using Parser 2.7.1.4, rubocop-ast 0.3.0, running on ruby 2.7.0 x86_64-darwin19)
```

koic:
Fixes #8673.

This PR returns only JSON result when specifying `--format=json` and `--stdin` to fix the JSON parse error.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cli/command/execute_runner.rb
+++ b/lib/rubocop/cli/command/execute_runner.rb
@@ -8,6 +8,9 @@ module RuboCop
       class ExecuteRunner < Base
         include Formatter::TextUtil
 
+        # Combination of short and long formatter names.
+        INTEGRATION_FORMATTERS = %w[h html j json ju junit].freeze
+
         self.command_name = :execute_runner
 
         def run
@@ -61,6 +64,11 @@ module RuboCop
         end
 
         def maybe_print_corrected_source
+          # Integration tools (like RubyMine) expect to have only the JSON result
+          # when specifying JSON format. Similar HTML and JUnit are targeted as well.
+          # See: https://github.com/rubocop-hq/rubocop/issues/8673
+          return if INTEGRATION_FORMATTERS.include?(@options[:format])
+
           # If we are asked to autocorrect source code read from stdin, the only
           # reasonable place to write it is to stdout
           # Unfortunately, we also write other information to stdout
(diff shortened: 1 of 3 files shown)


lib/rubocop/cli/command/execute_runner.rb
```diff
@@ -61,6 +61,11 @@ def display_error_summary(errors)
         end
 
         def maybe_print_corrected_source
+          # Integration tools (like RubyMine) expect to have only the JSON result
+          # when specifying JSON format.
+          # See: https://github.com/rubocop-hq/rubocop/issues/8673
+          return if @options[:format] == 'json'
```
bbatsov, line 67: Btw, shouldn't we do the same for the HTML formatter as well? Might be weird to append plain text to it.

koic: Sure! I think also JUnit formatter is similar, so I added HTML and JUnit formatters.

bbatsov:
Thanks!

PR #8701, opened 2020-09-11
[Fix #8621] Helpful Infinite Loop message

### LINKED ISSUE #8621: Infinite Loop message is not helpful
When there's a infinite loop, it would be helpful to know which Cop(s) are the culprit, i.e. which ones are still trying to do auto-corrections.

iSarCasm:
Closes #8621

Adds the information about which Cops exactly caused an infinite loop

Tracks the corrected offenses on each iteration. 

In the case of exceeding `MAX_ITERATIONS`, it just returns cop names from the latest iteration.
In the case of duplicated sources, it returns a list of iterations (with cop names) starting from the first time this exact source has already been generated. In most cases, I imagine there would only be 2 or 3 iterations between looped states but it should handle all cases.

Examples:
- `Test/ClassMustBeAModuleCop -> Test/ModuleMustBeAClassCop`
- `Test/ClassMustBeAModuleCop, Test/AtoB -> Test/ModuleMustBeAClassCop, Test/BtoA`
- `Test/AtoB -> Test/BtoC -> Test/CtoA`

-----------------

* [X] Added tests.

### DIFF
--- a/lib/rubocop/runner.rb
+++ b/lib/rubocop/runner.rb
@@ -11,9 +11,12 @@ module RuboCop
     class InfiniteCorrectionLoop < RuntimeError
       attr_reader :offenses
 
-      def initialize(path, offenses)
-        super "Infinite loop detected in #{path}."
-        @offenses = offenses
+      def initialize(path, offenses_by_iteration, loop_start: -1)
+        @offenses = offenses_by_iteration.flatten.uniq
+        root_cause = offenses_by_iteration[loop_start..-1]
+                     .map { |x| x.map(&:cop_name).uniq.join(', ') }
+                     .join(' -> ')
+        super "Infinite loop detected in #{path} and caused by #{root_cause}"
       end
     end
 
@@ -236,18 +239,21 @@ module RuboCop
 
     def do_inspection_loop(file)
       processed_source = get_processed_source(file)
-      offenses = []
+      # This variable is 2d array used to track corrected offenses after each
+      # inspection iteration. This is used to output meaningful infinite loop
+      # error message.
+      offenses_by_iteration = []
 
       # When running with --auto-correct, we need to inspect the file (which
       # includes writing a corrected version of it) until no more corrections
       # are made. This is because automatic corrections can introduce new
       # offenses. In the normal case the loop is only executed once.
-      iterate_until_no_changes(processed_source, offenses) do
+      iterate_until_no_changes(processed_source, offenses_by_iteration) do
         # The offenses that couldn't be corrected will be found again so we
         # only keep the corrected ones in order to avoid duplicate reporting.
-        offenses.select!(&:corrected?)
+        !offenses_by_iteration.empty? && offenses_by_iteration.last.select!(&:corrected?)
         new_offenses, updated_source_file = inspect_file(processed_source)
-        offenses.concat(new_offenses).uniq!
+        offenses_by_iteration.push(new_offenses)
 
         # We have to reprocess the source to pickup the changes. Since the
         # change could (theoretically) introduce parsing errors, we break the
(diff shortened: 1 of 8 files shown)


bbatsov:
The proposed changes look good to me overall. Let's also hear what @marcandre and @koic think about them, though.

spec/support/cops/c_to_a_cop.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Test
+      class CtoA < RuboCop::Cop::Base
```
bbatsov, line 6: Might be a good idea to a line to each cop mentioning it's used only for infinite loop tests.

lib/rubocop/runner.rb
```diff
@@ -236,18 +239,18 @@ def save_in_cache(cache, offenses)
 
     def do_inspection_loop(file)
       processed_source = get_processed_source(file)
-      offenses = []
+      offenses_by_iteration = []
```
bbatsov, line 242: You might want to explain a bit that variable, as it took me a while to figure out what exactly does it track.

lib/rubocop/runner.rb
```diff
@@ -11,9 +11,12 @@ class Runner # rubocop:disable Metrics/ClassLength
     class InfiniteCorrectionLoop < RuntimeError
       attr_reader :offenses
 
-      def initialize(path, offenses)
-        super "Infinite loop detected in #{path}."
-        @offenses = offenses
+      def initialize(path, offenses_by_iteration, loop_start = -1)
```
bbatsov, line 14: These days I always use keyword params instead of optional params, but in this case it's not a big deal.

marcandre: Agreed, but I forgot to mark the constructor as `@api private`.

marcandre (approved):
Fabulous! ♥️

Only nitpick is that I would convert an array to string with `.join(', ')` instead of `.to_s`. In the typical cases, `Some/Cop -> Some/OtherCop` is nicer than `[Some/Cop] -> [Some/OtherCop]`.

iSarCasm:
> Fabulous! ♥️
> 
> Only nitpick is that I would convert an array to string with `.join(', ')` instead of `.to_s`. In the typical cases, `Some/Cop -> Some/OtherCop` is nicer than `[Some/Cop] -> [Some/OtherCop]`.

How should it look like when there are more than 1 cop? In case of `["Test/ClassMustBeAModuleCop", "Test/AtoB"] -> ["Test/ModuleMustBeAClassCop", "Test/BtoA"]` for example

iSarCasm:
> Fabulous! ♥️
> 
> Only nitpick is that I would convert an array to string with `.join(', ')` instead of `.to_s`. In the typical cases, `Some/Cop -> Some/OtherCop` is nicer than `[Some/Cop] -> [Some/OtherCop]`.

So in case of more than 1 conflicting cop it should be `Test/ClassMustBeAModuleCop, Test/AtoB ->Test/ModuleMustBeAClassCop, Test/AtoB ` instead of `["Test/ClassMustBeAModuleCop", "Test/AtoB"] -> ["Test/ModuleMustBeAClassCop", "Test/BtoA"]`, right?

spec/support/cops/c_to_a_cop.rb
iSarCasm: Added class comment to `AtoB`, `BtoC`, `CtoA` and `BtoA`

lib/rubocop/runner.rb
iSarCasm: Added a variable explanation comment

lib/rubocop/runner.rb
iSarCasm: Changed to `def initialize(path, offenses_by_iteration, loop_start: -1)`

marcandre:
> P.S: Updated it already

Cool, LGTM 👍

koic (approved):
This looks good to me. I think this information can be a good hint to solve the infinite loop error 👍

bbatsov:
Thanks!

PR #8702, opened 2020-09-11
Add new cop `Lint/UselessTimes`

dvandersluis:
This cop looks for pointless uses of `Integer#times` and replaces or removes them as necessary (I found some of these in my codebase at work, sadly). When the integer is <= 0, the block will never be yielded to, so the entire thing can be removed. When the integer is 1, the block will only ever yield once, so the `1.times` call can (generally) be replaced with the block body.

Auto-correction is enabled in most cases, but because `Integer#times` returns its receiver (and the replacement won't), I've marked it as unsafe.

The cop will remove `0.times` and `-5.times` (etc.) calls, since they will never do anything. The cop will replace `1.times` with its block, and if the block arg is used at all, replace it with 1.

Auto-correction will not be attempted in a couple more complex cases, such as mutating the block arg within the block, or calling `Integer#times` not on its own line. See the tests for more details.

-----------------

* [x] Added tests.


bbatsov:
Great work!

PR #8707, opened 2020-09-12
Add new Lint/ConstantDefinitionInBlock cop

eugeneius:
Constants defined within a block are added to the current lexical scope, which means the block has no effect on where the constant is defined or how it can be accessed. I often see constants defined inside Rake tasks or RSpec `describe` blocks where the intent was clearly for them to be scoped to the block, but which actually leak into the top-level scope.

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/constant_defined_in_block.rb
```diff
@@ -0,0 +1,54 @@
...
+      #   # good
+      #   task :lint do
+      #     files_to_lint = Dir['lib/*.rb']
+      #   end
+      #
+      #   # good
+      #   describe 'making a request' do
+      #     let(:test_request) { Class.new }
+      #   end
+      class ConstantDefinedInBlock < Base
```
bbatsov, line 32: Probably we should name this `ConstantDefinitionInBlock` to be more consistent with existing cop names.

bbatsov:
Nice cop! I think we should also mention this in the Ruby Style Guide, as it might not be super obvious to some people.

lib/rubocop/cop/lint/constant_defined_in_block.rb
eugeneius: Good call, I updated the name.

eugeneius:
I opened https://github.com/rubocop-hq/ruby-style-guide/pull/841 to add a corresponding guideline to the Ruby Style Guide.

bbatsov:
Thanks! I'll merge the cop as it is, as I might cut a new release today, but we should add a link to the new guideline to the cop metadata.

ghiculescu:
FYI for anyone who comes across this, this cop isn't compatible with `sorbet`. Specifically, constants defined in blocks are [how enums work](https://sorbet.org/docs/tenum).

@marcandre do you want me to do a docs PR for this one too?

marcandre:
> @marcandre do you want me to do a docs PR for this one too?

Yes, that would be a great starting point. At a minimum we should document these cases, and afterwards see what is the best to address them. In this case, we could exclude `enum do`, or have an allowlist, for example...

PR #8729, opened 2020-09-16
Use token helpers from rubocop-ast

fatkodima:
Followup of https://github.com/rubocop-hq/rubocop/pull/8450
This pr uses new apis from https://github.com/rubocop-hq/rubocop-ast/pull/92

cc @marcandre


lib/rubocop/cop/layout/space_inside_hash_literal_braces.rb
```diff
@@ -85,12 +85,13 @@ def on_hash(node)
         private
 
         def hash_literal_with_braces(node)
-          tokens = processed_source.tokens
-          begin_index = index_of_first_token(node)
-          return unless tokens[begin_index].left_brace?
+          begin_index = processed_source.send(:first_token_index, node)
```
fatkodima, line 88: We should have made this methods public also, I think.

marcandre: Maybe, but I am not sure this is a good example here. Why not call `tokens_within`, use those and yield those?

fatkodima: Agreed, updated.
Do you have an idea on why CI fails on windows only?

lib/rubocop/cop/layout/space_inside_hash_literal_braces.rb
```diff
@@ -72,29 +72,15 @@ class SpaceInsideHashLiteralBraces < Base
...
+          return unless tokens.first.left_brace? && tokens.last.right_curly_brace?
 
-          hash_literal_with_braces(node) do |begin_index, end_index|
-            check(tokens[begin_index], tokens[begin_index + 1])
-            return if begin_index == end_index - 1
-
-            check(tokens[end_index - 1], tokens[end_index])
-          end
+          check(tokens[0], tokens[1])
+          check(tokens[-2], tokens[-1]) if tokens.size > 1
```
marcandre, line 79: Previous condition translates to `if tokens.size > 2`, no? (technically `!= 2` but can't be < 2, right?). Still I don't see how that can create an issue, it's just an optimization...

lib/rubocop/cop/layout/space_inside_hash_literal_braces.rb
marcandre: Not sure and I've got to run 🏃‍♂️

lib/rubocop/cop/layout/space_inside_hash_literal_braces.rb
fatkodima: Changed to `> 2`, but the same offense.
Funny, that I tested on ruby 2.7 on mac and saw 1 offense on one of the test files, but on CI (non windows) no offenses.

fatkodima: Ok, finally, I have figured it out - https://github.com/rubocop-hq/rubocop-ast/pull/116

The problem is that sorting is not stable (and this sometimes  can change the relative order of tokens like `@text=""  @pos=123...123` and `@text="}" @pos=123..124`) and this can lead to incorrect tokens returned by `ProcessedSource#tokens_within`.

lib/rubocop/cop/layout/space_inside_hash_literal_braces.rb
marcandre: Released 0.4.2; if you bump the requirement this should be good to go 👍

PR #8737, opened 2020-09-16
Use ResbodyNode#exceptions

fatkodima:


marcandre (changes requested):
LGTM, just worried about incompatibilities.

Could we keep the implementation in the mixin and either add a warning or mark it as `# @deprecated` instead?

fatkodima:
> Could we keep the implementation in the mixin and either add a warning or mark it as # @deprecated instead?

Done.

marcandre:
Awesome, thanks!

PR #8738, opened 2020-09-16
Enable autocorrection for `Style/DateTime`

dvandersluis:
`Style/DateTime` did not have autocorrection enabled, so I added it to replace `DateTime` with `Time`. I have marked this as unsafe because we are not checking if the date being used is historic or not (unless a `start` parameter is passed in, such as in `iso8601`, in which case an offense isn't registered anyways).

I did not enable autocorrection for `to_datetime` because the message just states to not use it.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/date_time.rb
+++ b/lib/rubocop/cop/style/date_time.rb
@@ -42,6 +42,8 @@ module RuboCop
       #   # good
       #   something.to_time
       class DateTime < Base
+        extend AutoCorrector
+
         CLASS_MSG = 'Prefer Time over DateTime.'
         COERCION_MSG = 'Do not use #to_datetime.'
 
@@ -63,13 +65,22 @@ module RuboCop
           return if historic_date?(node)
 
           message = to_datetime?(node) ? COERCION_MSG : CLASS_MSG
-          add_offense(node, message: message)
+
+          add_offense(node, message: message) do |corrector|
+            autocorrect(corrector, node)
+          end
         end
 
         private
 
         def disallow_coercion?
           !cop_config['AllowCoercion']
+        end
+
+        def autocorrect(corrector, node)
+          return if to_datetime?(node)
+
+          corrector.replace(node.receiver.loc.name, 'Time')
         end
       end
     end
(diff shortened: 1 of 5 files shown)


marcandre (approved):
LGTM

lib/rubocop/cop/style/date_time.rb
```diff
@@ -63,14 +65,24 @@ def on_send(node)
...
         private
 
         def disallow_coercion?
           !cop_config['AllowCoercion']
         end
+
+        def autocorrect(corrector, node)
+          return if to_datetime?(node)
+
+          replacement = node.receiver.source.start_with?('::') ? '::Time' : 'Time'
```
koic, line 83: I think it can be inspected without comparing strings.

```suggestion
          replacement = node.receiver.namespace&.cbase_type? ? '::Time' : 'Time'
```

dvandersluis: @koic sounds good, made the change. I was trying to use `ConstNode#absolute?` but was getting a `NoMethodError`, but this works!

marcandre: Actually, just replace the `node.receiver.short_name` with `Time` and that will handle both cases...

dvandersluis: @marcandre awesome, updated!

dvandersluis:
@koic @marcandre would this be okay to merge now?

koic:
The next release may be 0.91.1 (bug fix) and I'm not sure about next release version. I would like to merge when the next release is decided to be 0.92.

lib/rubocop/cop/style/date_time.rb
koic: @dvandersluis Maybe git push is forgotten?

dvandersluis:
@koic makes perfect sense!

lib/rubocop/cop/style/date_time.rb
dvandersluis: @koic updated again, it should be there again now! Sorry about that 🤦

koic: @dvandersluis Thank you!

koic (approved):

bbatsov:
I'll cut the new release later today.

dvandersluis:
@bbatsov updated for post 0.91.1.

bbatsov:
Thanks!

PR #8754, opened 2020-09-20
Ignore non-integer ranges in Style/RandomWithOffset

eugeneius:
This cop only works correctly for ranges with literal integer bounds.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/random_with_offset.rb
+++ b/lib/rubocop/cop/style/random_with_offset.rb
@@ -36,7 +36,7 @@ module RuboCop
             (send
               {nil? (const {nil? cbase} :Random) (const {nil? cbase} :Kernel)}
               :rand
-              {int irange erange}))
+              {int (irange int int) (erange int int)}))
         PATTERN
 
         def_node_matcher :rand_op_integer?, <<~PATTERN
@@ -44,7 +44,7 @@ module RuboCop
             (send
               {nil? (const {nil? cbase} :Random) (const {nil? cbase} :Kernel)}
               :rand
-              {int irange erange})
+              {int (irange int int) (erange int int)})
             {:+ :-}
             int)
         PATTERN
@@ -54,7 +54,7 @@ module RuboCop
             (send
               {nil? (const {nil? cbase} :Random) (const {nil? cbase} :Kernel)}
               :rand
-              {int irange erange})
+              {int (irange int int) (erange int int)})
             {:succ :pred :next})
         PATTERN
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/random_with_offset.rb
```diff
@@ -36,15 +36,15 @@ class RandomWithOffset < Base
             (send
               {nil? (const {nil? cbase} :Random) (const {nil? cbase} :Kernel)}
               :rand
-              {int irange erange}))
+              {int (irange (int _) (int _)) (erange (int _) (int _))}))
```
marcandre, line 39: No biggie, but `(int _)` => `int` is simpler

marcandre (commented):
I'm wondering if it wouldn't be better to mark this cop as unsafe instead of restricting it like this.
I'm a bit confused as to which cases are actually problematic...

lib/rubocop/cop/style/random_with_offset.rb
eugeneius: Thanks, updated!

eugeneius:
The previous behaviour was a crash, so I don't think marking as unsafe will cut it:

```
$ rubocop _0.91.0_ --debug --only Style/RandomWithOffset --stdin foo.rb <<< "rand(0..limit) + 1"
For /Users/eugene/test: Default configuration from /Users/eugene/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rubocop-0.91.0/config/default.yml
Inspecting 1 file
Scanning /Users/eugene/test/foo.rb
An error occurred while Style/RandomWithOffset cop was inspecting /Users/eugene/test/foo.rb:1:0.
undefined method `+' for nil:NilClass
/Users/eugene/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rubocop-0.91.0/lib/rubocop/cop/style/random_with_offset.rb:112:in `block in corrected_rand_op_integer'
/Users/eugene/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rubocop-0.91.0/lib/rubocop/cop/style/random_with_offset.rb:95:in `random_call'
/Users/eugene/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rubocop-0.91.0/lib/rubocop/cop/style/random_with_offset.rb:105:in `corrected_rand_op_integer'
/Users/eugene/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rubocop-0.91.0/lib/rubocop/cop/style/random_with_offset.rb:83:in `autocorrect'
/Users/eugene/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rubocop-0.91.0/lib/rubocop/cop/style/random_with_offset.rb:68:in `block in on_send'
```

This happens because the range bounds are assumed to be integer literals here later:

https://github.com/rubocop-hq/rubocop/blob/760d98ba562de63f66cfd789b77d065faf9e664f/lib/rubocop/cop/style/random_with_offset.rb#L147-L149

It might be possible to extend the cop to accept non-integer bounds (perhaps unsafely), but I think that can be a followup.

marcandre:
Cool, thanks!

PR #8766, opened 2020-09-22
[Fix #8764] Fix `Layout/CaseIndentation` to properly output annotated error messages

### LINKED ISSUE #8764: Layout/CaseIndentation: the name of rule is not printed in description
When `Layout/CaseIndentation` is detected, its name is not printed in description. Other rules it have in their description.

## Expected behavior

```
app/helpers/test_helper.rb:25:7: C: Layout/CaseIndentation: Indent when as deep as case.
    when :small then 16
    ^^^^
```

## Actual behavior

```
app/helpers/test_helper.rb:25:7: C: Indent when as deep as case.
    when :small then 16
    ^^^^
```

```ruby
    w_h = case size
      when :small then 16
      when :big then 32
      when :huge then 64
      else 24
          end
```

Run

```
rubocop --only Layout/CaseIndentation
```

## RuboCop version

```
$ rubocop -V
0.91.0 (using Parser 2.7.1.4, rubocop-ast 0.4.0, running on ruby 2.7.1 x86_64-darwin18)
```

dvandersluis:
`find_message` is overridden in `Layout::CaseIndentation` but did not actually annotate the message, so the cop name was not being shown in the output.

Fixes #8764.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/case_indentation.rb
+++ b/lib/rubocop/cop/layout/case_indentation.rb
@@ -105,7 +105,10 @@ module RuboCop
         end
 
         def incorrect_style(when_node)
-          add_offense(when_node.loc.keyword) do |corrector|
+          depth = indent_one_step? ? 'one step more than' : 'as deep as'
+          message = format(MSG, depth: depth, base: style)
+
+          add_offense(when_node.loc.keyword, message: message) do |corrector|
             detect_incorrect_style(when_node)
 
             whitespace = whitespace_range(when_node)
@@ -123,12 +126,6 @@ module RuboCop
           else
             unrecognized_style_detected
           end
-        end
-
-        def find_message(*)
-          depth = indent_one_step? ? 'one step more than' : 'as deep as'
-
-          format(MSG, depth: depth, base: style)
         end
 
         def base_column(case_node, base)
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/layout/case_indentation.rb
```diff
@@ -128,7 +128,7 @@ def detect_incorrect_style(when_node)
         def find_message(*)
           depth = indent_one_step? ? 'one step more than' : 'as deep as'
 
-          format(MSG, depth: depth, base: style)
+          super(nil, format(MSG, depth: depth, base: style))
```
koic, line 131: Thanks for the fixing. IMHO, there seems to be no reason for `find_message(*)` to be overridden. Can you remove this `find_message(*)` override definition and make it inline method?

```diff
diff --git a/lib/rubocop/cop/layout/case_indentation.rb b/lib/rubocop/cop/layout/case_indentation.rb
index 090bc73d7..5e6afb52f 100644
--- a/lib/rubocop/cop/layout/case_indentation.rb
+++ b/lib/rubocop/cop/layout/case_indentation.rb
@@ -105,7 +105,10 @@ module RuboCop
         end

         def incorrect_style(when_node)
-          add_offense(when_node.loc.keyword) do |corrector|
+          depth = indent_one_step? ? 'one step more than' : 'as deep as'
+          message = format(MSG, depth: depth, base: style)
+
+          add_offense(when_node.loc.keyword, message: message) do |corrector|
             detect_incorrect_style(when_node)

             whitespace = whitespace_range(when_node)
@@ -125,12 +128,6 @@ module RuboCop
           end
         end

-        def find_message(*)
-          depth = indent_one_step? ? 'one step more than' : 'as deep as'
-
-          super(nil, format(MSG, depth: depth, base: style))
-        end
-
         def base_column(case_node, base)
           case base
           when :case then case_node.location.keyword.column
```

dvandersluis: Definitely!

dvandersluis: @koic I've updated it now, thanks for the suggestion! 🙌

koic:
Thank you!

PR #8768, opened 2020-09-22
Fix a false positive for `Style/RedundantReturn` when a rescue has an else clause

### LINKED ISSUE #8767: Style/RedundantReturn removes non-redundant returns from begin-rescue-else-end blocks
When the `begin` block of a `begin-rescue-else-end` contains an explicit `return`, the `else` block is not executed.  Removing that `return` causes the `else` block to be executed.  Thus, the `return` is not redundant.  However, rubocop flags it as redundant.

--------

## Expected behavior

Rubocop should not consider the `return` redundant, since removing it alters behavior

## Actual behavior

Rubocop considers the `return` to be redundant, and removes it when the `--auto-correct` flag is supplied

```
$ cat test.rb
# frozen_string_literal: true

def try_to_return_true
  return true
rescue NoMethodError
  puts 'I can only return false'
  false
else
  puts 'I thought I already returned true'
  5
end

puts try_to_return_true
$ ruby test.rb 
true
$ rubocop --auto-correct test.rb 
Inspecting 1 file
C

Offenses:

test.rb:4:3: C: [Corrected] Style/RedundantReturn: Redundant return detected.
  return true
  ^^^^^^

$ ruby test.rb 
I thought I already returned true
5
```

## RuboCop version

```
$ rubocop -V
0.91.0 (using Parser 2.7.1.4, rubocop-ast 0.4.2, running on ruby 2.6.0 x86_64-linux)
```

fatkodima:
Closes #8767

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 * [#8720](https://github.com/rubocop-hq/rubocop/issues/8720): Fix an error for `Lint/IdentityComparison` when calling `object_id` method without receiver in LHS or RHS. ([@koic][])
+* [#8767](https://github.com/rubocop-hq/rubocop/issues/8767): Fix a false positive for `Style/RedundantReturn` when a rescue has an else clause. ([@fatkodima][])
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

PR #8778, opened 2020-09-23
Add `--regenerate-todo` command line option to rebuild .rubocop_todo.yml using its previous options

dvandersluis:
Added a new command line option `--regenerate-todo` which regenerates `.rubocop_todo.yml` using the options used the last time (extracted from the comment at the top of the todo file), or act as `--auto-gen-config` if there is no current valid todo file.

At my work, we frequently re-run `rubocop --auto-gen-config` with various options in order to keep the todo file small and relevant. Each time, we have to copy the command from the previous version of the file in order to have it use the same options. Sometimes this is accidentally missed, which results in the todo file completely changing due to the changed options. 

With this change I'm trying to prevent that, by just having rubocop reuse its previous config generation options when requested by just running `rubocop --regenerate-todo`.

```console
rubocop --auto-gen-config --no-auto-gen-timestamp --exclude-limit 1000

# ... later ...

rubocop --regenerate-todo
# uses exclude_limit: 1000, auto_gen_timestamp: false
```

I also made it so that it's possible to purposefully override the options with `--regenerate-todo` by passing them in as well. To facilitate this, I changed `--no-offense-counts`, `--auto-gen-only-exclude`, and `--no-auto-gen-timestamp` to have both positive and negative variants, so this will work for example:

```console
rubocop --regenerate-todo --auto-gen-timestamp
# uses exclude_limit: 1000, auto_gen_timestamp: true
```

-----------------

* [x] Added tests.


bbatsov:
I love this change! Great work! 🙇‍♂️

PR #8784, opened 2020-09-24
Fix #8650 for improved hidden files finder performance

### LINKED ISSUE #8650: Faster Find of hidden files
In profiling the delay between rubocop startup and the beginning of scanning files I discovered that rubocop will traverse the project directories/files twice.  Once for `all_files` and then again to find `hidden_files`.

```ruby
# lib/rubocop/target_finder.rb
def target_files_in_dir(base_dir = Dir.pwd)
  # ...
  all_files = find_files(base_dir, File::FNM_DOTMATCH)
  hidden_files = Set.new(all_files - find_files(base_dir, 0))
  #...
end

#...
def find_files(base_dir, flags)
  # ...
  Dir.glob(pattern, flags).select { |path| FileTest.file?(path) }
end
```

In the code above, `#find_files` is called twice which calls Dir.glob.  Each time with a different flag.  This is ineficient as it's touching the local disk twice for ever non-hidden file and it's checking every entry0for `FileTest.file?(path)` twice . 

A more efficient solution would be to only call `#find_files` once and then filter for hidden files.

```diff
- hidden_files = Set.new(all_files - find_files(base_dir, 0))
+ hidden_files = all_files.select {|file|file =~ /[\/|\\]\./ }
```

The regex looks for any file paths which include a '/.' (Linux) or ''\\.'' (Windows)

## Benchmarks

I tested this code change in a medium size Rails project.

### Project Stats Stats:

```
all_files: 123,727
hidden_files: 2,553
target_files: 405
```
Note: The majority of the files are in engines/my-engine/node_modules directory (which is a separate issue).

```ruby
require "benchmark/ips"

Benchmark.ips(7) do |x|
  x.report("glob:") { hidden_files = Set.new(all_files - find_files(base_dir, 0)) }
  x.report("select:") { hidden_files = all_files.select {|file|file =~ /\/\./ } }
  x.compare!
end
```

### Results:

```
Comparison:
              select:       23.6 i/s
                glob:        1.0 i/s - 23.99x  (± 0.00) slower

```

The above IPS benchmark shows that on this project the 'select' code can ran 23 times within 1 second, whereas the glob could only run once within a second.  In other words, glob is 23.99x slower.

## Additional context

If this code change is desired, I can create an MR.

tleish:
Faster find of hidden files in `TargetFinder` class which improves rubocop initial startup speed.

Benchmarks now show a 57x improvement in one large project tested:
```
Comparison:
             select:       50.6 i/s
               glob:        0.9 i/s - 57.38x  (± 0.00) slower
```

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/target_finder.rb
+++ b/lib/rubocop/target_finder.rb
@@ -5,6 +5,8 @@ module RuboCop
   # and picking ruby files.
   # @api private
   class TargetFinder
+    HIDDEN_PATH_SUBSTRING = "#{File::SEPARATOR}."
+
     def initialize(config_store, options = {})
       @config_store = config_store
       @options = options
@@ -55,7 +57,8 @@ module RuboCop
       # Support Windows: Backslashes from command-line -> forward slashes
       base_dir = base_dir.gsub(File::ALT_SEPARATOR, File::SEPARATOR) if File::ALT_SEPARATOR
       all_files = find_files(base_dir, File::FNM_DOTMATCH)
-      hidden_files = Set.new(all_files - find_files(base_dir, 0))
+      # use file.include? for performance optimization
+      hidden_files = all_files.select { |file| file.include?(HIDDEN_PATH_SUBSTRING) }
       base_dir_config = @config_store.for(base_dir)
 
       target_files = all_files.select do |file|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changes
 
 * [#8785](https://github.com/rubocop-hq/rubocop/pull/8785): Update TargetRubyVersion 2.8 to 3.0 (experimental). ([@em-gazelle][])
+* [#8650](https://github.com/rubocop-hq/rubocop/issues/8650): Faster find of hidden files in `TargetFinder` class which improves rubocop initial startup speed. ([@tleish][])
 
 ## 0.91.1 (2020-09-23)
 
@@ -4912,3 +4913,4 @@
 [@fsateler]: https://github.com/fsateler
 [@iSarCasm]: https://github.com/iSarCasm
 [@em-gazelle]: https://github.com/em-gazelle
+[@tleish]: https://github.com/tleish


marcandre:
57x is nice 😆
Could you not use `grep` with a regexp here?

tleish:
> Could you not use `grep` with a regexp here?

@marcandre - I previously tested this, but turns out that the proposed solution is faster.

```ruby
      require "benchmark/ips"

      Benchmark.ips(7) do |x|
        x.report("glob") { hidden_files = Set.new(all_files - find_files(base_dir, 0)) }
        x.report("select") { hidden_files = all_files.select { |file| file.include?(HIDDEN_FILE_PATTERN) } }
        x.report("grep") { hidden_files = all_files.grep(%r{#{File::SEPARATOR}\.}) }
        x.compare!
      end
```

```
Comparison:
             select:       52.3 i/s
               grep:       23.5 i/s - 2.22x  (± 0.00) slower
               glob:        0.9 i/s - 59.16x  (± 0.00) slower
```

marcandre:
> > Could you not use `grep` with a regexp here?
> 
> @marcandre - I previously tested this, but turns out that the proposed solution is faster.

Ah, right... Forgot about https://bugs.ruby-lang.org/issues/17030

tleish:
> Ah, right... Forgot about https://bugs.ruby-lang.org/issues/17030

FYI

```ruby
Benchmark.ips(7) do |x|
    x.report("glob") { hidden_files = Set.new(all_files - find_files(base_dir, 0)) }
    x.report("include?") { hidden_files = all_files.select { |file| file.include?(HIDDEN_FILE_PATTERN) } }
    x.report("=~") { hidden_files = all_files.select { |file| file =~ grep_hidden_files_pattern } }
    x.report("match?") { hidden_files = all_files.select { |file| file.match?(grep_hidden_files_pattern) } }
    x.report("grep") { hidden_files = all_files.grep(grep_hidden_files_pattern) }
    x.compare!
end
```

```
Comparison:
            include?:       50.2 i/s
              match?:       42.9 i/s - 1.17x  (± 0.00) slower
                grep:       24.6 i/s - 2.04x  (± 0.00) slower
                  =~:       22.9 i/s - 2.19x  (± 0.00) slower
                glob:        0.9 i/s - 56.91x  (± 0.00) slower
```

lib/rubocop/target_finder.rb
```diff
@@ -5,6 +5,8 @@ module RuboCop
   # and picking ruby files.
   # @api private
   class TargetFinder
+    HIDDEN_FILE_PATTERN = "#{File::SEPARATOR}."
```
bbatsov, line 8: I'm not sure that naming this "pattern" is a good idea, as it makes me think of regular expressions. :-)

lib/rubocop/target_finder.rb
```diff
@@ -55,7 +57,7 @@ def target_files_in_dir(base_dir = Dir.pwd)
       # Support Windows: Backslashes from command-line -> forward slashes
       base_dir = base_dir.gsub(File::ALT_SEPARATOR, File::SEPARATOR) if File::ALT_SEPARATOR
       all_files = find_files(base_dir, File::FNM_DOTMATCH)
-      hidden_files = Set.new(all_files - find_files(base_dir, 0))
+      hidden_files = all_files.select { |file| file.include?(HIDDEN_FILE_PATTERN) }
```
bbatsov, line 60: I'd add a note we're doing this for performance reasons, as someone might want to optimize it for readability or whatever.

lib/rubocop/target_finder.rb
tleish: updated to `HIDDEN_PATH_SUBSTRING`

lib/rubocop/target_finder.rb
tleish: adde comment

PR #8785, opened 2020-09-24
[Fix #8165] Require Parser gem 2.7.1.5

koic:
Fixes #8165 and Follows https://github.com/whitequark/parser/pull/729.

Parser 2.7.1.5 includes https://github.com/whitequark/parser/pull/717 and fixes #8165. And this PR updates Ruby 2.8 to Ruby 3.0.

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Added an entry to the [Changelog](https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format).
* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: https://chris.beams.io/posts/git-commit/

PR #8786, opened 2020-09-24
Ruby 2.8 is dead. Long live to Ruby 3.0

marcandre:


koic:
I opened the same PR #8785 😅

Issue #8391, opened 2020-07-24
Style/ArrayCoercion auto-correction not equal for nil values


andrewngo:
If a value can be nil, then ArrayCoercion auto-corrected change does not produce an equivalent change. It's a good suggestion but not always valid. 

## Expected behavior
Don't auto-correct unless certain

## Actual behavior
Changes code 

## Steps to reproduce the problem

```
test_value = nil

original_result = [test_value] unless test_value.is_a?(Array)
new_result = Array(test_value)

original_result == new_result # false

# original_result [nil]
# new_result []

```

## RuboCop version

Include the output of `rubocop -V` or `bundle exec rubocop -V` if using Bundler. Here's an example:

```
$ [bundle exec] rubocop -V
0.88.0 (using Parser 2.7.1.4, rubocop-ast 0.2.0, running on ruby 2.6.5 x86_64-darwin16)
```

marcandre:
Agreed, this cop is, at the very least, not safe.

Any object that `responds_to? :to_a` is an issue:

```ruby
Array({a: 1}) # => [[:a, 1]]  != [{a: 1}]
Array(Set[1, 2, 3]) # => [1, 2, 3]  != [Set[1, 2, 3]]
```

I'll make it as unsafe but the issue should remain open: does this cop deserve its badge?

0x0badc0de:
What about using `Array.new(1, value)`?
```
irb(main):001:0> Array.new(1, {a: 1})
=> [{:a=>1}]
irb(main):002:0> Array.new(1, Set[1, 2, 3])
=> [#<Set: {1, 2, 3}>]
irb(main):003:0> Array.new(1, 1..)
=> [1..]
```

0x0badc0de:
Ah, it won't work for arrays args as expected :(

0x0badc0de:
This one seems to work https://stackoverflow.com/a/29431497
```
irb(main):001:0> [{a: 1}].flatten(1)
=> [{:a=>1}]
irb(main):002:0> [Set[1, 2, 3]].flatten(1)
=> [#<Set: {1, 2, 3}>]
irb(main):003:0> [1..].flatten(1)
=> [1..]
irb(main):004:0> [Time.now].flatten(1)
=> [2020-07-27 17:07:10.792690033 +0300]
irb(main):005:0> [[1, 2]].flatten(1)
=> [1, 2]
irb(main):006:0> [[{a:1}, Time.now]].flatten(1)
[...]

Issue #8334, opened 2020-07-14
Style/ArrayCoercion does not work with Time objects


Aqualon:
The Style/ArrayCoercion cop introduced in https://github.com/rubocop-hq/rubocop/pull/8295 has some unexpected auto-correct behavior if dealing with `Time` objects.

## Expected behavior

Style/ArrayCoercion auto-correct should not break code if dealing with `Time` objects.

## Actual behavior

Style/ArrayCoercion auto-correct breaks if the wrapped variable is a `Time` object.

## Steps to reproduce the problem
We have code like this to wrap an `Time` object called `inserted_at` in an Array
```
inserted_at = [inserted_at] unless inserted_at.is_a? Array
```
This was replaced by auto-correct with
```
inserted_at = Array(inserted_at)
```

This leads to an array with each part of the Time object as value
```
inserted_at = Time.now
2020-07-14 14:54:33.339436 +0200
Array(inserted_at)
[33, 54, 14, 14, 7, 2020, 2, 196, true, "CEST"]
```

I don't know why this behavior happens, but it seems just calling Array on some object, my lead to unexpected breaks.

## RuboCop version
```
0.88.0 (using Parser 2.7.1.4, rubocop-ast 0.1.0, running on ruby 2.7.1 x86_64-darwin19)
```

Aqualon:
https://ruby-doc.org/core-2.7.1/Kernel.html#method-i-Array describes Array

> First tries to call to_ary on arg, then to_a. If arg does not respond to to_ary or to_a, returns an Array of length 1 containing arg.

The `Time` class provides to_a: https://ruby-doc.org/core-2.7.1/Time.html#method-i-to_a

So I assume for any object providing a `to_a` method, there may be unexpected behavior, because instead of `[my_object]` you could get whatever `my_object` returns from its `to_a` method.

marcandre:
Indeed. I think the equivalent of `Array(foo)` is `[*foo]`.

bbatsov:
Well, I guess it's safe to say that this cop is unsafe. Still, most of the time it does improve the code a bit.

Aqualon:
I'm not sure I'd call it an improvement if it may change the code to something, that behaves differently than before.

dvandersluis:
I think this should probably be closed now that #8792 is merged?

marcandre:
Indeed 👍

PR #8792, opened 2020-09-25
[Fix #8783] Disable `Style/ArrayCoercion` cop by default

### LINKED ISSUE #8783: Style/ArrayCoercion does not play nicely with Hashes
## Expected behaviour

The Style/ArrayCoercion is not safe because it converts hashes to arrays.

## Actual behaviour

```
values = [values] unless values.is_a?(Array)
# is corrected to
values = Array(values)
```

Which leads to values being:

```
values = {a: 'b'}
values = Array(values) # => [[:a, "b"]] but it should be [{a: 'b'}]
```

## RuboCop version

```
$ bundle exec rubocop -V
0.91.1 (using Parser 2.7.1.4, rubocop-ast 0.4.2, running on ruby 2.6.5 x86_64-linux)
```

koic:
Fixes #8783, #8391, and #8334.

This PR disables `Style/ArrayCoercion` cop. Because false positive will occur if the argument of `Array()` is not an array (e.g. Hash, Set), an array will be returned as an incompatibility result.

-----------------

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 * [#8785](https://github.com/rubocop-hq/rubocop/pull/8785): Update TargetRubyVersion 2.8 to 3.0 (experimental). ([@koic][])
 * [#8650](https://github.com/rubocop-hq/rubocop/issues/8650): Faster find of hidden files in `TargetFinder` class which improves rubocop initial startup speed. ([@tleish][])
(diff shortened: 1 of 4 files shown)


bbatsov:
Please, also add some note to the cop description about why it is disabled by default.

koic:
Sure! I've added a reason to disable this cop by default to the cop description.

PR #8796, opened 2020-09-25
Add new `Lint/HashCompareByIdentity` cop

fatkodima:
This implements a second part of https://rubystyle.guide/#identity-comparison


koic:
Thank you for the implementation. I was also considering this style rule implementation.
I'm not sure if user can always change hash creator which `compare_by_identity` method should be added. So, It may be simpler to split the roles with another `Lint/CompareByIdentity` (?) cop. They also have different auto-correction support. 

Cc @rubocop-hq/rubocop-core

marcandre (changes requested):
I ❤️ it when a cop improves our own code base 🎉

Even though both cases deal with comparing by identity, I wonder if we shouldn't split this into two cops:
1) Some people might not like `compare_by_identity` and want to disable that (but not the use of `equal?`
2) The `==` case is safe (I think 😅), the second case is not quite safe: although unlikely, the hash could store object ids and other stuff that need be compared by values.

Maybe @bbatsov has an opinion on this

*Edit*: Just read @koic's remark, glad to see we agree about splitting this 😄

marcandre (changes requested):
I ❤️ it when a cop improves our own code base 🎉

Even though both cases deal with comparing by identity, I wonder if we shouldn't split this into two cops:
1) Some people might not like `compare_by_identity` and want to disable that (but not the use of `equal?`
2) The `==` case is safe (I think 😅), the second case is not quite safe: although unlikely, the hash could store object ids and other stuff that need be compared by values.

fatkodima:
Extracted into a separate cop.

bbatsov:
Thanks!

PR #8797, opened 2020-09-25
[Fix #8782] Mark any condition with defined? as an unsafe autocorrection if it is not parenthesized

### LINKED ISSUE #8782: Style/TernaryParentheses still not necessarily safe autocorrect for `defined?` keyword
I have a project with this line in it:

```ruby
val = (defined? inc.name) ? inc.name : inc
```

`rubocop -a` autocorrects this to:

```ruby
val = defined? inc.name || inc
```

Which is clearly not the same thing because one returns a Boolean and another returns the `name`.

The correct way to fix this, IMO, to make Rubocop happy:

```ruby
val = defined?(inc.name) ? inc.name : inc
```

I thought this issue was common enough that I would've been able to find someone who had it before, and I did find:

https://github.com/rubocop-hq/rubocop/pull/3505
https://github.com/rubocop-hq/rubocop/issues/3450

But it seems like this problem has persisted...

Thoughts?

## RuboCop version

```
$ bundle exec rubocop -V
0.91.1 (using Parser 2.7.1.4, rubocop-ast 0.4.2, running on ruby 2.7.1 x86_64-darwin19)
```

dvandersluis:
Previously, the check for `defined?` in `unsafe_autocorrect?` only applied when the argument was a single send node with no receiver, but that meant that most times `defined?` was used in a ternary wrapped in parens, it would be autocorrected incorrectly (since `defined?` would then be called on the entire ternary instead of the argument). This change makes `unsafe_autocorrect?` return true if there is any `defined?` call without parentheses.

Fixes #8782.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/ternary_parentheses_spec.rb
+++ b/spec/rubocop/cop/style/ternary_parentheses_spec.rb
@@ -159,6 +159,50 @@ RSpec.describe RuboCop::Cop::Style::TernaryParentheses, :config do
           not (bar) ? a : b
         RUBY
       end
+
+      it 'registers an offense for defined? with variable in condition' do
+        expect_offense(<<~RUBY)
+          foo = defined?(bar) ? a : b
+                ^^^^^^^^^^^^^^^^^^^^^ Use parentheses for ternary conditions.
+        RUBY
+
+        expect_correction(<<~RUBY)
+          foo = (defined?(bar)) ? a : b
+        RUBY
+      end
+
+      it 'registers an offense for defined? with method chain in condition' do
+        expect_offense(<<~RUBY)
+          foo = defined?(bar.baz) ? a : b
+                ^^^^^^^^^^^^^^^^^^^^^^^^^ Use parentheses for ternary conditions.
+        RUBY
+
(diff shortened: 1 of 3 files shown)


dvandersluis:
@marcandre @bbatsov would you be able to take a look at this one too?

spec/rubocop/cop/style/ternary_parentheses_spec.rb
```diff
@@ -159,6 +159,50 @@
           not (bar) ? a : b
         RUBY
       end
+
+      it 'registers an offense for defined with variable in condition' do
```
bbatsov, line 163: defined -> defined?

bbatsov (approved):
Appart from my small remark the changes look good. Sorry about the slow response!

spec/rubocop/cop/style/ternary_parentheses_spec.rb
dvandersluis: Fixed!

dvandersluis:
@bbatsov no worries, updated with your requested changes now!

marcandre:
Thanks!

PR #8798, opened 2020-09-25
[Fix #5713] Be more explicit about how Style/CommentAnnotation works

### LINKED ISSUE #5713: Style/CommentAnnotation reports only first of multiple offending lines
Good morning everyone,

I stumbled upon some unexpected behavior when working on a project. I created a minimal working example [here](https://gist.github.com/jdno/ba9c04ad7f297a1e2745988d6810b6fc).

I am working on a class that has multiple `# TODO ...` annotations. Something like this:

```ruby
class ToBeDone
  ITEMS = [
    '', # TODO Item 1
    '', # TODO Item 2

    '', # TODO Item 3
  ].freeze
end
```

When running Rubocop, it reports a `Style/CommentAnnotation` offense, because the proper style would be `# TODO: ...`. But it does that only for the first and third occurrence, and not for the second one.

```
jandavid@STS ~/D/j/rubocop-todo-bug> rubocop
Inspecting 2 files
C.

Offenses:

todo.rb:5:11: C: Style/CommentAnnotation: Annotation keywords like TODO should be all upper case, followed by a colon, and a space, then a note describing the problem.
    '', # TODO Item 1
          ^^^^^
todo.rb:8:11: C: Style/CommentAnnotation: Annotation keywords like TODO should be all upper case, followed by a colon, and a space, then a note describing the problem.
    '', # TODO Item 3
          ^^^^^

2 files inspected, 2 offenses detected
```

Even after fixing the first and third line, the second line is never reported as an offense. This also means that `rubicon -a` will never pick that line up and fix it automatically.

## Expected behavior

When there are `Style/CommentAnnotation` in multiple lines that follow each other, Rubocop should report a violation for each individual line.

## Actual behavior

When there are `Style/CommentAnnotation` in multiple lines that follow each other, only the first offense is reported.

Check out the [minimal working example](https://gist.github.com/jdno/ba9c04ad7f297a1e2745988d6810b6fc). Download the Gist as a ZIP, extract the archive, execute `bundle` if necessary and run `rubocop` in it.

## RuboCop version

```
jandavid@STS ~/D/j/rubocop-todo-bug> rubocop -V
0.54.0 (using Parser 2.5.0.4, running on ruby 2.5.0 x86_64-darwin17)
```

dvandersluis:
The consensus in the issue is that the cop is working as expected, but the documentation does not explain well enough. This change adds an explicit test for comment blocks, as well as a documentation note.

Closes #5713.

-----------------

* [x] Added tests.

### DIFF
--- a/docs/modules/ROOT/pages/cops_style.adoc
+++ b/docs/modules/ROOT/pages/cops_style.adoc
@@ -1576,6 +1576,12 @@ folders = %x(find . -type d).split
 This cop checks that comment annotation keywords are written according
 to guidelines.
 
+NOTE: With a multiline comment block (where each line is only a
+comment), only the first line will be able to register an offense, even
+if an annotation keyword starts another line. This is done to prevent
+incorrect registering of keywords (eg. `review`) inside a paragraph as an
+annotation.
+
 === Examples
 
 [source,ruby]
--- a/lib/rubocop/cop/style/comment_annotation.rb
+++ b/lib/rubocop/cop/style/comment_annotation.rb
@@ -6,6 +6,12 @@ module RuboCop
       # This cop checks that comment annotation keywords are written according
       # to guidelines.
       #
+      # NOTE: With a multiline comment block (where each line is only a
+      # comment), only the first line will be able to register an offense, even
+      # if an annotation keyword starts another line. This is done to prevent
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #8803, opened 2020-09-27
Handle comments and blank interpolations in regexp parsed_tree

owst:
Pulling out a PR to add this functionality, as mentioned by @marcandre: https://github.com/rubocop-hq/rubocop/pull/8625#discussion_r486044238

-----------------

* [x] Added tests.


rubocop.gemspec
```diff
@@ -36,9 +36,9 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('parallel', '~> 1.10')
   s.add_runtime_dependency('parser', '>= 2.7.1.5')
   s.add_runtime_dependency('rainbow', '>= 2.2.2', '< 4.0')
-  s.add_runtime_dependency('regexp_parser', '>= 1.7')
+  s.add_runtime_dependency('regexp_parser', '>= 1.8')
   s.add_runtime_dependency('rexml')
-  s.add_runtime_dependency('rubocop-ast', '>= 0.5.0')
+  s.add_runtime_dependency('rubocop-ast', '>= 0.6.0')
```
owst, line 41: To pick up https://github.com/rubocop-hq/rubocop-ast/pull/124

rubocop.gemspec
```diff
@@ -36,9 +36,9 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('parallel', '~> 1.10')
   s.add_runtime_dependency('parser', '>= 2.7.1.5')
   s.add_runtime_dependency('rainbow', '>= 2.2.2', '< 4.0')
-  s.add_runtime_dependency('regexp_parser', '>= 1.7')
+  s.add_runtime_dependency('regexp_parser', '>= 1.8')
```
owst, line 39: To pick up https://github.com/ammar/regexp_parser/pull/68 and https://github.com/ammar/regexp_parser/pull/67

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* [#8803](https://github.com/rubocop-hq/rubocop/pull/8803): Extend RegexpNode#parsed_tree to handle regexps including interpolaton and comments. ([@owst][])
```
marcandre, line 7: ~^interpolaton^interpolation
Also, please back-quote RegexpNode#parsed_tree~

Might change altogether, read below

lib/rubocop/ext/regexp_node.rb
```diff
@@ -16,12 +16,16 @@ class << self
       @parsed_cache = {}
 
       # @return [Regexp::Expression::Root, nil]
-      def parsed_tree
-        return if interpolation?
+      def parsed_tree(interpolation: :ignore)
+        unless %i[ignore blank].include?(interpolation)
```
marcandre, line 20: When thinking about the fact that `:ignore` is confusing (it's "ignore the regexp itself", not the interpolations), the only good name I could come up with was `:return_nil` and that's a code smell.

That made me wonder if the best course of action isn't to remove the option altogether, and make blanking the default. What do you think? It should be the callers job to check for interpolation  `lint/mixed_regexp_capture_types` would need to do that).

Should be documented, and marked as breaking change in the Changelog.

lib/rubocop/ext/regexp_node.rb
```diff
@@ -40,6 +44,23 @@ def each_capture(named: ANY)
         self
       end
 
+      private
+
+      def with_interpolations_blanked
+        children.reject(&:regopt_type?).map do |child|
```
marcandre, line 50: There's only one `regopt` at the end, right? If so, `.reject(&:regopt_type?)` => `[0...-1]`.

owst: Good point - I don't think I'd realised that even for a Regexp without options the regopt node was still present. I think the original was better readability-wise, but the new version is probably fine with a comment 👍

lib/rubocop/ext/regexp_node.rb
owst: I agree, I'm not really a fan of `interpolation :ignore`. I like the idea of always blanking (that is, until we find a scenario where it isn't appropriate 😄). 

Having had a quick look at `lint/mixed_regexp_capture_types` I think it could also just use blanking, as we could catch strictly more issues (where the regexp contains mixed capture groups in addition to an interpolation)

I'll take a look at making this change, and report back

marcandre: Oh, you're right about `lint/mixed_regexp_capture_types` 🤦‍♂️

lib/rubocop/ext/regexp_node.rb
marcandre: How about I add a `RegexpNode#content_nodes` to `rubocop-ast`?

lib/rubocop/ext/regexp_node.rb
owst: So for now, to keep this PR focussed, I left `lint/mixed_regexp_capture_types` as is - i.e. it still does nothing in the presence of interpolation. A nice follow-up might be to change it, as I mentioned previously

lib/rubocop/ext/regexp_node.rb
owst: Hmm, my slight concern is that `RegexpNode#content` already exists and ignores interpolation (those with `begin_type?`) nodes, whereas `RegexpNode#content_nodes` would not ignore interpolation nodes. I'm not sure I have an alternative suggestion, though

marcandre: Yeah, I know. I would actually mark `content` as `# @deprecated`; I can't think of a single good use for that method.

marcandre:
Great stuff, thanks!

PR #8807, opened 2020-09-28
Fix a false positive for `Style/RedundantCondition`

koic:
This PR fixes the following incorrect autocorrect for `Style/RedundantCondition` when using assignment by hash key access.

### Before

The result of auto-correction is syntax error.

```console
% cat example.rb
if @columns_cache[table_name]
  @columns_cache[table_name]
else
  @columns_cache[table_name] = super(table_name)
end

% bundle exec rubocop -a --only Style/RedundantCondition
(snip)

Inspecting 1 file
C

Offenses:

example.rb:1:1: C: [Corrected] Style/RedundantCondition: Use double
pipes || instead.
if @columns_cache[table_name] ...
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

% cat example.rb
@columns_cache[table_name] || []=(table_name, super(table_name))

% ruby -c example.rb
example.rb:1: syntax error, unexpected '=', expecting end-of-input
...olumns_cache[table_name] || []=(table_name, super(table_name...
```

### After

The cop accepts the example case.

-----------------

* [x] Added tests.


dvandersluis:
Not sure if your example makes perfect sense here, because it should be `@columns_cache[table_name] ||= super(table_name)` since it was originally a setter, right? I'm not sure if setters are handled by this cop (doesn't appear so from the tests), another one, or not at all.

koic:
My example was incorrect and your feedback is perfectly right. In any case, there was an incompatible auto-correction, and first of all, I made it possible to accept the example case where hash assignment setter is used. I updated this PR. Thank you for your review!

PR #8814, opened 2020-09-29
Add new `Style/CollectionCompact` cop

### LINKED ISSUE #8566: Suggest compact over reject nil values
Rubocop could suggest replacing:

`{}.reject { |_k, v| v.nil? }` with `{}.compact`

fatkodima:
This cop checks for places where custom logic on rejection nils from arrays and hashes can be replaced with `{Array,Hash}#{compact,compact!}`.

```ruby
# bad
array.reject { |e| e.nil? }
array.select { |e| !e.nil? }

# good
array.compact

# bad
hash.reject! { |k, v| v.nil? }
hash.select! { |k, v| !v.nil? }

# good
hash.compact
```

Closes #8566


spec/rubocop/cop/style/array_and_hash_compact_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      hash.compact!
+    RUBY
+  end
+
+  it 'registers an offense and corrects when using `select/select!` to reject nils' do
+    expect_offense(<<~RUBY)
+      array.select { |e| !e.nil? }
+            ^^^^^^^^^^^^^^^^^^^^^^ Use `compact` instead of `select { |e| !e.nil? }`.
+      hash.select! { |k, v| !v.nil? }
+           ^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `compact!` instead of `select! { |k, v| !v.nil? }`.
```
tejasbubane, line 39: There should also be negative cases - selecting nil and rejecting non-nils. Although I cannot image where one would write something like that, ideally the cop should not flag those.

bbatsov:
Let's leave this one for after the 1.0 release next week.

koic:
It may be named `CollectionCompact` cop if https://bugs.ruby-lang.org/issues/17208 is accepted.

fatkodima:
Looks nicer. I would rename it even before that, wdyt?

koic:
Yeah, let's see for other opinions :-)

lib/rubocop/cop/style/array_and_hash_compact.rb
```diff
@@ -0,0 +1,85 @@
...
+      #
+      #   # good
+      #   array.compact
+      #
+      #   # bad
+      #   hash.reject! { |k, v| v.nil? }
+      #   hash.select! { |k, v| !v.nil? }
+      #
+      #   # good
+      #   hash.compact
```
koic, line 22: This may be a typo if it makes symmetry with the bad case.

```suggestion
      #   hash.compact!
```

bbatsov:
I'm fine with the suggested rename and the cop in general.

fatkodima:
Renamed the cop.

bbatsov:
Thanks!

PR #8815, opened 2020-09-29
Fix #8646 for faster find of all files

### LINKED ISSUE #8646: Slow find_files does not use Include configuration
A project with inline Rails engines can become slow with rubocops initial scan.  For example, a project that looks like:

```
.
├── app
├── bin
├── config
├── db
├── engines
│   └── my_engine
│       ├── app
│       ├── bin
│       ├── config
│       ├── db
│       ├── lib
│       ├── log
│       ├── node_modules
│       ├── spec
│       ├── tmp
│       └── vendor
├── lib
├── log
├── node_modules
├── public
├── spec
├── tmp
└── vendor

```

With the following configuration file:
```yaml
AllCops:
  Include:
    - 'app/**/*.rb'
    - 'config/**/*.rb'
    - 'lib/**/*.rb'
    - 'lib/**/*.rake'
    - 'spec/**/*.rb'
    - 'vendor/*/**/*.rb'
    - 'Rakefile*'
    - 'engines/*/app/**/*.rb'
    - 'engines/*/config/**/*.rb'
    - 'engines/*/lib/**/*.rb'
    - 'engines/*/lib/**/*.rake'
    - 'engines/*/spec/**/*.rb'
    - 'engines/*/vendor/*/**/*.rb'
    - 'engines/*/Rakefile*'
  Exclude:
    - 'bin/*'
    - 'app/assets/**/*'
    - 'app/javascript/**/*'
    - 'app/views/**/*'
    - 'tmp/**/*'
    - 'public/**/*'
    - 'node_modules/**/*'
    - 'vendor/**/*'
    - 'db/*'
    - 'engines/*/bin/*'
    - 'engines/*/app/assets/**/*'
    - 'engines/*/app/javascript/**/*'
    - 'engines/*/app/views/**/*'
    - 'engines/*/tmp/**/*'
    - 'engines/*/node_modules/**/*'
    - 'engines/*/db/**/*'
```

--------

## Expected behavior

I would expect that the file scanning would scan the "Include" directories, and "Exclude" the directories after before analyzing.

## Actual behavior

What actually happens is rubocop scans top level root directories, then filters for Exclude patterns
1) gets a list of all directories in the root folder at one level
2) subtracts any one-level directories listed in the Exclude config
3) scans all files
4) filters based on Ruby patterns

The problem is, node_modules folder will often have thousands of files.  The root node_modules can be skipped from the initial scan, but the `engines/my_engine/node_modules` cannot be skipped in the initial filescan (based on the configuration).

Because of this, in the project I'm testing the scanning takes 7 seconds just to get the list of files.  If I scan by only 

See description

## RuboCop version

0.77.0 (using Parser 2.6.5.0, running on ruby 2.6.5 x86_64-darwin18)

tleish:
After sleeping on it, I realized there was a simpler (less code) and slightly faster version of the previous https://github.com/rubocop-hq/rubocop/pull/8806.

By _recursively_ getting all directories and then at each directory level testing Rubocop Exclude on directories.  Once directories are found at all levels, we search for files.

In addition, I replaced many of the `base_dir` string concatenations with the safer `File#join`.

Benchmarks on a large project removes an additional 500+ milliseconds.

The following flamegraphs illustrate the percentage of processing `TargetFinder#target_files_in_dir` took in using the command `bundle exec rubocop -L` to list rubocop files.

#### Original Flamegraph

rubocop-original

#### Changes from https://github.com/rubocop-hq/rubocop/pull/8806

rubocop-update01

#### Changes from this MR

rubocop-update02

Note: I didn't add an entry to the Changelog as a previous one was already added.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/target_finder.rb
+++ b/lib/rubocop/target_finder.rb
@@ -82,46 +82,31 @@ module RuboCop
     # normal way (dir/**/*).
     def find_files(base_dir, flags)
       # get all wanted directories first to improve speed of finding all files
-      patterns = wanted_dir_patterns(base_dir, flags)
+      exclude_pattern = combined_exclude_glob_patterns(base_dir)
+      dir_flags = flags | File::FNM_PATHNAME | File::FNM_EXTGLOB
+      patterns = wanted_dir_patterns(base_dir, exclude_pattern, dir_flags)
+      patterns.map! { |dir| File.join(dir, '*') }
       # We need this special case to avoid creating the pattern
       # /**/* which searches the entire file system.
-      patterns = ["#{base_dir}/**/*"] if patterns.empty?
+      patterns = [File.join(dir, '**/*')] if patterns.empty?
 
       Dir.glob(patterns, flags).select { |path| FileTest.file?(path) }
     end
 
-    def wanted_dir_patterns(base_dir, flags)
-      exclude_pattern = combined_exclude_glob_patterns(base_dir)
-      flags = flags | File::FNM_PATHNAME | File::FNM_EXTGLOB | File::FNM_DOTMATCH
-      wanted_toplevel_dirs = toplevel_dirs(base_dir, flags) -
-                             excluded_dirs(base_dir)
-      wanted_toplevel_dirs.map! { |dir| dir << '/**/' }
-
-      Dir.glob(wanted_toplevel_dirs, flags)
-         .map { |dir| dir << '*' } # add file glob pattern to end of each dir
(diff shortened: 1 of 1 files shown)


bbatsov:
Nicely done! I think it's the first time I see someone sharing flamegraphs in a PR which is extra cool! I always think of @miharekar when I see a flamegraph (a friend of mine, who's a big proponent of the concept).

PR #8828, opened 2020-10-01
Document interaction of `Style/AccessorGrouping` and `Sorbet` [doc]

marcandre:
See https://github.com/rubocop-hq/rubocop/pull/8241#issuecomment-702304766

### DIFF
--- a/lib/rubocop/cop/style/accessor_grouping.rb
+++ b/lib/rubocop/cop/style/accessor_grouping.rb
@@ -7,6 +7,9 @@ module RuboCop
       # By default it enforces accessors to be placed in grouped declarations,
       # but it can be configured to enforce separating them in multiple declarations.
       #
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/style/accessor_grouping.rb
```diff
@@ -7,6 +7,9 @@ module Style
       # By default it enforces accessors to be placed in grouped declarations,
       # but it can be configured to enforce separating them in multiple declarations.
       #
+      # Note: `Sorbet` is not compatible with "grouped" style. Consider "separated" style
```
bbatsov, line 10: I don't think AsciiDoc handles admonition markers properly if they are not UPCASE like NOTE.

marcandre: Fixed, thanks for letting me know.

PR #8833, opened 2020-10-02
Add new `Style/ClassEqualityComparison` cop

fatkodima:
This cop enforces the use of `Object#instance_of?` instead of class comparison for equality.

```ruby
# bad
var.class == Date
var.class.equal?(Date)
var.class.eql?(Date)
var.class.name == "Date"

# good
var.instance_of?(Date)
```

Ran it on 30k files and got 130 offenses. So seems useful.

Not sure if this should be in `Style` or `Lint` (looks like `Style` for me).


bbatsov:
Yeah, I agree it's a style cop. Might be a good idea to mention this in the style guide as well.

docs/modules/ROOT/pages/cops_style.adoc
```diff
@@ -1052,6 +1052,45 @@ var.kind_of?(String)
...
+| Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
+
+| Pending
+| Yes
+| Yes
+| 0.93
+| -
+|===
+
+This cop enforces the use of `Object#instance_of?` instead of class comparison
```
bbatsov, line 1067: Btw, you can use `kind_of?` as well. I think we have some other cop for this, so that one should honor the config of the existing cop.

bbatsov:
You'll also need to rebase.

docs/modules/ROOT/pages/cops_style.adoc
fatkodima: `kind_of?` is an alias for `is_a?`, which checks also superclasses. So this should be `instance_of?` for sure.

fatkodima:
Rebased and added a link to style guide.

docs/modules/ROOT/pages/cops_style.adoc
bbatsov: I keep mixing those up. :-)

bbatsov:
Thanks!

PR #8836, opened 2020-10-02
Fix a false positive for `Lint/BooleanSymbol` when used within `%i[...]`

### LINKED ISSUE #8151: Issue with Lint/BooleanSymbol and %i[...]
```sh
$ echo "%i[foo false]" | rubocop -s test -a --only Lint/BooleanSymbol
Inspecting 1 file
W

Offenses:

test:1:8: W: [Corrected] Lint/BooleanSymbol: Symbol with a boolean name - you probably meant to use false.
%i[foo false]
       ^^^^^

====================
%i[foo false]
```

1) Looks like this isn't crashing by chance. There's no spec for this case

2) Autocorrection should not try to autocorrect this case

fatkodima:
Closes #8151


bbatsov:
Thanks!

PR #8838, opened 2020-10-02
Add new `Lint/RedundantSafeNavigation` cop

### LINKED ISSUE #8668: Add a cop to detect safe navigation uses with respond_to?

I noticed autocorrecting some of our older code that the suggested code by RuboCop is actually totally unnecessary and it seems like a place where codebases could be simplified.

Our original code was:

```ruby
next unless attrs && attrs.respond_to?(:[]) 
```

RuboCop suggested:

```ruby
next unless attrs&.respond_to?(:[]) 
```

There's really no need for the &. safe navigation though because even nil has respond_to?. This can just be autocorrected to:

```ruby
next unless attrs.respond_to?(:[]) 
```

A new cop to remove unnecessary safe navigation

:shrug:

## Additional context

🚫

fatkodima:
This cop checks for redundant safe navigation calls.

```ruby
# bad
attrs&.respond_to?(:[])
foo&.dup&.inspect

# good
attrs.respond_to?(:[])
foo.dup.inspect
```

Closes #8668


bbatsov:
Love it! Great work!

PR #8844, opened 2020-10-03
[Fix #8708] Fix bad regexp recognition in `Lint/OutOfRangeRegexpRef` when there are multiple regexps

### LINKED ISSUE #8708: Lint/OutOfRangeRegexpRef picks wrong regex
## Actual behavior

`Lint/OutOfRangeRegexpRef` seems confused when multiple regular expressions are present on a single line, and picks the _wrong_ Regex, leading to false positives _and_ true negatives.

In the following example `Lint/OutOfRangeRegexpRef` seems to pick the first Regex, in the `gsub` call (`/\s+/`), instead of the final Regex that is populating `$1` (`/foo(bar)/`):

```ruby
if "foo bar".gsub(/\s+/, "") =~ /foo(bar)/
  p $1
end
```
```
x.rb:2:5: W: Lint/OutOfRangeRegexpRef: Do not use out of range reference for the Regexp.
  p $1
    ^^
```

Rubocop won't complain anymore when the Regex in the `gsub` call is modified to capture, but still won't complain when changing the second regex to not capture:

```ruby
if "foo bar".gsub(/(\s+)/, "") =~ /foobar/
  p $1 # always nil
end
```

## RuboCop version

```
0.90.0 (using Parser 2.7.1.4, rubocop-ast 0.4.0, running on ruby 2.6.6 x86_64-linux-musl)
```

dvandersluis:
`Lint/OutOfRangeRegexpRef` was using the wrong regexp to determine how many capturing groups there are, in a few cases, that are now fixed. I also updated the error message to be a bit more straightforward because I found it hard to understand, but please let me know if you'd like me to reword or revert.

Also I have no idea what the plural of `regexp` should be. 😅 

Fixes #8708.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
+++ b/lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
@@ -19,7 +19,7 @@ module RuboCop
       #   puts $1 # => foo
       #
       class OutOfRangeRegexpRef < Base
-        MSG = 'Do not use out of range reference for the Regexp.'
+        MSG = '$%<backref>s is out of range (%<count>s regexp capture %<group>s detected).'
 
         REGEXP_RECEIVER_METHODS = %i[=~ === match].to_set.freeze
         REGEXP_ARGUMENT_METHODS = %i[=~ match grep gsub gsub! sub sub! [] slice slice! index rindex
@@ -56,9 +56,16 @@ module RuboCop
 
         def on_nth_ref(node)
           backref, = *node
-          return if @valid_ref.nil?
+          return if @valid_ref.nil? || backref <= @valid_ref
 
-          add_offense(node) if backref > @valid_ref
+          message = format(
+            MSG,
+            backref: backref,
+            count: @valid_ref.zero? ? 'no' : @valid_ref,
+            group: @valid_ref == 1 ? 'group' : 'groups'
+          )
+
+          add_offense(node, message: message)
         end
 
         private
--- a/lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
+++ b/lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
@@ -35,14 +35,13 @@ module RuboCop
           check_regexp(node.children.first)
         end
 
-        def on_send(node)
+        def after_send(node)
           @valid_ref = nil
 
-          if node.receiver&.regexp_type?
-            check_regexp(node.receiver)
-          elsif node.first_argument&.regexp_type? \
-            && REGEXP_ARGUMENT_METHODS.include?(node.method_name)
+          if regexp_first_argument?(node)
             check_regexp(node.first_argument)
+          elsif regexp_receiver?(node)
+            check_regexp(node.receiver)
           end
         end
 
@@ -79,6 +78,19 @@ module RuboCop
                        else
                          node.each_capture(named: false).count
                        end
+        end
+
+        def regexp_first_argument?(send_node)
+          send_node.first_argument&.regexp_type? \
+            && REGEXP_ARGUMENT_METHODS.include?(send_node.method_name)
+        end
+
+        def regexp_receiver?(send_node)
+          send_node.receiver&.regexp_type?
+        end
+
+        def nth_ref_receiver?(send_node)
+          send_node.receiver&.nth_ref_type?
         end
       end
     end
--- a/spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
+++ b/spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
@@ -5,10 +5,10 @@ RSpec.describe RuboCop::Cop::Lint::OutOfRangeRegexpRef do
 
   let(:config) { RuboCop::Config.new }
 
-  it 'registers an offense when references are used before any Regexp' do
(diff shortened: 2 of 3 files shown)


marcandre:
Update: I started reviewing, love the message changes 👍
It would be quite nice to have a `after_send` callback of some kind, instead of messing around with `ignore` things... I'm checking a few things on the `Commissionner` side.

dvandersluis:
@marcandre thanks! Let me know if you want me to change something.

marcandre:
Could you check if it also has this issue? #8708

dvandersluis:
I'm not sure what you mean? This PR is for that issue haha 😅

dvandersluis:
Oh I assume you meant #8862 / #8863 ? That's a different cop so there shouldn't be a conflict.

dvandersluis:
Updated to fix the changelog now that 0.93 is out.

marcandre (commented):
> Oh I assume you meant #8862 / #8863 ? That's a different cop so there shouldn't be a conflict.

Ooops, yes, my double mistake 🤦‍♂️, need coffee apparently

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -58,7 +58,8 @@ def on_nth_ref(node)
           backref, = *node
           return if @valid_ref.nil?
 
-          add_offense(node) if backref > @valid_ref
+          message = format(MSG, backref: backref, count: @valid_ref == 0 ? 'no' : @valid_ref)
```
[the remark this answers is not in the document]
dvandersluis: @marcandre absolutely! I've made this change now.

marcandre:
So I checked about adding `on_after_xxx` and it's sadly in a hotspot, it would cost a 5% slowdown which is pretty sad 😿
I could add it only just for `send`, wouldn't be so bad, or I could introduce some API to add a callback on the fly, but that feels a bit ugly...
So I'm tempted to just leave this as is (after the conflict is resolved...) even though the logic is quite convoluted just because we don't currently have a way do processing after a particular node's children are processed...

dvandersluis:
@marcandre fixed the conflict, happy to make other changes if you want, just let me know!

marcandre:
Update: I opened #8889; I'm waiting a day or two to see if the latest bug fix release is stable and I'll merge that, which would give you the `on_after_send` that would simplify the processing I believe.

marcandre:
`after_send` is now available...

dvandersluis:
@marcandre cool thanks, I’ll probably Be able to update this tomorrow!

marcandre:
I believe this or can be simplified without using an ignore list...

dvandersluis:
@marcandre updated to use `after_send`, thanks that really simplified it!

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -19,7 +19,7 @@ module Lint
       #   puts $1 # => foo
       #
       class OutOfRangeRegexpRef < Base
-        MSG = 'Do not use out of range reference for the Regexp.'
+        MSG = '$%<backref>s is out of range (%<count>s Regexp capture groups detected).'
```
bbatsov, line 22: I'd use `regexp` here without the capital letter. No idea why it was there in the first place.

spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
```diff
@@ -8,39 +8,39 @@
...
+           ^^ $3 is out of range (2 Regexp capture groups detected).
     RUBY
   end
 
   it 'registers an offense when out of range references are used for mix of numbered and named captures' do
     expect_offense(<<~RUBY)
       /(?<foo>FOO)(BAR)/ =~ "FOOBAR"
       puts $2
-           ^^ Do not use out of range reference for the Regexp.
+           ^^ $2 is out of range (1 Regexp capture groups detected).
```
bbatsov, line 35: It'd be a bit better to say "1 group", but it's not a big deal.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+### Bug fixes
```
bbatsov, line 5: Please, add a blank line after the heading.

spec/rubocop/cop/lint/out_of_range_regexp_ref_spec.rb
dvandersluis: I agree but I'm not sure how to easily achieve this?

dvandersluis: Never mind I just implemented it directly, should be ok.

dvandersluis:
@bbatsov made all the requested changes, thanks!

bbatsov:
Can you also address the changelog conflict?

marcandre:
> Can you also address the changelog conflict?

You can use `rake changelog:fix` and al. in the future 😄

dvandersluis:
@bbatsov @marcandre added the changelog item in `changelog/`

marcandre:
Outstanding work @dvandersluis 👍 
Thanks!

PR #8850, opened 2020-10-04
[Fix #7495] Documentation for `Lint/AmbiguousBlockAssociation` cop is confusing

### LINKED ISSUE #7495: Documentation for Lint/AmbiguousBlockAssociation cop is confusing
## Actual behaviour

Given following file `test.rb`:

```ruby
# frozen_string_literal: true

some_method a { |val| puts val }
```

When I execute command:

```sh
rubocop --force-default-config test.rb
```

I see following:

```
Inspecting 1 file
W

Offenses:

test.rb:3:1: W: Lint/AmbiguousBlockAssociation: Parenthesize the param a { |val| puts val } to make sure that the block will be associated with the a method call.
some_method a { |val| puts val }
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

## What documentation says

> Examples:
> 
> ```
> # bad
> some_method a { |val| puts val }
> 
> # good
> # With parentheses, there's no ambiguity.
> some_method(a) { |val| puts val }
> 
> # good
> # Operator methods require no disambiguation
> foo == bar { |b| b.baz }
> 
> # good
> # Lambda arguments require no disambiguation
> foo = ->(bar) { bar.baz }
> ```

(https://rubocop.readthedocs.io/en/latest/cops_lint/#lintambiguousblockassociation)

## What is the problem

The documentation describes a different case than the cop. Given aforementioned code sample:

```ruby
some_method a { |val| puts val }
```

The cop message says that this line should be fixed in a following way:

```ruby
some_method(a { |val| puts val })
```

Whereas documentation suggests that it should rather be fixed in following way:

```ruby
some_method(a) { |val| puts val }
```

Which is syntactically and semantically different.

Obviously, both suggested resolutions are correct in some cases and incorrect in others. Applying any of them results with `1 file inspected, no offenses detected`. Nevertheless, this discrepancy is somewhat confusing.

## Suggestion 1

IMHO cop documentation should illustrate both "good" resolutions, like in following:

```
# good
# With parentheses, there's no ambiguity.
some_method(a) { |val| puts val }
# or
some_method(a { |val| puts val })
```

I have no suggestion how to improve the cop message. It should be succinct, so perhaps the current one is good enough.

## Suggestion 2

Given the fact that this cop often reports a false positive for a popular RSpec idiom:

```ruby
expect { something }.not_to change { some_variable }
```

Documentation could propose to disable this cop for spec files (see #4222).

## RuboCop version

0.76.0 (using Parser 2.6.5.0, running on ruby 2.6.3 x86_64-darwin17)

AllanSiqueira:
Fixes: #7495 
Add an example in the documentation for `Lint/AmbiguousBlockAssociation` to illustrate both good solutions for this cop.

-----------------

* [x] Added tests.
* [x] Run `bundle exec rake default`. It executes all tests and RuboCop for itself and generates the documentation.

### DIFF
--- a/lib/rubocop/cop/lint/ambiguous_block_association.rb
+++ b/lib/rubocop/cop/lint/ambiguous_block_association.rb
@@ -15,6 +15,8 @@ module RuboCop
       #
       #   # good
       #   # With parentheses, there's no ambiguity.
+      #   some_method(a { |val| puts val })
+      #   # or (different meaning)
       #   some_method(a) { |val| puts val }
       #
       #   # good
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * [#8843](https://github.com/rubocop-hq/rubocop/issues/8843): Fix an incorrect autocorrect for `Lint/AmbiguousRegexpLiteral` when sending method to regexp literal receiver. ([@koic][])
 * [#8842](https://github.com/rubocop-hq/rubocop/issues/8842): Save actual status to cache, except corrected. ([@hatkyinc2][])
 * [#8835](https://github.com/rubocop-hq/rubocop/issues/8835): Fix an incorrect autocorrect for `Style/RedundantInterpolation` when using string interpolation for non-operator methods. ([@koic][])
+* [#7495](https://github.com/rubocop-hq/rubocop/issues/7495): Example for `Lint/AmbiguousBlockAssociation` cop. ([@AllanSiqueira][])
 
 ### Changes
 
@@ -4956,3 +4957,4 @@
 [@rdunlop]: https://github.com/rdunlop
 [@ghiculescu]: https://github.com/ghiculescu
 [@hatkyinc2]: https://github.com/hatkyinc2
+[@AllanSiqueira]: https://github.com/allansiqueira
--- a/docs/modules/ROOT/pages/cops_lint.adoc
+++ b/docs/modules/ROOT/pages/cops_lint.adoc
@@ -27,6 +27,8 @@ some_method a { |val| puts val }
 ----
 # good
 # With parentheses, there's no ambiguity.
+some_method(a { |val| puts val })
+# or (different meaning)
 some_method(a) { |val| puts val }
 
 # good


lib/rubocop/cop/lint/ambiguous_block_association.rb
```diff
@@ -15,6 +15,8 @@ module Lint
       #
       #   # good
       #   # With parentheses, there's no ambiguity.
+      #   some_method(a { |val| puts val  })
+      #   # or
```
marcandre, line 19: How about `# or (different meaning):`

marcandre (changes requested):
Good. I'd merge but there's a Changelog merge conflict, so might as well improve it and rebase 😆

marcandre:
Also the offense message would be best with quotes, if you'd like to do that too...

```sh
Parenthesize the param a { |val| puts val } to make sure that the block will be associated with the a method call.
# to
Parenthesize the param `a { |val| puts val }` to make sure that the block will be associated with the `a` method call.
```

lib/rubocop/cop/lint/ambiguous_block_association.rb
AllanSiqueira: Done it : )

AllanSiqueira:
> Also the offense message would be best with quotes, if you'd like to do that too...
> 
> ```shell
> Parenthesize the param a { |val| puts val } to make sure that the block will be associated with the a method call.
> # to
> Parenthesize the param `a { |val| puts val }` to make sure that the block will be associated with the `a` method call.
> ```

I search for this and seems like it already has quotes, there is somewhere who hasn't the quotes and I let pass?

marcandre:
> > Also the offense message would be best with quotes, if you'd like to do that too...
> > ```shell
> > Parenthesize the param a { |val| puts val } to make sure that the block will be associated with the a method call.
> > # to
> > Parenthesize the param `a { |val| puts val }` to make sure that the block will be associated with the `a` method call.
> > ```
> 
> I search for this and seems like it already has quotes, there is somewhere who hasn't the quotes and I let pass?

Oh, perfect. I got that message from the original issue only, didn't check.

marcandre:
Thanks for the PR! 👍

PR #8851, opened 2020-10-04
[Fix #8848] Fix a false positive for `Style/CombinableLoops` when using the same method with different arguments

### LINKED ISSUE #8848: Style/CombinableLoops incorrectly triggers on rails validates_each
The `Style/CombinableLoops` cop assumes that any method that ends with `_each` is an iterator, and then applies the cop logic to the method body. This rule triggers on the rails `validates_each` method, described [here](https://apidock.com/rails/ActiveModel/Validations/ClassMethods/validates_each), which is not actually an iterator.

The below code will trigger the cop:

```ruby
class Example < ActiveRecord::Base
  # Has two attributes :array_values, :other_array_values, both of which are arrays of strings

  validates_each :array_values, :other_array_values do |record, attribute, array_value|
    array_value.each do |array_value_item|
      record.errors.add(attribute, "invalid") unless is_valid_item?(array_value_item)
    end
  end

  def is_valid_item?(item)
    return item !== "invalid"
  end
end
```

--------

## Expected behavior

The `Style/CombinableLoops` should not necessarily trigger on methods with names that happen to end with `_each`, since that suffix doesn't necessarily imply that it's an iterator. 

## Actual behavior

The above code is not actually a combinable loop, because the `validates_each` is not really an iterator, but simply a method that can apply some validation logic to a value which may in turn be an enumerable. As such it should not trigger this cop.

See example above in the issue description above. There is no way to combine `validates_each` and the iteration over the `array_value` array.

## RuboCop version

```
└> bundle exec rubocop -V
0.92.0 (using Parser 2.7.1.5, rubocop-ast 0.7.1, running on ruby 2.7.1 x86_64-linux)
```

dvandersluis:
If the same iterable method is called multiple times with different arguments it is not a combinable loop, and thus a false positive for `Style/CombinableLoops`.

Fixes #8848.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [#8354](https://github.com/rubocop-hq/rubocop/issues/8354): Detect regexp named captures in `Style/CaseLikeIf` cop. ([@dsavochkin][])
 * [#8830](https://github.com/rubocop-hq/rubocop/issues/8830): Fix bad autocorrect of `Style/StringConcatenation` when string includes double quotes. ([@tleish][])
 * [#8807](https://github.com/rubocop-hq/rubocop/pull/8807): Fix a false positive for `Style/RedundantCondition` when using assignment by hash key access. ([@koic][])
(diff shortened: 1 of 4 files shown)


koic:
Thank you!

PR #8852, opened 2020-10-04
Added `IgnoredMethods` configuration to `Style/CombinableLoops`

dvandersluis:
Related to #8848, it probably makes sense to make `Style/CombinableLoops` use `IgnoredMethods`, since it's possible to have a method that matches the method name conditions without being combinable.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 * [#8782](https://github.com/rubocop-hq/rubocop/issues/8782): Fix incorrect autocorrection for `Style/TernaryParentheses` with `defined?`. ([@dvandersluis][])
 * [#8864](https://github.com/rubocop-hq/rubocop/issues/8864): Fix false positive for `Style/RedundantBegin` with a postfix `while` or `until`. ([@dvandersluis][])
 
+### Changes
+* [#8852](https://github.com/rubocop-hq/rubocop/pull/8852): Add `IgnoredMethods` configuration to `Style/CombinableLoops`. ([@dvandersluis][])
+
 ## 0.93.0 (2020-10-08)
 
 ### New features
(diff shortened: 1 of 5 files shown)


marcandre:
Now that #8851 is merged, your example no longer works (methods have different arguments)...

dvandersluis:
@marcandre oops, thanks! I had changed the test but not the example, I’ll fix it.

dvandersluis:
@marcandre updated!

marcandre:
Cool. Now why should these two `validates_each` not be combined?

dvandersluis:
@marcandre definitely true that you could. I was trying to come up with an example of something you might not want to combine despite being able to, this is maybe not the best example though. 

Overall for this I was just thinking based on the original issue that there is a use case to make this cop less restrictive without having to be manually disabled (because the cop just looks for a method that starts or ends with `each` but it’s conceivable that there are methods that match that aren’t enumerators).

marcandre:
The patch is well written and not particularly complex, but I'd wait until
we have someone providing an actual example of false positive.
For example, had it been merged originally, we might not have had the
report about different arguments...

dvandersluis:
To be fair, we didn't have the report about different arguments, I discovered it while trying to replicate the [original issue](https://github.com/rubocop-hq/rubocop/issues/8848) 😁 I originally had gone down this route because the example provided wasn't actually a failure, until it struck me that there might be another issue with multiple `validates_each` calls.

I'm ok with closing this though if you don't think it's necessary!

bbatsov:
I'm leaning towards closing the PR at this point, as I'm not sure it solves an actual problem. We can revisit it down the road if the need to do so arises.

PR #8853, opened 2020-10-04
[Fix #8842]: Add notification about cache being used to debug mode

### LINKED ISSUE #8842: When loading a cache, rubocop should let the user know

As a user that tried to debug an issue https://github.com/rubocop-hq/rubocop/issues/8804, and I'm guessing https://github.com/rubocop-hq/rubocop/issues/8755
I was surprised to learn that rubocop has a cache.
The behaviour and code path using the cache is different in some way, even if it's just due to code bugs
It might only be a debugging concern and not a normal golden path, but so is the cache path itself optimising..
It might improve issue reporting, handling and testing as it would be obvious in the output.
I think it's common to announce when cached is used in other utilities, but I can't at the moment name particular examples. 

Output a line at least in the default formater when cache is successfully loaded, indicating something like 
```
Cache loaded from ~/.cache/rubocop
```
Probably as the first line before the normal output.

I don't think formaters like json actually need it, but it can become an attribute in the summery if that makes more sense to be good cross formaters...

An alternative would probably to only add it to the `--debug` option, but than I'd recommend that would come with updating the issue template to run and paste an output with that.

Or just have running with no cache `-C false` in the issue submission as the MVP..

## Additional context

This is an example of a bug I think I found with the caching, where it doesn't save the issue status as it was discovered, it saves a hard-coded value
https://github.com/rubocop-hq/rubocop/pull/8841/files

hatkyinc2:
* When rubocop is ran in debug mode (`--debug`) and rubocop uses a cache, output the location of cache used. (This is to inform about possible differences between runs..)
* Encourage users submitting bugs to use debug mode.

-----------------

* [ ] Added tests.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -23,6 +23,7 @@ Describe here how you expected RuboCop to behave in this particular situation.
 ## Actual behavior
 
 Describe here what actually happened.
+Please use `rubocop --debug` when pasting rubocop output as it contains additional information.
 
 ## Steps to reproduce the problem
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #8874, opened 2020-10-09
[Fix #8871] Fix a false positive for `Style/RedundantBegin`

### LINKED ISSUE #8871: Style/RedundantBegin autocorrect breaks rhs begin-end
Auto-correcting `Style/RedundantBegin` with RuboCop v0.93 breaks code if the value of subject begin-end block is used as a method call parameter or as rhs of a binop.

No offense if value of subject begin-end block is rhs of an assignment, e.g. `@var ||= begin ...`

```ruby
# frozen_string_literal: true

puts begin
  'foo'
  'bar'
end

def f
  'hoge' || begin
    'fuga'
    'piyo'
  end
end

puts f
```

--------

## Expected behavior

These code should not offense nor auto-correct.

## Actual behavior

Auto-correction breaks code.

```
$ ruby main.rb 
bar
hoge
$ rubocop --cache false --auto-correct main.rb --only Style/RedundantBegin
Inspecting 1 file
C

Offenses:

main.rb:3:6: C: [Corrected] Style/RedundantBegin: Redundant begin block detected.
puts begin
     ^^^^^
main.rb:9:13: C: [Corrected] Style/RedundantBegin: Redundant begin block detected.
  'hoge' || begin
            ^^^^^

$ ruby main.rb 

piyo
$ 
```

## RuboCop version

```
$ rubocop -V
```

koic:
Fixes #8871.

This PR fixes a false positive for `Style/RedundantBegin` when using `begin` for method argument or part of conditions.

`begin` keyword may be redundant when using only one expression in `begin` of each issue case.
However, since it is the unintended case for #8822, I think it can be implemented as an enhancement different from this bug fix.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [#8869](https://github.com/rubocop-hq/rubocop/issues/8869): Fix a false positive for `Style/RedundantBegin` when using `begin` for or assignment and method call. ([@koic][])
 * [#8862](https://github.com/rubocop-hq/rubocop/issues/8862): Fix an error for `Lint/AmbiguousRegexpLiteral` when using regexp without method calls in nested structure. ([@koic][])
 * [#8872](https://github.com/rubocop-hq/rubocop/issues/8872): Fix an error for `Metrics/ClassLength` when multiple assignments to constants. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #8877, opened 2020-10-09
[Fix #8868] Fix false positives for `Lint/RedundantSafeNavigation`

### LINKED ISSUE #8868: Lint/RedundantSafeNavigation should ignore self-yielding methods
I've just updated to RuboCop 0.93.0 and received some suspicious false positives from Lint/RedundantSafeNavigation.

Safe navigation usages are intended to "check `nil`, and we expect an early return with `nil` result when the receiver was `nil`, so we can ease `nil` handling."
Self-yielding methods, `#tap` , `#yield_self`, and `#then`, in the other hand, yields the receiver into a block when called, possibly making method calls with the possibly-`nil` receiver, or pass `nil` to unsafe methods like `Time.at`.
So safe navigating into self-yielding methods should imply that an early return is intended, therefore I think the safe navigations are actually not redundant.

While I can assume that auto-correction may break code with unsafe cops, currently the possible actions I can take are either rewrite safe navigation to something like `array_or_nil.tap { |array| puts array[0] unless array.nil? }`, which is even more redundant, or disable this cop.

--------

## Expected behavior

`#tap` , `#yield_self`, and `#then` should not be flagged.

## Actual behavior

These calls are flagged.
Auto-correction produces code that passes `nil` to a block which may raise errors if `nil` is passed.

While documentation says `it can produce code that returns non nil`, raising is more than expected.

```
begin
  nil&.tap { |i| puts "p #{Time.at(i)}" }
rescue
  puts 'parse error'
end

time = begin
  nil&.yield_self { |i| Time.at(i) }
rescue
  'parse error'
end

puts "time = #{time.inspect}"
```

```
$ ruby main.rb
time = nil
$ rubocop --auto-correct-all --only Lint/RedundantSafeNavigation main.rb
Inspecting 1 file
W

Offenses:

main.rb:2:6: W: [Corrected] Lint/RedundantSafeNavigation: Redundant safe navigation detected.
  nil&.tap { |i| puts "p #{Time.at(i)}" }
     ^^^^^
main.rb:8:6: W: [Corrected] Lint/RedundantSafeNavigation: Redundant safe navigation detected.
  nil&.yield_self { |i| Time.at(i) }
     ^^^^^^^^^^^^

$ ruby main.rb
parse error
time = "parse error"
$
```

## RuboCop version

```
$ [bundle exec] rubocop -V
```

tejasbubane:
With `tap`, `yield_self` and `then` methods.

Closes #8868

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [#8872](https://github.com/rubocop-hq/rubocop/issues/8872): Fix an error for `Metrics/ClassLength` when multiple assignments to constants. ([@koic][])
 * [#8871](https://github.com/rubocop-hq/rubocop/issues/8871): Fix a false positive for `Style/RedundantBegin` when using `begin` for method argument or part of conditions. ([@koic][])
 * [#8875](https://github.com/rubocop-hq/rubocop/issues/8875): Fix an incorrect auto-correct for `Style/ClassEqualityComparison` when comparing class name. ([@koic][])
+* [#8868](https://github.com/rubocop-hq/rubocop/issues/8868): Fix false positives for `Lint/RedundantSafeNavigation` with `tap`, `yield_self` and `then` methods. ([@tejasbubane][])
 
 ## 0.93.0 (2020-10-08)
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -1767,6 +1767,9 @@ Lint/RedundantSafeNavigation:
     - public_send
     - send
     - __send__
+    - tap
+    - yield_self
+    - then
 
 Lint/RedundantSplatExpansion:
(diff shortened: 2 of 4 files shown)


dvandersluis:
As a general question, is it documented somewhere if users need to copy all the default `IgnoredMethods` if they want to add to it? I don't believe that if `IgnoredMethods` is set in config for a cop that supports it that it is merged with the default list.

bbatsov:
@dvandersluis See https://docs.rubocop.org/rubocop/0.93/configuration.html#inheritance

bbatsov:
Fixed by https://github.com/rubocop-hq/rubocop/pull/8884 @tejasbubane Thanks for proposing another solution!

PR #8881, opened 2020-10-10
Don't over use `extend`

marcandre:
The cop `Style/MethodCallWithArgsParentheses` does an `extend` at runtime in `initialize`.

This is not really needed in this case and this PR avoid that, while reducing the overall complexity and line count.

Other small changes were done in #7623 because of this; I didn't check if these should be reverted or not.

Note: To my surprise, I couldn't measure a difference performance-wise 😮


Drenmi:
Wow. This is pretty wild! The cop was essentially some sort of factory pattern that used inheritance. 😅 

I still think it would be nicer to use composition here, but I think this improvement is big enough to stand on its own.

bbatsov (approved):

marcandre:
Merged

PR #8113, opened 2020-06-07
Let format_offense add variable-width whitespace

eugeneius:
Followup to https://github.com/rubocop-hq/rubocop/pull/8069.

This allows `expect_offense` to be used when the dynamically generated part of the source code is not highlighted by the offense.

I couldn't find a spec in this repo that could be rewritten to use this option, but I have an open PR on RuboCop Rails that would benefit from it: https://github.com/rubocop-hq/rubocop-rails/pull/257/files#diff-d9cf3a5c58c7f3acbad8f9812ca8006bR70

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [ ] Added tests.
* [x] Added an entry to the [Changelog](https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format).
* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: https://chris.beams.io/posts/git-commit/

PR #8884, opened 2020-10-10
Rework `Lint/RedundantSafeNavigation` to be more safe

### LINKED ISSUE #8867: Lint/RedundantSafeNavigation doesn't account for ActiveSupport core-ext
I just updated to RuboCop 0.93.0. Lint/RedundantSafeNavigation is suggesting an incorrect action.

```
mileage&.to_s(:delimited)
       ^^^^^^^^^^^^^^^^^^
```

RuboCop seems to assume that the `&` is unnecessary, because `nil` responds to `to_s`. The problem is, `to_s(:delimited)` is a core-ext on `Numeric` that ActiveSupport adds.
https://github.com/rails/rails/blob/7905bdfd8b2ae50319cd7a9a74ee1f8c865d648d/activesupport/lib/active_support/core_ext/numeric/conversions.rb#L122-L123

```ruby
>> 123456.to_s(:delimited)   #=> "123,456"
>> nil.to_s(:delimited)
        2: from (irb):2
        1: from (irb):2:in `to_s'
ArgumentError (wrong number of arguments (given 1, expected 0))
```

--------

## Expected behavior

Don't flag these this line.

## Actual behavior

It flagged this line.

This code shouldn't get flagged.

```
val = 123_456
val&.to_s(:delimited)

val = nil
val&.to_s(:delimited)
```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.93.0 (using Parser 2.7.2.0, rubocop-ast 0.7.1, running on ruby 2.5.1 x86_64-linux)
```

### LINKED ISSUE #8868: Lint/RedundantSafeNavigation should ignore self-yielding methods

fatkodima:
Closes #8867 
Closes #8868 

I reworked this to make an offense only when used inside conditions and only for whitelisted methods.

cc @bbatsov @marcandre @natematykiewicz

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1757,16 +1757,14 @@ Lint/RedundantSafeNavigation:
   Description: 'Checks for redundant safe navigation calls.'
   Enabled: pending
   VersionAdded: '0.93'
+  AllowedMethods:
+    - instance_of?
+    - kind_of?
+    - is_a?
+    - eql?
+    - respond_to?
+    - equal?
   Safe: false
-  IgnoredMethods:
-    - to_c
-    - to_f
-    - to_i
-    - to_r
-    - rationalize
-    - public_send
-    - send
-    - __send__
 
 Lint/RedundantSplatExpansion:
   Description: 'Checks for splat unnecessarily being called on literals.'
--- a/docs/modules/ROOT/pages/cops_lint.adoc
+++ b/docs/modules/ROOT/pages/cops_lint.adoc
@@ -2918,25 +2918,38 @@ require 'unloaded_feature'
 |===
 
 This cop checks for redundant safe navigation calls.
-
-In the example below, the safe navigation operator (`&.`) is unnecessary
-because `NilClass` has methods like `respond_to?` and `dup`.
+`instance_of?`, `kind_of?`, `is_a?`, `eql?`, `respond_to?`, and `equal?` methods
+are checked by default. These are customizable with `AllowedMethods` option.
 
 This cop is marked as unsafe, because auto-correction can change the
 return type of the expression. An offending expression that previously
 could return `nil` will be auto-corrected to never return `nil`.
 
+In the example below, the safe navigation operator (`&.`) is unnecessary
+because `NilClass` has methods like `respond_to?` and `is_a?`.
+
 === Examples
 
 [source,ruby]
 ----
-# bad, because nil has these methods
-attrs&.respond_to?(:[])
-foo&.dup&.inspect
+# bad
+do_something if attrs&.respond_to?(:[])
 
 # good
-attrs.respond_to?(:[])
-foo.dup.inspect
+do_something if attrs.respond_to?(:[])
+
+# bad
+while node&.is_a?(BeginNode)
+  node = node.parent
+end
+
+# good
+while node.is_a?(BeginNode)
+  node = node.parent
+end
+
+# good - without `&.` this will always return `true`
+foo&.respond_to?(:to_a)
 ----
 
 === Configurable attributes
@@ -2944,8 +2957,8 @@ foo.dup.inspect
 |===
 | Name | Default value | Configurable values
 
-| IgnoredMethods
-| `to_c`, `to_f`, `to_i`, `to_r`, `rationalize`, `public_send`, `send`, `__send__`
+| AllowedMethods
+| `instance_of?`, `kind_of?`, `is_a?`, `eql?`, `respond_to?`, `equal?`
 | Array
 |===
 
--- a/lib/rubocop/cop/lint/redundant_safe_navigation.rb
+++ b/lib/rubocop/cop/lint/redundant_safe_navigation.rb
@@ -4,34 +4,52 @@ module RuboCop
   module Cop
     module Lint
       # This cop checks for redundant safe navigation calls.
-      #
-      # In the example below, the safe navigation operator (`&.`) is unnecessary
-      # because `NilClass` has methods like `respond_to?` and `dup`.
+      # `instance_of?`, `kind_of?`, `is_a?`, `eql?`, `respond_to?`, and `equal?` methods
+      # are checked by default. These are customizable with `AllowedMethods` option.
       #
       # This cop is marked as unsafe, because auto-correction can change the
       # return type of the expression. An offending expression that previously
       # could return `nil` will be auto-corrected to never return `nil`.
       #
+      # In the example below, the safe navigation operator (`&.`) is unnecessary
+      # because `NilClass` has methods like `respond_to?` and `is_a?`.
+      #
       # @example
-      #   # bad, because nil has these methods
-      #   attrs&.respond_to?(:[])
-      #   foo&.dup&.inspect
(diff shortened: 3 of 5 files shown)


marcandre (changes requested):
Good work.

I would remove `nil?` from the allowlist, since `foo&.nil?` is trivially falsey but `foo.nil?` isn't.

Unlikely, but `nil.methods - Object.new.methods #=> [:to_h, :&, :to_c, :to_f, :to_i, :to_a, :to_r, :rationalize, :|, :^]`, so `foo&.respond_to?(any_of_these)` shouldn't be flagged. Because of this, probably best to mark the autocorrect as unsafe (`foo&.respond_to?(variable)` can't be autocorrected safely).

fatkodima:
Good points 👍 Removed `nil?` and marked as unsafe.

marcandre:
Great. Would it be difficult to exclude `foo&.respond_to?(:to_a)` and others, when the argument is a literal symbol corresponding to one of the few methods that `NilClass` defines that may not be defined on other objects?

fatkodima:
Updated with suggestion.
This surely was easy, I just was lazy 😄

spec/rubocop/cop/lint/redundant_safe_navigation_spec.rb
```diff
@@ -2,53 +2,97 @@
...
   end
 
-  it 'does not register an offense when using non-nil method with `.&`' do
-    expect_no_offenses(<<~RUBY)
-      foo.&do_something
+  %i[while until].each do |loop_type|
+    it 'registers an offense and corrects when `&.` is used inside `#{loop_type}` condition' do
+      expect_offense(<<~RUBY)
+        #{loop_type} foo&.respond_to?(:bar)
+                 ^^^^^^^^^^^^^^^^^^^ Redundant safe navigation detected.
```
marcandre, line 43: Please use `_{loop_type}` (see #8113)

spec/rubocop/cop/lint/redundant_safe_navigation_spec.rb
```diff
@@ -2,53 +2,97 @@
...
+          do_something
+        end
+
+        begin
+          do_something
+        end #{loop_type} foo&.respond_to?(:bar)
+                     ^^^^^^^^^^^^^^^^^^^ Redundant safe navigation detected.
```
marcandre, line 50: Same here

config/default.yml
```diff
@@ -1757,16 +1757,13 @@ Lint/RedundantSafeNavigation:
   Description: 'Checks for redundant safe navigation calls.'
   Enabled: pending
   VersionAdded: '0.93'
-  Safe: false
```
marcandre, line 1760: I would keep it as `Safe: false` given that there are some cases can't handle perfectly.

docs/modules/ROOT/pages/cops_lint.adoc
```diff
@@ -2911,41 +2911,43 @@ require 'unloaded_feature'
...
 
 This cop checks for redundant safe navigation calls.
 
 In the example below, the safe navigation operator (`&.`) is unnecessary
-because `NilClass` has methods like `respond_to?` and `dup`.
-
-This cop is marked as unsafe, because auto-correction can change the
-return type of the expression. An offending expression that previously
-could return `nil` will be auto-corrected to never return `nil`.
+because `NilClass` has methods like `respond_to?` and `is_a?`.
```
marcandre, line 2923: Unsafe examples include: `var = :to_a; foo&.respond_to?(var)` and expressions with `||`/`&&` returning `false` instead of `nil`.

marcandre (changes requested):
Almost ready

fatkodima:
@marcandre Thank you for all reviews 💪 
Updated with your suggestions.

config/default.yml
```diff
@@ -1757,16 +1757,14 @@ Lint/RedundantSafeNavigation:
   Description: 'Checks for redundant safe navigation calls.'
   Enabled: pending
   VersionAdded: '0.93'
+  AllowedMethods:
```
bbatsov, line 1760: You should document this setting in the cop description.

lib/rubocop/cop/lint/redundant_safe_navigation.rb
```diff
@@ -5,33 +5,47 @@ module Cop
...
       # This cop checks for redundant safe navigation calls.
       #
-      # In the example below, the safe navigation operator (`&.`) is unnecessary
-      # because `NilClass` has methods like `respond_to?` and `dup`.
-      #
       # This cop is marked as unsafe, because auto-correction can change the
       # return type of the expression. An offending expression that previously
       # could return `nil` will be auto-corrected to never return `nil`.
       #
+      # Unsafe examples include: `var = :to_a; foo&.respond_to?(var)`
```
bbatsov, line 12: Isn't it better to drop a code block with unsafe examples here?

fatkodima:
Updated with suggestions. Let me know if there is anything else I can improve.

bbatsov:
Thanks!

PR #8885, opened 2020-10-10
[Fix #8761] Read `required_ruby_version` from gemspec file if it exists 

### LINKED ISSUE #8761: Read TargetRubyVersion from gemspec `required_ruby_version` if not specified

Gemspec/RequiredRubyVersion enforces that gemspec.required_ruby_version is set and then complains that it differs from TargetRubyVersion in .rubocop.yml even if it is not specified in .rubocop.yml.

Having to specify TargetRubyVersion in .rubocop.yml in addition to gemspec.required_ruby_version duplicates the information.

If TargetRubyVersion is not specified in .rubocop.yml it should be read from gemspec.required_ruby_version if present.

HeroProtagonist:
Closes #8761

- Added `GemspecFile` class to search upwards for `<project_base>.gemspec` and read from `required_ruby_version` if present in the file
- Placed it in the `Sources` array right before `Default`
- Created specs over new class
- Update docs to clarify which files are looked at for ruby version
- Updated changelog

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#8761](https://github.com/rubocop-hq/rubocop/issues/8761): Read `required_ruby_version` from gemspec file if it exists #8761. ([@HeroProtagonist][])
+
 ### Bug fixes
 
 * [#8499](https://github.com/rubocop-hq/rubocop/issues/8499): Fix `Style/IfUnlessModifier` and `Style/WhileUntilModifier` to prevent an offense if there are both first-line comment and code after `end` block. ([@dsavochkin][])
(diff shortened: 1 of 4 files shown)


bbatsov:
I'm fine with the proposed change, but we should document the new behavior as otherwise it might confuse people. 

> I guess the question to be answered is how they support those cases, where applicable, and then matching that behavior

As for bounded version - I guess you can check for an array and extract a minimum version from there. If only a maximum version is present there's nothing for us to do.

bbatsov:
You'll also need to rebase on top of `master`.

HeroProtagonist:
@bbatsov updated and rebased 👍

bbatsov:
Thanks!

PR #8889, opened 2020-10-13
Add `after_<type>` callbacks

marcandre:
I double-checked that this has basically no performance impact (less than ~0.1% overall) after #8785 (otherwise that would add ~4-5% overall)


bbatsov:
Won't `after_something` be a simpler name? `on_after` seems a bit confusing to me. :)

marcandre:
> Won't `after_something` be a simpler name? `on_after` seems a bit confusing to me. :)

I liked that it made it clear what where the callbacks: all methods starting with `on_`. I can check how more complicated the code is to handle `after_`.

bbatsov:
Yeah, I got why you named it this way, but the only reason why all callbacks started with `on_` is that they are all the same type of callbacks. I also believe it's quite uncommon to combine two prepositions in a name. I'm not that concerned about the naming per se, rather about people understanding what the new callbacks are about.

marcandre:
Changed `on_after` => `after`. It is definitely better 👍

May I merge this now, or would you rather wait for 1.0 to be out?

bbatsov:
@marcandre Merge it! 🚀

bbatsov (approved):

PR #8890, opened 2020-10-13
Mark `Style/SlicingWithRange` as safe

koic:
Follow https://github.com/rubocop-hq/rubocop/pull/7921.

This PR marks `Style/SlicingWithRange` as safe.

-----------------

* [ ] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -4116,8 +4116,7 @@ Style/SingleLineMethods:
 Style/SlicingWithRange:
   Description: 'Checks array slicing is done with endless ranges when suitable.'
   Enabled: pending
-  VersionAdded: '0.83'
-  Safe: false
+  VersionAdded: '0.94'
 
 Style/SoleNestedConditional:
   Description: >-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * [#8892](https://github.com/rubocop-hq/rubocop/issues/8892): Fix an error for `Style/StringConcatenation` when correcting nested concatenable parts. ([@fatkodima][])
 
+### Changes
+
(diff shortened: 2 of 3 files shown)


bbatsov:
Let's also add a note in the cop description explaning why the auto-correct is unsafe, so it's clearer to the end users.

koic:
Ah, I misunderstood it as `(begin_pos..-1)` 💦  I think this cop is safe now, and I updated it that way.

@zverok Please let me know if there is any reason why this cop is unsafe.

config/default.yml
```diff
@@ -4116,8 +4116,7 @@ Style/SingleLineMethods:
 Style/SlicingWithRange:
   Description: 'Checks array slicing is done with endless ranges when suitable.'
   Enabled: pending
-  VersionAdded: '0.83'
-  Safe: false
+  VersionAdded: '0.94'
```
bbatsov, line 4119: This should be `VersionChanged`.

zverok:
I don't believe this cop can be safe. Equivalence of `x..-1` and `x..` is guaranteed only for `Array#[]`, but as any class might have `#[]` method (with semantics we do not know from code analysis), generally those two are NOT equivalent. Examples with Ruby core classes:
```ruby
h = {-3..1 => 'foo', 0..5 => 'bar'}
h[-3..1] # => "foo" 
h[-3..] # => nil 

sum = proc { |ary| ary.sum }
sum[-3..-1] # => -6 
sum[-3..] # Hangs forever
```

While both examples are artificial, they demonstrate the principle: in general, `#[]` is not expected to treat endless range as range till `-1`

koic:
Ah, I was missing on that point of view. I think it's possible to add an unsafe reason to the documentation. I'll close this PR for now. @zverok Thank you for your explanation!

PR #8895, opened 2020-10-15
Add new `Lint/EmptyBlock` cop

fatkodima:
This cop checks for blocks without a body.
```ruby
# bad
items.each { |item| }

# good
items.each { |item| puts item }
```

Blocks without a body most of the time indicate useless code (left after refactoring, for example) or that some explanation about its emptiness should be given (using a comment).


bbatsov:
I'm guessing an empty one-line block with a comment after it would fine as well. E.g.

```
x = ->{} # no-op
```

lib/rubocop/cli/command/auto_genenerate_config.rb
```diff
@@ -90,7 +90,7 @@ def run_all_cops(line_length_contents)
         def reset_config_and_auto_gen_file
           @config_store = ConfigStore.new
           @config_store.options_config = @options[:config] if @options[:config]
-          File.open(AUTO_GENERATED_FILE, 'w') {}
+          File.open(AUTO_GENERATED_FILE, 'w') { |f| f.truncate(0) }
```
marcandre, line 93: `truncate` is redundant. Maybe just `{} # truncate`

fatkodima:
> I'm guessing an empty one-line block with a comment after it would fine as well. E.g.
x = ->{} # no-op

This will conflict with [Style/InlineComment](https://docs.rubocop.org/rubocop/0.86/cops_style.html#styleinlinecomment)

How about preceding comment then?
```ruby
# no-op
x = ->{}
```

marcandre:
> This will conflict with [Style/InlineComment](https://docs.rubocop.org/rubocop/0.86/cops_style.html#styleinlinecomment)

That cop is (thankfully) disabled by default. In any case, inline comment should be supported.

fatkodima:
Added support for inline comments.

config/default.yml
```diff
@@ -1474,6 +1474,12 @@ Lint/ElseLayout:
   Enabled: true
   VersionAdded: '0.17'
 
+Lint/EmptyBlock:
+  Description: 'This cop checks for blocks without a body.'
+  Enabled: pending
+  VersionAdded: '0.94'
```
bbatsov, line 1480: The next release will be 1.0 and all currently pending cops will be enabled there by default. I'll ponder if we should merge this one after 1.0.

lib/rubocop/cop/lint/empty_block.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # This cop checks for blocks without a body.
```
bbatsov, line 6: I'd add here also some explanation that such empty blocks are typically an oversight, so it's clearer what we're aiming for.

lib/rubocop/cop/lint/empty_block.rb
```diff
@@ -0,0 +1,40 @@
...
+      # @example
+      #   # bad
+      #   items.each { |item| }
+      #
+      #   # good
+      #   items.each { |item| puts item }
+      #
+      # @example AllowComments: true (default)
+      #   # good
+      #   items.each do |item|
```
bbatsov, line 17: We should have some examples with inline comments as well.

bbatsov:
Great work!

PR #8898, opened 2020-10-16
Update RuboCop/RSpec

marcandre:
Disabled one of the new cop (I already voiced my opinion on that cop in #8491 and others, we have >300 offenses in our codebase)

ANother cop is fixed in master but yet unreleased.

Otherwise nothing much here.


koic (approved):

marcandre:
Thanks for the quick review @koic

PR #8899, opened 2020-10-16
Update RuboCop/Performance

marcandre:
Builds on #8898 and updates `RuboCop/Performance`

New cop detected arrays that needed to be put into constants, and some places where `sum` can be used. Cop excluded from specs


koic (approved):
`private_constant` to constants under lib/rubocop/cop directory may be strict, anyway overall it looks good to me. Thank you!

marcandre:
Merged

PR #8902, opened 2020-10-16
[Fix #8901] Fix a false positive for `Naming/BinaryOperatorParameterName`

### LINKED ISSUE #8901: Discrepancy with Ruby Style Guide regarding regexp and BinaryOperatorParameterName cop
I implemented a class which defined `=~`. I was basing it off a class internal to Rails, and so I also utilized their parameter name of `re`. In addition to being too short, to my surprise, Rubocop informed me that I violated `Naming/BinaryOperatorParameterName` and should name my parameter `other`. I looked at the [Rubocop docs](https://docs.rubocop.org/rubocop/cops_naming.html#namingbinaryoperatorparametername), which linked me to the ruby style guide for the [`other` parameter](https://rubystyle.guide/#other-arg). It is my understanding that Rubocop is based on this guide.

The style guide states:
> When defining binary operators and operator-alike methods, name the parameter other for operators with "symmetrical" semantics of operands. Symmetrical semantics means both sides of the operator are typically of same or coercible types.
>
> Operators and operator-alike methods with symmetrical semantics (the parameter should be named `other`): `+`, `-`, `*`, `/`, `%`, `**`, `==`, `>`, `<`, `|`, `&`, `^`, `eql?`, `equal?`.

the style guide then states
> Operators with non-symmetrical semantics (the parameter should not be named `other`): `<<`, `[]` (collection/item relations between operands), `===` (pattern/matchable relations). 

I would assume that `=~` falls under pattern/matchable relations, but I might be wrong. I would expect in this case, that I would not be in violation of the `Naming/BinaryOperatorParameterName` cop, especially considering `=~` is not listed in the operators and operator-alike methods.

For example's sake, here is the aforementioned method that was in violation:

```ruby
def =~(re)
  re.match?(to_s)
end
```
--------

## Expected behavior
I expected to not be in violation of `Naming/BinaryOperatorParameterName` for this method.

## Actual behavior

```ruby
For /Users/zach/workspace/checklistpro: configuration from /Users/zach/workspace/checklistpro/.rubocop.yml
configuration from /Users/zach/.gem/ruby/2.7.2/gems/rubocop-performance-1.8.1/config/default.yml
configuration from /Users/zach/.gem/ruby/2.7.2/gems/rubocop-performance-1.8.1/config/default.yml
Default configuration from /Users/zach/.gem/ruby/2.7.2/gems/rubocop-0.93.1/config/default.yml
configuration from /Users/zach/.gem/ruby/2.7.2/gems/rubocop-rails-2.8.1/config/default.yml
configuration from /Users/zach/.gem/ruby/2.7.2/gems/rubocop-rails-2.8.1/config/default.yml
Inheriting configuration from /Users/zach/workspace/checklistpro/.rubocop_todo.yml
Inspecting 1 file
Scanning /Users/zach/workspace/checklistpro/app/models/jetpack_time_zone.rb
C

Offenses:

app/models/jetpack_time_zone.rb:28:10: C: Naming/BinaryOperatorParameterName: When defining the =~ operator, name its argument other.
  def =~(re)
         ^^
app/models/jetpack_time_zone.rb:28:10: C: Naming/MethodParameterName: Method parameter must be at least 3 characters long.
  def =~(re)
         ^^

Finished in 0.5917779998853803 seconds
```

Run the following code with `ruby hello.rb` to verify it works. Then run `rubocop hello.rb` and verify that the `def =~` method is in violation of `Naming/BinaryOperatorParameterName`.

image

```ruby
# hello.rb
# frozen_string_literal: true

# Example of Naming/BinaryOperatorParameterName cop
class Hello
  def initialize(name)
    @name = name
  end

  def =~(regex)
    regex.match?('Zach')
  end

  def self.greet(name)
    h = new(name)

    puts 'Hello Zach' if h =~ /#{name}/
  end
end

Hello.greet('Fred')
Hello.greet('Zach')
```
## RuboCop version
```
~/workspace/rubocop-example » rubocop -V                                                                                                
0.93.1 (using Parser 2.7.2.0, rubocop-ast 0.8.0, running on ruby 2.7.2 x86_64-darwin19)
```

------

I would be happy to submit a PR if this is something that should be corrected. If not, thanks for your time!

Zajn:
Defining the match operator `=~` triggered a false positive for
`Naming/BinaryOperatorParameterName`, which should not be triggered
for pattern/matchable relations according to the [Ruby style guide](https://rubystyle.guide/#other-arg).

The method which triggered this false positive for me was as follows:

```ruby
def =~(re)
  re.match?(to_s)
end
```

-----------------

* [X] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * [#8490](https://github.com/rubocop-hq/rubocop/pull/8490): **(Breaking)** Change logic for cop department name computation. Cops inside deep namespaces (5 or more levels deep) now belong to departments with names that are calculated by joining module names starting from the third one with slashes as separators. For example, cop `Rubocop::Cop::Foo::Bar::Baz` now belongs to `Foo/Bar` department (previously it was `Bar`). ([@dsavochkin][])
 * [#8692](https://github.com/rubocop-hq/rubocop/pull/8692): Default changed to disallow `Layout/TrailingWhitespace` in heredoc. ([@marcandre][])
 * [#8894](https://github.com/rubocop-hq/rubocop/issues/8894): Make `Security/Open` aware of `URI.open`. ([@koic][])
+* [#8901](https://github.com/rubocop-hq/rubocop/issues/8901): Fix false positive for `Naming/BinaryOperatorParameterName` when defining `=~`. ([@zajn][])
 
 ## 0.93.1 (2020-10-12)
 
@@ -4998,3 +4999,4 @@
 [@ghiculescu]: https://github.com/ghiculescu
 [@hatkyinc2]: https://github.com/hatkyinc2
 [@AllanSiqueira]: https://github.com/allansiqueira
+[@zajn]: https://github.com/zajn
--- a/lib/rubocop/cop/naming/binary_operator_parameter_name.rb
+++ b/lib/rubocop/cop/naming/binary_operator_parameter_name.rb
@@ -18,7 +18,7 @@ module RuboCop
               'name its argument `other`.'
 
         OP_LIKE_METHODS = %i[eql? equal?].freeze
-        EXCLUDED = %i[+@ -@ [] []= << === `].freeze
+        EXCLUDED = %i[+@ -@ [] []= << === ` =~].freeze
 
         def_node_matcher :op_method_candidate?, <<~PATTERN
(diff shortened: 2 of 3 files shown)


marcandre:
Super, thanks!

PR #8906, opened 2020-10-19
Fix a false positive for `Layout/SpaceAroundOperators`

koic:
This PR fixes a false positive for `Layout/SpaceAroundOperators` when upward alignment. Two false positive examples are shown.

1. `@output` and `@logger` assignments are aligned with the same margin

```ruby
@integer_message = 12345
@output  = StringIO.new
@logger  = Logger.new(@output)
```

2. `expected` and `tagging` assignments are aligned with a blank line in between

```ruby
expected = posts(:welcome)

tagging  = Tagging.all.merge!(includes: :taggable).find(taggings(:welcome_general).id)
assert_no_queries { assert_equal expected, tagging.taggable }
```

I got this issue feedback from @kamipo. Thank you!

cf: https://github.com/rails/rails/pull/36943

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [#8781](https://github.com/rubocop-hq/rubocop/issues/8781): Fix handling of comments in `Style/SafeNavigation` autocorrection. ([@dvandersluis][])
 * [#8907](https://github.com/rubocop-hq/rubocop/pull/8907): Fix an incorrect auto-correct for `Layout/ClassStructure` when heredoc constant is defined after public method. ([@koic][])
 * [#8889](https://github.com/rubocop-hq/rubocop/pull/8889): Cops can use new `after_<type>` callbacks (only for nodes that may have children nodes, like `:send` and unlike `:sym`). ([@marcandre][])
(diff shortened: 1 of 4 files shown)


bbatsov:
I think the docs should be expanded to cover this. In particular I'm not quite sure that alignment should stretch over blank lines, as they are typically used to separate logical blocks, although on the other hand it's hard to imagine there are two logical blocks if the end of the previous and the begining of the next block are assignments. Tricky case, but I'm willing to accept this change if we explain the behaviour clearly.

koic:
Thank you for your feedback. I've added the doc for this change.
Let me add a little explanation. This was a potential (breaking) change vertical alignment consisting of one or more whitespace around the operator at https://github.com/rubocop-hq/rubocop/pull/7211, for part of users it was an unexpected change.
e.g. https://github.com/rails/rails/pull/36943#issuecomment-521659529

On the other hand, I agree that there are cases where https://github.com/rubocop-hq/rubocop/pull/7211 behavior is preferred rather than vertical alignment. So, It may be better to make it configurable in future.

bbatsov:
We should also add good examples with vertical alignment but this can be done later. Thanks for working on this!

PR #8916, opened 2020-10-21
[Fix #8859] Add new `Lint/UnmodifiedReduceAccumulator` cop

### LINKED ISSUE #8859: [Style/EachWithObject] Add support for showing warning for positionally misplaced block parameters in `reduce` or `each_with_object` methods
```
Platform: x86_64-linux
Shell: /bin/zsh
Ruby Version: 2.7.1
RuboCop Version: 0.92.0
Using Bundler: true
```

I'm a beginner in the Ruby world and was just trying out Rubocop. Thanks for creating this wonderful tool. 
So when I was coding, I got a suggestion from `rubocop` to replace `reduce` with `each_with_object`. I learnt that there exists a method called `each_with_object` that way, which is super helpful. Thus I auto-formatted the code to follow `each_with_object` style and the error was fixed.

This is the new rubocop formatted code with `each_with_object` style:-
```ruby
  def proj_substrings(
    base_string: "Howdy partner, sit down! How's it going?",
    dictionary: %w[below up down]
  )
    dictionary.each_with_object(Hash.new(0)) do |word, acc|
      is_substring = base_string.include? word.downcase
      acc[word] += 1 if is_substring
    end
  end
# output: {"down"=>1}
```
**Then later after a few days, inorder to show others in my circle that Rubocop provides such an intelligent suggestion I tried to revert back to  the `reduce` based code format, like so:-**
```ruby
 def proj_substrings(
    base_string: "Howdy partner, sit down! How's it going?",
    dictionary: %w[below up down]
  )
    dictionary.reduce(Hash.new(0)) do |word, acc|
      is_substring = base_string.include? word.downcase
      acc[word] += 1 if is_substring
      acc
    end
  end
``` 
So after reverting it back to above code, I couldn't get any suggestion to use `each_with_object` over `reduce`. I tried all stuffs including explicitly enabling the `Style/EachWithObject:` in `.rubocop.yml`. But still no errors were found.

Maybe it's my lack of experience, I only later saw the error in above code. **When I reverted back to `reduce` based code, I didn't swap positions of the `accumulator` and `word`. In `reduce` method, accumulator should come as first block parameter and in `each_with_object` accumulator should come as second block parameter.** 
After making that modification by swapping positions of block params, I got the correct suggestion for `Style/EachWithObject:` for below code:-
```ruby
 def proj_substrings(
    base_string: "Howdy partner, sit down! How's it going?",
    dictionary: %w[below up down]
  )
    dictionary.reduce(Hash.new(0)) do |acc, word|
      is_substring = base_string.include? word.downcase
      acc[word] += 1 if is_substring
      acc
    end
  end
```

It would be super awesome, if `rubocop` has a rule to check whether the `accumulator` or the first block parameter is the one which is returned from `reduce` method block.
Example:- if I give:-  `dictionary.reduce(Hash.new(0)) {|acc, word| word}`, then rubocop should atleast give a warning that the `The accumulator "acc" is not the one which is returned.`

Similarly if using `each_with_object`, like so:- `dictionary.each_with_object(Hash.new(0)) { |acc, word|  acc[word]+=1 }`, it should give a warning or error which says `The accumulator "word"  is not the one being modified/accumulated`

i'm not sure of the alternatives. I'm new to Ruby.

## Additional context

You can paste the above mentioned code-blocks in a ruby file and just run `rubocop` on it to see for yourself, what I mean. If rubocop can show careless mistakes like the one I had done above, then it can save a lot of time for devs I believe. 

Thanks

dvandersluis:
Adds a cop to check for `reduce` or `inject` blocks that do not modify the accumulator. Closes #8859.

I tried to do this in the safest way, and avoid false positives as much as possible. The cop takes a look at all the return values from the block and makes some judgement on them (in order):
* If the block returns the accumulator indexed (ie. `reduce { |acc, el| acc[el] += 1 }`), it'll always register an offense because that will likely cause exceptions to be thrown.
* If the block returns the accumulator, or an expression containing it, in any branch, it is considered valid and no offenses will be registered.
* If the block returns an expression containing the element and another value (or just other values), it is considered valid because we cannot determine for sure what that value is.
* Otherwise, returns containing the element are registered as offenses. 

I think I covered a lot of edge cases (running this on rubocop itself helped there), but please let me know if I've missed something, I'd like to try to avoid having a new cop immediately have a bunch of bug reports 😅 

PS: I'm not 100% sold on the name and went back and forth for a while. Any suggests are appreciated.

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -1735,6 +1735,11 @@ Lint/RandOne:
   Enabled: true
   VersionAdded: '0.36'
 
+Lint/ReduceAccumulator:
+  Description: Checks for `reduce` or `inject` blocks that do not update the accumulator each iteration.
+  Enabled: pending
+  VersionAdded: '1.1'
```
dvandersluis, line 1741: @bbatsov not sure if this is the correct version now.

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,148 @@
...
+        def_node_search :expression_values, <<~PATTERN
+          `{
+            (%RuboCop::AST::Node::VARIABLES $_)
+            (%RuboCop::AST::Node::EQUALS_ASSIGNMENTS $_ ...)
+            (send (%RuboCop::AST::Node::VARIABLES $_) :<< ...)
+            (send nil? $_)
+            (dstr (begin {(send nil? $_) (%RuboCop::AST::Node::VARIABLES $_)}))
+            (%RuboCop::AST::Node::SHORTHAND_ASSIGNMENTS (%RuboCop::AST::Node::EQUALS_ASSIGNMENTS $_) ...)
+          }
+        PATTERN
```
dvandersluis, line 72: @marcandre I feel like you probably have a simpler way to do this :grin:

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,148 @@
...
+      #   (1..4).reduce(0) do |acc, i|
+      #     acc + i * 2
+      #   end
+      #
+      #   %w(a b c).reduce({}) do |acc, letter|
+      #     acc[letter] = true
+      #     acc
+      #   end
+      class ReduceAccumulator < Base
+        MSG = 'Ensure the accumulator will be modified by `%<method>s`.'
```
bbatsov, line 42: Might be a good idea to mention the accumulator name explicitly in the messages.

bbatsov:
I like the cop, but I find the name confusing. Perhaps something like `UnusedReduceAccumulator`? I'm totally open to better suggestions myself. :D

config/default.yml
bbatsov: It is.

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,148 @@
...
+              end
+            end
+          end
+        end
+
+        def block_arg_name(node, index)
+          node.arguments[index].node_parts[0]
+        end
+
+        # Look for the an index of the accumulator being returned
```
marcandre, line 118: ^an^

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,148 @@
...
+        # This is always an offense, in order to try to catch
```
marcandre, line 119: unfinished sentence

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,148 @@
...
+      # Looks for `reduce` or `inject` blocks where the value returned (implicitly or
+      # explicitly) does not include the accumulator. A block is considered valid as
+      # long as at least one return value includes the accumulator.
+      #
+      # If the accumulator is not included in the return value, then the entire
+      # block will just return a transformation of the last element value, and
+      # could be rewritten as such without a loop.
+      #
+      # Also catches instances where an index of the accumulator is returned, as
+      # this changes the type of object being retained.
```
marcandre, line 15: this *may* change

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,148 @@
...
+      #
+      # NOTE: For the purpose of reducing false positives, this cop only flags
+      # `reduce` blocks where an expression
```
marcandre, line 18: unfinished sentence

marcandre (changes requested):
I started a review, but I think this cop is too ambitious and it's in general not possible to do what it wants to do.
I fear that the risk of false positive is too high. Unless I'm mistaken, this would be a false positive, and it would be very difficult to distinguish:
```ruby
enum.inject do |acc, elem|
  x = [*acc, elem]
  x << 42 if foo
  x
end
```

dvandersluis:
@marcandre thanks for the feedback, your example actually won't trip the cop (I can add it as a test if you'd like, I have a whole section of `does not register an offense if the return value cannot be determined` tests already [here](https://github.com/rubocop-hq/rubocop/pull/8916/files#diff-0c9784fbdc0a5d6631a93ed08c611e5d5a570839ad04c4a2966ff02d344fc634R231)).

I missed it in the description, but the cop only flags return values where the element is the *only* variable, for this exact reason. I didn't want the cop to make any assumptions here so if it's not clearly just returning the element, it does not register.

dvandersluis:
> I like the cop, but I find the name confusing. Perhaps something like `UnusedReduceAccumulator`? I'm totally open to better suggestions myself. :D

@bbatsov I agree it's not the best name. `UnmodifiedReduceAccumulator`?

dvandersluis:
@bbatsov @marcandre I updated for your feedback so far. I've also renamed the cop to `UnmodifiedReduceAccumulator` for now, still happy to rename further if a better name comes up. I've also added more examples to the documentation.

bbatsov:
I like the new name and I'm fine with merging the cop in its current state, unless @marcandre has some objections.

bbatsov (approved):

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,164 @@
...
+            (lvar %1)
+            (lvasgn %1 ...)
+            (send (lvar %1) :<< ...)
+            (dstr (begin (lvar %1)))
+            (%RuboCop::AST::Node::SHORTHAND_ASSIGNMENTS (lvasgn %1))
+          }
+        PATTERN
+
+        def_node_search :expression_values, <<~PATTERN
+          `{
```
marcandre, line 78: I imagine this is an error, as you are doing `def_node_search` which will already go through descendants, right?

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,164 @@
...
+            (%RuboCop::AST::Node::VARIABLES $_)
+            (%RuboCop::AST::Node::EQUALS_ASSIGNMENTS $_ ...)
+            (send (%RuboCop::AST::Node::VARIABLES $_) :<< ...)
+            (send nil? $_)
```
marcandre, line 82: This mistakes the name of a method with the name of a variable

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,164 @@
...
+            (dstr (begin {(send nil? $_) (%RuboCop::AST::Node::VARIABLES $_)}))
```
marcandre, line 83: Same here (first part of `{}`)

marcandre (commented):
Sorry I didn't have the time to reply earlier, and I do have objections.

It checks for two scenarios, both of which can yield false positives for perfectly valid code (in the review)

I think this should either be reverted (at least until it is revised, but again I think it is not possible to do so), or this cop should remain disabled by default.

lib/rubocop/cop/lint/reduce_accumulator.rb
dvandersluis: Yep, my mistake!

lib/rubocop/cop/lint/reduce_accumulator.rb
dvandersluis: This is actually needed, because I'm looking for identifiers in general, not specifically variables. This is valid code and should not be flagged since we can't determine what `method` is from static analysis:

```ruby
(1..4).reduce(0) do |acc, el|
  method + el
end
```

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,164 @@
...
+      # returns in `reduce` blocks where the element is the only variable in
+      # the expression (since we will not be able to tell what other variables
+      # relate to via static analysis).
+      #
+      # @example
+      #
+      #   # bad
+      #   (1..4).reduce(0) do |acc, el|
+      #     el * 2
+      #   end
```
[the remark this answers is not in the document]
dvandersluis: >Bad example, acc is not used at all. There are already cops that deal with this, and Ruby too.

Do you mean `Lint/UnusedBlockArgument` here? While that's true it doesn't really convey the same thing that this cop does. (I don't see any other offenses for the example, so I think it's valid).

I think you meant false *positives* (false negatives are inherent here on purpose), but you're right that there is a class of false positives here that I did not consider. I'm looking at this now, thanks!

lib/rubocop/cop/lint/reduce_accumulator.rb
```diff
@@ -0,0 +1,164 @@
...
+      #
+      #   # bad, may raise a NoMethodError after the first iteration
+      #   %w(a b c).reduce({}) do |acc, letter|
+      #     acc[letter] = true
+      #   end
```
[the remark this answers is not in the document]
dvandersluis: I concede that that is a valid use that will be registered by the cop, but I'd argue that this is going to be *much* less prevalent than mistakenly forgetting to return the accumulator when setting a hash key.

I see a few options:
1. We can remove this behaviour (I think this isn't ideal, because I do think this is still a good check)
2. We can put it behind a conf variable that is not enabled by default
3. We can accept the false positives (if this was my code I'd just `# rubocop:disable` it, but maybe that's suboptimal).

What do we think? /cc @bbatsov

lib/rubocop/cop/lint/reduce_accumulator.rb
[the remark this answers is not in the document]
marcandre: > I don't see any other offenses for the example, so I think it's valid

specs only check for the current cop, you can't rely on that. 

> I think you meant false positives 

Yes, sorry 😅

lib/rubocop/cop/lint/reduce_accumulator.rb
marcandre: I'm confused as to why there's `nil?` then. For example `self.method + el`, or `el.method`?

marcandre: (BTW, please do not "resolve conversation" when there's no agreement)

lib/rubocop/cop/lint/reduce_accumulator.rb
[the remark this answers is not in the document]
dvandersluis: @marcandre no I meant I ran all of rubocop on the example haha, it raise `Lint/UnusedBlockArgument` and this cop. :grin:

lib/rubocop/cop/lint/reduce_accumulator.rb
dvandersluis: You're right there shouldn't be!

lib/rubocop/cop/lint/reduce_accumulator.rb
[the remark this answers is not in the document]
marcandre: > forgetting to return the accumulator when setting a hash key.

BTW this is a double error, as `inject` probably shouldn't be used in the first place in that case (see `Style/EachWithObject`)

lib/rubocop/cop/lint/reduce_accumulator.rb
dvandersluis: > (BTW, please do not "resolve conversation" when there's no agreement)

Sorry about that!

lib/rubocop/cop/lint/reduce_accumulator.rb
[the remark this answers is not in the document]
dvandersluis: Definitely true, but in this case `Style/EachWithObject` doesn't register (because the accumulator is not being returned).

lib/rubocop/cop/lint/reduce_accumulator.rb
marcandre: > You're right there shouldn't be!

Ok, if you remove `nil?`, then `el` and `self.el` are treated the same way, even though they are not related?

dvandersluis: Good call, I've fixed that now too. (btw I've been taking all your notes and applying them locally for the purpose of a new fix PR)

dvandersluis:
@marcandre I opened #8949 to continue this.

PR #8922, opened 2020-10-22
Fix rubocop comment directives handling of cops with multiple levels in department name

### LINKED ISSUE #8917: RuboCop::CommentConfig doesn't support X/Y/Z format cops
## Expected behavior

Now that the cop naming behavior has changed to support X/Y/Z cop format you should be able to enable/disable those via comments as well.

## Actual behavior

These comments aren't recognized. It looks like the COP_NAME_PATTERN regex pattern needs to be updated and a spec for the X/Y/Z format added..

Try to disabled a with an X/Y/Z name. Doesn't do anything.

## RuboCop version
```
$ [bundle exec] rubocop -V
1.0.0 (using Parser 2.7.2.0, rubocop-ast 1.0.0, running on ruby 2.7.2 x86_64-darwin19)
```

fatkodima:
Closes #8917

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ### Bug fixes
 
 * [#8913](https://github.com/rubocop-hq/rubocop/pull/8913): Fix an incorrect auto-correct for `Style/RedundantRegexpCharacterClass` due to quantifier. ([@ysakasin][])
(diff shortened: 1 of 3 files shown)


tas50:
Looks good to me. Thanks @fatkodima

koic:
Thanks!

PR #8924, opened 2020-10-22
Fix a false positives for `Bundler/DuplicatedGem` when a gem conditionally duplicated within `if-elsif` or `case-when` statements

### LINKED ISSUE #8918: False positive for Bundler/DuplicatedGem
Should not complain

```ruby
  if File.exist?(local_ast)
    gem 'rubocop', path: local_ast
  elsif ENV['RUBOCOP_VERSION'] == 'master'
    gem 'rubocop', git: 'https://github.com/rubocop-hq/rubocop.git'
  else
    gem 'rubocop', '>= 1.0'
  end
```

cc/ @tejasbubane 

I didn't check, but `case/when` should also work...

fatkodima:
Closes #8918


bbatsov:
The changes look good. I think we should expand the cop's documentation to cover better the cases that the cop can handle, as they are pretty basic right now (probably a remnant of the original implementation that was more limitted in its scope).

fatkodima:
The current documentation looks pretty good to me. I've expanded it with a section about conditional declaration.

bbatsov:
Thanks!

PR #8929, opened 2020-10-23
Add allowed method for Lint/Debugger

mfbmina:
This PR fixes https://github.com/rubocop-hq/rubocop/issues/8927.

It allows the user to add AllowedMethods for Lint/Debugger cop.

- [ ] Wrote good commit messages.
- [ ]  Commit message starts with [Fix #issue-number] (if the related issue exists).
- [ ]  Feature branch is up-to-date with master (if not - rebase it).
- [ ]  Squashed related commits together.
- [ ]  Added tests.
- [ ]  Added an entry to the Changelog if the new code introduces user-observable changes. See changelog entry format.
- [ ]  The PR relates to only one subject with a clear title and description in grammatically correct, complete sentences.
- [ ]  Run bundle exec rake default. It executes all tests and RuboCop for itself, and generates the documentation.


bbatsov:
You should also update the docs and the default configuration. And remove the hardcoded methods.

mfbmina:
@bbatsov please check again!

config/default.yml
```diff
@@ -1415,6 +1415,18 @@ Lint/Debugger:
   Enabled: true
   VersionAdded: '0.14'
   VersionChanged: '0.49'
+  AllowedMethods:
```
robotdana, line 1418: this seems the opposite of Allowed, should this be `DisallowedMethods`? `ReportedMethods`? `DebuggerMethods`? just `Methods`?

bbatsov: Yep, you're toally right. I like the name `DebuggerMethods` best, although I'm also fine with "ForbiddenMethods".

mfbmina: I also felt like this when implementing... shall we go with `DebuggerMethods` then?

bbatsov: Yep.

bbatsov:
> please check again!

Yout still have replace the hardcoded node patterns with something that's generated dynamically based on the cop's configuration.

mfbmina:
> Yout still have replace the hardcoded node patterns with something that's generated dynamically based on the cop's configuration.

My solution for that was adding a `ForbiddenReceivers` options which will help to specify which receivers are not allowed, like `binding` or `Kernel`. 

The only edge case that I couldn't cover and had to add a special check when was `Kernel.save_and_open_page` which should be allowed even tho `Kernel` is not an allowed receiver and `save_and_open_page` is a debugger method.

mfbmina:
@bbatsov @robotdana what you guys think?

config/default.yml
```diff
@@ -1415,6 +1415,22 @@ Lint/Debugger:
   Enabled: true
   VersionAdded: '0.14'
   VersionChanged: '0.49'
+  ForbiddenReceivers:
```
bbatsov, line 1418: We can name this "DebuggerReceivers" for consistency.

lib/rubocop/cop/mixin/forbidden_receivers.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # This module encapsulates the ability to forbid certain method receivers
+    # when parsing.
+    module ForbiddenReceivers
```
bbatsov, line 7: I don't think we need a module for this, so I'd fold it back into the cop.

bbatsov:
I think we're pretty close to merging this. I left a couple of small remarks. You'll also need to update the cop description to mention the configuration options and the rebase on top of master.

lib/rubocop.rb
```diff
@@ -69,6 +69,7 @@
 require_relative 'rubocop/cop/mixin/configurable_formatting'
 require_relative 'rubocop/cop/mixin/configurable_naming'
 require_relative 'rubocop/cop/mixin/configurable_numbering'
+require_relative 'rubocop/cop/mixin/debugger_methods'
```
bbatsov, line 72: I think this mixin should be inlined as well. I doubt we'll ever reuse it.

lib/rubocop/cop/lint/debugger.rb
```diff
@@ -72,8 +52,26 @@ def message(node)
           format(MSG, source: node.source)
         end
 
-        def binding_irb?(node)
-          binding_irb_call?(node)
+        def with_receiver?(node)
+          !!node.receiver
+        end
+
+        def special_rule?(node)
```
bbatsov, line 59: I don't get what's the purpose of this `special_rule?` method.

lib/rubocop/cop/lint/debugger.rb
```diff
@@ -72,8 +52,26 @@ def message(node)
...
+          return false unless node.receiver.is_a?(RuboCop::AST::ConstNode)
+          return true if node.receiver.const_name == 'Kernel' && node.method?(:save_and_open_page)
```
bbatsov, line 61: In particular it's very weird we hardcoded this.

lib/rubocop/cop/lint/debugger.rb
```diff
@@ -33,35 +34,14 @@ module Lint
       #     do_something
       #   end
       class Debugger < Base
-        MSG = 'Remove debugger entry point `%<source>s`.'
-
-        RESTRICT_ON_SEND = %i[
```
bbatsov, line 38: You should keep this constant, as it's special for RuboCop (it limits the `on_send` callbacks). Just initialize it with the data from the config.

mfbmina: The problem if I hardcode the values from the configuration in this constant is not recognizing new values from the config.

This test, for example, start failing: https://github.com/rubocop-hq/rubocop/blob/12bd06882eaca12bd33c77f31e8d976bc3ad0bed/spec/rubocop/cop/lint/debugger_spec.rb#L15-L20

What I did, for now, is initializing the constant but with an empty array.

lib/rubocop/cop/lint/debugger.rb
mfbmina: this is an edge case that I couldn't find a better fix. Maybe you can help me:

Even tho `Kernel` is a receiver and `save_and_open_page` is a debugger method together they should not be an offense.

Here is the test case for it: https://github.com/rubocop-hq/rubocop/blob/12bd06882eaca12bd33c77f31e8d976bc3ad0bed/spec/rubocop/cop/lint/debugger_spec.rb#L164-L166

lib/rubocop/cop/lint/debugger.rb
mfbmina: I also don't like it. Look above why I did it.

lib/rubocop/cop/lint/debugger.rb
bbatsov: Ah, yeah - that's a fair point.

lib/rubocop/cop/lint/debugger.rb
bbatsov: I'd just drop this test case. For such a small thing people should add a local exclusion rule. It just doesn't fit with having with our goal to have a configurable cop.

mfbmina: Cool! I will drop it! It was introduced here. https://github.com/rubocop-hq/rubocop/commit/83902a6a4b957130abcabdebe1c266a12ec5d138

lib/rubocop/cop/lint/debugger.rb
```diff
@@ -72,8 +51,23 @@ def message(node)
           format(MSG, source: node.source)
         end
 
-        def binding_irb?(node)
-          binding_irb_call?(node)
+        def with_receiver?(node)
```
bbatsov, line 54: I see you're using this method only once, so I'd inline it in `forbidden_receiver?`.

lib/rubocop/cop/lint/debugger.rb
```diff
@@ -72,8 +51,23 @@ def message(node)
...
+          !!node.receiver
+        end
+
+        def debugger_method?(name)
+          cop_config.fetch('DebuggerMethods', []).include?(name.to_s)
+        end
+
+        def forbidden_receiver?(node)
```
bbatsov, line 62: I'd name this `debugger_receiver?`.

bbatsov (approved):

PR #8934, opened 2020-10-25
Add new `Style/SwapValues` cop

fatkodima:
This cop enforces the use of shorthand-style swapping of 2 variables.
```ruby
# bad
tmp = x
x = y
y = tmp

# good
x, y = y, x
```

This will be mostly useful for newcomers from other languages into ruby, because when you used it once you will always remember about its existence.


bbatsov:
Great idea! Seems that we should also add this to the style guide.

docs/modules/ROOT/pages/cops_style.adoc
```diff
@@ -9897,6 +9897,35 @@ end
...
+
+| Pending
+| Yes
+| Yes (Unsafe)
+| 1.1
+| -
+|===
+
+This cop enforces the use of shorthand-style swapping of 2 variables.
+It is marked as unsafe, because while autocorrecting it can erroneously remove
```
bbatsov, line 9913: The wording here is a bit misleading, as the cop is safe and only the auto-correct is unsafe.

fatkodima:
Added link to the style guide (https://github.com/rubocop-hq/ruby-style-guide/pull/853) and changed wording.

bbatsov:
Thanks!

pvdb:
There's an issue #10394 with this cop that I just ran into.

Issue #8127, opened 2020-06-09
Old specs need love


marcandre:
Many cop specs still use `inspect_source` or `autocorrect_source`. We'll like to deprecated this (see #8003), so they should be refactored to use `expect_offense` + `expect_correction` instead (I'd love to have an even nicer API for this).

rrosenblum:
Some of these are much easier to convert than others. Some of the tests are as easy as merging an inspect test with and autocorrect test and rewriting a couple of lines. 

The tests that make use of `shared_examples` will likely need to be rewritten to not use example groups. Since the code is dynamic, you likely won't be able to get the highlighting to work in a clean way.

I haven't found a way to use `expect_offense` on any test that points to blank line. `expect_offense` doesn't have a way to process highlight that doesn't really point to anything https://github.com/rubocop-hq/rubocop/blob/3b0552b9989db97104cfe0fc30515e9b6ae549f2/spec/rubocop/cop/layout/empty_lines_around_method_body_spec.rb#L42-L51

```ruby
  it 'registers an offense for class method body starting with a blank' do
    expect_offense(<<~RUBY)
      def Test.some_method

      ^ Extra empty line detected at method body beginning.
        do_something
      end
    RUBY
  end
```
```shell
Failures:

  1) RuboCop::Cop::Layout::EmptyLinesAroundMethodBody registers an offense for class method body starting with a blank
     Failure/Error: expect(actual_annotations.to_s).to eq(expected_annotations.to_s)

       expected: "def Test.some_method\n\n^ Extra empty line detected at method body beginning.\n  do_something\nend\n"
            got: "def Test.some_method\n\n Extra empty line detected at method body beginning.\n  do_something\nend\n"

       (compared using ==)

       Diff:
       @@ -1,6 +1,6 @@
        def Test.some_method

       -^ Extra empty line detected at method body beginning.
       + Extra empty line detected at method body beginning.
          do_something
        end

     # ./lib/rubocop/rspec/expect_offense.rb:125:in `expect_offense'
     # ./spec/rubocop/cop/layout/empty_lines_around_method_body_spec.rb:43:in `block (2 levels) in <top (required)>'

Finished in 0.04196 seconds (files took 1.03 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/rubocop/cop/layout/empty_lines_around_method_body_spec.rb:42 # RuboCop::Cop::Layout::EmptyLinesAroundMethodBody registers an offense for class method body starting with a blank
[...]

PR #8936, opened 2020-10-25
refactor the following layout specs to use 'expect_offense':

mechos3d:
Part of #8127
Refactored specs in spec/rubocop/cop/layout/ to use 'expect_offense' matcher :  

- end_of_line_spec.rb, 
- extra_spacing_spec.rb, 
- indentation_width_spec.rb,
- trailing_empty_lines_spec.rb, 
- trailing_whitespace_spec.rb
-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/rspec/expect_offense.rb
+++ b/lib/rubocop/rspec/expect_offense.rb
@@ -113,8 +113,6 @@ module RuboCop
 
       # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       def expect_offense(source, file = nil, severity: nil, **replacements)
-        source = "#{source.join("\n")}\n" if source.is_a?(Array)
-
         source = format_offense(source, **replacements)
         RuboCop::Formatter::DisabledConfigFormatter
           .config_to_allow_offenses = {}
@@ -146,8 +144,6 @@ module RuboCop
       end
 
       def expect_correction(correction, loop: true)
-        correction = "#{correction.join("\n")}\n" if correction.is_a?(Array)
-
         raise '`expect_correction` must follow `expect_offense`' unless @processed_source
 
         iteration = 0
@@ -188,8 +184,6 @@ module RuboCop
       end
 
       def expect_no_offenses(source, file = nil)
-        source = "#{source.join("\n")}\n" if source.is_a?(Array)
-
         offenses = inspect_source(source, file)
 
         expected_annotations = AnnotatedSource.parse(source)
--- a/lib/rubocop/rspec/expect_offense.rb
+++ b/lib/rubocop/rspec/expect_offense.rb
@@ -113,6 +113,8 @@ module RuboCop
 
       # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       def expect_offense(source, file = nil, severity: nil, **replacements)
+        source = "#{source.join("\n")}\n" if source.is_a?(Array)
+
         source = format_offense(source, **replacements)
         RuboCop::Formatter::DisabledConfigFormatter
           .config_to_allow_offenses = {}
@@ -144,6 +146,8 @@ module RuboCop
       end
 
       def expect_correction(correction, loop: true)
+        correction = "#{correction.join("\n")}\n" if correction.is_a?(Array)
+
         raise '`expect_correction` must follow `expect_offense`' unless @processed_source
 
         iteration = 0
@@ -184,6 +188,8 @@ module RuboCop
       end
 
       def expect_no_offenses(source, file = nil)
+        source = "#{source.join("\n")}\n" if source.is_a?(Array)
+
         offenses = inspect_source(source, file)
 
         expected_annotations = AnnotatedSource.parse(source)
--- a/spec/rubocop/cop/layout/end_of_line_spec.rb
+++ b/spec/rubocop/cop/layout/end_of_line_spec.rb
@@ -1,9 +1,6 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::Layout::EndOfLine, :config do
-  let(:message)          { 'Carriage return character missing.' }
-  let(:detected_message) { 'Carriage return character detected.' }
-
   shared_examples 'all configurations' do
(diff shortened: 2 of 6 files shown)


spec/rubocop/cop/layout/trailing_empty_lines_spec.rb
```diff
@@ -106,25 +99,33 @@
     end
 
     it 'registers an offense for multiple trailing blank lines' do
-      offenses = inspect_source(<<~RUBY)
+      expect_offense(
+        ['x = 0',
```
marcandre, line 103: Please use heredoc instead of array

spec/rubocop/cop/layout/end_of_line_spec.rb
```diff
@@ -1,56 +1,63 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::Layout::EndOfLine, :config do
+  let(:message)          { 'Carriage return character missing.' }
```
marcandre, line 4: Could you please copy paste these messages instead? Most specs are like this. Note you can abbreviate with `[...]` when the message is long.

lib/rubocop/rspec/expect_offense.rb
```diff
@@ -113,6 +113,8 @@ def format_offense(source, **replacements)
 
       # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       def expect_offense(source, file = nil, severity: nil, **replacements)
+        source = "#{source.join("\n")}\n" if source.is_a?(Array)
```
marcandre, line 116: I'm :-1: on this, I much prefer if we use Heredocs instead of arrays of lines which aren't nearly as readable. Let me know if I missed a use-case.

spec/rubocop/cop/layout/end_of_line_spec.rb
```diff
@@ -1,56 +1,63 @@
...
-      expect(cop.offenses.empty?).to be(true)
+      expect_no_offenses('')
     end
   end
 
   shared_examples 'iso-8859-15' do |eol|
     it 'can inspect non-UTF-8 encoded source with proper encoding comment' do
-      inspect_source_file(["# coding: ISO-8859-15#{eol}",
-                           "# Euro symbol: \xa4#{eol}"].join("\n"))
+      inspect_source_file(<<~RUBY)
```
marcandre, line 15: Would it be possible to change this to use `expect_offense`?

marcandre (changes requested):
Good PR. There are a few things I'm requesting, we can discuss them, or else please apply them everywhere. Thanks!

lib/rubocop/rspec/expect_offense.rb
mechos3d: Now I agree that it looks more like a hacky workaround than a good solution (because it is :) ).          
As to why I did this: for trailing whitespaces. 
In cases like this - it's not clearly seen are there whitespaces, how much of them are there.       
Also some folks have removing trailing whitespace done automatically by the IDE on save - so one must be careful with these specs.
```ruby
    expect_offense(<<~RUBY)
        
      ^^ Trailing whitespace detected.\n
    RUBY
```
But later I've found a helper method `trailing_whitespace` in (`/spec/support/misc_helper.rb`) - I think it will solve the problem like this: 
 ```ruby
    expect_offense(<<~RUBY)
      #{trailing_whitespace}  
      ^^ Trailing whitespace detected.\n
    RUBY
```    
What do you think ?

File: spec/rubocop/cop/layout/end_of_line_spec.rb (hunk 4 above)
Comment (mechos3d) [reply spec/rubocop/cop/layout/end_of_line_spec.rb:15]: Yeah, it would be.   
I wanted to do it in a separate PR, but we can solve this one right here actually.         
The main goal of this spec - is to test how rubocop works with a file in non-utf8 encoding.  
So reading from a file and respecting magic comment seems like an important part of the process.      

Although maybe it's safe to test it this way: 
```ruby
expect_offense(<<~RUBY.force_encoding('ISO-8859-15'))
   .... 
RUBY
```       
Dealing with the magic comments is not part of rubocop's functionality, so there's no need to test this, we can rely on ruby itself on that. 
What do you think, is the `force_encoding` approach ok?

mechos3d: `def expect_offense(source, file = nil, severity: nil, **replacements)`
Oh, I see now.     
Another way would be just to send some temporary file path to `expect_offense` as an argument..

marcandre: Oh, I missed the encoding part. No problem then (maybe add a comment)

lib/rubocop/rspec/expect_offense.rb
marcandre: Indeed. BTW RuboCop will flag and auto-correct trailing whitespaces in heredocs now too.

spec/rubocop/cop/layout/trailing_empty_lines_spec.rb
```diff
@@ -30,67 +30,60 @@
...
 
       RUBY
-      expect(offenses.size).to eq(1)
-      expect(offenses.first.message).to eq('3 trailing blank lines detected.')
+
+      expect_correction("\n")
     end
 
     it 'registers an offense for no final newline after assignment' do
       offenses = inspect_source('x = 0')
```
mechos3d, line 57: There's another interesting place - the case doesn't work with `expect_offense`, because typing
```
expect_offense(<<~RUBY)
  x = 0
       ^ Final newline is missing
RUBY
```
Would actually have a newline after `x = 0` and will not cause an offense.  

I have some thoughts on how to work around it (possibly with another specific matcher), but I would like to leave it to a separate PR

marcandre: 👍 😆

marcandre: That's perfect then.

marcandre:
Great work, thank you!

PR #8941, opened 2020-10-25
Add new `Lint/ToEnumArguments` cop

### LINKED ISSUE #7753: Cop idea: Not enough arguments for enum_for/to_enum

We have an idiom of `enum_for`/`to_enum` method for defining an iterator.

```ruby
class C
  def each_item(&block)
    return enum_for(__method__) unless block_given?

    iterate_someting_with_block block
  end
end

c = C.new

# We can call C#each with and without a block.
c.each_item do |item|
end
c.each_item.map do |item|
end
```

It is a useful idiom! But we sometimes make a bug if the iterator receives arguments. For example:

```ruby
class C
  def each_item(type, &block)
    return enum_for(__method__) unless block_given?

    iterate_something_with_block type, block
  end
end
c = C.new

# It works well.
c.each_item(:foo) do |item|
end

# But it does not work well because enum_for should receive `type` but actually doesn't.
c.each_item(:foo).map do |item|
end
```

It should be `return enum_for(__method__, type) unless block_given?`.

In this case, it is not a serious problem because `c.each_item(:foo).map{}` raises an `ArgumentError`.
But it will be serious when the argument is optional, like `def each_item(type = :default, &block)`.
It ignores `type` argument silently.

Add a new cop to detect this problem.

Nothing

## Additional context

* to_enum is an alias of enum_for
* Probably we should focus the idiom, it means the cop should accept any `enum_for` without `return enum_for(__method__) unless block_given?` form.

fatkodima:
Closes #7753 

This cop ensures that `to_enum`/`enum_for`, called for the current method, has correct arguments.
```ruby
# bad
def method(x, y = 1)
  return to_enum(__method__, x) # `y` is missing
end
      
# good
def method(x, y = 1)
  return to_enum(__method__, x, y)
end
```

I have made this mistake myself 😄 - [proof](https://github.com/mperham/sidekiq/pull/4328/files)


bbatsov:
I don't think I've ever used those methods, so I'll leave it to the others to provide feedback for that cop. Some more rationale might be useful, as others like me might first think this is something about enumerations. :D

spec/rubocop/cop/lint/to_enum_arguments_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      def m(x, &block)
+        return to_enum(:m, x) unless block_given?
+      end
+    RUBY
+  end
+
+  it 'does not register an offense when enumerator is created for other method' do
+    expect_no_offenses(<<~RUBY)
+      def m(x)
+        return to_enum(:not_m) unless block_given?
```
marcandre, line 112: I'm not sure what's the use-case for allowing this tbh. If receiver is `self`, then method should match...

marcandre (commented):
Good cop! I've made this kind of mistake a few times.

Need to get to bed, just checked the tests, looks good, I'd be even stricter.

marcandre:
Also, could you support forwarding please? (`...`, Ruby 2.7+)

fatkodima:
Did you mean something like
```ruby
def meth(...)
  return to_enum(...) unless block_given?
end
```
Is there a usecase for this?

It does not make much sense to me, compared to
```ruby
def meth(...)
  return to_enum(:meth, ...) unless block_given?
end
```
which is an invalid syntax, unfortunately.

Also, currently the first example passes without violation.

marcandre:
> It does not make much sense to me, compared to
> 
> ```ruby
> def meth(...)
>   return to_enum(:meth, ...) unless block_given?
> end
> ```

This is what I meant, and it is Ruby 3.0+ only, my bad. Still would be nice to do

fatkodima:
Added a `TODO` for the future, as it is easy to implement, but hard to test before 3.0 is released.

marcandre:
> Added a `TODO` for the future, as it is easy to implement, but hard to test before 3.0 is released.

As you wish. FYI, we already run our CI against `ruby-head`, and testing locally is easy with `rvm install ruby-head`.

marcandre (approved):

marcandre:
@bbatsov ok to merge?

bbatsov:
Yep.

bquorning:
I am not sure I completely understand this cop. I am not sure that the following code should be flagged:

```rb
def foo
  bar = to_enum(:baz).count
  bar > 42
end
```

Is it a bug, or have I just misunderstood this cop?

bquorning:
> Is it a bug, or have I just misunderstood this cop?

I opened an issue: #12885

PR #8945, opened 2020-10-26
Fix changelog item generation

dvandersluis:
Re #8930. I found a couple issues when using `rake changelog:fix` to write a changelog item for another PR. Foremost, `[Fix #xxxx]` as requested in the contributing guidelines was not being parsed properly (it was only looking for `Fixes`, now it accepts both). `tasks/changelog.rb` did not have a spec, so I added a basic one that just covers this issue but is not exhaustive.

I also updated the tests in `project_spec.rb`:
* Only links at the start of a changelog item are evaluated (eg. 7th item here: https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md#bug-fixes-18)
* Issues on other rubocop-hq repos are accepted (eg. last item here: https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md#bug-fixes-18)
* Ensures that a changelog item does not have `[Fix #xxxx]` in it.
* Flag `[#x]` as invalid in a changelog item

The last item is potentially contentious because the rake task prefaces items with `* [#x](https://github.com/rubocop-hq/rubocop/pull/x):` if a ref ID cannot be found, which will then cause `bundle exec rake default` to fail. This is kind of a paradox if there isn't already an issue to refer to (ie. sometimes in practice the link is to the PR that merges the change, which doesn't exist before `rake default` is run), so I'm not 100% sure about this, but I think the test is valuable in general to ensure we don't end up with `[#x]` in the changelog.

/cc @marcandre 

-----------------

* [x] Added tests.

### DIFF
--- a/tasks/changelog.rb
+++ b/tasks/changelog.rb
@@ -51,7 +51,7 @@ class Changelog
     end
 
     def extract_id(body)
-      /^\[Fixes #(\d+)\] (.*)/.match(body)&.captures || [nil, body]
+      /^\[Fix(?:es)? #(\d+)\] (.*)/.match(body)&.captures || [nil, body]
     end
 
     def str_to_filename(str)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3079,7 +3079,7 @@
 * [#2146](https://github.com/rubocop-hq/rubocop/pull/2146): Add STDIN support. ([@caseywebdev][])
 * [#2175](https://github.com/rubocop-hq/rubocop/pull/2175): Files that are excluded from a cop (e.g. using the `Exclude:` config option) are no longer being processed by that cop. ([@bquorning][])
 * `Rails/ActionFilter` now handles complete list of methods found in the Rails 4.2 [release notes](https://github.com/rails/rails/blob/4115a12da1409c753c747fd4bab6e612c0c6e51a/guides/source/4_2_release_notes.md#notable-changes-1). ([@MGerrior][])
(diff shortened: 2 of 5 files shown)


marcandre:
> `tasks/changelog.rb` did not have a spec, so I added a basic one that just covers this issue but is not exhaustive.

Oh, right, I forgot to copy it over from `rubocop-ast`. Done in #8946.

You changes are great. How about moving your spec to `spec/tasks/changelog_spec.rb`?

spec/spec_helper.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+$LOAD_PATH.unshift("#{__dir__}/../tasks")
```
marcandre, line 3: Best not mess with `$LOAD_PATH`. In my spec I simply use `require_relative`

tasks/changelog.rb
```diff
@@ -51,7 +51,7 @@ def last_commit_title
     end
 
     def extract_id(body)
-      /^\[Fixes #(\d+)\] (.*)/.match(body)&.captures || [nil, body]
+      /^\[Fix(?:es)? #(\d+)\] (.*)/.match(body)&.captures || [nil, body]
```
marcandre, line 54: Cool, didn't know we could write either 😅

spec/spec_helper.rb
dvandersluis: I had that first but it was about 5 layers of `../`. Don't mind doing that though if desired!

dvandersluis:
@marcandre updated, merged my spec into yours, thanks!

marcandre (approved):
Good job 💪 ! Thanks 😄

PR #8949, opened 2020-10-26
Fix issues with `Lint/UnmodifiedReduceAccumulator`

dvandersluis:
Fix for issues brought up after merge in #8916 (please see discussion there).

-----------------

* [x] Added tests.


dvandersluis:
Still unresolved: given that checking for returning an index of the accumulator might be a false positive, how should we proceed?

<blockquote>
I see a few options:

1. We can remove this behaviour (I think this isn't ideal, because I do think this is still a good check)
2. We can put it behind a conf variable that is not enabled by default
3. We can accept the false positives (if this was my code I'd just `# rubocop:disable` it, but maybe that's suboptimal).
</blockquote>

bbatsov:
> We can accept the false positives (if this was my code I'd just # rubocop:disable it, but maybe that's suboptimal).

I have a feeling there won't be many of those, so we can launch the cop and see what kind of feedback we'll get.

dvandersluis:
@bbatsov not sure what to do for that ruby 2.4 spec failure? I ran it locally on 2.4 and it passes, and it doesn't seem to be about code I've changed here.

PR #8954, opened 2020-10-27
[Fix #8948] Fix autocorrection for Style/RedundantRegexpCharacterClass with %r

### LINKED ISSUE #8948: Invalid autocorrection with Style/RedundantRegexpCharacterClass
## Expected behavior

Regex autocorrect is a valid regex.

## Actual behavior

The offense is partially corrected and an invalid regex is produced. This is a similar problem to the one that was fixed in https://github.com/rubocop-hq/rubocop/pull/8913, but the issue still happens against current master with that fix.

Autocorrect:

```ruby
MATCHES_CHEF_GEM ||= %r{/chef-[\d]+\.[\d]+\.[\d]+}.freeze
```

Correct is to:

```ruby
MATCHES_CHEF_GEM ||= %r{/che-[\d+.[\d+.[\d+}.freeze
```

## RuboCop version

```
master
```

ysakasin:
This commit uses `RegexpParser::Map#body` instead of `Parser::Source::Range#adjust` in Style/RedundantRegexpCharacterClass.

When regular expression are written by %r literal, RedundantRegexpCharacterClass made incorrect loc as following.
```
%r{abc[\d]+}
    ^^^^^^
```

This bug are caused by calling`node.loc.begin.adjust` to get range of character class with %r regexp.
`Parser::Source::Range#adjust` adjust based on `Parser::Source::Range#begin_pos` but it with %r literal doesn't indicate start of regular expression source, point to location of '%'.
So, RedundantRegexpCharacterClass makes incorrect loc if %r literal are used in code.

```
r = %r{abc}
    ^      # `node.loc.begin.begin_pos`
```

```
r = %r{abs}
       ^   # `node.loc.begin.end_pos`
```

Ref. #8948 

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * [#8917](https://github.com/rubocop-hq/rubocop/issues/8917): Fix rubocop comment directives handling of cops with multiple levels in department name. ([@fatkodima][])
 * [#8918](https://github.com/rubocop-hq/rubocop/issues/8918): Fix a false positives for `Bundler/DuplicatedGem` when a gem conditionally duplicated within `if-elsif` or `case-when` statements. ([@fatkodima][])
 * [#8933](https://github.com/rubocop-hq/rubocop/pull/8933): Fix an error for `Layout/EmptyLinesAroundAccessModifier` when the first line is a comment. ([@matthieugendreau][])
+* [#8954](https://github.com/rubocop-hq/rubocop/pull/8954): Fix autocorrection for Style/RedundantRegexpCharacterClass with %r. ([@ysakasin][])
 
 ### Changes
 
--- a/docs/modules/ROOT/pages/cops_style.adoc
+++ b/docs/modules/ROOT/pages/cops_style.adoc
@@ -8230,6 +8230,12 @@ r = /[\s]/
 # good
 r = /\s/
 
+# bad
+r = %r{/[b]}
+
+# good
+r = %r{/b}
+
 # good
 r = /[ab]/
 ----
--- a/lib/rubocop/cop/style/redundant_regexp_character_class.rb
+++ b/lib/rubocop/cop/style/redundant_regexp_character_class.rb
@@ -19,6 +19,12 @@ module RuboCop
       #   # good
       #   r = /\s/
       #
+      #   # bad
+      #   r = %r{/[b]}
+      #
+      #   # good
+      #   r = %r{/b}
+      #
       #   # good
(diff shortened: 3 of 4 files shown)


owst:
Thanks for fixing this @ysakasin!

I [recently](https://github.com/rubocop-hq/rubocop/pull/8896) added `RegexpNode#parsed_tree_expr_loc`: https://github.com/rubocop-hq/rubocop/blob/0259bed3df30e7d0e18645e651a683b890d03db4/lib/rubocop/ext/regexp_node.rb#L42-L44 which correctly uses `loc.begin.end`. I wonder if we could alter that method so that it accepts an optional explicit `length` argument that you could use in this PR?

What about something like:
```ruby
def parsed_tree_expr_loc(expr, loc_size: expr.full_length)
  loc.begin.end.adjust(begin_pos: expr.ts, end_pos: expr.ts).resize(loc_size)
end
```

ysakasin:
@owst Thanks reviewing!

I modify to use `RegexpNode#parsed_tree_expr_loc` and add keyword argument `loc_size` to it.

marcandre:
It's very clucky to have to provide `parsed_tree_expr_loc` with the right answer.
At minimum, that method should check the type of the node and adjust if need be (like here).
Probably better would be to monkey patch Regexp's parser Nodes and add a `loc` method that would provide us with the right information.
For a char class like here, you could use `loc.expression` and that's it.

ysakasin:
Like following?

```ruby
        def each_redundant_character_class(node)
          each_single_element_character_class(node) do |char_class|
            next unless redundant_single_element_character_class?(node, char_class)

            # This means that: '['.length + len_char_class_source + ']'.length
            len_char_class = 1 + char_class.expressions.first.text.length + 1

            yield node.loc.expression.adjust(begin_pos: char_class.ts).resize(len_char_class)
          end
        end
```

marcandre:
Not what I meant. Let me prepare a PR

marcandre:
I created #8960. I believe that the correction would then simply be to `remove` the `loc.begin` and `loc.end`.

ysakasin:
@ysakasin Thanks adding awesome method!
I replaced `RegexpNode#parsed_tree_expr_loc` with `RegexpParser::Map#body`.

ysakasin:
@marcandre Thanks adding awesome method!
I replaced `RegexpNode#parsed_tree_expr_loc` with `RegexpParser::Map#body`.

bbatsov:
Looks great! Thanks!

PR #8955, opened 2020-10-27
Fix a false positive for `Lint/ToEnumArguments`

koic:
Follow https://github.com/rubocop-hq/rubocop/pull/8941#discussion_r511770777.

This PR fixes the following false positive for `Lint/ToEnumArguments` when enumerator is created for another method.

```console
% cd path/to/rubocop-performance
% bundle exec rubocop lib/rubocop/cop/performance/redundant_block_call.rb
Inspecting 1 file
W

Offenses:

lib/rubocop/cop/performance/redundant_block_call.rb:83:19: W:
Lint/ToEnumArguments: Ensure you correctly provided all the arguments.
          calls = to_enum(:blockarg_calls, body, argname)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

The following is a snippet of `Performance/RedundantBlockCall` cop that is a real world use-case.

```ruby
def_node_search :blockarg_calls, <<~PATTERN
  (send (lvar %1) :call ...)
PATTERN

...

def calls_to_report(argname, body)
  return [] if blockarg_assigned?(body, argname)

  calls = to_enum(:blockarg_calls, body, argname)

  return [] if calls.any? { |call| args_include_block_pass?(call) }

  calls
end
```

https://github.com/rubocop-hq/rubocop-performance/blob/v1.8.1/lib/rubocop/cop/performance/redundant_block_call.rb#L80-L88

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/to_enum_arguments_spec.rb
+++ b/spec/rubocop/cop/lint/to_enum_arguments_spec.rb
@@ -106,11 +106,10 @@ RSpec.describe RuboCop::Cop::Lint::ToEnumArguments do
     RUBY
   end
 
-  it 'registers an offense when enumerator is created for another method' do
-    expect_offense(<<~RUBY)
+  it 'does not register an offense when enumerator is created for another method' do
+    expect_no_offenses(<<~RUBY)
       def m(x)
-        return to_enum(:not_m) unless block_given?
-               ^^^^^^^^^^^^^^^ Ensure you correctly provided all the arguments.
+        return to_enum(:not_m, x)
       end
     RUBY
   end
--- a/lib/rubocop/cop/lint/to_enum_arguments.rb
+++ b/lib/rubocop/cop/lint/to_enum_arguments.rb
@@ -33,11 +33,7 @@ module RuboCop
         RESTRICT_ON_SEND = %i[to_enum enum_for].freeze
 
         def_node_matcher :enum_conversion_call?, <<~PATTERN
-          (send {nil? self} {:to_enum :enum_for} $_ $...)
-        PATTERN
-
-        def_node_matcher :method_name?, <<~PATTERN
-          {(send nil? :__method__) (sym %1)}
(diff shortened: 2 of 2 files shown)


spec/rubocop/cop/lint/to_enum_arguments_spec.rb
```diff
@@ -106,11 +106,10 @@ def m(x, &block)
...
   end
 
-  it 'registers an offense when enumerator is created for another method' do
-    expect_offense(<<~RUBY)
+  it 'does not register an offense when enumerator is created for another method' do
+    expect_no_offenses(<<~RUBY)
       def m(x)
-        return to_enum(:not_m) unless block_given?
-               ^^^^^^^^^^^^^^^ Ensure you correctly provided all the arguments.
+        return to_enum(:not_m, x)
```
fatkodima, line 112: But when `return to_enum(:not_m)` (as in the original) is used - this should also not register an offense. But it erroneously generates one.

marcandre:
Not sure that's a false negative. Correct code should be:

```ruby
calls = blockarg_calls(body, argname)
```

marcandre:
PS: The code above does the search twice (at least), so it could be improved to:

```ruby
def calls_to_report(argname, body)
  return [] if blockarg_assigned?(body, argname)

  to_enum(:blockarg_calls, body, argname).map do |call|
    return [] if args_include_block_pass?(call)
    call
  end
end
```

spec/rubocop/cop/lint/to_enum_arguments_spec.rb
marcandre: It's a code smell I think. Maybe parameter (default = don't accept)?

koic:
> Not sure that's a false negative.

Right! I'm wrong. the example is not eligible for false positive use case 💦  So I'll close this PR because I'd like to observe feedback after the cop will be released as a pending cop. Thank you always!

PR #8956, opened 2020-10-27
[Fix #8950] Add `IgnoredMethods` and `IgnoredClasses` to `Lint/NumberConversion`

### LINKED ISSUE #8950: Lint/NumberConversion suggests 30.minutes.to_i be turned into Integer(30.minutes, 10), which is invalid
```
app/controllers/concerns/signed.rb:22:19: W: Lint/NumberConversion: Replace unsafe number conversion with number class parsing, instead of using 30.minutes.to_i, use stricter Integer(30.minutes, 10).
      expires_in: 30.minutes.to_i,
                  ^^^^^^^^^^^^^^^
```

```
[1] pry(main)> Integer(30.minutes, 10)
ArgumentError: base specified for non string value
from (pry):1:in `Integer'
```

```
$ [bundle exec] rubocop -V
1.0.0 (using Parser 2.7.2.0, rubocop-ast 1.0.0, running on ruby 2.6.6 x86_64-darwin19)
```

dvandersluis:
Fixes #8950, to an extent. There are going to inherently be a number of false positives possible from this cop since it applies to basically any receiver of `to_i`, etc. other than `Time` and `DateTime`, which were hardcoded. Since this cop is disabled by default we probably don't have a lot of external coverage here (searching github for Lint/NumberConversion returns a lot of inline disables though!)

Instead, the classes to ignore are now configurable, and `IgnoredMethods` was added as well (so the rails duration helpers could be allowed in configuration). This of course isn't going to help when assigned to a variable, because those are checked as well:

```ruby
x = 10.minutes
x.to_i # will be registered
```

Another potential issue is that `Integer(10.minutes)` will work but `Integer(10.minutes, 10)`, which is the default suggestion/correction, does not work when the first argument is not a string. Should I change that as well to only add the base if the first arg is a known string?

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_add_ignoredmethods_and_ignoredclasses_to.md
+++ b/changelog/change_add_ignoredmethods_and_ignoredclasses_to.md
@@ -1,0 +1,1 @@
+* [#8950](https://github.com/rubocop-hq/rubocop/issues/8950): Add `IgnoredMethods` and `IgnoredClasses` to `Lint/NumberConversion`. ([@dvandersluis][])
--- a/spec/rubocop/cop/lint/number_conversion_spec.rb
+++ b/spec/rubocop/cop/lint/number_conversion_spec.rb
@@ -1,10 +1,6 @@
 # frozen_string_literal: true
 
-RSpec.describe RuboCop::Cop::Lint::NumberConversion do
-  subject(:cop) { described_class.new(config) }
-
-  let(:config) { RuboCop::Config.new }
-
+RSpec.describe RuboCop::Cop::Lint::NumberConversion, :config do
   context 'registers an offense' do
     it 'when using `#to_i`' do
       expect_offense(<<~RUBY)
@@ -120,6 +116,16 @@ RSpec.describe RuboCop::Cop::Lint::NumberConversion do
       RUBY
     end
 
+    it 'when `#to_i` called without a receiver' do
+      expect_no_offenses(<<~RUBY)
+        to_i
+      RUBY
+    end
+  end
+
+  context 'IgnoredClasses' do
+    let(:cop_config) { { 'IgnoredClasses' => %w[Time DateTime] } }
+
     it 'when using Time' do
       expect_no_offenses(<<~RUBY)
         Time.now.to_i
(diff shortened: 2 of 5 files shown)


spec/rubocop/cop/lint/number_conversion_spec.rb
```diff
@@ -149,10 +155,21 @@
...
+    let(:cop_config) { { 'IgnoredMethods' => %w[minutes] } }
+
+    it 'does not register an offense for an ignored method' do
       expect_no_offenses(<<~RUBY)
-        to_i
+        10.minutes.to_i
+      RUBY
+    end
+
+    it 'does not register an offense for other methods' do
```
koic, line 169:
```suggestion
    it 'registers an offense for other methods' do
```

koic:
I think this approach is fine. For #8950, default.yml of rubocop-rails can the following settings:

```yaml
Lint/NumberConversion:
  IgnoredMethods:
    - seconds
    - second
    - hours
    - hour
    (snip)
```

And the following methods can be added.
https://github.com/rails/rails/blob/v6.0.3.4/activesupport/lib/active_support/core_ext/numeric/time.rb

changelog/change_add_ignoredmethods_and_ignoredclasses_to.md
```diff
@@ -0,0 +1 @@
+* [#8950](https://github.com/rubocop-hq/rubocop/pull/8950): Add `IgnoredMethods` and `IgnoredClasses` to `Lint/NumberConversion`. ([@dvandersluis][])
```
koic, line 1:
```suggestion
* [#8950](https://github.com/rubocop-hq/rubocop/issues/8950): Add `IgnoredMethods` and `IgnoredClasses` to `Lint/NumberConversion`. ([@dvandersluis][])
```

dvandersluis:
Oh nice great suggestion @koic, I’ll make a PR to rubocop-rails tomorrow!

bbatsov:
It'd be nice if we updated the cop description with explanation of its limitations and why we had to disable it by default.

changelog/change_add_ignoredmethods_and_ignoredclasses_to.md
dvandersluis: I'll fix this but FYI this is coming from the `rake changelog:*` tasks. I assume it's done like this because it's impossible to tell whether the reference is an issue or PR, but github redirects `pull/xxx` to `issues/xxx` if it's actually an issue so I assume that's why @marcandre did it this way.

dvandersluis:
@bbatsov added to the cop description. Should I make it not use a base with `Integer()` if not given a string?

changelog/change_add_ignoredmethods_and_ignoredclasses_to.md
marcandre: Agree it doesn't actually matter. Still, I'll change the script so it writes `issues/nnn` when there is `[Fix #nnn]` in the commit message, and  `pulls/x` if there isn't (assuming that the link will point to the PR to be created)

bbatsov:
I guess that would miss all cases where the first argument is not a literal, so probably we're better off not doing so, at least for now. Thanks for tackling this!

PR #8962, opened 2020-10-28
Use `public_send` instead of `send` when public sendable

koic:
In Ruby, there seems to be a trend to separate `send` into `public_send` and `__send__`.
https://github.com/ruby/ruby/commit/3198e7ab

~RuboCop does not override `send`, so this PR enables `Style/Send` cop to prevent use `send` method.~
~In future, `Style/Send` may be able to move to `Lint/Send`, but for now I think it's good to stay with `Style` department.~
~Also, `Style/Send` remains disabled by default because considering user applications overriding `send` method.~

This PR changes to `public_send` to suppress unnecessary scope to(private, protected) method calls. On the other hand, this PR will not make the change to `__send__` because the change to `__send__` is premature
https://bugs.ruby-lang.org/issues/17100

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/commissioner.rb
+++ b/lib/rubocop/cop/commissioner.rb
@@ -97,7 +97,7 @@ module RuboCop
       def trigger_responding_cops(callback, node)
         @callbacks[callback]&.each do |cop|
           with_cop_error_handling(cop, node) do
-            cop.send(callback, node)
+            cop.public_send(callback, node)
           end
         end
       end
@@ -133,7 +133,7 @@ module RuboCop
         name = node.method_name
         @restricted_map[event][name]&.each do |cop|
           with_cop_error_handling(cop, node) do
-            cop.send(event, node)
+            cop.public_send(event, node)
           end
         end
       end
--- a/lib/rubocop/cop/force.rb
+++ b/lib/rubocop/cop/force.rb
@@ -31,7 +31,7 @@ module RuboCop
         cops.each do |cop|
           next unless cop.respond_to?(method_name)
 
-          cop.send(method_name, *args)
+          cop.public_send(method_name, *args)
         end
       end
 
--- a/lib/rubocop/cop/lint/useless_access_modifier.rb
+++ b/lib/rubocop/cop/lint/useless_access_modifier.rb
@@ -255,7 +255,7 @@ module RuboCop
               PATTERN
             end
 
-            send(matcher_name, child)
+            public_send(matcher_name, child)
           end
         end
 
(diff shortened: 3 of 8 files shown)


marcandre:
I don't mind `public_send`, but I do find `__send__` very ugly and I'd prefer if it was restricted for the very rare case where we do not know the class of the receiver and where it is possible that `send` might be overriden.
I feel that apart from some basic forwarding / callback libraries, there's not need to use `__send__`.

marcandre:
BTW, this "trend" is because the current `Ractor` class defines `send` for communication to a `Ractor`. A few people (me included) [would prefer that `send` *not* be used](https://bugs.ruby-lang.org/issues/17100).

koic:
Thank you for the information. AFAIK, this separation is also beginning to apply in the rails/rails repository. The following is an example:
https://github.com/rails/rails/commit/09939f8657e13ab5eb638593ed7ec9ef959a5f60

In any case, I think it is possible to change to `public_send` to suppress unnecessary scope to (private, protected) method calls. On the other hand, I think it's possible to wait for a while to change to `__send__` (in that case, disable `Style/Send` again). 

So, I will update the PR to not make the change to` __send__` if the change to `__send__` is premature :-)

marcandre:
> Thank you for the information. AFAIK, this separation is also beginning to apply in the rails/rails repository. The following is an example:
> [rails/rails@09939f8](https://github.com/rails/rails/commit/09939f8657e13ab5eb638593ed7ec9ef959a5f60)

This example is a good example where it is important to test that the method is public. Use `public_send` in these cases. For example in `RuboCop`, it's not a bad idea that our `Commissioner` use `public_send`.

> (And the change to `__send__` hasn't started yet in the rails/rails repo 💦 )

Very different concern.

> So, I will update the PR to not make the change to` __send__` if the change to `__send__` is premature :-)

Sounds great to me 👍

koic:
Thanks for your advice! I updated this PR.

bbatsov:
> RuboCop does not override send, so this PR enables Style/Send cop to prevent use send method.

Did you reconsider this part, as I don't see the cop enabled for RuboCop?

bbatsov:
>  I do find __send__ very ugly

I do find everything wrapped in underscores to be very ugly. :D For me such names go against the spirit of Ruby, but thankfully I rarely have to use of them.

koic:
I didn't update the PR description 💦 and I updated it. So, `Style/Send` cop remains disabled.

bbatsov:
Got it. Thanks!

PR #8964, opened 2020-10-28
Renamed `Naming/VariableNumber` cop to `Naming/IdentifierNumber`. Extended this cop to handle symbols and method names

fatkodima:
This PR renames and extends an existing cop to also handle symbols and method names, as in the style guide https://rubystyle.guide/#snake-case-symbols-methods-vars-with-numbers

### DIFF
--- a/changelog/change_variable_number.md
+++ b/changelog/change_variable_number.md
@@ -1,0 +1,1 @@
+* [#8964](https://github.com/rubocop-hq/rubocop/pull/8964): Extend `Naming/VariableNumber` cop to handle method names and symbols. ([@fatkodima][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2411,14 +2411,18 @@ Naming/VariableName:
     - camelCase
 
 Naming/VariableNumber:
-  Description: 'Use the configured style when numbering variables.'
+  Description: 'Use the configured style when numbering symbols, methods and variables.'
+  StyleGuide: '#snake-case-symbols-methods-vars-with-numbers'
   Enabled: true
   VersionAdded: '0.50'
+  VersionChanged: '1.2'
   EnforcedStyle: normalcase
   SupportedStyles:
     - snake_case
     - normalcase
     - non_integer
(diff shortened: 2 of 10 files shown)


koic:
RuboCop has reached 1.0, so we will need to consider providing it to prevent breaking changes.
@rubocop-hq/rubocop-core

lib/rubocop/config_obsoletion.rb
```diff
@@ -31,6 +31,7 @@ class ConfigObsoletion
       'Lint/UnneededSplatExpansion' => 'Lint/RedundantSplatExpansion',
       'Naming/UncommunicativeBlockParamName' => 'Naming/BlockParameterName',
       'Naming/UncommunicativeMethodParamName' => 'Naming/MethodParameterName',
+      'Naming/VariableNumber' => 'Naming/IdentifierNumber',
```
koic, line 34: IMHO, 1.0 stable may require a (new) mechanism that can provide a migration period to users for deprecated cop name until 2.0 is reached.

bbatsov: Yeah, I totally agree. It'd be nice to have a mechanism which allows us to change the name of a cop without breaking configuration. E.g. cops can have a primary name and a deprecated name or something like this.

bbatsov:
Yeah, definitely we won't be renaming any cops unless we introduce a way to do so without breaking the user configs.

bbatsov:
@fatkodima I've been thinking we can make the changes without renaming the cop, as probably the impact for the identifiers will be relatively small. We can mitigate it further with boolean configuration flags for each identifier type.

fatkodima:
Yes, as an option. Updated with your suggestion.

bbatsov:
Method params were represented as locals in the AST and therefore are covered as well even before your changes, right? It'd be nice to reflect this in the cop examples.

fatkodima:
Added examples with arguments to the docs.

bbatsov:
Thanks!

PR #8965, opened 2020-10-29
Using refinements

marcandre:
There are many methods of `Cop::Base` that are not meant to be called by subclasses. Marking them `private` has no impact.

I was just made aware of [how refinements can be used to restrict access](https://bugs.ruby-lang.org/issues/17288#note-9). Using this we can completely hide these methods and insure that no cop is written using them (not realizing they are not meant to be called).

This is just a very rough POC for the time being, I'm not sure when I'll have time to really work on this.

There's an issue with stubs, I'm confident this can be resolved and be better.

A quick test revealed no performance hit.

Another place where it could be nice to use refinements is in the extensions (`ext/`), like my previous changes on `regexp_parser`'s Nodes.

I'm quite excited to realize that I've been wrongly neglecting refinements until now.

### DIFF
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative 'base/internals'
+
 module RuboCop
   module Cop
     # A scaffold for concrete cops.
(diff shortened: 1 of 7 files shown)


marcandre:
Note: I've not checked JRuby though, but read that refinements support is supposed to be good now 🤞

bbatsov:
@marcandre Do you plan to revisit this or can we close it? I'm trying to get the list of active PRs to under 25, so they can fit on a single page. :-)

marcandre:
Yeah, let's close this, there's not a real need for it.

PR #8967, opened 2020-10-29
Add new `Style/NegatedIfElseCondition` cop

fatkodima:
This cop checks for uses of `if-else` and ternary operators with a negated condition which can be simplified by inverting condition and swapping branches.

```ruby
# bad
if !x
  do_something
else
  do_something_else
end

# good
if x
  do_something_else
else
  do_something
end
      
# bad
!x ? do_something : do_something_else

# good
x ? do_something_else : do_something

# bad 
if x != y
  do_something
else
  do_something_else
end

# good
if x == y
  do_something_else
else
  do_something
end
```

I'm sure, there is a better name for this cop.


lib/rubocop/cop/style/inverted_if_else.rb
```diff
@@ -0,0 +1,98 @@
...
+      #   # bad
+      #   !x ? do_something : do_something_else
+      #
+      #   # good
+      #   x ? do_something_else : do_something
+      #
+      class InvertedIfElse < Base
+        extend AutoCorrector
+
+        MSG = 'Invert the negative condition and swap branches.'
```
bbatsov, line 33: negative -> negated

and what the conditional's branches (or something like this)

bbatsov:
Nicely done! I only had a tiny bit of feedback about the message. I'm also thinking a better name for the cop might be "NegatedIfElseCondition".

bbatsov:
Thanks!

PR #8968, opened 2020-10-29
[Fix #8966] Fix spaces in empty parens.

### LINKED ISSUE #8966: Layout/SpaceInsideParens not ignoring empty parentheses.
## Expected behavior

I have the following configuration: 

```yml
Layout/SpaceInsideParens:
  EnforcedStyle: space
```

This should enforce spaces inside parentheses but enforce _no_ spaces in empty parentheses, as per the docs: 

```
# bad
f(3)
g = (a + 3)
y( )

# good
f( 3 )
g = ( a + 3 )
y()
```

## Actual behavior

Have been using this fine on 0.61.1 but after upgrading to 1.1.0, we're now seeing offences for empty parentheses, like this: 

```
 C: Layout/SpaceInsideParens: No space inside parentheses detected.
      subject{ super().to_s }
                     ^
```

1. Set configuration to enforce spaces in parentheses.

2. Use no space in an empty parentheses. 

3. Run rubocop.

## RuboCop version

```
$ rubocop -V
1.1.0 (using Parser 2.7.2.0, rubocop-ast 1.1.0, running on ruby 2.6.5 x86_64-darwin19)
  - rubocop-performance 1.8.1
  - rubocop-rails 2.8.1
```

joshuapinter:
According to the docs, empty parens should always have no spaces. However, when `spaces` was set as the `style` for `SpaceInsideParens`, the Cop would register an offence if there was not a space inside the empty parens.

Empty parens should not have spaces.

This commit fixes `SpaceInsideParens` Cop when the `style` is `spaces` so that it doesn't require a space inside the parens. Moreover, it registers an offence when there _is_ a space inside empty parens.

Additionally, a test was added for the default `style` to ensure there are no spaces inside empty parens.

And finally, the docs were updated to show that spaces are not permitted with _either_ `style`.

Note, had to extract out `process_with_space_style` to avoid method length Cop offence.

Fixes https://github.com/rubocop-hq/rubocop/issues/8966.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/space_inside_parens.rb
+++ b/lib/rubocop/cop/layout/space_inside_parens.rb
@@ -11,10 +11,12 @@ module RuboCop
       #   # bad
       #   f( 3)
       #   g = (a + 3 )
+      #   f( )
       #
       #   # good
       #   f(3)
       #   g = (a + 3)
+      #   f()
       #
       # @example EnforcedStyle: space
       #   # The `space` style enforces that parentheses have a space at the
@@ -44,11 +46,7 @@ module RuboCop
           @processed_source = processed_source
 
           if style == :space
-            each_missing_space(processed_source.tokens) do |range|
-              add_offense(range, message: MSG_SPACE) do |corrector|
-                corrector.insert_before(range, ' ')
-              end
-            end
+            process_with_space_style(processed_source)
           else
             each_extraneous_space(processed_source.tokens) do |range|
               add_offense(range) do |corrector|
@@ -59,6 +57,21 @@ module RuboCop
         end
 
         private
+
+        def process_with_space_style(processed_source)
+          processed_source.tokens.each_cons(2) do |token1, token2|
+            each_extraneous_space_in_empty_parens(token1, token2) do |range|
+              add_offense(range) do |corrector|
+                corrector.remove(range)
+              end
+            end
+            each_missing_space(token1, token2) do |range|
+              add_offense(range, message: MSG_SPACE) do |corrector|
+                corrector.insert_before(range, ' ')
+              end
+            end
+          end
+        end
 
         def each_extraneous_space(tokens)
           tokens.each_cons(2) do |token1, token2|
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,13 @@
 
 * [#8912](https://github.com/rubocop-hq/rubocop/pull/8912): Fix `Layout/ElseAlignment` for `rescue/else/ensure` inside `do/end` blocks with assignment. ([@miry][])
 * [#8971](https://github.com/rubocop-hq/rubocop/issues/8971): Fix a false alarm for `# rubocop:disable Lint/EmptyBlock` inline comment with `Lint/RedundantCopDisableDirective`. ([@koic][])
+### New features
+
+### Bug fixes
+
+* [#8966](https://github.com/rubocop-hq/rubocop/issues/8966): Fix `Layout/SpaceInsideParens` to enforce no spaces in empty parens for all styles. ([@joshuapinter][])
```
bbatsov, line 13: Might be better ot add your changelog entry under the new `changelog` folder, to avoid having to deal with changelog conflicts.

joshuapinter: Oh, this must be _very_ new. I'll move it now...

joshuapinter: Done! Back over to you.

lib/rubocop/cop/layout/space_inside_parens.rb
```diff
@@ -60,6 +58,19 @@ def on_new_investigation
 
         private
 
+        def process_with_space_style(processed_source)
+          each_missing_space(processed_source.tokens) do |range|
+            add_offense(range, message: MSG_SPACE) do |corrector|
+              corrector.insert_before(range, ' ')
+            end
+          end
+          each_extraneous_space_in_empty_parens(processed_source.tokens) do |range|
```
bbatsov, line 67: What's the performance impact of that second pass over the tokens? I'm wondering if we can't combine the two passes for the sake of efficiency.

joshuapinter: Good question. I kept them separate because the offences for empty parens is quite different than the typical cop. I'll run some benchmarks but if you're leaning towards combining them, that might be the best approach for clarity, regardless of the benchmarks.

joshuapinter: Okay, refactored into a single pass. Benchmark was nearly negligible but it's certainly much cleaner and more efficient now. Take a look when you get a sec.

bbatsov:
Thanks!

joshuapinter:
👍

PR #8969, opened 2020-10-30
Add pre-commit support.

adithyabsk:
Drafts a pre-commit support file based on discussion from https://github.com/rubocop-hq/rubocop/issues/7827

-----------------

  * Can be fixed prior to merge
* [x] Added tests.
  * No relevant tests
  * No user facing changes

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,8 @@
+-   id: rubocop
+    name: rubocop
+    entry: rubocop
+    language: ruby
+    'types': [ruby]
+    args: []
+    require_serial: false
+    additional_dependencies: []


koic:
Thank you for opening the PR (and related issue). I think it's interesting as a means. However, I use a lot of (temporary) commits in local development before pushing and I don't run rubocop every time commits. I think it can be set in users layer, not (officially) provided in this repository. Thank you.

adithyabsk:
@koic thanks for the reply though I think you may misunderstand how pre-commit works. The software is actually quite flexible and I in fact use it only to [validate my code prior to pushing](https://pre-commit.com/#pre-commit-during-push).

Additionally, this allows for linting steps to be managed by a single service and allows for the bundling of other convenience steps such [EOF fixing, checking yaml config syntax, and removing trailing whitespace](https://pre-commit.com/hooks.html).

If you do not see this as an ease of use integration for rubocop users, I can understand.[ For my own purposes](https://github.com/adithyabsk/mirrors-rubocop), I have in fact set up a "mirror" repository that integrates with pre-commit that automatically updates when a new version of rubocop is released. If this is the direction you'd rather proceed, I think it might be beneficial to mention this ease of use integration in the docs. (I can open a separate PR)

But I'd much prefer that this functionality be integrated upstream.

adithyabsk:
@koic just thought I'd follow up here.

jdufresne:
I would find this change useful. Pre-commit hooks help ensure code meets a standard before it gets to CI or review. It saves developer time.

@koic It is up to each user whether or not they wish to install the hook. One must first do:

```
$ pre-commit install
```

Before the user will see the effect of the pre-commit hook. So in your situation, you'd be welcome to simply not install the hook. You can always manually run hooks too to ensure linting passes the same as your teammates:

```
$ pre-commit run --all-files
```

Another option for you is the git argument `-n` or `--no-verify` option which will, when provided, skip commit hooks:

```
$ git commit -n -m 'WIP'
```

I really like pre-commit hooks and use all the techniques I've listed above.

bbatsov:
Well, I don't really understand the purpose of this hook myself, and I doubt that anyone will unless they know what to look for and what's it supposed to do. We can't just be adding configuration files with no explanations. 

To be clear - I get the purpose and the benefits of pre-commit hooks, but I don't quite get what this one does (I assume it just runs RuboCop on Ruby files) and I'm not quite sure how people are supposed to use it. This can easily be in the docs instead of committed as a file.

jdufresne:
First, thanks for the response and getting back to this. I appreciate the time.

I think there may be some confusion about what this change does so let me back up and explain a bit.

This _does not_ install a git pre-commit for the rubocop project itself. It _does not_ interfere or change the workflow of rubocop developers.

What this does is add an entry point for _rubocop users_ to use the [pre-commit](https://pre-commit.com/) framework with their project. The entry point allows users to add a rubocop pre-commit hook with minimal configuration and installation. Without this entry point, additional configuration is required, pushing the burden onto every Ruby project that would like this feature.

The pre-commit framework allows users to add a configuration file to their project to add common, repetitive git pre-commits with a minimal amount of configuration. This removes the need for users of Ruby rubocop, JavaScript prettier, or Python black to configure these tools manually.

Here is the framework's introduction:

> Git hook scripts are useful for identifying simple issues before submission to code review. We run our hooks on every commit to automatically point out issues in code such as missing semicolons, trailing whitespace, and debug statements. By pointing these issues out before code review, this allows a code reviewer to focus on the architecture of a change while not wasting time with trivial style nitpicks.
>
> As we created more libraries and projects we recognized that sharing our pre-commit hooks across projects is painful. We copied and pasted unwieldy bash scripts from project to project and had to manually change the hooks to work for different project structures.
>
> We believe that you should always use the best industry standard linters. Some of the best linters are written in languages that you do not use in your project or have installed on your machine. For example scss-lint is a linter for SCSS written in Ruby. If you’re writing a project in node you should be able to use scss-lint as a pre-commit hook without adding a Gemfile to your project or understanding how to get scss-lint installed.
>
> We built pre-commit to solve our hook issues. It is a multi-language package manager for pre-commit hooks. You specify a list of hooks you want and pre-commit manages the installation and execution of any hook written in any language before every commit. pre-commit is specifically designed to not require root access. If one of your developers doesn’t have node installed but modifies a JavaScript file, pre-commit automatically handles downloading and building node to run eslint without root.

This PR makes it simple for users of both rubocop and pre-commit to make them work together. With this change, I can add the following file to my project:


```yml
# .pre-commit-config.yaml
repos:
  - repo: https://github.com/rubocop-hq/rubocop.git
    rev: v1.7.0
    hooks:
      - id: rubocop
```

And after that, every commit to _my project_ will run rubocop on a git commit.

Notice the file added by this PR is `.pre-commit-hooks.yaml` and not `.pre-commit-config.yaml`. `.pre-commit-hooks.yaml` is for tools to provide an entry point and `.pre-commit-hooks.yaml` is for projects to consume those entry points.

So I just want to say again, this does nothing to change the rubocop developers workflow. This is only about providing access to pre-commit users.

bbatsov:
Thanks for the detailed explanations. This makes sense to me. I'll reopen the PR, but for it to get merged we need the same level of details added to the docs here https://docs.rubocop.org/rubocop/1.7/integration_with_other_tools.html

jdufresne:
@adithyabsk Let me know if you'd like a hand finishing this PR. I'm happy finish the doc changes.

jdufresne:
It seems this PR has stalled, so I've continued the work in #9385. Thanks @adithyabsk for the initial work!

PR #8970, opened 2020-10-30
[Fix #8925] Add `--display-time` option for outputting time.

### LINKED ISSUE #8925: Option for displaying elapsed time without using `--debug`.

When you run `rubocop` it doesn't show the elapsed time. However, when you run `rubocop --debug` it outputs `Finished in 3.1285830000415444 seconds` at the end, but it also shows you the path of every file it is scanning.

I'd like to get the elapsed time without getting all the debug information. We use this in our CI to routinely monitor the performance to see if there are bottlenecks, etc.

We have tried using `time rubocop` and it works, but not as cleanly and varies from system to system (e.g. macOS to Ubuntu to CodeShip servers).

If you are against outputting this by default, perhaps a simple option like `--time` or something that will output the elapsed time.

As mentioned above,

> We have tried using `time rubocop` and it works, but not as cleanly and varies from system to system (e.g. macOS to Ubuntu to CodeShip servers).

## Additional context

I'm happy to PR this if it's something you think is valuable and we decide on the details.

Thanks! I love this library!

joshuapinter:
This shows the elapsed time of the Rubocop command without needing to show all of the output of `--debug`.

### Before

```bash
$ rubocop
...
865 files inspected, 218 offenses detected, 148 offenses auto-correctable
$
```

### After

```bash
$ rubocop --display-time
...
865 files inspected, 218 offenses detected, 148 offenses auto-correctable
Finished in 16.146678999997675 seconds
$
```

Fixes #8925.

-----------------

* [x] Added tests.
  - I tried testing this through the CLI specs but because it occurs outside of the CLI - essentially wrapping the CLI command itself - it wasn't working. Any thoughts or tips on testing this?

### DIFF
--- a/changelog/new_display_time_command_line_option.md
+++ b/changelog/new_display_time_command_line_option.md
@@ -1,0 +1,1 @@
+* [#8925](https://github.com/rubocop-hq/rubocop/issues/8925): Add `--display-time` option for displaying elapsed time of `rubocop` command. ([@joshuapinter][])
--- a/exe/rubocop
+++ b/exe/rubocop
@@ -13,5 +13,5 @@ time = Benchmark.realtime do
   result = cli.run
 end
 
-puts "Finished in #{time} seconds" if cli.options[:debug]
+puts "Finished in #{time} seconds" if cli.options[:debug] || cli.options[:display_time]
 exit result
--- a/lib/rubocop/options.rb
+++ b/lib/rubocop/options.rb
@@ -145,6 +145,7 @@ module RuboCop
         end
       end
 
+      option(opts, '--display-time')
       option(opts, '--display-only-failed')
     end
 
@@ -469,6 +470,7 @@ module RuboCop
                                          'if no format is specified.'],
       fail_level:                       ['Minimum severity (A/R/C/W/E/F) for exit',
                                          'with error code.'],
+      display_time:                     'Display elapsed time in seconds.',
       display_only_failed:              ['Only output offense messages. Omit passing',
                                          'cops. Only valid for --format junit.'],
       display_only_fail_level_offenses:
--- a/spec/rubocop/cli/cli_options_spec.rb
+++ b/spec/rubocop/cli/cli_options_spec.rb
@@ -789,6 +789,26 @@ RSpec.describe RuboCop::CLI, :isolated_environment do
     end
   end
 
+  describe '--display-time' do
+    before do
+      create_file('example1.rb', '')
+    end
+
+    regex = /Finished in [0-9]*\.[0-9]* seconds/
+
+    context 'without --display-time' do
+      it 'does not display elapsed time in seconds' do
+        expect(`rubocop example1.rb`).not_to match(regex)
+      end
+    end
+
+    context 'with --display-time' do
+      it 'displays elapsed time in seconds' do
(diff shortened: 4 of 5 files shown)


bbatsov:
I'm fine with the proposed change in general, but it seems to me an option name like `--display-time` would probably be better. We can also skip a shorthand for it, as we don't have that many of those available, so I'd be careful with their assignment.

joshuapinter:
@bbatsov Yup, was thinking something like `--display-elapsed-time` but `--display-time` is nice and succinct while still being descriptive.

Quick Question: For testing, I tried testing this through the CLI specs but because it occurs outside of the CLI - essentially wrapping the CLI command itself - it wasn't working. Any thoughts or tips on testing this?

Thanks!

joshuapinter:
Rebased and updated to `--display-time` now.

CHANGELOG.md
```diff
@@ -6,6 +6,15 @@
 
 * [#8912](https://github.com/rubocop-hq/rubocop/pull/8912): Fix `Layout/ElseAlignment` for `rescue/else/ensure` inside `do/end` blocks with assignment. ([@miry][])
 * [#8971](https://github.com/rubocop-hq/rubocop/issues/8971): Fix a false alarm for `# rubocop:disable Lint/EmptyBlock` inline comment with `Lint/RedundantCopDisableDirective`. ([@koic][])
+### New features
```
bbatsov, line 9: The changelog seems messed up. I think you didn't rebase properly.

joshuapinter: Yes, yes it does. Sorry, will look at that right meow.

joshuapinter: Alright, that looks a little better.

bbatsov:
> Quick Question: For testing, I tried testing this through the CLI specs but because it occurs outside of the CLI - essentially wrapping the CLI command itself - it wasn't working. Any thoughts or tips on testing this?

Can't you just emulate the `--debug` tests?

joshuapinter:
> > Quick Question: For testing, I tried testing this through the CLI specs but because it occurs outside of the CLI - essentially wrapping the CLI command itself - it wasn't working. Any thoughts or tips on testing this?
> 
> Can't you just emulate the `--debug` tests?

Good idea but the `--debug` tests seem to be contained inside `Cli` and not testing `rubocop/exe/rubocop`, which is where the elapsed time is displayed.

Regardless, I used it as a bit of a base and added tests for `--display-time` with and without the command line option, regex'ing the `STDOUT` from the command.

Ready for your review again, and thanks for your help.

bbatsov:
> Good idea but the --debug tests seem to be contained inside Cli and not testing rubocop/exe/rubocop, which is where the elapsed time is displayed.

My bad. I didn't get what you mean initially. I guess it'd be best to move the timing logic to the `CLI` class. I guess we were just lazy in the past. :-)

joshuapinter:
> > Good idea but the --debug tests seem to be contained inside Cli and not testing rubocop/exe/rubocop, which is where the elapsed time is displayed.
> 
> My bad. I didn't get what you mean initially. I guess it'd be best to move the timing logic to the `CLI` class. I guess we were just lazy in the past. :-)

:+1:  All good. `CLI` was my first guess too.

Issue #8564, opened 2020-08-18
Redefine what counts as a branch on ABC Size


jafuentest:
I was thinking if it would be possible to not count getter methods as a branch. I know this has been asked before (can't find the issue now, but I de remember reading something about it) and I'm just wondering if it may be possible with a plugin instead of a configuration setting.

The reason for it is pretty simple, take this method for example

```ruby
def search
  @posts = model.active.visible_by(current_user)
    .search(params[:q], skip: @skip)

  @posts = model.some_process(@posts, current_user)
  @posts = model.another_process(@posts, current_user)

  render 'pages/search/page', locals: { model: model }
end
```

It's ABC would be `<3, 15, 0> = 15.3`

Now, if instead of calling model and current_user every time, you assign their values to a variable you get:

```ruby
def search
  the_model = model
  user = current_user
  @posts = the_model.active.visible_by(user)
    .search(params[:q], skip: @skip)

  @posts = the_model.some_process(@posts, user)
  @posts = the_model.another_process(@posts, user)

  render 'pages/search/page', locals: { model: the_model }
end
```

It's ABC would be `<5, 10, 0> = 11.18`

So the reason this bothers me is that the last method is considered the least complex one, but in reality it takes more time for someone to read through it.

Branches are defined as "an explicit forward program branch out of scope", I would say getter methods in ruby are not really "explicit" as they pretty much look like just any variable. But I'm aware that's highly a matter of personal opinion.

So wrapping up, is there a way to not count any getter method (defined for this purpose as any method that receives no arguments) as branch while calculating ABC? I had a look at `AbcSizeCalculator` class but I can't figure out where the `node.type` comes from in the `branch?(node)` method

marcandre:
It is indeed tricky. As you realize, it's impossible to know if, in your example, `model` is just a `attr_reader` or a method that makes a web request synchronously, thanks to the dynamic nature of Ruby.

The `AbcSize` metric was tweaked extensively recently (#8037) and also we reduced some repeated conditional paths (#8408). I didn't add a setting for that last tweak as it would be plausible for Ruby to actually optimize this internally. In this particular case, Ruby can not optimize this (even if it is an `attr_reader`, as the instance variable `@model` can change at any time). So I don't think we can justify not counting it by default.

That said, we could discuss a setting to discount repeated `send` without arguments. In this example:

```
def foo
  model.do_something  # branch +2
[...]

PR #8986, opened 2020-11-01
Allow Metrics/AbcSize skip wide array/hash literals with nested method calls

razum2um:
We had code pieces like 

```
def builder_meth
  ActiveRecordModel.new(
    attr1: x.attr1,
    attr2: x.attr2,
    attr3: y.attr3,
    attr4: y.attr4
  )
end
```

The current defaults for `Metrics/AbcSize` grows at double speed along with column count. 
Now some alternative approaches:

- You may advise rewriting using `merge` + extract methods for getting `attr1` / `attr2` together,
   but really, the code looks very clean, there's nothing more **simple** than a hash declaration, isn't it?

- You may advise increasing AbcSize's limit,
   but for a non-declarative, **imperative** code is really good, so we'd rather not

I think this is very similar to what they done in https://github.com/rubocop-hq/rubocop/pull/8159

p.s. but still, I think even declarations should contribute at least something to branches, if not empty, hence added `@branches += 1` for every `folded` thing - but this is questionable, maybe it makes sense not completely `-=` but `-= (.. / 2).ceil` - what do you think?

### DIFF
--- a/changelog/new_add_count_as_one_for_abc_size.md
+++ b/changelog/new_add_count_as_one_for_abc_size.md
@@ -1,0 +1,1 @@
+* [#8986](https://github.com/rubocop-hq/rubocop/pull/8986): Add new `CountAsOne` option for `Metric/AbcSize` cop. ([@razum2um][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2072,6 +2072,7 @@ Metrics/AbcSize:
   # The ABC size is a calculated magnitude, so this number can be an Integer or
   # a Float.
   IgnoredMethods: []
+  CountAsOne: []
   Max: 17
 
 Metrics/BlockLength:
--- a/docs/modules/ROOT/pages/cops_metrics.adoc
+++ b/docs/modules/ROOT/pages/cops_metrics.adoc
@@ -17,12 +17,40 @@ configured maximum. The ABC size is based on assignments, branches
 (method calls), and conditions. See http://c2.com/cgi/wiki?AbcMetric
 and https://en.wikipedia.org/wiki/ABC_Software_Metric.
 
+You can set literals you want to fold with `CountAsOne`.
+Available are: 'array', 'hash'. Each literal
+will be counted as one branch regardless of its values branch counts,
+but only if any value has branches (e.g. is/includes a method call)
+
+=== Examples
+
+==== CountAsOne: ['array', 'hash']
+
+[source,ruby]
+----
+def m
+  Klass.new([
+    private_m.foo,
+    private_m.bar
+  ]) # 1 (construstor) + 1 (array includes calls)
+
+  Klass.new(
+    key: private_m.foo,
+    bar: private_m.bar
+  ) # 1 (construstor) + 1 (hash values includes calls)
+end
+----
+
 === Configurable attributes
 
 |===
 | Name | Default value | Configurable values
 
 | IgnoredMethods
+| `[]`
+| Array
+
+| CountAsOne
 | `[]`
 | Array
 
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -358,6 +358,7 @@ require_relative 'rubocop/cop/lint/void'
 require_relative 'rubocop/cop/metrics/utils/iterating_block'
 require_relative 'rubocop/cop/metrics/cyclomatic_complexity'
 # relies on cyclomatic_complexity
+require_relative 'rubocop/cop/metrics/abc_size_result'
 require_relative 'rubocop/cop/metrics/utils/abc_size_calculator'
 require_relative 'rubocop/cop/metrics/utils/code_length_calculator'
 require_relative 'rubocop/cop/metrics/abc_size'
--- a/lib/rubocop/cop/metrics/abc_size.rb
+++ b/lib/rubocop/cop/metrics/abc_size.rb
@@ -7,6 +7,26 @@ module RuboCop
       # configured maximum. The ABC size is based on assignments, branches
       # (method calls), and conditions. See http://c2.com/cgi/wiki?AbcMetric
       # and https://en.wikipedia.org/wiki/ABC_Software_Metric.
+      #
+      # You can set literals you want to fold with `CountAsOne`.
+      # Available are: 'array', 'hash'. Each literal
+      # will be counted as one branch regardless of its values branch counts,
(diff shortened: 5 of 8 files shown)


marcandre:
Thanks for raising your concern.

> The current defaults for Metrics/AbcSize grows at double speed

You could store `y` in a local variable and it would actually speed up (moderately) your method and your AbcSize should grown at single speed in the branch condition.

As for the PR, I don't think that it really addresses the actual issue.The issue I see is that multiple calls to `y` could probably be counted as one (i.e. "assume repeated branch calls are simple `attr_reader` or cached methods", at least first depth) and/or not count at all some method calls that can be assumed to be builtin (e.g. `[]`).

razum2um:
@marcandre thanks for your thoughts, I've edited example multiple times to get most clean one, but apparently failed :)

First of all, `x` & `y` are _mostly_ already lazy cached `attr_reader`s and a local variable for this looks awkward, but there are cases when some values come from arguments and the key's count is just bigger. The main idea here: tables can be legitimately wide and all keys should be side-by-side, no `{}#merge`

I still think this PR has very tight relation to `CountAsOne` in MethodSize. Ofc, there that is a pure formatting issue and inline hash literal is total equivalent (except line size).

But here I would like to distinguish the declarative initialisation from imperative one.
Compare this example, which I would be glad to continue restricting by `AbcSize` like today (branch=16, but also due setters):

```
Klass.new do |obj|
  obj.a = a.a
  obj.b = b.b
  obj.c = c.c
  obj.d = d.d
  obj.e = e.e
end
```

Declarative initialisation however carries less cognitive load (imo).
Strictly speaking, this already had less AbcSize (but it's only so because there are no setters):

```
Klass.new(
  a: a.a
  b: b.b
  c: c.c
  d: d.d
  e: e.e
)
```

As I've said I'm not sure by totally squashing the whole hash to 1 branch either.
Your idea about deduplicated counting calls to methods with `self` receiver looks interesting.
Maybe we should strip only top-level `:send`s only when receiver is `self` (nil) from hash values?

marcandre:
It's still not clear to me how "where" it's used (building a hash/array vs other things) is that important.

If you are calling 10 different methods (columns) on your object, it seems to me that there is no possible justification for them not to count as 10 branch calls, unless maybe if you explicitly put them in an IgnoreList (because you know they are simple attributes, say). To me the fact that you are putting them in an array / hash makes no difference to the size of the method.

See also discussion in #8564 (has nothing to do with hash/array).

razum2um:
> there is no possible justification for them not to count as 10 branch calls
> because you know they are simple attributes

This hits the point. One cannot add all instance methods to `IgnoredMethods` and I'd not to use `@ivar` because need laziness. Regarding local variable I agree with @jafuentest from linked PR (thanks for this) that readability `getter/lvar` outweighs perf cost, btw double getter access barely differs with lvar even after 100M iterations: https://gist.github.com/razum2um/133c41a2b9d69db31a469ed8989421bb)

I think we're searching how to tell rubocop counting getters (lazy attributes) differently, and in the linked PR you're right, that ruby cannot allow us to declare that zero-parameter (even memoized) getter calls are idempotent. But this is a **very** strong rule of thumb across whole ecosystem that for a single object they should (agree?). I think @matz gave us possibility to omit parens exactly having this in mind.

I hear your idea about "discount repeated send without arguments" - but we cannot just skip (or even discount) *all* getter costs across whole codebase. I'm talking about **object integrity**.

If you're literally defining a hash/array and fetching values into 2 keys via a single getter - you definitely want to fetch data from a _single_ object (by convention). Local vars will be explicit, but again, never saw such explicitness during 10 yrs with ruby. That's why "how "where" it's used is that important"

bbatsov:
For the record - I'm fine with the proposed PR, mostly because of:

> I still think this PR has very tight relation to CountAsOne in MethodSize. Ofc, there that is a pure formatting issue and inline hash literal is total equivalent (except line size).

Which is true. I was initially opposed to twisting straighforward metrics like `MethodSize` with special cases, but there was a lot of demand for this from the end users, and I guess the arguments made for it are relevant here as well.

marcandre:
I'd like to repeat the formatting vs content idea. The fact that one couldn't write a method in a "vertical style" but that the same content with fewer `\n` was acceptable *is* an issue imo. I see `MethodLength` as a poor man's equivalent of the other complexity metrics. It basically counts the number of `\n` in a method, which is not super releated to it's complexity.

OTOH, `AbcSize` aims to be more about how much "content" there is. How long the variable names are, on how many lines this is written, these things don't count. What counts is how many method calls, conditions and assignments there are.

With this setting, my feat is that it will be very tempting to "game" the AbcSize metric, even if it will render the code less readable.

```
def foo # => AbcSize too high
  # ... very complex expression
end

# easy solution!!!

def foo # => AbcSize is ok!
  [
     # ... very complex expression (same as before)
  ].first
end
```

Note that after  #9053, the original observation that each hash key was growing as double would no longer be true.
I should follow up on #9053 though

razum2um:
@marcandre that was such a brilliant `foo` workaround example 😂 that I would never welcome in my codebase. https://github.com/rubocop-hq/rubocop/pull/9053 is way better, go ahead 🚀

marcandre:
I thought of an alternative solution to allow (bigger) methods that serialize data (in case there is interest).

Something like `GroupHashValues` that would apply a different way to count hash values.

```ruby
def foo(x, y)
  {
    attr1: x.foo.bar.baz,  # size = 3
    attr2: y.convert(42).to_s, # size = 2
    attr3: y.other # size = 1
  }
end
```
Currently, the branches are simply added (3 + 2 + 1 = 6).
Maybe we should consider another norm, like the [euclidian norm](https://en.wikipedia.org/wiki/Norm_(mathematics)#Euclidean_norm): `Math.sqrt(3**2 + 2 **2 + 1**2) # => 3.7`

For a more typical hash that has 8 values, each with 2 branches say, instead of 8*2 == 16 we'd get 5.6.

The idea of applying a norm is to maintain that more values, or more complex ones increase the size but not as fast.

I we want this, we should only consider for hash literals; I don't see a case where arrays should be counted this way, and I'd even exclude keyword arguments too.

An alternative norm is the [maximum norm](https://en.wikipedia.org/wiki/Norm_(mathematics)#Maximum_norm_(special_case_of:_infinity_norm,_uniform_norm,_or_supremum_norm)) where only the maximum value would count.

Let me know

PR #9001, opened 2020-11-06
Add new `Lint/EmptyClass` cop

fatkodima:
This cop checks for metaclasses without a body.

```ruby
  # bad
  class Foo
    class << self
    end
  end

  class << obj
  end

  # good
  class Foo
    class << self
      attr_reader :bar
    end
  end

  class << obj
    attr_reader :bar
  end
```


bbatsov:
I'm wondering if this shouldn't be part of a more generic `EmptyClass` cop which can check for empty metaclass and an empty class that doesn't inherit from anything.

fatkodima:
Ha, I was 100% sure that there is already a cop for empty classes 😄 
Then it should be also extended for modules, I think?

Empty modules and metaclasses are always useless, while empty classes are sometimes used and useful (for example, empty classes for some results `class ResultSuccess; end`, etc). 

Maybe there is a better name for the cop then, if we will combine checks for classes, modules and metaclasses?

marcandre:
> Empty modules and metaclasses are always useless,

Metaclasses: yes. Modules: not necessarily:

```ruby
module MyNameSpace
end

require_relative 'foo'
require_relative 'bar'
# ...

# foo.rb:
class MyNameSpace::Foo
end

# same with bar
```

bbatsov:
Still, I've never seen a namespace module created empty, but I guess someone might be using them in this way. We can always have a flag for this.

marcandre:
> Still, I've never seen a namespace module created empty

Here's [one I wrote myself](https://github.com/deep-cover/deep-cover/blob/master/core_gem/lib/deep_cover/tools.rb).

I have an issue with cops that flag code that is either *valid* or *innocuous*, especially when rare. [The extremely negative feelings](https://youtu.be/OltCr8AWpWw?t=2886) it might generate can completely outweigh the (potentially negligible) gain. Remember, you've never seen an empty module, so clearly it's not a common mistake that gains much from being flagged and corrected.

fatkodima:
> Modules: not necessarily:

Ah, forgot about this case. I have seen a couple of examples like this.

Thanks for the link to keynote, haven't seen that. It was fun 😄 Personally, I find `Lint` and spacing-like cops the most useful for everyone. `Style`, `Naming`, etc cops can not satisfy everyone, so projects configure them for their specific needs.

The cops I absolutely hate, when enabled, are `Style/GuardClause` and much-much less than previous -`Style/IfUnlessModifier`.

The first of them leads to code like
```ruby
return ... if ...
return ... unless ...
3 more times return ...
do_something
```

Even, forcing me to write something like
```ruby
def method
  return unless condition

  do_something
end
```
drives me mad a little.

fatkodima:
Changed the cop's name to `EmptyClass`. And now this cop checks for classes in addition to metaclasses. Decided against checking for empty modules, at least for now.

bbatsov:
Simple is good. :-) Thanks!

PR #9002, opened 2020-11-06
Update to `rubocop-rspec` 2.0

marcandre:


marcandre:
I'm taking the liberty of disabling `RSpec/StubbedMock`, [see here why](https://github.com/rubocop-hq/rubocop-rspec/pull/226#issuecomment-723337232)

bquorning:
I know that `RSpec/StubbedMock` can be controversial, but I actually the three offenses it points out in the RuboCop spec suite are valid. In all three cases, `expect` can be exchanged for `allow`.

I’ll open a PR and mention you there.

PR #9004, opened 2020-11-07
Remove obsolete gem `SafeYAML` compatibility

marcandre:
AFAIK, there is no longer any reason to use `SafeYAML`.

This is what was causing issues on `ruby-head`, the gem is incompatible with keyword parameters.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,6 @@ gem 'test-queue'
 gem 'yard', '~> 0.9'
 
 group :test do
-  gem 'safe_yaml', require: false
   gem 'webmock', require: false
 end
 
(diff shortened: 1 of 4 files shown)


bbatsov (approved):

aav7fl:
@marcandre Would it have been better to `warn` instead of raising an error?

I understand the importance of letting users know they should remove that dependency if they were using RuboCop in isolation.

But I have another project (Jekyll) that I use to build my site before testing. Jekyll is pulling in `SafeYAML` during the build portion which then breaks my RuboCop tests.

Or should I be isolating my build/test portions and uninstalling gems between steps?

marcandre:
Let's discuss in #9035 👍

PR #9006, opened 2020-11-07
Enable RSpec/StubbedMock

bquorning:
I know this cop can be controversial, and maybe we don't want to *always* enforce its recommended style.

But in all three cases found in our spec suite, the mock (defined by `expect`) is entirely unnecessary - using a stub (with `allow`) is enough. If any of the stubs are changed, other expectations in each example will fail.

-----------------

* [x] Added tests.
* [-] Added an entry (file) to the [changelog folder](https://github.com/rubocop-hq/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

### DIFF
--- a/spec/rubocop/result_cache_spec.rb
+++ b/spec/rubocop/result_cache_spec.rb
@@ -175,7 +175,7 @@ RSpec.describe RuboCop::ResultCache, :isolated_environment do
       context 'when team external_dependency_checksum changes' do
         it 'is invalid' do
           cache.save(offenses)
-          expect(team).to(
+          allow(team).to(
             receive(:external_dependency_checksum).and_return('bar')
           )
           cache2 = described_class.new(
@@ -188,7 +188,7 @@ RSpec.describe RuboCop::ResultCache, :isolated_environment do
       context 'when team external_dependency_checksum is the same' do
         it 'is valid' do
           cache.save(offenses)
-          expect(team).to(
+          allow(team).to(
             receive(:external_dependency_checksum).and_return('foo')
           )
           cache2 = described_class.new(
@@ -329,8 +329,8 @@ RSpec.describe RuboCop::ResultCache, :isolated_environment do
       end
 
       it 'doesn\'t raise an exception' do
-        expect(FileUtils).to receive(:mkdir_p).with(start_with(cache_root))
-                                              .and_raise(error)
+        allow(FileUtils).to receive(:mkdir_p).with(start_with(cache_root))
+                                             .and_raise(error)
         expect { cache.save([]) }.not_to raise_error
         expect($stderr.string).to eq(<<~WARN)
           Couldn't create cache directory. Continuing without cache.


pirj (approved):
Thanks!

pirj:
Some more context: https://github.com/rubocop-hq/rubocop-rspec/pull/226#issuecomment-723494927

marcandre:
I agree they are not *necessary*.

I'd like to hear an argument for them not being *harmful* (or even for not being *useful*)

Here's my argument to their usefullness.

These `expect` act as sanity check that the implementation being tested does indeed call said method as expected. I see no value in removing these sanity checks.

For example, if the caching mechanism is changed to call something else than `external_dependency_checksum` and if this PR is merged in, only one of the two modified specs would fail. With the current `expect` they would both fail.

pirj:
> only one of the two modified specs would fail. With the current expect they would both fail.

Do you really need both of them to fail to see that something's not right?
I understand that if both would fail, it would probably take less time to figure out what is the root cause since the `expect` would fail.

This is what I meant in:
> properly balanced examples


To be fair, there was another valid argument just recently, that [unused stubs would hang](https://github.com/rubocop-hq/rubocop-rspec/issues/1074) in specs while not actually called from the code.


I can think of something like this to address all of those concerns at once:

```ruby
      it 'uses external_dependency_checksum under the hood' do
        cache.save(offenses)
        expect(team).to receive(:external_dependency_checksum)
        cache2 = described_class.new(
          file, team, options, config_store, cache_root
        )
        cache2.valid? # not sure if this line is necessary, and what calls `external_dependency_checksum`, `initialize` or `valid?`
      end

      context 'when team external_dependency_checksum changes' do
        it 'is invalid' do
          cache.save(offenses)
          allow(team).to(
            receive(:external_dependency_checksum).and_return('bar')
          )
          cache2 = described_class.new(
            file, team, options, config_store, cache_root
          )
          expect(cache2.valid?).to eq(false)
        end
      end

      context 'when team external_dependency_checksum is the same' do
        it 'is valid' do
          cache.save(offenses)
#          This in no way makes it clear that "the checksum is the same"
#          allow(team).to(
#            receive(:external_dependency_checksum).and_return('foo')
#          )
          cache2 = described_class.new(
            file, team, options, config_store, cache_root
          )
          expect(cache2.valid?).to eq(true)
        end
      end
```

But pretty sure someone will be dissatisfied with it anyway, not meaning anyone particular :D

I believe this all boils down to "what is more effective: be able to track down the root cause quickly when the implementation of SUT changes and some spec break, or to be able to understand how SUT works by looking at the spec".
I'm personally more biased towards the latter, but don't insist that it's the only acceptable option.

marcandre:
> This in no way makes it clear that "the checksum is the same"

Agreed, would be better with simply `expect(team).to receive(:external_dependency_checksum) # sanity check` (assuming that works, idk).

> But pretty sure someone will be dissatisfied with it anyway

IUC, this is adding a whole test just to isolate the "sanity check" part from the rest, just so it satisfies this "rule". You were right to be pretty sure 😅

> I believe this all boils down to...

To me it boils down to an idealized abstract theory that doesn't jive with pragmatic reality.

This is the same as with `MultipleMemoizedHelpers` and I must admit being still frustrated about that being a default, despite being clearly more problematic and even less fixable.

I will cut the debate short this time, as I would rather spend my time elsewhere. I officially have no objection to whichever way to turn this cop back on.

bbatsov:
I'm also ambivalent towards the proposed change. Feel free to merge it or not, but let's not keep the PR open for an extended period of time.

bquorning:
How about this compromise – I fix the offenses, but don’t enable the cop?

pirj:
Or reverse compromise - we disable the cop for this spec?
From that, we'll see if it helps improving or stands in the way of specs yet to be created and adjust accordingly.

marcandre:
> I'm also ambivalent

It's not clear who the "also" refers to, but since this comment follows mine, I'll underline that I am not ambivalent at all, only that I'm not going to make any further opposition to whatever changes are deemed positive.

bbatsov:
> How about this compromise – I fix the offenses, but don’t enable the cop?

I like this compromise and I don't want this PR to stay open forever. Merged. :-)

PR #9008, opened 2020-11-08
Mark `Style/InfiniteLoop` as unsafe

marcandre:
I noticed this while backporting `Ractor`. In particular `Ractor::ClosedError` and `CloseQueueError` are `StopIteration`

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -3356,12 +3356,14 @@ Style/ImplicitRuntimeError:
   VersionAdded: '0.41'
 
 Style/InfiniteLoop:
-  Description: 'Use Kernel#loop for infinite loops.'
+  Description: >-
+                 Use Kernel#loop for infinite loops.
+                 This cop is unsafe in the body may raise a `StopIteration` exception.
+  Safe: false
   StyleGuide: '#infinite-loop'
   Enabled: true
   VersionAdded: '0.26'
   VersionChanged: '0.61'
-  SafeAutoCorrect: true
 
 Style/InlineComment:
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/style/infinite_loop.rb
```diff
@@ -5,6 +5,10 @@ module Cop
     module Style
       # Use `Kernel#loop` for infinite loops.
       #
+      # This cop is marked as unsafe as if the rule does not necessarily
```
bbatsov, line 8: I think the `if` here is a mistake, as I'm having hard time understanding the sentence. :-)

config/default.yml
```diff
@@ -3356,12 +3356,16 @@ Style/ImplicitRuntimeError:
   VersionAdded: '0.41'
 
 Style/InfiniteLoop:
-  Description: 'Use Kernel#loop for infinite loops.'
+  Description: >-
+                 Use Kernel#loop for infinite loops.
+                 This cop is marked as unsafe as if the rule does not necessarily
```
bbatsov, line 3361: Same here. I generally try to keep those description short, so we don't end up duplicating too much. At some point we'll probably have to start auto-generating them from the cop classes, but that's not a big deal.

marcandre: Shortened!

lib/rubocop/cop/style/infinite_loop.rb
marcandre: Oops, good catch.

PR #9010, opened 2020-11-09
`Metrics/ParameterLists` supports `MaxOptionalParameters` config parameter

fatkodima:
This is an implementation of https://rubystyle.guide/#keyword-arguments-vs-optional-arguments

Not sure if something like `IgnoredMethods` should be added - like for this https://github.com/rubocop-hq/rubocop/compare/master...fatkodima:too_many_default_arguments-cop?expand=1#diff-3b2d7d0e5589b3568a72de6447fa1095b2e9bae437f299ab43196979d0d50c01R36

### DIFF
--- a/changelog/new_add_max_optional_parameters_to_parameter_lists_cop.md
+++ b/changelog/new_add_max_optional_parameters_to_parameter_lists_cop.md
@@ -1,0 +1,1 @@
+* [#9010](https://github.com/rubocop-hq/rubocop/pull/9010): `Metrics/ParameterLists` supports `MaxOptionalParameters` config parameter. ([@fatkodima][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2185,8 +2185,10 @@ Metrics/ParameterLists:
   StyleGuide: '#too-many-params'
   Enabled: true
   VersionAdded: '0.25'
+  VersionChanged: '<<next>>'
   Max: 5
   CountKeywordArgs: true
+  MaxOptionalParameters: 3
 
 Metrics/PerceivedComplexity:
   Description: >-
--- a/lib/rubocop/cop/metrics/parameter_lists.rb
+++ b/lib/rubocop/cop/metrics/parameter_lists.rb
@@ -36,14 +36,43 @@ module RuboCop
       #   # good (assuming Max is 3)
       #   def foo(a, b, c, d: 1)
       #   end
+      #
+      # This cop also checks for the maximum number of optional parameters.
+      # This can be configured using the `MaxOptionalParameters` config option.
+      #
+      # @example MaxOptionalParameters: 3 (default)
+      #   # good
+      #   def foo(a = 1, b = 2, c = 3)
+      #   end
+      #
+      # @example MaxOptionalParameters: 2
+      #   # bad
+      #   def foo(a = 1, b = 2, c = 3)
+      #   end
+      #
       class ParameterLists < Base
         include ConfigurableMax
 
         MSG = 'Avoid parameter lists longer than %<max>d parameters. ' \
               '[%<count>d/%<max>d]'
+        OPTIONAL_PARAMETERS_MSG = 'Method has too many optional parameters. [%<count>d/%<max>d]'
 
         NAMED_KEYWORD_TYPES = %i[kwoptarg kwarg].freeze
         private_constant :NAMED_KEYWORD_TYPES
+
+        def on_def(node)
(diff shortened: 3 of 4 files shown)


bbatsov:
I think the real problem is not a particular number of default arguments, but having any of them in the first place, and mixing them with keyword arguments. I'd just discourage their usage completely, but that's just me. @rubocop-hq/rubocop-core wha t do you think?

marcandre:
There already a cop for too many positional arguments I believe, but the default maximum is super high (5?).

My personal preference:
* public methods: max of 1 positional or rest argument, everything else is keyword argument ("Rails convention")
* private methods: no maximum, but never any default for positional arguments. It's typically to pass some context around, so typically the arguments passed will be clear on the call site.

fatkodima:
My intention with this cop was to avoid ugliness like 
```ruby
def do_something(a = nil, b = nil, c = nil, d = nil)
end

do_something(nil, nil, nil, value)
``` 
and more general
```ruby
def do_something(a = 1, b = 2, c = 3)
end

do_something(1, 2, 4) # you have to repeat default values for "a" and "b" to call with different "c"
```
> but having any of them in the first place, and mixing them with keyword arguments

agreed with mixing

Feel free to close this if is not worth it.

bbatsov:
Btw, I was just reminded that we already have a cop `Style/OptionalArguments` (yeah, I know we messed up its name), and it seems natural to me to just add the limit for optional arguments there.

bbatsov:
As for keyword vs optional params - I'd love for us to have some cop that enforces more or less what @marcandre suggested, but I think that's orthogonal to @fatkodima's current suggestion.

fatkodima:
> and it seems natural to me to just add the limit for optional arguments there.

This is what I wanted 👍 So added that config.

lib/rubocop/cop/style/optional_arguments.rb
```diff
@@ -17,14 +20,30 @@ module Style
...
+      #   # bad
+      #   def foo(a = 1, b = 2, c = 3)
+      #   end
+      #
       class OptionalArguments < Base
+        include ConfigurableMax
+
         MSG = 'Optional arguments should appear at the end ' \
               'of the argument list.'
+        TOO_MANY_MSG = 'Method has too many optional arguments. [%<count>d/%<max>d]'
```
bbatsov, line 39: It should say "parameters". In general you can fix the terminology everywhere except in the cop's name, as we're checking the definition not the call sites.

fatkodima: Yeah, that is correct that those are "parameters"  and not "arguments". But renaming it to "parameters", I think, may introduce some confusion, because this cop name includes "arguments" in its name, `def on_def(node)` - `node` also has `#arguments` method (and looks like `parser` gem does not have a notion of "parameters", but "arguments" everywhere - for calls and definitions). And seems like `rubocop`s own codebase uses "arguments" in most of the places where really it is about "parameters". Personally, I'm using both words interchangeably, and the real meaning depends on the context.
So should I fix the terminology?

bbatsov: Those are fair points, although I'd disagree that RuboCop uses mostly arguments (just going over the names of the cops shows otherwise). Not that big of a deal though, it's just I never use the terms interchangeably and this terminology confuses me. I guess I'm one old grumpy coder. :D

lib/rubocop/cop/style/optional_arguments.rb
```diff
@@ -17,14 +20,30 @@ module Style
...
+      #   # bad
+      #   def foo(a = 1, b = 2, c = 3)
+      #   end
+      #
       class OptionalArguments < Base
-        MSG = 'Optional arguments should appear at the end ' \
+        include ConfigurableMax
+
+        MSG = 'Optional parameters should appear at the end ' \
               'of the argument list.'
```
bbatsov, line 38: argument -> parameter

bbatsov:
Btw, I was thinking that we can also have this check in the other cops `Metrics/ParameterLists`. I'm on the fence which method is a better home for it.

fatkodima:
Forgot about its existence. `Metrics/ParameterLists` seems a better place to me, since the original cop deals with default parameters placement (at the end or not), but the mentioned one with counts of parameters.

bbatsov:
Okay, let's go for it then. Sorry that I'm making you do so much extra work, I just have trouble remembering everything we've done so far. 😆

fatkodima:
No worries. I have a lot of free time 😄

PR #9016, opened 2020-11-09
[Fix #9015] Update `Lint/EmptyBlock` to allow for empty lambdas.

### LINKED ISSUE #9015: `Lint/EmptyBlock` should not apply to empty lamdas
It is often useful to have an empty lambda as a callable placeholder (for instance to stub out a callable or as an initial value in dynamic function composition. `Lint/EmptyBlock` flags these, but there isn't really a better way to express these (`-> { nil }` is not better here IMO). This should apply to both `lambda` and `->`.

--------

## Expected behavior

Rubycop should not register any offenses about this code:

```ruby
expect(Foo).to receive(:callable).and_return(-> {})
```

## Actual behavior

Rubocop registers an offense.

```
Offenses:

spec/test_spec.rb:1:46: W: Lint/EmptyBlock: Empty block detected.
expect(Foo).to receive(:callable).and_return(-> {})
                                             ^^^^^
```

## RuboCop version

```
1.2.0 (using Parser 2.7.2.0, rubocop-ast 1.1.1, running on ruby 2.6.2 x86_64-darwin18)
  - rubocop-performance 1.8.1
  - rubocop-rails 2.8.1
  - rubocop-rspec 2.0.0
```

dvandersluis:
Update `Lint/EmptyBlock` to allow for empty lambdas. Fixes #9015.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1505,7 +1505,9 @@ Lint/EmptyBlock:
   Description: 'This cop checks for blocks without a body.'
   Enabled: pending
   VersionAdded: '1.1'
+  VersionChanged: '1.3'
   AllowComments: true
+  AllowEmptyLambdas: true
 
 Lint/EmptyClass:
   Description: 'Checks for classes and metaclasses without a body.'
--- a/docs/modules/ROOT/pages/cops_lint.adoc
+++ b/docs/modules/ROOT/pages/cops_lint.adoc
@@ -1131,12 +1131,14 @@ end
 | Yes
 | No
 | 1.1
-| -
+| 1.3
 |===
 
 This cop checks for blocks without a body.
 Such empty blocks are typically an oversight or we should provide a comment
 be clearer what we're aiming for.
+
+Empty lambdas are ignored by default.
 
 === Examples
 
@@ -1173,12 +1175,40 @@ end
 items.each { |item| } # TODO: implement later (inline comment)
 ----
 
+==== AllowEmptyLambdas: true (default)
+
+[source,ruby]
+----
+# good
+allow(subject).to receive(:callable).and_return(-> {})
+
+placeholder = lambda do
+end
+(callable || placeholder).call
+----
+
+==== AllowEmptyLambdas: false
+
+[source,ruby]
+----
+# bad
+allow(subject).to receive(:callable).and_return(-> {})
+
+placeholder = lambda do
+end
+(callable || placeholder).call
+----
+
 === Configurable attributes
 
 |===
 | Name | Default value | Configurable values
 
 | AllowComments
+| `true`
+| Boolean
+
+| AllowEmptyLambdas
 | `true`
 | Boolean
 |===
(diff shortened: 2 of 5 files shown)


koic:
This looks like a reasonable change to me if an example for `lambda do(\n)end` and `-> {}` to good cases are added.

dvandersluis:
@koic sounds good I’ll add one!

dvandersluis:
@koic updated!

docs/modules/ROOT/pages/cops_lint.adoc
```diff
@@ -1147,6 +1149,11 @@ items.each { |item| }
 
 # good
 items.each { |item| puts item }
+
+allow(subject).to receive(:callable).and_return(-> {})
+
+lambda do
+end
```
koic, line 1156: Ah, this was a bad example I wrote 💦  IMHO, the case where empty `{}` is often used is in the one-liner.
E.g. `-> {}` or `lambda {}`.
So, I'd like to hear other opinions as to whether it should be allowed even in the case of multi-liner.

bbatsov:
It may be better to have this as a configuration option enabled by default. Not sure about that, though, as I guess empty lambdas are relatively common.

dvandersluis:
I found a bunch of them in rails/rails as well as my work repo. I don't think that this is an uncommon pattern.

dvandersluis:
If you want it to be an option though I'll fix it up in the morning!

docs/modules/ROOT/pages/cops_lint.adoc
dvandersluis: I think it makes sense to allow both, because `Style/Lambda` exists and is configurable to both kinds.

koic:
Yeah, I've also seen `-> {}` and `lambda {}` than `-> { nil }` and `lambda { nil }`. IMHO, these one-liner lambda empty blocks `{}` can be allowed by configuration default.

bbatsov:
Fine by me. Probably we need a cop for empty lamba with a `nil` in it too. :-)

dvandersluis:
@bbatsov @koic updated to add `AllowEmptyLambdas` configuration, default on. Also updated the documentation examples a bit further.

config/default.yml
```diff
@@ -1505,7 +1505,9 @@ Lint/EmptyBlock:
   Description: 'This cop checks for blocks without a body.'
   Enabled: pending
   VersionAdded: '1.1'
+  VersionChanged: '1.3'
   AllowComments: true
+  AllowEmptyLambdas: true
```
koic, line 1510: Can you make the configuration name consistent with either `AllowEmptyLambda` or `AllowEmptyLambdas`?

dvandersluis: 🤦 fixing!

dvandersluis: Fixed now, sorry about that.

bbatsov:
Looks good to me. Thanks!

PR #9024, opened 2020-11-11
[Fix #9023] Mark unsafe for `Style/CollectionCompact`.

### LINKED ISSUE #9023: Incorrect autocorrection by Style/CollectionCompact
The new `Style/CollectionCompact` cop may change some code behavior. It assumes that `reject` can be replaced with `compact` even when it shouldn't. Probably it is applying to an `Array#reject` a correction that would only be valid for a `Hash#reject`. See below for an example.

--------

## Expected behavior
When a `reject` method is called for an array with a block with two arguments, and this block only checks if the second argument is nil, this `reject` shouldn't be replaced with a `compact`. 

## Actual behavior
`reject` gets replaced with `compact` in that scenario.

* Create this ruby file (lets call it `test.rb`):
```ruby
x = [[1, nil], [2, 3], [4, nil]]
x = x.reject { |_a, b| b.nil? }
puts x.size
```
* Run it
```sh
$ ruby test.rb
1
```
* Autocorrect this file with rubocop, selecting only the Style/CollectionCompact cop:
```sh
$ bundle exec rubocop --only Style/CollectionCompact test.rb -a
(some lines about new cops not enabled omited)
Inspecting 1 file
Scanning /home/wendel/projects/voltera-api/test.rb
C

Offenses:

test.rb:2:7: C: [Corrected] Style/CollectionCompact: Use compact instead of reject { |_x, y| y.nil? }.
x = x.reject { |_x, y| y.nil? }
      ^^^^^^^^^^^^^^^^^^^^^^^^^

```
* File content is now:
```ruby
x = [[1, 2], [3, nil], [4, nil]]
x = x.compact
puts x.size
```
* Run the modified script:
```sh
$ ruby test.rb
3
```
* The output is now 3 instead of 1

## RuboCop version

```
$ bundle exec rubocop -V
1.2.0 (using Parser 2.7.2.0, rubocop-ast 1.1.1, running on ruby 2.7.1 x86_64-linux)
  - rubocop-performance 1.8.1
  - rubocop-rails 2.8.1
  - rubocop-rake 0.5.1
```

koic:
Fixes #9023.

This PR marks unsafe for `Style/CollectionCompact` by default because false positives may occur in the nil check of block arguments to the receiver object.
For example, `[[1, 2], [3, nil]].reject { |first, second| second.nil? }` and `[[1, 2], [3, nil]].compact` are not compatible. This will work fine when the receiver is a hash object.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_mark_unsafe_for_collection_compact.md
+++ b/changelog/change_mark_unsafe_for_collection_compact.md
@@ -1,0 +1,1 @@
+* [#9023](https://github.com/rubocop-hq/rubocop/issues/9023): Mark unsafe for `Style/CollectionCompact`. ([@koic][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2833,7 +2833,9 @@ Style/ClassVars:
 Style/CollectionCompact:
   Description: 'Use `{Array,Hash}#{compact,compact!}` instead of custom logic to reject nils.'
   Enabled: pending
+  Safe: false
   VersionAdded: '1.2'
+  VersionChanged: '1.3'
 
 # Align with the style guide.
 Style/CollectionMethods:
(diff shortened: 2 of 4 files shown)


marcandre (approved):
👍

PR #9028, opened 2020-11-11
[Fix #9011] Mark autocorrection for `Lint/Loop` as unsafe

### LINKED ISSUE #9011: Autocorrect of Lint/Loop can be unsafe
##  Expected behavior

Safe auto correct of Lint/Loop should not introduce errors in the code

## Actual behavior

A `begin`...`end` `until` loop introducing a local variable gets corrected to a `loop do` .. `break if` .. `end` structure which has its own local variable scope, potentially breaking the code

The following code:

```ruby
def some_check?
  true
end

def generate_key
  42
end

begin
  key = generate_key
end until some_check?

puts key
```

gets auto corrected (run `rubocop -a` on the file) by the `Lint/Loop` cop to the following:

```ruby
def some_check?
  true
end

def generate_key
  42
end

loop do
  key = generate_key
  break if some_check?
end

puts key
```

which generates the following error instead of printing `42`:

```
test.rb:14:in `<main>': undefined local variable or method `key' for main:Object (NameError)
```

## RuboCop version

```
$ [bundle exec] rubocop -V
1.2.0 (using Parser 2.7.2.0, rubocop-ast 1.1.0, running on ruby 2.7.2 x86_64-linux)
```

dvandersluis:
The case in #9011 

```ruby
loop do
  key = generate_key
  break if some_check?
end

puts key
```

cannot be detected by static analysis, so instead I've marked autocorrection for `Lint/Loop` as unsafe and added an explanation.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_mark_autocorrection_for_lintloop_as.md
+++ b/changelog/change_mark_autocorrection_for_lintloop_as.md
@@ -1,0 +1,1 @@
+* [#9011](https://github.com/rubocop-hq/rubocop/issues/9011): Mark autocorrection for `Lint/Loop` as unsafe. ([@dvandersluis][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -1659,7 +1659,8 @@ Lint/Loop:
   StyleGuide: '#loop-with-break'
   Enabled: true
   VersionAdded: '0.9'
-  VersionChanged: '0.89'
+  VersionChanged: '1.3'
+  Safe: false
 
 Lint/MissingCopEnableDirective:
   Description: 'Checks for a `# rubocop:enable` after `# rubocop:disable`.'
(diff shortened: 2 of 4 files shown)


marcandre:
> cannot be detected by static analysis

Are you sure?

Also, like I wrote in #9008, `loop` rescues `StopIteration`, so it's never strictly equivalent.

dvandersluis:
@marcandre I was simplifying, we could I suppose detect every lvar in the begin and look for it in the following code but there are always edge cases and I don’t think it’s worthwhile. 

Didn’t know about `StopIteration` but then that’s another reason to make it unsafe. I can change it to be unsafe as a whole and not just for autocorrection.

marcandre:
> I can change it to be unsafe as a whole and not just for autocorrection.

dvandersluis:
Updated!

marcandre:
Thanks!

PR #9037, opened 2020-11-12
Fix `required_ruby_version` issue when using `Gem::Requirement`

cetinajero:
This fixes an issue when the `required_ruby_version` uses `Gem::Requirement` to define the Ruby version.

Gemspec:

```
...
s.required_ruby_version = Gem::Requirement.new('< 3.0.0')
...
```

Backtrace using v1.3.0:

```
undefined method `match' for nil:NilClass
296/home/travis/build/grupopv/jekyll-tasks/vendor/bundle/ruby/2.5.0/gems/rubocop-1.3.0/lib/rubocop/target_ruby.rb:165:in `version_from_str'
297/home/travis/build/grupopv/jekyll-tasks/vendor/bundle/ruby/2.5.0/gems/rubocop-1.3.0/lib/rubocop/target_ruby.rb:144:in `find_version'
298/home/travis/build/grupopv/jekyll-tasks/vendor/bundle/ruby/2.5.0/gems/rubocop-1.3.0/lib/rubocop/target_ruby.rb:22:in `initialize'
```

Related to https://github.com/rubocop-hq/rubocop/issues/8761

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/target_ruby.rb
+++ b/lib/rubocop/target_ruby.rb
@@ -141,7 +141,15 @@ module RuboCop
           return versions.compact.min
         end
 
+        return gem_requirement_version(version) if version.send_type?
+
         version_from_str(version.str_content)
+      end
+
+      def gem_requirement_version(version)
+        gem_requirement = version.children.last
+        versions = gem_requirement.children.map { |v| version_from_str(v) }
+        versions.compact.min
       end
 
       def gemspec_filename
--- a/lib/rubocop/target_ruby.rb
+++ b/lib/rubocop/target_ruby.rb
@@ -140,20 +140,11 @@ module RuboCop
         version = version_from_gemspec_file(file)
         return if version.nil?
 
-        if version.array_type?
-          versions = version.children.map { |v| version_from_str(v.str_content) }
-          return versions.compact.min
-        end
-
-        return gem_requirement_version(version) if gem_requirement? version
+        requirement = version.children.last
+        return version_from_array(version) if version.array_type?
+        return version_from_array(requirement) if gem_requirement? version
 
         version_from_str(version.str_content)
-      end
-
-      def gem_requirement_version(version)
-        gem_requirement = version.children.last
-        versions = gem_requirement.children.map { |v| version_from_str(v) }
-        versions.compact.min
       end
 
       def gemspec_filename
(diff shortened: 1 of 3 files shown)


lib/rubocop/target_ruby.rb
```diff
@@ -141,9 +141,17 @@ def find_version
           return versions.compact.min
         end
 
+        return gem_requirement_version(version) if version.send_type?
```
bbatsov, line 144: Probably we should do a more precise node match here, as there's a slight chance some other code might be used here.

lib/rubocop/target_ruby.rb
```diff
@@ -141,9 +141,17 @@ def find_version
...
+
         version_from_str(version.str_content)
       end
 
+      def gem_requirement_version(version)
```
bbatsov, line 149: As this is still extracted from the gemspec it should probably be part of that existing method.

lib/rubocop/target_ruby.rb
cetinajero: Thanks bbatsov!

There is now a more precise match pattern.

lib/rubocop/target_ruby.rb
cetinajero: Initially I had this logic on the existing method (`find_version`), but I got the _Assignment Branch Condition_ and the _Cyclomatic complexity_ too high, that's why I created a new method.

Any suggestions @bbatsov?

lib/rubocop/target_ruby.rb
marcandre: Looks good! Maybe you can merge it with `if version.array_type?`, also capturing the versions, and you can reuse the same `map`, `compact`, `min`. Maybe not though, just a thought

cetinajero: I'll look into it @marcandre!

marcandre: `{ (array $str+) ... } ` can help, and there's https://nodepattern.herokuapp.com/ to help test things.

cetinajero: So now the `map`, `compact` and `min` are reused and I definitely think it looks more DRY.

I managed to merge it by creating a `version_from_array` method but didn't do any new `NodePattern`.

lib/rubocop/target_ruby.rb
cetinajero: New arrangement based on https://github.com/rubocop-hq/rubocop/pull/9037#discussion_r523170441, let me know your thoughts @bbatsov .

bbatsov:
Looks good! Thanks!

PR #9043, opened 2020-11-13
Add an option to redirect offence reports to stderr

knu:
This adds `--stderr` to make it easier to use `--stdin` in shell scripts, editor integration, CI workflow, etc.

In response to feedbacks, I altered this PR from changing the behavior of `--stdin` to adding a new option. 
<details>
<summary>Here's the original description.</summary>
This makes editor integration much easier to implement by eliminating the need to process the output.

In the short term, existing editor plugins that expect `====================` will be broken, but it should be relatively easy to fix.

For example, [vscode-ruby-rubocop](https://github.com/misogi/vscode-ruby-rubocop/blob/master/src/rubocop.ts) can check if stderr output contains (or ends with) `====================` to see if it needs to delete warnings from stdout output. (to support any version of rubocop)

Eventually, they just need to simply use the stdout output.
</details>

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_redirect_offence_reports_to_stderr_when.md
+++ b/changelog/new_redirect_offence_reports_to_stderr_when.md
@@ -1,0 +1,1 @@
+* [#9043](https://github.com/rubocop-hq/rubocop/pull/9043): Add `--stderr` to write all output to stderr except for the autocorrected source. ([@knu][])
--- a/docs/modules/ROOT/pages/usage/basic_usage.adoc
+++ b/docs/modules/ROOT/pages/usage/basic_usage.adoc
@@ -240,6 +240,9 @@ $ rubocop --only Rails/Blank,Layout/HeredocIndentation,Naming/FileName
 | `--show-cops`
 | Shows available cops and their configuration.
 
+| `--stderr`
+| Write all output to stderr except for the autocorrected source. This is especially useful when combined with `--auto-correct` and `--stdin`.
+
 | `-s/--stdin`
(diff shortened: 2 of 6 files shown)


bbatsov:
I get the rationale, but won't this break the existing editor plugins?

dvandersluis:
Yeah my concern is editors that you can’t just update by pulling a repo are going to be broken until they can be updated (like rubymine).

knu:
An easiest fix is to add `2>&1` to the command line, and we can inform that or submit PRs to the developers of those known plugins.

https://docs.rubocop.org/rubocop/integration_with_other_tools.html

knu:
Would it better to keep the behavior and add a new option, like `--stderr` for example?

bbatsov:
I think so. Might also be a good idea to invite some of the plugin authors to share their perspective on this.

knu:
OK, I changed so.  A new option `--stderr` redirects all output except for the autocorrected source to stderr.

tas50:
Thanks for changing this to be an option. That'll make things a lot easier for those of us that parse rubocop output in editor plugins / other tools

bbatsov:
I think you should add some section to the docs explaining the use-case for this option, otherwise I doubt people will make use of it.

knu:
I updated the description of this option mentioning it is useful when combined with `--auto-correct` and `--stdin`.

bbatsov:
Thanks!

PR #9045, opened 2020-11-14
Have `cut_release` handle "config/default" and generate cops doc.

marcandre:
* task `cut_release` will replace the string "<<next_version>>" in the default.yml file with the version being cut (matching is permissive: quoted or unquoted, with "-" or "_", etc.)
* task `generate_cops_documentation` now does nothing and only prints out that doc generation is now automatic
* processing code moved to new task `update_cops_documentation` (no public description) that is triggered by `cut_release`
* new cop generator uses this by default instead of the next minor version

Did I forget anything?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -234,9 +234,6 @@ jobs:
       - run:
           name: Check documentation syntax
           command: bundle exec rake documentation_syntax_check
-      - run:
-          name: Build documentation and verify that doc files are in sync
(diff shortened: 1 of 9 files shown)


bbatsov:
> <<next_version>>

I think we can shorten this to `<<next>>`. That would simply the matching a regexp a bit. 

> Did I forget anything?

I don't think so.

marcandre:
Changed to `<<next>>` 👍

PR #9051, opened 2020-11-15
[Fix #9050] Fix a false positive for `Style/NegatedIfElseCondition`

### LINKED ISSUE #9050: Style/NegatedIfElseCondition: False positive on `if !!`
Using `!!` isn't negation, it's conversion to Boolean. So this should be valid code:

```ruby
if !!my_method_or_variable
  "true condition"
else
  "false condition"
end

# Fails with:
Style/NegatedIfElseCondition: Invert the negated condition and swap the if-else branches.
  if !!my_method ...
  ^^^^^^^^^^^^^^
```

--------

## Expected behavior

Don't fail when `!!` is used in an if/else block. i.e. this should be valid:

```ruby
if !!my_method_or_variable
  "true condition"
else
  "false condition"
end
```

## Actual behavior

The Style/NegatedIfElseCondition cop fails with:
> Style/NegatedIfElseCondition: Invert the negated condition and swap the if-else branches

Add an if/else such as the following to a ruby file and run the Style/NegatedIfElseCondition cop:

```ruby
if !!my_method
  "a"
else
  "b"
end
```

## RuboCop version

```
$ rubocop -V
1.3.0 (using Parser 2.7.2.0, rubocop-ast 1.1.1, running on ruby 2.6.6 x86_64-darwin19)
  - rubocop-performance 1.8.1
  - rubocop-rails 2.8.1
  - rubocop-rake 0.5.1
```

koic:
Fixes #9050.

This PR fixes a false positive for `Style/NegatedIfElseCondition` when `if` with `!!` condition'.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/negated_if_else_condition.rb
+++ b/lib/rubocop/cop/style/negated_if_else_condition.rb
@@ -35,6 +35,8 @@ module RuboCop
 
         NEGATED_EQUALITY_METHODS = %i[!= !~].freeze
 
+        def_node_matcher :double_negation?, '(send (send _ :!) :!)'
+
         def self.autocorrect_incompatible_with
           [Style::InverseMethods, Style::Not]
         end
@@ -47,7 +49,7 @@ module RuboCop
           return unless if_else?(node)
 
           condition = node.condition
-          return unless negated_condition?(condition)
+          return if double_negation?(condition) || !negated_condition?(condition)
 
           type = node.ternary? ? 'ternary' : 'if-else'
           add_offense(node, message: format(MSG, type: type)) do |corrector|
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/negated_if_else_condition.rb
```diff
@@ -35,6 +35,8 @@ class NegatedIfElseCondition < Base
 
         NEGATED_EQUALITY_METHODS = %i[!= !~].freeze
 
+        def_node_matcher :double_negative?, '(send (send _ :!) :!)'
```
bbatsov, line 38: Let's name this `double_negation?` to be consistent with the terminology we've used so far.

koic: Sure! I updated it.

PR #9052, opened 2020-11-15
[Fixes #9049] Have `Lint/ToEnumArguments` accept `__callee__`

### LINKED ISSUE #9049: Lint/ToEnumArguments: False positive on `to_enum(__callee__)`
Given the following code, Lint/ToEnumArguments reports an offense:

```ruby
def my_method
  return to_enum(__callee__) unless block_given?
  # ...
end

# Offense:
Lint/ToEnumArguments: Ensure you correctly provided all the arguments
  return to_enum(__callee__) unless block_given?
         ^^^^^^^^^^^^^^^^^^^
```

Whereas, in the same method, just changing `__callee__` to the name of the method passes:

```ruby
def my_method
  return to_enum(:my_method) unless block_given?
  # ...
end

# Passes ^
```

--------

## Expected behavior

Since [`__callee__`](https://ruby-doc.org/core-2.6.5/Kernel.html#method-i-__callee__) 
> Returns the called name of the current method as a Symbol. If called outside of a method, it returns nil.

... this should be an acceptable (and even preferred) method of using `to_enum`, because it allows the method name to be changed without requiring a change to the internals of the method to match.

## Actual behavior

A somewhat misleading offense is detected/reported:
> Ensure you correctly provided all the arguments

Just add this method to a ruby file and run Lint/ToEnumArguments on it:

```ruby
def my_method
  to_enum(__callee__)
end

def my_method2
  to_enum(:my_method2)
end
```

Both should pass, but only the 2nd one does.

## RuboCop version

```
$ rubocop -V
1.3.0 (using Parser 2.7.2.0, rubocop-ast 1.1.1, running on ruby 2.6.6 x86_64-darwin19)
  - rubocop-performance 1.8.1
  - rubocop-rails 2.8.1
  - rubocop-rake 0.5.1
```

marcandre:


bbatsov:
I think it'd be good to update the documentation examples to mention it.

koic:
I removed "auto-merge" label because it is preferable for the doc to be updated. Thank you for fixing.

marcandre:
> I removed "auto-merge" label because it is preferable for the doc to be updated. Thank you for fixing.

Thanks! I was going to beat it 😆

koic (approved):
👍

PR #9053, opened 2020-11-15
[Fixes #8564] `Metrics/AbcSize`: Add optional discount for repeated "attributes"

### LINKED ISSUE #8564: Redefine what counts as a branch on ABC Size

marcandre:
That was more involved that I had hoped for...

I'm unsure as to the naming of the new config option; within the code I feel that `discount_repeated_attributes` was easiest to think about, but for the config I'm proposing `CountRepeatedAttributes` instead of `Discount...`. Maybe `IgnoreRepeatedAttributes` would be better?


marcandre:
Also wondering if certain methods should never be considered as attributes, say `pop shift drop receive new save sort delete next reset flush`. Most are unlikely to be repeated on the same object, but some could:

```ruby
def process
  foo(queue.pop)
  bar(queue.pop) # This second `pop` should always be counted...
  # ...
end
```

Worth it?

lib/rubocop/cop/metrics/abc_size.rb
```diff
@@ -7,6 +7,27 @@ module Metrics
       # configured maximum. The ABC size is based on assignments, branches
       # (method calls), and conditions. See http://c2.com/cgi/wiki?AbcMetric
       # and https://en.wikipedia.org/wiki/ABC_Software_Metric.
+      #
+      # You can have repeated "attributes" calls count as a single "branch".
+      # For this purpose, attributes are any method with no argument; no attempt
```
bbatsov, line 12: and an implicit/current object receiver I guess.

bbatsov (approved):
I like your proposal. Definitely makes a lot of sense.

lib/rubocop/cop/metrics/abc_size.rb
marcandre: Not necessarily:
```ruby
foo.bar.baz # => 3 branches
foo.bar.qux # => 1 more branch only (`foo` and `bar` are "repeated")

n = 42
n.to_s # => 1 branch
n.to_s.to_i # => just 1 more (`to_s` "repeated")
```

marcandre:
> I like your proposal. Definitely makes a lot of sense.

Thanks. So `CountRepeatedAttributes` is good?

bbatsov:
I think it's fine - I don't have a strong preference for it or `IgnoreRepeatedAttributes`.

marcandre:
Last remaining question:

> Also wondering if certain methods should never be considered as attributes, say `pop shift drop receive new save sort delete next reset flush`. Most are unlikely to be repeated on the same object, but some could:
> 
> ```ruby
> def process
>   foo(queue.pop)
>   bar(queue.pop) # This second `pop` should always be counted...
>   # ...
> end
> ```
> 
> Worth it?

bbatsov:
Seems like an overkill to me, but I'd be fine with adding some special handling for them (ideally in the form of configuration).

marcandre:
Ok, then this is mergeable as is then 👍

Rebased just to be sure (maybe there are new style issues)

bbatsov:
Great! It's heading to production later today or tomorrow with the next release.

PR #9058, opened 2020-11-16
[Fix #8993] Allow `ExcludedMethods` config of `Metrics/MethodLength` cop to contain regex

### LINKED ISSUE #8993: Metrics/MethodLength ExcludedMethods Regex

Maybe this one is actually working, but I couldn't do it. It would be nice if you could add a regular expression to the ExludedMethods array.

I would love to add Regex to ExludedMethods, or to know how if it's already possible.

Something like this:
`ExcludedMethods: [!ruby/regexp \b\w*params\w*\b\$/, 'create', 'update', 'new', 'edit']`
meaning all methods which have **params** string in it would be ignored, trying to archive that my controllers and their's permitted params are always skipped when MethodLenght check is being controlled.

Ofc I could skip this check for whole controllers folder, but I want it to be as much precise as possible.

tejasbubane:
Closes #8993

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changes
 
 * [#8788](https://github.com/rubocop-hq/rubocop/issues/8788): Change `Style/Documentation` to not trigger offense with only macros. ([@tejasbubane][])
+* [#8993](https://github.com/rubocop-hq/rubocop/issues/8993): Allow `ExcludedMethods` config of `Metrics/MethodLength` cop to contain regex. ([@tejasbubane][])
 
 ## 1.3.1 (2020-11-16)
 
--- a/lib/rubocop/cop/metrics/method_length.rb
+++ b/lib/rubocop/cop/metrics/method_length.rb
@@ -36,7 +36,7 @@ module RuboCop
 
         def on_def(node)
           excluded_methods = cop_config['ExcludedMethods']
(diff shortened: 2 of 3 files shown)


bbatsov:
I'm wondering if this should be mentioned in the docs, as I have a feeling we're not consistent with this behavior across cops. Probably somewhere it's a string match and somewhere a regexp match.

marcandre:
> Probably somewhere it's a string match and somewhere a regexp match.

Ideally it should be general (string or regexp) everywhere though, no?

tejasbubane:
> Ideally it should be general (string or regexp) everywhere though, no?

I think we have some inconsistency with the configuration naming. `ExcludedMethods` is used only by two cops: `Metrics/BlockLength` and `Metrics/MethodLength`. Rest use `IgnoredMethods`.

bbatsov:
Yep, I was going to mention this as well, but it slipped my mind. I think we should unify this and just add two possible names for the problematic cops, so that we don't break backwards compatibility. 

> Ideally it should be general (string or regexp) everywhere though, no?

Yep. I guess we should go for regex everywhere, as it's more flexible.

bbatsov:
@dvandersluis I see you liked the suggestion I made, so I'm wondering if you'd be willing to implement this. 😄

dvandersluis:
@bbatsov definitely! I will take a look in the next couple days.

PR #9066, opened 2020-11-17
Update `Lint/UnmodifiedReduceAccumulator` to accept blocks which return in the form `accumulator[element]`

### LINKED ISSUE #9059: False positive of UnmodifiedReduceAccumulator for `self`
I'm working on [R18n gem](https://github.com/r18n/r18n) and I've got a new offense from RuboCop.

--------

## Expected behavior

No offenses.

## Actual behavior

The offense:

```
r18n-core/lib/r18n-core/translation.rb:171:41: W: Lint/UnmodifiedReduceAccumulator: Do not return an element of the accumulator in reduce.
      keys.reduce(self) { |result, key| result[key] }
                                        ^^^^^^^^^^^
```

```ruby
# Custom class with `#[]` method
def dig(*keys)
  keys.reduce(self) { |result, key| result[key] }
end
````

## RuboCop version

```
$ [bundle exec] rubocop -V
1.3.1 (using Parser 2.7.2.0, rubocop-ast 1.1.1, running on ruby 2.7.2 x86_64-linux)
```

### PR DESCRIPTION
**Pull request by dvandersluis**
Reduce blocks that just return something in the form `accumulator[element]` are now accepted by `Lint/UnmodifiedReduceAccumulator` (the issue asked for all methods on the accumulator to be allowed but this was the only exception previously due to the check for accumulator index). Other accumulator indexes are still offenses.

Fixes #9059.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/unmodified_reduce_accumulator_spec.rb
+++ b/spec/rubocop/cop/lint/unmodified_reduce_accumulator_spec.rb
@@ -373,12 +373,40 @@ RSpec.describe RuboCop::Cop::Lint::UnmodifiedReduceAccumulator do
         RUBY
       end
 
+      it 'does not register an offense when calling a method on the accumulator with the element' do
+        expect_no_offenses(<<~RUBY)
+          foo.#{method} { |result, key| result.method(key) }
+        RUBY
+      end
+
+      it 'registers an offense when returning an index of the accumulator' do
+        expect_offense(<<~RUBY)
+          %w(a b c).#{method}({}) do |acc, letter|
+            acc[foo]
+            ^^^^^^^^ Do not return an element of the accumulator in `#{method}`.
+          end
+        RUBY
+      end
+
       it 'registers an offense when returning an index setter on the accumulator' do
         expect_offense(<<~RUBY)
           %w(a b c).#{method}({}) do |acc, letter|
-            acc[letter] = true
-            ^^^^^^^^^^^^^^^^^^ Do not return an element of the accumulator in `#{method}`.
+            acc[foo] = bar
+            ^^^^^^^^^^^^^^ Do not return an element of the accumulator in `#{method}`.
           end
+        RUBY
+      end
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: spec/rubocop/cop/lint/unmodified_reduce_accumulator_spec.rb [hunk 1]
```diff
@@ -373,15 +373,42 @@
...
 
+      it 'does not register an offense when returning accumulator[element]' do
+        expect_no_offenses(<<~RUBY)
+          foo.#{method} { |result, key| result[key] }
+        RUBY
+      end
+
+      it 'does not register an offense when returning accumulator[element]=' do
+        expect_no_offenses(<<~RUBY)
+          foo.#{method} { |result, key| result[key] = foo }
```
marcandre, line 408: Just curious, but is there a legitimate use case for this form?

dvandersluis: Probably not, other than it being of the form `result.send <anything>, key` as in the issue. I can remove it.

dvandersluis: Updated to remove the exception for `[]=`

marcandre: Ah ah, you got me, I mispoke 😆.
Thanks for updating it; maybe someone will come up with a good usecase but I can't think of one.

marcandre: > Ah ah, you got me, I mispoke 😆.

Well, I take that back, `result.send <anything>, key` doesn't actually fit `:[]=`; that requires an extra argument! My honor is safe 🤣

dvandersluis: Yeah I guess I interpreted it as `key, ...` but in any case all good now 😄

PR #9072, opened 2020-11-20
`Lint/MissingSuper`: exclude `method_missing` and `respond_to_missing?`

marcandre:
Contrary to the other hooks, these have very valid cases to be completely taken over:

```ruby
# accept everything:
def method_missing(method, *args)
  nil
end

# or delegate all
def method_missing(method, *args)
  delegate_to.send(method, *args)
end
```

While a class has a "right" to expect its other hooks to be called, these hooks aren't necessarily called the same way.

From https://github.com/testdouble/standard/issues/195#issuecomment-729863480


bbatsov:
I'm fine with the proposed change, but I think it should be clearly explained in the cop's documentation why are those hooks treated differently.

marcandre:
I did my best while staying concise.

bbatsov:
Looks good! Thanks!

PR #9075, opened 2020-11-21
[Fix #8857] Modify --auto-gen-config regarding Exclude and inheritance

### LINKED ISSUE #8857: .rubocop_todo.yml mentions file explicitly Excluded in .rubocop.yml
## Expected behavior

When running `rubocop --auto-gen-config --auto-gen-only-exclude --exclude-limit 100` on a project where the following is present in `.rubocop.yml`:
```
Metrics/MethodLength:
  Exclude:
    - path/to/excluded/file.rb
```
I expect that the `path/to/excluded/file.rb` will not appear in `.rubocop_todo.yml` under the `Metrics/MethodLength` cop since I have already explicitly asked Rubocop to Exclude it. I know the Exclude is working because when I run `rubocop` by itself, no offenses are reported on that file even though there are offenses present within it.

## Actual behavior

The following appears in `.rubocop_todo.yml`:

```
# Offense count: 6
# Configuration parameters: CountComments, Max, CountAsOne, ExcludedMethods.
Metrics/MethodLength:
  Exclude:
    - 'path/to/included/file.rb'
    - 'path/to/excluded/file.rb'
```

Note that the `path/to/included/file.rb` contains violations of the `Metrics/MethodLength` rule. While that is also true of `path/to/excluded/file.rb`, we explicitly Exclude that file in our `.rubocop.yml` for that specific cop.

Note that if we delete `path/to/included/file.rb` and run `rubocop --auto-gen-config --auto-gen-only-exclude --exclude-limit 100` again, neither of the files appears as offenders in the `.rubocop_todo.yml`.

1. Create two local Ruby files, both containing violations of `Metrics/MethodLength`
2. Create a `.rubocop.yml` file that contains a rule to Exclude only one of these files for that specific cop:
```
Metrics/MethodLength:
  Exclude:
    - path/to/excluded/file.rb
```
3. Run `rubocop --auto-gen-config --auto-gen-only-exclude --exclude-limit 100`
4. Observe that the `path/to/excluded/file.rb` appears under `Metrics/MethodLength` in `.rubocop_todo.yml`

## RuboCop version

```
0.92.0 (using Parser 2.7.2.0, rubocop-ast 0.7.1, running on ruby 2.6.6 x86_64-darwin19)
```

jonas054:
When we generate `Exclude` properties in `.rubocop_todo.yml`, we already check if the same `Exclude` appears in existing configuration. The file list is then copied into the generated file. This doesn't work as intended for `Exclude` properties in custom configuration. They override (by default) the same properties inherited from `.rubocop_todo.yml`, resulting in a configuration that will not hide all the offenses when `rubocop` is next run. The expected behavior is that `--auto-gen-config` should hide all offenses.

This change makes `rubocop --auto-gen-config` warn about this situation when it occurs. If `inherit_mode` is set to `merge` for `Exclude` properties, the behavior is pretty much as expected today, but we tweak the behavior here too, and skip the copying of file names that will be excluded anyway due to the merging.


bbatsov:
Nice improvement!

PR #9076, opened 2020-11-21
[Fix #7338] Handle assignment with []= in MultilineMethodCallIndentation

### LINKED ISSUE #7338: False positive for `MultilineMethodCallIndentation` with `[]` on the left side of assignment
There is false positive for `MultilineMethodCallIndentation` with `[]` on the left side of assignment.

--------

## Expected behavior

Without offenses.

## Actual behavior

With offenses.

```yaml
# .rubocop.yml
Layout/MultilineMethodCallIndentation:
  EnforcedStyle: indented
```

```ruby
# test.rb
# frozen_string_literal: true

params[:key] =
  foo
    .bar
    .baz
```

```
> bundle exec rubocop test.rb
Inspecting 1 file
C

Offenses:

test.rb:5:5: C: Layout/MultilineMethodCallIndentation: Use 2 (not 4) spaces for indenting an expression in an assignment spanning multiple lines.
    .bar
    ^^^^
test.rb:6:5: C: Layout/MultilineMethodCallIndentation: Use 2 (not 4) spaces for indenting an expression in an assignment spanning multiple lines.
    .baz
    ^^^^
```

## RuboCop version

```
0.74.0 (using Parser 2.6.4.0, running on ruby 2.6.4 x86_64-linux)
```

jonas054:
The `[]=` operator is a send node in the AST, which we need to consider while traveling up the ancestor hierarchy of send nodes, looking for the base of alignment.

### DIFF
--- a/changelog/fix_MultilineMethodCallIndentation_indexed_assignment.md
+++ b/changelog/fix_MultilineMethodCallIndentation_indexed_assignment.md
@@ -1,0 +1,2 @@
+* [#7338](https://github.com/rubocop-hq/rubocop/issues/7338): Handle assignment with `[]=` in `MultilineMethodCallIndentation`. ([@jonas054][])
+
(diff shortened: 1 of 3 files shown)


bbatsov:
Looks good! Thanks!

PR #9078, opened 2020-11-21
[Fix #9064] Add three new config options for `Layout/EmptyLineBetweenDefs` cop

### LINKED ISSUE #9064: Apply Layout/EmptyLineBetweenDefs to class definitions as well

The `Layout/EmptyLineBetweenDefs` cop only applies between two functions, so the following pundit policy code raises no offenses.
```ruby
class UserPolicy < ApplicationPolicy
  class Scope < Scope
    def resolve
      scope.all
    end
  end
  def show?
    true
  end
end
```
I would like a cop to raise an offense about how there's no empty line between the end of the `Scope` class and the `def show?` line.

My preference would be to modify the `Layout/EmptyLineBetweenDefs` to also apply to class definitions. This is within the intent of the cop that definitions should be visually spaced out to help readability.

Options are to:
1) Modify the cop in place.
	- Pro: gets the most adoption with no user configuration
	- Con: raises offenses on code that was previously inoffensive
2) Add a configuration option to the cop
	- Pro: this is safer since it lets people toggle the behaviour
	- Con: more complicated configuration
3) Add a new cop
	- Pro: safe option for users that doesn't require more configuration
	- Con: Possible interactions with `Layout/EmptyLineBetweenDefs` cop

## Additional context

There's nothing specifically addressing this situation in the Ruby style guide so I can't be 100% certain my suggestion is correct. If there's no consensus that it is, then option 2 or 3 will let people choose their own preference.

I'd be happy to make the PR once people agree on which option to go with.

tejasbubane:
* EmptyLineBetweenMethodDefs
* EmptyLineBetweenClassDefs
* EmptyLineBetweenModuleDefs

This cop previous had ability to check for empty lines between method
definitions only. This commit allows the cop to check for empty lines between class
and module definitions as well.

All three configs are enabled by default.

Closes #9064

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -463,6 +463,10 @@ Layout/EmptyLineBetweenDefs:
   StyleGuide: '#empty-lines-between-methods'
   Enabled: true
   VersionAdded: '0.49'
+  VersionChanged: <<next>>
+  EmptyLineBetweenMethodDefs: true
+  EmptyLineBetweenClassDefs: true
+  EmptyLineBetweenModuleDefs: true
   # If `true`, this parameter means that single line method definitions don't
   # need an empty line between them.
   AllowAdjacentOneLineDefs: false
--- a/lib/rubocop/cop/layout/empty_line_between_defs.rb
+++ b/lib/rubocop/cop/layout/empty_line_between_defs.rb
@@ -13,7 +13,8 @@ module RuboCop
       # `AllowAdjacentOneLineDefs` configures whether adjacent
       # one-line method definitions are considered an offense.
       #
-      # @example
+      # @example EmptyLineBetweenMethodDefs: true (default)
+      #   # checks for empty lines between method definitions.
       #
       #   # bad
       #   def a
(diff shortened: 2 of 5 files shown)


lib/rubocop/cop/layout/empty_line_between_defs.rb
```diff
@@ -13,6 +13,10 @@ module Layout
       # `AllowAdjacentOneLineDefs` configures whether adjacent
       # one-line method definitions are considered an offense.
       #
+      # `ApplyToClassDefs` checks for empty lines around class definitions.
```
bbatsov, line 16: I think those names and descriptions are misleading, as there is already a cop checking for empty lines around class bodies.

bbatsov: I mean mostly in the way that when enabled those settings should also cover different definition types as in the issue that was reported - e.g. you'll want space between a module definition and a method definition and so on. The documentation needs to be expanded and more examples should be added.

bbatsov:
Overall the change seems good to me, apart from my remarks about improving the docs. It'd be nice to also come up with better names for those options, as I don't like them much, but I can live with them.

config/default.yml
```diff
@@ -468,6 +468,8 @@ Layout/EmptyLineBetweenDefs:
   AllowAdjacentOneLineDefs: false
   # Can be array to specify minimum and maximum number of empty lines, e.g. [1, 2]
   NumberOfEmptyLines: 1
+  ApplyToClassDefs: false
```
bbatsov, line 471: Perhaps we can name this `EmptyLineBetweenClassDefs` instead? Probably it'd be best if there was also a configuration options for methods (`EmptyLineBetweenMethodDefs`) so this cop would get more flexible. 

I'd also enable those new options by default, as I doubt they'll cause a lot of offenses. The version changed should be updated to `<<next>>` as well.

bbatsov:
I've added a few more remarks, but now I feel confident in the direction I'd like us to take on this. :-)

tejasbubane:
@bbatsov changes done

bbatsov:
Looks good! Thanks!

jtannas:
Thanks @tejasbubane!
Sorry for not getting it myself in time :sweat_smile:

swiknaba:
I like the new cop! However, for cases such as:

```ruby
class CustomerFooError    < StandardError; end
class CustomerBarError    < StandardError; end
class CustomerFooBarError < StandardError; end
class CustomerOtherError  < StandardError; end
```

it would be amazing if we could have `EmptyLineBetweenClassDefs` as a string instead of a boolean, with values like:
* strict (applies always)
* skip_one_liners (always, except for one-line class definitions)
* disabled

What do you think?

bbatsov:
I think you're looking for `AllowAdjacentOneLineDefs`. The cop is not new, btw - it used to work only with method defs in the past, but this PR expanded its coverage.

PR #9079, opened 2020-11-22
Improve the gemspec to load only the necessary files without the git utility

piotrmurach:
This makes the gemspec independent of the git utility by replacing the
files attribute value with Ruby Dir for listing all the required files.
In addition, we exclude bin folder as these files are strictly development
concerns and can cause issues when distributed. The same goes for the
assets which are unnecessary for the distributed package. Both the README.txt
and LICENSE.txt are no longer mentioned as well as they are automatically
included in the generated file list via the extra_rdoc_files attribute.
This change reduces the overall package size from 440K down to 432K.

The version file is now loaded with the require_relative method to
increase performance and skip unnecessary lookup.

-----------------

* [x] Added tests.

### DIFF
--- a/rubocop.gemspec
+++ b/rubocop.gemspec
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
-$LOAD_PATH.unshift File.expand_path('lib', __dir__)
-require 'rubocop/version'
-require 'English'
+require_relative 'lib/rubocop/version'
 
 Gem::Specification.new do |s|
   s.name = 'rubocop'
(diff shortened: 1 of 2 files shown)


rubocop.gemspec
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
-$LOAD_PATH.unshift File.expand_path('lib', __dir__)
-require 'rubocop/version'
+require_relative 'lib/rubocop/version'
 require 'English'
```
bbatsov, line 4: I think now you can remove this, as it was needed for `$RS` that was removed.

piotrmurach: Good catch! Another nice side effect of the change.

bbatsov (approved):

bbatsov:
The failure is unrelated. Seems the `parallel` gem now requires Ruby 2.5, which is a bummer.

piotrmurach:
Thanks for merging! 

Looking at the commits in the `parallel` repository, the Ruby upgrade wasn't necessitated by any internal change. I would expect this to happen in `2.0` release. Hopefully, the maintainer can be persuaded to amend given the `rubocop` depends on it.

bbatsov:
I've started a conversation with him here https://github.com/grosser/parallel/commit/d191db712917a13c0beee3bc6b6c0a45b8181521

piotrmurach:
Sounds like you've reached a 'compromise'. It would be good for the dependency to still guarantee the same level of support as the upstream gem, otherwise, the `2.4` version would be a bit worse off even if bundler can reconcile appropriate version.

PR #9081, opened 2020-11-23
[Fix #7726] Handle method call chains inside square brackets

### LINKED ISSUE #7726: False positive for `MultilineMethodCallIndentation` inside `[]`
There is false positive for `MultilineMethodCallIndentation` inside `[]` (code as key).

(Probably related to #7338)

--------

## Expected behavior

Without offenses.

## Actual behavior

With offenses.

```yaml
# .rubocop.yml
Layout/MultilineMethodCallIndentation:
  EnforcedStyle: indented
```

```ruby
# test.rb

# frozen_string_literal: true

foo[
  bar
    .baz
]
```

```
Inspecting 1 file
C

Offenses:

test.rb:5:5: C: Layout/MultilineMethodCallIndentation: Align .baz with bar on line 4.
    .baz
    ^^^^

```

## RuboCop version

```
$ bundle exec rubocop -V
0.80.0 (using Parser 2.7.0.2, running on ruby 2.7.0 x86_64-linux)
```

jonas054:
Searching for the root of a method call chain should only consider method calls with a dot operator. We should stop if it's another kind of send node.

This is an improvement on a recent fix for the same cop. That commit only fixed the problem for the `[]=` operator. This solution is more comprehensive.

### DIFF
--- a/changelog/fix_MultilineMethodCallIndentation_inside_brackets.md
+++ b/changelog/fix_MultilineMethodCallIndentation_inside_brackets.md
@@ -1,0 +1,1 @@
+* [#7726](https://github.com/rubocop-hq/rubocop/issues/7726): Fix `MultilineMethodCallIndentation` indentation inside square brackets. ([@jonas054][])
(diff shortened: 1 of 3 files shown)


Drenmi (approved):
🇸🇪

koic:
Thanks!

PR #9085, opened 2020-11-23
[Fix #9083] Fix `Style/RedundantArgument` cop raising offense for more than one argument

### LINKED ISSUE #9083: `Style/RedundantArgument` report for cases like `foo.split(' ', 2)`
Seems `Style/RedundantArgument` didn't know about second argument [`limit`](https://ruby-doc.org/core-2.7.0/String.html#method-i-split) for `split` method

--------

1. Create simple `test.rb` script with content:
```
foo = 'a b c d e'
bar = foo.split(' ', 2)
puts(bar[0] == 'a')
puts(bar[1] == 'b c d e')
```
2. Run rubocop on it and get:
```
rubocop -A test.rb 
Inspecting 1 file
C

Offenses:

test.rb:2:7: C: Style/RedundantArgument: Argument ' ' is redundant because it is implied by default.
bar = foo.split(' ', 2)
      ^^^^^^^^^^^^^^^^^

```

Seems in that case this cop should just ignore it, if limit is set

## RuboCop version
```
rubocop -V
1.4.0 (using Parser 2.7.2.0, rubocop-ast 1.1.1, running on ruby 2.7.2 x86_64-linux)
  - rubocop-performance 1.9.0
  - rubocop-rspec 2.0.0

```

tejasbubane:
Also
* Add check for explicit receiver must be present.
* Refactor tests for the cop.

Closes #9083

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_argument.rb
+++ b/lib/rubocop/cop/style/redundant_argument.rb
@@ -40,6 +40,8 @@ module RuboCop
         MSG = 'Argument %<arg>s is redundant because it is implied by default.'
 
         def on_send(node)
+          return if node.receiver.nil?
+          return if node.arguments.count != 1
           return unless redundant_argument?(node)
 
           add_offense(node, message: format(MSG, arg: node.arguments.first.source))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Bug fixes
 
 * [#9082](https://github.com/rubocop-hq/rubocop/pull/9082): Fix gemspec to include assets directory. ([@javierav][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/redundant_argument.rb
```diff
@@ -40,6 +40,7 @@ class RedundantArgument < Cop
         MSG = 'Argument %<arg>s is redundant because it is implied by default.'
 
         def on_send(node)
+          return if node.arguments.count != 1
```
bbatsov, line 43: I think you should also return directly if there's no explicit receiver. I don't see a check for this currently.

tejasbubane: added

bbatsov:
Thanks!

PR #9091, opened 2020-11-24
Have `Naming/VariableNumber` accept _1, _2, ...

marcandre:
These are now special implicit parameters.


bbatsov:
Might also be considered a bug-fix, but I guess a change is fine as well. Thanks for fixing this!

PR #9093, opened 2020-11-25
[Fix #7670] Handle offenses inside heredocs for --disable-uncorrectable

### LINKED ISSUE #7670: infinite loop when using --disable-uncorrectable when offense in HEREDOC
## Expected behavior

Rubocop would add `# rubocop:todo Naming/VariableName` at top of HEREDOC and re-enable at end of HEREDOC. The same issue exists for %Q tags.

## Actual behavior

Excessive amount of `# rubocop:todo Naming/VariableName` are generated within the heredoc instead of on top of it and program crashes.

Given the following ruby:

```ruby
# frozen_string_literal: true

def our_function
  ourVariable = "foo"
  script = <<~JS
    <script>
      window.stuff = "#{ourVariable}"
    </script>
  JS
  puts(script)
end
```

When running: `rubocop --disable-uncorrectable --safe-auto-correct`, you get an infinite loop detected error, and the following code is generated:

```ruby
# frozen_string_literal: true

def our_function
  ourVariable = 'foo' # rubocop:todo Naming/VariableName
  script = <<~JS
    <script>
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # etc, etc
      window.stuff = "#{ourVariable}"
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # rubocop:todo Naming/VariableName
      # etc
```

## RuboCop version

0.79.0

jonas054:
The `# rubocop:todo` and `# rubocop:enable` must be placed outside the heredoc. Otherwise they won't have any effect and we will just add more and more comments until we crash.

### DIFF
--- a/lib/rubocop/cop/autocorrect_logic.rb
+++ b/lib/rubocop/cop/autocorrect_logic.rb
@@ -57,16 +57,11 @@ module RuboCop
         # The empty offense range is an edge case that can be reached from the Lint/Syntax cop.
         return nil if offense_range.empty?
 
-        @heredoc_ranges ||= begin
-          ranges = []
-          processed_source.ast.each_descendant do |node|
-            if node.respond_to?(:heredoc?) && node.heredoc?
-              ranges << node.loc.expression.join(node.loc.heredoc_end)
-            end
-          end
-          ranges
+        heredoc_nodes = processed_source.ast.each_descendant.select do |node|
+          node.respond_to?(:heredoc?) && node.heredoc?
         end
-        @heredoc_ranges.find { |range| range.contains?(offense_range) }
+        heredoc_nodes.map { |node| node.loc.expression.join(node.loc.heredoc_end) }
+                     .find { |range| range.contains?(offense_range) }
       end
 
       def range_of_first_line(range)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/autocorrect_logic.rb
```diff
@@ -39,16 +39,36 @@ def autocorrect_enabled?
...
+            disable_offense_before_and_after(range_by_lines(range))
+          end
         end
       end
 
+      def surrounding_heredoc(offense_range)
+        # The empty offense range is an edge case that can be reached from the Lint/Syntax cop.
+        return nil if offense_range.empty?
+
+        @heredoc_ranges ||= begin
```
jonas054, line 60: Maybe this is premature optimization. It caches the ranges based on the assumption that we come here many times for one cop's inspection of one file.

jonas054: I pushed a second commit, which I will squash if the reviewer agrees that it's better without the caching.

bbatsov:
The changes look solid to me. Thanks!

PR #9094, opened 2020-11-25
[Fix #8053] Fix an incorrect auto-correct for `Style/AndOr`

### LINKED ISSUE #8053: Auto-correct for Style/AndOr does not handle precedence change correctly
The auto-correct for Style/AndOr incorrectly replaces expressions like:

`a or b and c`

with

`a || b && c`

rather than

`(a || b) && c`

I wasn't able to find an open issue that relates to this issue.  If this behavior is unavoidable for some reason, then I recommend not marking Style/AndOr a "safe" auto-correctable cop.

--------

## Expected behavior

I expect it to correctly handle the change in precedence between `and` and `or` (which have the same precedence and `&&` and `||` (`&&` has higher precedence).

## Actual behavior

Running `rubocop --safe-auto-correct` on this code:

```
#!/usr/bin/ruby
# frozen_string_literal: true

t1 = (true or false and false)
print("#{t1}\n")
```
 results in:
```
#!/usr/bin/ruby
# frozen_string_literal: true

t1 = (true || false && false)
print("#{t1}\n")
```
The original code prints `false`, but the resulting code prints `true`.

See above.

## RuboCop version
```
vagrant@vagrant:/vagrant/mushroom-observer$ rubocop -V
0.83.0 (using Parser 2.7.1.2, running on ruby 2.6.6 x86_64-linux)
```
I originally discovered this using 0.75.1 at work and replicated on a local personal project that is running 0.83.0.

koic:
Fixes #8053.

This PR fixes an incorrect auto-correct for `Style/AndOr` when `or` precedes `and`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_an_incorrect_autocorrect_for_style_and_or.md
+++ b/changelog/fix_an_incorrect_autocorrect_for_style_and_or.md
@@ -1,0 +1,1 @@
+* [#8053](https://github.com/rubocop-hq/rubocop/issues/8053): Fix an incorrect auto-correct for `Style/AndOr` when `or` precedes `and`. ([@koic][])
--- a/lib/rubocop/cop/style/and_or.rb
+++ b/lib/rubocop/cop/style/and_or.rb
@@ -72,6 +72,8 @@ module RuboCop
             end
 
             corrector.replace(node.loc.operator, node.alternate_operator)
+
+            keep_operator_precedence(corrector, node)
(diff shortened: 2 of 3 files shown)


bbatsov:
Good work! I only wish I had seen this patch before cutting 1.4.2, but it is how it is. :-)

PR #9096, opened 2020-11-25
[Fix #9095] use merged_config instead of config for pending new cop check 

### LINKED ISSUE #9095: Respect NewCop value from merged_config
Rubocop `1.4.2` shows a warning about pending cops, even if the `NewCops: enabled` is set in a `required` config file.
This happens because the `possible_new_cops?` methods in the [ConfigLoader](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/config_loader.rb#L112) checks against `config` and not against `merged_config`.

#### Warning Text:
```
The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file.

Please also note that can also opt-in to new cops by default by adding this to your config:
  AllCops:
    NewCops: enable
```
--------

## Expected behavior
I expect that for the pending new cops check the merged configuration is considered and no warning is show when the `NewCops` setting is configured in a shared config file.

## Actual behavior
The `NewCops` setting is not considered and the warning about pending new cops is shown.

Use 2 configuration files. 
```yml
# File: .rubocop.yml
require:
- base

# NewCops is not set here
#
#  AllCops:
#    NewCops: enable
```

Base
```yml
# File: base.yml
AllCops:
  NewCops: enable
```

## RuboCop version

```
$ [bundle exec] rubocop -V
1.4.2 (using Parser 2.7.2.0, rubocop-ast 1.2.0, running on jruby 2.5.7 java)
  - rubocop-rspec 2.0.0
```

## Solution
A possible solution is, as described above in the intro, to use `merged_config` instead of `config` for the check in the [ConfigLoader](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/config_loader.rb#L112).

ThomasKoppensteiner:
Fix for issue #9095

### DIFF
--- a/changelog/fix_fix_9095_use_merged_config_instead_of.md
+++ b/changelog/fix_fix_9095_use_merged_config_instead_of.md
@@ -1,0 +1,1 @@
+* [#9096](https://github.com/rubocop-hq/rubocop/pull/9096): Fix #9095 use merged_config instead of config for pending new cop check. ([@ThomasKoppensteiner][https://github.com/ThomasKoppensteiner])
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -109,7 +109,7 @@ module RuboCop
         end
 
         merge_with_default(config, config_file).tap do |merged_config|
-          warn_on_pending_cops(merged_config.pending_cops) unless possible_new_cops?(config)
+          warn_on_pending_cops(merged_config.pending_cops) unless possible_new_cops?(merged_config)
         end
       end
 
--- a/spec/rubocop/config_loader_spec.rb
+++ b/spec/rubocop/config_loader_spec.rb
@@ -1443,6 +1443,51 @@ RSpec.describe RuboCop::ConfigLoader do
     end
   end
 
+  describe 'when pending cops exist', :isolated_environment do
+    subject(:from_file) do
+      described_class.configuration_from_file('.rubocop.yml')
+    end
+
+    before do
+      create_empty_file('.rubocop.yml')
+
+      # Setup similar to https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/rspec/inject.rb#L16
+      # and https://github.com/runtastic/rt_rubocop_defaults/blob/master/lib/rt_rubocop_defaults/inject.rb#L21
+      config = RuboCop::Config.new(parent_config)
+      described_class.instance_variable_set(:@default_configuration, config)
+    end
+
+    context 'when NewCops is set in a required file' do
+      let(:parent_config) do
+        {
+          'AllCops' => {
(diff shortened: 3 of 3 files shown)


bbatsov:
Did you manage to figure out when the regression was introduced? Was this in 1.3? 

Also - ideally you should add some spec illustrating the impact of the fix.

ThomasKoppensteiner:
@bbatsov the code was last changed within [this commit](https://github.com/rubocop-hq/rubocop/commit/d3f6b097a8fc65a64d050b5844a1a4fe2505d503), but before it also used `config`.
I think the issue is more related to the introduction of the `NewCops` option and how it disables the warning.

ThomasKoppensteiner:
For me it looks like the issue is related to the monkey patching which happens in [rubocop-rspec](https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/rspec/inject.rb#L16) or [rt_rubocop_defaults](https://github.com/runtastic/rt_rubocop_defaults/blob/master/lib/rt_rubocop_defaults/inject.rb#L21). At least I couldn't reproduce it in otherwise.

ThomasKoppensteiner:
Btw. is [this override](https://github.com/rubocop-hq/rubocop/blame/master/spec/support/suppress_pending_warning.rb) still needed, when `Rainbow.enabled = false` is set in the `spec_helper.rb`?

ThomasKoppensteiner:
The following spec fails for me on `master` as well `bundle exec rspec spec/rubocop/cop/style/redundant_regexp_escape_spec.rb:523`.

ThomasKoppensteiner:
@bbatsov and @koic anything else I can do here?

koic:
> The following spec fails for me on master as well bundle exec rspec spec/rubocop/cop/style/redundant_regexp_escape_spec.rb:523.

This failure has been resolved by https://github.com/rubocop-hq/rubocop/pull/9102. Can you rebase with the latest master branch?

bbatsov:
Seems to me something went wrong with the rebase, as a lot of commits on master appeared in the diff for this branch.

bbatsov:
Also - don't forget to add a bug-fix entry in the `changelog` folder.

ThomasKoppensteiner:
> Seems to me something went wrong with the rebase, as a lot of commits on master appeared in the diff for this branch.

Thank you for pointing this out. Did it again, now with command line only.

bbatsov:
Thanks!

ThomasKoppensteiner:
@bbatsov do you already know, when this will be released? What is the typical release cycle of rubocop?

bbatsov:
I plan to do the next release today or tomorrow. Typically I'm aiming for one small release each week/couple of weeks, but I don't pursue any particular release cadence.

PR #9098, opened 2020-11-25
Unification of `IgnoredMethods`

dvandersluis:
As mentioned by @bbatsov at https://github.com/rubocop-hq/rubocop/pull/9058#issuecomment-729918462

1. The two cops that were configured using `ExcludedMethods` instead of `IgnoredMethods` (`Metrics/BlockLength` and `Metrics/MethodLength`) have been changed to use `IgnoredMethods` going forward. I changed the `IgnoredMethods` mixin so that it can be configured to allow for those two cops to continue accepting `ExcludedMethods` without adding that config key to the reset of the cops.
2. All cops that use the `IgnoredMethods` mixin now can accept an array of both strings and regexes, and both are applied.

Questions:
* ~Is there a good way to document this so that users will know they can use regexes for `IgnoredMethods` everywhere now?~ I added a new section to configuration.adoc.
* Is there a way to add a deprecation warning when running rubocop with `ExcludedMethods` in the configuration so that users can update their config?

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/metrics/block_length.rb
+++ b/lib/rubocop/cop/metrics/block_length.rb
@@ -44,7 +44,8 @@ module RuboCop
         LABEL = 'Block'
 
         def on_block(node)
-          return if excluded_method?(node)
+          return if ignored_method?(node.method_name)
+          return if method_receiver_excluded?(node)
           return if node.class_constructor? || node.struct_constructor?
 
           check_code_length(node)
@@ -52,11 +53,13 @@ module RuboCop
 
         private
 
-        def excluded_method?(node)
+        def method_receiver_excluded?(node)
           node_receiver = node.receiver&.source&.gsub(/\s+/, '')
           node_method = String(node.method_name)
 
           ignored_methods.any? do |config|
+            next unless config.is_a?(String)
+
             receiver, method = config.split('.')
 
             unless method
--- a/lib/rubocop/cop/metrics/block_length.rb
+++ b/lib/rubocop/cop/metrics/block_length.rb
@@ -12,6 +12,10 @@ module RuboCop
       # Available are: 'array', 'hash', and 'heredoc'. Each literal
       # will be counted as one line regardless of its actual size.
       #
+      #
+      # NOTE: The `ExcludedMethods` configuration is deprecated and only kept
+      # for backwards compatibility. Please use `IgnoredMethods` instead.
+      #
       # @example CountAsOne: ['array', 'heredoc']
       #
       #   something do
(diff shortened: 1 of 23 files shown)


lib/rubocop/cop/metrics/block_length.rb
```diff
@@ -33,23 +33,29 @@ module Metrics
...
-        def excluded_method?(node)
+        def method_receiver_excluded?(node)
           node_receiver = node.receiver&.source&.gsub(/\s+/, '')
           node_method = String(node.method_name)
 
-          excluded_methods.any? do |config|
+          ignored_methods.any? do |config|
+            next unless config.is_a?(String)
+
             receiver, method = config.split('.')
```
dvandersluis, line 59: `Metrics/BlockLength` has a special case for `IgnoredMethods`, where it allows for values to be fully qualified methods, and it uses this to test receivers. I kept this logic of course, but this is now an exception from how all the other `IgnoredMethods` cops work.

bbatsov:
> Is there a good way to document this so that users will know they can use regexes for IgnoredMethods everywhere now?

Can't think of anything other than something in the "Configuration" part of the docs. 

> Is there a way to add a deprecation warning when running rubocop with ExcludedMethods in the configuration so that users can update their config?

Yes - look for `ConfigObsoletion`.

dvandersluis:
> Yes - look for ConfigObsoletion.

Looks like `OBSOLETE_PARAMETERS` from `ConfigObsoletion` gives an error instead of allowing it, which makes my configuration not very useful since rubocop won't run anyways. What do you suggest @bbatsov?

```
$ rubocop
Error: obsolete parameter ExcludedMethods (for Metrics/BlockLength) found in .rubocop.yml
`ExcludedMethods` has been renamed to `IgnoredMethods`.
obsolete parameter ExcludedMethods (for Metrics/MethodLength) found in .rubocop.yml
`ExcludedMethods` has been renamed to `IgnoredMethods`.
```

dvandersluis:
`spec/rubocop/config_loader_spec.rb:740` is failing in CI but not via `bundle exec rake default`. Anyone have any ideas?

bbatsov:
I'm assuming some dep got updated on the CI. 

> which makes my configuration to still be able to use ExcludedMethods not very useful since rubocop won't run anyways. What do you suggest @bbatsov?

Damn, I had forgotten that in the past we already raised error. I guess we should make the code there a bit more flexible and make it show errors or warnings.

dvandersluis:
> Damn, I had forgotten that in the past we already raised error. I guess we should make the code there a bit more flexible and make it show errors or warnings.

I'll take a look at it!

dvandersluis:
@bbatsov I think everything has been taken care of now. I've added to the documentation and the deprecation message is now set up.

dvandersluis:
One other thing -- I needed to keep `ExcludedMethods` in `config/default.yml` or rubocop will complain about unsupported parameters. However, this will add it to the doc for the cop. I added a note for both cops to explain why there's both, hopefully that will be sufficient.

bbatsov:
Fantastic work! 🙇‍♂️

PR #9101, opened 2020-11-25
Add new `Style/UseNamedBlockArgumentForSimpleMethodCalls` cop

dabroz:
Add new `Style/UseNamedBlockArgumentForSimpleMethodCalls` cop

I wanted `array.map { _1.name }` to be corrected to `array.map(&:name)` but instead of adding more stuff to `SymbolProc` it seemed better to add a new cop instead that could be later picked up with `SymbolProc`.

-----------------

* [x] Added tests.


dabroz:
`RuboCop::Cop::Style::RedundantRegexpEscape with multibyte characters removes the escape character at the right position` seems to be broken on `master`.

marcandre:
I think this is a promising cop, but it should really go both ways (e.g. if there's only a single param, and the block is single line, then use `_1` instead).

Independently, I think `SymbolProc` should be fixed to catch `{ _1.foo }`

bbatsov:
Totally agree with @marcandre.

config/default.yml
```diff
@@ -4554,6 +4554,11 @@ Style/UnpackFirst:
   Enabled: true
   VersionAdded: '0.54'
 
+Style/UseNamedBlockArgumentForSimpleMethodCalls:
+  Description: "{ _1.method } should not be used because it won't be picked up by SymbolProc cop."
```
bbatsov, line 4558: Btw, that's no longer true, after a recent change. :-)

bbatsov:
It seems to me a better rationale for this cop would be to stick to their explicit or implicit params, with the option to apply this only to one-line blocks or to all blocks.

bbatsov:
I think @dvandersluis already fixed this one in `SymbolProc`.

PR #9112, opened 2020-11-27
Add new cop `Lint/UnexpectedBlockArity`

dvandersluis:
Follows https://github.com/rubocop-hq/rubocop/pull/9071#issuecomment-730946877:

>  Maybe it should be a different cop that checks the block arity for methods that are known to have arity 2. There are less of them now that `Hash#each` is being changed to arity 1. Maybe just  `Enumerable#each_with_object`, `min`, `max`, `minmax`, `slice_when`, `sort`? Maybe we even have such a cop already?

This new cop checks for method blocks where the given block args don't match what the method is known to need. It is configurable using `Methods`, which is preloaded with all the `Enumerable` methods that take 2 args, as per the above comment.

-----------------

* [x] Added tests.


bbatsov:
Overall the cop looks great. My only suggestion would be to ignore method calls without receivers, as those are going to be more likely to be false positives.

dvandersluis:
@bbatsov makes sense! I've updated it.

bbatsov:
Thanks!

inkstak:
Would it be possible to add a deny list using regexp ?

We have a lot of `sort` blocks in jbuilder templates which are json properties.

```
json.sort do
  json.order       @current_order
  json.direction @current_order_direction
end
```

dvandersluis:
@inkstak configuring the cop like this probably would solve this for you:

```yaml
Lint/UnexpectedBlockArity:
  Exclude:
    - '**/*.jbuilder'
```

inkstak:
@dvandersluis Yes this is my current workaround, but it's too bad that this helpful cop doesn't cover regular `Enum#sort` methods present in our templates.

dvandersluis:
There isn't really a way for rubocop to know if `sort` is `Enumerable#sort` or another sort. Generally when cops exclude methods it's done across the board, not based on receiver (because in general we can't tell what the receiver is).

marcandre:
(until we plug into `rbs`...)

inkstak:
I actually suspect we can't tell what type is the receiver, but isn't it impossible to exclude cases using the name of the receiver or the whole line ? I humbly ask, the subtleties of rubocop parser are still beyonds me.

```
Lint/UnexpectedBlockArity:
  Allowlist:
    - 'json.sort'
    - /\Ajson\.sort do\Z/
```

PR #9113, opened 2020-11-28
Fix a false positive for `Style/MethodCallWithoutArgsParentheses`

koic:
This PR fixes the following false positive for `Style/MethodCallWithoutArgsParentheses` when assigning to a default argument with the same name.

```console
% cat example.rb
def foo(test = test())
end

% ruby -c example.rb
Syntax OK

% rubocop --only Style/MethodCallWithoutArgsParentheses -a example.rb
Inspecting 1 file
C

Offenses:

example.rb:1:20: C: [Corrected] Style/MethodCallWithoutArgsParentheses:
Do not use parentheses for method calls with no arguments.
def foo(test = test())
                   ^^

% cat example.rb
def foo(test = test)
end

% ruby -c example.rb
example.rb:1: circular argument reference - test
```

It prevents `example.rb:1: circular argument reference - test`.

-----------------

* [x] Added tests.


bbatsov:
Nice catch!

PR #9118, opened 2020-11-29
Fix Layout/ClassStructure public_class_methods ordering

timlkelly:
The `Layout/ClassStructure` rule does not enforce the order of `public_class_methods`. Given the default ordering, I would expect public class methods to appear before the initializer. 

Expected:
```
# person.rb
class Person
  def self.some_method; end
  def initialize; end
end
```

However, when in the incorrect order, it does not raise any offenses:

```
# person.rb
class Person
  def initialize; end
  def self.some_method; end
end
```

```
❯ rubocop person.rb --only Layout/ClassStructure
warning: parser/current is loading parser/ruby27, which recognizes
warning: 2.7.2-compliant syntax, but you are running 2.7.1.
Inspecting 1 file
.

```

```
❯ rubocop -V
warning: parser/current is loading parser/ruby27, which recognizes
warning: 2.7.2-compliant syntax, but you are running 2.7.1.
1.4.2 (using Parser 2.7.2.0, rubocop-ast 1.2.0, running on ruby 2.7.1 x86_64-darwin19)
```

---

The problem is that `Rubocop::Cop::Layout::ClassStructure#walk_over_nested_class_definition` is ignoring class method nodes because `#classify` returns `class_methods` while the configuration's `ExpectedOrder` provides `public_class_methods`.

I thought it would be easier to update the value in `HUMANIZED_NODE_TYPE` instead of changing the configuration to match, as that would require users to update their `ExpectedOrder` configurations.

With that said, this now catches several `Layout/ClassStructure` offenses in the codebase. Here is the output from `bundle exec rake default`

```
Offenses:

lib/rubocop/config.rb:36:5: C: Layout/ClassStructure: public_class_methods is supposed to appear before initializer.
    def self.create(hash, path) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/rubocop/cop/base.rb:95:7: C: Layout/ClassStructure: public_class_methods is supposed to appear before initializer.
      def self.joining_forces; end
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/rubocop/cop/base.rb:330:7: C: Layout/ClassStructure: public_class_methods is supposed to appear before private_methods.
      def self.builtin? ...
      ^^^^^^^^^^^^^^^^^
lib/rubocop/cop/corrector.rb:71:7: C: Layout/ClassStructure: public_class_methods is supposed to appear before initializer.
      def self.source_buffer(source) ...
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/rubocop/cop/metrics/utils/abc_size_calculator.rb:32:11: C: Layout/ClassStructure: constants is supposed to appear before public_class_methods.
          ARGUMENT_TYPES = %i[arg optarg restarg kwarg kwoptarg kwrestarg blockarg].freeze
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/rubocop/cop/registry.rb:213:7: C: Layout/ClassStructure: public_class_methods is supposed to appear before initializer.
      def self.all ...
      ^^^^^^^^^^^^
lib/rubocop/cop/team.rb:28:7: C: Layout/ClassStructure: public_class_methods is supposed to appear before initializer.
      def self.new(cop_or_classes, config, options = {}) ...
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/rubocop/target_ruby.rb:197:5: C: Layout/ClassStructure: constants is supposed to appear before public_class_methods.
    SOURCES = [RuboCopConfig, RubyVersionFile, BundlerLockFile, GemspecFile, Default].freeze
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tasks/changelog.rb:122:3: C: Layout/ClassStructure: public_class_methods is supposed to appear before initializer.
  def self.pending? ...
  ^^^^^^^^^^^^^^^^^

1205 files inspected, 9 offenses detected, 9 offenses auto-correctable
RuboCop failed!
```

Some of these are easy changes, like [tasks/changelog.rb:122:3](https://github.com/rubocop-hq/rubocop/blob/master/tasks/changelog.rb#L122)

While others seem a little more complex, for instance [lib/rubocop/cop/base.rb:330:7](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/base.rb#L330-L336). This appears to be declaring a private class method, however, this node does not seem to know that it should be private. From my understanding, this rule does not provide guidance on where private class methods should be positioned in the class. So I did not want to make any further changes without a discussion of what the expected behavior should be.

Separately, I noticed that the `ExpectedOrder` in the `.rubocop.yml` file for this project lists `instance_methods`, which I don't think corresponds to anything and should be changed to `public_methods` to match the value in `HUMANIZED_NODE_TYPE`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_layout_class_structure_class_methods.md
+++ b/changelog/fix_layout_class_structure_class_methods.md
@@ -1,0 +1,1 @@
+* [#9118](https://github.com/rubocop-hq/rubocop/pull/9118): Fix Layout/ClassStructure public_class_methods ordering. ([@timlkelly][])
--- a/lib/rubocop/cop/layout/class_structure.rb
+++ b/lib/rubocop/cop/layout/class_structure.rb
@@ -139,7 +139,7 @@ module RuboCop
 
         HUMANIZED_NODE_TYPE = {
           casgn: :constants,
-          defs: :class_methods,
+          defs: :public_class_methods,
           def: :public_methods,
           sclass: :class_singleton
         }.freeze
--- a/spec/rubocop/cop/layout/class_structure_spec.rb
+++ b/spec/rubocop/cop/layout/class_structure_spec.rb
@@ -325,4 +325,16 @@ RSpec.describe RuboCop::Cop::Layout::ClassStructure, :config do
       end
     RUBY
   end
+
+  context 'with initializer before public_class_methods' do
+    it 'creates an offense' do
+      expect_offense <<~RUBY
+        class Person
+          def initialize; end
+          def self.some_method; end
+          ^^^^^^^^^^^^^^^^^^^^^^^^^ `public_class_methods` is supposed to appear before `initializer`.
+        end
+      RUBY
+    end
+  end
 end


bbatsov:
> I thought it would be easier to update the value in HUMANIZED_NODE_TYPE instead of changing the configuration to match, as that would require users to update their ExpectedOrder configurations.

Yeah, this seems reasonable/less painful. 

>  From my understanding, this rule does not provide guidance on where private class methods should be positioned in the class. So I did not want to make any further changes without a discussion of what the expected behavior should be.

I've rarely used private class methods, so I never thought much on the subject. Probably it makes sense for them to appear after public class methods. @rubocop-hq/rubocop-core Your input here would be appreciated.

marcandre:
> where private class methods should appear

Are we talking about a style like `private_class_method def self.foo ...`, and not `class << self; private def foo`?

I don't personally agree that class methods should precede instance methods... Also, I think that RuboCop's codebase usually doesn't use many class methods, but currently they are often mixed in a more thematic way. For example `Cop::Base` has many class methods interspersed with instance methods.

So it looks like what we currently have is more like:

```
  ExpectedOrder:
      - module_inclusion
      - constants
      - initializer
      - [public_methods, public_class_methods]
      - protected_methods
      - private_methods
```

Should we support that type of config?

timlkelly:
> Are we talking about a style like `private_class_method def self.foo ...`, and not `class << self; private def foo`?

Honestly, I've never found myself writing a private class method, in either format. I suspected that given the `public_class_methods` value in this cop, there would be a `private_` variation. 

I don't have a strong argument for either, or for the ordering of them in the class itself. Allowing for mixed placement of instance and class methods does give more flexibility for people to tune this rule to work for their projects. I suppose the next step is to decide on the standard/default, be it either strict ordering or the suggested mixed option, so that this rule can be used effectively. I am interested in continuing to work on this rule to address the issue with class methods, however, I'd want to define the desired outcome before starting on any work.

> Any maybe `singleton_class` could be distinguished from `public_class_methods` for those of us that like it (✋)

Is there an example of this in the Rubocop codebase? I believe I know what you're referring to, but it's something I don't often use.

marcandre:
I am doing a heavy refactor of Layout/ClassStructure in #9575, best not to work on this for now...

timlkelly:
@marcandre given #9575 is open, would it be best to close this pull request? It seems like the issue is being tracked in #9738.

bbatsov:
Yeah, I guess we can close this PR.

PR #9130, opened 2020-11-30
[Fix #9122] Added notification if any gems are loaded that have rubocop extensions.

### LINKED ISSUE #9122: Nudge about standard extensions

Although it has been long time since we [extracted `rubocop-rails` from core](https://github.com/rubocop-hq/rubocop/issues/5976) some users never discover our extensions. I have seen quite a few modern Rails projects using only `rubocop`.

I would like when `rubocop` is run, it recognizes and prints a log message suggesting standard plugins like `rubocop-rails`, `rspec`, `minitest`, `rake`, `sequel` - whichever applicable to the project. 

An option to disable this message.

A generic log message suggesting to have a look at our extensions.

dvandersluis:
The notification is output at the end, and only when using a few specified formatters that make sense to me. It is also disabled for CI and if some rubocop options are set that I didn't think it made sense to be included for (`--only`, `--debug`, `--list-target-files` and `--out` (since this is possibly used for something programmatic, I didn't want to change it).

The message looks like this:

```
Notice: Based on detected gems, the following RuboCop extension libraries might be helpful:
  * rubocop-rake (http://github.com/rubocop-hq/rubocop-rake)

This notification can be disabled by adding this to your config:
  AllCops:
    NotifyForExtensions: false
```

I also wanted to have a way to disable it so that it doesn't end up being spammy to users who aren't interested, so it is configured by `AllCops/NotifyForExtensions` (which can either take a hash of extensions and gems, or `false`). I did it this way so that someone could disable the notification for just `rubocop-rake` for instance, if they're never planning on using it.

I would love to discuss any choices I made and make any changes!

-----------------

* [ ] Added tests.


bbatsov:
I love the proposed design and I might even ship this with the release I've planned for later today, as I don't really have any concerns about it. It's actually better than what I had in mind, as originally I was thinking of mapping extensions to the presence of certain files/folders in the project. You came up with a granular and flexible configuration and I agree that something like this doesn't make sense for every formatter. 

I'm assuming in the unlikely event of a project with only a `.gemspec` we might run into some issues, but that's not a big deal. As far as I'm concerned all we need is some documentation is this is pure gold. Thanks! 🙇

lib/rubocop/cli/command/notify_for_extensions.rb
```diff
@@ -0,0 +1,67 @@
...
+        # Combination of short and long formatter names.
+        INCLUDED_FORMATTERS = %w[p progress fu fuubar pa pacman].freeze
+
+        self.command_name = :notify_for_extensions
+
+        def run
+          return if skip? || extensions.none?
+
+          puts
+          puts 'Notice: Based on gems detected by Rubocop, the following '\
```
bbatsov, line 18: Rubocop -> RuboCop ;-)

dvandersluis: Fixed 🤦

dvandersluis:
@bbatsov thanks! My biggest worry is how to make this be useful without being spammy.

I'll write up some documentation for it in the morning, thanks for the reminder!

config/default.yml
```diff
@@ -139,6 +139,16 @@ AllCops:
   # from the lock file.) If the Ruby version is still unresolved, RuboCop will
   # use the oldest officially supported Ruby version (currently Ruby 2.4).
   TargetRubyVersion: ~
+  # Determines if a notification for extension libraries should be shown when
+  # rubocop is run. Keys are the name of the extension, and values are an array
+  # of gems in the Gemfile that the extension is suggested for, if not already
+  # included.
+  NotifyForExtensions:
```
bbatsov, line 146: I'm also wondering if `SuggestExtensions` is not better terminology.

dvandersluis: Works for me!

lib/rubocop/cli/command/notify_for_extensions.rb
```diff
@@ -0,0 +1,67 @@
...
+          puts
+          puts 'Notice: Based on detected gems, the following '\
+            'RuboCop extension libraries might be helpful:'
+
+          extensions.each do |extension|
+            puts "  * #{extension} (http://github.com/rubocop-hq/#{extension})"
+          end
+
+          puts
+          puts 'You can opt out of this notification by adding the following to your config:'
```
bbatsov, line 26: I think notification sounds a bit too formal and we can just go with `message` here.

lib/rubocop/cli/command/notify_for_extensions.rb
```diff
@@ -0,0 +1,67 @@
...
+        # Combination of short and long formatter names.
+        INCLUDED_FORMATTERS = %w[p progress fu fuubar pa pacman].freeze
+
+        self.command_name = :notify_for_extensions
+
+        def run
+          return if skip? || extensions.none?
+
+          puts
+          puts 'Notice: Based on detected gems, the following '\
```
bbatsov, line 18: I'm pondering if `Tip` is not a better fit than `Notice`. Not a big deal, though.

bbatsov: Or "Suggestion".

dvandersluis:
@bbatsov oh I was writing tests and documentation 😅 I'll open another PR

PR #9131, opened 2020-12-01
[Fix #9128] Fix an incorrect auto-correct for `Style/ClassAndModuleChildren`

### LINKED ISSUE #9128: Style/ClassAndModuleChildren cop incorrectly corrects parent `class` to `module`
RuboCop incorrectly expands compact style class reopenings to use `module` for the parent if the parent is a `Class`.

--------

## Expected behavior

Describe here how you expected RuboCop to behave in this particular situation.

### Input

```ruby
# frozen_string_literal: true

# Foo class
class Foo
  # Foo::Bar class
  class Bar
    # Foo::Bar#baz method
    def baz
      'baz'
    end
  end
end

class Foo::Bar
  # Foo::Bar#quux method
  def quux
    'quux'
  end
end
```

### Expected output

```ruby
# frozen_string_literal: true

# Foo class
class Foo
  # Foo::Bar class
  class Bar
    # Foo::Bar#baz method
    def baz
      'baz'
    end
  end
end

class Foo
  class Bar
    # Foo::Bar#quux method
    def quux
      'quux'
    end
  end
end
```

## Actual behavior

### Input

```ruby
# frozen_string_literal: true

# Foo class
class Foo
  # Foo::Bar class
  class Bar
    # Foo::Bar#baz method
    def baz
      'baz'
    end
  end
end

class Foo::Bar
  # Foo::Bar#quux method
  def quux
    'quux'
  end
end
```

### Expected output

```ruby
# frozen_string_literal: true

# Foo class
class Foo
  # Foo::Bar class
  class Bar
    # Foo::Bar#baz method
    def baz
      'baz'
    end
  end
end

module Foo
  class Bar
    # Foo::Bar#quux method
    def quux
      'quux'
    end
  end
end
```

### Executing the bad correction code

```console
$ cat rubocop_compact_autocorrect_bad.rb
# frozen_string_literal: true

# Foo class
class Foo
  # Foo::Bar class
  class Bar
    # Foo::Bar#baz method
    def baz
      'baz'
    end
  end
end

module Foo
  class Bar
    # Foo::Bar#quux method
    def quux
      'quux'
    end
  end
end
$ ruby rubocop_compact_autocorrect_bad.rb
rubocop_compact_autocorrect_bad.rb:14:in `<main>': Foo is not a module (TypeError)
```

bundle exec rubocop -A [source]

## RuboCop version

```
$ bundle exec rubocop -V
1.3.1 (using Parser 2.7.2.0, rubocop-ast 1.1.1, running on ruby 2.6.6 x86_64-darwin20)
```

koic:
Fixes #9128.

This PR fixes an incorrect auto-correct for `Style/ClassAndModuleChildren` when namespace is defined as a class in the same file.

It is still undetectable when class definition exists in the different file, but it can be detected in the same file. This PR solves the latter.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_incorrect_autocorrect_for_class_and_module_children.md
+++ b/changelog/fix_incorrect_autocorrect_for_class_and_module_children.md
@@ -1,0 +1,1 @@
+* [#9128](https://github.com/rubocop-hq/rubocop/issues/9128): Fix an incorrect auto-correct for `Style/ClassAndModuleChildren` when namespace is defined as a class in the same file. ([@koic][])
--- a/lib/rubocop/cop/style/class_and_module_children.rb
+++ b/lib/rubocop/cop/style/class_and_module_children.rb
@@ -55,13 +55,18 @@ module RuboCop
           padding = ((' ' * indent_width) + leading_spaces(node)).to_s
           padding_for_trailing_end = padding.sub(' ' * node.loc.end.column, '')
 
-          replace_keyword_with_module(corrector, node)
+          replace_namespace_keyword(corrector, node)
           split_on_double_colon(corrector, node, padding)
           add_trailing_end(corrector, node, padding_for_trailing_end)
         end
 
-        def replace_keyword_with_module(corrector, node)
-          corrector.replace(node.loc.keyword, 'module')
+        def replace_namespace_keyword(corrector, node)
+          class_definition = node.left_sibling&.each_node(:class)&.find do |class_node|
+            class_node.identifier == node.identifier.namespace
+          end
+          namespace_keyword = class_definition ? 'class' : 'module'
+
+          corrector.replace(node.loc.keyword, namespace_keyword)
         end
 
         def split_on_double_colon(corrector, node, padding)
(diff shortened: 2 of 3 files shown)


bbatsov:
Nice fix! 🙇

lopopolo:
Why does this autocorrect in cases where correctness cannot be verified?

siegfault:
@lopopolo there may be more context in https://github.com/rubocop-hq/rubocop/pull/4900, but while the cop is enabled, the autocorrection for the cop is marked as unsafe (https://github.com/rubocop-hq/rubocop/blob/afcbb6a077c82fbf6ddd62889fdbf671312e73f8/config/default.yml#L2812-L2840). My understanding is that without any config change, that does not autocorrect unless someone passes `--auto-correct-all`. I think this rules falls into the third bucket (

> Some automatic corrections might change (slightly) the semantics of the code, meaning they’d produce code that’s mostly equivalent to the original code, but not 100% equivalent. We call such auto-correct behaviour "unsafe".

) of https://docs.rubocop.org/rubocop/usage/auto_correct.html. Hope that answers it, but let me know if I misunderstood!

marcandre:
What @siegfault said :-)

I would add that in this case, the code is *never truly equivalent* as it changes constant lookup.

```ruby
X = 1
module Foo
  X = 2
  module Bar
    def self.info
      X
     end
  end
end
module Foo::Bar
  def self.info2
    X
  end
end
Foo::Bar.info # => 2
Foo::Bar.info2 # => 1
```

PR #9138, opened 2020-12-01
[Fix #9080] Make `Lint/ShadowingOuterVariable` aware of `Ractor`

### LINKED ISSUE #9080: `Lint/ShadowingOuterLocalVariable` and Ractor
It would be nice if `Lint/ShadowingOuterLocalVariable` was aware of `Ractor.new`:

```ruby
def foo(*args)
  Ractor.new(*args) do |*args| # => Shadowing outer local variable - args.
    # ...
  end
end
```

This is a pattern that should be *encouraged*, as the block given to `Ractor.new` should not access outer variables, so shadowing the outer local variables seems to me to be a nice and natural way to go, no?

tejasbubane:
Block given to `Ractor.new` should not access outer variables, so shadowing the outer local variables should be encouraged.

Closes #9080 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/shadowing_outer_local_variable.rb
+++ b/lib/rubocop/cop/lint/shadowing_outer_local_variable.rb
@@ -8,6 +8,14 @@ module RuboCop
       # given by `ruby -cw` prior to Ruby 2.6:
       # "shadowing outer local variable - foo".
       #
+      # NOTE: Shadowing of variables in block passed to `Ractor.new` is allowed
+      # because `Ractor` should not access outer variables.
+      # eg. following syle is encouraged:
+      #
+      #   worker_id, pipe = env
+      #   Ractor.new(worker_id, pipe) do |worker_id, pipe|
+      #   end
+      #
       # @example
       #
       #   # bad
@@ -34,12 +42,17 @@ module RuboCop
       class ShadowingOuterLocalVariable < Base
         MSG = 'Shadowing outer local variable - `%<variable>s`.'
 
+        def_node_matcher :ractor_block?, <<~PATTERN
+          (block (send (const nil? :Ractor) :new ...) ...)
+        PATTERN
+
         def self.joining_forces
           VariableForce
         end
 
         def before_declaring_variable(variable, variable_table)
           return if variable.should_be_unused?
+          return if ractor_block?(variable.scope.node)
 
           outer_local_variable = variable_table.find_variable(variable.name)
           return unless outer_local_variable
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/shadowing_outer_local_variable.rb
```diff
@@ -34,12 +37,17 @@ module Lint
       class ShadowingOuterLocalVariable < Base
         MSG = 'Shadowing outer local variable - `%<variable>s`.'
 
+        def_node_matcher :ractor_block?, <<~PATTERN
+          (block (send (const nil? :Ractor) :new ...) ...)
```
marcandre, line 41: Could you use `#global_const?(:Ractor)` instead?

tejasbubane: Thanks. For some reason `global_const` method is not found. Am I missing something?

lib/rubocop/cop/lint/shadowing_outer_local_variable.rb
```diff
@@ -8,6 +8,9 @@ module Lint
       # given by `ruby -cw` prior to Ruby 2.6:
       # "shadowing outer local variable - foo".
       #
+      # Note: Shadowing of variables in block passed to `Ractor.new` is allowed
```
bbatsov, line 11: `Note:` -> `NOTE:`, so AsciiDoc would interpret it properly.

bbatsov: I'm guessing you can also add some example with `Ractor.new` and extend the note saying this applies only to Ruby 3+.

bbatsov:
The PR seems fine to me, but I'm not familiar with `Ractor`, so I'll leave it to @marcandre to approve/merge it.

lib/rubocop/cop/lint/shadowing_outer_local_variable.rb
marcandre: Oh, right, it's a method of `Node` only, but since the receiver could be `nil` you'd have to write `[!nil? global_const?(:Ractor)]`. Let's merge it as is and I'll figure something out.

lib/rubocop/cop/lint/shadowing_outer_local_variable.rb
tejasbubane: Added

lib/rubocop/cop/lint/shadowing_outer_local_variable.rb
tejasbubane: @marcandre Changed back to `const nil?`

marcandre: Cool, thanks. I created https://github.com/rubocop-hq/rubocop-ast/issues/160

PR #9139, opened 2020-12-01
[Fix #8985] Fix `Style/StringConcatenation` autocorrect generating invalid ruby

### LINKED ISSUE #8985: Style/StringConcatenation autocorrect doesn't properly escape double quotes when concatenated via single quotes
## Expected Behavior

When Rubocop encounters `'"' + "foo" + '"'` it should be autocorrected to `'"foo"'`

## Actual Behavior 
When Rubocop 1.1.0 encounters `'"' + "foo" + '"'` it currently autocorrects to `""foo""` thus leading to a syntax error:

```
test.rb:3:3: C: [Corrected] Style/StringConcatenation: Prefer string interpolation to string concatenation.
p '"' + "foo" + '"'
  ^^^^^^^^^^^^^^^^^
test.rb:3:5: E: Lint/Syntax: unexpected token tIDENTIFIER
p ""foo""
```

Create a Ruby file with the following line and run `rubocop -A` in its directory
`'"' + "foo" + '"'` 

## RuboCop version

```
$ rubocop -V
1.1.0 (using Parser 2.7.2.0, rubocop-ast 1.1.0, running on ruby 2.7.2 x86_64-linux)
```
## Note

This is a bug specific to double quotes in single quotes. `"\"" + 'foo' + "\""` is properly autocorrected to `'"foo"'`.

tejasbubane:
When single quoted strings

Closes #8985 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/string_concatenation.rb
+++ b/lib/rubocop/cop/style/string_concatenation.rb
@@ -106,7 +106,13 @@ module RuboCop
               end
             end
 
-          "\"#{interpolated_parts.join}\""
+          "\"#{handle_quotes(interpolated_parts).join}\""
+        end
+
+        def handle_quotes(parts)
+          parts.map do |part|
+            part == '"' ? '\"' : part
+          end
         end
 
         def single_quoted?(str_node)
--- a/spec/rubocop/cop/style/string_concatenation_spec.rb
+++ b/spec/rubocop/cop/style/string_concatenation_spec.rb
@@ -158,4 +158,26 @@ RSpec.describe RuboCop::Cop::Style::StringConcatenation do
       RUBY
     end
   end
+
+  context 'empty quotes' do
+    it 'registers offense and corrects' do
+      expect_offense(<<-RUBY)
+        '"' + "foo" + '"'
+        ^^^^^^^^^^^^^^^^^ Prefer string interpolation to string concatenation.
+        '"' + "foo" + "'"
+        ^^^^^^^^^^^^^^^^^ Prefer string interpolation to string concatenation.
+        "'" + "foo" + '"'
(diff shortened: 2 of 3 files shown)


spec/rubocop/cop/style/string_concatenation_spec.rb
```diff
@@ -158,4 +158,26 @@
...
+        "'" + "foo" + '"' + "bar"
+        ^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer string interpolation to string concatenation.
+      RUBY
+
+      expect_correction(<<-RUBY)
+        "'foo'"
+        "'foo'"
+        "'foo'"
+        "'foo'bar"
+      RUBY
```
dvandersluis, line 180: This doesn't look right to me, it's changing the actual string value now. The original examples returned these values:

```
"foo"
"foo'
'foo"
'foo"bar
```

tejasbubane: @dvandersluis You are right. I have fixed this now.

lib/rubocop/cop/style/string_concatenation.rb
```diff
@@ -106,7 +106,13 @@ def replacement(parts)
               end
             end
 
-          "\"#{interpolated_parts.join}\""
+          "\"#{handle_empty_quotes(interpolated_parts).join}\""
+        end
+
+        def handle_empty_quotes(parts)
```
bbatsov, line 112: What do you mean by "empty quotes" here? Is this an empty string that's basically just ""?

tejasbubane: Renamed to `handle_quotes`. I mean handle when parts contain quote only like `'"'` or `"'"`.

bbatsov:
Thanks!

PR #9143, opened 2020-12-02
[Fix #9125] Allow ConfigObsoletion to be extended by other RuboCop libraries

### LINKED ISSUE #9125: Better support for packages: renamed / obsolete cops / config
It would be nice if packages had an API to tap into our [config_obsoletion class](https://github.com/rubocop-hq/rubocop/blob/5784d45f6c17d15c7aa72cb6d727352819351bcb/lib/rubocop/config_obsoletion.rb)

See https://github.com/rubocop-hq/rubocop-performance/issues/198#issuecomment-733497898

dvandersluis:
Adds the ability to add config obsoletions to extension libraries. Fixes #9125.

In order to facilitate this, I did two things, broken into separate commits.

1. I updated the obsoletions to be defined in a yaml file instead of a series of hashes. My rationale here is that it'll be simpler to configure in YAML than to try to reopen RuboCop's hashes (yes they weren't frozen but they probably should have been). In order to facilitate this, I refactored `ConfigObsoletion` and introduced `Rule` objects to encapsulate each type of rule. Each rule class is responsible for determining how it is violated, and what message to show in that case. The YAML file is parsed and converted to rules, which are then iterated over to find issues.
2. I extended the above to allow `ConfigObsoletion` to be given additional filenames to load. This will allow extensions to create their own version of `config/obsoletions.yml` (or whatever they want to call it) and tell RuboCop to load it. I also added to the documentation to explain how to use this.

Extension libraries can configure like so:

```ruby
RuboCop::ConfigObsoletion.files << filename
```

As always, feedback is welcome and I'm happy to make any changes.

-----------------

* [x] Added tests.


config/obsoletion.yml
```diff
@@ -0,0 +1,196 @@
...
+  Style/OpMethod: Naming/BinaryOperatorParameterName
+  Style/SingleSpaceBeforeFirstArg: Layout/SpaceBeforeFirstArg
+  Style/UnneededCapitalW: Style/RedundantCapitalW
+  Style/UnneededCondition: Style/RedundantCondition
+  Style/UnneededInterpolation: Style/RedundantInterpolation
+  Style/UnneededPercentQ: Style/RedundantPercentQ
+  Style/UnneededSort: Style/RedundantSort
+
+# Cops that were moved to a different department
+moved:
```
bbatsov, line 42: Do we need `renamed` and `moved` separately only to provide better messages?  In general it's easy to infer from the old name/new name that a cop has been renamed or moved, so I'm wondering if we should not simplify this.

bbatsov:
I like the spirit of the change overall. 

One thing I'm pondering with respect to cop renames is changing the current behavior from failing to warning and somehow accounting the for the obsolete config in RuboCop internally (e.g. merging the obsolete config and updating the cop name there or something along those lines). This will give us the flexibility to rename cops in minor releases (we clearly have many poorly named cops for historical reasons and it'd be nice if we could tackle those gradually).

As for how useful the current suggestion is for the extensions I'll defer to @koic and the @rubocop-hq/rubocop-rspec team.

docs/modules/ROOT/pages/extensions.adoc
```diff
@@ -83,6 +83,77 @@ Upstream best practices and coding conventions for downstream compatibility.
...
+If you're writing an extension, you can tie some of its configuration into RuboCop.
+
+=== Config Obsoletions
+When a cop that has been released is later renamed or removed, or one of its parameters is, RuboCop can output error messages letting users know to update their configuration to the newest values. If any obsolete configurations are encountered, RuboCop will output an error message and quit.
+
+You can tie your extension into this system by creating your own `obsoletions.yml` file and letting RuboCop know where to find it:
+
+[source,ruby]
+----
+RuboCop::ConfigObsoletion.files << filename
```
pirj, line 97: Does it make sense to make it clear that it accepts an absolute path?
From extension's code, it might not be obvious that `File.expand_path` should be called.

lib/rubocop/config_obsoletion.rb
```diff
@@ -4,298 +4,102 @@ module RuboCop
...
-
-    def obsolete_enforced_style
-      OBSOLETE_ENFORCED_STYLES.map do |params|
-        obsolete_enforced_style_message(params[:cop], params[:parameter],
-                                        params[:enforced_style],
-                                        params[:alternative])
+    # Default rules for obsoletions are in config/obsoletion.yml
+    # Additional rules files can be added with `RuboCop::ConfigObsoletion.files << filename`
+    def load_rules
+      rules = self.class.files.each_with_object({}) do |filename, hash|
```
pirj, line 51: Just in case if some extension decides to override RuboCop's own cops, maybe `.reverse`? Or mix in `DEFAULT_RULE_FILE` right here, e.g. `[DEFAULT_RULE_FILE, *self.class.files].each...`?

lib/rubocop/config_obsoletion.rb
```diff
@@ -4,298 +4,102 @@ module RuboCop
...
-          next
+    # Cop rules are keyed by the name of the original cop
+    def load_cop_rules(rules)
+      rules.each_with_object([]) do |(rule_type, data), out|
+        objects = data.map do |configuration|
+          COP_RULE_CLASSES[rule_type].new(@config, *configuration)
         end
 
-        messages
+        out.concat(objects.compact)
```
pirj, line 70: Cosmetic: it's just a matter of a personal preference, and leaving this completely up to you.
```ruby
rules.flat_map do |rule_type, data|
  data.map do |configuration|
    COP_RULE_CLASSES[rule_type].new(@config, *configuration)
  end
end
```

or
```ruby
rules.flat_map do |rule_type, data|
  rule_class = COP_RULE_CLASSES[rule_type]
  data.map do |configuration|
    rule_class.new(@config, *configuration)
  end
end
```

lib/rubocop/config_obsoletion/changed_enforced_styles.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module RuboCop
+  class ConfigObsoletion
+    # Encapsulation of a ConfigObsoletion rule for changing a parameter
+    # @api private
+    class ChangedEnforcedStyles < ParameterRule
+      def violated?
+        style = config[cop]&.detect { |key, _| key.start_with?(parameter) }
+        style && style[1] == value
```
pirj, line 10: Might be just
```ruby
def violated?
  config[cop]&.detect { |key, actual_value| key.start_with?(parameter) && actual_value == value }
end
```

lib/rubocop/config_obsoletion/changed_enforced_styles.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module RuboCop
+  class ConfigObsoletion
+    # Encapsulation of a ConfigObsoletion rule for changing a parameter
+    # @api private
+    class ChangedEnforcedStyles < ParameterRule
+      def violated?
+        style = config[cop]&.detect { |key, _| key.start_with?(parameter) }
```
pirj, line 9: What are `key` and `parameter` here? Why can't we `==` them?

lib/rubocop/config_obsoletion/parameter_rule.rb
```diff
@@ -0,0 +1,44 @@
...
+        @parameter = parameter
+        @metadata = metadata
+      end
+
+      def parameter_rule?
+        true
+      end
+
+      def violated?
+        @config[cop]&.key?(parameter)
```
pirj, line 22: There's an `attr_reader` for `@config` (in other sibling rule classes as well).

pirj (approved):
Wow, a nice addition, thanks!
I suspected that given the config inject trick, config obsoletion for extensions does not exist.

Special thanks for `Configuring an Extension` doc section!

All notes are cosmetic, LGTM! 💯

lib/rubocop/config_obsoletion.rb
dvandersluis: `self.class.files` is defaulted to `[DEFAULT_RULE_FILE]` with this in mind (so that `files << ...` will append but the default will continue to be first.

I'm not sure that an extension _should_ override something from RuboCop, but if it does, it probably should be applied?

docs/modules/ROOT/pages/extensions.adoc
dvandersluis: That's a good point, I'll update.

config/obsoletion.yml
dvandersluis: I've merged the two together now, thanks for the suggestion!

lib/rubocop/config_obsoletion/changed_enforced_styles.rb
dvandersluis: I've changed it to just be a straight up `==`, we can extend it in the future if something comes up to make it necessary.

lib/rubocop/config_obsoletion.rb
dvandersluis: Love it, thanks!

dvandersluis:
@bbatsov @pirj updated with your suggestions, thanks! I left one comment unresolved because I'm not sure if there's something to be done there or not.

> One thing I'm pondering with respect to cop renames is changing the current behavior from failing to warning and somehow accounting the for the obsolete config in RuboCop internally (e.g. merging the obsolete config and updating the cop name there or something along those lines). This will give us the flexibility to rename cops in minor releases (we clearly have many poorly named cops for historical reasons and it'd be nice if we could tackle those gradually).

I agree that there is more that can be done here. It'd be nice if old config was automatically rewritten to the new one instead of raising an error (and having everything just be a warning like we did for #9098). I think with the rules extracted to objects it should be feasible to do this, but as a future task.

pirj:
Looks good!

I'm a bit concerned about the removal of `start_with?(param)`, I hope the exhaustive test coverage will prevent any regressions.
Just in case, `start_with?` was introduced [here](https://github.com/rubocop-hq/rubocop/commit/1dc32483c59ba3e61c783c09555df0615c3cc3a4#diff-b35fca184e57be3f44600ddadf6c2ceee757410050664099239e990a903d85a9R619).

bbatsov:
> I agree that there is more that can be done here. It'd be nice if old config was automatically rewritten to the new one instead of raising an error (and having everything just be a warning like we did for #9098). I think with the rules extracted to objects it should be feasible to do this, but as a future task.

Yeah, I totally agree. I like doing changes in smaller increments in general, just wanted to share my musings based on the journey to 1.0.

dvandersluis:
> I'm a bit concerned about the removal of start_with?(param)

Yeah I added a test specifically for that rule which was previously missing. I tracked it down, the obsoletion type was added in #7113, and `Layout/IndentationConsistency` is the only cop using it. Looks like it previously was only the parameter `EnforcedStyle` so this shouldn't break anything as far as I see.

bbatsov (approved):
I'll merge this after we cut 1.5.2.

PR #9153, opened 2020-12-02
[Fix #9144] Add `aggressive` and `conservative` enforced styles for `Style/StringConcatenation` cop

### LINKED ISSUE #9144: Style/StringConcatenation mistakenly assumes that the + operator means concatenation, e.g. with Pathname
## Expected behavior

Rubocop should not suggest replacing an expression that returns a `Pathname` with an expression that returns a `String`, even if we use the `+` operator to combine a `Pathname` and `String` (which [returns a `Pathname`](https://ruby-doc.org/stdlib-2.6.3/libdoc/pathname/rdoc/Pathname.html#method-i-2B))

## Actual behavior

```
1.5.0 (using Parser 2.7.2.0, rubocop-ast 1.3.0, running on ruby 2.6.6 x86_64-linux)
Inspecting 1 file
C

Offenses:

bad.rb:4:7: C: Style/StringConcatenation: Prefer string interpolation to string concatenation.
puts (Pathname.new('/') + 'test').class
      ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Here are the files I used to produce the error:

`Gemfile`
```ruby
source 'https://rubygems.org'
gem 'rubocop', '=1.5.0'
```

`bad.rb`
```ruby
# frozen_string_literal: true

require 'pathname'
puts (Pathname.new('/') + 'test').class
```

> Note that running the `bad.rb` will print `Pathname`

### Faster way to reproduce the issue

```Dockerfile
FROM ruby:2.6-slim
ARG TESTFILE=bad.rb

RUN true \
  && echo "source 'https://rubygems.org'" > Gemfile \
  && echo "gem 'rubocop', '=1.5.0'" >> Gemfile \
  && cat Gemfile \
  && bundle install

RUN true \
  && echo "# frozen_string_literal: true\n" > $TESTFILE \
  && echo "require 'pathname'" >> $TESTFILE \
  && echo "puts (Pathname.new('/') + 'test').class" >> $TESTFILE \
  && cat $TESTFILE

RUN ruby $TESTFILE
RUN bundle exec rubocop -V && bundle exec rubocop --debug -D $TESTFILE
```

Saving this as `Dockerfile` and running `docker build .` should show the entire process

tejasbubane:
Closes #9144

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_string_concatenation_enforced_styles.md
+++ b/changelog/change_string_concatenation_enforced_styles.md
@@ -1,0 +1,1 @@
+* [#9144](https://github.com/rubocop/rubocop/issues/9144): Add `aggressive` and `conservative` modes of operation for `Style/StringConcatenation` cop. ([@tejasbubane][])
--- a/lib/rubocop/cop/style/string_concatenation.rb
+++ b/lib/rubocop/cop/style/string_concatenation.rb
@@ -15,17 +15,36 @@ module RuboCop
       # lines, this cop does not register an offense; instead,
       # `Style/LineEndConcatenation` will pick up the offense if enabled.
       #
-      # @example
+      # Two modes are supported:
+      # 1. `aggressive` style checks and corrects all occurrences of `+` where
+      # either the left or right side of `+` is a string literal.
+      # 2. `conservative` style on the other hand, checks and corrects only if
+      # left side (receiver of `+` method call) is a string literal.
+      # This is useful when the receiver is some expression that returns string like `Pathname`
+      # instead of a string literal.
+      #
+      # @example Mode: aggressive (default)
       #   # bad
       #   email_with_name = user.name + ' <' + user.email + '>'
+      #   Pathname.new('/') + 'test'
       #
       #   # good
       #   email_with_name = "#{user.name} <#{user.email}>"
       #   email_with_name = format('%s <%s>', user.name, user.email)
+      #   "#{Pathname.new('/')}test"
       #
       #   # accepted, line-end concatenation
       #   name = 'First' +
       #     'Last'
+      #
+      # @example Mode: conservative
+      #   # bad
+      #   'Hello' + user.name
+      #
+      #   # good
+      #   "Hello #{user.name}"
+      #   user.name + '!!'
+      #   Pathname.new('/') + 'test'
       #
       class StringConcatenation < Base
         include Util
@@ -52,10 +71,15 @@ module RuboCop
           return if line_end_concatenation?(node)
 
           topmost_plus_node = find_topmost_plus_node(node)
+          parts = collect_parts(topmost_plus_node)
+          return unless parts[0..-2].any? { |receiver_node| offensive_for_mode?(receiver_node) }
 
-          parts = []
-          collect_parts(topmost_plus_node, parts)
+          register_offense(topmost_plus_node, parts)
+        end
 
+        private
+
+        def register_offense(topmost_plus_node, parts)
           add_offense(topmost_plus_node) do |corrector|
             correctable_parts = parts.none? { |part| uncorrectable?(part) }
             if correctable_parts && !corrected_ancestor?(topmost_plus_node)
@@ -67,7 +91,10 @@ module RuboCop
           end
         end
 
-        private
+        def offensive_for_mode?(receiver_node)
+          mode = cop_config['Mode'].to_sym
+          mode == :aggressive || mode == :conservative && receiver_node.str_type?
+        end
 
         def line_end_concatenation?(node)
           # If the concatenation happens at the end of the line,
@@ -87,7 +114,7 @@ module RuboCop
           current
         end
 
-        def collect_parts(node, parts)
+        def collect_parts(node, parts = [])
           return unless node
 
           if plus_node?(node)
--- a/spec/rubocop/cop/style/string_concatenation_spec.rb
+++ b/spec/rubocop/cop/style/string_concatenation_spec.rb
@@ -203,4 +203,36 @@ RSpec.describe RuboCop::Cop::Style::StringConcatenation, :config do
       RUBY
     end
   end
+
+  context 'Mode = conservative' do
+    let(:cop_config) { { 'Mode' => 'conservative' } }
+
+    context 'when first operand is not string literal' do
+      it 'does not register offense' do
+        expect_no_offenses(<<~RUBY)
+          user.name + "!!"
+          user.name + "<"
+        RUBY
+      end
+    end
+
+    context 'when first operand is string literal' do
+      it 'registers offense' do
+        expect_offense(<<~RUBY)
+          "Hello " + user.name
+          ^^^^^^^^^^^^^^^^^^^^ Prefer string interpolation to string concatenation.
+          "Hello " + user.name + "!!"
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer string interpolation to string concatenation.
+          user.name + "<" + "user.email" + ">"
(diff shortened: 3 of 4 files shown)


ianfixes:
This is a small step forward, but I want to be clear that my aim in opening #9144 was simply to show that the _+ operator_ could not be trusted to indicate a concatenation operation.  What's being addressed in this PR is only my very contrived counterexample.  

A more common counterexample would be in a form like this:

```ruby
def for_example(a_pathname)
    a_pathname + "a_string"
end
```

RuboCop really has no business pretending that string concatenation is the _only_ thing that `+ "a_string"` could be accomplishing.

bbatsov:
I also don't like the proposed change as I don't think the problem is solvable by ignoring some cases. Fundamentally, the problem is that with operator overloading we can never know what happens in certain cases and we can only guess. The cop is already marked as unsafe because of this. The only thing I can think of is to add a mode that ignores cases where the first operand is not a string literal.

tejasbubane:
> This is a small step forward, but I want to be clear that my aim in opening #9144 was simply to show that the _+ operator_ could not be trusted to indicate a concatenation operation. What's being addressed in this PR is only my very contrived counterexample.

Yes I had the same thing in mind, but thought `Pathname` being stdlib made sense to add an exception.

> The only thing I can think of is to add a mode that ignores cases where the first operand is not a string literal.

That sounds a lot better. I will change this PR accordingly.

ianfixes:
> The only thing I can think of is to add a mode that ignores cases where the first operand is not a string literal.

Cross posting my latest comment from #9144, I think this could be expanded to say "only consider cases where _the first and last operands_ in a chain of `+` operators are string lterals"

```ruby
def some_fn(var_of_unknown_type)
  puts "hello" + "world"                       # string concatenation 
  puts var_of_unknown_type + "world"           # insufficient info to decide
  puts "hello" + var_of_unknown_type           # string concatenation (I think? Is there a case where that's not true?)
  puts "hello" + var_of_unknown_type + "world" # string concatenation 
end
```

tejasbubane:
@bbatsov Changed to use a config option.

lib/rubocop/cop/style/string_concatenation.rb
```diff
@@ -11,16 +11,24 @@ module Style
       # In those cases, it might be useful to extract statements to local
       # variables or methods which you can then interpolate in a string.
       #
-      # @example
+      # @example EnforcedStyle: all (default)
```
bbatsov, line 14: I'd expand the cop docs to explain the safety issues and the implications of the using the different styles.

bbatsov:
I'm also thinking we can have some more generic names for the enforced style (e.g. `conservative` and `aggressive`) or perhaps it'd be even better if this was some boolean flag like `SafeMode`. We're not really supporting different styles here, we're rather limitting the scope of the cop.

bbatsov:
Ping :-)

tejasbubane:
I forgot about this PR. I will fix this in a day or two.

bbatsov:
@tejasbubane another friendly reminder about this PR :-)

tejasbubane:
@bbatsov Changes done. Sorry for the delay.

lib/rubocop/cop/style/string_concatenation.rb
```diff
@@ -15,7 +15,13 @@ module Style
       # lines, this cop does not register an offense; instead,
       # `Style/LineEndConcatenation` will pick up the offense if enabled.
       #
-      # @example
+      # Two modes are supported:
```
bbatsov, line 18: You should also mention how this ties to the cop safety - e.g. with the `Pathname` example.

bbatsov:
Seems I totally forgot about this myself. After some further thought let's not use `EnforcedStyle`, but rather some other configuration option name (e.g. `Mode/Detection`), as that's not really a different style just an instruction how hard to try to find offenses. If someone from @rubocop/rubocop-core can suggest a better config setup - be my guest.

bbatsov (changes requested):
Overall things look good. We just need to add some safety documentation and adopt some dedicated configuration option for the two modes of operation.

spec/rubocop/cop/style/string_concatenation_spec.rb
```diff
@@ -201,4 +201,33 @@
...
   end
+
+  context 'EnforcedStyle = conservative' do
+    let(:cop_config) { { 'EnforcedStyle' => 'conservative' } }
+
+    context 'when first operand is not string literal' do
+      it 'does not register offense' do
+        expect_no_offenses(<<~RUBY)
+          user.name + "!!"
+          user.name + "<" + "user.email" + ">"
```
bquorning, line 212: Actually, I would expect an offense here. Not on `user.name + "<"`, but on `"<" + "user.email"` (and `"user.email" + ">"`).

bbatsov: Good catch!

ianfixes: I disagree.

```ruby
# yes, I know that a variable called user.name is unlikely to be a Pathname
user.name = Pathname.new("user.name")
result = user.name + "<" + "user.email" + ">"
puts "result is #{result}" # result is user.name/</user.email/>
                           # which is in no way equivalent to "user.name<user.email>"
```

tejasbubane:
Review comments fixed. 

* Changed config name to `Mode`. 
* Fixed the issue @bquorning pointed out. 
* Added `Pathname` example to cop documentation to show how the config ties to cop safety.

spec/rubocop/cop/style/string_concatenation_spec.rb
ianfixes: @tejasbubane -- I think the fix that was called for here was in error.  See my above comment

lib/rubocop/cop/style/string_concatenation.rb
```diff
@@ -15,18 +15,37 @@ module Style
...
+      # 1. `aggressive` style checks and corrects all occurrences of `+` where
+      # either the left or right side of `+` is a string literal.
+      # 2. `conservative` style on the other hand, checks and corrects only if
+      # left side (receiver of `+` method call) is a string literal.
+      # This is useful when the receiver is some expression that returns string like `Pathname`
```
ianfixes, line 23: `Pathname.new` returns a `Pathname`, not a string.  `Pathname.new("x") + "y"` also returns a `Pathname`.

I think this comment should read 
> This is useful when the receiver is an object whose `+` operator accepts a string, but does not return a string -- such as `Pathname`

changelog/change_string_concatenation_enforced_styles.md
```diff
@@ -0,0 +1 @@
+* [#9144](https://github.com/rubocop/rubocop/issues/9144): Add `aggressive` and `conservative` enforced styles for `Style/StringConcatenation` cop. ([@tejasbubane][])
```
bbatsov, line 1: Those are not enforced styles, but "modes of operation".

spec/rubocop/cop/style/string_concatenation_spec.rb
```diff
@@ -203,4 +203,36 @@
       RUBY
     end
   end
+
+  context 'EnforcedStyle = conservative' do
```
bbatsov, line 207: Mode

bbatsov:
@tejasbubane ping :-)

tejasbubane:
@bbatsov Minor wording changes done.

bbatsov (approved):

PR #9154, opened 2020-12-02
Relax regexp_parser requirement.

marcandre:
See https://github.com/rubocop-hq/rubocop/pull/9102#issuecomment-737487246

PR #9162, opened 2020-12-03
Support regexp_parser 1.8 and 2.x series

koic:
Follow https://github.com/rubocop-hq/rubocop/pull/9154, https://github.com/rubocop-hq/rubocop/issues/9155, and https://github.com/rubocop-hq/rubocop/pull/9102.

This is a step towards the widespread use of regexp_parser 2.0.

RuboCop core accepts regexp_parser 1.8 (#9154) , but several code is already incompatible with regexp_parser 1.8. It can cause issue because these combination of versions.

Therefore, this PR ports code for regexp_parser 1.8 that will never be maintained from https://github.com/rubocop-hq/rubocop/pull/9102.

Implementation of this patch, code is intentionally duplicated because it is evaluated only when the class is defined. Also, since obsoleted code for regexp_parser 1.8 is assumed to never be maintained, the target to be removed is clear.

To be honest, it's ugly as implementation, but I think it has the least impact for RuboCop 1.x series users.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_regexp_escape.rb
+++ b/lib/rubocop/cop/style/redundant_regexp_escape.rb
@@ -80,14 +80,30 @@ module RuboCop
           delimiters.include?(char)
         end
 
-        def each_escape(node)
-          node.parsed_tree&.traverse&.reduce(0) do |char_class_depth, (event, expr)|
-            yield(expr.text[1], expr.ts, !char_class_depth.zero?) if expr.type == :escape
+        if Gem::Version.new(Regexp::Parser::VERSION) >= Gem::Version.new('2.0')
+          def each_escape(node)
+            node.parsed_tree&.traverse&.reduce(0) do |char_class_depth, (event, expr)|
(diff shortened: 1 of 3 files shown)


marcandre:
Thanks, I was too lazy to do this 😅

> several code is already incompatible with regexp_parser 1.8.

Only for not-utf-8 Regexp, right?

koic:
Yeah, this patch going to address issues with multibyte characters Regexp rather than non-utf8 Regexp :-)

knu (commented):
Thanks!

bbatsov:
The fix is ugly indeed, but I think it's the best we can do. Thanks! 🙇‍♂️

ShockwaveNN:
Thank you for your fix, It's very sad for me that some basic ruby libraries are not very active maintained.
Even on level "accept PR with less strict versions, without any extra features"

But this PR allow to use latest version of `rubocop` even using forbidden dependencies

PR #9169, opened 2020-12-05
Add obsoletion message when config has Performance/Rails cops in it

dvandersluis:
Ran into this today when helping someone update rubocop. Performance cops were still in their config but no longer existing, and we did not have an obsoletion for it.

This adds a new type of cop obsoletions, where cops are extracted to another gem. It checks that the gem is not installed before outputting the error message (using the same bundle parsing as #9168, which will rebase away when that is merged) to ensure that configs are not marked as obsolete if `rubocop-performance` is active.

-----------------

* [x] Added tests.


bbatsov:
Probably you can add the `Rails` cops there as well. They were also extracted at some point from the main gem.

dvandersluis:
@bbatsov updated to include Rails too now.

dvandersluis:
Test failures appear to be unrelated.

bbatsov:
> This config should probably be extracted to rubocop-rails and removed from here.

Yeah, definitely.

lib/rubocop/config_obsoletion.rb
```diff
@@ -8,7 +8,8 @@ class ConfigObsoletion
     COP_RULE_CLASSES = {
       'renamed' => RenamedCop,
       'removed' => RemovedCop,
-      'split' => SplitCop
+      'split' => SplitCop,
```
bbatsov, line 11: I just realized that in the spirit of puns we should change "removed" to "retired". 😆 Probably we can also do something about "transfers between departments". 😆

bbatsov:
Great work!

lib/rubocop/config_obsoletion.rb
dvandersluis: Hahaha yes!

PR #9031, opened 2020-11-12
[Fix #8978] Update `Layout/LineLength` autocorrection to be able to handle method calls with long argument lists

dvandersluis:
Added the ability to autocorrect a long argument list when it's not a hash.

I think part of the confusion in #8978 is that `Layout/LineLength` marks every offense as correctable even though only a subset are. I'm not 100% sure if inheriting from `Base` instead of `Cop` would help here, but because `autocorrect` is defined on the cop, it's considered to be always autocorrectable. In order to try to get around this, I added a second commit which overrides `correctable?` so that only offense the cop can actually correct are marked. I'm happy to remove this if it's too much of a hack or not desirable!

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop-hq/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.
* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: https://chris.beams.io/posts/git-commit/

PR #9176, opened 2020-12-06
[Fix #9175] Don't count autocorrects that will not be performed

### LINKED ISSUE #9175: Wrong number of autocorrections
When running `rubocop -a --only=Layout/LineLength`, it displays a total of 12 offenses, of which 7 were autocorrected. In fact, none of those offenses were corrected.

```
app/models/user.rb:77:151: C: [Corrected] Layout/LineLength: Line is too long. [151/150]
    last_push_token = user_tokens.where.not(push_token: [nil, '']).where.not(updated_at: nil).where.not(last_sign_in: nil).order('updated_at ASC').last
```

Here, for example, an offense was supposedly corrected, although nothing in this line was changed. And, sure enough, next time I run the same command, the offense shows up again.

I am not sure what is causing the problem as it seems to occur randomly on different line lengths (150-190 characters in my case).

--------

## Expected behavior

Either not to show that the offense was corrected, or correct it once and for all.

In rubocop.yml, specify 
```
Layout/LineLength:
  Max: 150
```

Then run the autocorrection on the line above.

## RuboCop version

```
  - rubocop-faker 1.1.0
  - rubocop-performance 1.9.1
  - rubocop-rails 2.8.1
  - rubocop-rake 0.5.1
  - rubocop-rspec 2.0.1
```

dvandersluis:
In #9031, in order to try to stop `Layout/LineLength` from reporting potential corrections that could not be made, I overrode `correctable?` to only return true if there is a range that can be broken. However, I was not aware that `correctable?` is also called before any cops are run to partition cops into autocorrectable and not:

https://github.com/rubocop-hq/rubocop/blob/291aec7f04775dbe8aaff0c6a2d7328aaca0427b/lib/rubocop/cop/team.rb#L66-L91

My overridden `correctable?` was therefore causing `Layout/LineLength` to not actually autocorrect, ie. #9175.

I have changed it now to override `correction_strategy` instead and return `:unsupported` if no correction can be made for the code.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_fix_layoutlinelength_autocorrection.md
+++ b/changelog/fix_fix_layoutlinelength_autocorrection.md
@@ -1,0 +1,1 @@
+* [#9175](https://github.com/rubocop-hq/rubocop/issues/9175): Fix autocorrection reporting when autocorrection is not always possible. ([@dvandersluis][])
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -340,20 +340,28 @@ module RuboCop
       end
 
       # @return [Symbol, Corrector] offense status
-      def correct(range)
+      def correct(range, &block)
         status = correction_strategy
 
-        if block_given?
-          corrector = Corrector.new(self)
+        corrector = init_corrector(&block)
+
+        status = attempt_correction(range, corrector) if status == :attempt_correction
+        status = :unsupported if status == :uncorrected && corrector&.empty?
+
+        [status, corrector]
+      end
+
+      # @return Corrector
+      def init_corrector
+        return unless block_given?
+
+        Corrector.new(self).tap do |corrector|
           yield corrector
+
           if !corrector.empty? && !self.class.support_autocorrect?
             raise "The Cop #{name} must `extend AutoCorrector` to be able to autocorrect"
           end
         end
-
-        status = attempt_correction(range, corrector) if status == :attempt_correction
-
-        [status, corrector]
       end
 
       # @return [Symbol] offense status
--- a/lib/rubocop/cop/correctors/string_literal_corrector.rb
+++ b/lib/rubocop/cop/correctors/string_literal_corrector.rb
@@ -7,16 +7,14 @@ module RuboCop
       extend Util
 
       class << self
-        def correct(node, style)
+        def correct(corrector, node, style)
           return if node.dstr_type?
 
-          lambda do |corrector|
-            str = node.str_content
-            if style == :single_quotes
-              corrector.replace(node, to_string_literal(str))
-            else
-              corrector.replace(node, str.inspect)
-            end
+          str = node.str_content
+          if style == :single_quotes
+            corrector.replace(node, to_string_literal(str))
+          else
+            corrector.replace(node, str.inspect)
           end
         end
       end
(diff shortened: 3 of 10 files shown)


marcandre:
Good catch. Some methods like `correctable?` should probably be hidden as they are not meant to be overridden, nor used by anybody else than the engine itself. Hopefully that will be part of #8965

marcandre:
OTOH, this is not the best fix.

I think that `Offense#correctable?` should be changed to also check if there is a non-empty corrector or not. The Cop might have to be converted to the modern API (descending from `Cop::Base` instead of `Cop::Cop`) but I would guess not

dvandersluis:
@marcandre definitely agree (and in fact I called this a "hack" on #9031), but I was working around what is available. I tried playing around with having the corrector lambda return `:unsupported` and allowing it to get the status from the lambda, but then I was running into strange edge cases when autocorrection loops (like a super long long hits the cop multiple times).

Maybe [`Base#correct`](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/base.rb#L343-L357) would be a good place to set the status to `unsupported` if `corrector.empty?`

marcandre:
> Maybe [`Base#correct`](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/base.rb#L343-L357) would be a good place to set the status to `unsupported` if `corrector.empty?`

That would be fine too 👍

dvandersluis:
Alright this ended up being a bigger change than expected.

Rather than overriding `correction_strategy` as before, I have now changed `Base#correct` to return `unsupported` status if the corrector is empty, as discussed above. This actually improves some things, as previously we would emit a <code>more offenses can be corrected with &#96;rubocop -A&#96;</code> message when autocorrection was not possible at all (see changes in `cli_autocorrect_spec.rb`).

However, this also caused some cops to not display the `n auto-correctable` message. I've taken all the ones that were flagged by tests and changed them to inherit `Base` and fix, but I am not sure at the moment if there are any others that may be problematic. My guess is cops that still inherit `Cop` and autocorrect conditionally might be affected.

/cc @marcandre @bbatsov

marcandre:
> This actually improves some things, as previously we would emit a more offenses can be corrected with `rubocop -A` message when autocorrection was not possible at all 

Perfect 👍

Before merging, I'd like to understand why it doesn't work for the old style cops, let me check more fully

dvandersluis:
@marcandre this is an example of a test that was failing:

https://github.com/rubocop-hq/rubocop/blob/8b8b79db125a603f1727b319129fb55a7e0dc9ff/spec/rubocop/cli_spec.rb#L1049-L1054

The `1 offense auto-correctable` part of the output was missing, because of this:

https://github.com/rubocop-hq/rubocop/blob/55f1dfbb5ffcdce3458cc7b6912af35c7ed8ecfd/lib/rubocop/cop/mixin/string_help.rb#L10-L21

The `add_offense` on line 17 has its own block which prevents autocorrection from happening. This lead to a cascade of updating cops that relied on the `StringHelp` module and `StringLiteralCorrector` (technically I probably could have fixed it without changing the base class, but I figured two birds one stone 😆)

I took a look and I couldn't find any others that looked like they'd be problematic but I wanted to flag it in case I'm missing something.

marcandre:
Yeah, I agree this is more difficult than it should.

Part of it is my fault, I don't think there was a good reason to keep the auto-correct logic so complicated for the old style cops.

Another one is that `attempt_correction` was wrongly returning `:uncorrected` instead of `:unsupported` (my bad 😅).

Here's my attempt at fixing all this: #9190. Does it makes sense to you?

dvandersluis:
I'm going to close this PR and open a new PR with the converted cops since #9190 supercedes this.

PR #9178, opened 2020-12-06
[Fix #9174] Handle send nodes with unparenthesized arguments in `Style/SoleNestedConditional`

### LINKED ISSUE #9174: False autocorrection when merging "exclude?" in outer conditional
When running `rubocop -a` on the following code, it autocorrects it to something syntactically wrong.

```
if condition
     array << device if (deviceArray1 + deviceArray2 + deviceArray3).exclude? device
end
```

Result:
```
if condition && (deviceArray1 + deviceArray2 + deviceArray3).exclude? device
     array << device 
end
```

RuboCop even throws an error in Lint/Syntax on the newly corrected line, so it is obviously a wrong correction. The correction stems from Style/SoleNestedConditional

--------

## Expected behavior

Not to make my code syntactically wrong. Maybe add brackets?

## Actual behavior

A wrong autocorrection

## RuboCop version

```
  - rubocop-faker 1.1.0
  - rubocop-performance 1.9.1
  - rubocop-rails 2.8.1
  - rubocop-rake 0.5.1
  - rubocop-rspec 2.0.1
```

dvandersluis:
If a condition is rewritten by `Style/SoleNestedConditional` that was an unparenthesized send node, it would be rewritten to a syntax error. This changes it to wrap the send node in parens.

Fixes #9174.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/sole_nested_conditional_spec.rb
+++ b/spec/rubocop/cop/style/sole_nested_conditional_spec.rb
@@ -266,6 +266,82 @@ RSpec.describe RuboCop::Cop::Style::SoleNestedConditional, :config do
     RUBY
   end
 
+  context 'when the inner condition has a send node without parens' do
+    context 'in guard style' do
+      it 'registers an offense and corrects' do
+        expect_offense(<<~RUBY)
+          if foo
+            do_something if ok? bar
+                         ^^ Consider merging nested conditions into outer `if` conditions.
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          if foo && (ok? bar)
+            do_something
+          end
+        RUBY
+      end
+    end
+
+    context 'in modifier style' do
+      it 'registers an offense and corrects' do
+        expect_offense(<<~RUBY)
+          if foo
+            if ok? bar
+            ^^ Consider merging nested conditions into outer `if` conditions.
+              do_something
+            end
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          if foo && (ok? bar)
+              do_something
+            end
+        RUBY
+      end
+    end
+  end
+
+  context 'when the inner condition has a send node with parens' do
+    context 'in guard style' do
+      it 'registers an offense and corrects' do
+        expect_offense(<<~RUBY)
+          if foo
+            do_something if ok?(bar)
+                         ^^ Consider merging nested conditions into outer `if` conditions.
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          if foo && ok?(bar)
+            do_something
+          end
+        RUBY
+      end
+    end
+
+    context 'in modifier style' do
+      it 'registers an offense and corrects' do
+        expect_offense(<<~RUBY)
+          if foo
+            if ok?(bar)
+            ^^ Consider merging nested conditions into outer `if` conditions.
+              do_something
+            end
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          if foo && ok?(bar)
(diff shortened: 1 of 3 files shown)


dvandersluis:
FYI - the original correction, `if x && y z` _would_ be fixed by `Style/MethodCallWithArgsParentheses`, and this does work if you run rubocop with `--only Style/SoleNestedConditional, Style/MethodCallWithArgsParentheses`. However, when the set of cops is not limited like this, `Lint/Syntax` catches it first and `MethodCallWithArgsParentheses` is unable to correct the code.

It'd be ideal I think if there was a way to immediately queue up a follow up cop after a cop's autocorrection completes (so that logic doesn't have to be duplicated, but I couldn't find a way to do so.

spec/rubocop/cop/style/sole_nested_conditional_spec.rb
```diff
@@ -266,6 +266,44 @@
...
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          if foo && (ok? bar)
+            do_something
+          end
+        RUBY
+      end
+    end
```
koic, line 285: The following case already in parentheses seem to unneeded a new wrap. Can you add the following case?

From:

```ruby
         if foo
            do_something if ok?(bar)
                         ^^ Consider merging nested conditions into outer `if` conditions.
          end
```

To:

```ruby
          if foo && ok?(bar)
            do_something
          end
```

spec/rubocop/cop/style/sole_nested_conditional_spec.rb
```diff
@@ -266,6 +266,44 @@
...
+        RUBY
+
+        expect_correction(<<~RUBY)
+          if foo && (ok? bar)
+              do_something
+            end
+        RUBY
+      end
+    end
+  end
```
koic, line 305: Can you add a test for the case?

From:

```
          if foo
            if ok?(bar)
            ^^ Consider merging nested conditions into outer `if` conditions.
              do_something
            end
          end
```

To:

```ruby
          if foo && ok?(bar)
              do_something
            end
```

dvandersluis:
@koic thanks will make those changes!

dvandersluis:
@koic fixed now, thanks!

spec/rubocop/cop/style/sole_nested_conditional_spec.rb
```diff
@@ -266,6 +266,82 @@
...
+      it 'registers an offense and corrects' do
+        expect_offense(<<~RUBY)
+          if foo
+            do_something if ok? bar
+                         ^^ Consider merging nested conditions into outer `if` conditions.
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          if foo && (ok? bar)
```
koic, line 280: A little worrisome is that `foo && ok?(bar)` is (maybe) more natural code rather than `foo && (ok? bar)` for user 😅

dvandersluis: Yeah, I agree, but I couldn't get it working otherwise. If I try to correct the `condition` node I get a clobbering error. Any suggestions?

koic: Unfortunately I don't have an immediate idea 💦 , but I think it can be improved later. First, I'm going to merge this PR to prevent the syntax error. Thank you for your consideration and fixing!

dvandersluis: Thanks @koic!

PR #9187, opened 2020-12-08
[Proposal] Update formatters to output `[Correctable]` for correctable offenses

dvandersluis:
Some cops only are able to correct in specific situations, but it is unclear from the output which offenses are auto-correctable and which aren't (I kept running into this while working on #9176 and found it confusing). This adds a tag to the output for offenses which are correctable but not corrected yet:

```
Offenses:

test.rb:5:41: C: [Correctable] Layout/LineLength: Line is too long. [123/40]
  attr_accessor :name, :unit_number, :street_name, :city, :province, :postal_code, :location, :direction, :country, :region
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
test.rb:11:41: C: [Correctable] Layout/LineLength: Line is too long. [113/40]
    Model.where(username: username, email: email, first_name: first_name, last_name: last_name, country: country)
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
test.rb:15:41: C: Layout/LineLength: Line is too long. [82/40]
    # this is a really long comment that is longer than `Layout/LineLength` allows
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
test.rb:16:41: C: Layout/LineLength: Line is too long. [76/40]
    here.is.a.long.chain.of.methods.that.exceeds.the.max.allowed.line.length
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 4 offenses detected, 2 offenses auto-correctable
```

I updated all the formatters that were already reporting `[Corrected]`. Interesting, the json formatter already exposes `correctable` so this would be a good parallel.

I think this would be useful to show what can be automatically corrected, especially like in my above example where with just `2 offenses auto-correctable` it's impossible to know which 2 of the 4 will be corrected.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_update_formatters_to_output_correctable.md
+++ b/changelog/change_update_formatters_to_output_correctable.md
@@ -1,0 +1,1 @@
+* [#9187](https://github.com/rubocop-hq/rubocop/pull/9187): Update formatters to output `[Correctable]` for correctable offenses. ([@dvandersluis][])
--- a/lib/rubocop/formatter/emacs_style_formatter.rb
+++ b/lib/rubocop/formatter/emacs_style_formatter.rb
@@ -27,6 +27,8 @@ module RuboCop
             "[Todo] #{offense.message}"
           elsif offense.corrected?
             "[Corrected] #{offense.message}"
+          elsif offense.correctable?
+            "[Correctable] #{offense.message}"
           else
             offense.message
           end
--- a/lib/rubocop/formatter/simple_text_formatter.rb
+++ b/lib/rubocop/formatter/simple_text_formatter.rb
@@ -90,6 +90,8 @@ module RuboCop
             green('[Todo] ')
           elsif offense.corrected?
             green('[Corrected] ')
+          elsif offense.correctable?
+            yellow('[Correctable] ')
           else
             ''
           end
--- a/lib/rubocop/formatter/tap_formatter.rb
+++ b/lib/rubocop/formatter/tap_formatter.rb
@@ -71,6 +71,8 @@ module RuboCop
             '[Todo] '
           elsif offense.corrected?
             '[Corrected] '
+          elsif offense.correctable?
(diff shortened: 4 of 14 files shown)


bbatsov:
I like the idea, but I wondering if it wouldn't break some tools that might be parsing the output of these messages (e.g. some editor plugins). Unlikely, but who knows... Perhaps we can make this controlled via some cmdflag to be on safe side?

dvandersluis:
> I like the idea, but I wondering if it wouldn't break some tools that might be parsing the output of these messages (e.g. some editor plugins).

I was thinking about that as well, and hoping that anything parsing rubocop output is not using one of the text-based formatters, but you're right it's certainly possible. What do you have in mind for the flag? If it's something to opt into I doubt anyone would ever enable.

bbatsov:
> What do you have in mind for the flag?

I was thinking of an opt-out (e.g. `--compact-messages`), but maybe I'm overthinking it.  Officially those messages are not part of the API, so we can just release an update and see what will happen. 😆 @rubocop-hq/rubocop-core What do you think ?

koic:
I totally agree with @bbatsov's opinion.

I also like this PR proposal. And I was worried about some tools that analyze the text format. However, since RuboCop provides JSON and XML formats, I think that it is essentially preferable to choice those programable API formats instead of analyzing the text format.

So (I'm still a little worried, but) I also think we can merge, release without new opt-out command and see what will happen.

bbatsov:
Let's live dangerously! 😆

PR #9190, opened 2020-12-08
[Fix #9175] Fix status for offenses that are not correctable.

### LINKED ISSUE #9175: Wrong number of autocorrections

marcandre:
This addresses a few things:
1) Calling `add_offense` or `add_offense { |corrector| # don't auto-correct }` should be handled the same way. An empty corrector is now replaced by `nil`.
2) Tweaks how the status is handled. Status `:uncorrected` is reserved for cases where there is a corrector but it is not applied.
3) Old style cops' `auto_correct` method is now called even when run without auto-correction. This way we can know if they can auto-correct or not. It also simplifies logic.

I should have done 3) during my refactor (I don't see how it could be incompatible)

### DIFF
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -288,13 +288,6 @@ module RuboCop
         @current_corrector&.merge!(corrector) if corrector
       end
 
-      def correction_strategy
-        return :unsupported unless correctable?
-        return :uncorrected unless autocorrect?
-
-        :attempt_correction
-      end
-
       ### Reserved for Commissioner:
 
       def current_offense_locations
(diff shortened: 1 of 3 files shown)


bbatsov:
Looks good to me.

dvandersluis (approved):
Cleaner than my solution. :shipit:

dvandersluis:
Now that #9187 is merged the changed tests might fail now because they shouldn't be marked as `[Correctable]` with this PR. Also probably makes sense to have tests specifically for not-correctable and for unsafe-correctable like I did in #9176.

PR #9204, opened 2020-12-10
Freeze global registry when testing

marcandre:
This insures that there are no side effects from running a particular spec.
Use `:restore_registry` / `RuboCop::Cop::Registry.with_temporary_global` if
need to modify registry (e.g. with `stub_cop_class`).

See also https://github.com/rubocop-hq/rubocop/pull/9197#issuecomment-742325730

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -19,6 +19,11 @@ require 'rubocop/rspec/support'
 # in ./support/ and its subdirectories.
 Dir["#{File.dirname(__FILE__)}/support/**/*.rb"].sort.each { |f| require f }
 
+RuboCop::Cop::Registry.global.freeze
+# This insures that there are no side effects from running a particular spec.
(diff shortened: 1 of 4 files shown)


marcandre:
Oh, an extra tiny documentation commit sneaked in there 😅

spec/spec_helper.rb
```diff
@@ -19,6 +19,8 @@
 # in ./support/ and its subdirectories.
 Dir["#{File.dirname(__FILE__)}/support/**/*.rb"].sort.each { |f| require f }
 
+RuboCop::Cop::Registry.global.freeze
```
bbatsov, line 22: Might be a good idea to add a comment describing why we freeze it in the specs.

marcandre: Agreed, I moved the commit message just after the freeze 👍

dvandersluis:
Great idea @marcandre!

PR #9209, opened 2020-12-10
[Fix #9208] Use Array#bsearch instead of Array#include? to detect hid…

### LINKED ISSUE #9208: Hidden directories containing a large number of files impacts performance
Finding hidden files can have a performance impact when there are hidden
directories in the working directory that contain many files, as the
`hidden_files` array can become rather large, resulting in a long
start-up time while the target files are determined. Using a binary search
can cut down on this start-up time considerably. 

As an example, on my machine, I have a `.bundle` directory containing my local 
bundle, which contains tens of thousands of files. This causes the `target_files`
array to be quite large as a result. Scanning on these using `Array#include?` 
becomes at worst O(n), while using a binary search will be on average O(log n).
This change reduced the start-up time on my Rubocop scans from around 
100 seconds to around 15 seconds when analyzed using `Benchmark`.

## Expected behavior

Quick start-up time.

## Actual behavior

Potentially slow start-up time when a large number of hidden files are present.

Create a hidden directory containing a large number of files and run Rubocop.
The issue will worse as the number of files increases appreciably.

## RuboCop version

```
1.6.1 (using Parser 2.7.2.0, rubocop-ast 1.3.0, running on ruby 2.7.2 x86_64-darwin18)
  - rubocop-performance 1.9.1
  - rubocop-rspec 2.0.1
```

dark-panda:
Finding hidden files can have a performance impact when there are hidden
directories in the working directory that contain many files, as the
`hidden_files` array can become rather large, resulting in a long
start-up time while the target files are determined. Using a binary search
can cut down on this start-up time considerably. 

As an example, on my machine, I have a `.bundle` directory containing my local 
bundle, which contains tens of thousands of files. This causes the `target_files`
array to be quite large as a result. Scanning on these using `Array#include?` 
becomes at worst O(n), while using a binary search will be on average O(log n).
This change reduced the start-up time on my Rubocop scans from around 
100 seconds to around 15 seconds when analyzed using `Benchmark`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5267,3 +5267,4 @@
 [@ThomasKoppensteiner]: https://github.com/ThomasKoppensteiner
 [@PhilCoggins]: https://github.com/PhilCoggins
 [@tas50]: https://github.com/tas50
+[@dark-panda]: https://github.com/dark-panda
--- a/changelog/change_use_arraybsearch_instead_of_arrayinclude.md
+++ b/changelog/change_use_arraybsearch_instead_of_arrayinclude.md
@@ -1,0 +1,1 @@
+* [#9208](https://github.com/rubocop-hq/rubocop/issues/9208): Use Array#bsearch instead of Array#include? to detect hidden files. ([@dark-panda][])
--- a/lib/rubocop/target_finder.rb
+++ b/lib/rubocop/target_finder.rb
@@ -58,7 +58,7 @@ module RuboCop
       base_dir = base_dir.gsub(File::ALT_SEPARATOR, File::SEPARATOR) if File::ALT_SEPARATOR
       all_files = find_files(base_dir, File::FNM_DOTMATCH)
       # use file.include? for performance optimization
-      hidden_files = all_files.select { |file| file.include?(HIDDEN_PATH_SUBSTRING) }
+      hidden_files = all_files.select { |file| file.include?(HIDDEN_PATH_SUBSTRING) }.sort
       base_dir_config = @config_store.for(base_dir)
 
       target_files = all_files.select do |file|
@@ -70,7 +70,9 @@ module RuboCop
 
     def to_inspect?(file, hidden_files, base_dir_config)
       return false if base_dir_config.file_to_exclude?(file)
-      return true if !hidden_files.include?(file) && ruby_file?(file)
+      return true if !hidden_files.bsearch do |hidden_file|
+        file <=> hidden_file
+      end && ruby_file?(file)
 
       base_dir_config.file_to_include?(file)
     end


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#9208](https://github.com/rubocop-hq/rubocop/issues/9208): Use Array#bsearch instead of Array#include? to detect hidden files. ([@dark-panda][])
+
```
dvandersluis, line 8: Please use `rake changelog:change` (or `changelog:fix`, but I think this makes more sense as a change) instead of directly editing the CHANGELOG.

dark-panda:
Oh, cool, I didn’t realize there was a rake task. Must have missed it, I was in a rush out the door when I submitted the PR. Will reroll the patch tonight and re-submit.

marcandre:
> Oh, cool, I didn’t realize there was a rake task. Must have missed it, I was in a rush out the door when I submitted the PR. Will reroll the patch tonight and re-submit.

It's not mandatory though; the rake task will merge existing entries. It avoids conflicts and is faster to use.

bbatsov:
> This change reduced the start-up time on my Rubocop scans from around
100 seconds to around 15 seconds when analyzed using Benchmark.

That's quite impressive, although I hope we can do even better than 15 secs down the road. Let's just get the CI green and that's good to go.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#9208](https://github.com/rubocop-hq/rubocop/issues/9208): Use Array#bsearch instead of Array#include? to detect hidden files. ([@dark-panda][])
```
bbatsov, line 7: I assume the build is failing, because you forgot to add a link to your handle to the bottom of the file. Or you can just generate the changelog entry as @dvandersluis suggested. This is the way. 😆

dark-panda: Ah, yeah, that's likely it. Will fix it up now.

dark-panda:
A few things to note on this:

* My use case might be somewhat different than other people, as others may not have tens of thousands of hidden files. In my case, the `.bundle` directory had like close to 100,000 files in it due to several Ruby versions being in there. If I had move it to, say, `vendor/bundle`, the issue is lessened. 

* I have a second version of the patch that improves performance a little bit more by further reducing the number of comparisons and checks needed on hidden files by assuming that if a directory is hidden then all of the files in it are also hidden. In this case, we don't actually perform checks on all of the files in `/path/to/project/.bundle/**/*`, we instead extract `/path/to/project/.bundle` as a hidden path and then use `String#start_with?` to check inclusion in the `hidden_files` array. However, the difference between this patch and this further enhancement isn't as pronounced as I'd hoped it would be, to the point where it's statistically irrelevant for most use cases, so I went with this simplified approach instead.

bbatsov:
Thanks!

PR #9210, opened 2020-12-10
[Fix #9205] Update `Naming/MemoizedInstanceVariableName` to handle dynamically defined methods

### LINKED ISSUE #9205: False positive Naming/MemoizedInstanceVariableName
```ruby
class Test
  def self.inherited(klass)
    klass.define_method(:values) do
      @values ||= do_something
    end
  end

  inherited(self)
end
```

## Expected behavior

Memoized variable @values matches method name `values`.

## Actual behavior

Rebocop reports an offence:

Naming/MemoizedInstanceVariableName: Memoized variable @values does not match method name inherited. Use @inherited instead.

## RuboCop version

```
bundle exec rubocop -V
1.6.1 (using Parser 2.7.2.0, rubocop-ast 1.3.0, running on ruby 2.6.6 x86_64-linux-musl)
```

dvandersluis:
`Naming/MemoizedInstanceVariableName` was previously not aware of dynamically defined methods, and when defined inside another method (such as a module's `inherited` callback), it used the wrong value as the method name.

Fix to be able to find `define_method` and `define_singleton_method` and use that as the node to check against, so that the recommended ivar name is correct.

Fixes #9205.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/naming/memoized_instance_variable_name_spec.rb
+++ b/spec/rubocop/cop/naming/memoized_instance_variable_name_spec.rb
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::Naming::MemoizedInstanceVariableName, :config do
+  it 'does not register an offense when or-assignment-based memoization is used outside a method definition' do
+    expect_no_offenses(<<~RUBY)
+      @x ||= y
+    RUBY
+  end
+
   context 'with default EnforcedStyleForLeadingUnderscores => disallowed' do
     let(:cop_config) do
       { 'EnforcedStyleForLeadingUnderscores' => 'disallowed' }
@@ -94,7 +100,7 @@ RSpec.describe RuboCop::Cop::Naming::MemoizedInstanceVariableName, :config do
           RUBY
         end
 
-        it 'does not registers an offense when method has leading `_`' do
+        it 'does not register an offense when method has leading `_`' do
           expect_no_offenses(<<~RUBY)
             def _foo
               @foo ||= :foo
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/naming/memoized_instance_variable_name_spec.rb
```diff
@@ -194,6 +200,83 @@ def initialize
...
+                  klass.define_method(:values) do
+                    @values ||= do_something
+                  end
+                end
+              RUBY
+            end
+          end
+
+          context 'when the method does not match' do
+            it 'does not register an offense' do
```
koic, line 240: This looks like a typo.

```suggestion
            it 'registers an offense' do
```

bbatsov:
Great changes, just make sure to also reflect them in the cop's examples.

spec/rubocop/cop/naming/memoized_instance_variable_name_spec.rb
koic: There are likely to be some similar typos :-)

dvandersluis: Thanks @koic!

dvandersluis:
@bbatsov updated the docs

spec/rubocop/cop/naming/memoized_instance_variable_name_spec.rb
```diff
@@ -194,6 +200,83 @@ def initialize
...
+                  klass.define_method(:values) do
+                    @values ||= do_something
+                  end
+                end
+              RUBY
+            end
+          end
+
+          context 'when the method does not match' do
+            it 'it registers an offense' do
```
koic, line 240: Can you update this one?

```suggestion
            it 'registers an offense' do
```

dvandersluis: 🤦

dvandersluis: Fixed!

koic:
Nice! Thank you!

PR #9221, opened 2020-12-13
Allow simple math for `Lint/BinaryOperatorWithIdenticalOperands` cop

### LINKED ISSUE #8482: Lint/BinaryOperatorWithIdenticalOperands warning for simple math

1. Create test.rb file with 
```
x = 1
y = x * x * 2
puts y
```
2. Runing rubocop on it result:
```
Offenses:

1.rb:2:5: W: Lint/BinaryOperatorWithIdenticalOperands: Binary operator * has identical operands.
y = x * x * 2
```
I'm thinking this case is out of scope of `Lint/BinaryOperatorWithIdenticalOperands`, just a simple math calculation

This code can be fixed by 
```
y = x ^ 2 * 2
```

But I don't think original code is bad

## RuboCop version

```
$ [bundle exec] rubocop -V
0.89.0 (using Parser 2.7.1.4, rubocop-ast 0.3.0, running on ruby 2.7.1 x86_64-linux)

```

fatkodima:
Fixes #8482 

I have removed several operators (`-`, `/`, `%`, `|` and `^`) from the ignore list, because each of them makes no sense with identical operands and looks like a bug (for example, `whatever` `-` `whatever` is always `0`, etc).

```ruby
# good
x * x
x + x
...

# bad
x / x
x % x
...
```

### DIFF
--- a/lib/rubocop/cop/lint/binary_operator_with_identical_operands.rb
+++ b/lib/rubocop/cop/lint/binary_operator_with_identical_operands.rb
@@ -17,6 +17,7 @@ module RuboCop
       #
       # @example
       #   # bad
+      #   x / x
       #   x.top >= x.top
       #
       #   if a.x != 0 && a.x != 0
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/lint/binary_operator_with_identical_operands.rb
```diff
@@ -27,15 +28,19 @@ module Lint
...
       #   end
       #
+      #   # good
+      #   x + x
+      #   1 << 1
+      #
       class BinaryOperatorWithIdenticalOperands < Base
         MSG = 'Binary operator `%<op>s` has identical operands.'
-        MATH_OPERATORS = %i[+ - * / % ** << >> | ^].to_set.freeze
+        MATH_OPERATORS = %i[+ * ** << >>].to_set.freeze
```
bbatsov, line 37: I think this name is now a bit misleading, as it doesn't relay the fact those are the ops that are considered fine.

fatkodima: Agreed. Updated.

bbatsov:
Thanks!

PR #9226, opened 2020-12-14
Add new `Lint/AmbiguousAssignment` cop

### LINKED ISSUE #9223: New lint cop to check mistypes in `+=` or `-=`
Inspired by https://rules.sonarsource.com/ruby/type/Bug/RSPEC-2757 and https://www.viva64.com/en/w/v588/

I think many people did mistakes like the following
```ruby
x =+ 2
```

fatkodima:
Closes #9223

### DIFF
--- a/changelog/new_ambiguous_assignment_cop.md
+++ b/changelog/new_ambiguous_assignment_cop.md
@@ -1,0 +1,1 @@
+* [#9223](https://github.com/rubocop-hq/rubocop/issues/9223): Add new `Lint/AmbiguousAssignment` cop. ([@fatkodima][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -1361,6 +1361,11 @@ Layout/TrailingWhitespace:
 #################### Lint ##################################
 ### Warnings
 
+Lint/AmbiguousAssignment:
+  Description: 'Checks for mistyped shorthand assignments.'
+  Enabled: pending
+  VersionAdded: '<<next>>'
+
 Lint/AmbiguousBlockAssociation:
   Description: >-
                  Checks for ambiguous block association with method when param passed without
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -249,6 +249,7 @@ require_relative 'rubocop/cop/layout/space_inside_string_interpolation'
 require_relative 'rubocop/cop/layout/trailing_empty_lines'
 require_relative 'rubocop/cop/layout/trailing_whitespace'
 
+require_relative 'rubocop/cop/lint/ambiguous_assignment'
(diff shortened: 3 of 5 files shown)


lib/rubocop/cop/lint/suspicious_assignment.rb
```diff
@@ -0,0 +1,59 @@
...
+      #
+      class SuspiciousAssignment < Base
+        include RangeHelp
+        extend AutoCorrector
+
+        MSG = 'Suspicious assignment detected. Did you mean `%<op>s`?'
+
+        SIMPLE_ASSIGNMENT_TYPES = %i[lvasgn ivasgn cvasgn gvasgn casgn].freeze
+
+        MISTYPES = {
```
bbatsov, line 27: I think something like `MISTYPED_ASSIGNMENTS" or "MISTAKES" would be a better name here.

lib/rubocop/cop/lint/suspicious_assignment.rb
```diff
@@ -0,0 +1,59 @@
...
+module RuboCop
+  module Cop
+    module Lint
+      # This cop checks for mistyped shorthand assignments.
+      #
+      # @example
+      #   # bad
+      #   x =- y
+      #   x =+ y
+      #   x =! y
```
bbatsov, line 12: Btw, perhaps `x =* y` should be handled as well. It does work. :-)

dvandersluis:
Maybe `AmbiguousAssignment` would be a better name?

dvandersluis:
Also, I think we need to handle ensure that `x = -y` is allowed, since it'll parse to the same AST as `x =- y`, and is valid. The offense message should probably suggest that `x =- y` could be either `x -= y` or `x = -y` since both are valid and we can't know what they were going for.

bbatsov:
I think that's why there's no auto-correct for the cop - it's hard to tell if that's a bug or poorly formatted code. That being said it seems to me the code simply checks in the source if the operator is right after the `=`,  so it wouldn't result in a false positive for `x = -y`. That being said, some tests to cover this and `+/*y` would be useful.

fatkodima:
For me both names (current and `AmbiguousAssignment`) are good enough, not sure which is better.
Expanded tests.

dvandersluis:
`AmbiguousAssignment` would be parallel with the other Lint cops (`AmbiguousOperator`, `AmbiguousBlockAssociation`, `AmbiguousRegexpLiteral`), but I'd leave it up to @bbatsov

bbatsov:
Let's go with Ambiguous. Consistent naming is a good thing.

fatkodima:
Changed the name.

bbatsov:
Thanks!

PR #9232, opened 2020-12-15
Fix wrong offense in Style/SymbolProc

tdeo:
While upgrading to rubocop `1.6.1`, I noticed that the `Style/SymbolProc` was registering an offense on the following code:
```ruby
array.map { _2.first }
```
And suggesting to autocorrect to `array.map(&:first)`, which is incorrect in such case.

I believe this was introduced in https://github.com/rubocop-hq/rubocop/pull/9127

I suggest changing the cop to only register offenses for the `_1` numbered block argument.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_style_symbol_proc
+++ b/changelog/fix_a_false_positive_for_style_symbol_proc
@@ -1,0 +1,1 @@
+* [#9232](https://github.com/rubocop-hq/rubocop/pull/9232): Fix `Style/SymbolProc` registering wrong offense when using a symbol numbered block argument greater than 1, i.e. `[[1, 2]].map { _2.succ }`. ([@tdeo][])
--- a/lib/rubocop/cop/style/symbol_proc.rb
+++ b/lib/rubocop/cop/style/symbol_proc.rb
@@ -25,7 +25,7 @@ module RuboCop
         def_node_matcher :symbol_proc?, <<~PATTERN
           ({block numblock}
             ${(send ...) (super ...) zsuper}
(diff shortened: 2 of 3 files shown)


dvandersluis:
Thanks for catching this!

https://github.com/rubocop-hq/rubocop/blob/701468b5dfea527d991fba5ebd7b181d6750a1a8/lib/rubocop/cop/style/symbol_proc.rb#L25-L30

You can fix this more simply by changing line 28 to `${(args (arg _)) 1}` (ie. match only numblocks with 1 numeric arg).

tdeo:
> You can fix this more simply by changing line 28 to `${(args (arg _)) 1}` (ie. match only numblocks with 1 numeric arg). Then the rest of the code doesn't need to be changed.
This is indeed a much better solution, thanks

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### Bug fixes
 
+* [#9232](https://github.com/rubocop-hq/rubocop/pull/9232): Fix `Style/SymbolProc` registering wrong offense when using a symbol numbered block argument greater than 1, i.e. `[[1, 2]].map { _2.succ }`. ([@tdeo][])
```
bbatsov, line 9: You misplaced this changelog entry. Now they go under the `changelog` dir and this file is auto-generated on release.

tdeo: Indeed, read the checklist a bit too fast and assumed it still was the same, I amended the commit

marcandre:
Great, thanks!

PR #9238, opened 2020-12-16
[Fix #9237] Add `IgnoredPatterns` configuration to `Lint/UnreachableLoop`

### LINKED ISSUE #9237: Lint/UnreachableLoop false positive with should_receive and raise in block
## Expected behavior

rubocop should not identify this as an unreachable loop.

## Actual behavior

rubocop identifies this as an unreachable loop.

test_spec.rb:

```ruby
# frozen_string_literal: true

class SomeClass; end

describe 'test' do
  it 'test' do
    SomeClass.should_receive(:test).exactly(2).times { raise StandardError }
  end
end
```

```
For /private/tmp: Default configuration from /Users/me/.rvm/gems/ruby-2.4.2/gems/rubocop-1.6.1/config/default.yml
Inspecting 1 file
Scanning /private/tmp/test_spec.rb
Loading cache from /Users/me/.cache/rubocop_cache/0fe4423234f49f9a7a29f384c05d385bdc87c27f/6d7a3b621ca1730e04accd938619e4bdab66cfb1/cbd2ebc1b9a3baa10e15b9523bb86be7ea95cc25
W

Offenses:

test_spec.rb:7:5: W: Lint/UnreachableLoop: This loop will have at most one iteration.
    SomeClass.should_receive(:test).exactly(2).times { raise StandardError }
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Finished in 0.10298199998214841 seconds

```

## RuboCop version

```
$ rubocop -V
warning: parser/current is loading parser/ruby24, which recognizes
warning: 2.4.10-compliant syntax, but you are running 2.4.2.
1.6.1 (using Parser 2.7.2.0, rubocop-ast 1.3.0, running on ruby 2.4.2 x86_64-darwin18)
```

dvandersluis:
In #9237:

```
test_spec.rb:7:5: W: Lint/UnreachableLoop: This loop will have at most one iteration.
    SomeClass.should_receive(:test).exactly(2).times { raise StandardError }
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

`Lint/UnreachableLoop` is registering on `times` here because it is an [`enumerator_method`](https://github.com/rubocop-hq/rubocop-ast/blob/master/lib/rubocop/ast/node/mixin/method_identifier_predicates.rb#L10-L13). However, it is actually an RSpec method in this case.

Since we don't know what the receiver is, `IgnoredPatterns` can now be used to have the block receiver be allowed. For instance:

```yaml
Lint/UnreachableLoop:
  IgnoredPatterns:
    - !ruby/regexp /exactly\(\d+\)\.times/
```

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2058,6 +2058,11 @@ Lint/UnreachableLoop:
   Description: 'This cop checks for loops that will have at most one iteration.'
   Enabled: true
   VersionAdded: '0.89'
+  VersionChanged: <<next>>
+  IgnoredPatterns:
+    # RSpec uses `times` in its message expectations
+    # eg. `exactly(2).times`
+    - !ruby/regexp /(exactly|at_least|at_most)\(\d+\)\.times/
 
 Lint/UnusedBlockArgument:
   Description: 'Checks for unused block arguments.'
--- a/lib/rubocop/cop/lint/unreachable_loop.rb
+++ b/lib/rubocop/cop/lint/unreachable_loop.rb
@@ -9,6 +9,12 @@ module RuboCop
       # In rare cases where only one iteration (or at most one iteration) is intended behavior,
       # the code should be refactored to use `if` conditionals.
       #
+      # NOTE: Block methods that are used with `Enumerable`s are considered to be loops.
+      #
+      # `IgnoredPatterns` can be used to match against the block receiver in order to allow
+      # code that would otherwise be registered as an offense (eg. `times` used not in an
+      # `Enumerable` context).
+      #
       # @example
       #   # bad
       #   while node
@@ -70,7 +76,16 @@ module RuboCop
       #     raise NotFoundError
       #   end
       #
+      #   # bad
+      #   2.times { raise ArgumentError }
+      #
+      # @example IgnoredPatterns: [/(exactly|at_least|at_most)\(\d+\)\.times/] (default)
+      #
+      #   # good
+      #   exactly(2).times { raise StandardError }
       class UnreachableLoop < Base
+        include IgnoredPattern
+
         MSG = 'This loop will have at most one iteration.'
 
         def on_while(node)
@@ -91,6 +106,8 @@ module RuboCop
           return false unless node.block_type?
 
           send_node = node.send_node
+          return false if matches_ignored_pattern?(send_node.source)
+
           send_node.enumerable_method? || send_node.enumerator_method? || send_node.method?(:loop)
         end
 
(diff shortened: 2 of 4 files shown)


lib/rubocop/cop/lint/unreachable_loop.rb
```diff
@@ -70,7 +76,16 @@ module Lint
       #     raise NotFoundError
       #   end
       #
+      #   # bad
+      #   2.times { raise ArgumentError }
+      #
+      # @example IgnoredPatterns: [/exactly\(\d+\)\.times/]
```
koic, line 82: I think this cop can ignore `exactly`, `at_least`, and `at_most` by default. I think it's much less likely that these will be false negatives in non-RSpec context than false positives in RSpec context.

```suggestion
      # @example IgnoredPatterns: [/(exactly|at_least|at_most)\(\d+\)\.times/] (default)
```

dvandersluis: @koic sounds good! There's a whole bunch more methods that it picks up but this is a good start.

(Currently: `all?`, `any?`, `chain`, `chunk`, `chunk_while`, `collect`, `collect_concat`, `count`, `cycle`, `detect`, `downto`, `drop`, `drop_while`, `each`, `each_cons`, `each_entry`, `each_slice`, `each_with_index`, `each_with_object`, `entries`, `filter`, `find`, `find_all`, `find_index`, `first`, `flat_map`, `grep`, `grep_v`, `group_by`, `include?`, `inject`, `lazy`, `loop`, `map`, `map!`, `max`, `max_by`, `member?`, `min`, `min_by`, `minmax`, `minmax_by`, `none?`, `one?`, `partition`, `reduce`, `reject`, `reject!`, `reverse_each`, `select`, `select!`, `slice_after`, `slice_before`, `slice_when`, `sort`, `sort_by`, `sum`, `take`, `take_while`, `times`, `to_a`, `to_h`, `to_set`, `uniq`, `upto`, `zip`)

dvandersluis: Updated!

koic (approved):
👍

config/default.yml
```diff
@@ -2058,6 +2058,9 @@ Lint/UnreachableLoop:
   Description: 'This cop checks for loops that will have at most one iteration.'
   Enabled: true
   VersionAdded: '0.89'
+  VersionChanged: <<next>>
+  IgnoredPatterns:
```
bbatsov, line 2062: I'd add a comment explaining what the default configuration targets, as it's not immediately obviously, especially if you're not using RSpec.

dvandersluis: Added a comment, let me know if you think it needs more info.

bbatsov: I think you may have forgotten to push.

dvandersluis: Oops! Sorry, pushed now.

dvandersluis: @bbatsov any changes necessary do you think?

bbatsov: Nope. I think we're good here.

PR #9241, opened 2020-12-16
[Fix #9235] Allow `--only` and `--except` to be able to properly qualify cops added by require.

### LINKED ISSUE #9235: `--only` and ambiguous cop name
Versions: rubocop 1.6.1, rubocop-rspec 2.0.1
This file:
```ruby
RSpec.describe 'something' do
  let(:field_X) { 1 }
end
```
...produces (beside others) this offense:
```
spec/foo_spec.rb:2:7: C: RSpec/VariableName: Use snake_case for variable names.
  let(:field_X) { 1 }
      ^^^^^^^^
```
...but, when I am trying to use `--only RSpec/VariableName` (actually, on a large project, checking each cop's impact):
```
$ rubocop --only RSpec/VariableName spec/
--only option: RSpec/VariableName has the wrong namespace - should be Naming
Inspecting 1 file
.

```
(Notice the warning about namespace)

Shouldn't be this way, I believe :)

dvandersluis:
Previously, cop names passed to `--only` or `--except` were qualified immediately, but because the configuration is not yet loaded when `Options.parse` is called, it could not properly resolve cops that are added by an extension or a local `require`. This was generally not a problem because unresolved cops were accepted, but in the case of a cop that shares a name with a default cop, this resulted in raising a warning that the department name was wrong, and in fact loading the wrong cop.

Instead, qualifying cop names is now done lazily, so that they are only qualified right before they are needed. This ensures that the full registry is present and external cops can be properly detected.

Fixes #9235.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_allow_only_and_except_to_be_able_to.md
+++ b/changelog/change_allow_only_and_except_to_be_able_to.md
@@ -1,0 +1,1 @@
+* [#9235](https://github.com/rubocop-hq/rubocop/issues/9235): Allow `--only` and `--except` to be able to properly qualify cops added by require. ([@dvandersluis][])
--- a/lib/rubocop/options.rb
+++ b/lib/rubocop/options.rb
@@ -92,14 +92,7 @@ module RuboCop
           raise OptionArgumentError, message
         end
 
-        @options[:"#{option}"] =
-          if list.empty?
-            ['']
-          else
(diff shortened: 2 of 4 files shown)


bbatsov:
Great solution to the problem at hand!

PR #9246, opened 2020-12-17
[Fix #9225] Fix Style/LambdaCall ignoring further offenses after opposite style is detected

### LINKED ISSUE #9225: Style/LambdaCall skips subsequent checks after finding a violation
My colleague @kelvinngsl and I might have found an unexpected behaviour in `Style/LambdaCall`. It might also be related to Style/LambdaCall's implementation of `ConfigurableEnforcedStyle` mixin.

With `EnforcedStyle: braces`, when the first `blk.call` is found, it is reported as an offense, but all subsequent code that violates the braces code style got skipped.

With `EnforcedStyle: call`, when the first `blk.()` is found, it is reported as an offense, but all subsequent code that violates the braces code style got skipped.

I have included example and steps to reproduce in the section below.

Some finding that might be helpful: 
- Looking at the source code, I think this is related to the use of `opposite_style_detected` in `lib/rubocop/cop/style/lambda_call.rb`. Once an opposite style is detected, the `ConfigurableEnforcedStyle` mixin turned the config into `{ 'Enabled' => false }`, thus ignoring any subsequent violations.

If this is indeed an issue, I am willing to give fixing it a try. However I'm not really sure where to start, so some pointer would be appreciated 🙇 

---

- I've searched open issues and it does not seem that this is already reported.
- I've checked https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/style/lambda_call.rb and it is the same as the rubocop in my machine `(1.5.2)`.

--------

## Expected behavior

RuboCop should detect all `blk.call` violations when `EnforcedStyle: braces`. Conversely, RuboCop should detect all `blk.()` violations when `EnforcedStyle: call`.

## Actual behavior

It seems that once RuboCop found the first violation, the rest of the violations get ignored, so not all of the violations end up getting reported.

The following example illustrates an exaggeration of the unexpected behaviour by adding an inline disable. The root cause should be as reported above, which is RuboCop ignoring subsequent violations.

The following example uses `EnforcedStyle: braces`. However, the same behaviour also happens with `EnforcedStyle: call` and when the codebase uses the opposite pattern.

### rubocop.yml:
```yaml
Style/LambdaCall:
  EnforcedStyle: braces
```

### test.rb:
```ruby
def test1(&blk)
  blk.()
end

def test2(&blk)
  blk.call # rubocop:disable Style/LambdaCall
end

def test3(&blk)
  blk.call
end

def test4(&blk)
  blk.call
end
``` 

### RuboCop result:
```shell
$ bundle exec rubocop test.rb

Inspecting 1 file
.

```

## RuboCop version
```
$ bundle exec rubocop -V
1.5.2 (using Parser 2.7.2.0, rubocop-ast 1.3.0, running on ruby 2.6.6 x86_64-darwin18)
```

sswander:
- The Style/LambdaCall Cop used to ignore all offenses after encountering an opposite style because during the first encounter, we set `no_acceptable_style!` and in the next iterations we early returned nil for `opposite_style_detected` check. This caused the Cop to never got around to adding the offense
- This PR fixes the behaviour by letting the Cop add the offense first before checking for `opposite_style_detected`
- This PR also updates tests to illustrate the multiple opposite style scenarios

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_bug_in_stylelambdacall.md
+++ b/changelog/fix_bug_in_stylelambdacall.md
@@ -1,0 +1,1 @@
+* [#9225](https://github.com/rubocop-hq/rubocop/issues/9225): Fix Style/LambdaCall ignoring further offenses after opposite style is detected. ([@sswander][])
--- a/lib/rubocop/cop/style/lambda_call.rb
+++ b/lib/rubocop/cop/style/lambda_call.rb
@@ -27,8 +27,9 @@ module RuboCop
         def on_send(node)
           return unless node.receiver
 
-          if offense?(node) && opposite_style_detected
+          if offense?(node)
             add_offense(node) do |corrector|
+              opposite_style_detected
               autocorrect(corrector, node)
             end
           else
--- a/spec/rubocop/cop/style/lambda_call_spec.rb
+++ b/spec/rubocop/cop/style/lambda_call_spec.rb
@@ -27,6 +27,22 @@ RSpec.describe RuboCop::Cop::Style::LambdaCall, :config do
         x.call(a, b)
       RUBY
     end
+
+    it 'registers an offense for correct + multiple opposite styles' do
+      expect_offense(<<~RUBY)
+        x.call(a, b)
+        x.(a, b)
+        ^^^^^^^^ Prefer the use of `lambda.call(...)` over `lambda.(...)`.
+        x.(a, b)
+        ^^^^^^^^ Prefer the use of `lambda.call(...)` over `lambda.(...)`.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        x.call(a, b)
+        x.call(a, b)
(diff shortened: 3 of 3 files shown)


changelog/fix_bug_in_stylelambdacall.md
```diff
@@ -0,0 +1 @@
+* [#9225](https://github.com/rubocop-hq/rubocop/issues/9225): Fix bug in Style/LambdaCall. ([@sswander][])
```
fatkodima, line 1: Better to add a more specific message here, like the PR title.

sswander: Thanks @fatkodima ! Updated.

bbatsov:
Thanks!

PR #9254, opened 2020-12-18
[Fix #9244] When a cop defined in an extension is explicitly enabled, ensure that it remains enabled.

### LINKED ISSUE #9244: Cannot enable rubocop-rspec derived cops on DisabledByDefault
To enable only certain cops derived from rubocop-rspec, I configured `DisabledByDefault: true` and tried to enable `RSpec/EmptyExampleGroup` like this:

```yaml
# .rubocop.yml
require:
  - rubocop-rspec

AllCops:
  DisabledByDefault: true

RSpec/EmptyExampleGroup:
  Enabled: true
```

ref: https://github.com/rubocop-hq/rubocop/pull/7923

## Expected behavior

I expected `RSpec/EmptyExampleGroup` to be enabled like this:

```console
$ bundle install
$ bundle exec rubocop
Inspecting 3 files
.C.

Offenses:

spec/example_spec.rb:2:3: C: RSpec/EmptyExampleGroup: Empty example group detected.
  context 'with empty example group' do
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

3 files inspected, 1 offense detected
```

## Actual behavior

But `RSpec/EmptyExampleGroup` is not enabled:

```console
$ bundle install
$ bundle exec rubocop
Inspecting 3 files
...

3 files inspected, no offenses detected
```

```ruby
# Gemfile
source 'https://rubygems.org'

gem 'rspec'
gem 'rubocop'
gem 'rubocop-rspec'
```

```yaml
# .rubocop.yml
require:
  - rubocop-rspec

AllCops:
  DisabledByDefault: true

RSpec/EmptyExampleGroup:
  Enabled: true
```

```ruby
# spec/example_spec.rb
RSpec.describe 'Example' do
  context 'with empty example group' do
  end
end
```

## RuboCop version

```
$ bundle exec rubocop -V
1.6.1 (using Parser 2.7.2.0, rubocop-ast 1.3.0, running on ruby 2.7.2 x86_64-linux)
  - rubocop-rspec 2.0.1
```

dvandersluis:
With `DisabledByDefault: true` in a config, `Enabled: false` is automatically added to every key in the default configuration.

In normal operation, the configuration from any config files (including via inheritence) is then merged into that default configuration, so that if a cop is explicitly enabled it overrides the default disable (in `ConfigLoaderResolver#merge_with_default`). This works even if `inherit_from` or `inherit_gem` are specified.

However, extensions inject their configuration in when they are required, and the injection code that extensions use also calls `merge_with_default`. In that case, the **extension's** configuration gets merged into rubocop's default, which then becomes the default configuration (note, depending on how many extensions are `require`d by the user, this process may happen multiple times). When that configuration gets the local config merged into it with `DisabledByDefault: true`, if that extension's configuration had a department config key, it would get `Enabled: false` set on it. Then, when the specific cop is checked for if it's enabled, the department value would override the local value, and the cop would be disabled.

In order to fix this, cops with `Enabled: true` in their configuration after all the merging happens are always considered to be enabled, and the department is not checked. To facilitate this, when config files are being merged (either through inheritence or through `merge_with_default`), if a department is disabled any previously defined cops that are Enabled are flipped to be disabled (as expected, since configurations are meant to override other configurations they inherit from).

--- 

Example:

```yaml
# .rubocop.yml
require:
  - rubocop-rspec
  
AllCops:
  DisabledByDefault: true

RSpec/Focus:
  Enabled: true
```

Before:
```
$ rubocop test_spec.rb
Inspecting 1 file
.

```

After:
```
$ rubocop test_spec.rb
Inspecting 1 file
C

Offenses:

test_spec.rb:31:3: C: RSpec/Focus: Focused spec found.
  fit 'does something' do
  ^^^^^^^^^^^^^^^^^^^^

```

Fixes #9244.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/config_loader_spec.rb
+++ b/spec/rubocop/config_loader_spec.rb
@@ -535,13 +535,18 @@ RSpec.describe RuboCop::ConfigLoader do
       end
     end
 
-    context 'when a department is disabled' do
+    context 'when a department is disabled', :restore_registry do
       let(:file_path) { '.rubocop.yml' }
 
       shared_examples 'resolves enabled/disabled for all cops' do |enabled_by_default, disabled_by_default|
+        before { stub_cop_class('RuboCop::Cop::Foo::Bar::Baz') }
+
         it "handles EnabledByDefault: #{enabled_by_default}, " \
            "DisabledByDefault: #{disabled_by_default}" do
           create_file('grandparent_rubocop.yml', <<~YAML)
+            Naming/FileName:
+              Enabled: pending
+
             Metrics/AbcSize:
               Enabled: true
 
@@ -550,6 +555,9 @@ RSpec.describe RuboCop::ConfigLoader do
 
             Lint:
               Enabled: false
+
+            Foo/Bar/Baz:
+              Enabled: true
           YAML
           create_file('parent_rubocop.yml', <<~YAML)
             inherit_from: grandparent_rubocop.yml
@@ -558,6 +566,12 @@ RSpec.describe RuboCop::ConfigLoader do
               Enabled: false
 
             Metrics/AbcSize:
+              Enabled: false
+
+            Naming:
+              Enabled: false
+
+            Foo/Bar:
               Enabled: false
           YAML
           create_file(file_path, <<~YAML)
@@ -599,8 +613,8 @@ RSpec.describe RuboCop::ConfigLoader do
           # Enabled in grandparent config, department disabled in parent.
           expect(enabled?('Metrics/PerceivedComplexity')).to be(false)
 
-          # Pending in default config, department disabled in grandparent.
-          expect(enabled?('Lint/StructNewOverride')).to be(false)
+          # Pending in grandparent config, department disabled in parent.
+          expect(enabled?('Naming/FileName')).to be(false)
 
           # Department disabled in child config.
           expect(enabled?('Style/Alias')).to be(false)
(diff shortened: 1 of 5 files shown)


dvandersluis:
This bug is really an artifact of the lack of plugin system, which I have some ideas about how we can get into place. However, given that we need to keep supporting the existing injection system for backwards compatibility, this needed to be fixed regardless.

spec/rubocop/config_loader_spec.rb
```diff
@@ -599,8 +605,8 @@ def enabled?(cop)
           # Enabled in grandparent config, department disabled in parent.
           expect(enabled?('Metrics/PerceivedComplexity')).to be(false)
 
-          # Pending in default config, department disabled in grandparent.
-          expect(enabled?('Lint/StructNewOverride')).to be(false)
+          # Pending in grandparent config, department disabled in parent.
+          expect(enabled?('Naming/FileName')).to be(false)
```
dvandersluis, line 609: The previous test was for a cop that wasn't actually `pending` anymore, so I updated it to not rely on the default config not changing.

dvandersluis:
I expected the department calculation change in #9258 to affect this. Now that it's merged, I've updated to ensure that nested departments work as well. /cc @mvz @bbatsov

bbatsov:
> This bug is really an artifact of the lack of plugin system, which I have some ideas about how we can get into place.

You're totally right. That's one of the major items we kept discussing for years and never really got to doing. It's nice you're thinking in that direction.

PR #9276, opened 2020-12-22
Add support for endless methods to `Style/SingleLineMethods`.

dvandersluis:
`Style/SingleLineMethods` would raise an error when encountering a Ruby 3.0 endless method. I updated it to ignore single line methods (and I am working on a new cop for endless methods).

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/single_line_methods.rb
+++ b/lib/rubocop/cop/style/single_line_methods.rb
@@ -6,6 +6,8 @@ module RuboCop
       # This cop checks for single-line method definitions that contain a body.
       # It will accept single-line methods with no body.
       #
+      # Endless methods added in Ruby 3.0 are also accepted by this cop.
+      #
       # @example
       #   # bad
       #   def some_method; body end
@@ -15,6 +17,7 @@ module RuboCop
       #   # good
       #   def self.resource_class=(klass); end
       #   def @table.columns; end
+      #   def some_method() = body
       #
       # @example AllowIfMethodIsEmpty: true (default)
       #   # good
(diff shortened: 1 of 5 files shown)


dvandersluis:
`documentation_syntax_check` was failing on the ruby 3.0 syntax in examples, so I've updated the base parser it uses to be `Parser::Ruby30` instead of `Parser::Ruby27`.

lib/rubocop/cop/style/single_line_methods.rb
```diff
@@ -32,6 +35,7 @@ class SingleLineMethods < Base
 
         def on_def(node)
           return unless node.single_line?
+          return if node.endless?
```
bbatsov, line 38: Shouldn't `#single_line?` return true on this case as well?

bbatsov:
I'll merge this, even though it seems something's wrong with the `single_line?` method.

lib/rubocop/cop/style/single_line_methods.rb
bbatsov: My bad. I just realized the cop treats specially empty definitions and now endless definition. I guess we need to improve its description a bit.

dvandersluis: @bbatsov just realized my PR description was a bit wonky so I've fixed that.

`single_line?` likely returns true for an endless method (since they generally should be on a single line I believe), however, a single line method as far as this cop is concerned is one that is of the form `def x; something end`, and the correction code was failing because `node.loc.end` doesn't exist for an endless method.

bbatsov: Yeah, I figured this out with a bit of a delay. 😆

PR #9279, opened 2020-12-22
Fix `Style/MethodCallWithArgsParentheses` for endless method definitions

dvandersluis:
When an endless method definition calls another method with arguments, parentheses are required or it will be a syntax error. Therefore, parentheses are required even for the `omit_parentheses` style. If there are no arguments, the method call will continue to be matched to the enforced style as before.

-----------------

* [x] Added tests.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -23,7 +23,7 @@ Metrics/MethodLength:
 # Offense count: 8
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 132
+  Max: 135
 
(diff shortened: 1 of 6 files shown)


bbatsov:
> When an endless method definition calls another method with arguments, parentheses are required or it will be a syntax error. Therefore, parentheses are required even for the omit_parentheses style. If there are no arguments, the method call will continue to be matched to the enforced style as before.

I dislike those methods even more now. 😆 You should put this and the other required parens (for defs with no params) into the rationale of the style guide entry.

PR #9280, opened 2020-12-22
[Fix #8938] Fix some `ConfigurableEnforcedStyle` cops so they output `Exclude` file lists in `--auto-gen-config` runs

### LINKED ISSUE #8938: 1.0 does not add Layout/SpaceInsideBlockBraces and Layout/SpaceInsideHashLiteralBraces excluding files to .rubocop_todo.yml
# 0.89

https://github.com/marutosi/redmine/blob/82b58d98dff30edd6884696b596b00aba2050369/Gemfile#L95

```
  gem 'rubocop', '~> 0.89.0'
  gem 'rubocop-performance', '~> 1.5.0'
  gem 'rubocop-rails', '~> 2.5.0'
```

https://github.com/marutosi/redmine/blob/82b58d98dff30edd6884696b596b00aba2050369/.rubocop_todo.yml#L762

```
Layout/SpaceInsideBlockBraces:
  Exclude:
    - 'app/controllers/activities_controller.rb'
    - 'app/controllers/application_controller.rb'
              .
              .
              .

Layout/SpaceInsideHashLiteralBraces:
  Exclude:
    - 'app/controllers/imports_controller.rb'
              .
              .
              .
```

# 1.0

https://github.com/marutosi/redmine/blob/f730543ddee2d4bc4e547908f711df3320a31d18/Gemfile#L95

```
  gem 'rubocop', '~> 1.0.0'
  gem 'rubocop-performance', '~> 1.8.0'
  gem 'rubocop-rails', '~> 2.8.0'
```

https://github.com/marutosi/redmine/blob/f730543ddee2d4bc4e547908f711df3320a31d18/.rubocop_todo.yml#L866

```
Layout/SpaceInsideBlockBraces:
  EnforcedStyle: no_space

Layout/SpaceInsideHashLiteralBraces:
  EnforcedStyle: no_space

```

h-lame:
[Fixes #8938]

The expected use of the API provided by `ConfigurableEnforcedStyle` is
for the cop to call `correct_style_detected` if the cop detects no
offense or to call one of the other `xxx_style_detected` api methods if
it does detect an offense, and it should also call `add_offense` to
register the offense. Doing this updates `config_to_allow_offenses` which
is used during `--auto-gen-config` to decide how to configure the cop for
the current code under review so that regardless of the styles used the
code passes review.

This PR fixes errors in several cops that use this API incorrectly:

* `Layout/SpaceAroundEqualsInParameterDefault`
* `Layout/SpaceBeforeBlockBraces`
* `Style/For`
* `Style/RaiseArgs`
* `Style/AccessModifierDeclarations`
* `Style/MethodDefParentheses`
* `Style/SpaceInsideHashLiteralBraces`
* `Style/SpaceInsideBlockBraces`

I eyeballed all the cops that include this module and the ones above are the only ones that both use the `xxx_style_detected` APIs and use them incorrectly.  The majority of the cops that include this module _don't_ use the `xxx_style_detected` APIs at all.

Note: `Style/LambdaCall` also exhibits this error, but there's an independent PR (#9246) and issue (#9225) recorded to fix that cop.

# TODO

* ~Added tests~
It's not clear to me how to write tests for this change as the error is in the output of a `--auto-gen-config` run which there don't appear to be any examples of.  We could write tests that assert the internal API is called correctly, but the rest of the suite prefers asserting on outcomes so this doesn't feel like the correct approach.  I'd appreciate any advice on the required testing needed.
  - [Added an `InternalAffairs` cop instead](https://github.com/rubocop-hq/rubocop/pull/9280#issuecomment-750524967)
* ~Squashed related commits together~
I've put single commits per cop, would you prefer I smoosh them all into one commit? 
  - [No need](https://github.com/rubocop-hq/rubocop/pull/9280#issuecomment-750025586)

-----------------

* [x] Added tests.
* [X] Added an entry (file) to the [changelog folder](https://github.com/rubocop-hq/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.


bbatsov:
> It's not clear to me how to write tests for this change as the error is in the output of a --auto-gen-config run which there don't appear to be any examples of. We could write tests that assert the internal API is called correctly, but the rest of the suite prefers asserting on outcomes so this doesn't feel like the correct approach. I'd appreciate any advice on the required testing needed.

Probably it'd be best to create some InternalAffairs cop that makes sure the API is used properly. I'll likely cut a new release on Chrismas for Ruby 3, so it'd be nice if we wrapped this up in time for it. 

> I've put single commits per cop, would you prefer I smoosh them all into one commit?

No need.

h-lame:
> Probably it'd be best to create some InternalAffairs cop that makes sure the API is used properly. I'll likely cut a new release on Chrismas for Ruby 3, so it'd be nice if we wrapped this up in time for it.

Added `InternalAffairs/StyleDetectedApiUse` (not 100% sold on the name, but 🤷).  This cop checks that the style detected API is used correctly:

1. The methods aren't used in conditionals - maybe this is overkill, but from what I saw, if they were it was a mistake.  The methods aren't predicates to be used for checking something, they mutate state and you can't rely on the return value in a conditional setting.
2. If the `correct_style_detected` method is called then so is one of the negative versions, and vice-versa.

This cop isn't perfect because to truly check 2. we'd need to trace the full path through the cop in question and work out that all branching paths that through which lead to a `correct_style_detected` can be traced through the opposite path to a negative version.  However, I think it's good enough because running it on "master" before all the commits I added it raised offences for 8 of the cops I identified as being using the API in error.  The one it couldn't detect `Style/SpaceInsideBlockBraces` was the most complicated one because while it does call `correct_style_detected` and a negative version, it calls `correct_style_detected` correctly for `EnforcedStyle` offences, but incorrectly calls the negative ones for `EnforcedStyleForEmptyBraces` offences.

Also, it checks all the cops, should it attempt to only do so if the code is a cop that includes `ConfigurableEnforcedStyle`?  This feels hard because some cops include the module indirectly through other modules, and I don't think an AST parser can be aware of that.

FWIW: because the fix for `Style/LambdaCall` lives in #9246 introducing this cop means the "build" for this PR no longer passes.  Would it make sense to cherry-pick the commit from there into here, or just do some judicious merge + rebase ordering?

bbatsov:
I've merged the other PR, so we should good. Great work! Thanks!

PR #9283, opened 2020-12-23
Add new `Style/HashExcept` cop

koic:
This PR adds new `Style/HashExcept` cop.

This cop checks for usages of `Hash#reject` and `Hash#select` methods that can be replaced with `Hash#except` method.

This cop should only be enabled on Ruby version 3.0 or higher.
(`Hash#except` was added in Ruby 3.0.)

```ruby
# bad
{foo: 1, bar: 2, baz: 3}.reject {|k, v| k == :bar }
{foo: 1, bar: 2, baz: 3}.select {|k, v| k != :bar }

# good
{foo: 1, bar: 2, baz: 3}.except(:bar)
```

cf. https://bugs.ruby-lang.org/issues/15822

-----------------

* [x] Added tests.


marcandre:
You're fast 😆

Technically, it's only if using `eql?` that they are equivalent.

```
{0 => :foo}.except(0.0)
# => {0=>:foo}
{0 => :foo}.reject {|k, v| k == 0.0 }
# => {}
```

Maybe if running in safe mode we should be more careful, so `k == :symbol` and `k.eql? 0` is ok but not `k == var`?

koic:
Oh, I couldn't catch it! I prefer being safe, so I updated this PR that way! For safe detection, it is limited to commonly used string and symbol comparisons when used `==`. Thank you always!

lib/rubocop/cop/style/hash_except.rb
```diff
@@ -0,0 +1,93 @@
...
+      #
+      class HashExcept < Base
+        include RangeHelp
+        extend TargetRubyVersion
+        extend AutoCorrector
+
+        minimum_target_ruby_version 3.0
+
+        MSG = 'Use `%<prefer>s` instead.'
+        RESTRICT_ON_SEND = %i[reject select].freeze
```
bbatsov, line 32: I think you forgot `filter` and `find_all` here.

bbatsov, line 32: I think you forgot about `filter`, `keep_if` and `delete_if`. I don't use those Hash methods often, but I recall there were a ton of aliases there.

koic: I added `filter` method. On the other hand, `delete_if` and` keep_if` methods  are not checked because they are incompatible with `except` method.

```ruby
# except
h = {foo: 1, bar: 2, baz: 3}
=> {:foo=>1, :bar=>2, :baz=>3}
h.except(:bar)
=> {:foo=>1, :baz=>3}
h
=> {:foo=>1, :bar=>2, :baz=>3}

# delete_if
h = {foo: 1, bar: 2, baz: 3}
=> {:foo=>1, :bar=>2, :baz=>3}
h.delete_if {|k, v| k == :bar }
=> {:foo=>1, :baz=>3}
h
=> {:foo=>1, :baz=>3}

# keep_if
h = {foo: 1, bar: 2, baz: 3}
=> {:foo=>1, :bar=>2, :baz=>3}
h.keep_if {|k, v| k != :bar }
=> {:foo=>1, :baz=>3}
h
=> {:foo=>1, :baz=>3}
```

I updated documentation and test on this.

bbatsov:
Thanks!

luke-hill:
@koic Question about this new cop.

I've noticed this pulling through on ruby 2.7 - However it's pulling through as `Rails/` - I'm guessing this method is something which was added to the raw ruby implementation in ruby 3.0 (Previously I'm guessing it was in AS?)

PR #9294, opened 2020-12-27
Upgrade rubocop-rspec

bquorning:
Upgrade rubocop-rspec.

-----------------

* [ ] Added tests.
* [ ] Added an entry (file) to the [changelog folder](https://github.com/rubocop-hq/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.


bquorning:
@pirj After upgrading rubocop-rspec, we run into the same type of problem as was reported in https://github.com/rubocop-hq/rubocop-rspec/issues/1111 ([example](https://app.circleci.com/pipelines/github/rubocop-hq/rubocop/3629/workflows/97c2dfca-5f05-421e-a381-e305adbfaffd/jobs/163148)). I wonder which is the best approach here.

pirj:
Wild guess that for a file named `spec/rubocop/cli/cli_autocorrect_spec.rb` top-level example groups should look something like:
```ruby
RSpec.describe RuboCop::CLI, 'CLI autocorrect', :isolated_environment do
```
or even better to name the file ``spec/rubocop/cli/autocorrect_spec.rb` and define the group as:
```ruby
RSpec.describe RuboCop::CLI, 'autocorrect', :isolated_environment do
```
The second, string argument is an "additional doc", and even though it's [discouraged to be used](https://github.com/rspec/rspec-core/pull/2681#discussion_r361811453), from my point of view it might come handy here.

Alternatively, it can be defined as:
```ruby
RSpec.describe 'RuboCop::CLI autocorrect', :isolated_environment do
```
but you'll lose `described_class` and implicitly-defined subject.

bbatsov:
Any updates here?

bquorning:
PR updated.

.rubocop_todo.yml
```diff
@@ -1,55 +1,54 @@
...
 # Configuration parameters: Prefixes.
 # Prefixes: when, with, without
 RSpec/ContextWording:
   Enabled: false
 
-# Offense count: 3810
+# Offense count: 5579
 # Configuration parameters: Max.
 RSpec/ExampleLength:
   Enabled: false
```
pirj, line 45: Semi-related. I believe this has no single chance to be fixed, can be moved to `.rubocop.yml`.

.rubocop_todo.yml
```diff
@@ -59,7 +58,7 @@ RSpec/ExpectOutput:
     - 'spec/rubocop/result_cache_spec.rb'
     - 'spec/rubocop/target_finder_spec.rb'
 
-# Offense count: 434
+# Offense count: 365
 RSpec/MultipleExpectations:
   Max: 25
```
pirj, line 63: This cop doesn't seem to apply to the established project style.

spec/rubocop/cli/auto_gen_config_spec.rb
```diff
@@ -2,8 +2,8 @@
 
 require 'timeout'
 
-RSpec.describe RuboCop::CLI, :isolated_environment do
-  subject(:cli) { described_class.new }
+RSpec.describe 'RuboCop::CLI --auto-gen-config', :isolated_environment do # rubocop:disable RSpec/DescribeClass
```
pirj, line 5: Unrelated to this PR. It seems that our `RSpec/FilePath` has two deficiencies:
 - it ignores non-constant docstring. If you happen to change it to `describe 'foo bar'`, it ignores it. Not sure why we've decided to do it this way
 - implies that the second argument is a method name in `describe RuboCop::CLI, '--auto-gen-config'`, strips dashes and suggests `autogenconfig` to be a part of the file name. In this case, this doesn't apply so well. Writing it like this triggers `RSpec/DescribeMethod` though

I can file a ticket if you think it's important

pirj (approved):
Looks great, thanks!

.rubocop_todo.yml
bquorning: I agree. It would good if someone went over the TODO file to determine which are expected to be fixed, and which are merely reflecting project style.

bbatsov:
Thanks!

PR #9310, opened 2020-12-30
[Fix #9304] Do not register offense for `Style/ExplicitBlockArgument` when the `yield` arguments are not an exact match with the block arguments.

### LINKED ISSUE #9304: Style/ExplicitBlockArgument autocorrect doesn't take into if it's yielding the same args
Given

```ruby
def each
  each_row do |row, index|
    yield row
  end
end
```

## Expected behaviour

`rubocop -A` should at worst do nothing.

## Actual behavior

```ruby
def each(&block)
  each_row(&block)
end
```

This now populates the 2nd argument of `each`, so it's not equivalent.

Run `rubocop -A` on:

```ruby
def each_row
  9.times.each_with_index do |row, index|
    yield row, index
  end
end

def each
  each_row do |row, index|
    yield row
  end
end
```

## RuboCop version

```
$ bundle exec rubocop -V
0.89.1 (using Parser 2.7.2.0, rubocop-ast 0.3.0, running on ruby 2.7.1 x86_64-darwin19)
```

dvandersluis:
When the block arguments and yield arguments are not an exact match, no offense is raised. Fixes #9304.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -3271,9 +3271,8 @@ Style/ExplicitBlockArgument:
                   that just passes its arguments to another block.
   StyleGuide: '#block-argument'
   Enabled: true
-  # May change the yielding arity.
-  Safe: false
   VersionAdded: '0.89'
+  VersionChanged: <<next>>
 
 Style/ExponentialNotation:
   Description: 'When using exponential notation, favor a mantissa between 1 (inclusive) and 10 (exclusive).'
--- a/lib/rubocop/cop/style/explicit_block_argument.rb
+++ b/lib/rubocop/cop/style/explicit_block_argument.rb
@@ -6,6 +6,9 @@ module RuboCop
       # This cop enforces the use of explicit block argument to avoid writing
       # block literal that just passes its arguments to another block.
       #
+      # NOTE: This cop only registers an offense if the block args match the
+      # yield args exactly.
+      #
       # @example
       #   # bad
       #   def with_tmp_dir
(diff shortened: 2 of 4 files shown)


lib/rubocop/cop/style/explicit_block_argument.rb
```diff
@@ -6,6 +6,9 @@ module Style
       # This cop enforces the use of explicit block argument to avoid writing
       # block literal that just passes its arguments to another block.
       #
+      # NOTE: This cop only registers an offense if the block args match the
```
bbatsov, line 9: Does this change make the cop safe?

dvandersluis: I think so! There might be a case I'm not thinking about though.

bbatsov: Well, I guess you can change `Safe` to `true` and we'll see if we're lucky. :-)

dvandersluis: Updated!

config/default.yml
```diff
@@ -3219,9 +3219,9 @@ Style/ExplicitBlockArgument:
                   that just passes its arguments to another block.
   StyleGuide: '#block-argument'
   Enabled: true
-  # May change the yielding arity.
-  Safe: false
+  Safe: true
```
koic, line 3222: Safe is true by default, so it can be omitted.

```suggestion
```

dvandersluis: 🤦 Fixed now, thanks.

dvandersluis:
@bbatsov @koic any other changes needed here?

dvandersluis:
@bbatsov I think we can include this in the bug fix release, what do you think?

bbatsov:
Consider it included! 😄

PR #9313, opened 2020-12-30
[Fix #9312] Fix `Layout/FirstHashElementLineBreak` for single line hash

### LINKED ISSUE #9312: Layout/FirstHashElementLineBreak not applying to single element hash
FirstHashElementLineBreak doesn't apply to the below case:

```ruby
{:foo => {
  :bar => "baz",
}}
```

I want it to produce:
```ruby
{
  :foo => {
    :bar => "baz",
  },
}
```

FirstHashElementLineBreak skips the outer hash because it only has a single element. I think the outer hash still counts as  "multi-line" even though it only has one element.

muirdm:
Fix Layout/FirstHashElementLineBreak to consider below as a multi-line
hash even though it only has one element:

    {:foo => {
      :bar => "baz",
    }}

Now FirstHashElementLineBreak will correct the above to:

    {
      :foo => {
        :bar => "baz",
      },
    }

Fixes #9312.


bbatsov:
The proposed change looks good. Thanks!

PR #9315, opened 2020-12-31
[Fix #9314] Fix an incorrect auto-correct for `Style/RedundantReturn`

### LINKED ISSUE #9314: Style/RedundantReturn produces a syntax error when safe-correcting methods that return arrays
When a method returns something in the form `return (a), b` or `return a, (b)`, these are safe-corrected to a syntactically invalid format and not `[a, b]` as expected.

Below I have a reproduction which only triggers the `Style/RedundantReturn` correction, however the same behavior is also observed when `Style/RedundantParentheses` is applied (in the case where `a` is a literal, for example).

--------

Start with a valid method returning multiple values:

```
❯ cat foo.rb
def complex_statement
  1
end
def return_multiple
  return (complex_statement && complex_statement), 2
end
puts return_multiple
❯ ruby foo.rb
1
2
```

## Expected behavior

```
❯ cat foo.rb
def complex_statement
  1
end
def return_multiple
  [complex_statement && complex_statement, 2]
end
puts return_multiple
```

## Actual behavior

The file is no longer valid after safe-correcting:

```
❯ rubocop -a foo.rb
Inspecting 1 file
E

Offenses:

foo.rb:5:3: C: [Corrected] Style/RedundantReturn: Redundant return detected. To return multiple values, use an array.
  return (complex_statement && complex_statement), 2
  ^^^^^^
foo.rb:5:43: E: Lint/Syntax: unexpected token tCOMMA
(Using Ruby 2.6 parser; configure using TargetRubyVersion parameter, under AllCops)
  (complex_statement && complex_statement), 2
                                          ^

```

```
❯ cat foo.rb
def complex_statement
  1
end
def return_multiple
  (complex_statement && complex_statement), 2
end
puts return_multiple
❯ ruby foo.rb
foo.rb:5: syntax error, unexpected ',', expecting end
...tatement && complex_statement), 2
```

## RuboCop version

```
❯ ruby --version
ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-darwin19]
❯ rubocop -V
1.7.0 (using Parser 2.7.2.0, rubocop-ast 1.3.0, running on ruby 2.6.6 x86_64-darwin19)
```

koic:
Fixes #9314.

This PR fixes an incorrect auto-correct for `Style/RedundantReturn` when multiple return values have a parenthesized return value.

There may be anther approach that modifies `ReturnNode#arguments` of RuboCop AST (in future).

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_incorrect_autocorrect_for_style_redundant_return.md
+++ b/changelog/fix_incorrect_autocorrect_for_style_redundant_return.md
@@ -1,0 +1,1 @@
+* [#9314](https://github.com/rubocop-hq/rubocop/issues/9314): Fix an incorrect auto-correct for `Style/RedundantReturn` when multiple return values have a parenthesized return value. ([@koic][])
--- a/lib/rubocop/cop/style/redundant_return.rb
+++ b/lib/rubocop/cop/style/redundant_return.rb
@@ -66,7 +66,7 @@ module RuboCop
         end
 
         def correct_with_arguments(return_node, corrector)
-          if return_node.arguments.size > 1
+          if return_node.children.size > 1
             add_brackets(corrector, return_node)
           elsif hash_without_braces?(return_node.first_argument)
             add_braces(corrector, return_node.first_argument)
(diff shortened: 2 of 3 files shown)


bbatsov:
Looks good!

PR #9320, opened 2020-12-31
Support unicode-display_width v2

dduugg:
Adds support for v2 of `unicode-display_width`. This version emits a [warning](https://github.com/janlelis/unicode-display_width/blob/main/lib/unicode/display_width/no_string_ext.rb#L3-L6) when requiring the `no_string_ext` version, since omitting `String` extensions is the default. To compensate, I've reverted to the normal `require`, but this will of course include `String` extensions when the gem version is < 2. (It may make sense to require the `unicode-display_width`version be >= 2, to avoid unintentionally using a conflicting `String` API, let me know which approach you'd prefer.)

The only use of this library appears to be an invocation of `Unicode::DisplayWidth.of` in `lib/rubocop/cop/mixin/alignment.rb`, so this should be compatible with the v2 library [changes](lib/rubocop/cop/mixin/alignment.rb).

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/change_support_unicodedisplay_width_v2.md
+++ b/changelog/change_support_unicodedisplay_width_v2.md
@@ -1,0 +1,1 @@
+* [#9320](https://github.com/rubocop-hq/rubocop/pull/9320): Support unicode-display_width v2. ([@dduugg][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #9323, opened 2021-01-01
Add project_spec test to ensure that `Safe: true` is not given in a config, since it is redundant

dvandersluis:
Re https://github.com/rubocop-hq/rubocop/pull/9310#discussion_r550726758

Added a test to make sure `Safe: true` isn't added to the configuration, as it is the default value.


bbatsov:
Looks good! Thanks!

koic:
Nice! 👍

PR #9324, opened 2021-01-01
Add new `Lint/DeprecatedConstants` cop

koic:
This PR adds new `Lint/DeprecatedConstants` cop.

This cop checks for deprecated constants.

It has `DeprecatedConstants` option. If there is an alternative method, you can set the value as a string. Or set `null` to the value if there is no alternative.
By default, `NIL`, `TRUE`, `FALSE` and `Random::DEFAULT` are configured.

```ruby
# bad
NIL
TRUE
FALSE
Random::DEFAULT # Return value of Ruby 2 is `Random` instance, Ruby 3.0 is `Random` class.

# good
nil
true
false
Random.new # `::DEFAULT` has been deprecated in Ruby 3, `.new` is compatible with Ruby 2.
```

The following are these sources of information.

- Remove (deprecated) `NIL`, `TRUE`, and `FALSE` ... https://github.com/ruby/ruby/commit/62554ca
- Deprecate `Random::DEFAULT` ... https://bugs.ruby-lang.org/issues/17351

-----------------

* [x] Added tests.


bbatsov:
Nice cop, but it might also be a good idea to add some data about the Ruby version where something got deprecated to the config as well. This will allow us to have more informative messages.

koic:
That's good idea! I added deprecated version for deprecated constants.

lib/rubocop/cop/lint/deprecated_constants.rb
```diff
@@ -0,0 +1,76 @@
...
+
+        def consntant_name(node, nested_constant_name)
+          return nested_constant_name unless node.namespace.const_type?
+
+          consntant_name(node.namespace, "#{node.namespace.short_name}::#{nested_constant_name}")
+        end
+
+        def message(good, bad, deprecated_version)
+          if deprecated_version
+            deprecated_message = ", which is deprecated in Ruby #{deprecated_version}"
```
bbatsov, line 60: "deprecated since" would be a slightly better wording here

koic: I see! Thank your for the suggestion!

config/default.yml
```diff
@@ -1469,6 +1469,25 @@ Lint/DeprecatedClassMethods:
...
+  # Or set `null` to the value if there is unknown deprecated version.
+  #
+  # DeprecatedConstants:
+  #   'DEPRECATED_CONSTANT': ['alternative_value', 'deprecated_version']
+  #
+  DeprecatedConstants:
+    'NIL': ['nil', '2.4']
+    'TRUE': ['true', '2.4']
+    'FALSE': ['false', '2.4']
+    'Random::DEFAULT': ['Random.new', '3.0']
```
koic, line 1489: I'm not sure whether array is good. For example, hash may be preferable. Please let me know if you have feedback 😅

bbatsov: I agree that hash would be better. This would spare us from having to use explicit `null`s.

koic: > This would spare us from having to use explicit `null`s.

Sure! That makes sense :-) I updated it!

bbatsov:
Thanks!

PR #9326, opened 2021-01-03
Add new `Lint/NumberedParameterAssignment` cop

koic:
This cop checks for uses of numbered parameter assignment.
It emulates the following warning in Ruby 2.7:

```console
% ruby -ve '_1 = :value'
ruby 2.7.2p137 (2020-10-01 revision 5445e04352)
[x86_64-darwin19]
-e:1: warning: `_1' is reserved for numbered parameter; consider another name
```

Assiging to numbered parameter (from `_1` to `_9`) cause an error in Ruby 3.0.

```console
% ruby -ve '_1 = :value'
ruby 3.0.0p0 (2020-12-25 revision 95aff21468)
[x86_64-darwin19]
-e:1: _1 is reserved for numbered parameter
```

NOTE: The parametered parameters are from `_1` to `_9`. This cop checks `_10` and
above as well to prevent confusion.

```ruby
# bad
_1 = :value

# good
non_numbered_parameter = :value
```

This cop checks numbered (block) parameter other than `_1` to `_9` (e.g.: `_0`, `_10`),
it is placed in the `Naming` department instead of `Lint` department.

-----------------

* [x] Added tests.


bbatsov:
I like the cop, but I find the name a bit confusing, as the assignment is just the only place where we can check about the naming reliably, but orthogonal to the naming itself. I'm wondering if this shouldn't be incorporated in `Naming/VariableName` instead. If we keep it as a separate cop it's probably more of a Lint cop, than I naming cop, though. Not sure what the others think.

bbatsov:
@rubocop-hq/rubocop-core If anyone has suggestions how to name this better or whether to integrate it into `Naming/VariableName`, please share them with us. :-)

Rakefile
```diff
@@ -114,6 +115,10 @@ task documentation_syntax_check: :yard_for_generate_documentation do
                  # `Lint/CircularArgumentReference` cop's example.
                  elsif cop == RuboCop::Cop::Lint::CircularArgumentReference
                    Parser::Ruby26.new(RuboCop::AST::Builder.new)
+                 # Ruby 3.0 raises an syntax error in
```
jonas054, line 118: `an` -> `a`

jonas054:
@koic You wrote that you put the new cop in `Naming` instead of `Lint`, so it sounds like you first considered `Lint` but then changed your mind. What was your reason for that? To me it seems more natural to put it in `Lint` and to keep it as a separate cop. For one thing, it reports on severity level `Convention` now, while it's syntax error in Ruby 3.0.

koic:
Yeah, what I originally tried to focus on was that assigning to numbered parameters (`_1` to` _9`) will cause syntax error in Ruby 3.0. OTOH, assigning to `_0`, ` _10` doesn't cause syntax error, so I put it in `Naming` department, but this is a side effect 💦  (It's not essentially the cop's aim).

So I reconsidered `Lint` as the comfortable department and updated this PR.  Thank you for your consideration

new_add_new_naming_numbered_parameter_assignment_cop.md
```diff
@@ -0,0 +1 @@
+* [#9999](https://github.com/rubocop-hq/rubocop/pull/9999): Add new `Lint/NumberedParameterAssignment` cop. ([@koic][])
```
jonas054, line 1: I'm not sure what this file is. It shouldn't be here, right?

jonas054 (changes requested):

new_add_new_naming_numbered_parameter_assignment_cop.md
koic: Oops! This is a mistake and I removed this file 💦

jonas054 (approved):

bbatsov:
Let's merge the PR after the next bugfix release. I plan to do it Monday morning.

PR #9338, opened 2021-01-06
Improve comments handling in couple of auto-corrections

### LINKED ISSUE #9233: Wrong auto correction Style/SoleNestedConditional
I got a wrong auto-correction with `Style/SoleNestedConditional` cop and without `Style/IfUnlessModifier` cop.

`.rubocop.yml`
```yaml
AllCops:
  NewCops: disable

Style/IfUnlessModifier:
  Enabled: false
```

`test.rb`
```ruby
# an example file

def func
  # foo
  foo

  # hoge
  if foo
    if bar
      buz
    end
  end
end
```

Run:
```
$ rubocop -a --debug --only Style/SoleNestedConditional test.rb
```

## Expected behavior

```ruby
# an example file

def func
  # foo
  foo

  # hoge
  if foo && bar
    buz
  end
end
```

## Actual behavior

Output:
```
For /Users/thinca/tmp/rubocop-test: configuration from /Users/thinca/tmp/rubocop-test/.rubocop.yml
Default configuration from /Users/thinca/tmp/rubocop-test/vendor/bundle/ruby/2.7.0/gems/rubocop-1.6.1/config/default.yml
Inspecting 1 file
Scanning /Users/thinca/tmp/rubocop-test/test.rb
Loading cache from /Users/thinca/.cache/rubocop_cache/fd9e92937dd32541de61425da45d7dfa0c54bb9e/5c3a0012aae88f33b221aa9bbc6dcfba8f2100c7/9f77710eeb664a239190bf981ba7f5ff6c702a02
C

Offenses:

test.rb:9:5: C: [Corrected] Style/SoleNestedConditional: Consider merging nested conditions into outer if conditions.
    if bar
    ^^

Finished in 0.12683500000275671 seconds
```

`test.rb`
```ruby
# an example file

def func
  # foo
  foo

  # hoge
  # an example file
# foo
# hoge
if foo && bar
      buz
    end
end
```

- Extra comments are appears
- Strange indentation

```
$ mkdir -p ~/tmp/rubocop-test
$ cd ~/tmp/rubocop-test
$ (put .rubocop.yml and test.rb. rubocop is already installed.)
$ rubocop -a --debug --only Style/SoleNestedConditional test.rb
```

## RuboCop version

```
$ rubocop -V
1.6.1 (using Parser 2.7.2.0, rubocop-ast 1.3.0, running on ruby 2.7.2 x86_64-darwin19)
```

Darhazer:
I learned the hard way that the `comments_before_line` won't stop until the very first line in the processed file, and that it doesn't care whether the comments begin their own line or not. So I replaced it in the 2 cops that use it.

In one of them (that uses the CommentsHelp mixin) the code already tried to limit the damage by looking for consecutive lines, so it would only break if there is a node with inline comment and missing separating line. 

The other one however would completely mess the comments, bringing all of the unrelated one to the new position, so I added spec for that case. It also fixes #9233 

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_handling_of_comments_during_autocorrection.md
+++ b/changelog/fix_handling_of_comments_during_autocorrection.md
@@ -1,0 +1,1 @@
+* [#9233](https://github.com/rubocop-hq/rubocop/issues/9233): Fix `Style/SoleNestedConditional` copying non-relevant comments during auto-correction. ([@Darhazer][])
--- a/lib/rubocop/cop/mixin/comments_help.rb
+++ b/lib/rubocop/cop/mixin/comments_help.rb
@@ -22,16 +22,7 @@ module RuboCop
       end
 
       def begin_pos_with_comment(node)
-        annotation_line = node.first_line - 1
-        first_comment = nil
-
-        processed_source.comments_before_line(annotation_line)
-                        .reverse_each do |comment|
-          if comment.location.line == annotation_line
-            first_comment = comment
-            annotation_line -= 1
-          end
-        end
+        first_comment = processed_source.ast_with_comments[node].first
 
         start_line_position(first_comment || node)
       end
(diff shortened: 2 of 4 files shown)


bbatsov:
Looks good! Thanks!

koic:
Thank you @Darhazer!

PR #9339, opened 2021-01-06
[Fix #9337] Add `AllowedIdentifiers` to `Naming/VariableName`.

### LINKED ISSUE #9337: Support AllowedIdentifiers option for Naming/VariableName cop
There exist situations where non-standard naming should be permitted. In our case, we have a `cvss_vectorString` column name that we need to support.

The `Naming/VariableNumber` cop has an `AllowedIdentifiers` configuration option with some sensible defaults based on offenses built into Ruby itself. I suggest adding a similar `AllowedIdentifiers` option to the `Naming/VariableName` cop with a default value of an empty array.

The only other way we've found to work around this issue is to disable the cop for individual offenses as they arise.

Thank you for your work! 💛

dvandersluis:
Also extracts shared logic from `Naming/VariableNumber` to a new `AllowedIdentifiers` mixin, and I updated it to allow global variables to be allowed too. Fixes #9337.

-----------------

* [x] Added tests.


bbatsov:
Nice improvement!

PR #9351, opened 2021-01-08
Fix sole nested conditional handle parens

magneland:
Fix edge case in `Style/SoleNestedConditional` auto-correct.

I think it only applies to an outer `unless` with multiple conditions, but without parens.

The addition of parentheses is correct.
However, the `!` is the wrong place.
It should be before the parentheses.


bbatsov:
Thanks!

PR #9352, opened 2021-01-08
[Fix #9349] Fix a false positive for `Lint/MultipleComparison`

### LINKED ISSUE #9349: Lint/MultipleComparison should not be marked safe
I am using a SQL query dsl with code like this:

```ruby
query.where { date >= range.begin & date < range.end }
```

rubocop autocorrects this to:

```ruby
query.where { date >= range.begin & date && range.begin & date < range.end }
```

which has different semantics.

(see follow-up comment for my original error here^^ 🤦)

--------

## Expected behavior

This cop seems to make wrong assumptions about the safety of operating on relational operators. Whatever the case may be, this should not change my code to something that could be semantically different (by default).

## Actual behavior

It autocorrects to semantically different code as outlined above.

Run `rubocop -a` on source files with the snippets above.

## RuboCop version

```
1.7.0 (using Parser 3.0.0.0, rubocop-ast 1.3.0, running on ruby 2.6.6 x86_64-darwin19)
  - rubocop-performance 1.9.2
  - rubocop-rspec 2.1.0
  - rubocop-rspec 2.1.0
```

Side note: that duplicate `rubocop-rspec` is not a typo, that's what the command outputs. Not sure if there's something wrong there as well or some quirk to my config^

koic:
Fixes #9349.

This PR fixes a false positive for `Lint/MultipleComparison` when using `&`, `|`, and `^` (DSL) set operation operators.

e.g. `Arel::Match`'s DSL methods.
https://github.com/rails/rails/blob/v6.1.1/activerecord/lib/arel/math.rb#L21-L23

I think that the cop can probably allow using these operators by default.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/multiple_comparison.rb
+++ b/lib/rubocop/cop/lint/multiple_comparison.rb
@@ -11,14 +11,10 @@ module RuboCop
       # @example
       #
       #   # bad
-      #
       #   x < y < z
       #   10 <= x <= 20
       #
-      # @example
-      #
       #   # good
-      #
       #   x < y && y < z
       #   10 <= x && x <= 20
       class MultipleComparison < Base
@@ -26,6 +22,7 @@ module RuboCop
 
         MSG = 'Use the `&&` operator to compare multiple values.'
         COMPARISON_METHODS = %i[< > <= >=].freeze
+        SET_OPERATION_OPERATORS = %i[& | ^].freeze
         RESTRICT_ON_SEND = COMPARISON_METHODS
 
         def_node_matcher :multiple_compare?, <<~PATTERN
@@ -34,6 +31,9 @@ module RuboCop
 
         def on_send(node)
           return unless (center = multiple_compare?(node))
+          # It allows multiple comparison using `&`, `|`, and `^` set operation operators.
+          # e.g. `x >= y & y < z`
+          return if center.send_type? && SET_OPERATION_OPERATORS.include?(center.method_name)
 
           add_offense(node) do |corrector|
             new_center = "#{center.source} && #{center.source}"
--- a/changelog/fix_false_positive_for_lint_multiple_comparison.md
+++ b/changelog/fix_false_positive_for_lint_multiple_comparison.md
@@ -1,0 +1,1 @@
+* [#9349](https://github.com/rubocop-hq/rubocop/issues/9349): Fix a false positive for `Lint/MultipleComparison` when using `&`, `|`, and `^` set operation operators in multiple comparison. ([@koic][])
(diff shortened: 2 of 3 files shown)


koic:
And I added a commit to this PR that marks `Lint/MultipleComparison` as unsafe auto-correction because comparison of `1 < 2 < 3` is an error of `true < 3` and is semantically behaviour different from `1 < 2 && 2 < 3`.
https://github.com/rubocop-hq/rubocop/issues/9349#issuecomment-756569781

lib/rubocop/cop/lint/multiple_comparison.rb
```diff
@@ -7,25 +7,26 @@ module Lint
       # multiple value. However, we can't use the comparison in Ruby. However,
       # the comparison is not syntax error. This cop checks the bad usage of
       # comparison operators.
+      # It allows multiple comparison using `&`, `|`, and `^` set operation operators.
```
bbatsov, line 10: I'm a bit confused why this is a problem at all - looking at the defintion of the cop it shouldn't be looking at something like this at all. (unless I'm missing the point here)

lib/rubocop/cop/lint/multiple_comparison.rb
```diff
@@ -7,25 +7,26 @@ module Lint
...
+      #
+      # The cop is marked as unsafe auto-correction because comparison of `1 < 2 < 3` is
```
bbatsov, line 12: I think this example is confusing - I get it, but I don't get how is it relevant to the cop and its safety. It'd be better to illustrate this with some case where it's more obvious that someone intentially wanted a chain a comparsion ops.

lib/rubocop/cop/lint/multiple_comparison.rb
```diff
@@ -34,6 +35,7 @@ class MultipleComparison < Base
 
         def on_send(node)
           return unless (center = multiple_compare?(node))
+          return if center.send_type? && SET_OPERATION_OPERATORS.include?(center.method_name)
```
bbatsov, line 38: I think we'll definitely have to explain here, why those set operations matter at all.

bbatsov:
@koic Don't forget about this PR. Basically right now I'm not quite sure why we need any sort of handling for set operators, so some more explanations would be useful. Is this PR adding support for them just for the sake of Arel's DSL? Shouldn't we make this some configuration in this case - after all if you're not using Arel, treating them as logical operators doesn't make sense.

koic:
I'm sorry I was late to come back to this PR. I rethinked that Arel's specific issue handling can only be configured in RuboCop Rails. Also, it was not good option to mark as unsafe about non-working code `1 > 2 > 3` will work by auto-correction.
So, I have reverted the `SafeAutoCorrect` configuration change in the PR, and leaving only a fix to a false positive for `x> y & y> z` that should not be registered offense.

bbatsov:
Thanks!

PR #9358, opened 2021-01-09
Add restrictive_version_specifiers to Bundler/GemComment

RobinDaugherty:
The problem I'm trying to solve:

There are multiple reasons one might lock the version of a gem in the Gemfile:

1. When a minimum version of the gem is needed to support the project. (Like `">= 2.0"`.)
2. When a specific version of the gem has a known bug. (Should be done like `"!= 2.4.0"`.)
3. When a major version of the gem is incompatible with the project. (Usually looks like `"< 3.0"`.)
4. For no reason. Some gems even provide instructions like "add `gem 'mygem', '~> 2.3'` to your Gemfile".

The last two reasons lead to a project's dependencies falling behind because the `bundle update` process will not update to 3.0 without manual intervention. This would cause dependencies to continue to fall behind and in some cases even limit other non-locked dependencies because of cross-dependencies.

To combat this, good practice would include a periodic cleanup of Gemfile version specifications that limit updating of gems.

Another good practice is a rule that when a gem has a limiting version specifier like items 3 and 4 above, there's a comment in the Gemfile explaining the reason. This allows the periodic cleanup process to be more effective, since one can determine whether a bug is fixed, or know what kind of breakage or test failure to look for when updating the gem past the version specification.

Currently, the `version_specifiers` option for this cop checks a gem specification for any version specifier, requiring a gem comment if one exists. This means that a non-limiting version specification like ">= 1.0" without a justification would warrant a complaint.

This new option, `limiting_version_specifiers`, requires a gem comment only for version specifications that are limiting in nature, but not for other version specifications.

So the following version specifications are permitted without comment:
```rb
gem "rubocop", ">= 1.0"
gem "rubocop", "!= 2.3.4"
```

while the following require an accompanying gem comment:

```rb
gem "rubocop", "< 3.0"
gem "rubocop", ">= 1.0", "< 3.0"
gem "rubocop", "~> 2.0"
gem "rubocop", "= 2.3.5"
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/bundler/gem_comment.rb
+++ b/lib/rubocop/cop/bundler/gem_comment.rb
@@ -43,6 +43,18 @@ module RuboCop
       #   # Version 2.1 introduces breaking change baz
       #   gem 'foo', '< 2.1'
       #
+      # @example OnlyFor: ['limiting_version_specifiers']
+      #   # bad
+      #
+      #   gem 'foo', '< 2.1'
+      #
+      #   # good
+      #
+      #   gem 'foo', '>= 1.0'
+      #
+      #   # Version 2.1 introduces breaking change baz
+      #   gem 'foo', '< 2.1'
+      #
       # @example OnlyFor: ['version_specifiers', 'github']
       #   # bad
       #
@@ -64,6 +76,8 @@ module RuboCop
         MSG = 'Missing gem description comment.'
         CHECKED_OPTIONS_CONFIG = 'OnlyFor'
         VERSION_SPECIFIERS_OPTION = 'version_specifiers'
+        LIMITING_VERSION_SPECIFIERS_OPTION = 'limiting_version_specifiers'
+        LIMITING_VERSION_PATTERN = /<|~>/.freeze
         RESTRICT_ON_SEND = %i[gem].freeze
 
         # @!method gem_declaration?(node)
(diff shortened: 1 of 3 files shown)


bbatsov:
I'm fine with the proposed change, but I think you should expand the cop's documentation to account for its rationale.

lib/rubocop/cop/bundler/gem_comment.rb
```diff
@@ -64,6 +76,8 @@ class GemComment < Base
         MSG = 'Missing gem description comment.'
         CHECKED_OPTIONS_CONFIG = 'OnlyFor'
         VERSION_SPECIFIERS_OPTION = 'version_specifiers'
+        LIMITING_VERSION_SPECIFIERS_OPTION = 'limiting_version_specifiers'
```
bbatsov, line 79: I'm wondering if `restrictive`/`restricting` won't be better wording here.

bbatsov:
@RobinDaugherty ping :-)

lib/rubocop/cop/bundler/gem_comment.rb
RobinDaugherty: Agreed, I'm going to change it to "restrictive version".

bbatsov (approved):

CHANGELOG.md
```diff
@@ -5485,6 +5485,7 @@
 [@wcmonty]: https://github.com/wcmonty
 [@nguyenquangminh0711]: https://github.com/nguyenquangminh0711
 [@chocolateboy]: https://github.com/chocolateboy
+[@RobinDaugherty]: https://github.com/RobinDaugherty
```
bbatsov, line 5488: I think that's not needed, as it will added automatically once we regen the changelog.

bbatsov:

RobinDaugherty:
✅ I've rebased it.

bbatsov:
Thanks!

PR #9359, opened 2021-01-09
[Fix #8281] Improve WhileUntilModifier detection and correction

### LINKED ISSUE #8281: Style/WhileUntilModifier must copy logic from Style/IfUnlessModifier
There's a lot of logic in `IfUnlessModifier` that is missing from `WhileUntilModifier`.

1) Looses comments:

```ruby
while bar # important comment about to be nuked
  baz
end
# corrected to
baz while bar
```

2) Incorrect correction for assignment:

```ruby
foo = while bar
  baz
end
# corrected to
foo = baz while bar
```

This is logic that `IfUnlessModifier` has.

It should be moved into their shared module, and the corresponding tests should be moved in their shared context.

3) Doesn't detect inline cases:

```ruby
while foo; bar; end # => no offense
```

Again, `IfUnless` does the right thing. Same logic should be applied, tests should be moved in their shared context.

Found while checking #8273. See also #8282 and #8283

Darhazer:
The extracted code was already there, it just had to be re-used. Moved relevant tests to the shared examples.

-----------------

* [x] Added tests.


marcandre:
Looks good.
Is the moving of many specs directly related to this change?

marcandre:
(and wouldn't the `if` form also benefit from the same?)

changelog/fix_while_until_modifier_autocorrection.md
```diff
@@ -0,0 +1 @@
+* [#8281](https://github.com/rubocop-hq/rubocop/issues/8281) Fix Style/WhileUntilModifier handling comments and assignment when correcting to modifier form. ([@Darhazer][])
```
bbatsov, line 1: I think you forgot `:` here.

Darhazer: Sorry about that :blush: Fixed

Darhazer:
> Is the moving of many specs directly related to this change?

The change makes those pass. You mentioned in the original ticket that "the corresponding tests should be moved in their shared context", and I agree - the actual code is also in the shared module. Copy/pasting those specs doesn't make sense.

> (and wouldn't the if form also benefit from the same?)

What do you mean?

marcandre:
> > Is the moving of many specs directly related to this change?
> 
> The change makes those pass.

Oh, perfect then. I thought they might have already been passing

> > (and wouldn't the if form also benefit from the same?)
> 
> What do you mean?

There's a spec being added to `while_until_modifier_spec`. Shouldn't that spec be added to `condition_modifier_cop.rb ` instead (and made to pass)?
I.e.
```ruby
# bad
if foo; bar; end
# good
bar if foo
```

Darhazer:
> There's a spec being added to while_until_modifier_spec. Shouldn't that spec be added to condition_modifier_cop.rb instead (and made to pass)?

Sure. Done. Thank you :bow:

marcandre (approved):
LGTM 👍

bbatsov:
Thanks!

PR #9341, opened 2021-01-07
Use IRB instead of Pry

koic:
This PR uses IRB instead of Pry.

Since IRB has many enhancements since Ruby 2.7.
https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/

It replace this PR with the standard REPL IRB.

Since use of IRB is limited to RuboCop development tool, there is no impact on end users.

TIPS: If you cannot do syntax highlighting or multi-line editing with IRB of Ruby 2.7 or higher, please check the following setting.

```ruby
# ~/.irbrc

IRB.conf[:USE_READLINE] = true
```

If `IRB.conf[:USE_READLINE]` is true, either remove it or set to `false`, Reline is used instead of Readline.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [ ] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [ ] Added an entry (file) to the [changelog folder](https://github.com/rubocop-hq/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #9360, opened 2021-01-10
Allow `pry` to be activated in rspec if in the local gemfile.

dvandersluis:
Re: https://github.com/rubocop-hq/rubocop/pull/9341#issuecomment-757504942

#9341 replaced pry with irb but it is useful for debugging. This change will just allow pry to be used in rspec if it's activated from Gemfile.local.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -11,6 +11,12 @@ require 'webmock/rspec'
 
 require_relative 'core_ext/string'
 
+begin
+  require 'pry'
+rescue LoadError
+  # Pry is not activated.
+end
+
 # Require supporting files exposed for testing.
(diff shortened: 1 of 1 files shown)


spec/spec_helper.rb
```diff
@@ -11,6 +11,12 @@
 
 require_relative 'core_ext/string'
 
+begin
+  require 'pry'
+rescue LoadError
+  # pry is not activated
```
koic, line 17: This is a nitpick :-)

```suggestion
  # Pry is not activated.
```

dvandersluis: Fixed!

koic (approved):

koic:
This looks good!

PR #9382, opened 2021-01-15
[Fix #9374] Fix autocorrection for `Layout/LineLength` when the first argument to a send node is a overly long hash pair

### LINKED ISSUE #9374: Really long strings as arguments autocorrect to invalid syntax
```
$ bundle exec rubocop -V
1.8.1 (using Parser 3.0.0.0, rubocop-ast 1.4.0, running on ruby 2.7.2 x86_64-linux)
```

No configuration.

Input file:

```rb
class MyClass
  def show
    render arg1: 'a really long string blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah', arg2: 'blah'
  end
end
```

Use `rubocop -a myfile.rb`. Result:

```rb
class MyClass
  def show
    render 
arg1: 'a really long string blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah', arg2: 'blah'
  end
end
```

Notice `arg1` is now on the next line after the method call.

```
$ ruby -wc myfile.rb
myfile.rb:4: syntax error, unexpected ':', expecting `end'
arg1: 'a really long string blah b...
myfile.rb:4: syntax error, unexpected ',', expecting `end'
...blah blah blah blah blah blah', arg2: 'blah'
```

## Expected behavior

Autocorrected file is valid syntax.

dvandersluis:
When the first argument of a method call is a hash, and the method is not parenthesized, autocorrection would move the argument to a new line, which would cause a syntax error:

```ruby
foo x: a_really_long_value_that_triggers_the_line_length_cop

# becomes

foo
x: a_really_long_value_that_triggers_the_line_length_cop
#=> SyntaxError: unexpected ':', expecting end-of-input
```

Now, if there is more than one pair, the line break will be added after the first argument instead of before it so that the syntax error is avoided.

Fixes #9374.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/layout/line_length_spec.rb
+++ b/spec/rubocop/cop/layout/line_length_spec.rb
@@ -596,6 +596,75 @@ RSpec.describe RuboCop::Cop::Layout::LineLength, :config do
         end
       end
 
+      context 'with a hash with a too long first item' do
+        context 'when parenthesized' do
+          it 'corrects' do
+            expect_offense(<<~RUBY)
+              foo(abc: '10000000000000000000000000000000000000000000000000000', def: '1000')
+                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Line is too long. [78/40]
+            RUBY
+
+            expect_correction(<<~RUBY)
+              foo(
+              abc: '10000000000000000000000000000000000000000000000000000', def: '1000')
+            RUBY
+          end
+        end
+
+        context 'when the hash is parenthesized' do
+          it 'corrects' do
+            expect_offense(<<~RUBY)
+              foo({ abc: '10000000000000000000000000000000000000000000000000000', def: '1000' })
+                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Line is too long. [82/40]
+            RUBY
+
+            expect_correction(<<~RUBY)
+              foo({#{trailing_whitespace}
+              abc: '10000000000000000000000000000000000000000000000000000', def: '1000' })
+            RUBY
+          end
+        end
+
+        context 'when not parenthesized' do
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/layout/line_length_spec.rb
```diff
@@ -596,6 +596,76 @@ def baz(bar)
...
+            it 'moves the 2nd element to a new line' do
+              expect_offense(<<~RUBY)
+                foo abc: '10000000000000000000000000000000000000000000000000000', ghi: '1000'
+                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Line is too long. [77/40]
+              RUBY
+
+              expect_correction(<<~RUBY, loop: false)
+                foo abc: '10000000000000000000000000000000000000000000000000000',#{trailing_whitespace}
+                ghi: '1000'
+              RUBY
```
dvandersluis, line 650: `loop: false` is needed here otherwise the second pass fails with `Correction target buffer 70093973917840 name:"(string)" is not current 70093955918280 name:"(string)" under investigation`. I've tested out the same code directly in rubocop and it doesn't have an issue so I'm not sure what's going on with `expect_correction`.

config/default.yml
```diff
@@ -906,7 +906,7 @@ Layout/LineLength:
   StyleGuide: '#max-line-length'
   Enabled: true
   VersionAdded: '0.25'
-  VersionChanged: '1.4'
+  VersionChanged: <<next>>
```
bbatsov, line 909: That's not needed in the case of a bugfix.

dvandersluis:
The jruby failure appears to be unrelated.

koic:
Yeah, I rerun the failed test and it succeeded. Thank you.
rubocop-circleci

dvandersluis:
Thanks @koic!

PR #9386, opened 2021-01-17
[Fix #5388] Add new `Style/UnlessLogicalOperators` cop.

### LINKED ISSUE #5388: Feature request: Checks that `unless` is not used with multiple conditions.
I was worried about the poor readability of code using `unless` with multiple conditions.

```ruby
unless foo && bar
  something
end
```

Looking at the code like the above, I first think about De Morgan's law, rewrite it and proceed with implementation. This is a very painful task.

In order to prevent this problem, I created a cop called `Lint/UnlessMultipleCondtions`.
https://github.com/wata727/rubocop/commit/da4ed994d5fd263c4e08d4df98ea6cd716244ca3

However, when this Cop is executed to RuboCop, the result is as follows:

```
...

lib/rubocop/config_loader.rb:128:23: W: Lint/UnlessMultipleConditions: Avoid using unless with multiple conditions.
        return unless hash['AllCops'] && hash['AllCops'][version]
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1101 files inspected, 140 offenses detected
RuboCop failed!
```

I don't mind fixing these problems, but I doubt if this fix will be accepted by the community.

Please tell me your opinion :)

caalberts:
[Fix #5388] Add new `Style/UnlessLogicalOperators` cop.

This cop checks for the use of `&&` and `||` in an unless condition.
-----------------

* [x] Added tests.


bbatsov:
It seems to me it'd better if this cop had some degree of configuration - e.g. it could enforce only simple conditions (no logical operators) or allow something like the currently suggested default.

caalberts:
@bbatsov I agree it would be better to have configurable styles. I just wasn't sure what are the options to go with, so I started with this one first.

Do you have suggestions on what to name the options?

caalberts:
@bbatsov I added the 2 configurations suggested. What do you think?

config/default.yml
```diff
@@ -4700,6 +4700,16 @@ Style/UnlessElse:
   Enabled: true
   VersionAdded: '0.9'
 
+Style/UnlessLogicalOperators:
+  Description: >-
+                 Checks for use of logical operators in an unless condition.
+  Enabled: pending
```
koic, line 4706: It seems too strict to enable this rule by default in future (or `NewCops: enable`). So I think this cop should be disabled by default.

```suggestion
  Enabled: false
```

caalberts: Thanks. I was hoping to get feedback on this.

config/default.yml
```diff
@@ -4700,6 +4700,16 @@ Style/UnlessElse:
...
   VersionAdded: '0.9'
 
+Style/UnlessLogicalOperators:
+  Description: >-
+                 Checks for use of logical operators in an unless condition.
+  Enabled: false
+  VersionAdded: '<<next>>'
+  EnforcedStyle: no_mixed_logical_operator
+  SupportedStyles:
+    - no_mixed_logical_operator
```
bbatsov, line 4710: I'd probably go with names like `forbid_logical_operators` and `forbid_mixed_logical_operators` or something along those lines. Such names seem clearer to me.

lib/rubocop/cop/style/unless_logical_operators.rb
```diff
@@ -0,0 +1,57 @@
...
+      #   return unless a && b || c
+      #   return unless a || b || c
+      #   return unless a && b && c
+      #
+      #   # good
+      #   return unless a?
+      class UnlessLogicalOperators < Base
+        include ConfigurableEnforcedStyle
+
+        NO_MIXED_LOGICAL_OPERATOR_MSG = 'Do not use both && and || in an unless condition.'
```
bbatsov, line 30: It's best to put the operators in backticks, so they'd be coloured in the terminal output.

bbatsov:
I don't see any tests/mentions of `and` and `or`. Obviously we have to cover those as well.

bbatsov:
@caalberts ping :-)

config/default.yml
caalberts: Sounds good to me. I'll update the PR.

caalberts:
> tests/mentions of and and or

@bbatsov that's a great catch! Thanks for highlighting. I'll update the PR with the necessary changes.

lib/rubocop/cop/style/unless_logical_operators.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        def mixed_logical_operator?(node)
+          or_with_and?(node) ||
+            and_with_or?(node) ||
+            mixed_precedence_and?(node) ||
+            mixed_precedence_or?(node)
+        end
+
+        def mixed_precedence_and?(node)
+          node.source.include?('&&') && node.source.include?('and')
```
caalberts, line 70: I'm not sure if there is a better way to distinguish `a && b && c` and `a && b and c`.

Using node matcher doesn't work because both have the following AST:
```
(and
  (and
    (send nil :a)
    (send nil :b))
  (send nil :c))
```

Same goes for `a || b || c` and `a || b or c`.

caalberts: I considered using a node matcher for `a and b && c`, using `(if (and ... (and ...)) ...)` to match the following AST:
```
(and
  (send nil :a)
  (and
    (send nil :b)
    (send nil :c)))
```

However, if we need to check the `node.source` anyway, then it's unnecessary to have the node matcher.

caalberts:
@bbatsov Could you take a look again please?

config/default.yml
```diff
@@ -4700,6 +4700,16 @@ Style/UnlessElse:
...
   VersionAdded: '0.9'
 
+Style/UnlessLogicalOperators:
+  Description: >-
+                 Checks for use of logical operators in an unless condition.
+  Enabled: false
+  VersionAdded: '<<next>>'
+  EnforcedStyle: forbid_mixed_logical_operator
+  SupportedStyles:
+    - forbid_mixed_logical_operator
```
bbatsov, line 4710: Please, make those plural (operators).

lib/rubocop/cop/style/unless_logical_operators.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # This cop checks for the use of logical operators in an unless condition.
```
bbatsov, line 6: You can add a bit of rationale here - e.g. this cop exists to discourage hard to read/process usages of `unless`.

lib/rubocop/cop/style/unless_logical_operators.rb
```diff
@@ -0,0 +1,79 @@
...
+      #   return unless a or b
+      #   return unless a and b
+      #
+      #   # good
+      #   return unless a
+      #   return unless a?
+      class UnlessLogicalOperators < Base
+        include ConfigurableEnforcedStyle
+
+        FORBID_MIXED_LOGICAL_OPERATOR = 'Do not use mixed logical operators in an unless condition.'
```
bbatsov, line 37: Put `unless` in backticks please.

lib/rubocop/cop/style/unless_logical_operators.rb
```diff
@@ -0,0 +1,79 @@
...
+        FORBID_LOGICAL_OPERATOR = 'Do not use a logical operator in an unless condition.'
```
bbatsov, line 38: any logical operators

lib/rubocop/cop/style/unless_logical_operators.rb
bbatsov: That's ok. Technically you don't need to use `include?`, though as you can do an equality check instead (and I guess it'd be faster).

lib/rubocop/cop/style/unless_logical_operators.rb
bbatsov: In the rationale you should also briefly explain the purpose of the two supported styles.

bbatsov:
Just added a couple more small remarks.

lib/rubocop/cop/style/unless_logical_operators.rb
caalberts: Could you share an example how we can do an equality check on this?

bbatsov: Isn't this just a regular string? How about `==`? :-)

bbatsov (approved):

PR #9391, opened 2021-01-19
[Fixes #9340] Added `info` Severity level to allow offenses to be listed but not return a non-zero error code

### LINKED ISSUE #9340: There should be an `info` severity level
There should be an `info` severity level for things that are worth pointing out but are not show-stoppers for our codebase. I don't really understand the current severity levels in RuboCop and how they should be used, and the thought process behind them doesn't seem to be documented.

For JavaScript I use ESLint which has 3 levels:
- `off`
- `warn`
- `error`

With ESLint, I set rules such as line length ([max-len](https://eslint.org/docs/rules/max-len)) to 'warn' because they're helpful to see (i.e. in your editor) but not critical to fix.

Here are RuboCop's severity levels, in order:
- `refactor`
- `convention` - default for most cops
- `warning`
- `error`
- `fatal`

If I want the `Layout/LineLength` cop notify (i.e. in my editor) but not cause RuboCop to fail (i.e. return code != 0), I'd have to either:
- Set the severity for that cop to `refactor`, a name which makes no sense
- [what I want] Be able to set some cops to severity of `info`, a level which currently doesn't exist
- Set the default severity for all cops to `error` ([not possible](https://github.com/rubocop-hq/rubocop/issues/1624)), then set specific cops I want to warn to `warning` and run rubocop with `--fail-level E`

Maybe I'm misunderstanding how the `warning` severity level should be used. It seems `warning` should be lower than the default, but it's not. Before I dug into the docs, I assumed setting a cop to `warning` was "backing it off" from the default of `convention`, but that's not true with how Rubocop's severity levels work.

Thanks for the time you take to work on RuboCop. 🙏

dvandersluis:
Cops can now be configured to have `Severity: info` which will cause offenses to be reported but not cause the `rubocop` command to return `1` if there are any. It can be overridden by `--fail-level` and works with `--display-only-fail-level-offenses`.

Fixes #9340.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_added_info_severity_level_to_allow.md
+++ b/changelog/new_added_info_severity_level_to_allow.md
@@ -1,0 +1,1 @@
+* [#9340](https://github.com/rubocop-hq/rubocop/pull/9340): Added `info` Severity level to allow offenses to be listed but not return a non-zero error code. ([@dvandersluis][])
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -498,8 +498,11 @@ Style/Alias:
 Each cop has a default severity level based on which department it belongs
 to. The level is normally `warning` for `Lint` and `convention` for all the
 others, but this can be changed in user configuration. Cops can customize their
-severity level. Allowed values are `refactor`, `convention`, `warning`, `error`
+severity level. Allowed values are `info`, `refactor`, `convention`, `warning`, `error`
 and `fatal`.
+
+Cops with severity `info` will be reported but will not cause `rubocop` to return
+a non-zero value.
 
 There is one exception from the general rule above and that is `Lint/Syntax`, a
 special cop that checks for syntax errors before the other cops are invoked. It
--- a/lib/rubocop/cop/severity.rb
+++ b/lib/rubocop/cop/severity.rb
@@ -6,10 +6,10 @@ module RuboCop
     class Severity
       include Comparable
 
-      NAMES = %i[refactor convention warning error fatal].freeze
+      NAMES = %i[info refactor convention warning error fatal].freeze
 
       # @api private
-      CODE_TABLE = { R: :refactor, C: :convention,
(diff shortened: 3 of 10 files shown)


tas50:
@dvandersluis Thanks for putting this together. This is going to be very useful for us in cookstyle. Right now we have to override the fail level to be 1 level higher so we can do a similar thing: https://github.com/chef/cookstyle/blob/master/bin/cookstyle#L10

I look forward to removing that hack from our codebase.

bbatsov:
Let's think a bit more on this, before making any changes. In general I'm not very happy with the existing severity levels and how we're using them, so let's discuss a bit more our options before adding anything new.

bbatsov:
As an example - perhaps better severities would be "low", "medium", "high", "error". At least semantically this sounds more like severities to me. :-)

dvandersluis:
@bbatsov sure, I just wanted to get something in place we can discuss on. I think it makes sense to have a `info` level in general, because the semantics of it are different than `low`, `medium`, etc. Otherwise, I think it's a good idea to simplify the severity levels, but if we change them we're going to have to map the old values because otherwise a lot of people are going to get errors that their configurations are invalid 😄

phallstrom:
For me, the most important use case is the ability to mark cops as informational so I see them in my editor integrations, but that they don't fail the build -- and most critically don't make me have to specify `--fail-level` on the CLI. Either make `info` not fail the build by default, or let me specify `FailLevel` in the yml config.

bbatsov:
> Either make info not fail the build by default, or let me specify FailLevel in the yml config.

Hmm, I thought we had already done this, but I guess we haven't.

bbatsov:
I've decided to merge this, as it doesn't add a lot of complexity and solves a problem some people have today. We can create a separate task to discuss simplifying/improving the severity concept in general. Perhaps just removing a couple of odd levels like `refactor` and `error` or `fatal` would be enough, but we need to think more about this.

PR #9396, opened 2021-01-20
Add new `Style/IfWithBooleanLiteralBranches` cop

koic:
This PR adds new `Style/IfWithBooleanLiteralBranches` cop that checks for redundant `if` with boolean literal branches.

It checks only conditions to return boolean value (`true` or `false`) for safe detection.
The conditions to be checked are comparison methods, predicate methods, and double negative.

```ruby
# bad
if foo == bar
  true
else
  false
end

# bad
foo == bar ? true : false

# good
foo == bar
```

And this PR auto-corrected the following offense by the new cop.

```console
% bundle exec rake
(snip)

Offenses:

lib/rubocop/cop/style/ternary_parentheses.rb:123:47: C: [Correctable]
Style/IfWithBooleanLiteralBranches: Remove redundant ternary operator
with boolean literal branches.
            non_complex_expression?(condition) ? false : true
                                              ^^^^^^^^^^^^^^^

1246 files inspected, 1 offense detected, 1 offense auto-correctable
RuboCop failed!
```

-----------------

* [x] Added tests.


lib/rubocop/cop/style/if_with_boolean_literal_branches.rb
```diff
@@ -0,0 +1,94 @@
...
+      #
+      class IfWithBooleanLiteralBranches < Base
+        extend AutoCorrector
+
+        MSG = 'Remove redundant %<keyword>s with boolean literal branches.'
+
+        def_node_matcher :if_with_boolean_literal_branches?, <<~PATTERN
+          {
+            (if #return_boolean_value? (true) (false))
+            (if #return_boolean_value? (false) (true))
```
marcandre, line 32: You can use `(if #return_boolean_value? {(true) (false) | (false) (true)}) `

spec/rubocop/cop/style/if_with_boolean_literal_branches_spec.rb
```diff
@@ -0,0 +1,419 @@
...
+          true
+        else
+          false
+        end
+      RUBY
+    end
+
+    it 'does not register an offense when using `if foo? && bar && baz?` with boolean literal branches' do
+      expect_no_offenses(<<~RUBY)
+        if foo? && bar && baz?
```
marcandre, line 319: Why no offense here? I understand not wanting to inverse a complex condition, but to me anything not opposite is better simplified, no?

marcandre (commented):
That's a great cop imo

lib/rubocop/cop/style/if_with_boolean_literal_branches.rb
koic: Cool ❤️

spec/rubocop/cop/style/if_with_boolean_literal_branches_spec.rb
koic: It's my mistake. 💦 I refined the conditions of `&&` and `||` again. Updated implementation registers an offense if it can be assumed that a boolean value will be returned.

spec/rubocop/cop/style/if_with_boolean_literal_branches_spec.rb
```diff
@@ -0,0 +1,511 @@
...
+      RUBY
+
+      expect_correction(<<~RUBY)
+        foo? && bar && baz?
+      RUBY
+    end
+
+    it 'does not register an offense when using `if foo? || bar || baz?` with boolean literal branches' do
+      expect_no_offenses(<<~RUBY)
+        if foo? || bar || baz?
```
marcandre, line 336: Why do you prefer not simplifying this one?

koic: This is because the return value changes from `true` to `"hi"` if `bar` is not a boolean value. Example:

```ruby
def foo?
  false
end

def bar
  "hi"
end

def baz?
  false
end

foo? || bar || baz? ? true : false #=> true
foo? || bar || baz? #=> "hi"
```

So this aims for safe detection, which keeps the return value as a boolean value.

bbatsov:
Great addition as well! Technically speaking we can extend this to cover `case` as well, but I guess it's quite uncommon to write such code.

spec/rubocop/cop/style/if_with_boolean_literal_branches_spec.rb
marcandre: Ah, got it. Thanks @koic

PR #9397, opened 2021-01-21
Add documentation for Layout/MultilineAssignmentLayout `SupportedTypes`

Bhacaz:
I was looking for exactly the configurable attributes `SupportedTypes` for the cop Layout/MultilineAssignmentLayout.  But, I had to look in the code to found out this attribute and in `config/default.yml` to found the available parameters. I thought it might be useful to add this info in the documentation. 

UPDATE

To make `SupportedTypes` appear in the documentation, I had to change the name of the attribute, I named it `AssignmentTypes`.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/multiline_assignment_layout.rb
+++ b/lib/rubocop/cop/layout/multiline_assignment_layout.rb
@@ -31,6 +31,32 @@ module RuboCop
       #   foo = if expression
       #     'bar'
       #   end
+      #
+      # @example SupportedTypes: ['block', 'case', 'class', 'if', 'kwbegin', 'module'] (default)
+      #   # good
+      #   foo =
+      #     if expression
+      #       'bar'
+      #     end
+      #
+      #   # good
+      #   foo =
+      #     [1].map do |i|
+      #       i + 1
+      #     end
+      #
+      # @example SupportedTypes: ['block']
+      #   # good
+      #   foo = if expression
+      #     'bar'
+      #   end
+      #
+      #   # good
+      #   foo =
+      #     [1].map do |i|
+      #       'bar' * i
+      #     end
+      #
       class MultilineAssignmentLayout < Base
         include CheckAssignment
         include ConfigurableEnforcedStyle


bbatsov:
The documentation is auto-generated from the cop source, so you'll actually have to update it instead.

Bhacaz:
@bbatsov 

> The documentation is auto-generated from the cop source, so you'll actually have to update it instead.

Sorry I totally missed that. But now I understand why the attribute wasn’t present in the documentation.

--- 

The reason `SupportedStyles` wasn't appearing in the documentation is because of this line: https://github.com/rubocop-hq/rubocop/blob/bfb6a0607a19531ac8e1f20fe27a30a9da29dede/lib/rubocop/cops_documentation_generator.rb#L120

I don't know if was on purpose to name it that way (I don't think so), but I think it will be interesting to add documentation about that config attribute.

I had two choice that imply to change the config name used by the cop:
1. Use the pattern `EnforcedTypes`/`SupportedTypes`. Which required to specified every possible supported values and use  `EnforcedTypes` in the code instead of `SupportedTypes`.
2. Change the config name, `AssignmentTypes` and the docs will only specified that it required an `Array`.

bbatsov:
> I don't know if was on purpose to named it that way (I don't think so), but I think it will be interesting to add documentation about that config attribute.

@koic Do you remember why we did it like this? I don't want to change the attribute name just to fix the docs, as that'd be a breaking change.

koic:
I agree with @bbatsov. I also think that renaming option is a breaking change, so it's better to only update the documentation for `SupportedTypes` now.

JFYI, It seems that the option has been named `SupportedTypes` for a long time.
https://github.com/rubocop-hq/rubocop/pull/2361

Bhacaz:
@koic 
I also think that a breaking change is too must to fix the documentation. What I can see, is to have a complet documentation for that attribute it might required that. The table of _Configurable attribute_ looks for the pattern `Supported*/Enforced*` names and to configure that attribute in a `.rubocop.yml` file we must used `EnforcedTypes` (instead of the actual `SupportedStyles`).

I have a suggestion, I can only add example in the cop code for that attribute and it will be shown in the documentation, but `SupportedTypes` will be missed in the table of _Configurable attribute_.

<details>

image

</details>

Let me know what you think.

koic:
I see. It's not perfect as configuration table, but I think it can be merged as it  is as a first step (I think that examples has been improved by this patch). And I think RuboCop 2.0 is an opportunity for breaking change to the better option name.

changelog/change_add_supported_types_documentation_multiline_assignment.md
```diff
@@ -0,0 +1 @@
+* [#9397](https://github.com/rubocop-hq/rubocop/issues/9397): Update Layout/MultilineAssignmentLayout examples to add `SupportedTypes` attribute in documentation. ([@bhacaz][])
```
koic, line 1: Can you remove the entry as there is no change in product behavior to users?

koic:
Thanks!

PR #9398, opened 2021-01-21
[Fix #7766] Rename rescue body vars when renaming exception name

### LINKED ISSUE #7766: [Corrected] Naming/RescuedExceptionsVariableName: Use e instead of error.
Ruby 2.7.0
Rubocop 0.80.0

I have some code:

```ruby
# frozen_string_literal: true

def main
  raise
rescue StandardError => error
  error = {
    error_message: error.message
  }
  puts error
end
```

I run CLI rubocop with argument `--auto-correct`:

```bash
$ rubocop --auto-correct main.rb
```

↓

```
Inspecting 1 file
W

Offenses:

main.rb:5:25: C: [Corrected] Naming/RescuedExceptionsVariableName: Use e instead of error.
rescue StandardError => error
                        ^^^^^
main.rb:6:3: W: Lint/UselessAssignment: Useless assignment to variable - error.
  error = {
  ^^^^^

```

Rubocop corrected my code, renamed rescued exception variable name from `error` to `e`.
But this correcting is changed code logic.

Before use Rubocop method `main()` return Hash: `{ error: { error_message: String} }`.
After use return `RuntimeError`.

asterite:
Fixes #7766

The old code used `node.body&.each_descendant(:lvar)` to replace the exception variable. The problem is, that doesn't work if the rescue body is a single variable with the name of the exception variable. Using `each_node` will consider that case too.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Bug fixes
 
+* [#7766](https://github.com/rubocop-hq/rubocop/issues/7766): Rename rescue body vars when renaming exception name. ([@asterite][])
 * [#9342](https://github.com/rubocop-hq/rubocop/issues/9342): Fix an error for `Lint/RedundantDirGlobSort` when using `collection.sort`. ([@koic][])
 * [#9304](https://github.com/rubocop-hq/rubocop/issues/9304): Do not register an offense for `Style/ExplicitBlockArgument` when the `yield` arguments are not an exact match with the block arguments. ([@dvandersluis][])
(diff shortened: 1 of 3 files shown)


marcandre:
This PR is good, but it only solves this example, not the one in #7766 (needs to also check for `lvasgn`)

asterite:
Oh, I thought the issue was only when the body was a single variable. That's why I didn't fix the other one. I don't think I'll have time to do that, sorry. But the fix should be similar to this PR's code.

dvandersluis:
@marcandre I fixed that example in #9400.

PR #9401, opened 2021-01-21
Improve project_spec tests when expected configuration for a cop is missing.

dvandersluis:
Small test update to show better error messages when configuration is missing for a (new) cop, rather than failing.

### DIFF
--- a/spec/project_spec.rb
+++ b/spec/project_spec.rb
@@ -40,17 +40,18 @@ RSpec.describe 'RuboCop Project', type: :feature do
 
     it 'has a nicely formatted description for all cops' do
       cop_names.each do |name|
-        description = config[name]['Description']
-        expect(description.nil?).to be(false)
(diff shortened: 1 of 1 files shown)


spec/project_spec.rb
```diff
@@ -70,8 +71,9 @@ def expected
 
     it 'has a period at EOL of description' do
       cop_names.each do |name|
-        description = config[name]['Description']
+        next unless config[name]
 
+        description = config[name]['Description']
```
koic, line 76: Others have been replaced with `dig`, is there any reason why this was not replaced?

dvandersluis: Because there's no point in having two tests fail for there being no description IMO. The description one will fail if there is no description and this one will be skipped.

bbatsov:
Thanks!

PR #9402, opened 2021-01-21
Add new `Lint/TripleQuotes` cop

dvandersluis:
Follows https://github.com/rubocop-hq/rubocop/issues/9394#issuecomment-763899981.

Looks for instances of triple (or quintuple, or septuple, etc.) quotes and registers an offense for them as they do not have a special meaning in Ruby but could cause confusion that they do.

The case of useless implicit concatenation (ie. `'a' 'b' 'c'`, `'' ''`, etc.) is not registered by this cop to keep things simple (but I'm going to look at adding a cop for that as well).

-----------------

* [x] Added tests.


dvandersluis:
On further thought, maybe it might be worthwhile to handle triple quotes and other implicit string concatenation in the same `Lint/ImplicitStringContatenation` cop...

lib/rubocop/cop/lint/triple_quotes.rb
```diff
@@ -0,0 +1,67 @@
...
+      #
+      #   # bad
+      #   '''
+      #     A string
+      #   '''
+      #
+      #   # good
+      #   'A string'
+      #
+      #   # good
```
marcandre, line 33: We should state that multiline singlequoted strings are ok:
```ruby
# ok
"
  A string
"
```

marcandre (commented):
LGTM, just a small comment

marcandre:
> On further thought, maybe it might be worthwhile to handle triple quotes and other implicit string concatenation in the same `Lint/ImplicitStringContatenation` cop...

Either way works for me. Separate cop makes sense though as `ImplicitStringContatenation` could be disabled, for example, but triple quotes is kind of different.

lib/rubocop/cop/lint/triple_quotes.rb
dvandersluis: Makes sense! Added it, thanks.

lib/rubocop/cop/lint/triple_quotes.rb
```diff
@@ -0,0 +1,72 @@
...
+      #     A string
+      #   """
+      #
+      #   # bad
+      #   '''
+      #     A string
+      #   '''
+      #
+      #   # good
+      #   'A string'
```
koic, line 31: It may be better to mention that this representation is incompatible with the bad case for line breaks and indentation.

dvandersluis: Thanks I rearranged the examples to be clearer.

lib/rubocop/cop/lint/triple_quotes.rb
```diff
@@ -0,0 +1,72 @@
...
+
+        MSG = 'Delimiting a string with multiple quotes has no effect, use a single quote instead.'
+
+        def on_dstr(node)
+          return if (empty_str_nodes = empty_str_nodes(node)).none?
+
+          opening_quotes = node.source.scan(/(?<=\A)['"]*/)[0]
+          return if opening_quotes.size < 3
+
+          # If the entire `dstr` is empty `str` nodes, leave one so that there is
```
pirj, line 53: What do you think of "If `dstr` node is composed of empty strings only, keep one"?

spec/rubocop/cop/lint/triple_quotes_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+        STRING
+      RUBY
+    end
+  end
+
+  it 'does not register an offense for implicit concatenation' do
+    expect_no_offenses(<<~RUBY)
+      '' ''
+      'a''b''c'
+      'a''''b'
```
pirj, line 110: I believe this one line is offensive as well, but it's not a `dstr`, and not such a big deal anyway.

pirj (approved):
Incredible, thank you!

lib/rubocop/cop/lint/triple_quotes.rb
dvandersluis: Sounds good!

spec/rubocop/cop/lint/triple_quotes_spec.rb
dvandersluis: Yeah it's not technically a "triple quote" so I didn't want to cover it if we're going to have a separate cop for these situations (as discussed above), but I wanted it to be in the tests for clarity there.

lib/rubocop/cop/lint/triple_quotes.rb
dvandersluis: Updated, I worded it _slightly_ differently but I think it's clearer now.

koic (approved):
This looks good to me.

bbatsov:
Nicely done!

PR #9405, opened 2021-01-22
Improve documentation for Style/EvalWithLocation cop

taichi-ishitani:
This PR is to improve documentation for `Style/EvalWithLocation` cop.
There are no description and example for the case that a string variable is given as a code string like below.

```ruby
code = <<-RUBY
def do_something
end
RUBY

eval code
```

I added description and example to make this case clear.

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/change_improve_documentation_for_style_eval_with_location_cop.md
+++ b/changelog/change_improve_documentation_for_style_eval_with_location_cop.md
@@ -1,0 +1,1 @@
+* [#9405](https://github.com/rubocop-hq/rubocop/pull/9405): Improve documentation for `Style/EvalWithLocation` cop. ([@taichi-ishitani][])
--- a/lib/rubocop/cop/style/eval_with_location.rb
+++ b/lib/rubocop/cop/style/eval_with_location.rb
@@ -31,6 +31,18 @@ module RuboCop
       #     def do_something
       #     end
       #   RUBY
+      #
+      # This cop works only when a string literal is given as a code string.
+      # No offence is reported if a string variable is given as below:
+      #
+      # @example
+      #   # not checked
+      #   code = <<-RUBY
+      #     def do_something
+      #     end
+      #   RUBY
+      #   eval code
+      #
       class EvalWithLocation < Base
         MSG = 'Pass `__FILE__` and `__LINE__` to `eval` method, ' \
               'as they are used by backtraces.'


bbatsov:
Hmm, I wonder why we'd treat this case differently. Perhaps an oversight on our part and we should treat a variable param just a string literal. @dvandersluis what do you think?

taichi-ishitani:
I think we cannot know where the given string variable is defined so the cop does not check this case.

bbatsov:
This is true about a string that came externally, but for any other string parameter it's pretty much the same as for an explicit string literal IMO. E.g. in your example you can supply the location metadata.

taichi-ishitani:
Ah, I understand. You're thinking the cop should check as much as possible.
Do following examples match your thought?

Examples which the cop should check:

```ruby
def eval_string
  code = <<~CODE
    def do_something
    end
  CODE
  eval(code) # report offense
end

def eval_string_from_file(file)
  code = File.read(file)
  eval(code, file, 1)
end
```

Examples which the cop does not check:

```ruby
def eval_string(s)
  eval(s)
end

def eval_code
  eval(@code)
end
```

dvandersluis:
@taichi-ishitani I think what @bbatsov is getting at is that your examples are still calls to `eval` without a file and location given.

In other words, your `eval_string` method would be better done as:

``` ruby
def eval_string
  code = <<~CODE
    def do_something
    end
  CODE
  eval(code, binding, __FILE__, __LINE__) # this line is changed
end
```

We don't need to know what the code is doing for this cop, just that eval is called without the proper arguments.

However, I'm not sure if we should autocorrect when not given a string, because the cop sets a value relative to `__LINE__` and as you said we don't know where `code` comes from (unless we always just set it to `__LINE__` with no offset when not given a string).

bbatsov:
Yeah, that's exactly what I was referring to. Thanks for explaining it much better than me! :D

taichi-ishitani:
```ruby
def eval_string
  code = <<~S


    foo
  S
  instance_eval(code, __FILE__, __LINE__)
end
eval_string
```

We can get following back traces when we invoke the above code:

```
$ ruby with_location_info.rb
with_location_info.rb:9:in `eval_string': undefined local variable or method `foo' for main:Object (NameError)
        from with_location_info.rb:7:in `instance_eval'
        from with_location_info.rb:7:in `eval_string'
        from with_location_info.rb:9:in `<main>'
```

The 1st line of the back trace shows incorrect location.
Therefore, I think we should not set the location info if we cannot know where a string variable is defined correctly.

bbatsov:
Fair enough. I think updating the docs is the simplest thing we can do right now. Thanks!

PR #9413, opened 2021-01-24
[Fix #7902] Change `Link/NumberConversion` to detect symbol form of conversion methods

### LINKED ISSUE #7902: Lint/NumberConversion doesn't handle `to_i` in symbol form

tejasbubane:
Closes #7902

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_number_conversion_to_handle_symbols.md
+++ b/changelog/change_number_conversion_to_handle_symbols.md
@@ -1,0 +1,1 @@
+* [#7902](https://github.com/rubocop-hq/rubocop/issues/7902): Change `Lint/NumberConversion` to detect symbol form of conversion methods. ([@tejasbubane][])
--- a/lib/rubocop/cop/lint/number_conversion.rb
+++ b/lib/rubocop/cop/lint/number_conversion.rb
@@ -25,12 +25,18 @@ module RuboCop
       #   '10'.to_i
       #   '10.2'.to_f
       #   '10'.to_c
+      #   ['1', '2', '3'].map(&:to_i)
+      #   foo.try(:to_f)
+      #   bar.send(:to_c)
       #
       #   # good
       #
       #   Integer('10', 10)
       #   Float('10.2')
       #   Complex('10')
+      #   ['1', '2', '3'].map { |i| Integer(i, 10) }
+      #   foo.try { |i| Float(i) }
+      #   bar.send { |i| Complex(i) }
       #
       # @example IgnoredMethods: [minutes]
       #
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/lint/number_conversion.rb
```diff
@@ -52,22 +52,33 @@ class NumberConversion < Base
...
+          (send $_ ${#{METHODS}})
+        PATTERN
+
+        def_node_matcher :to_method_symbol, <<~PATTERN
+          {(send _ $_ ${(sym ${#{METHODS}})} ...)
+           (send _ $_ ${(block_pass (sym ${#{METHODS}}))} ...)}
         PATTERN
 
         def on_send(node)
+          handle_to_method(node)
```
bbatsov, line 69: Why not "handle_conversion_method"(as_symbol)? I find the current naming confusing.

changelog/change_number_conversion_to_handle_symbols.md
```diff
@@ -0,0 +1 @@
+* [#7902](https://github.com/rubocop-hq/rubocop/issues/7902): Change `Link/NumberConversion` to detect symbol form of conversion methods. ([@tejasbubane][])
```
bbatsov, line 1: Link or Lint? ;-)

lib/rubocop/cop/lint/number_conversion.rb
```diff
@@ -76,13 +87,31 @@ def on_send(node)
...
+        end
 
         def correct_method(node, receiver)
           format(CONVERSION_METHOD_CLASS_MAPPING[node.method_name],
                  number_object: receiver.source)
         end
 
+        def correct_sym_method(to_method)
+          body = format(CONVERSION_METHOD_CLASS_MAPPING[to_method], number_object: 'i')
+          "{ |i| #{body} }"
```
bbatsov, line 112: I'd probably suggest here something like `&method(:Integer)` as it's a bit more compact and closer to their original code.

bbatsov:
You also need to update the cop documentation to reflect the extended capabilities of the cop.

lib/rubocop/cop/lint/number_conversion.rb
tejasbubane: For `Integer` we pass in base 10 as second argument. We cannot do that with `&method(:Integer)` right?

tejasbubane: Should we use explicit block for `Integer` and method-to-proc for `Float` and `Complex`?

tejasbubane:
@bbatsov Comments resolved - except one comment.

lib/rubocop/cop/lint/number_conversion.rb
bbatsov: I guess it'd best to stay uniform with those. You can keep them as blocks.

bbatsov:
Thanks!

PR #9419, opened 2021-01-26
[Fix #9415] Change `Layout/ClassStructure` to detect inline modifiers

### LINKED ISSUE #9415: False negatives in Layout/ClassStructure with inline access modifiers
The `Layout/ClassStructure` cop fails to surface private/public method ordering violations when inline access modifiers are used. Given the file `ab.rb` with:

```
class A
  private def foo
  end

  public def bar
  end
end

class B
  private
  def foo
  end

  public
  def bar
  end
end
```
--------

## Expected behavior

`Layout/ClassStructure` should be agnostic to visibility modifier and flag both definitions of `bar` (`A#bar` and `B#bar`).

## Actual behavior
Only `B#bar` is flagged:
```
rubocop ab.rb --only Layout/ClassStructure
Inspecting 1 file
C

Offenses:

ab.rb:15:3: C: [Correctable] Layout/ClassStructure: public_methods is supposed to appear before private_methods.
  def bar ...
  ^^^^^^^

```

☝️ 

## RuboCop version

```
1.8.1 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 3.0.0 x86_64-darwin20)
  - rubocop-performance 1.9.2
  - rubocop-rake 0.5.1
  - rubocop-rspec 2.1.0
  - rubocop-sorbet 0.5.1
```

AndreiEres:
Fix https://github.com/rubocop-hq/rubocop/issues/9415

* Added check for inline def modifiers. Now they can be handle as `public_methods`, `protected_methods`, and `private methods` respectively.
* Fixed autocorrection range for methods with modifier.
* Fixed test that now throws new offence for inline modifier but was created for another reason.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/layout/class_structure_spec.rb
+++ b/spec/rubocop/cop/layout/class_structure_spec.rb
@@ -172,9 +172,6 @@ RSpec.describe RuboCop::Cop::Layout::ClassStructure, :config do
 
         private :other_public_method
 
-        private def something_inline
-        end
-
         def yet_other_public_method
         end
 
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/layout/class_structure_spec.rb
```diff
@@ -321,4 +318,27 @@ def do_something
...
     RUBY
   end
+
+  it 'registers an offense and corrects when private method with inline modifier is declared before public' do
+    expect_offense(<<~RUBY)
+      class A
+        private def foo
+        end
+
+        public def bar
```
dduugg, line 328: Does this still flag and autocorrect the offense if the explicit `public` isn't there? Is that captured in a test?

AndreiEres: Thanks, I've added the test

bbatsov:
The changes look good. Probably we should add some doc examples of a class using inline modifiers, but that's orthogonal to the current PR. Thanks!

PR #9424, opened 2021-01-27
Fix an incorrect auto-correct for `Style/ClassMethodsDefinitions`

koic:
This PR fixes the following incorrect auto-correct for `Style/ClassMethodsDefinitions` when defining class methods with `class << self` and there is no blank line between method definition and attribute accessor.

```console
% cat example.rb
class A
  class << self
    def three
    end
    attr_reader :two
  end
end
```

```console
% bundle exec rubocop --only Style/ClassMethodsDefinitions -a
(snip)

Inspecting 1 file
C

Offenses:

example.rb:2:3: C: [Corrected] Style/ClassMethodsDefinitions: Do not
define public methods within class << self.
class << self ...
^^^^^^^^^^^^^

```

## Before

The syntax is broken as follows.

```console
% cat example.rb
class A
  class << self    attr_reader :two
  end

  def self.three
  end
end

% ruby -c example.rb
example.rb:2: syntax error, unexpected local variable or method,
expecting ';' or '\n'
class << self    attr_reader :two
example.rb:7: syntax error, unexpected `end', expecting end-of-input
```

## After

Safe auto-correction is done.

```console
% cat example.rb
class A
  class << self
  attr_reader :two
  end

  def self.three
  end
end
```

NOTE: This change may leave unwanted whitespace, which can be resolved with `Layout/EmptyLinesAroundClassBody`. On the other hand, cannot auto-correction for syntax error, so this PR fixes the syntax error first priority.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_an_incorrect_autocorrect_for_class_methods_definitions.md
+++ b/changelog/fix_an_incorrect_autocorrect_for_class_methods_definitions.md
@@ -1,0 +1,1 @@
+* [#9424](https://github.com/rubocop-hq/rubocop/pull/9424): Fix an incorrect auto-correct for `Style/ClassMethodsDefinitions` when defining class methods with `class << self` and there is no blank line between method definition and attribute accessor. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

Issue #8724, opened 2020-09-15
Style/ConstantVisibility needs ♥️


marcandre:
The `Style::ConstantVisibility` cop has potential but could be made much more useful.

False positives: it should detect `# @api private` (or `# @api public`) in the associated comment, and also detect if the parent module is itself private (either via `private_constant` or `@api private`).

Modules vs literals: it should allow a config `'ExcludeModules'` to restrict detection of literals only (`EXAMPLE = 42`, and not `SomeClass = Struct.new()`). Typically nested modules and classes are public, but literals should be private.

Auto-correction: it should offer auto-correction. A config option like `DefaultVisibility` could make it much nicer. Values could be: `'private_constant'`, `'public_constant'`, `'@api private'`, or `'@api public'`. Autocorrector should be mindful of constants containing heredocs and/or assignments ending with comments.

bbatsov:
Yeah, totally agreed. Lots of room for improvement there.

bbatsov:
Closing as stale — not actionable as filed. If there are specific improvements needed for `Style/ConstantVisibility`, please open a focused issue.

PR #9427, opened 2021-01-27
Add `IgnoreModules` configuration to `Style/ConstantVisibility`

tejasbubane:
To not register offense for module definitions.

Issue #8724

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/constant_visibility.rb
+++ b/lib/rubocop/cop/style/constant_visibility.rb
@@ -26,6 +26,24 @@ module RuboCop
       #     public_constant :BAZ
       #   end
       #
+      # @example IgnoreModules: false (default)
+      #   # bad
+      #   class Foo
+      #     MyClass = Struct.new()
+      #   end
+      #
+      #   # good
+      #   class Foo
+      #     MyClass = Struct.new()
+      #     public_constant :MyClass
+      #   end
+      #
+      # @example IgnoreModules: true
+      #   # good
+      #   class Foo
+      #     MyClass = Struct.new()
+      #   end
+      #
       class ConstantVisibility < Base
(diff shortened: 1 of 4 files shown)


dvandersluis:
I believe we've stopped using `Excluded` in configuration keys, and generally use `Ignored` instead, so `IgnoredModules` would probably be a preferable name here!

lib/rubocop/cop/style/constant_visibility.rb
```diff
@@ -26,20 +26,43 @@ module Style
...
           return if visibility_declaration?(node)
+          return if excluded_module?(node)
 
           message = message(node)
           add_offense(node, message: message)
         end
 
         private
 
+        def excluded_module?(node)
```
bbatsov, line 62: I think that API is misleading, as modules will get excluded when the config is true, but from the function name it seems that we're checking whether some some module is excluded.

lib/rubocop/cop/style/constant_visibility.rb
```diff
@@ -26,20 +26,43 @@ module Style
...
+      #   end
+      #
       class ConstantVisibility < Base
         MSG = 'Explicitly make `%<constant_name>s` public or private using ' \
               'either `#public_constant` or `#private_constant`.'
 
         def on_casgn(node)
           return unless class_or_module_scope?(node)
           return if visibility_declaration?(node)
+          return if excluded_module?(node)
```
bbatsov, line 54: I'd do this like `return if exclude_modules? && module?(node)` or something along those lines. I think it reflects a lot better what the code is supposed to be doing.

tejasbubane:
Changed config name to `IgnoreModules` and refactored the predicate method. @dvandersluis @bbatsov

bbatsov:
Thanks!

PR #9436, opened 2021-01-28
[Fix #9429] Fix `Style/NegatedIfElseCondition` autocorrect to keep comments in correct branch

### LINKED ISSUE #9429: NegatedIfElseCondition autocorrect leaves comments behind
Autocorrect this code:

```ruby
if !true
  # It wasn't true
  "Not true"
else
  "True!"
end
```

And you get:

```ruby
if true
  # It wasn't true
  "True!"
else
  "Not true"
end
```

The "not true" code path "#It wasn't true" comment didn't move and is now in the "true" code path falsely marking the "True!" response as "#It wasn't true". True fact.

Rubocop 1.5.2

tejasbubane:
Closes #9429

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/negated_if_else_condition_spec.rb
+++ b/spec/rubocop/cop/style/negated_if_else_condition_spec.rb
@@ -129,6 +129,64 @@ RSpec.describe RuboCop::Cop::Style::NegatedIfElseCondition, :config do
     RUBY
   end
 
+  it 'moves comments to correct branches during autocorrect' do
+    expect_offense(<<~RUBY)
+      if !condition.nil?
+      ^^^^^^^^^^^^^^^^^^ Invert the negated condition and swap the if-else branches.
+        # part B
+        # and foo is 39
+        foo = 39
+      else
+        # part A
+        # and foo is 42
+        foo = 42
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      if condition.nil?
+        # part A
+        # and foo is 42
+        foo = 42
+      else
+        # part B
+        # and foo is 39
+        foo = 39
+      end
+    RUBY
+  end
+
+  it 'works with comments and multiple statements' do
+    expect_offense(<<~RUBY)
+      if !condition.nil?
+      ^^^^^^^^^^^^^^^^^^ Invert the negated condition and swap the if-else branches.
+        # part A
+        # and foo is 1 and bar is 2
+        foo = 1
+        bar = 2
+      else
+        # part B
+        # and foo is 3 and bar is 4
+        foo = 3
+        bar = 4
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      if condition.nil?
+        # part B
+        # and foo is 3 and bar is 4
+        foo = 3
+        bar = 4
+      else
+        # part A
+        # and foo is 1 and bar is 2
+        foo = 1
+        bar = 2
+      end
+    RUBY
+  end
+
   it 'does not register an offense when negating condition for `if-elsif`' do
     expect_no_offenses(<<~RUBY)
       if !x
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/negated_if_else_condition_spec.rb
```diff
@@ -129,6 +129,33 @@
     RUBY
   end
 
+  it 'moves comments to correct branches during autocorrect' do
+    expect_offense(<<~RUBY)
+      if !condition.nil?
+      ^^^^^^^^^^^^^^^^^^ Invert the negated condition and swap the if-else branches.
+        # part B
+        # and foo is 39
+        foo = 39
```
koic, line 138: It seems that a syntax error occurs due to auto-correction when there are multiple statements as shown below.

```diff
        # part B        
        # and foo is 39
        foo = 39
+       foo = 40
```

Also, the following case does not seem to be corrected as expected.

```diff
        # part B        
        # and foo is 39
        foo = 39
+       # and foo is 40
+       foo = 40
```

bbatsov:
@tejasbubane ping :-)

tejasbubane:
@koic @bbatsov Fixed.

spec/rubocop/cop/style/negated_if_else_condition_spec.rb
```diff
@@ -129,6 +129,64 @@
...
+        # and foo is 42
+        foo = 42
+      else
+        # part B
+        # and foo is 39
+        foo = 39
+      end
+    RUBY
+  end
+
```
koic, line 158: Thank you for the fixing. There's one more thing. Can you add and pass the following test case ending in comments? (I think it's an edge case 😅 )

```ruby
  it 'registers and corrects an offense when using negated condition and the end of branches are comment' do
     expect_offense(<<~RUBY)
       if !condition.nil?
       ^^^^^^^^^^^^^^^^^^ Invert the negated condition and swap the if-else branches.
         # part B
         # and foo is 39
         foo = 39
         # end of `then` body comment
       else
         # part A
         # and foo is 42
         foo = 42
         # end of `else` body comment
       end
     RUBY

     expect_correction(<<~RUBY)
       if condition.nil?
         # part A
         # and foo is 42
         foo = 42
         # end of `else` body comment
       else
         # part B
         # and foo is 39
         foo = 39
         # end of `then` body comment
       end
     RUBY
   end
```

koic: @tejasbubane Can you fix this issue with this PR? I think it's an edge case, so I think it can be merged and improved even at present if you don't have time for this case.

tejasbubane: @koic I tried handling this case but seems like this needs to be fixed at parser level. Below is output of `processed_source.ast_with_comments`. Note the comment just above else if taken as part of else statement. Is there any other rubocop API that I can use and I am unaware of?

```
irb(#<RuboCop::Cop::Style::NegatedIfElseCondition:0x00005588841deb30>):007:0> pp processed_source.ast_with_comments
{s(:lvasgn, :foo,
  s(:int, 39))=>
  [#<Parser::Source::Comment (string):2:3 "# part B">,
   #<Parser::Source::Comment (string):3:3 "# and foo is 39">],
 s(:lvasgn, :foo,
  s(:int, 42))=>
  [#<Parser::Source::Comment (string):5:3 "# end of `then` body comment">,
   #<Parser::Source::Comment (string):7:3 "# part A">,
   #<Parser::Source::Comment (string):8:3 "# and foo is 42">],
 s(:if,
  s(:send,
    s(:send,
      s(:send, nil, :condition), :nil?), :!),
  s(:lvasgn, :foo,
    s(:int, 39)),
  s(:lvasgn, :foo,
    s(:int, 42)))=>
  [#<Parser::Source::Comment (string):10:3 "# end of `else` body comment">]}
```

koic: Thank you for investigating. This issue may require deeper investigation to be resolved. And this PR has improved auto-correction, so let's merge it with the current status. Thank you again!

PR #9443, opened 2021-01-28
[Fix #9437] Improve offense message when there is an allowed range of empty lines.

### LINKED ISSUE #9437: `Layout/EmptyLineBetweenDefs` with multiple empty lines gives confusing message

In case of multiple lines between method defs, the message shown by `Layout/EmptyLineBetweenDefs` is `Use empty lines between method definitions`. This is slightly confusing and can be improved.

1. For default config `NumberOfEmptyLines=1`: Mention there are multiple lines detected.
2. For `NumberOfEmptyLines=[min, max]`: Mention the number of lines detected does not match config.

NA

## Additional context

Steps to reproduce:

```
$ cat test.rb
def foo; end


def bar; end

$ rubocop test.rb
Inspecting 1 file
C

Offenses:

test.rb:3:1: C: [Correctable] Layout/EmptyLines: Extra blank line detected.
test.rb:4:1: C: [Correctable] Layout/EmptyLineBetweenDefs: Use empty lines between method definitions.
def bar; end
^^^^^^^

1 file inspected, 2 offenses detected, 2 offenses auto-correctable
```

```
$ rubocop -V
1.9.0 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 3.0.0 x86_64-linux)
```

dvandersluis:
When there is a range allowed for empty lines, the offense message takes that into account.

Fixes #9437.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/empty_line_between_defs.rb
+++ b/lib/rubocop/cop/layout/empty_line_between_defs.rb
@@ -88,7 +88,7 @@ module RuboCop
         include RangeHelp
         extend AutoCorrector
 
-        MSG = 'Use empty lines between %<type>s definitions.'
+        MSG = 'Expected %<expected>s between %<type>s definitions; found %<actual>d.'
 
         def self.autocorrect_incompatible_with
           [Layout::EmptyLines]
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/layout/empty_line_between_defs.rb
```diff
@@ -89,6 +89,8 @@ class EmptyLineBetweenDefs < Base
         extend AutoCorrector
 
         MSG = 'Use empty lines between %<type>s definitions.'
+        MSG_RANGE = '%<min>d..%<max>d empty lines between %<type>s definitions expected; ' \
+                    'found %<actual>d.'
```
tejasbubane, line 93: I feel `between %<min> and %<max> (both inclusive)` would read better in a message than the range syntax.

lib/rubocop/cop/layout/empty_line_between_defs.rb
```diff
@@ -157,14 +159,18 @@ def module_candidate?(node)
...
-                 when :def, :defs
-                   :method
-                 else
-                   node.type
-                 end
-          format(MSG, type: type)
+        def message(node, count: nil)
+          type = node_type(node)
+
+          if allowance_range?
```
tejasbubane, line 165: I would be great if for default config `NumberOfEmptyLines=1` and multiple lines, the message mentions that `%<n>d lines were detected, 1 expected`.

lib/rubocop/cop/layout/empty_line_between_defs.rb
dvandersluis: I originally had written it less terse but I think rubyists will understand range syntax without having to be overly verbose.

lib/rubocop/cop/layout/empty_line_between_defs.rb
dvandersluis: Thanks for the suggestion @tejasbubane! Now the message shows the actual and expected count even when it's not a range.

PR #9461, opened 2021-01-29
[Fix #9459] Add `AllowedMethods` option to `Style/IfWithBooleanLiteralBranches`

### LINKED ISSUE #9459: False positive with Style/IfWithBooleanLiteralBranches and nonzero?
The `nonzero?` method returns self if num is not zero, nil otherwise, and not a boolean. Causing false positive with the `Style/IfWithBooleanLiteralBranches` cop with code such as `val.nonzero? ? true : false`

--------

## Expected behavior

The cop should not flag statements such as `val.nonzero? ? true : false`

## Actual behavior

```
 Style/IfWithBooleanLiteralBranches: Remove redundant ternary operator with boolean literal branches.
      NS::Browser.get(url(path)).code.nonzero? ? true : false
```

Run the cop against code like `val.nonzero? ? true : false` , for example https://github.com/wpscanteam/CMSScanner/blob/master/lib/cms_scanner/web_site.rb#L74

## RuboCop version

```
1.9.0 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 2.7.2 x86_64-linux)
  - rubocop-performance 1.9.2
```

koic:
Fixes #9459.

This PR adds `AllowedMethods` option to `Style/IfWithBooleanLiteralBranches` and set `nonzero?` as default value.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_add_allowed_method_option_to_if_with_boolean_literal_branche.md
+++ b/changelog/new_add_allowed_method_option_to_if_with_boolean_literal_branche.md
@@ -1,0 +1,1 @@
+* [#9459](https://github.com/rubocop-hq/rubocop/issues/9459): Add `AllowedMethods` option to `Style/IfWithBooleanLiteralBranches` and set `nonzero?` as default value. ([@koic][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -3521,6 +3521,8 @@ Style/IfWithBooleanLiteralBranches:
   Description: 'Checks for redundant `if` with boolean literal branches.'
   Enabled: pending
   VersionAdded: '1.9'
+  AllowedMethods:
+    - nonzero?
 
 Style/IfWithSemicolon:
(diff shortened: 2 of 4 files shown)


dvandersluis:
Should this cop maybe be `Safe: false`? Any predicate method (or really any method at all) has a chance of the same unsafe correction.

marcandre:
> Should this cop maybe be Safe: false? 

I think so.

marcandre:
I would at least turn on the auto-correct as being unsafe. Problem is that right correction in case part of the condition isn't strictly `true`/`false` is, to my taste, to use `!!`, but by default that's supposed to be bad...

koic:
Thank you for the conversations! I opened #9476 that marks `Style/IfWithBooleanLiteralBranche` as unsafe auto-correction.

bbatsov:
Thanks!

PR #9471, opened 2021-01-29
[Fix #9465] Update `Metrics/ParameterLists` to be able to write `MaxOptionalParameters` in rubocop_todo.yml

### LINKED ISSUE #9465: Metrics/ParameterLists ignores Max value generate by rubocop --auto-gen-config

1. Create `test.rb` file with content:
```
# frozen_string_literal: true

def foo(param1, param2, param3 = true, param4 = false, param5 = false, param6 = false)
  puts param1
  puts param2
  puts param3
  puts param4
  puts param5
  puts param6
end

```
2. run `rubocop --auto-gen-config` and get this `.rubocop_todo.yml`
```
# This configuration was generated by
# `rubocop --auto-gen-config`
# on 2021-01-29 13:25:51 UTC using RuboCop version 1.9.0.
# The point is for the user to remove these configuration records
# one by one as the offenses are removed from the code base.
# Note that changes in the inspected code, or installation of new
# versions of RuboCop, may require this file to be generated again.

# Offense count: 2
# Configuration parameters: CountKeywordArgs, MaxOptionalParameters.
Metrics/ParameterLists:
  Max: 6

# Offense count: 4
# Configuration parameters: AllowedMethods.
# AllowedMethods: respond_to_missing?
Style/OptionalBooleanParameter:
  Exclude:
    - 'test.rb'
```

3. `Metrics/ParameterLists` max is 6, so it should be no warning for running rubocop now, but there is one:
```
rubocop
...
Offenses:

test.rb:3:1: C: Metrics/ParameterLists: Method has too many optional parameters. [4/3]
def foo(param1, param2, param3 = true, param4 = false, param5 = false, param6 = false) ...
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```
## RuboCop version

```
1.9.0 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 2.7.2 x86_64-linux)
```

dvandersluis:
`Metrics/ParameterLists` adds offenses for `MaxOptionalParameters`, but when used with `--auto-gen-config` it does not configure that in the todo file, so there will still be registered offenses. This change adds that configuration option as well.

~Note that I had to slightly change the `ConfigurableMax` mixin so that I could use it twice for this cop (since there are two "max" configurations), without duplicating the code.~

This PR deprecates `ConfigurableMax` in favour of `ExcludeLimit` which adds a class macro for defining a configuration option with an exclude limit and can be called multiple times in the same class.

Fixes #9465.

-----------------

* [x] Added tests.


bbatsov:
I'm wondering if it won't be easier and safer to actually introduce a new more flexible mixin and we can just deprecate the old one. I was never found of the `ConfigurableMax` name anyways. :D

dvandersluis:
@bbatsov what do you have in mind? The key part for `ConfigurableMax` is that it just keeps track of the maximum value (which I'd otherwise have to duplicate for the second configuration).

bbatsov:
My main concern is that your change looks like a breaking change to the public API. No idea if extensions actually use it, but I guess some probably do.

dvandersluis:
Oh yeah that’s a really good point. I’ll write a new one and deprecate the old one.

dvandersluis:
Updated now to deprecate, but not change, `ConfigurableMax`, and add `ExcludeLimit` instead.

bbatsov:
Great work!

PR #9478, opened 2021-01-30
Style/LegacyHash cop, prohibiting Hash[] method

zverok:
```ruby
# bad
Hash[ary]
^^^^^^^^^ Prefer ary.to_h to Hash[ary].

# good
ary.to_h

# bad
Hash[a, b, c, d]
^^^^^^^^^^^^^^^^ Prefer literal hash to Hash[arg1, arg2, ...].

# good
{a => b, c => d}

# bad
Hash[]
^^^^^^ Prefer literal hash to Hash[arg1, arg2, ...].

# good
{}
```

Correctly identifies un-autocorrecable cases (`Hash[*foo]`, and an odd number of arguments).
I am open to suggestions regarding the name, it is my best guess :)

-----------------

* [x] Added tests.


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#9478](https://github.com/rubocop-hq/rubocop/pull/9478): Add new `Style/LegacyHash` cop. ([@zverok][])
+
```
tejasbubane, line 8: Please refer [this][1] for creating changelog entries. It will create separate files under `/changelog` which are then merged into this main changelog during release.

[1]: https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#creating-changelog-entries

lib/rubocop/cop/style/legacy_hash.rb
```diff
@@ -0,0 +1,74 @@
...
+      #   Hash[key1, value1, key2, value2]
+      #
+      #   # good
+      #   {key1 => value1, key2 => value2}
+      class LegacyHash < Base
+        extend AutoCorrector
+
+        MSG_TO_H = 'Prefer ary.to_h to Hash[ary].'
+        MSG_LITERAL = 'Prefer literal hash to Hash[arg1, arg2, ...].'
+        MSG_SPLAT = 'Prefer array_of_pairs.to_h to Hash[*array].'
```
tejasbubane, line 25: These messages should be have dynamic variable names. Please refer to other cops like [this one][1]

[1]: https://github.com/rubocop-hq/rubocop/blob/77db90e6fa0f45ea7cb0b6ec4bee23faf66a385a/lib/rubocop/cop/style/eval_with_location.rb#L42-L45

lib/rubocop/cop/style/legacy_hash.rb
```diff
@@ -0,0 +1,74 @@
...
+          #   Hash[a1, a2, a3, a4] => {a1 => a2, a3 => a4}
+          #   ...but don't suggest correction if there is odd numbe of them (it is a bug)
+          node.arguments.count == 1 ? single_argument(node) : multi_argument(node)
+        end
+
+        private
+
+        def single_argument(node)
+          if node.arguments.first.splat_type?
+            add_offense(node, message: MSG_SPLAT)
```
tejasbubane, line 47: We should auto-correct this too.

spec/rubocop/cop/style/legacy_hash_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      Hash[]
+      ^^^^^^ Prefer literal hash to Hash[arg1, arg2, ...].
+    RUBY
+
+    expect_correction(<<~RUBY)
+      {}
+    RUBY
+  end
+
+  it 'does not tries to correct multi-argument Hash with odd number of arguments' do
```
tejasbubane, line 37: For wordings of these example descriptions, please refer: https://github.com/rubocop-hq/rubocop/blob/77db90e6fa0f45ea7cb0b6ec4bee23faf66a385a/lib/rubocop/cop/internal_affairs/example_description.rb

lib/rubocop/cop/style/legacy_hash.rb
zverok: I contemplated this, but I prefer not to. The real code that caused offense might look like this (Rubocop's own codebase):
```ruby
Hash[PERCENT_LITERAL_TYPES.map do |type|
  [type, preferred_delimiters_config[type] ||
     preferred_delimiters_config['default']]
end]
```
...and that's not the limit, block inside `Hash[]` can take many lines, so the suggestion would look really cumbersome. That's why I've decided to stick with abstract recommendations.

lib/rubocop/cop/style/legacy_hash.rb
zverok: I don't think so. With splat, there are basically two possibilities:
1. Code inside `Hash[]` produces a flat list of keys and values by design (say, it reads a file which has `[key, value, key, value]` split line-by-line)
2. Code actually produces an array of pairs, and then `flatten(1)` added to the end of it to match `Hash[]` signature.

In both cases, it should be reviewed and redecided by human. 
The only working suggestion we can produce is
```diff
- Hash[*something]
+ something.each_slice(2).to_h
```
...and in most cases, it would not be a good suggestion.

lib/rubocop/cop/style/legacy_hash.rb
dvandersluis: @zverok that's a fair point. I think maybe one case we can be more explicit about is `Hash[]` to `{}`.

spec/rubocop/cop/style/legacy_hash_spec.rb
zverok: I don't understand this suggestion, sorry. What exactly should be fixed here (except for "tries"→"try" :joy:)?

config/default.yml
```diff
@@ -3623,6 +3623,11 @@ Style/LambdaCall:
     - call
     - braces
 
+Style/LegacyHash:
+  Description: 'Avoid Hash[] in favor of ary.to_h or literal hashes.'
+  Enabled: pending
+  VersionAdded: '1.10'
```
dvandersluis, line 3629:
```suggestion
  VersionAdded: <<next>>
```

We don't hardcode next versions anymore, `<<next>>` is metadata that is automatically updated when a release is made.

spec/rubocop/cop/style/legacy_hash_spec.rb
dvandersluis: Yeah after correcting "tries" I think it's fine. The internal affairs cop is just to make sure that we don't have descriptions that contradict the actual test.

lib/rubocop/cop/style/legacy_hash.rb
zverok: @dvandersluis I actually believe it is the tiniest percent of actual `Hash::[]` usages in code, and didn't want to complicate the code for the rare case.

lib/rubocop/cop/style/legacy_hash.rb
```diff
@@ -0,0 +1,74 @@
...
+      #   {key1 => value1, key2 => value2}
+      class LegacyHash < Base
+        extend AutoCorrector
+
+        MSG_TO_H = 'Prefer ary.to_h to Hash[ary].'
+        MSG_LITERAL = 'Prefer literal hash to Hash[arg1, arg2, ...].'
+        MSG_SPLAT = 'Prefer array_of_pairs.to_h to Hash[*array].'
+        RESTRICT_ON_SEND = %i[[]].freeze
+
+        def_node_matcher :hash_from_array?, '(send (const nil? :Hash) :[] ...)'
```
dvandersluis, line 28: Might be useful to catch `::Hash` as well:

```suggestion
        def_node_matcher :hash_from_array?, '(send (const {nil? cbase} :Hash) :[] ...)'
```

lib/rubocop/cop/style/legacy_hash.rb
```diff
@@ -0,0 +1,74 @@
...
+        def on_send(node)
+          return unless hash_from_array?(node)
+
+          # There are several cases:
+          # If there is one argument:
+          #   Hash[ary] => ary.to_h
+          #   Hash[*ary] => don't suggest corrections
+          # If there is 0 or 2+ arguments:
+          #   Hash[a1, a2, a3, a4] => {a1 => a2, a3 => a4}
+          #   ...but don't suggest correction if there is odd numbe of them (it is a bug)
```
dvandersluis, line 39:
```suggestion
          #   ...but don't suggest correction if there is odd number of them (it is a bug)
```

lib/rubocop/cop/style/legacy_hash.rb
dvandersluis: Fair enough, I was basing on your example cases.

bbatsov:
It seems to me that a name like `ArrayToHashConversation` or something along those lines would be better, as it's both more neutral (allows us to make this configurable) and more descriptive of what the cop actually checks for.

zverok:
@bbatsov This was my first guess, but...
```ruby
Hash[:a, 1, :b, 2]
```
...has no array to be seen :) (And we really have a few of those in the codebase, so they can happen in the wild!)

I thought about configurability and (maybe too bold for me) decided there should be no reason to prefer `Hash[foo]` to `foo.to_h`, the `.to_h` is clearly superior.

bbatsov:
Fair points. Maybe simply `HashConversion` then? I still think that `LegacyHash` is not the most informative name.

zverok:
@bbatsov Yeah, that's ideal.
Updated cop name, edited commit history to make more sense, pweeze check!

bbatsov (approved):
Looks great. I'll merge it after we cut 1.9.1.

lib/rubocop/cop/mixin/preferred_delimiters.rb
```diff
@@ -35,10 +35,10 @@ def preferred_delimiters
             ensure_valid_preferred_delimiters
 
             if preferred_delimiters_config.key?('default')
-              Hash[PERCENT_LITERAL_TYPES.map do |type|
+              PERCENT_LITERAL_TYPES.map do |type|
                 [type, preferred_delimiters_config[type] ||
                   preferred_delimiters_config['default']]
-              end]
+              end.to_h
```
jdufresne, line 41: I believe this could be further simplified to avoid the `map`

```ruby
              PERCENT_LITERAL_TYPES.to_h do |type|
                [type, preferred_delimiters_config[type] ||
                  preferred_delimiters_config['default']]
              end
```

dvandersluis: `to_h` with a block was only added in ruby 2.6, but RuboCop supports 2.4, so we can’t use that.

jdufresne: Ah I see. Thanks for explaining that.

bbatsov:
@zverok Can you squash the fixup commit?

zverok:
@bbatsov Done

changelog/new_style_hash_conversion_cop.md
```diff
@@ -0,0 +1 @@
+* [#9478](https://github.com/rubocop-hq/rubocop/pull/9478): Add new `Style/LegacyHash` cop. ([@zverok][])
```
bbatsov, line 1: Just noticed that we need to update the name here as well.

zverok: @bbatsov fixed

Issue #9452, opened 2021-01-29
StyleGuideBaseURL can't be set on nested departments


tas50:
## Expected behavior

Full disclosure: I might be missing something here. Trust but verify.

So https://github.com/rubocop-hq/rubocop/pull/7295 enabled setting the `StyleGuideBaseURL` config at the department level, which is great because I have 200 cops in a few different departments and I want to push users to our docs when they encounter those cops. It's not really something I want to include in the config 200 times. Per department is 😘.

The problem is I have nested departments like `Chef/Style/UsePlatformHelpers`

In the config if I set 

```yaml
Chef/Style:
    StyleGuideBaseURL: https://docs.chef.io/whatever
```

and then later define the cop with a `StyleGuide` config like this:

```yaml
Chef/Style/UsePlatformHelpers:
  Description: Use platform? and platform_family? helpers to check node platform in resources and recipes
  StyleGuide: '#chefstyleuseplatformhelpers'
  Enabled: true
  VersionAdded: '5.6.0'
  Exclude:
    - '**/metadata.rb'
    - '**/libraries/*'
    - '**/Berksfile'
```

You would think I'd end up with:

```
R: [Correctable] Chef/Style/UsePlatformHelpers: Use platform? and platform_family? helpers to check a node's platform (https://docs.chef.io/whatever#chefstyleuseplatformhelpers)
      version << rhel_append_version if node['platform_family'] == 'rhel' &&
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## Actual behavior

You get the BaseURL defined in AllCops in the rubocop project config:

```
R: [Correctable] Chef/Style/UsePlatformHelpers: Use platform? and platform_family? helpers to check a node's platform (https://rubystyle.guide#chefstyleuseplatformhelpers)
      version << rhel_append_version if node['platform_family'] == 'rhel' &&
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

If I change it at the AllCops level in my config it works fine, but not at the department level.

## RuboCop version

This is in within Cookstyle in a RuboCop 1.9 branch here: https://github.com/chef/cookstyle/tree/1.9

```
$ [bundle exec] rubocop -V
1.9.0 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 2.7.2 x86_64-darwin20)
  - rubocop-performance 1.9.2
```

[...]

PR #9480, opened 2021-01-30
[#9452] `StyleGuideBaseURL` not functioning with nested departments

tas50:
Currently, you can set the StyleGuideBaseURL config for cops that ship
within RuboCop or other RuboCop plugins that use a DEPT/COP format. If
the project has nested departments this doesn't work as the message
annotator does not properly split the cop name to determine the
department to lookup in the config.

With this change, just the cop name is split off from the full DEPT/COP
string allowing the StyleGuideBaseURL to function with nested
departments.

We have nested departments in Cookstyle like `Chef/Deprecations/Foo` and
we'd like to be able to enable this config by default so we can point
our users at our documentation and define the URL at the department level.

Signed-off-by: Tim Smith <tsmith@chef.io>


bbatsov:
Looks good! Thanks!

PR #9496, opened 2021-02-08
Add new `Gemspec/DateAssignment` cop

koic:
Follow https://github.com/rubocop-hq/rubocop-rails/issues/432.

This PR adds new `Gemspec/DateAssignment` cop.

This cop checks that `date =` is not used in gemspec file. It is set automatically when the gem is packaged.

```ruby
# bad
  s.name = 'your_cool_gem'
  spec.date = Time.now.strftime('%Y-%m-%d')
end

# good
  s.name = 'your_cool_gem'
end
```

RubyGems doesn't expect the value to be set.
https://github.com/rubygems/rubygems/blob/be08d8307eda3b61f0ec0460fe7fbcf647b526e6/lib/rubygems/specification.rb#L1679-L1681

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -203,6 +203,13 @@ Bundler/OrderedGems:
 
 #################### Gemspec ###############################
 
+Gemspec/DateAssignment:
+  Description: 'Checks that `date =` is not used in gemspec file, it is set automatically when the gem is packaged.'
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  Include:
+    - '**/*.gemspec'
+
 Gemspec/DuplicatedAssignment:
   Description: 'An attribute assignment method calls should be listed only once in a gemspec.'
   Enabled: true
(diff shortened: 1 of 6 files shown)


config/default.yml
```diff
@@ -203,6 +203,13 @@ Bundler/OrderedGems:
 
 #################### Gemspec ###############################
 
+Gemspec/DateAssignment:
+  Description: 'Checks that `date =` is not used in gemspec file, it is set automatically when the gem is packaged.'
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  Include:
+    - '**/*.gemspec'
```
dvandersluis, line 211: This will include gemspecs in vendored gems, which may have the date set but be out of the control of the developer.

koic: Good point. However this setting is the same as other Gemspec department's cops. These cops hasn't received feedback on that issue, so I think we see reactions with this setting.

bbatsov:
Love it!

PR #9504, opened 2021-02-10
[Fix #9500] Update `Lint/Debugger` so that only specific receivers for debug methods lead to offenses

### LINKED ISSUE #9500: Lint/Debugger complains about using "console" when it is not appropriate
If I define and use a `console` method, `attr_reader :console` or a minitest/spec `let(:console)` Lint/Debugger says I should "Remove debugger entry point console".

I have not `require`d a debugger, nor have I added one to a `Gemfile`, so I cannot be entering a debugger, I'm working with a method or accessor that is appropriately named for my usage and is in-scope.  No `method_missing` magic is involved.

## Expected behavior

Lint/Debugger does not complain by default about methods I define that happen collide with a debugging tool I have not even loaded a gem for.

I expect zero lint warnings for the code included below.

## Actual behavior

```
$ cat t.rb
require "stringio"
require "minitest/spec"

describe "bug" do
  let(:console) { StringIO.new }
  let(:thing) { [console] }

  it "does not complain about console" do
    assert thing
  end
end

class T1
  def console
  end

  def test
    console.puts "this is not a debugger"
  end
end

class T2
  attr_reader :console

  def test
    console.puts "this is not a debugger"
  end
end

module T3
  def console
    $stdeout
  end
end

class T4
  include T3

  def test
    console.puts "this is not a debugger"
  end
end

$ rubocop -l t.rb
Inspecting 1 file
W

Offenses:

t.rb:6:18: W: Lint/Debugger: Remove debugger entry point console.
  let(:thing) { [console] }
                 ^^^^^^^
t.rb:18:5: W: Lint/Debugger: Remove debugger entry point console.
    console.puts "this is not a debugger"
    ^^^^^^^
t.rb:26:5: W: Lint/Debugger: Remove debugger entry point console.
    console.puts "this is not a debugger"
    ^^^^^^^
t.rb:40:5: W: Lint/Debugger: Remove debugger entry point console.
    console.puts "this is not a debugger"
    ^^^^^^^

1 file inspected, 4 offenses detected
$ ls -a
.	..	t.rb
```

Run `rubocop -l` on test embedded above

## RuboCop version

```
$ rubocop -V
1.9.1 (using Parser 3.0.0.0, rubocop-ast 1.3.0, running on ruby 2.6.6 x86_64-darwin19)
```

dvandersluis:
Prior to #8929, `Lint/Debugger` looked for a specific receiver/method pair, but that PR changed it so that any of the `DebuggerReceivers` would apply, *or also* no receiver. This caused regressions (#7636 and #9500, for instance) because method names specified as debugger by this cop would now register offenses when called with no receiver.

In order to fix it, I changed the configuration for `Lint/Debugger` a bit, but in a way that should not cause any major incompatibility. The allow list is now specified including its receiver so that the receiver can be checked. Existing configurations that just pass in a method name will be treated as expecting no receiver (there's a slight incompatibility there because previously `foo` would be matched by `Kernel.foo` but I think this is ok (and is resolvable by adding `Kernel.foo` to the configuration).

I've also grouped the debugger methods by source (it's often confusing to me where each method comes from), which also allows an end-user to disable a pre-defined group of methods by specifying `Byebug: ~` for example.

Finally I've deprecated the `DebuggerReceivers` configuration as it is no longer useful.

Fixes #9500.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1446,23 +1446,31 @@ Lint/Debugger:
   Description: 'Check for debugger calls.'
   Enabled: true
   VersionAdded: '0.14'
-  VersionChanged: '0.49'
-  DebuggerReceivers:
-    - binding
-    - Kernel
-    - Pry
+  VersionChanged: <<next>>
+  DebuggerReceivers: [] # deprecated
   DebuggerMethods:
-    - debugger
-    - byebug
-    - remote_byebug
-    - pry
-    - remote_pry
-    - pry_remote
-    - console
-    - rescue
-    - save_and_open_page
-    - save_and_open_screenshot
-    - irb
+    # Groups are available so that a specific group can be disabled in
+    # a user's configuration, but are otherwise not significant.
+    Kernel:
+      - binding.irb
+    Byebug:
+      - byebug
+      - remote_byebug
+      - Kernel.byebug
+      - Kernel.remote_byebug
+    Capybara:
+      - save_and_open_page
+      - save_and_open_screenshot
+    Pry:
+      - binding.pry
+      - binding.remote_pry
+      - binding.pry_remote
+      - Pry.rescue
+    Rails:
+      - debugger
+      - Kernel.debugger
+    WebConsole:
+      - binding.console
 
 Lint/DeprecatedClassMethods:
   Description: 'Check for deprecated class method calls.'
--- a/spec/rubocop/cop/lint/debugger_spec.rb
+++ b/spec/rubocop/cop/lint/debugger_spec.rb
@@ -1,69 +1,104 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::Lint::Debugger, :config do
-  context 'with configured debugger methods' do
+  context 'with the DebuggerMethods configuration' do
     let(:cop_config) do
       { 'DebuggerMethods' => %w[custom_debugger] }
     end
 
-    it 'does not report an offense for a byebug call' do
+    it 'does not register an offense for a byebug call' do
       expect_no_offenses(<<~RUBY)
         byebug
       RUBY
     end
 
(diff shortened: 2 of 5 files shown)


spec/rubocop/cop/lint/debugger_spec.rb
```diff
@@ -87,120 +122,176 @@
...
+    it 'does not register an offense for `console` without a receiver' do
+      expect_no_offenses('console')
+    end
   end
 
-  it 'reports an offense for save_and_open_page with Kernel' do
-    expect_offense(<<~RUBY)
-      Kernel.save_and_open_page
-      ^^^^^^^^^^^^^^^^^^^^^^^^^ Remove debugger entry point `Kernel.save_and_open_page`.
-    RUBY
```
dvandersluis, line 168: This test was [discussed in #8929 ](https://github.com/rubocop-hq/rubocop/pull/8929#pullrequestreview-524890240) as being there because it was not possible to exclude it in the way the cop was changed, but capybara does not actually add the method to `Kernel` so it's a false positive that I've now removed.

config/default.yml
```diff
@@ -1446,23 +1446,31 @@ Lint/Debugger:
   Description: 'Check for debugger calls.'
   Enabled: true
   VersionAdded: '0.14'
-  VersionChanged: '0.49'
-  DebuggerReceivers:
-    - binding
-    - Kernel
-    - Pry
+  VersionChanged: <<next>>
+  DebuggerReceivers: [] # deprecated
```
dvandersluis, line 1450: I didn't want users to get error messages about an invalid configuration (in case they've set this) so I've added it to the obsoletion config (with warning severity) but kept the configuration.

bbatsov:
Fantastic work! Quite the improvement over the previous configuration setup of the cop.

Issue #9329, opened 2021-01-03
`TargetRubyVersion` and `inherit` loading error


marcandre:
If my `.rubocop.yml` starts with:

```
inherit_from: .rubocop_todo.yml

AllCops:
  DisplayCopNames: true
  TargetRubyVersion: 2.4
  NewCops: enable
# ...
```

And my gemspec has:
```ruby
  spec.required_ruby_version = Gem::Requirement.new('>= 2.3.0')
```

Then I get an error:
```
$ rubocop
Error: RuboCop found unsupported Ruby version 2.3 in `required_ruby_version` parameter (in ractor-cache.gemspec). 2.3-compatible analysis was dropped after version 0.81.
Supported versions: 2.4, 2.5, 2.6, 2.7, 3.0
```

If I remove the line `inherit_from`, I don't get an error.

There should be no error (assuming I specify `TargetRubyVersion: 2.4` as I did)

AndreiEres:
@marcandre Maybe we can close this issue?

marcandre:
Indeed! Thanks

PR #9515, opened 2021-02-14
Required ruby version fixes

### LINKED ISSUE #9482: `TargetRuby` and `required_ruby_version` crash
Followup to #9329

There seem to be multitude issues with `TargetRuby` and `required_ruby_version` of gemspec:
* while `<= 3.0.0` will use `DEFAULT_VERSION` (good), `>= 2.0.0` will use `2.0.0` which will then raise an error and stop (really bad)
* same for list of versions
* does not understand `>` and `<`, `!=` and `~>`.

Code should probably use https://docs.ruby-lang.org/en/2.6.0/Gem/Requirement.html and find the minimal correct version that is `satisfied_by?`

HeroProtagonist:
Closes #9482 and #9329

The class to match `required_ruby_version` from `.gemspec` files could return invalid ruby versions. 

- This PR makes it so that it will only return a value from the `KNOWN_RUBIES` array. Specifically the minimal known ruby will be returned.
- The version parsing is also greatly improved by leveraging existing classes from the `Gem` library.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/target_ruby.rb
+++ b/lib/rubocop/target_ruby.rb
@@ -168,11 +168,7 @@ module RuboCop
         right_hand_side = version_from_gemspec_file(file)
         return if right_hand_side.nil?
 
-        version = version_from_right_hand_side(right_hand_side)
-        requirement = Gem::Requirement.new(version)
-        match = KNOWN_RUBIES.each.lazy.map { |r| Matcher.new(r, requirement) }.detect(&:satisfied?)
-
-        match.version.to_f if match
+        find_minimal_known_ruby(right_hand_side)
       end
 
       def gemspec_filename
@@ -196,37 +192,21 @@ module RuboCop
         if right_hand_side.array_type?
           version_from_array(right_hand_side)
         elsif gem_requirement?(right_hand_side)
-          remove_patch_from_version(right_hand_side.children.last.value)
+          right_hand_side.children.last.value
         else
-          remove_patch_from_version(right_hand_side.value)
+          right_hand_side.value
         end
       end
 
       def version_from_array(array)
-        array.children.map { |v| remove_patch_from_version(v.value) }
+        array.children.map(&:value)
       end
 
-      def remove_patch_from_version(str)
-        str.to_s.match(/^(?<prefix>>=?|<=?|~>|!=)?\s*(?<version>\d+(?:\.\d+)*)/) do |md|
-          "#{md[:prefix]}#{md[:version].to_f}"
-        end
-      end
+      def find_minimal_known_ruby(right_hand_side)
+        version = version_from_right_hand_side(right_hand_side)
+        requirement = Gem::Requirement.new(version)
 
-      # Helper to check if the requirement is satisfied by a given known ruby version
-      # @api private
-      class Matcher
-        def initialize(ruby_version, requirement)
-          @gem_version = Gem::Version.new(ruby_version)
-          @requirement = requirement
-        end
-
-        def satisfied?
-          @requirement.satisfied_by?(@gem_version)
-        end
-
-        def version
-          @gem_version.version
-        end
+        KNOWN_RUBIES.detect { |v| requirement.satisfied_by?(Gem::Version.new("#{v}.99")) }
       end
     end
 
--- a/lib/rubocop/target_ruby.rb
+++ b/lib/rubocop/target_ruby.rb
@@ -165,14 +165,14 @@ module RuboCop
         file = gemspec_filepath
         return unless file && File.file?(file)
 
-        version = version_from_gemspec_file(file)
-        return if version.nil?
+        right_hand_side = version_from_gemspec_file(file)
+        return if right_hand_side.nil?
 
-        requirement = version.children.last
-        return version_from_array(version) if version.array_type?
-        return version_from_array(requirement) if gem_requirement? version
+        version = version_from_right_hand_side(right_hand_side)
+        requirement = Gem::Requirement.new(version)
+        match = KNOWN_RUBIES.each.lazy.map { |r| Matcher.new(r, requirement) }.detect(&:satisfied?)
 
-        version_from_str(version.str_content)
+        match.version.to_f if match
       end
 
       def gemspec_filename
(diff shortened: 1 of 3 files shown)


bbatsov (approved):

lib/rubocop/target_ruby.rb
```diff
@@ -192,14 +192,40 @@ def version_from_gemspec_file(file)
...
-          md[:version].to_f
+      def remove_patch_from_version(str)
+        str.to_s.match(/^(?<prefix>>=?|<=?|~>|!=)?\s*(?<version>\d+(?:\.\d+)*)/) do |md|
+          "#{md[:prefix]}#{md[:version].to_f}"
+        end
+      end
+
+      # Helper to check if the requirement is satisfied by a given known ruby version
+      # @api private
+      class Matcher
```
marcandre, line 217: I think it would be best to do without this class and use the `Gem::Requirement` as a matcher. Note that matchers typically better constructed without the value to match; that value is passed to the matching method (like `satisfied_by?`).

lib/rubocop/target_ruby.rb
```diff
@@ -192,14 +192,40 @@ def version_from_gemspec_file(file)
...
       def version_from_array(array)
-        versions = array.children.map { |v| version_from_str(v.is_a?(String) ? v : v.str_content) }
-        versions.compact.min
+        array.children.map { |v| remove_patch_from_version(v.value) }
       end
 
-      def version_from_str(str)
-        str.match(/^(?:>=|<=)?\s*(?<version>\d+(?:\.\d+)*)/) do |md|
-          md[:version].to_f
+      def remove_patch_from_version(str)
```
marcandre, line 209: I see that you went through the trouble of parsing the requirement nevertheless... How about instead mapping our versions like 2.7 => '2.7.999'? This would have the desired effect, except in the case of exact Ruby version requirements (which are a mistake), simplify the code and avoid redoing the parsing that we shouldn't be doing.

spec/rubocop/target_ruby_spec.rb
```diff
@@ -308,33 +308,89 @@
...
                   s.licenses = ['MIT']
                 end
               HEREDOC
 
             create_file(gemspec_file_path, content)
             expect(target_ruby.version).to eq default_version
           end
+
+          it 'sets target_ruby from inclusive range' do
+            content =
```
marcandre, line 340: No obligation, but consider using `let` or `before_each` to avoid duplicating all these lines so many times.

spec/rubocop/target_ruby_spec.rb
```diff
@@ -308,33 +308,89 @@
...
           end
 
-          it 'sets target_ruby from required_ruby_version from inclusive range' do
+          it 'sets target_ruby from exacts string version' do
             content =
               <<-HEREDOC
                 Gem::Specification.new do |s|
                   s.name = 'test'
-                  s.required_ruby_version = '>= 3.0.0'
+                  s.required_ruby_version = '2.7.4'
```
marcandre, line 316: I wouldn't even test this. There's is no reason I can think of to specify such a requirement. We could even consider raising an offense in one of our cops for this, but I doubt it happens in practice.

spec/rubocop/target_ruby_spec.rb
```diff
@@ -308,33 +308,89 @@
...
           end
 
-          it 'sets default target_ruby from exclusive range' do
+          it 'sets target_ruby to default from unknown ruby' do
             content =
               <<-HEREDOC
                 Gem::Specification.new do |s|
                   s.name = 'test'
-                  s.required_ruby_version = '< 3.0.0'
+                  s.required_ruby_version = '2.3.0'
```
marcandre, line 330: Same here, testing for `< 2.3.0` is fine, but not `= 2.3.0`

marcandre (changes requested):
Much better!
I think it can be improved, simplified, and avoid parsing the requirement ourselves

HeroProtagonist:
@marcandre Thanks for the review. Updated the PR and I _think_ its ready now 🤞

lib/rubocop/target_ruby.rb
```diff
@@ -192,15 +188,27 @@ def version_from_gemspec_file(file)
...
-      def version_from_str(str)
-        str.match(/^(?:>=|<=)?\s*(?<version>\d+(?:\.\d+)*)/) do |md|
-          md[:version].to_f
-        end
+      def find_minimal_known_ruby(right_hand_side)
+        version = version_from_right_hand_side(right_hand_side)
+        requirement = Gem::Requirement.new(version)
+        gem_version = ->(ruby_version) { Gem::Version.new(ruby_version) }
+
+        match = KNOWN_RUBIES.each.lazy.map(&gem_version).detect { |v| requirement.satisfied_by?(v) }
```
marcandre, line 210: Can't these lines be replaced by:
```
match = KNOWN_RUBIES.detect { |v| requirement.satisfied_by?(Gem::Version.new(v) }
```

I'd actually use `Gem::Version.new("#{v}.99")` so that it works ok for `>= '2.6.1'`

In general, storing a lambda in a local variable is an anti-pattern. Also, it's typically best to use `lazy` only when actually needed.

spec/rubocop/target_ruby_spec.rb
```diff
@@ -294,82 +294,82 @@
...
           let(:gemspec_file_path) { File.join(base_path, 'example.gemspec') }
 
-          it 'sets target_ruby from required_ruby_version from exact version' do
+          it 'sets target_ruby from inclusive range' do
             content =
               <<-HEREDOC
                 Gem::Specification.new do |s|
                   s.name = 'test'
-                  s.required_ruby_version = '2.7.4'
+                  s.required_ruby_version = '>= 2.6.0'
```
marcandre, line 302: It should also work for `>= 2.6.1`

marcandre (changes requested):
Yes, it is almost ready 😅

lib/rubocop/target_ruby.rb
HeroProtagonist: Only one question. The `.99` would cause a not equal check like the ones I have [here](https://github.com/rubocop-hq/rubocop/pull/9515/files#diff-9ac8fa18e6383d5acc2f64f888efd29cd782ee80a85fa2c6e3090421f0d3e228R344) and [here](https://github.com/rubocop-hq/rubocop/pull/9515/files#diff-9ac8fa18e6383d5acc2f64f888efd29cd782ee80a85fa2c6e3090421f0d3e228R410) to fail. Maybe that is ok? Not sure how common it is to be specifying not equal versions

marcandre: Seems to me it is an error to use `!=`. In any case, if someone did specifies `!= 2.4.0`, that doesn't exclude `2.4.1` etc which are in the `2.4` series, so it seems like an ok target.

HeroProtagonist: Ok 1 more thing if we had `'> 2.4.1'` then it would match to `2.4` due to the `.99`, that seems wrong?

marcandre: I think that's the best behavior. Imagine that there is a bug in 2.4.1 that is fixed in Ruby 2.4.2+; the gem author wants to make sure we don't try to run the gem in 2.4.1 or below as the bug can be triggered by code in the gem. The gem still uses the "2.4.x" features and syntax, so target of 2.4 is best.

HeroProtagonist: Ok push the updates :)

marcandre (approved):
Good stuff! 💪
Thanks 😄

PR #9527, opened 2021-02-19
[#9526] Add `AllowSplatArgument` option to `Style/HashConversion`

### LINKED ISSUE #9526: Style/HashConversion splatted array should not error
## Expected behavior

It would be nice to enable `Style/HashConversion` but not flag `Hash[*values]` usage as invalid.

## Actual behavior

Getting the following error when using splatting an array to a hash:

``` 
snippet.rb:4:16: C: Style/HashConversion: Prefer array_of_pairs.to_h to Hash[*array].
  def to_env = Hash[*values]
```

Create the following files locally to reproduce:

``` ruby
# File: .rubocop.yml
Style/HashConversion:
  Enabled: true
```

``` ruby
# frozen_string_literal: true

# File: snippet.rb
Pair = Struct.new :key, :value do
  def to_env = Hash[*values]
end
```

Now run the following:

`rubocop snippet.rb`

Notice you'll get the following error:

``` 
snippet.rb:4:16: C: Style/HashConversion: Prefer array_of_pairs.to_h to Hash[*array].
  def to_env = Hash[*values]
```

## RuboCop version

```
# rubocop -V
1.10.0 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 3.0.0 x86_64-darwin20)
```

koic:
Resolves #9526.

This PR add `AllowSplatArgument` option to `Style/HashConversion` and the option is true by default.
This is because `Hash[*ary]` is not allowed because it is strict.

As discussed in https://github.com/rubocop-hq/rubocop/pull/9478#discussion_r567259779, the correction code from `Hash[*arg]` is not simply determined. So, I think it's better to allow it by default as a simple notation.

On the other hand, there may be case where some users want to use the existing behavior, so this PR have made it optional.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/hash_conversion.rb
+++ b/lib/rubocop/cop/style/hash_conversion.rb
@@ -6,6 +6,10 @@ module RuboCop
       # This cop checks the usage of pre-2.1 `Hash[args]` method of converting enumerables and
       # sequences of values to hashes.
       #
+      # Correction code from splat argument (`Hash[*ary]`) is not simply determined. For example,
+      # `Hash[*ary]` can be replaced with `ary.each_slice(2).to_h` but it will be complicated.
+      # So, `AllowSplatArgument` option is true by default to allow splat argument for simple code.
+      #
       # @example
       #   # bad
       #   Hash[ary]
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/style/hash_conversion.rb
```diff
@@ -18,6 +18,15 @@ module Style
       #
       #   # good
       #   {key1 => value1, key2 => value2}
+      #
+      # @example AllowSplatArgument: true (default)
```
bbatsov, line 22: Might be a good idea to add some explanation for the need for some such configuration option, as without the context from the ticket I don't think that'd be obvious.

koic: Sure! I added an explanation for the configuration option.

PR #9535, opened 2021-02-24
[Fix #9525] Add `AllowIfMethodWithArguments` option to `Style/SymbolProc`

### LINKED ISSUE #9525: Add option to Style/SymbolProc to ignore methods with parameters

No.

My application has some code like this:

```ruby
on_scheduled_jobs(worker_class_name, some_arg) do |job|
  job.delete
end

after_transition(after_commit: true, to: :deleted) do |project|
  project.clear_indexes
end
```

Rubocop detects offense and suggests change my code to:
```ruby
on_scheduled_jobs(worker_class_name, first_arg_c, &:delete)

after_transition(after_commit: true, to: :deleted, &:clear_indexes)
```

But in some scenarios, using symbol as proc isn't the best option for code style because don't follow other block definitions style. I know I have some alternatives like disable `Style/SymbolProc`, add methods to IgnoredMethods option or use `# rubocop:disable Style/SymbolProc` for some cases, but I think a option to ignore methods with parameters would be nice.

Add option (maybe a EnforcedStyle??) to ignore `Style/SymbolProc` for methods with parameters, or option to only enable `Style/SymbolProc` for methods without parameters.

For example:
```yaml
# rubocop.yml
Style/SymbolProc:
  IgnoreMethodsWithParameters: true
```
```ruby
# code.rb
users.each { |u| u.send_email } # Detects offense

after_transition(to: :paid) do |payment| # Doesn`t detect offense
   payment.send_notification
end
  
```

koic:
## Summary

Fixes #9525.

This PR adds `AllowIfMethodWithArguments` option to `Style/SymbolProc`.

### `AllowIfMethodWithArguments: false` (default)

```ruby
# bad
something.do_something(foo) { |o| o.bar }

# good
something.do_something(foo, &:bar)
```

### `AllowIfMethodWithArguments: true`

```ruby
# good
something.do_something(foo) { |o| o.bar }
```

If user prefer a style that allows block for method with arguments, the user can set `true` to `AllowIfMethodWithArguments`.

## Other Information

It may be possible to consider defaulting to `AllowIfMethodWithArguments: true`, but at the time of this patch `Style/SymbolProc` cop respect keeping the existing behavior.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/symbol_proc.rb
+++ b/lib/rubocop/cop/style/symbol_proc.rb
@@ -5,6 +5,9 @@ module RuboCop
     module Style
       # Use symbols as procs when possible.
       #
+      # If you prefer a style that allows block for method with arguments,
+      # please set `true` to `AllowMethodsWithArguments`.
+      #
       # @example
       #   # bad
       #   something.map { |s| s.upcase }
@@ -12,6 +15,17 @@ module RuboCop
       #
       #   # good
       #   something.map(&:upcase)
+      #
+      # @example AllowMethodsWithArguments: false (default)
+      #   # bad
+      #   something.do_something(foo) { |o| o.bar }
+      #
+      #   # good
+      #   something.do_something(foo, &:bar)
+      #
+      # @example AllowMethodsWithArguments: true
+      #   # good
+      #   something.do_something(foo) { |o| o.bar }
       class SymbolProc < Base
         include RangeHelp
         include IgnoredMethods
@@ -42,6 +56,7 @@ module RuboCop
             return if proc_node?(dispatch_node)
             return if %i[lambda proc].include?(dispatch_node.method_name)
             return if ignored_method?(dispatch_node.method_name)
+            return if allow_if_method_has_argument?(node)
             return if node.block_type? && destructuring_block_argument?(arguments_node)
 
             register_offense(node, method_name, dispatch_node.method_name)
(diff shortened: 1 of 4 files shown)


stephannv (approved):

lib/rubocop/cop/style/symbol_proc.rb
```diff
@@ -103,6 +118,10 @@ def begin_pos_for_replacement(node)
             node.loc.begin.begin_pos
           end
         end
+
+        def allow_if_mothod_has_argument?(node)
```
stephannv, line 122: Please, fix this typo: `mothod`. Line 59 too.

bbatsov:
I'm fine with the proposed patch, but I think the name of the configuration option is quite confusing. `AllowMethodsWith(Additional)Arguments` seems a bit better to me (as you can argue that the block is an argument as well), but I'm wondering if we can do better. @rubocop-hq/rubocop-core who's best at naming here? 😆

dvandersluis:
👍 for `AllowMethodsWithArguments`. It's kind of a hard concept to name. I was thinking something like `MethodsWithBlockArgumentOnly` but that's clumsier IMO.

koic:
Thank you for the feedback! I renamed the option `AllowMethodsWithArguments`.

> I was thinking something like `MethodsWithBlockArgumentOnly`

I think this name is also a good candidate :-)

bbatsov:
Thanks!

PR #9538, opened 2021-02-25
Bug: Layout/FirstArgumentIndentation false negative when method name is one special character

TSMMark:
It is uncommon but completely valid to have methods names like `|`, `+`, `~`, `&`, etc. Some prominent libraries make use of this. For example, Sequel ORM has `Sequel.|` method which can be used to build a SQL `OR` clause, along with `Sequel.&`, `Sequel.~`, etc.

Example:

```
3.0.0 :005 > def |(a, b) = a + b
 => :|
3.0.0 :006 > self.|(1, 2)
 => 3
3.0.0 :007 > def +(a, b) = a + b
 => :+
3.0.0 :008 > self.+(1, 2)
 => 3
```

The Layout/FirstArgumentIndentation cop does not seem to register an offense when these special method names are invoked, as reproduced in the added specs in this branch.

I took a look thru some of the relevant source code of FirstArgumentIndentation, but as a cop author novice it's not obvious to me exactly what needs to change. 

---

Before submitting the PR make sure the following are checked:

* [ ] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [ ] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [ ] Feature branch is up-to-date with `master` (if not - rebase it).
* [ ] Squashed related commits together.
* [x] Added tests.
* [ ] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [ ] Added an entry (file) to the [changelog folder](https://github.com/rubocop-hq/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #9542, opened 2021-02-25
Fix `Layout/FirstArgumentIndentation` for operator methods not called as operators.

dvandersluis:
Continuation of #9538.

Operators were explicitly allowed by `Layout/FirstArgumentIndentation`, however when an operator method is called as a method (ie. not as a bare operator), the cop should still apply.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/first_argument_indentation.rb
+++ b/lib/rubocop/cop/layout/first_argument_indentation.rb
@@ -153,7 +153,7 @@ module RuboCop
 
         def on_send(node)
           return if style != :consistent && enforce_first_argument_with_fixed_indentation?
-          return if !node.arguments? || node.operator_method?
+          return if !node.arguments? || bare_operator?(node)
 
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/layout/first_argument_indentation.rb
```diff
@@ -167,6 +167,10 @@ def autocorrect(node)
 
         private
 
+        def bare_operator?(node)
+          node.operator_method? && !node.dot?
```
TSMMark, line 171: 👏

TSMMark (approved):
I added this commit hash ref to Gemfile and tested running rubocop one file and it seems to solve the issue perfectly. Awesome!

koic:
This change looks good. Can you squash the commits into one and add the co-authors to the squashed commit message?
https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

PR #9546, opened 2021-02-26
Add rescue nil config to suppressed exceptions

corroded:
Based off: https://github.com/rubocop/rubocop/pull/7297

Also the comment by @bbatsov https://github.com/rubocop/rubocop/pull/7297#issuecomment-523501395

> If some people disagree we can extend the existing behaviour with some config option, but I definitely imagine cases where a return value of nil might be the reasonable thing to do.

I personally have been bitten by explicit `rescue nil`s and would prefer to flag it. I've set `AllowRescueNil` to default to `true` so as not to break existing configurations. Setting it to false renders `rescue nil` and 

```
begin
  do_something
rescue
  nil
end
```

as offences.
-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2055,8 +2055,9 @@ Lint/SuppressedException:
   StyleGuide: '#dont-hide-exceptions'
   Enabled: true
   AllowComments: true
+  AllowNil: true
   VersionAdded: '0.9'
-  VersionChanged: '0.81'
+  VersionChanged: <<next>>
 
 Lint/SymbolConversion:
   Description: 'Checks for unnecessary symbol conversions.'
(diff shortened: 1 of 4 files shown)


config/default.yml
```diff
@@ -2055,8 +2055,9 @@ Lint/SuppressedException:
   StyleGuide: '#dont-hide-exceptions'
   Enabled: true
   AllowComments: true
+  AllowRescueNil: true
```
bbatsov, line 2058: I think `AllowNil` would be a better name here.

bbatsov:
Apart from my small remark I'm fine with the proposed change.

config/default.yml
corroded: Makes sense! Will fix it!

bbatsov (approved):

PR #9547, opened 2021-02-26
[Fix #9533] Fix calc of multiline hash arg length for metrics length cops

### LINKED ISSUE #9533: Metrics/BlockLength (and alike) should be able to ignore multi-line method calls

I’m always frustrated when RuboCop complains when I changed a method call with multiple parameters to span multiple lines and thus broke the block length limit.
While it helps to use keyword arguments and add `hash` to `CountAsOne` this is not yet perfect.

For example, the statement
```ruby
foo(
  arg1: 1,
  arg2: 2,
  arg3: 3,
)
```
counts as 3 lines, but
```ruby
foo({
  arg1: 1,
  arg2: 2,
  arg3: 3,
})
```
only counts as 1.

I’d like an option for `CountAsOne` like `method` or alike which allows to break a method call onto multiple lines without counting towards the block length limit.

For keyword args it helps to use `hash` in `CountAsOne` but the start and end line are still counted unless you explicitly wrap the args into a hash.
For positional args one could wrap them into `*[…]` but this lets `Lint/RedundantSplatExpansion` complain.

## Additional context

Probably the same should be added to `Class-`, `Method-` and `ModuleLength`.

koic:
Fixes #9533.

This PR makes calculation of length of multiline hash argument with and without brace the same.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_make_metrics_length_cops_aware_of_multi_line_kwargs.md
+++ b/changelog/fix_make_metrics_length_cops_aware_of_multi_line_kwargs.md
@@ -1,0 +1,1 @@
+* [#9533](https://github.com/rubocop/rubocop/issues/9533): Make metrics length cops aware of multi-line kwargs. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #9553, opened 2021-03-01
Add autocorrect for `Style/UnlessLogicalOperators`

SeekingMeaning:
This PR adds autocorrect for `unless ... &&` and `unless ... ||`

For example:

```ruby
return unless a || b
```

is auto-corrected to:

```ruby
return if !a && !b
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_add_autocorrect_for_styleunlesslogicaloperators.md
+++ b/changelog/new_add_autocorrect_for_styleunlesslogicaloperators.md
@@ -1,0 +1,1 @@
+* Add autocorrect for `Style/UnlessLogicalOperators`. ([@SeekingMeaning][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Will the auto-corrector handle cases like `return unless !a || !b` or `return unless !a && b`?

dvandersluis:
I’m not so sure that correcting by inverting the clause is that useful. The resulting clauses are pretty complicated and not easy to read. It’d be a lot clearer to just refactor the condition (for example, create a method to encapsulate the logic and then call `unless condition?`), but autocorrect can’t do that.

bbatsov:
Good point. I was thinking about the same myself - those negations don't really make anything easier to read, quite the contrary in  test examples. Just replacing the `unless` with an equivalent `if` is not always an improvement.

kadirselcuk (approved):

PR #9557, opened 2021-03-05
Add example for style trailing body on method definition

koic:
Follow https://github.com/rubocop/rubocop/pull/9548#issuecomment-787423830.

This PR adds example for `Style/TrailingBodyOnMethodDefinition` cop and removes a redundant condition from the cop. I noticed the redundant condition when writing the example description.


bbatsov:
Thanks!

PR #9559, opened 2021-03-06
[Fix #9505] Disable cops on next line with directive comment

### LINKED ISSUE #9505: Ability to disable cop for next line
## Problem

When disabling a cop, we have two options, each with something I consider a problem:
- Disabling a cop for a line with a comment at the end of it. Sometimes this results in a line that is too long and it's uncomfortable to read
- Disabling a rule in a block using `rubocop:disable` and `rubocop:enable`. This could be cumbersome if I want to disable for a single line

## Solution

I think a directive to disable the next line would be good, similar to what [eslint has](https://eslint.org/docs/2.13.1/user-guide/configuring#disabling-rules-with-inline-comments). It could be `rubocop:disable:next-line` or `rubocop:disable-next-line` for example

## Additional context

There's already [an old closed issue](https://github.com/rubocop-hq/rubocop/issues/4569) regarding this topic, sorry for the duplicate. I would like to know if there's support for a feature like this and/or if the rubocop team would still be willing to review a PR for this.

AndreiEres:
For https://github.com/rubocop/rubocop/issues/9505

- Added checking for `disable-next-line` to DirectiveComment
- Added ability to disable cops on next line with directive comment to CommentConfig
- Added notes about feature to docs

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/comment_config.rb
+++ b/lib/rubocop/comment_config.rb
@@ -153,6 +153,8 @@ module RuboCop
     # so that `Lint/RedundantCopEnableDirective` can register offenses correctly.
     def handle_switch(directive, names, extras)
       directive.cop_names.each do |name|
+        next if directive.disabled_next_line?
+
         if directive.disabled?
           names[name] += 1
         elsif (names[name]).positive?
--- a/lib/rubocop/comment_config.rb
+++ b/lib/rubocop/comment_config.rb
@@ -68,6 +68,8 @@ module RuboCop
       # Disabling cops after comments like `#=SomeDslDirective` does not related to single line
       if !comment_only_line?(directive.line_number) || directive.single_line?
         analyze_single_line(analysis, directive)
+      elsif directive.disabled_next_line?
+        analyze_next_line(analysis, directive)
       elsif directive.disabled?
         analyze_disabled(analysis, directive)
       else
@@ -79,6 +81,13 @@ module RuboCop
       return analysis unless directive.disabled?
 
       line = directive.line_number
+      start_line = analysis.start_line_number
+
+      CopAnalysis.new(analysis.line_ranges + [(line..line)], start_line)
+    end
+
+    def analyze_next_line(analysis, directive)
+      line = directive.line_number + 1
       start_line = analysis.start_line_number
 
       CopAnalysis.new(analysis.line_ranges + [(line..line)], start_line)
--- a/lib/rubocop/directive_comment.rb
+++ b/lib/rubocop/directive_comment.rb
@@ -14,7 +14,7 @@ module RuboCop
     # @api private
     COPS_PATTERN = "(all|#{COP_NAMES_PATTERN})"
     # @api private
-    DIRECTIVE_COMMENT_MODES = '(?:disable|disable-next-line|enable|todo)'
+    DIRECTIVE_COMMENT_MODES = '(?:disable|disable-next-line|enable|todo|todo-next-line)'
     # @api private
     DIRECTIVE_COMMENT_REGEXP = Regexp.new(
       "# rubocop : (#{DIRECTIVE_COMMENT_MODES})\\b #{COPS_PATTERN}".gsub(' ', '\s*')
@@ -57,7 +57,7 @@ module RuboCop
 
     # Checks if this directive disables cops only on next line
     def disabled_next_line?
-      %w[disable-next-line].include?(mode)
+      %w[disable-next-line todo-next-line].include?(mode)
     end
 
     # Checks if this directive enables cops
--- a/lib/rubocop/directive_comment.rb
+++ b/lib/rubocop/directive_comment.rb
@@ -14,9 +14,10 @@ module RuboCop
     # @api private
     COPS_PATTERN = "(all|#{COP_NAMES_PATTERN})"
     # @api private
+    DIRECTIVE_COMMENT_MODES = '(?:disable|disable-next-line|enable|todo)'
+    # @api private
     DIRECTIVE_COMMENT_REGEXP = Regexp.new(
-      "# rubocop : ((?:disable|enable|todo))\\b #{COPS_PATTERN}"
-        .gsub(' ', '\s*')
+      "# rubocop : (#{DIRECTIVE_COMMENT_MODES})\\b #{COPS_PATTERN}".gsub(' ', '\s*')
     )
 
     def self.before_comment(line)
@@ -52,6 +53,11 @@ module RuboCop
     # Checks if this directive disables cops
     def disabled?
       %w[disable todo].include?(mode)
+    end
+
+    # Checks if this directive disables cops only on next line
+    def disabled_next_line?
+      %w[disable-next-line].include?(mode)
     end
 
     # Checks if this directive enables cops
--- a/spec/rubocop/comment_config_spec.rb
+++ b/spec/rubocop/comment_config_spec.rb
@@ -179,8 +179,7 @@ RSpec.describe RuboCop::CommentConfig do
     end
 
     it 'supports disabling cops on next line' do
-      expect(disabled_lines_of_cop('Style/EmptyMethod'))
-        .to eq([7, 8, 9, 56])
+      expect(disabled_lines_of_cop('Style/EmptyMethod')).to eq([7, 8, 9, 56])
     end
   end
 
--- a/spec/rubocop/comment_config_spec.rb
+++ b/spec/rubocop/comment_config_spec.rb
@@ -211,9 +211,7 @@ RSpec.describe RuboCop::CommentConfig do
     let(:source) do
       <<~RUBY
         # rubocop:enable Metrics/MethodLength, Security/Eval
-        def some_method
-          puts 'foo'
-        end
+        def some_method; end
       RUBY
     end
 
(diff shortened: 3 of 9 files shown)


spec/rubocop/comment_config_spec.rb
```diff
@@ -178,5 +181,10 @@ def disabled_lines_of_cop(cop)
       expect(disabled_lines_of_cop('Layout/LeadingCommentSpace'))
         .to eq([7, 8, 9, 50])
     end
+
+    it 'supports disabling cops for next line' do
+      expect(disabled_lines_of_cop('Style/DoubleNegation'))
+        .to eq([7, 8, 9, 56])
```
rrosenblum, line 187: The `disable all` from above threw me at first here. I was expecting to only see `56` at first.

lib/rubocop/comment_config.rb
```diff
@@ -20,6 +20,12 @@ class CommentConfig
         .gsub(' ', '\s*')
     )
 
+    # @api private
+    COMMENT_NEXT_LINE_DIRECTIVE_REGEXP = Regexp.new(
+      "# rubocop : ((?:disable-next-line))\\b #{COPS_PATTERN}"
+        .gsub(' ', '\s*')
```
rrosenblum, line 26: I didn't realize that we supported so much white space in the comments. I see we're doing the same thing above so this looks good to me.

rrosenblum (approved):
The change looks good to me. You'll have to rebase because there are now merge conflicts

lib/rubocop/directive_comment.rb
```diff
@@ -54,6 +55,11 @@ def disabled?
       %w[disable todo].include?(mode)
     end
 
+    # Checks if this directive disables cops only on next line
+    def disabled_next_line?
+      %w[disable-next-line].include?(mode)
+    end
```
dvandersluis, line 61: I think we should support `todo-next-line` as well to mirror how `disable`/`todo` work.

spec/rubocop/directive_comment_spec.rb
```diff
@@ -198,4 +198,20 @@
...
+      ['when disabled next line', '# rubocop:disable-next-line all', true],
+      ['when disabled current line', 'def foo # rubocop:disable all', false],
+      ['when disabled few lines', '# rubocop:disable Foo/Bar', false]
+    ].each do |example|
+      context example[0] do
+        let(:text) { example[1] }
+
+        it { is_expected.to eq example[2] }
+      end
+    end
```
dvandersluis, line 215: This could be replaced by a shared example for easier reading.

AndreiEres:
@dvandersluis @bbatsov 
That feature also required investigation as another one about disabling all department with directive comment.

AndreiEres:
@dvandersluis @bbatsov
So, I fixed new issues with redundant disabling and enabling. Now it seems like ready to use.

AndreiEres:
Covered edge cases. 
```
# rubocop:disable-next-line Foo/Bar
do_something
# rubocop:enable Foo/Bar <-- Lint/RedundantCopEnableDirective
```
```
# rubocop:disable-next-line Foo/Bar
do_something # rubocop:disable Foo/Bar <-- Lint/RedundantCopDisableDirective
```

lib/rubocop/directive_comment.rb
AndreiEres: Supported

spec/rubocop/directive_comment_spec.rb
AndreiEres: Replaced

AndreiEres:
@dvandersluis @bbatsov 
I covered few edge cases and fixed issues appeared with that new feature. Now it seems like ready to use in my opinion :-)

bbatsov:
I saw the PR and the conversation, but I have responded yet mostly because I'm still not convinced we need this to begin with. One can already do single-line disables with inline comments and adding something for the single line that follows the comment seems oddly specific, even if ESLint is doing it. I think there's be a lot more value in a comment that actions on the next expression in the sense of a definition, conditional, method with a block, etc. 

Someone has to make a strong case for those `next-line` comments before I'm ready to approve them.

ngouy:
TBH it's much more readable (I guess it's only a point of view) to have a next_line disable rather than adding it to an existing line

```rb
# rubocop:disable:next_line ACop/WithAVeryVeryLongName
my_line_is_already_long_enought = "because my code is endented and my basic config is 80/120 lines"
```

vs
```
my_line_is_already_long_enought = "because my code is endented and my basic config is 80/120 lines" # # rubocop:disable ACop/WithAVeryVeryLongName
```

or in a screen when you have many sections : 

image

ngouy:
> I think there's be a lot more value in a comment that actions on the next expression in the sense of a definition, conditional, method with a block, etc.

I 💯 agree with that

But in another hand, it may seem like a bad argument, but wouldn't it be really much more complicated to integrate/maintain? Will someone gave it a try?

If this solution
- works and helps many people looking for it
- only adds new features without touching the existing
- and that has no negative impact on existing or new users

what is the issue here?
Is it about altering the codebase without being sure about the utility of this feature?
Will this feature prevent / will complicate the integration of what you describe if it happens?

Or put it otherwise: what are the cons?

sandstrom:
This is awesome! I was looking for this feature and about to ask about it, then saw that it's already almost completed! 🎉 😄

bbatsov:
Well, there hasn't been much traction here recently so I'm not sure if we'll continue in this direction. I still think it's preferable to have something that operates on the expression level and I don't see much demand for the "next-line" approach. If I recall correctly @AndreiEres wasn't particularly interested in "next-line" for his own needs (please, correct me if I'm wrong), so we may just close this PR.

thisismydesign:
FWIW in my experience, many devs are used to that functionality from eslint and miss it for rubocop.

bbatsov:
Well, that's what started the discussion here originally, but I don't think that copying features blindly from other tools is always a good idea. I know that `eslint` is very popular, but there's always room for improvement. :-)

thisismydesign:
Not blindly, as I said many of us use and love that feature. What's wrong with it, so that it would an improvement not to have it?

bbatsov:
As I said several times already, I don't see what exactly does it solve. It adds some complexity to our codebase (as evidenced by this PR) and it's limited to a very specific use-case - an expression that's exactly one line long. As this can already be handled with both fenced comments and inline comments I don't want to go overboard with a 3rd solution for the same problem. I said I like the idea of a special comment that disables a check for the span of the next expression and no one has made a strong case for `next_line` other than "eslint has it" (I think fencing is fine for long lines).

bbatsov:
Also - in general I prefer naming like `# rubocop:disable:next_line`, `# rubocop:disable:next_expr`, etc. I think it reads a bit better.

thisismydesign:
Here are a few arguments:

- I think "eslint has it and a lot of ppl use it" is a pretty good argument. I don't know what is unclear about what this solves. It's the go-to solution in every eslint codebase I've ever seen for disabling rules because it's short and easy to read. So I don't understand in what way "it's limited to a very specific use-case". If it is, then that very specific use case happens all the time.

- This solution is superior to the other 2 because they are more cumbersome (fenced) or harder to read (inline). I can only refer back to eslint again: I almost always use next-line in practice.

- This has a PR for it. This has been a functionality that many of us were missing since forever. Maybe next expression is the coolest thing but when will it happen if ever?

- Does next expression actually make sense? I didn't look into it, did someone already investigate and make sure it's feasible? E.g. if I want to disable a rule about blocks on top of a block, what if it has more blocks inside of it that I wouldn't want to apply it to? I think there can be cases where it will cause confusion and false negatives. Are other libraries providing this feature? Do we know if people use it? I don't think so, but please correct me. On the face of it, it looks to me that we would be discarding a very popular feature in favor of an untested idea.

- "It adds some complexity to our codebase" Fair enough, and it seems to me this is the real issue here, but it is being supported by "this would be a useless / misdirected feature" which is definitely not the case imo.

bbatsov:
That fact that someone submitted something doesn't make it automatically a good idea. The fact that another tool has some feature doesn't automatically make it a good idea. I know that people have copied bad ideas from CIDER and RuboCop just because they are popular. :-) 

> Does next expression actually make sense? I didn't look into it, did someone already investigate and make sure it's feasible? E.g. if I want to disable a rule about blocks on top of a block, what if it has more blocks inside of it that I wouldn't want to apply it to?

Yeah, it's feasible (the boundaries of the expression are something we can get from the parser) and yeah, it will affected some nested block in case of more complex expressions (which should reduce the need for fenced blocks a lot). `next-line` doesn't make much sense for a multi-line expression anyways, so nothing lost here. Anyways, I get your perspective and it's appreciated, but in the end of the day I remain unconvinced of both the demand and the utility of the next-line approach. 

In the end of the day, my role as a maintainer is to find the right balance between the needs of our users and the quality and maintainability of our codebase. That often involves rejecting seemingly good ideas, and that's fine. As I wrote this it did strike me that a variant of this that applies to the next X lines might be an improvement, as at least it won't be super specific.

ngouy:
> I remain unconvinced of both the demand and the utility

That is what can be frustrating for users.
A lot (maybe not a lot but at least some) people are asking for that, some ever are willing to help develop it
Some arguments are given

And the arguments about this feature on the other side is:
- I'm not convinced its useful (which can be seen as a very opiniated not-really-an-argument)

I understand as a maintainer you have constraints and a huge responsibility, and that we all benefit from what you are doing. Thank you for that 👍 

But when a "no" comes it's always appreciated when arguments other than "I don't feel it" is given while some of us gave clear and detailed arguments on why they miss it.

bbatsov:
There's nothing more I can say on the subject. I'm sorry if someone's offended by my reasoning, but at the end of the day the burden for all decisions lies with me. Some decisions are going to be disappointing to some people, there's no recipe to make everyone happy.

bbatsov:
@alfuken It's easy to complain, but it's not so easy to maintain a huge project for over a decade. Sounds wild, right?

Yet here we are...

On a less bitter note - I've said everything I had to say on the subject [here](https://github.com/rubocop/rubocop/pull/9559#issuecomment-930110498). If someone reworks proposal this to work on expressions instead of lines I'll definitely reconsider it.
Btw, your generic complaints are certainly not helping make the case for why is feature so important - so far no one explained why exactly would this special comment be so important to have. You can't make me reconsider anything by shaming me, but you can make a lot of progress with laying down some good arguments about the merits of something.

PEZ:
@alfuken, For a maintainer of a project it is not enough that some other people (even many) see something as useful. You _have_ to see how it is useful yourself, or you will be maintaining a mess some steps down the road. You have to see so much value in it that it motivates the maintenance burden it adds. All features add maintenance burden.

It is great that people are ready to put in the time to implement, but even that is not enough. It is still the maintainer that will need to maintain every single line of code in the project. Maintaining code where you don't understand the value it adds is not fun at all.

This is open source stuff. If you really want a feature, but fail to show @bbatsov how it is useful, you can always maintain your own fork. Then that feature will be maintained by the right person, the person who sees the value with it.

fmnoise:
@alfuken please read this first
https://gist.github.com/richhickey/1563cddea1002958f96e7ba9519972d9

ngouy:
@alfuken I am kinda with you. But we (at least I) really lack of context on global view on this project: Even though I participated here and there, for me the experience was the classical "the more I participate the more I realize I don't know sh*t" (pardon my french)

I guess if someone comes up with a perfect solution it won't get rejected.

But it doesn't exist and I think that even if a lot of people want that (or have the impression they need that), it's not that simple (It's like peace. Everyone wants it. But no one came up with a good solution so far)

We can and should trust the maintainers to take the best decisions. If they wouldn't be able to do that, this project wouldn't be so mature (even though it shouldn't be an excuse to challenge them here and there 😉,  but it already has been done on this subject).

fmnoise:
@alfuken have you read this? https://gist.github.com/richhickey/1563cddea1002958f96e7ba9519972d9

bbatsov:
I'm getting really tired of this, so I'll just lock this thread. I'm fine with getting bad feedback, but I'm not fine with wasting my time in conversations that are leading nowhere. 

If someone is willing to implement a variation of the original proposal that's working on the next expressions (regardless of how many lines it is) (e.g. `#rubocop:disable:next`) I'd be happy to consider it. In case it's not clear so far:

- I get the use-case
- I don't think it warrants a special solution, if we can just tackle the broader "next expression" problem, which solves the original problem and then some
- I get some of you are frustrated, and that's OK

> After all, negative feedback is no less important that affirmative. No one dared telling Putin the truth and the bad news, all reports were positive, reassuring, and oh so great! Now look at what this has led to. And just because one is not Putin, it does not mean that one should not be receiving a feedback in which someone expressess disappointment or other not-so-happy things.

I'll admit this part amused me. Everyone's more than welcome to complain to me about anything, but if they don't lay out their arguments in a way to actually speaks to me they are just wasting their time and mine. I tell people there are better ways to solve a particular issue, yet they are so fucking attached to their solution I end up being compared with a crazy dictator.

PR #9562, opened 2021-03-08
Change Style/Documentation to not require documentation for ClassMethods modules

Drenmi:
## Background

It is a common idiom in ruby to define a module named `ClassMethods` which is automatically extended when it's parent module is included, e.g.:

```ruby
module Foo
  module ClassMethods
    def bar
      # ...
    end
  end

  def self.included(klass)
    klass.extend(ClassMethods)
  end
end
```

This inner module does not need additional documentation beyond that of the parent module.

## What is this change?

This PR makes it so that `Style/Documentation` does not register offenses on modules named `ClassMethods`.

---

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* Change `Style/Documentation` to not require documentation for `ClassMethods` modules. ([@drenmi][])
+
 ## 1.11.0 (2021-03-01)
 
 ### New features
--- a/changelog/change_documentation_to_allow_class_methods.md
+++ b/changelog/change_documentation_to_allow_class_methods.md
@@ -1,0 +1,1 @@
+* Change `Style/Documentation` to not require documentation for `ClassMethods` modules. ([@drenmi][])
--- a/lib/rubocop/cop/style/documentation.rb
+++ b/lib/rubocop/cop/style/documentation.rb
@@ -76,6 +76,11 @@ module RuboCop
           (send nil? {:public_constant :private_constant} ({sym str} _))
         PATTERN
 
+        # @!method class_methods?(node)
+        def_node_matcher :class_methods?, <<~PATTERN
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* Change `Style/Documentation` to not require documentation for `ClassMethods` modules. ([@drenmi][])
```
koic, line 7: No direct editing to the CHANGELOG.md is required because existence of changelog/change_documentation_to_allow_class_methods.md :-)

koic:
Often `ClassMethods` is a module of `ActiveSupport::Concern`, so I feel that it is based on Rails. I'm wondering if RuboCop core should support it by default. At least I think it needs to be mentioned in the documentation for users.

marcandre:
I think defining a `ClassMethods` module is a pattern used outside of Rails too, I'm 👍 with this proposal.

koic:
> defining a `ClassMethods` module is a pattern used outside of Rails too

Ah I see! It would be perfect if mentioned in the documentation.

bbatsov:
I'm fine with the proposed change, but I also agree it should be documented. I'm also thinking it'd better to have a list of `AllowedModules` or something like this and just make this the default value for the list.

Drenmi:
> I'm fine with the proposed change, but I also agree it should be documented. I'm also thinking it'd better to have a list of AllowedModules or something like this and just make this the default value for the list.

Sure! I can make this happen. 👍

bbatsov:
Thanks!

bbatsov:
@Drenmi Ping. :-) If you don't have time to make this configurable you can just resolve the merge conflict and merge it.

Drenmi:
It seems someone already [beat me to it](https://github.com/rubocop/rubocop/commit/3f10d2e063aaa7e3058fb06d625b1552a469f146). 🙂

PR #5361, opened 2017-12-30
[Fix #5350] Resolve Metrics/LineLength false offenses with double quotes

garettarrowood:
See #5350 for description of bug.

[Here is the root of the problem](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/metrics/line_length.rb#L129). For some reason, when `URI::DEFAULT_PARSER.make_regexp` is passed a double quoted string, it returns MatchData [here](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/metrics/line_length.rb#L116) that is one character shorter.  When this finally reaches `allowed_uri_position?`, the two test examples report different ending characters.

Here are the <MatchData> objects themselves:

#### Single quotes

```
=> #<MatchData
 "https://github.com/bbatsov/rubocop/commit/3b48d8bdf5b1c2e05e35061837309890f04ab08c'"
 1:"https"
 2:nil
 3:nil
 4:"github.com"
 5:nil
 6:nil
 7:"/bbatsov/rubocop/commit/3b48d8bdf5b1c2e05e35061837309890f04ab08c'"
 8:nil
 9:nil>
```

#### Double quotes

```
=> #<MatchData
 "https://github.com/bbatsov/rubocop/commit/3b48d8bdf5b1c2e05e35061837309890f04ab08c"
 1:"https"
 2:nil
 3:nil
 4:"github.com"
 5:nil
 6:nil
 7:"/bbatsov/rubocop/commit/3b48d8bdf5b1c2e05e35061837309890f04ab08c"
 8:nil
 9:nil>
```

-----------------

* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](../blob/master/CONTRIBUTING.md#changelog-entry-format).
* [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.
* [x] Run `rake default` or `rake parallel`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: http://chris.beams.io/posts/git-commit/

PR #9568, opened 2021-03-08
[Fix #9558] Fix inconsistency when dealing with URIs that are wrapped in single quotes vs double quotes

### LINKED ISSUE #9558: Inconsistent LineLength behavior with URL-like strings
When using double-quoted strings, a long URL-like string unexpectedly triggers a LineLength offense when it is used as argument to a method call using parentheses. This does not happen for single-quoted strings.

## Expected behavior

None of the following lines should trigger a LineLength offense:

```ruby
  foo = "http://www.example.com/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/'"
  bar("http://www.example.com/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/'")
  bar('http://www.example.com/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/')
  bar "http://www.example.com/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/'"
```

## Actual behavior

An LineLength offense is output for the line

```ruby
  bar("http://www.example.com/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/'")
```

1. Create a file `foo.rb` containing
```ruby
# frozen_string_literal: true

def long
  foo = "http://www.example.com/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/'"
  bar("http://www.example.com/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/'")
  bar('http://www.example.com/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/')
  bar "http://www.example.com/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/foo/bar/baz/'"
  foo
end
```
2. Create a file `.rubocop.yml` containing
```yaml
AllCops:
  NewCops: enable
```
3. Run `rubocop`

## RuboCop version

```
$ rubocop -V
1.11.0 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 3.0.0 x86_64-linux)
```

dvandersluis:
The `LineLength` cop uses `LineLengthHelp#allowed_uri_position?` to determine if the URI is in an allowed position. Previously it checked that the URL either ended the line, or ended one character before the end of the line:

https://github.com/rubocop/rubocop/blob/291aec7f04775dbe8aaff0c6a2d7328aaca0427b/lib/rubocop/cop/mixin/line_length_help.rb#L26-L30

It looks like this was added originally in #5361... also to handle disparity between single and double quotes. 🤔 

As it turns out, ruby's default URI parser (which we use to match parts of strings as a URL), accepts `'` as a valid character, but does not accept `"`. It also matches `)`. Therefore, the URL extracted from the string differs depending on the type of quote being used:

```ruby
regexp = URI::DEFAULT_PARSER.make_regexp(['http', 'https'])
regexp.match(%q[http://foo.com/')])
=> #<MatchData "http://foo.com/')" 1:"http" 2:nil 3:nil 4:"foo.com" 5:nil 6:nil 7:"/')" 8:nil 9:nil>
regexp.match(%q[http://foo.com/")])
=> #<MatchData "http://foo.com/" 1:"http" 2:nil 3:nil 4:"foo.com" 5:nil 6:nil 7:"/" 8:nil 9:nil>
```

This was leading to the failure in the issue, because for the single quoted string, the "URL" ends on the last character (despite the `')` not actually being a part of the URL!), but for the double quoted string, the parser actually ends at the "real" end of the URL, and then `allowed_uri_position?` fails.

This change updates the `uri_range` to also capture any trailing characters until the next whitespace, and then just checks that the `uri_range` ends at the end of the line. This fixes the issue while also continuing to register an offense for subsequent words.

Fixes #9558.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_fix_inconsistency_when_dealing_with_uris.md
+++ b/changelog/fix_fix_inconsistency_when_dealing_with_uris.md
@@ -1,0 +1,1 @@
+* [#9558](https://github.com/rubocop/rubocop/issues/9558): Fix inconsistency when dealing with URIs that are wrapped in single quotes vs double quotes. ([@dvandersluis][])
--- a/lib/rubocop/cop/mixin/line_length_help.rb
+++ b/lib/rubocop/cop/mixin/line_length_help.rb
@@ -24,9 +24,7 @@ module RuboCop
       end
 
       def allowed_uri_position?(line, uri_range)
-        uri_range.begin < max_line_length &&
-          (uri_range.end == line_length(line) ||
-           uri_range.end == line_length(line) - 1)
+        uri_range.begin < max_line_length && uri_range.end == line_length(line)
       end
 
       def line_length(line)
@@ -40,6 +38,14 @@ module RuboCop
         begin_position, end_position = last_uri_match.offset(0).map do |pos|
           pos + indentation_difference(line)
         end
+
+        # Extend the end position until the start of the next word, if any.
+        # This allows for URIs that are wrapped in quotes or parens to be handled properly
+        # while not allowing additional words to be added after the URL.
+        if (match = line[end_position..line_length(line)]&.match(/^\S+(?=\s|$)/))
+          end_position += match.offset(0).last
+        end
+
         return nil if begin_position < max_line_length &&
                       end_position < max_line_length
 
--- a/spec/rubocop/cop/layout/line_length_spec.rb
+++ b/spec/rubocop/cop/layout/line_length_spec.rb
@@ -124,6 +124,27 @@ RSpec.describe RuboCop::Cop::Layout::LineLength, :config do
       end
     end
 
+    context 'and the excessive characters include part of a URL in double quotes' do
+      it 'does not include the quote as part of the offense' do
+        expect_offense(<<-RUBY)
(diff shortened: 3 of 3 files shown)


bbatsov (approved):
Nicely done!

Issue #9563, opened 2021-03-08
CommentConfig needs ♥️


AndreiEres:
This class parses the special `rubocop:disable` comments in a source and provides a way to check if each cop is enabled at arbitrary line.
`lib/rubocop/comment_config.rb`

## Problem

1. First of all it looks like a huge class. Over 200 lines.

2. Class has many public methods, but tested only `cop_enabled_at_line?`.

3. There are many cases with direct using private constant:

```
lib/rubocop/directive_comment.rb:
  16:       match = comment.text.match(CommentConfig::COMMENT_DIRECTIVE_REGEXP)

lib/rubocop/cop/lint/redundant_cop_disable_directive.rb:
  177:           match = comment.text.match(CommentConfig::COMMENT_DIRECTIVE_REGEXP)

lib/rubocop/cop/mixin/documentation_comment.rb:
  48:         CommentConfig::COMMENT_DIRECTIVE_REGEXP.match?(comment.text)

lib/rubocop/cop/mixin/line_length_help.rb:
  19:         comment.text.match?(CommentConfig::COMMENT_DIRECTIVE_REGEXP)
  82:         before_comment, = line.split(CommentConfig::COMMENT_DIRECTIVE_REGEXP)

lib/rubocop/cop/style/disable_cops_within_source_code_directive.rb:
  73:           match = CommentConfig::COMMENT_DIRECTIVE_REGEXP.match(comment.text)

lib/rubocop/ext/processed_source.rb:
  8:         @comment_config ||= CommentConfig.new(self)
```

## Soluttion

Refactor class. We have `DirectiveComment` that can get part of functionality. Add additional specs.

## One more thing

Later we can add `disable-next-line` feature. See: https://github.com/rubocop/rubocop/issues/9505

AndreiEres:
Here is first suggestion https://github.com/rubocop/rubocop/pull/9567

AndreiEres:
Related PR https://github.com/rubocop/rubocop/pull/9567

AndreiEres:
For third item I moved directive comment matching to DirectiveComment. 
https://github.com/rubocop/rubocop/pull/9569

Comments welcome :-)

AndreiEres:
Added readability to CommentConfig spec source by replacing string joining with heredoc.
https://github.com/rubocop/rubocop/pull/9577

Also comments welcome :-)

AndreiEres:
And added missing spec for CommentConfig#cop_disabled_line_ranges
[...]

PR #9589, opened 2021-03-10
Replace CommentConfig#directive_parts with DirectiveComment instance

AndreiEres:
Part of https://github.com/rubocop/rubocop/issues/9563

- CommentConfig#directive_parts replaced  with DirectiveComment instance.
- Further we're going to continue moving part of responsibility to DirectiveComment

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/directive_comment.rb
+++ b/lib/rubocop/directive_comment.rb
@@ -6,6 +6,8 @@ module RuboCop
   # cops it contains.
   class DirectiveComment
     # @api private
+    REDUNDANT_COP = 'Lint/RedundantCopDisableDirective'
+    # @api private
     COP_NAME_PATTERN = '([A-Z]\w+/)*(?:[A-Z]\w+)'
     # @api private
     COP_NAMES_PATTERN = "(?:#{COP_NAME_PATTERN} , )*#{COP_NAME_PATTERN}"
@@ -47,10 +49,29 @@ module RuboCop
       @match_captures ||= comment.text.match(DIRECTIVE_COMMENT_REGEXP)&.captures
     end
 
+    # Checks if this directive disables cops
+    def disabled?
+      %w[disable todo].include?(mode)
+    end
+
+    # Checks if all cops specified in this directive
+    def all_cops?
+      cops == 'all'
+    end
+
+    # Returns array of specified in this directive cop names
+    def cop_names
+      @cop_names ||= all_cops? ? all_cop_names : parsed_cop_names
+    end
+
     private
 
     def parsed_cop_names
       (cops || '').split(/,\s*/)
+    end
+
+    def all_cop_names
+      Cop::Registry.global.names - [REDUNDANT_COP]
     end
   end
 end
--- a/lib/rubocop/directive_comment.rb
+++ b/lib/rubocop/directive_comment.rb
@@ -21,18 +21,11 @@ module RuboCop
       line.split(DIRECTIVE_COMMENT_REGEXP).first
     end
 
-    attr_reader :comment
+    attr_reader :comment, :mode, :cops
 
     def initialize(comment)
       @comment = comment
-    end
-
-    # Return all the cops specified in the directive
-    def cops
-      return unless match_captures
-
-      cops_string = match_captures[1]
-      cops_string.split(/,\s*/).uniq.sort
+      @mode, @cops = match_captures
     end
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/directive_comment.rb
```diff
@@ -53,5 +48,31 @@ def range
...
+      return @disabled if defined? @disabled
+
+      @disabled = %w[disable todo].include?(switch)
+    end
+
+    def mentioned_all?
+      return @mentioned_all if defined? @mentioned_all
+
+      @mentioned_all = cops_string == 'all'
+    end
```
dvandersluis, line 62: This doesn't really seem like memoization is useful here.

lib/rubocop/directive_comment.rb
```diff
@@ -21,18 +23,11 @@ def self.before_comment(line)
       line.split(DIRECTIVE_COMMENT_REGEXP).first
     end
 
-    attr_reader :comment
+    attr_reader :comment, :switch, :cops_string
```
dvandersluis, line 26: I think we need a more descriptive name than `switch` because I'm not even sure what it means here. Maybe `mode`?

Also since you removed `cops` I think `@cops` would be more straightforward than `@cops_string`.

lib/rubocop/directive_comment.rb
```diff
@@ -42,7 +37,7 @@ def single_line?
 
     # Checks if this directive contains all the given cop names
     def match?(cop_names)
-      cops == cop_names.uniq.sort
+      parsed_cop_names&.uniq&.sort == cop_names.uniq.sort
```
dvandersluis, line 40: It'd be good if `parsed_cop_names` just returned `[]` instead of nil so that we wouldn't need this chain of `&.`

lib/rubocop/directive_comment.rb
AndreiEres: There are two places in CommentConfig when we check if mentioned all cops. Another one is here: https://github.com/rubocop/rubocop/blob/3148062e12160c33adecc4501bca316ce0e261b1/lib/rubocop/comment_config.rb#L169-L172

That why I chose memoization. But if you still think that it useless, I'll remove that ^-)

lib/rubocop/directive_comment.rb
AndreiEres: Thanks, replaced

lib/rubocop/directive_comment.rb
AndreiEres: thanks, fixed

AndreiEres:
@dvandersluis Could you take a look?

lib/rubocop/directive_comment.rb
```diff
@@ -53,5 +48,31 @@ def range
...
       @match_captures ||= comment.text.match(DIRECTIVE_COMMENT_REGEXP)&.captures
     end
+
+    def disabled?
+      return @disabled if defined? @disabled
+
+      @disabled = %w[disable todo].include?(mode)
+    end
+
+    def mentioned_all?
```
bbatsov, line 58: This doesn't seem like an optional name to me. Perhaps `all_cops?` would be better or something along those lines.

lib/rubocop/directive_comment.rb
bbatsov: Unless some benchmark proves decent performance gains I'd remove it, as it certainly makes the code more obscure.

lib/rubocop/directive_comment.rb
```diff
@@ -53,5 +48,31 @@ def range
     def match_captures
       @match_captures ||= comment.text.match(DIRECTIVE_COMMENT_REGEXP)&.captures
     end
+
+    def disabled?
```
bbatsov, line 52: I'd add YARD metadata to all public methods, as this in a relative important public API class.

lib/rubocop/directive_comment.rb
AndreiEres: Ok, done

lib/rubocop/directive_comment.rb
AndreiEres: Replaced

lib/rubocop/directive_comment.rb
AndreiEres: Added

bbatsov:
Thanks!

PR #9605, opened 2021-03-16
[Fixes #9487] Mark Naming/MemoizedInstanceVariableName as unsafe

### LINKED ISSUE #9487: Naming/MemoizedInstanceVariableName false positive
```
$ bundle exec rubocop -V
1.9.1 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 2.7.2 x86_64-linux)
```

No config

Input file

```rb
# frozen_string_literal: true

module MyModule
  # ...

  def f
    @myvar = other_method1
    @myvar ||= other_method2
  end
end
```

## Expected behavior

No error. The method isn't memoizing a value, it is recalculated every run.

## Actual behavior

```
$ bundle exec rubocop foo.rb 
Inspecting 1 file
C

Offenses:

foo.rb:3:1: C: Style/Documentation: Missing top-level module documentation comment.
module MyModule
^^^^^^
foo.rb:8:5: C: Naming/MemoizedInstanceVariableName: Memoized variable @myvar does not match method name f. Use @f instead.
    @myvar ||= other_method2
    ^^^^^^

```

A workaround to avoid the false positive would be to write the code as:

```rb
  def f
    @myvar = other_method1 || other_method2
  end
```

However, depending on the size and complexity of the expression, that may not be preferred. The assignment to the instance variable is the desirable side effect.

Suggestion: If final `||=` expression had a previous unconditional assignment, then this offense should not occur.

marcandre:


marcandre:
The cop would be improved if it was aware of the pattern:

```ruby
@foo = something
@foo ||= something_else
```

Even with this, it would still be unsafe though.

bbatsov:
I think it's a good to add a note in the cop's documentation explaining why the cop is unsafe.

marcandre:
Good point, thanks.

PR #9610, opened 2021-03-17
[Fix #9563] Refactor CommentConfig

### LINKED ISSUE #9563: CommentConfig needs ♥️

AndreiEres:
Last part of https://github.com/rubocop/rubocop/issues/9563

- Refactored CommentConfig
- Extra logic moved to DirectiveComment

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/comment_config.rb
+++ b/lib/rubocop/comment_config.rb
@@ -40,13 +40,13 @@ module RuboCop
     private
 
     def extra_enabled_comments_with_names(extras:, names:)
-      each_directive do |comment, cop_names, disabled|
-        next unless comment_only_line?(comment.loc.expression.line)
+      each_directive do |directive|
+        next unless comment_only_line?(directive.line_number)
 
-        if !disabled && enable_all?(comment)
-          handle_enable_all(names, extras, comment)
+        if directive.enabled_all?
+          handle_enable_all(directive, names, extras)
         else
-          handle_switch(cop_names, names, disabled, extras, comment)
+          handle_switch(directive, names, extras)
         end
       end
 
@@ -56,9 +56,11 @@ module RuboCop
     def analyze # rubocop:todo Metrics/AbcSize
       analyses = Hash.new { |hash, key| hash[key] = CopAnalysis.new([], nil) }
 
-      each_mentioned_cop do |cop_name, disabled, line, single_line|
-        analyses[cop_name] =
-          analyze_cop(analyses[cop_name], disabled, line, single_line)
+      each_directive do |directive|
+        directive.cop_names.each do |cop_name|
+          cop_name = qualified_cop_name(cop_name)
+          analyses[cop_name] = analyze_cop(analyses[cop_name], directive)
+        end
       end
 
       analyses.each_with_object({}) do |element, hash|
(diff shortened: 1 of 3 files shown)


bbatsov:
Did you check if any of those changes are going to affect the extensions? I'm not sure to which extent they might be using `CommentConfig` directly.

AndreiEres:
@bbatsov 
Good point.

I've checked this extensions for `/CommentConfig|comment_config/`:
- rubocop-performance
- rubocop-rails
- rubocop-rspec
- rubocop-minitest
- rubocop-rake
- rubocop-sequel
- rubocop-thread_safety
- rubocop-require_tools
- rubocop-i18n
- cookstyle
- rubocop-packaging
- rubocop-sorbet

And I've found one inclusion in `cookstyle` that allready broken by previous changes: https://github.com/chef/cookstyle/blob/master/lib/rubocop/monkey_patches/comment_config.rb

It is a monkey patch to use `# cookstyle:disable`. We move that responsibility from CommentConfig to DirectiveComment.

lib/rubocop/comment_config.rb
```diff
@@ -31,22 +31,21 @@ def extra_enabled_comments
       )
     end
 
-    def comment_only_line?(line_number)
-      non_comment_token_line_numbers.none? do |non_comment_line_number|
-        non_comment_line_number == line_number
-      end
+    # Checks if line contains only comment
+    def line_with_comment_only?(line_number)
```
bbatsov, line 35: What's the reason to rename this?

AndreiEres: I think it was misunderstanding what method do. First time I've read `comment_only_line` as it's comments only current line. Like so `def foo # rubocop:disable all`. But it really means another thing: line has only comment.

bbatsov: Well, it was meant to mean `comment-only line`, which is pretty much the same. If that's the only reason I'd rather go back to the original name to reduce the scope of the change to whatever's really meaningful.

AndreiEres: ok. I'll return it back

AndreiEres: it's done

bbatsov:
Thanks!

tas50:
> 
> And I've found one inclusion in `cookstyle` that's already broken by previous changes: https://github.com/chef/cookstyle/blob/master/lib/rubocop/monkey_patches/comment_config.rb
> 
> 
> 
> It is a monkey patch to use `# cookstyle:disable`. We move that responsibility from CommentConfig to DirectiveComment.
> 
> I've informed guys about that in issue https://github.com/chef/cookstyle/issues/839
> 
> 
> 

Thanks for the heads up. I figured we'd need to update that code when I saw all these refactors starting. We pin to an exact version of RuboCop so there's no impact to us until we go to update.

PR #9612, opened 2021-03-17
Allow parentheses in safe navigation calls for `Style/MethodCallWithArgsParentheses`

gsamokovarov:
While calls like `foo&.[] bar` do compile, I won't blame people if they
want to leave the parentheses so they know what's going on, even if
they chose to use `EnforcedStyle: omit_parentheses`.

I'm proposing we legitimize the parens in `foo&.[](bar)` and any other
safe navigation calls. I don't think this deserves extra configuration
either.

What do you folks think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


bbatsov:
It seems to me this has little to do with safe navigation and more to do with operator method names, as doing `something.[] bar` is pretty much the same.

gsamokovarov:
This makes sense, will open another PR with a change that allows parens for operator methods.

gsamokovarov:
👉 https://github.com/rubocop/rubocop/pull/9620

PR #9615, opened 2021-03-18
Add new `Style/StringChars` cop

koic:
## Summary

This PR adds new `Style/StringChars` cop. The cop checks for uses of `String#split` with empty string or regexp literal argument.

```ruby
# bad
string.split(//)
string.split('')

# good
string.chars
```

Since Ruby 2.0, the behavior of `String#chars` and `String#split(//)` is the same.

### Ruby 1.9 and lower

```console
% ruby -ve "p 'foo'.split(//)"
ruby 1.9.3p551 (2014-11-13 revision 48407) [x86_64-darwin13.0.2]
["f", "o", "o"]
% ruby -ve "p 'foo'.chars"
ruby 1.9.3p551 (2014-11-13 revision 48407) [x86_64-darwin13.0.2]
#<Enumerator: "foo":chars>
```

### Ruby 2.0 and higher

```console
% ruby -ve "p 'foo'.split(//)"
ruby 2.0.0p648 (2015-12-16 revision 53162) [x86_64-darwin13.0.2]
["f", "o", "o"]
% ruby -ve "p 'foo'.chars"
ruby 2.0.0p648 (2015-12-16 revision 53162) [x86_64-darwin13.0.2]
["f", "o", "o"]
```

Writing `String#chars` would pretty express the intent of the code.

## Additional Information

This cop is intended to use `String#chars` method that express readable code.
And it has a different purpose than `Performance/RedundantSplitRegexpArgument` cop.
https://docs.rubocop.org/rubocop-performance/cops_performance.html#performanceredundantsplitregexpargument

Therefore, it will be added as a new `Style` cop.

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -4493,6 +4493,11 @@ Style/StderrPuts:
   Enabled: true
   VersionAdded: '0.51'
 
+Style/StringChars:
+  Description: 'checks for uses of `String#split` with empty string or regexp literal argument.'
```
bbatsov, line 4497: I guess you should mention that the cops is not safe and add some notes about the safety.

bbatsov: Actually, probably it's quite unlikely that some other class would define a split method that takes exactly the same args. :-)

bbatsov (approved):

bbatsov:
Great cop - simple and useful at the same time!

config/default.yml
koic: Thank you for the review. I marked the cop as unsafe and added a description to safety.
Initially, I didn't mark it as unsafe because I think it's less likely to be false positives 😅

PR #9620, opened 2021-03-19
Allow parentheses in operator method calls for `Style/MethodCallWithArgsParentheses`

gsamokovarov:
While calls like `data&.[] value`, `value.=== other` do compile, I won't
blame people if they want to leave the parentheses so they know what's
going on.

I'm proposing we legitimize the parens when calling operator methods.  I
don't think this deserves extra configuration.


bbatsov:
The change is fine, but it should be reflected in the cop documentation via some examples, otherwise I don't think the users will figure out how this works. And you have to move that changelog entry. :-)

bbatsov:
Thanks!

PR #9623, opened 2021-03-19
[Fix #9621] Add `consistent` style to `Lint/SymbolConversion` to require all symbol keys in a hash to use the same convention

### LINKED ISSUE #9621: Lint/SymbolConversion to provide an additional "consistent" option

For hashes where keys are mixed, wouldn't having a consistent key structure be more readable than be forced to use symbols for some keys and strings for the rest - 
```ruby
def method
  {
    'a': 1,
    'b-a': 2
  }
end
```
raises a `Lint/SymbolConversion` violation.

```ruby
# .rubocop.yml

Lint/SymbolConversion:
  EnforcedStyle: consistent
```
No violations for:
```ruby
def method
  {
    'a': 1,
    'b-a': 2
  }
end
```
(OR)
```ruby
def method
  {
    a: 1,
    b: 2
  }
end
```
violates for :
```ruby
def method
  {
    a: 1,
    'b-a': 2
  }
end
```

dvandersluis:
The `consistent` style works the same as the previous default style (now called `strict`), in that symbol conversions should be explicit. However, in the case of a hash with a mix of symbol keys requiring quoting and not, this style now enforces that all keys are quoted for consistency.

Fixes #9621.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/symbol_conversion.rb
+++ b/lib/rubocop/cop/lint/symbol_conversion.rb
@@ -6,6 +6,12 @@ module RuboCop
       # This cop checks for uses of literal strings converted to
       # a symbol where a literal symbol could be used instead.
       #
+      # There are two possible styles for this cop.
+      # `strict` (default) will register an offense for any incorrect usage.
+      # `consistent` additionally requires hashes to use the same style for
+      # every symbol key (ie. if any symbol key needs to be quoted it requires
+      # all keys to be quoted).
+      #
       # @example
       #   # bad
       #   'string'.to_sym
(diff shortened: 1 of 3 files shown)


dvandersluis:
@rubocop/rubocop-core one thing that I'm not sure about is that `Lint` cops generally do not have configurable styles; `Lint/InheritException` is the only other one. However, this behaviour can't be enforced by a separate `Style` cop unless `Lint/SymbolConversion` is updated to ignore some cases if that cop is enabled, which is another can of worms. Maybe someone has an idea?

Separately, I am using the configuration from `Style/StringLiterals` to determine how to quote the keys. We don't currently have anything checking for consistency of quote style in quoted symbol literals.

bbatsov:
> @rubocop/rubocop-core one thing that I'm not sure about is that Lint cops generally do not have configurable styles; Lint/InheritException is the only other one. However, this behaviour can't be enforced by a separate Style cop unless Lint/SymbolConversion is updated to ignore some cases if that cop is enabled, which is another can of worms. Maybe someone has an idea?

Most Lint cops shouldn't have different supported styles, but I guess that's OK in the current situation. Alternatively this can also be some flag like "AllowConsistentSymbolsInHash" or something along those lines. I'm fine with the current approach as well.

> Separately, I am using the configuration from Style/StringLiterals to determine how to quote the keys. We don't currently have anything checking for consistency of quote style in quoted symbol literals.

I think that's fine.

lib/rubocop/cop/lint/symbol_conversion.rb
```diff
@@ -21,10 +27,26 @@ module Lint
       #   :underscored_symbol
       #   :'hyphenated-string'
       #
+      # @example EnforcedStyle: consistent
```
bbatsov, line 30: Might be a good idea to add the `strict` case for hashes as well, so it's clearer what this means for them.

dvandersluis: @bbatsov I added additional documentation, thanks!

lib/rubocop/cop/lint/symbol_conversion.rb
```diff
@@ -21,10 +27,49 @@ module Lint
       #   :underscored_symbol
       #   :'hyphenated-string'
       #
+      # @example EnforcedStyle: strict
```
koic, line 30: Can you mark `(default)` to the EnforcedStyle?

```suggestion
      # @example EnforcedStyle: strict (default)
```

dvandersluis: Yep! Sorry, I forgot to.

dvandersluis: Fixed.

bbatsov:
Thanks!

PR #9624, opened 2021-03-20
[Fix #9325] Make Exclude config parameters merge by default

### LINKED ISSUE #9325: RuboCop does not merge Exclude by default and therefore scans vendor/bundle
RuboCop currently does not merge `Exclude` by default.
I think this is a significant usability bug, because it makes RuboCop harder to use in combination with the very common Bundler path `vendor/bundle`.

Notably, it causes extra troubles for ruby/setup-ruby in https://github.com/ruby/setup-ruby/issues/136 and RuboCop seems by far the most common tool to be tripped up by files under `vendor/`. Other tools seem to deal with it just fine by default, or be much less frequently used.

So that even though RuboCop does exclude `vendor/**/*` [in the default config](https://github.com/rubocop-hq/rubocop/blob/e12d1eadea402f955aa9be16ff6bd7def6a970d8/config/default.yml#L66),
that has actually no effect at all (!) when using a `.rubocop.yml` such as:
```yml
AllCops:
  Exclude:
    - 'gemfiles/**/*'
```

--------

## Expected behavior

RuboCop should by default ignore `vendor/**/*`, even if the config defines extra `AllCops:` -> `Exclude:`.

## Actual behavior

An extra `AllCops:` -> `Exclude:` overrides the default, which I believe is always unintentional.
This results in confusing errors, for instance when seeing an old `TargetRubyVersion` from a gem under `vendor/bundle`, such as:
```
$ bundle exec rubocop
Error: Unsupported Ruby version 2.1 found in `TargetRubyVersion` parameter (in vendor/bundle/ruby/2.5.0/gems/rainbow-3.0.0/.rubocop.yml). 2.1-compatible analysis was dropped after version 0.58.
Supported versions: 2.2, 2.3, 2.4, 2.5, 2.6
```
If it does not error, it would also slow down RuboCop significantly by scanning all these extra files.

This problem will also cause `.git`, `node_modules` and `tmp` to be scanned, while they should not.

## Workaround

```yml
inherit_mode:
  merge:
    - Exclude
```
(https://docs.rubocop.org/rubocop/0.88/configuration.html#merging-arrays-using-inherit_mode)

I don't think it's a good workaround. RuboCop should work fine with the most used Bundler path without any extra manual step.
It is confusing that the default `Exclude` paths are completely forgotten whenever adding a custom `Exclude`.

Related: https://github.com/rubocop-hq/rubocop/issues/288#issuecomment-664524299 #7819 and I guess other issues, this is counter-intuitive and confusing behavior.

jonas054:
First of all, I'm not convinced that this is a good idea. Let's discuss.

It turned out to be trickier to implement than I had anticipated. There was code here and there that depended on the inherit mode being `override` for `Exclude`. The changes in pretty central parts of the logic makes me worry that we're opening up a can of worms. More issues will be reported down the line. That sort of thing.

And also, the complexities of configuration inheritance are getting worse with this change IMO. It's a bit weird that `Exclude`s are merged while `Include`s are overridden. I experimented with letting `Include`s be merged as well, but that's a much bigger change since default configuration uses `Include` to let specific cops only inspect one kind of file.

So that's the case against. :slightly_smiling_face: Here's the description:

Add the `inherit_mode` setting in default configuration to make `Exclude` parameters in user configuration merge with the same `Exclude` settings in inherited configuration. The result of `rubocop --auto-gen-config` is affected by this change, since we otherwise copy values from inherited `Exclude` settings into the generated file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* [#9325](https://github.com/rubocop/rubocop/issues/9325): Make the `Exclude` configuration parameter merge by default. ([@jonas054][])
+
 ## 1.11.0 (2021-03-01)
 
 ### New features
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,5 +1,9 @@
 # Common configuration.
 
+inherit_mode:
+  merge:
+    - Exclude
+
 AllCops:
   RubyInterpreters:
     - ruby
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -89,10 +89,11 @@ Other types, such as `AllCops` / `Include` (an array), are overridden by the
 child setting.
 
 Arrays override because if they were merged, there would be no way to
-remove elements in child files.
+remove elements in child files. An exception to this rule is `Exclude`, which
+is an array but is still merged because that is more convenient for most users.
 
-However, advanced users can still merge arrays using the `inherit_mode` setting.
-See "Merging arrays using inherit_mode" below.
+Whether individual array parameters are overridden or merged is controlled
+through the `inherit_mode` setting. See "Merging arrays using inherit_mode" below.
 
 === Inheriting from another configuration file in the project
 
@@ -193,6 +194,26 @@ value.
 This applies to explicit inheritance using `inherit_from` as well as implicit
 inheritance from https://github.com/rubocop/rubocop/blob/master/config/default.yml[the default configuration].
 
+The setting for `inherit_mode` in the default configuration is:
+
+[source,yaml]
+----
+inherit_mode:
+  merge:
+    - Exclude
+----
+
+This makes `Exclude` a special case, being merged by default even though it is
+an array. If this behavior is not desired, user configuration should have the
+following setting:
+
+[source,yaml]
+----
+inherit_mode:
+  override:
+    - Exclude
+----
+
 Given the following config:
 
 [source,yaml]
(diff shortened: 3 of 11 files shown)


jonas054:
I don't understand the build failure for Ruby 2.4, and I can't reproduce it locally.

jonas054:
By the way, this is a change to default configuration, which means it requires bumping the major version of RuboCop. Right?

bbatsov:
Yes. That's why we have to weigh the pros and cons of this very carefully.

eregon:
Looks good to me.

I think almost nobody ever wants to scan files under:
```yaml
  Exclude:
    - 'node_modules/**/*'
    - 'tmp/**/*'
    - 'vendor/**/*'
    - '.git/**/*'
```

And it can still be done with
```yaml
inherit_mode:
  override:
    - Exclude
```
if desired (as documented).

So I'm not sure as a RuboCop user that requiring a new major version for this is really needed.
OTOH, if releasing RuboCop 2.x would be something reasonably soon then I wouldn't really mind.

Another way to do this which would avoid changing the default `inherit_mode` for `Exclude` would be to have something like `DefaultExclude:` in addition to `Exclude:` and then merge both of these arrays.
Then it would solve the issue, and for people wanting no default excludes, it'd be quite clear:
```yaml
AllCops:
  DefaultExclude: []
```

bbatsov:
> So I'm not sure as a RuboCop user that requiring a new major version for this is really needed.

I'm afraid that anything that will break RuboCop's behaviour for some users requires a major release. There are no plans for RuboCop 2.0 at this point, but it's bound to happen at some point.

jonas054:
Am I using the milestone feature correctly? I set 2.0 as milestone for this one, but does that mean that the PR should be merged exactly when we release 2.0 -- not before?

bbatsov:
@jonas054 Yeah, you're using it correctly.

pirj:
It would allow removing those tweaks from extensions: [`rubocop-rspec`](https://github.com/rubocop/rubocop-rspec/blob/cf272285131d2296c91b20f3ae2ce239cf4c0c7e/config/default.yml#L115), [`rubocop-rails`](https://github.com/rubocop/rubocop-rails/blob/3a937065ca8e63e97c5cbb408cafd6ab3d4e0db2/config/default.yml#L3).

pirj:
> I experimented with letting Includes be merged as well, but that's a much bigger change since default configuration uses Include to let specific cops only inspect one kind of file.

It may seem that `Include` is semantically different. Where `Exclude` means "don't check those", and it makes sense to accumulate definitions from different places (default RuboCop config, gems, project root config, project sub-directory config). `Include` feels different, and is supposed to mean "**only** check **those**", e.g.
```yaml
Include:
 - '**/Gemfile'
```

After checking the `Include` statements from `config/default.yml` of `rubocop`, `rubocop-rspec` and `rubocop-rails`, I couldn't come up with an example where users would suffer from `Include` merging by default. It was a revelation that it might make sense to add those as well, e.g.:
```
Rails/AddColumnIndex:
  Include:
    - db/migrate/*.rb
```
E.g. on my current project we have multiple databases, and migrations for the second one are in `db/misc-migrate`.
Having `Include` in override mode would skip the cop for `db/migrate`.

> much bigger change

Code-wise, or impact-wise bigger?

jonas054:
> Code-wise, or impact-wise bigger?

I was thinking _impact_ when I wrote that. My intuition says that changing the meaning from _include only_ to _include also_ would be big, but I haven't conducted any experiments to prove it.

Aqualon:
Another aspect on this situation is using `rubocop --auto-gen-config` as suggested in https://docs.rubocop.org/rubocop/configuration.html#automatically-generated-configuration. If you use it, you may get `Exclude:` set for certain cops, disabling the default excludes of those cops with the default inherit_mode, resulting in the end in rubocop failing even so you just tried to silence all complaints via the `.rubocop_todo.yml`.

Manually adding
```
inherit_mode:
  merge:
    - Exclude
```
fixes this, but it feels confusing this happens.

PR #9625, opened 2021-03-20
Allow parentheses in yield arguments for `Style/MethodCallWithArgsParentheses`

gsamokovarov:
When using the `omit_parentheses` style, we don't recognize `yield`
argument calls and require parentheses omission. If we yield multiple
values, we cannot omit the parentheses in method calls as this won't be
valid Ruby. We currently register offenses for:

```ruby
# Example 1
yield File.basename(path)
      ^^^^^^^^^^^^^^^^^^^ Omit parentheses for method calls with arguments.

# Correction...
yield File.basename path # 👈 This compiles.

# Example 2
yield path, File.basename(path)
            ^^^^^^^^^^^^^^^^^^^ Omit parentheses for method calls with arguments.

# Correction...
yield path, File.basename path  # 👈 This DOES NOT!!!
```

If we treat yield arguments as any other method call arguments, we'll
allow the parentheses and let people write valid Ruby in the cases
above.


bbatsov:
The suggested change looks good, just move the changelog entry to the `changelog` folder.

gsamokovarov:
Thanks, moved the changelog.

bbatsov:
Thanks!

PR #9629, opened 2021-03-22
Calls with braces in interpolation for `Style/MethodCallWithArgsParentheses`

gsamokovarov:
If we are calling methods in string interpolation with the `omit_parentheses`
style for the `Style/MethodCallWithArgsParentheses` cop, they are
currently forced to be without parentheses:

```ruby
"#{part 1}, #{part 2}, #{part 3}"
```

My coworkers at Dext want to be able to put spaces while calling methods
in interpolation, and I don't blame them. Having spaces reduces the
noise of following a method call in a string, especially if the string
happens to have multiple interpolations in it.

I want to add the `AllowParenthesesInStringInterpolation` cop configuration
that allows the use of parentheses in string interpolation calls.

```ruby
"#{part(1)}, #{part(2)}, #{part(3)}"
```

### DIFF
--- a/changelog/new_add_allow_parentheses_in_string_interpolation_config_to_style_method_call_with_args_parentheses.md
+++ b/changelog/new_add_allow_parentheses_in_string_interpolation_config_to_style_method_call_with_args_parentheses.md
@@ -1,0 +1,1 @@
+* [#9629](https://github.com/rubocop/rubocop/issues/9629): Add `AllowParenthesesInStringInterpolation` configuration to `Style/MethodCallWithArgsParentheses` to allow parenthesized calls in string interpolation. ([@gsamokovarov][])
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#9629](https://github.com/rubocop/rubocop/issues/9629): Add `AllowParenthesesInStringInterpolation` configuration to `Style/MethodCallWithArgsParentheses` to allow parenthesized calls in string interpolation. ([@gsamokovarov][])
```
bbatsov, line 7: The changelog entries should now go to the `changelog` folder. (this file is auto-generated from them on release)

gsamokovarov: Thanks, moved the entry to its own file. Will do so to my other PRs as well.

gsamokovarov:
I have moved the changelog and rewrote the documentation examples.

PR #9632, opened 2021-03-23
[Fix #5953] Fix a false positive for `Style/AccessModifierDeclarations`

### LINKED ISSUE #5953: Style/AccessModifierDeclarations false positives
I just upgraded to the latest rubocop and received some peculiar offenses from Style/AccessModifierDeclarations offenses. I use a number of different tools that generate methods automatically, including [`anima`](https://github.com/mbj/anima), for example. But sometimes I apply a different access modifier to the generated method(s). Style/AccessModifierDeclarations flags this as "inlined", when it is not anything I would consider inline.

--------

## Expected behavior

I expect the following example (and various similar cases) not to raise an offense with the default rubocop configuration.

```ruby
class Foo
  include Anima.new(:foo) # Defines constructor which takes the single keyword argument :foo and generates a public reader.

  private :foo
end
```

If, for some reason, this is the intended behavior and I'm missing something here, I think the error message should be made clearer. I think this cop should ignore access modifier adjustments to methods it can't statically detect.

## Actual behavior

```bash
test.rb:4:3: C: Style/AccessModifierDeclarations: private should not be inlined in method definitions.
  private :foo
  ^^^^^^^
```

Run rubocop with Style/AccessModifierDeclarations's style set to "group" (the default).

## RuboCop version

```bash
$ rubocop -V
0.57.0 (using Parser 2.5.1.0, running on ruby 2.5.1 x86_64-darwin17)
```

koic:
Fixes #5953.

This PR fixes a false positive for `Style/AccessModifierDeclarations` when using `module_function` with symbol.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_style_access_modifier_declarations.md
+++ b/changelog/fix_a_false_positive_for_style_access_modifier_declarations.md
@@ -1,0 +1,1 @@
+* [#5953](https://github.com/rubocop/rubocop/issues/5953): Fix a false positive for `Style/AccessModifierDeclarations` when using `module_function` with symbol. ([@koic][])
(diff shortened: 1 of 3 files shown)


pboling (approved):
Thank you! 🚀 LGTM

Drenmi (approved):

PR #9633, opened 2021-03-23
Fix an incorrect auto-correct for `Lint/NumberConversion`

koic:
For example, the following auto-correction test code is invalid syntax:

```ruby
"1,2,3,foo,5,6,7,8".split(',').map({ |i| Integer(i, 10) })
```

This PR removes invalid parentheses and auto-corrects it to valid syntax:

```ruby
"1,2,3,foo,5,6,7,8".split(',').map { |i| Integer(i, 10) }
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/number_conversion.rb
+++ b/lib/rubocop/cop/lint/number_conversion.rb
@@ -105,6 +105,8 @@ module RuboCop
               corrected_method: correct_sym_method(to_method)
             )
             add_offense(node, message: message) do |corrector|
+              remove_parentheses(corrector, node) if node.parenthesized?
+
               corrector.replace(sym_node, correct_sym_method(to_method))
             end
           end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/number_conversion.rb
```diff
@@ -120,6 +122,11 @@ def correct_sym_method(to_method)
           "{ |i| #{body} }"
         end
 
+        def remove_parentheses(corrector, node)
+          corrector.replace(node.loc.begin, ' ')
+          corrector.remove(node.loc.end)
+        end
```
dvandersluis, line 128: We have two other `remove_parentheses` methods defined on various cops (although they work slightly differently in each case) maybe it'd be good to normalize that at some point.

dvandersluis (approved):
👍 looks good!

lib/rubocop/cop/lint/number_conversion.rb
koic: Yeah. I didn't normalize here because each one has different behavior, but I'll look at it separately :-)

PR #9634, opened 2021-03-23
[Fix #9622] Fixed `Style/BisectedAttrAccessor` autocorrection to handle multiple bisected attrs in the same macro

### LINKED ISSUE #9622: Style/BisectedAttrAccessor AutoCorrect fails when multiple variables are defined
When using the Style/BisectedAttrAccessor Auto Correct function when multiple variables are defined, it only corrects the first variable, and deletes the other variables

--------

## Expected behavior

      # Bad
       attr_writer :name, :limit, :size
       attr_reader :name, :limit, :size
     
     # Good
       attr_accessor :name, :limit, :size

## Actual behavior

      # Bad
       attr_writer :name, :limit, :size
       attr_reader :name, :limit, :size
     
      # Good
       attr_accessor :name

Notice the lack of :limit and :size

Run rubocop against the  `#Bad` code

## RuboCop version

```
1.9.1 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 2.7.2 x86_64-darwin19)
  - rubocop-rails 2.9.1
  - rubocop-rspec 2.2.0
 ```

dvandersluis:
This ended up being a pretty heavy refactor of the `Style/BisectedAttrAccessor` cop.

Previously we were doing multiple passes at determining what macros are relevant for the cop, and which names within them. Despite that, if a macro had multiple names and was rewritten, we'd lose context of what other names there were and only the first one would survive the autocorrection.

Instead, I have now extracted a `Macro` value object that encapsulates all the data for the macro, including visibility (since macros of different visibilities are not combined), bisected and non-bisected attribute names, etc. When investigating a class, any macros found have objects created, and they are retained to be used to correct the class as a whole after wards (but we still register offenses on individual macro attributes as before).

I have also added some more edge case tests that I encountered as I was working on this.

Fixes #9622.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_fixed_stylebisectedattraccessor.md
+++ b/changelog/fix_fixed_stylebisectedattraccessor.md
@@ -1,0 +1,1 @@
+* [#9622](https://github.com/rubocop/rubocop/issues/9622): Fixed `Style/BisectedAttrAccessor` autocorrection to handle multiple bisected attrs in the same macro. ([@dvandersluis][])
--- a/lib/rubocop/cop/style/bisected_attr_accessor.rb
+++ b/lib/rubocop/cop/style/bisected_attr_accessor.rb
@@ -19,47 +19,61 @@ module RuboCop
       #   end
       #
       class BisectedAttrAccessor < Base
-        include VisibilityHelp
+        require_relative 'bisected_attr_accessor/macro'
+
+        include RangeHelp
         extend AutoCorrector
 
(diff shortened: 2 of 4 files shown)


koic:
Great work! Can you add a changelog entry?

dvandersluis:
@koic 🤦🏻‍♂️ I knew I forgot something. Thanks!

koic (approved):

bbatsov:
Great work! 🏅

PR #9637, opened 2021-03-25
Parenthesized forwarded args in `Style/MethodCallWithArgsParentheses` `omit_parentheses`

gsamokovarov:
The new-style arguments delegation included in Ruby 2.7 and expanded in
Ruby 3.0 is required to be parenthesized, since the syntax is clashing with the
endless `start...` range, which can be started with `start ...` as well. Currently,
the most iffy cop of them all is issuing offenses for:

```ruby
def delegated_call(...)
  @proxy.call(...)
             ^^^^^ Omit parentheses for method calls with arguments.
end
```

Removing the parens will  start an endless range with the beginning being the
result of the method call without arguments. See the parse trees below for more
details:

```bash
-> ruby-parse -e "def foo(...); bar(...); end"
(def :foo
  (args
    (forward-arg))
  (send nil :bar
    (forwarded-args)))

-> ruby-parse -e "def foo(...); bar ...; end"
(def :foo
  (args
    (forward-arg))
  (erange
    (send nil :bar) nil))

-> ruby-parse -e "def foo(...); bar...; end"
(def :foo
  (args
    (forward-arg))
  (erange
    (send nil :bar) nil))
```

We don't want that, so allow the parentheses and leave this gotcha to
the poor developer that's gonna run onto it.

### DIFF
--- a/changelog/fix_parentheses_in_forwarded_args_for_style_method_call_with_args_parentheses.md
+++ b/changelog/fix_parentheses_in_forwarded_args_for_style_method_call_with_args_parentheses.md
@@ -1,0 +1,1 @@
+* [#9637](https://github.com/rubocop/rubocop/issues/9637): Allow parentheses for forwarded args in `Style/MethodCallWithArgsParentheses`'s `omit_parentheses` style to avoid endless range ambiguity. ([@gsamokovarov][])
--- a/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
+++ b/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
@@ -40,8 +40,12 @@ module RuboCop
       #     to `true` allows the presence of parentheses in such a method call
       #     even with arguments.
       #
-      # NOTE: Parens are required around a method with arguments when inside an
-      # endless method definition (>= Ruby 3.0).
+      # NOTE: Parentheses are still allowed in cases where omitting them
+      # results in ambiguous or syntactically incorrect code. For example,
+      # parentheses are required around a method with arguments when inside an
+      # endless method definition introduced in Ruby 3.0.  Parentheses are also
+      # allowed when forwarding arguments with the triple-dot syntax introduced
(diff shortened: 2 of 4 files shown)


changelog/fix_parentheses_in_forwarded_args_for_style_method_call_with_args_parentheses.md
```diff
@@ -0,0 +1 @@
+* [#9627](https://github.com/rubocop/rubocop/issues/9627): Allow parentheses for forwarded args in `Style/MethodCallWithArgsParentheses`'s `omit_parentheses` style to avoid endless range ambiguity. ([@gsamokovarov][])
```
koic, line 1: This looks like a typo :-)

```suggestion
* [#9637](https://github.com/rubocop/rubocop/pull/9637): Allow parentheses for forwarded args in `Style/MethodCallWithArgsParentheses`'s `omit_parentheses` style to avoid endless range ambiguity. ([@gsamokovarov][])
```

gsamokovarov: So I didn't predict it after all 😅. Thanks for noticing!

bbatsov:
> The new-style arguments delegation included in Ruby 2.7 and expanded in
Ruby 3.0 is required to be parenthesized, since the syntax is clashing with the
endless start... range, which can be started with start ... as well.

I'd add something along those lines to the cop's documentation. 

> the most iffy cop of them all

Well said! This has to be the cop with the most special cases and bugs in the history of RuboCop! 😆

gsamokovarov:
But boy, do we love this style in Dext! Maybe there is a better approach to tackle the parentheses omission as in this form we have to enumerate the cases where omitting them leads to ambiguous or invalid code. 😅

I have updated the docs to hint that there are other cases we allow parentheses even if we are enforcing the `omit_parentheses` style.

bbatsov:
Thanks!

PR #9639, opened 2021-03-25
Fix `Style/RedundantBegin` removing comments on assignment statement correction

marcotc:
`Style/RedundantBegin` cop currently removes comments that occur
before the first child node of the `begin` block. This happens
because comments are skipped when navigating the node tree.

Currently,
```ruby
var ||= begin
  # comment
  foo
end
```
will correct to
```ruby
var ||= foo
```

This PR changes that correction to
```ruby
# comment
var ||= foo
```

Other comments in the vicinity (before and after `begin/end` block,
and after first child node) are still preserved, as they
already were before this change.

-----------------

* [x] Added tests.


koic:
Thanks!

PR #9647, opened 2021-03-30
Auto disable parallel when combined with --auto-correct, --auto-gen-config, or --fail-fast

rrosenblum:
This came up in a conversation in https://github.com/rubocop/rubocop/pull/9640. We probably should have made it work this way from the start instead of raising an exception.

It felt a big weird to handle this inside of a "validation". Please let me know if there is a better place for the code or if there is better error message.

I left in the code to delete the parallel flag from the rake task to avoid printing the warning about parallel being ignored.

### DIFF
--- a/changelog/change_auto_disable_parallel.md
+++ b/changelog/change_auto_disable_parallel.md
@@ -1,0 +1,1 @@
+* [#9647](https://github.com/rubocop/rubocop/pull/9647): The parallel flag will now be automatically ignored when combined with `--auto-correct`, `--auto-gen-config`, or `-F/--fail-fast`. Previously, an error was raised and execution stopped. ([@rrosenblum][])
--- a/lib/rubocop/options.rb
+++ b/lib/rubocop/options.rb
@@ -343,21 +343,24 @@ module RuboCop
                                    'false is not allowed.'
       end
 
-      validate_parallel_with_combo_option
+      disable_parallel_when_invalid_combo
     end
 
-    def validate_parallel_with_combo_option
+    def disable_parallel_when_invalid_combo
       combos = {
-        auto_gen_config: '-P/--parallel uses caching to speed up execution, ' \
-                         'while --auto-gen-config needs a non-cached run, ' \
-                         'so they cannot be combined.',
-        fail_fast: '-P/--parallel cannot be combined with -F/--fail-fast.',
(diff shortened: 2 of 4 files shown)


changelog/change_auto_disable_parallel.md
```diff
@@ -0,0 +1 @@
+* The parallel flag will now be automatically ignored when combined with `--auto-correct`, `--auto-gen-config`, or `-F/--fail-fast`. Previously, an error was raised and execution stopped. ([@rrosenblum][])
```
dvandersluis, line 1: You should add a link to this PR at the start of the message.

lib/rubocop/options.rb
```diff
@@ -362,16 +362,19 @@ def validate_parallel
 
     def validate_parallel_with_combo_option
```
dvandersluis, line 363: Maybe we should rename this method now?

dvandersluis (approved):
@rrosenblum thanks! I was going to get to this but hadn't gotten a chance yet. This is a great implementation.

One note: maybe your changed tests should expect that there is no parallel key in options after parsing?

lib/rubocop/options.rb
rrosenblum: I had the same thought. I couldn't think of great name for what this is doing. I think this is the first time we're disabling incompatible configs so maybe a new method entry point? Maybe something like `fix_invalid_parallel_combos` or `disable_invalid_parallel_option`?

rrosenblum:
> One note: maybe your changed tests should expect that there is no parallel key in options after parsing?

@dvandersluis I like that idea, any thoughts on how to implement the check? `options` doesn't appear to expose the underlying data structure for easy checking. Best I can think of is `expect(options.instance_variable_get('@options').keys).to_not include(:parallel)`

dvandersluis:
Sorry Ryan, I had intended to give you an answer to your question and totally forgot. We could conceivably add a delegate to `keys` or `key?` (if that exists) inside the `Option` class if you want, but if you don't want to change that going through the ivar seems to be the only way.

rrosenblum:
It's cool. Unless you have strong feeling about it, I'm inclined to use `instance_variable_get` in the tests for now. I rather not expose the internal structure of options just to help with this 1 test. If there are some other use cases, then it could be worth while.

bbatsov:
Thanks!

PR #9648, opened 2021-03-31
Drop Ruby 2.4 support

koic:
This PR drops Ruby 2.4 support. The next release will be the time to drop Ruby 2.4 support.

- https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/
- https://docs.rubocop.org/rubocop/1.12/compatibility.html#support-matrix

It includes two workaround separated commits for JRuby.

-----------------

* [x] Added tests.


marcandre:
Cool. Could you update `.mergify.yml` also please?

koic:
Ah, I get it! Thank you for your review!

tas50:
Other than the general code cleanup is there a driver for dropping Ruby 2.4 support?

marcandre:
It's been the policy to drop a version of Ruby every year.

Pros are:
* cleanup
* we can use more recent features (without needing `backports` gem 😅 )
* encourages the community to move forward

Cons are:
* leaves out the part of the community that has difficulty moving forward

bbatsov:
There are more details here https://docs.rubocop.org/rubocop/1.12/compatibility.html#support-matrix

Admittedly there's no rush to drop support for Ruby 2.4, but knowing that it has been unsupported for 1 year now and very few people are stuck on it (upgrading to 2.5 is trivial) I don't see much of an issue in dropping it. That being said - one can argue that's also a breaking change now...

marcandre:
> one can argue that's also a breaking change now...

If you are referring to semantic versioning, I don't feel it is. Anytime we bump the requirement for any dependency could be seen as a breaking change...

bbatsov:
You've convinced me! 😆 I agree with this line of reasoning, but I wonder if all of users subscribe to it. Again - I don't expect much impact from this change, as we're talking about an unsupported version of Ruby. The transition from 1.8 to 1.9 was painful because it required significant code changes in certain cases, but even since it the updates have been so smooth that I doubt many people are stuck on a particular version of Ruby.

bbatsov (approved):

marcandre:
The main reason one could be stuck on a particular version of Ruby is because they use a particular version of Rails... Rails is quite a bit more difficult to upgrade than Ruby.

Drenmi:
Friends don't let friends stay on ancient versions of Ruby.

It ultimately ends up being a disservice to both the individual and the community, and committing to backwards compatibility wastes time in the past that could be spent getting us into the future. 🙂

bbatsov:
One additional data point:

image

Twitter wouldn't allow me to add more than 4 answers, so it's hard to say how many people are on Ruby 2.4, but probably half of those 10%.

lib/rubocop/comment_config.rb
```diff
@@ -32,9 +32,7 @@ def extra_enabled_comments
     end
 
     def comment_only_line?(line_number)
-      non_comment_token_line_numbers.none? do |non_comment_line_number|
-        non_comment_line_number == line_number
-      end
+      non_comment_token_line_numbers.none?(line_number)
```
bbatsov, line 35: Do we have a cop for this and `any?` already?

bbatsov:
I'll merge this later next week unless someone violently objects.

lib/rubocop/comment_config.rb
koic: Yes. RuboCop Performance has `Performance/RedundantEqualityComparisonBlock` cop introduced by https://github.com/rubocop/rubocop-performance/pull/213.

https://docs.rubocop.org/rubocop-performance/1.10/cops_performance.html#performanceredundantequalitycomparisonblock

tagliala:
> That being said - one can argue that's also a breaking change now...

Hi, I think this may be considered a breaking change in the scenario I'm experiencing here: Travis CI builds failing for all versions of Ruby except 2.4: https://travis-ci.org/github/DavyJonesLocker/client_side_validations/builds/767962700

Starting from version 1.0, I've locked the RuboCop development dependency to the minor version number, assuming it is safe to upgrade as per SemVer.

Now I could stop supporting Ruby 2.4 on the other gem, but having a major bump in RuboCop would have prevented the build to fail when trying to upgrade an unrelated dependency (`puppeteer-core-9.0.0`, which is also a major btw, and I thought that it was responsible of the breaking change)

marcandre:
There's a big difference between a runtime dependency and a dev/test-only dependency. Which version you decide to support is not quite related to `rubocop`.

My recommendations:

1) Remove the `add_development_dependency` in your gemspec and move them in your Gemfile, possibly conditional on the Ruby version

2) Have the rubocop run be done only once, for whichever version of Ruby you want, as an independent step in your CI. This has the added benefit of making it clear if a failed CI is due to styling or actual test failure. It's also more efficient. We have this setup for the `rubocop-ast` gem.

PR #9678, opened 2021-04-05
Fix #7977 Add redundant line break cop

### LINKED ISSUE #7977: Force code on one line if it is possible

I would like a cop that forces ruby code as much as possible on 1 line. I am integrating RuboCop way more strictly in a long-living project. It's going pretty well but the autocorrection sometimes adds more lines than needed IMO. To go even further I would like it if I can enforce code to be on 1 line as much as possible (until the max linewidth is reached). What happens now, for instance, is this:
```ruby
method_call(other_method_call(foo: "this is a verry loooooooong string", bar: 1234), foobar: 123)
```
is corrected to:
```ruby
method_call(
  other_method_call(
    foo: "this is a verry loooooooong string",
    bar: 1234
  ),
  foobar: 123
)
```
And I would like:
```ruby
method_call(
  other_method_call(foo: "this is a verry loooooooong string", bar: 1234),
  foobar: 123
)
```
both are accepted with my current config

A cop with this behavior

```ruby
# bad
{
  foo: 1,
  bar: 2
}

#bad
foo(
  123,
  'bar'
)

# good
{ foo: 1, bar: 2 }

# good
foo(123, 'bar')
```

I dug pretty deep in RuboCop docs but could not find any thing. Prettier for ruby does do it like this way but I would like to use 1 tool instead of the combination of a lot.

If possible I would be open to help write this. But I never did anything like linting/formatting so I don't think it will be an easy task

jonas054:
A long time in the making, I finally got around to pushing this change. It might not be everybody's cup of tea, and I've set `Enabled` to `false` for the new cop in anticipation of some push-back. It's enabled for RuboCop sources, though.

The idea is that after we changed the default max line length from 80 to 120, and the max line length within our own code base to 100, there are a lot of lines that are split for a reason that's no longer valid. 

Having a cop that joins lines together might be seen as somewhat heavy-handed. The line breaks might be there for readability purposes. On the other hand, that's very subjective and adhering to a common style might be more important.

The risk for merge conflicts after running auto-correct with this cop in one branch is significant.

On the plus side, this is one more step towards `rubocop -A --only Layout` as a full-featured code formatter.


bbatsov:
@jonas054 Great work!

I'm fine with merging the PR as is it, mostly so you won't have to rebase, but I think it might be nice to make add some config options to the cop - e.g. to ignore blocks or certain methods.

lib/rubocop/cop/layout/redundant_line_break.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Layout
+      # This cop checks whether certain expressions, e.g. method calls, that could fit
+      # completely on a single line, are broken up into multiple lines unnecessarily.
+      #
+      # @example
```
bbatsov, line 9: You should probably metnion some other type of cases here - like blocks, method chains, and `\` separated lines, so it's more obvious what the cop can tackle.

lib/rubocop/cop/layout/redundant_line_break.rb
```diff
@@ -0,0 +1,95 @@
...
+        end
+
+        def suitable_as_single_line?(node)
+          !comment_within?(node) &&
+            node.each_descendant(:if, :case, :kwbegin, :def).none? &&
+            node.each_descendant(:dstr, :str).none?(&:heredoc?) &&
+            node.each_descendant(:begin).none? { |b| b.first_line != b.last_line }
+        end
+
+        def convertible_block?(node)
```
bbatsov, line 59: I wonder if we don't need something like this for convertable `def`s.

exe/rubocop
```diff
@@ -9,9 +9,7 @@ require 'benchmark'
 cli = RuboCop::CLI.new
 result = 0
 
-time = Benchmark.realtime do
-  result = cli.run
-end
+time = Benchmark.realtime { result = cli.run }
```
koic, line 12: I think it's better to accept `do ... end` line break by default, because `do ... end` and `{ ... }` have different meanings.

lib/rubocop/cop/layout/redundant_line_break.rb
jonas054: Good idea.

lib/rubocop/cop/layout/redundant_line_break.rb
jonas054: Possibly in the future, but currently the cop doesn't have any support for joining method definition lines, so the problem doesn't come up. Did I understand your comment correctly?

exe/rubocop
jonas054: Let me check if I understand what you mean. Is it

1. that `do ... end` and `{ ... }` bind differently so changing between them could change the semantics of the code
2. that you might choose one or the other to convey a meaning to the reader, like a value being returned from the block (["Weirich style"](https://github.com/rubocop/ruby-style-guide/issues/162))
3. something else :slightly_smiling_face: 

Becase if it's 1, then it's taken care of by `Style/BlockDelimiters` since `Layout/RedundantLineBreak` only auto-corrects to `time = Benchmark.realtime do result = cli.run end`

bbatsov: I agree it's fine to convert the blocks, unless this changes the meaning of the code due to the different precedence.

bbatsov:
I'm afraid you'll need to rebase on top of `master`.

exe/rubocop
koic: I didn't have enough words 💦  I have two concerns.

1. I'm concerned that it will be an unsafe cop because the operator precedence will change. 
2. It's possible that user intentionally writes it on multi-line instead of single line. IMHO, it feels better to respect readability intended by user by default (I'm not sure about one-liner `{ ... }` may not always be better).

Therefore, it seems safer to configure it to allow multi-line blocks by default.

jonas054: 1. In the `convertible_block?` method I put the condition `send_node.parenthesized? || !send_node.arguments?` to make sure the cop only reports blocks where `do ... end` and `{ ... }` behave the same.
2. I agree that a configuration option for blocks would be nice. It's probably as you say that some people may chose `do .. end` because they think it's more readable. @bbatsov Should I add a configuration option now, or can it wait since the cop is disabled by default? @koic Do you see any changes made by `rubocop -a` in this PR where you think that `do ... end` was better?

bbatsov: I think it's fine to make changes to the cop later. Cops that are disabled by default rarely get used anyways. Likely we'll be the only user for the cop for a while. :-)

marcandre:
I notice some nice changes, in particular strings broken in two now in one line 👍 

> It might not be everybody's cup of tea, and I've set Enabled to false for the new cop in anticipation of some push-back. 

🤚 

> Having a cop that joins lines together might be seen as somewhat heavy-handed. 

Indeed!

I feel that doing this for blocks is quite problematic, especially for `each`:

```ruby
# before
affected.each do |node|
  @classification[node][:visibility] = visibility
end
# after
affected.each { |node| @classification[node][:visibility] = visibility }
```

Looking at the changes in this PR, I have difficulty actually preferring any of the block changes.

> The line breaks might be there for readability purposes. On the other hand, that's very subjective and adhering to a common style might be more important

I would prioritize respecting developer's taste, minimizing developer frustration.

Now that we have "corrected" the maximum line-length change, could we please disable it again in RuboCop?

jonas054:
@marcandre Thanks for the feedback! 

If it's only the changes in block layout that you don't particularly like, wouldn't it be better if I reverse those changes and also change the setting to `InspectBlocks: false` internally?

marcandre:
@jonas054 You are right, it is better to change the setting to `InspectBlocks: false` internally than disabling. This way we can better detect other cases that might be problematic (if any).

As to reverting the changes to the blocks, I feel it might be best too (it seems to me that most are less clear this way), but I haven't checked in depth. I'm fine with whatever you decide in that regards 👍

bbatsov:
It seems to me that just relaxing the config will be enough. I went over the diff again and apart from a few cases of nested blocks I really don't see anything becoming materially harder to read. Yeah, it's undeniable that multi-line blocks stand out a bit more and it's easier to add more expressions to them, but I think it's perfectly fine to be using mostly single-line blocks when they are applicable. Anyways, I don't have a strong preference - I don't have any issues with both types of blocks.

marcandre:
>  I really don't see anything becoming materially harder to read. 

@bbatsov I'm curious, did you see cases that were materially easier to read?

bbatsov:
> I'm curious, did you see cases that were materially easier to read?

Nope. For me it was more or less the same before and after the changes. My point was mostly that having a trivial block on a single line is very common, that's why it seems excessive to revert everything. I'd rather just change instances where we feel the multi-line setup added more value (e.g. nested blocks).

PR #9682, opened 2021-04-07
[Fix #9681] Fix an incorrect auto-correct for `Style/RedundantBegin`

### LINKED ISSUE #9681: Style/RedundantBegin should be marked as unsafe for auto-correction
Style/RedundantBegin cop might auto-correct code incorrectly, leading to a performance issue or even a broken build. See example below:

--------

## Precondition

Given the following example class

```ruby
class Test
  def initialize
    @first_run = false
  end

  def memoized
    @memoized ||= begin
      'memoized' if first_run
    end
  end

  def first_run
    # this might be an expensive operation
    fail 'already run!' if @first_run

    @first_run = true
  end
end
```

and the following test code:
```ruby
test = Test.new
2.times { test.memoized }
```

## Expected behavior

The code should not be auto-corrected, or maybe auto-corrected as the following:

```ruby
class Test
  def initialize
    @first_run = false
  end

  def memoized
    @memoized ||= ('memoized' if first_run)
  end

  def first_run
    # this might be an expensive operation
    fail 'already run!' if @first_run

    @first_run = true
  end
end
```

## Actual behavior

The code is auto-corrected as follows:

```ruby
class Test
  def initialize
    @first_run = false
  end

  def memoized
    @memoized ||= 'memoized' if first_run
  end

  def first_run
    # this might be an expensive operation
    fail 'already run!' if @first_run

    @first_run = true
  end
end
```

which makes the test code raise an exception.

## RuboCop version

1.12.1 (using Parser 3.0.1.0, rubocop-ast 1.4.1, running on ruby 2.6.6 x86_64-linux)

koic:
Fixes #9681.

This PR fixes an incorrect auto-correct for `Style/RedundantBegin` when using modifier `if` single statement in `begin`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_incorrect_autocorrect_for_style_redundant_begin.md
+++ b/changelog/fix_incorrect_autocorrect_for_style_redundant_begin.md
@@ -1,0 +1,1 @@
+* [#9681](https://github.com/rubocop/rubocop/issues/9681): Fix an incorrect auto-correct for `Style/RedundantBegin` when using modifier `if` single statement in `begin` block. ([@koic][])
--- a/lib/rubocop/cop/style/redundant_begin.rb
+++ b/lib/rubocop/cop/style/redundant_begin.rb
@@ -109,7 +109,10 @@ module RuboCop
         def replace_begin_with_statement(corrector, offense_range, node)
           first_child = node.children.first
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

Issue #7486, opened 2019-11-06
Regression of #4198 - Lint/AmbiguousBlockAssociation


luke-hill:
--------

## Expected behavior

`expect { move_direction }.not_to change { current_room_id }` should not raise any offenses
`expect { move_direction }.to change { current_room_id }.to(new_room_id)` should not raise any offenses

## Actual behavior

`expect { move_direction }.not_to change { current_room_id }` raises an offense
`expect { move_direction }.to change { current_room_id }.to(new_room_id)` does not raise an offense

## Steps to reproduce the problem

Code is attached above. I added the working variety to show comparison.

## RuboCop version

Include the output of `rubocop -V` or `bundle exec rubocop -V` if using Bundler. Here's an example:

```
$ [bundle exec] rubocop -V
0.76.0 (using Parser 2.6.5.0, running on ruby 2.6.3 x86_64-linux)
```

mvz:
I'm still seeing this issue in RuboCop 0.83.0.

kreintjes:
Still an issue in 0.91.0.

kreintjes:
@luke-hill @mvz apparently it is desired behaviour, because the syntax is ambiguous. You can fix it as suggested here: https://github.com/rubocop-hq/rubocop/issues/8732#issuecomment-696913077

In some cases it might also be possible to write the matches as `change(object, :attribute)` instead of using a block, as suggested here: https://github.com/rubocop-hq/rubocop/issues/4198#issuecomment-290335794

Or you can disable the cop in this specific case as suggested by bbatsov here: https://github.com/rubocop-hq/rubocop/issues/4198#issuecomment-290332916

luke-hill:
4198 is impossible here. As I don't have a method on my object output, just the object.

The issue is more than not_to is behaving different to to

mvz:
This seems to be the same behaviour as mentioned in https://github.com/rubocop-hq/rubocop/issues/4198#issuecomment-290338661. I'm not sure if that was actually ever fixed?

As an aside, maybe `change(self, :current_room_id)` works?

reedstonefood:
If you're on rubocop 1.13 or later, then you can use `IgnoreMethods` (as mentioned just above this comment) to deal with this scenario. https://github.com/rubocop/rubocop/pull/9685

jimbali:
Can't you just put parantheses around the change method to make it unambiguous?
```
expect { move_direction }.not_to(change { current_room_id })
```

[...]

PR #9685, opened 2021-04-07
[Fix #9684] Support `IgnoredMethods` option for `Lint/AmbiguousBlockAssociation`

### LINKED ISSUE #9684: Support `IgnoredMethods` option for `Lint/AmbiguousBlockAssociation`
The proposed enhancement would allow for better customization for some common DSL, such as RSpec `change`:

```ruby
expect { do_something }.to change { object.attribute }
```

(example taken from: [https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher))

gPrado:
This PR adds support for `IgnoredMethods` parameter in `Lint/AmbiguousBlockAssociation` cop to allow better DSL support.
It fixes #9684 and is also related with #7486.

### DIFF
--- a/lib/rubocop/cop/lint/ambiguous_block_association.rb
+++ b/lib/rubocop/cop/lint/ambiguous_block_association.rb
@@ -6,6 +6,8 @@ module RuboCop
       # This cop checks for ambiguous block association with method
       # when param passed without parentheses.
       #
+      # This cop can customize ignored methods with `IgnoredMethods`.
+      #
       # @example
       #
       #   # bad
@@ -26,7 +28,14 @@ module RuboCop
       #   # good
       #   # Lambda arguments require no disambiguation
       #   foo = ->(bar) { bar.baz }
+      #
+      # @example IgnoredMethods: [change]
+      #
+      #   # good
+      #   expect { do_something }.to change { object.attribute }
       class AmbiguousBlockAssociation < Base
+        include IgnoredMethods
+
         MSG = 'Parenthesize the param `%<param>s` to make sure that the ' \
               'block will be associated with the `%<method>s` method ' \
               'call.'
@@ -50,7 +59,8 @@ module RuboCop
         end
 
         def allowed_method?(node)
-          node.assignment? || node.operator_method? || node.method?(:[])
+          node.assignment? || node.operator_method? || node.method?(:[]) ||
+            ignored_method?(node.last_argument.send_node.source)
         end
 
         def message(send_node)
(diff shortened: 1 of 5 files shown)


bbatsov:
Two points from me:

1. Generally it's a bit weird for Lint cops to have configuration - e.g. the built-in Ruby lint that this cop emulates will always give you a warning in this case and there's nothing one can do about this. You'll notice that extremely few Lint cops have any form of configuration what-so-ever. 
2. We've started to move away from "Ignored" and we prefer "Allowed" these days. 

I was also under the impression that the built-in Ruby warning has discouraged this particular style of using RSpec and it's not as common as it used to be. 

I will consider the suggested change once the PR is ready, but I'm not making any promises that it will be merged.

gPrado:
Thanks for your comments, @bbatsov 

I understand your point regarding Lint cops, but the intention for this configuration is exactly to enable support for such specific DSL, when you have absolute knowledge of the ambiguity risks. In the issue I mentioned the [change](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher) RSpec matcher, but the same offenses are also reported when using the [satisfy](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/satisfy-matcher) matcher. Currently, the only alternative left is to disable the cop entirely for (e.g.) all RSpec files.

I don't think an Allowed rule would work better here, since this is a very general Lint cop.

What do you think about a stricter configuration?
e.g. Instead of an array with ignored methods (e.g. `["change"]`), it could be configured with an array of ignored pair methods (e.g. `[["to", "change"], ["to_not", "change"]]`). Maybe, it would make the intention even clearer when configuring this cop.

bbatsov:
> I don't think an Allowed rule would work better here, since this is a very general Lint cop.

I don't get this part - "AllowedMethods" and "IgnoredMethods" mean exactly the same. It's just a different way of phrasing this.

lib/rubocop/cop/lint/ambiguous_block_association.rb
```diff
@@ -26,7 +28,14 @@ module Lint
       #   # good
       #   # Lambda arguments require no disambiguation
       #   foo = ->(bar) { bar.baz }
+      #
+      # @example IgnoredMethods: [change]
+      #
+      #   # good
+      #   expect { do_something }.to change { object.attribute }
       class AmbiguousBlockAssociation < Base
+        include IgnoredMethods
```
bbatsov, line 37: Ah, I had forgotten the mixing was named like this. You can escape from legacy. :D

bbatsov:
> What do you think about a stricter configuration?

I'd be down with this, as it's going to yield less false negatives, but it will add extra verbosity. It's hard to say how much of an impact it'd have in practice.

bbatsov:
No one had anything else to add, so it seems we can merge this. Thanks!

PR #9686, opened 2021-04-08
[Fix #9683] Fix an incorrect auto-correct for `Style/HashConversion`

### LINKED ISSUE #9683: Style/HashConversion auto-correct generates invalid code
The `Style/HashConversion` cop may generate invalid code under some specific scenarios, see example below.

--------

## Expected behavior

Given the following code:

```ruby
Hash[[1,2,3].zip]
# => {1=>nil, 2=>nil, 3=>nil} 
```

RuboCop is expected to not register an offense or to auto-correct it with an equivalent code.

## Actual behavior

RuboCop auto-corrects it as follows:

```ruby
[1,2,3].zip.to_h
# ArgumentError (wrong array length at 0 (expected 2, was 1))
```

## RuboCop version

1.12.1 (using Parser 3.0.1.0, rubocop-ast 1.4.1, running on ruby 2.6.6 x86_64-linux)

koic:
Fixes #9683.

This PR fixes an incorrect auto-correct for `Style/HashConversion` when using `zip` method in `Hash[]`.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/hash_conversion.rb
+++ b/lib/rubocop/cop/style/hash_conversion.rb
@@ -61,16 +61,38 @@ module RuboCop
         def single_argument(node)
           first_argument = node.first_argument
           if first_argument.hash_type?
-            add_offense(node, message: MSG_LITERAL_HASH_ARG) do |corrector|
-              corrector.replace(node, "{#{first_argument.source}}")
-            end
+            register_offense_for_hash(node, first_argument)
           elsif first_argument.splat_type?
             add_offense(node, message: MSG_SPLAT) unless allowed_splat_argument?
+          elsif use_zip_method_without_argument?(first_argument)
+            register_offense_for_zip_method(node, first_argument)
           else
             add_offense(node, message: MSG_TO_H) do |corrector|
               replacement = first_argument.source
               replacement = "(#{replacement})" if requires_parens?(first_argument)
               corrector.replace(node, "#{replacement}.to_h")
+            end
+          end
+        end
+
+        def use_zip_method_without_argument?(first_argument)
+          return false unless first_argument&.send_type?
+
+          first_argument.method?(:zip) && first_argument.arguments.empty?
+        end
+
+        def register_offense_for_hash(node, hash_argument)
+          add_offense(node, message: MSG_LITERAL_HASH_ARG) do |corrector|
+            corrector.replace(node, "{#{hash_argument.source}}")
+          end
+        end
+
+        def register_offense_for_zip_method(node, zip_method)
+          add_offense(node, message: MSG_TO_H) do |corrector|
+            if zip_method.parenthesized?
+              corrector.insert_before(zip_method.loc.end, '[]')
+            else
+              corrector.insert_after(zip_method, '([])')
             end
           end
         end
--- a/changelog/fix_incorrect_autocorrect_for_style_hash_conversion.md
+++ b/changelog/fix_incorrect_autocorrect_for_style_hash_conversion.md
@@ -1,0 +1,1 @@
+* [#9683](https://github.com/rubocop/rubocop/issues/9683): Fix an incorrect auto-correct for `Style/HashConversion` when using `zip` method without argument in `Hash[]`. ([@koic][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/hash_conversion.rb
```diff
@@ -45,6 +45,7 @@ class HashConversion < Base
 
         def on_send(node)
           return unless hash_from_array?(node)
+          return if node.arguments.any? { |argument| argument.send_type? && argument.method?(:zip) }
```
bbatsov, line 48: I don't like very much hardcoding such special cases. At the very least this should be mentioned in the docs.

bbatsov: Also are there other methods that might produce some output similar to that of `zip`?

bbatsov:
One more thing - if this `zip` actually had params everything would work just fine, which makes the use case extra weird. 

```ruby
[1, 3, 4, 5, 5,6].zip([1,2,3,4,5,6]).to_h
# => {1=>1, 3=>2, 4=>3, 5=>5, 6=>6}
```

koic:
> One more thing - if this `zip` actually had params everything would work just fine, which makes the use case extra weird.

I overlooked the case of passing argument to `zip`.

> Also are there other methods that might produce some output similar to that of `zip`?

Similar cases were considered, such as `combination`, `permutation`, `repeated_combination`, and `repeated_permutation`.
But I think they no need to be considered as they will also occur in `ArgumentError` with `Hash[]`.

The following is an example. These behave differently depending on whether the argument are odd or even, but they occur the same error in bad case (`Hash[]`).

## `combination` with odd number argument

```ruby
% irb
irb(main):001:0> Hash[[1, 2, 3, 4].combination(2)]
(irb):1:in `[]': odd number of arguments for Hash (ArgumentError)
        from (irb):1:in `<main>'
        from /Users/koic/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:13:in `<top (required)>'
        from /Users/koic/.rbenv/versions/3.0.1/bin/irb:23:in `load'
        from /Users/koic/.rbenv/versions/3.0.1/bin/irb:23:in `<main>'
irb(main):002:0> [1, 2, 3, 4].combination(2).to_h
=> {1=>4, 2=>4, 3=>4}
```

## `combination` with even number argument

```ruby
irb(main):003:0> Hash[[1, 2, 3, 4].combination(3)]
(irb):3:in `[]': odd number of arguments for Hash (ArgumentError)
        from (irb):3:in `<main>'
        from /Users/koic/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:13:in `<top (required)>'
        from /Users/koic/.rbenv/versions/3.0.1/bin/irb:23:in `load'
        from /Users/koic/.rbenv/versions/3.0.1/bin/irb:23:in `<main>'
irb(main):004:0> [1, 2, 3, 4].combination(3).to_h
(irb):4:in `combination': element has wrong array length (expected 2, was 3) (ArgumentError)
        from (irb):4:in `each'
        from (irb):4:in `to_h'
        from (irb):4:in `<main>'
        from /Users/koic/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:13:in `<top (required)>'
        from /Users/koic/.rbenv/versions/3.0.1/bin/irb:23:in `load'
        from /Users/koic/.rbenv/versions/3.0.1/bin/irb:23:in `<main>'
irb(main):005:0> [1, 2, 3, 4].combination(1).to_h
```

As far as I searched, only `zip` with no arguments seems to be a special case. So, I would like to update this PR with an implementation and documentation specific to this `zip` special case.

bbatsov:
Fair enough.

koic:
I updated this PR. Thank you!

gPrado:
Maybe a simpler alternative is to update the auto-correction code to handle this special case:

```ruby
Hash[[1,2,3].zip]
#  => {1=>nil, 2=>nil, 3=>nil} 
```

may be auto-corrected to:

```ruby
[1,2,3].zip([]).to_h
# => {1=>nil, 2=>nil, 3=>nil} 
```

bbatsov:
I like this idea!

bbatsov:
Thanks!

PR #9692, opened 2021-04-10
[Fix #9691] Add configuration option InspectBlocks to RedundantLineBreak

### LINKED ISSUE #9691: Add configuration parameter to Layout/RedundantLineBreak
In #9678, @koic raised a concern that users might prefer to keep existing line breaks within blocks when using the new cop `Layout/RedundantLineBreak`. A configuration parameter should be added to control whether blocks are inspected or not by this cop.

I have a fix for this in the works. A PR is coming up.

jonas054:
Set the default value for this new configuration option to `false`, since we suspect that many users might consider the `do`..`end` variant with line breaks more readable.

For RuboCop's own sources we set it to `true`, i.e., the same behavior as before the parameters existed, at least for now.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -49,6 +49,7 @@ Layout/ClassStructure:
 
 Layout/RedundantLineBreak:
   Enabled: true
+  InspectBlocks: true
 
 Layout/TrailingWhitespace:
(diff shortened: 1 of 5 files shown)


bbatsov:
Looks good. Thanks!

PR #9702, opened 2021-04-16
Avoid parser warning

marcandre:
The culprit was actually a Cop that was doing a `require 'parser/current'` that was avoidable and not actually the best thing to do.


bbatsov:
Ha, I always though this message was coming from somewhere else and we'd need to redirect the output or something along those lines to suppress. Better late than never! Thanks!

marcandre:
I thought I'd have to use `$VERBOSE = nil` before loading `parser`, but we actually load a particular version depending on the config so no need.

PR #9703, opened 2021-04-17
[Fix #9636] Resolve symlinks when excluding directories

### LINKED ISSUE #9636: Symlinks pointing to excluded directories are not excluded
After upgrading to Rubocop 0.93.0 or a more recent version, symbolic links pointing to directories that are excluded in `.rubocop.yml` are no longer excluded.

The behavior change was introduced by this PR: https://github.com/rubocop/rubocop/pull/8815

I'm not sure whether you'd consider this a bug, but it's definitely a behavior change that caught us off guard while trying to upgrade Rubocop to a more recent version.

## Expected behavior

With Rubocop 0.92.0 or older versions, symbolic links pointing to directories that are excluded in `.rubocop.yml` were themselves excluded.

E.g. with the following file tree:
```
$ ls -lR .
total 16
-rw-r--r--   1 ob  staff  109 Mar 23 22:34 .rubocop.yml
-rw-r--r--  1 ob  staff  172 Mar 23 22:29 Gemfile
-rw-r--r--  1 ob  staff  631 Mar 23 22:29 Gemfile.lock
drwxr-xr-x  3 ob  staff   96 Mar 23 22:21 build
drwxr-xr-x  3 ob  staff   96 Mar 23 22:26 foo

./build:
total 0
drwxr-xr-x  3 ob  staff  96 Mar 23 22:24 protoc_out

./build/protoc_out:
total 8
-rw-r--r--  1 ob  staff  764 Mar 23 22:22 ping_pb.rb

./foo:
total 0
lrwxr-xr-x  1 ob  staff  19 Mar 23 22:26 protoc_out -> ../build/protoc_out
```

and a `.rubocop.yml` as follows:
```yaml
AllCops:
  TargetRubyVersion: 2.7
  Exclude:
    - build/protoc_out/**/*
```

then the contents fo `foo/protoc_out` were excluded, because the link resolves to `build/protoc_out` which is itself excluded:
```
$ bundle exec rubocop -V
0.92.0 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 2.7.2 x86_64-darwin19)
$ bundle exec rubocop --list-target-files --debug
For /Users/ob/workspace/ruby/rubocop-test-link: configuration from /Users/ob/workspace/ruby/rubocop-test-link/.rubocop.yml
Default configuration from /Users/ob/workspace/ruby/rubocop-test-link/vendor/bundle/ruby/2.7.0/gems/rubocop-0.92.0/config/default.yml
Gemfile
Finished in 0.08337699994444847 seconds
```

## Actual behavior

With Rubocop 0.93.0 or more recent, symbolic links pointing to directories that are excluded in `.rubocop.yml` are no longer excluded:
```
$ bundle exec rubocop -V
0.93.0 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 2.7.2 x86_64-darwin19)
$ bundle exec rubocop --list-target-files --debug
For /Users/ob/workspace/ruby/rubocop-test-link: configuration from /Users/ob/workspace/ruby/rubocop-test-link/.rubocop.yml
Default configuration from /Users/ob/workspace/ruby/rubocop-test-link/vendor/bundle/ruby/2.7.0/gems/rubocop-0.93.0/config/default.yml
Gemfile
foo/protoc_out/ping_pb.rb
Finished in 0.08175699971616268 seconds
```

Set up a file structure as above, and run Rubocop 0.93.0 or more recent.

## RuboCop version

```
$ bundle exec rubocop -V
0.93.0 (using Parser 3.0.0.0, rubocop-ast 1.4.1, running on ruby 2.7.2 x86_64-darwin19)
```

ob-stripe:
Resolve symlinks when excluding directories, i.e. if `foo/**/*` is present in the Exclude list and there is a `bar` symlink pointing to `foo`, files under `bar` will be properly excluded.

Fixes #9636.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/target_finder.rb
+++ b/lib/rubocop/target_finder.rb
@@ -97,7 +97,11 @@ module RuboCop
       base_dir = base_dir.gsub('/{}/', '/\{}/')
       dirs = Dir.glob(File.join(base_dir.gsub('/**/', '/\**/'), '*/'), flags)
                 .reject do |dir|
-                  dir.end_with?('/./', '/../') || File.fnmatch?(exclude_pattern, dir, flags)
+                  next true if dir.end_with?('/./', '/../')
+                  next true if File.fnmatch?(exclude_pattern, dir, flags)
+
+                  File.symlink?(dir.chomp('/')) && File.fnmatch?(exclude_pattern,
+                                                                 "#{File.realpath(dir)}/", flags)
                 end
       dirs.flat_map { |dir| wanted_dir_patterns(dir, exclude_pattern, flags) }.unshift(base_dir)
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5555,3 +5555,4 @@
 [@kachick]: https://github.com/kachick
 [@corroded]: https://github.com/corroded
 [@osyo-manga]: https://github.com/osyo-manga
+[@ob-stripe]: https://github.com/ob-stripe
--- a/changelog/fix_resolve_symlinks_when_excluding_directories.md
+++ b/changelog/fix_resolve_symlinks_when_excluding_directories.md
@@ -1,0 +1,1 @@
+* [#9636](https://github.com/rubocop/rubocop/issues/9636): Resolve symlinks when excluding directories. ([@ob-stripe][])
--- a/spec/rubocop/target_finder_spec.rb
+++ b/spec/rubocop/target_finder_spec.rb
@@ -447,6 +447,33 @@ RSpec.describe RuboCop::TargetFinder, :isolated_environment do
         expect(found_basenames).to include('ruby5.rb')
       end
     end
+
+    it 'resolves symlinks when looking for excluded directories' do
+      create_link('link', 'dir1')
+
+      config = instance_double(RuboCop::Config)
+      exclude_property = { 'Exclude' => [File.expand_path('dir1/**/*')] }
+      allow(config).to receive(:for_all_cops).and_return(exclude_property)
+      allow(config_store).to receive(:for).and_return(config)
+
+      expect(found_basenames).not_to include('ruby1.rb')
+      expect(found_basenames).to include('ruby3.rb')
+    end
+
+    it 'can exclude symlinks as well as directories' do
+      Dir.mktmpdir do |tmpdir|
+        create_empty_file(File.join(tmpdir, 'ruby5.rb'))
+        create_link('link', tmpdir)
+
+        config = instance_double(RuboCop::Config)
+        exclude_property = { 'Exclude' => [File.expand_path('link/**/*')] }
+        allow(config).to receive(:for_all_cops).and_return(exclude_property)
+        allow(config_store).to receive(:for).and_return(config)
+
+        expect(found_basenames).not_to include('ruby5.rb')
+        expect(found_basenames).to include('ruby3.rb')
+      end
+    end
   end
 
   describe '#target_files_in_dir' do
--- a/spec/support/file_helper.rb
+++ b/spec/support/file_helper.rb
@@ -24,4 +24,13 @@ module FileHelper
   def create_empty_file(file_path)
     create_file(file_path, '')
   end
+
+  def create_link(link_path, target_path)
+    link_path = File.expand_path(link_path)
+
+    dir_path = File.dirname(link_path)
+    FileUtils.makedirs dir_path unless File.exist?(dir_path)
+
+    FileUtils.symlink(target_path, link_path)
+  end
 end


bbatsov:
Did you measure the performance impact of the fix? I guess some hit is unavoidable, but I'm curious how significant it is.

bbatsov:
To be clear - I feel this fixes a legitimate bug, but given the history of how this bug was introduced I want to make sure that resolving the real paths is not going to cause a performance issue.

ob-stripe:
@bbatsov Thanks for the quick reply!

I just did some very basic performance testing, running the following script on my laptop (Macbook Pro 2017, ruby 2.7.2p137) from the root of the rubocop repository:
```ruby
require 'benchmark'
require_relative './lib/rubocop'

config = RuboCop::ConfigStore.new
options = {force_exclusion: false, debug: false}
target_finder = RuboCop::TargetFinder.new(config, options)

Benchmark.bm do |x|
  x.report { 1_000.times { target_finder.target_files_in_dir(__dir__) } }
end
```

Results for Rubocop 1.12.1:
```
       user     system      total        real
  52.574793  30.188860  82.763653 ( 88.034613)
```

Results for Rubocop 1.12.1 + my patch:
```
       user     system      total        real
  52.337204  34.995932  87.333136 ( 91.022406)
```

It looks like the additional `realpath` calls add about 16% of system time and about 5% of total time.

Let me know if there are additional tests you'd like me to run!

bbatsov:
@rubocop/rubocop-core What do you think, guys? It's not a big performance hit, but it's still a step back. On the other hand I do believe that symlinks should be resolved.

lib/rubocop/target_finder.rb
```diff
@@ -97,7 +97,9 @@ def wanted_dir_patterns(base_dir, exclude_pattern, flags)
       base_dir = base_dir.gsub('/{}/', '/\{}/')
       dirs = Dir.glob(File.join(base_dir.gsub('/**/', '/\**/'), '*/'), flags)
                 .reject do |dir|
-                  dir.end_with?('/./', '/../') || File.fnmatch?(exclude_pattern, dir, flags)
+                  dir.end_with?('/./', '/../') ||
+                    File.fnmatch?(exclude_pattern, dir, flags) ||
+                    File.fnmatch?(exclude_pattern, "#{File.realpath(dir)}/", flags)
```
dvandersluis, line 102: I think line 101 is redundant here, and that might help performance. I ran your benchmark and got this:

master:
```
$ bundle exec ruby benchmark.rb
       user     system      total        real
  68.409056  14.974495  83.383551 ( 84.883239)
```

patched:
```
$ bundle exec ruby benchmark.rb
       user     system      total        real
  67.697482  15.706053  83.403535 ( 84.581582)
```

bbatsov: Great point!

ob-stripe: That would make it impossible to exclude paths by their symlink names though, and would change the current behavior. I've added a test for this scenario in a new commit.

dvandersluis: Alright, thanks for that added test.

We should still avoid doing multiple `File.fnmatch?` when `dir` is not a symlink (which is probably the majority case by far). Something like (could be cleaned up probably):

```ruby
dirs = Dir.glob(File.join(base_dir.gsub('/**/', '/\**/'), '*/'), flags)
          .reject do |dir|
            next true if dir.end_with?('/./', '/../')
          
            if File.symlink?(dir.chomp('/'))
              File.fnmatch?(exclude_pattern, dir, flags) ||
                File.fnmatch?(exclude_pattern, "#{File.realpath(dir)}/", flags)
            else
              File.fnmatch?(exclude_pattern, dir, flags)
            end
          end
```

```
       user     system      total        real
  66.163276  14.268255  80.431531 ( 81.158242)
```

marcandre:
Looks good, and what @dvandersluis said 👍

bbatsov:
Agreed. Let's wait for the changes proposed by @dvandersluis to be applied and I'll merge this. I was planning to do a new release today, but I can wait for this PR to make it into the release.

lib/rubocop/target_finder.rb
ob-stripe: The symlink check makes sense. I've updated the PR with your suggestion.

dvandersluis (approved):

bbatsov:
Thanks!

ob-stripe:
Thank you all! Looking forward to the release, as this issue was preventing us from upgrading :)

PR #9706, opened 2021-04-18
[Fix #5576] Treat relativity of Include parameters correctly

### LINKED ISSUE #5576: Subdirectory .rubocop.yml with any rules besides inheritance causes some cops to fail
As part of a Rails 5 upgrade I've come across the new style for Rails controller testing. Thankfully this is covered and has autocorrect available in the `Rails/HttpPositionalArguments` Cop. Unfortunately this cop simply either does not run or does not work when I have a `spec/.rubocop.yml`. I've commented out the entire file and removed all the contents before saving and running rubocop again. Until I actually remove the file the cop fails.

I'm not sure how many other cops this problem extends to. The spec local .rubocop.yml is just used to change some linting, metric, and rubocop-rspec behaviors. I do not believe the contents matters as the presence of a blank file triggers the error as well.

--------

## Expected behavior

`rubocop --rails --only HttpPositionalArguments` should list ~150 failures in my test suite.

## Actual behavior

No offenses are detected

If I simply do `touch spec/.rubocop.yml` before running rubocop I am not notified of the HttpPositionalArguments failures.

## RuboCop version

```
$ rubocop -V
0.52.1 (using Parser 2.4.0.2, running on ruby 2.5.0 x86_64-darwin16)
```

jonas054:
For configuration files whose name begins with `.rubocop`, the paths in `Include` parameters are relative to the directory of the configuration file.

When another configuration file inherits from such a file, we cannot just copy the `Include` paths. We must modify them so they select the files that were originally intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 * [#7977](https://github.com/rubocop/rubocop/issues/7977): Add `Layout/RedundantLineBreak` cop. ([@jonas054][])
 * [#9691](https://github.com/rubocop/rubocop/issues/9691): Add configuration parameter `InspectBlocks` to `Layout/RedundantLineBreak`. ([@jonas054][])
 
+### Bug fixes
+
+* [#5576](https://github.com/rubocop/rubocop/issues/5576): Fix problem with inherited `Include` parameters. ([@jonas054][])
+
 ## 1.12.1 (2021-04-04)
 
 ### Bug fixes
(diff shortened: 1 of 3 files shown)


bbatsov:
That's one important fix! Thanks!

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 * [#7977](https://github.com/rubocop/rubocop/issues/7977): Add `Layout/RedundantLineBreak` cop. ([@jonas054][])
 * [#9691](https://github.com/rubocop/rubocop/issues/9691): Add configuration parameter `InspectBlocks` to `Layout/RedundantLineBreak`. ([@jonas054][])
 
+### Bug fixes
```
bbatsov, line 10: Seems you really don't like the new way to add changelog entries. 😆

jonas054: On the contrary. I like it a lot. I just forgot that it existed. :slightly_smiling_face:

PR #9709, opened 2021-04-18
[Fix #9689] Treat parens around array items the same for children and deeper descendants

### LINKED ISSUE #9689: Unsafe auto-correct for Style/RedundantParentheses
When auto-correcting a redundant parenthesis inside an array with assignments in it, the resulting code is invalid Ruby. e.g. consider:

```ruby
[
 x = (
   1
 ),
 y = 2
]
```

Produces the invalid syntax:

```ruby
[
  x =
    1
  ,
  y = 2
]
```

--------

## Expected behavior

Should either be marked unsafe, or autocorrect to:

```ruby
[
 x = 1,
 y = 2
]
```

Auto-correct code above with `Style/RedundantParenteses`

## RuboCop version

```
1.12.1 (using Parser 3.0.1.0, rubocop-ast 1.4.1, running on ruby 2.6.5 x86_64-darwin19)
```

dvandersluis:
Previously there was a disparity between these two code blocks:

```ruby
[
 (
   1
 ),
 2
]
```

```ruby
[
 x = (
   1
 ),
 y = 2
]
```

The first was accepted as "plausible" but the second registered an offense (and was autocorrected to a syntax error because the comma in an array cannot be on its own line).

This change allows both to be treated the same by looking for any ancestor to be an array or hash rather than just the direct parent.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_treat_parens_around_array_items_the_same.md
+++ b/changelog/fix_treat_parens_around_array_items_the_same.md
@@ -1,0 +1,1 @@
+* [#9689](https://github.com/rubocop/rubocop/issues/9689): Treat parens around array items the same for children and deeper descendants. ([@dvandersluis][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Looks good! Thanks!

PR #9575, opened 2021-03-09
Layout/ClassStructure fixes [wip]

marcandre:
This is a complete refactoring.

Bug fixes:

1) Correctly distinguishes private / protected / public, including constants and class methods
2) Offenses raised are now complete, accurate, and minimal [#9738]
3) `attr_accessor`/`writer`/`reader` and `initializer` will now default to `methods` if not specified in the order
4) All other known bugs are fixed

PR #9724, opened 2021-04-21
Extract constant term from loop.

marcandre:
The case of the dynamic constant was the only one where we wouldn't move the code


marcandre:
Including in #9575

PR #9727, opened 2021-04-22
[Fix #7669] Add Bundler/GemVersion cop

### LINKED ISSUE #7669: Feature request: Bundler/GemsSpecifySemanticVersion
I'd like to enforce that gems in my Gemfile are all specified with a semantic version string -- e.g. the `~>1.7` in

```ruby
  gem 'foobar_thing', '~>1.7'
```

My understanding is that there are no good reasons to omit this version specifier; the "worst case" that still counts as a best practice is to simply use a greater-than operator.

I have a basic implementation of this that doesn't choke on special cases like GitHub-based dependencies:

```ruby
# Load all declared gems from gemfile
dependencies = Bundler::Definition.build('Gemfile', nil, nil).dependencies

# Separate git-based dependencies from others (proper processing between ref / version TBD)
git_deps, other_deps = dependencies.partition { |d| d.source.class == Bundler::Source::Git }

# Version of "0" is the default if unspecified
unversioned_deps = other_deps.select {|d| d.requirement.requirements[0][1].version == "0"}

# print results
puts "Github-based gems: #{git_deps.size}"
git_deps.map(&:name).sort.each { |n| puts "  #{n}" }
puts
puts "Gemfile gems without semver: #{unversioned_deps.size}"
unversioned_deps.map(&:name).sort.each { |n| puts "  #{n}" }
```

Is this something that could be added to the list of [Bundler Cops](https://rubocop.readthedocs.io/en/latest/cops_bundler/)?

timlkelly:
- `Bundler/GemVersion` enforces that gem version declarations are either required or prohibited in the `Gemfile`.
- `Bundler/GemVersion` is disabled by default. 
- Gems can be omitted from this check by adding the gem to the `AllowedGems` configuration option.

I originally wrote this cop for a private repository and then added more configuration options after seeing issue #7669. 

The AST checks and regex caught everything I could throw at it, but I realize that is probably limited compared to the larger Ruby ecosystem as a whole. I would appreciate any feedback you may have on those checks.

References:
https://github.com/rubocop/rubocop/discussions/9720
https://github.com/rubocop/rubocop/issues/7669

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -174,6 +174,18 @@ Bundler/GemComment:
   IgnoredGems: []
   OnlyFor: []
 
+Bundler/GemVersionDeclaration:
+  Description: 'Requires or prohibits gem version declarations.'
+  Enabled: false
+  VersionAdded: '1.13.0'
```
timlkelly, line 180: Should this be left blank? `1.13.0` is already published. I wasn't sure what to put here.

config/default.yml
```diff
@@ -174,6 +174,18 @@ Bundler/GemComment:
...
+  EnforcedStyle: 'required'
+  SupportedStyles:
+    - 'required'
+    - 'prohibited'
+  Include:
+    - '**/Gemfile'
```
timlkelly, line 186: I saw that other `Bundler` cops also include `*.gemfile` and `gems.rb`. I've personally never worked with this format of gem file, so I was unsure what exactly to test.

timlkelly:
The CircleCI builds do not provide a link to check their status. Is there a way I can check on their status? They seem to be queued but haven't been processed as far as I can tell.

config/default.yml
bbatsov: The format is exactly the same, it's just a different filename.

lib/rubocop/cop/bundler/gem_version_declaration.rb
```diff
@@ -0,0 +1,101 @@
...
+      #  # bad
+      #  gem 'rubocop', '>= 1.10.0'
+      #
+      #  # bad
+      #  gem 'rubocop', '>= 1.5.0', '< 1.10.0'
+      #
+      class GemVersionDeclaration < Base
+        include ConfigurableEnforcedStyle
+
+        REQUIRED_MSG = 'Gem version declaration is required.'
```
bbatsov, line 38: declaration -> specification.

config/default.yml
```diff
@@ -174,6 +174,18 @@ Bundler/GemComment:
...
   OnlyFor: []
 
+Bundler/GemVersionDeclaration:
+  Description: 'Requires or prohibits gem version declarations.'
+  Enabled: false
+  VersionAdded: '<<next>>'
+  EnforcedStyle: 'required'
+  SupportedStyles:
+    - 'required'
+    - 'prohibited'
```
bbatsov, line 184: I think we've typically used "forbidden" in the codebase, but you'll have to check this.

config/default.yml
```diff
@@ -174,6 +174,18 @@ Bundler/GemComment:
...
+  Include:
+    - '**/Gemfile'
+  IgnoredGems: []
```
bbatsov, line 187: Lately I've favoured more names like "AllowedGems" instead of "IgnoredGems". Sadly, our use of both is a bit inconsistent.

bbatsov:
I'd probably name this just "GemVersion" or something like this. Overall, I think that's a nice cop to have.

config/default.yml
dvandersluis: +1 for `AllowedGems`.

lib/rubocop/cop/bundler/gem_version_declaration.rb
```diff
@@ -0,0 +1,101 @@
...
+        PROHIBITED_MSG = 'Gem version declaration is prohibited.'
+        VERSION_DECLARATION_REGEX = /^[~<>=]*\s?[0-9.]+/.freeze
```
dvandersluis, line 40:
```suggestion
        VERSION_DECLARATION_REGEX = /^\s*[~<>=]*\s*[0-9.]+/.freeze
```

Multiple spaces are valid in bundler before and after the operator. We might also want to be more specific about the format here, but it probably doesn't _really_ matter.

lib/rubocop/cop/bundler/gem_version_declaration.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        # @!method gem_declaration?(node)
+        def_node_matcher :gem_declaration?, '(send nil? :gem str ...)'
```
dvandersluis, line 43: This method is in `Bundler::GemComment` as well, it might be useful to extract them to a shared mixin.

config/default.yml
timlkelly: Forbidden definitely seems to be more common than prohibited, good to know!

lib/rubocop/cop/bundler/gem_version_declaration.rb
timlkelly: I've added the suggested change.

> We also could be more specific about the format here

Do you mind detailing what you mean by this? Or is it the change you provided?

lib/rubocop/cop/bundler/gem_version_declaration.rb
timlkelly: Yup, definitely. I did copy this from `Bundler::GemComment` to get it working originally. But I completely agree that a mixin will DRY this up.

lib/rubocop/cop/bundler/gem_version_declaration.rb
dvandersluis: What I mean is the Regexp will match `......` or `>>>>> 7` or other random strings that aren’t version specifications, but again it probably doesn't really matter. 

It could be more strict with something like `^\s*(?:~>|[<>]?=?)?\s*[0-9]+(\.[0-9]+)*`

bbatsov:
Great work! Thanks!

PR #9728, opened 2021-04-22
Add `PreferHashRocketsForQuotedSymbols` to `Style/HashSyntax`

sambostock:
## TL;DR

This adds `PreferHashRocketsForQuotedSymbols` to `Style/HashSyntax`, allowing `{ "foo": 123 }` to be reported as an offence.

## Why?

Given the elegant syntax for symbol hash keys
```ruby
{
  foo: 123,
}
```
developers often try to use it in scenarios where they need a string key
```ruby
{
  'foo': 123,
}
```
not realizing the key remains a symbol, and that they needed to use
```ruby
{
  'foo' => 123,
}
```
instead.

While there are legitimate uses for quoting a symbol (if it includes special characters), in most cases this is not the intended use case.

Therefore, the `PreferHashRocketsForQuotedSymbols` option is added to `Style/HashSyntax`, to give the option to disallow the error-prone syntax.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/hash_syntax.rb
+++ b/lib/rubocop/cop/style/hash_syntax.rb
@@ -54,6 +54,8 @@ module RuboCop
       #   # good
       #   {a: 1, b: 2}
       #   {:c => 3, 'd' => 4}
+      #
+      # TODO: Add to docs
       class HashSyntax < Base
         include ConfigurableEnforcedStyle
         include RangeHelp
@@ -144,6 +146,12 @@ module RuboCop
              # similarly for other non-alnum final characters (except quotes,
              # to prefer { "x y": 1 } over { :"x y" => 1 }).
              !/[\p{Alnum}"']\z/.match?(sym_name)
+            return false
+          end
+
+          if cop_config['PreferHashRocketsForQuotedSymbols'] &&
+              # Prefer { :"x y" => 1 } over { "x y": 1 }.
+              /\A['"]|['"]\z/.match?(sym_name)
             return false
           end
 
--- a/spec/rubocop/cop/style/hash_syntax_spec.rb
+++ b/spec/rubocop/cop/style/hash_syntax_spec.rb
@@ -124,6 +124,48 @@ RSpec.describe RuboCop::Cop::Style::HashSyntax, :config do
         end
       end
 
+      context 'if PreferHashRocketsForQuotedSymbols is true' do
+        let(:cop_config_overrides) { { 'PreferHashRocketsForQuotedSymbols' => true } }
+
+        it 'accepts hash rockets when key is double quoted symbol' do
+          expect_no_offenses('x = { :"fancy symbol" => 0 }')
+        end
+
+        it 'accepts hash rockets when key is single quoted symbol' do
+          expect_no_offenses("x = { :'fancy symbol' => 0 }")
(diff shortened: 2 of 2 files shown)


dvandersluis:
This is already covered by `Lint/SymbolConversion`:

```
$ rubocop -A --only Lint/SymbolConversion test.rb
Inspecting 1 file
W

Offenses:

test.rb:2:3: W: [Corrected] Lint/SymbolConversion: Unnecessary symbol conversion; use foo: instead.
  'foo': 123,
  ^^^^^
```

Which results in:

```ruby
{
  foo: 123,
}
```

sambostock:
Ahh, interesting. Thanks for the info @dvandersluis! It can be hard to find cops unless you know exactly what to look for. Perhaps there would be value in a "Related Cops" section in the documentation.

sambostock:
Actually, it looks like there still isn't a way to disallow usage like this without disabling Ruby 1.9 Hash syntax entirely:

```ruby
{
  "symbols requiring quoting":     123,
  "look like strings to new devs": 456
}
```
`Lint/SymbolConversion` would not catch these, because they do in-fact require quoting (because of the spaces).

Moreover, usage of the `consistent` style would mean that a single quoted-Symbol make the entire hash have to switch to quotes.

```ruby
{
  "pt-br": "Brazilian Portuguese", # This requires quotes due to dash character.
  "en":    "English",              # Therefore these must be quoted too,
  "fr":    "French",               # due to EnforcedStyle: consistent
}
```

In a large codebase with many devs new to Ruby, it may be preferable to enforce that even a single quoted-Symbol appearing in a Hash is an offence, requiring either the removal of redundant quotes, or switching to hash-rocket syntax, so as to be unambiguous even to the untrained eye.

```ruby
{
  "pt-br" => "Brazilian Portuguese", # Use HashRocket for all keys due to this one
  "en"    => "English",
  "fr"    => "French",
}

{
  unrelated_hash:              true, # Without needing to use HashRockets in all Hashes
  with_simple_keys:            true,
  still_using_ruby_1_9_syntax: true
}
```

bbatsov:
> Perhaps there would be value in a "Related Cops" section in the documentation.

Love this suggestion, btw! //cc @koic

lib/rubocop/cop/style/hash_syntax.rb
```diff
@@ -147,6 +149,12 @@ def acceptable_19_syntax_symbol?(sym_name)
             return false
           end
 
+          if cop_config['PreferHashRocketsForQuotedSymbols'] &&
```
dvandersluis, line 152: Please add this configuration key to `config/default.yml` with a default value. We also need to add docs/examples to the cop's top comment for this configuration.

dvandersluis:
I'm not 100% sure I understand the point of this. I think your thinking is that `:"foo bar" => true` makes it clearer that the key is a symbol than `"foo bar": true`, but I'm not sure someone new to Ruby will realize that either form is a symbol and not a string. Why would you want to have a special case for quoted symbols (ie. needing hashrockets), but not for any other symbol key?

It seems to me like you almost would want to use hashrockets everywhere for consistency, or have a cop that disallows hash keys to be symbols that need to be quoted (which might be a good custom cop for your codebase?)

PR #9737, opened 2021-04-25
[Fix #9733] Add cop Layout/SingleLineBlockChain

### LINKED ISSUE #9733: Add Layout/RedundantLineBreak EnforcedStyle

By the way, thanks for adding this cop. Looking forward to making more use of it. :tada:

Would it be possible to add an `EnforcedStyle` to this cop so one could configure the cop to *avoid nested blocks*? Example:

``` yaml
Layout/RedundantLineBreak:
  Enabled: true
  EnforcedStyle: suffixed_blocks_only
```

Not married to my choice of words so if there is a better way to describe this, go for it. Mostly trying to identify a way in which to define this.

For example, default behavior will reformat code as follows:

``` ruby
# No
example.select { |item| item.cond? }.join('-')
```

However, I'd like to ensure this remains untouched as follows:

``` ruby
# Yes
example.select { |item| item.cond? }
       .join('-')
```

The reason is due to the block syntax since I'd like to avoid chaining blocks unless they are at the end. For example this, would be acceptable:

``` ruby
# Yes
example.split.each { |item| put item }
```

Notice the block is only at the end of the chain. This reads well because the eye can quickly scan a series of chained messages ending with a block. By slipping the block in the middle, reading the code becomes jarring, especially if multiple blocks are chained together (as multi-chained blocks can get out of hand quickly and would like to discourage that use).

I don't know if any other alternative solutions to this problem at the moment.

## Additional context

If it helps, I do consider message chaining with blocks to be an antipattern and have written up a more detailed reason for avoiding this usage in this [article](https://www.alchemists.io/articles/ruby_antipatterns/#_message_chains) in case it's of help.

jonas054:
This cop checks if method calls are chained onto single line blocks. It considers that a line break before the dot improves the readability of the code. It is disabled by default since the style it enforces is probably not universally preferred.

The `Layout/RedundantLineBreak` needs to stay away from offenses that `Layout/SingleLineBlockChain` has a conflicting opinion about, if `Layout/SingleLineBlockChain` is enabled.


bbatsov:
Looks good to me!

P.S. First I thought this cop had something to do with "Blockchain". 😆

PR #9748, opened 2021-04-29
Prevent infinite loops during symlink traversal

Tonkpils:
https://github.com/rubocop/rubocop/pull/8815 introduced a traversal strategy that used recursion.
https://github.com/rubocop/rubocop/pull/9703 then fixed an issue with this traversal which accounted for directories and symlinks.

When a symlink points to a parent directory that contains that symlink it'll cause this to go into a loop until the filename is too long for glob to handle.

We prevent this by checking for the inclusion of a symlink's real path in the base directory's realpath. If the base directory's path starts with the symlink's destination then we are in a loop and should skip processing the directory

-----------------

* [x] Added tests.


bbatsov:
So many symlink-related problems. :-) Thanks!

PR #9751, opened 2021-04-29
Fix string literal autocorrect

etiennebarrie:
With this config:

```
Style/StringLiterals:
  EnforcedStyle: double_quotes
```

`'\\'` is not autocorrected to `"\\"` which it could

and

`'#$foo'` is autocorrected to `"\#$foo"` which it shouldn't, because it adds an escape.

Sorry I did those two in the same PR, but that would have added conflicts otherwise, since they change the same line.
Let me know if you'd really rather have 2 PRs.

-----------------

* [x] Added tests.


bbatsov:
Thanks!

PR #9755, opened 2021-05-01
[Fix #9734] Add new `Style/TopLevelMethodDefinitions` cop

### LINKED ISSUE #9734: add cop for "global method"
## feature request

We have a project were many people contribute that don't know ruby well, they often write methods that are global, I'd like to stop them
this also often happens in things like Rakefile or *.rake files

```ruby
class Foo
  def not_global <- Good
  end
end
def this_is_global <- Bad
end
```

`git grep '^def ` works, but would be nice if it's in rubocop

tejasbubane:
Closes #9734

-----------------

* [x] Added tests.


lib/rubocop/cop/style/global_method.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # This cop looks for definitions of global methods.
```
koic, line 6: Can you document the reason why it is disabled by default?
https://github.com/rubocop/rubocop/issues/9734#issuecomment-826048992

config/default.yml
```diff
@@ -3421,6 +3421,11 @@ Style/FrozenStringLiteralComment:
     - never
   SafeAutoCorrect: false
 
+Style/GlobalMethod:
```
bbatsov, line 3424: I'm not sure whether "global" is the right terminology for this. I prefer to think of those as "top-level" method definitions. @rubocop/rubocop-core Ideas how to name this are welcome.

lib/rubocop/cop/style/global_method.rb
```diff
@@ -0,0 +1,46 @@
...
+      #     end
+      #   end
+      #
+      #   # good
+      #   Struct.new do
+      #     def some_method
+      #     end
+      #   end
+      class GlobalMethod < Base
+        MSG = 'Do not use global methods.'
```
bbatsov, line 35: Maybe "Do not define methods at the top-level." instead?

lib/rubocop/cop/style/global_method.rb
bbatsov: I'd probably add a bit more rationale about the cop in general here. I guess some people might not realize where those top-level definitions end up and why there are problematic.

bbatsov:
I think you can also use `define_method` at the top-level, which such a cop should probably detect.

config/default.yml
koic: > I prefer to think of those as "top-level" method definitions. 

I agree with this. `TopLevelMethodDefinitions` cop may be clear for the term "top-level". There may be other better names.

tejasbubane:
@bbatsov @koic Changed as per your comments.

dvandersluis:
I probably wouldn't want these to be allowed anyways, but I could see an argument for allowing a `defs` node where the receiver isn't `self`, since in that case, the method will end up scoped to an object and not be global.

ie.

```ruby
def MyClass.foo
  # ...
end
```

I'm fine with this being disallowed but anyone have an opinion there?

koic:
I think it can be allowed if a receiver is explicitly specified. I think top-level methods that are implicitly declared to `Object` without specifying a receiver should be prevented.

config/default.yml
```diff
@@ -4632,6 +4632,11 @@ Style/TernaryParentheses:
     - require_parentheses_when_complex
   AllowSafeAssignment: true
 
+Style/TopLevelMethodDefinitions:
```
bbatsov, line 4635: Let's use singular here (Definition) to be aligned with how most of the cops are named.

bbatsov:
You forgot to update the commit message itself. 

Btw, I'm thinking that it would be hard to add some check whether the file being analyzed is a script (if it has a shebang) and disable to the cop just for scripts, but probably that's an overkill at this point.

bbatsov:
I guess we should also document this in the Ruby style guide. I'm surprised we haven't done so already.

tejasbubane:
@bbatsov Changed to singular and created PR for styleguide.

bbatsov:
Thanks!

PR #9756, opened 2021-05-01
[Fix #9731] Fix two autocorrection issues for `Style/NegatedIfElseCondition`

### LINKED ISSUE #9731: Style/NegatedIfElseCondition bugs with specific conditions
The Style/NegatedIfElseCondition cop bugs with very specific conditions with error `Parser::Source::TreeRewriter detected clobbering`.

```
1 error occurred:
An error occurred while Style/NegatedIfElseCondition cop was inspecting ...
```
--------

Create a file with the code below and run rubocop on it:

```ruby
if !a_condition
  # comment
  flag = false
  do_a
end

if !another_condition
  # comment
  flag = false
  do_b
else
  do_c
end
```

This seems very specific since it works if you remove the else block, if you remove the comments...

## RuboCop version

```
$ bundle exec rubocop -V
1.12.1 (using Parser 3.0.1.0, rubocop-ast 1.4.1, running on ruby 2.6.7 x86_64-darwin20)
  - rubocop-performance 1.9.2
  - rubocop-rails 2.9.1
  - rubocop-rspec 2.2.0
```

dvandersluis:
Previously, when swapping the `if` and `else` branches, if one of the branches only had a single statement, the branch would end up not being spaced out properly.

Additionally, when there is a node with a comment inside the `if` that has an identical node with a comment *before* the `if` statement, `CommentsHelp.begin_pos_with_comment` returns an incorrect range, which was causing a clobbering error. I fixed this by creating ranges of the entire body of the if and else branches, so that they can just be swapped without relying on `ProcessedSource#ast_with_comments`. See rubocop/rubocop-ast#179.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/negated_if_else_condition.rb
+++ b/lib/rubocop/cop/style/negated_if_else_condition.rb
@@ -29,7 +29,6 @@ module RuboCop
       #
       class NegatedIfElseCondition < Base
         include RangeHelp
-        include CommentsHelp
         extend AutoCorrector
 
         MSG = 'Invert the negated condition and swap the %<type>s branches.'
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/negated_if_else_condition_spec.rb
```diff
@@ -50,6 +50,21 @@
...
+        do_something :
+        do_something_else # comment
+    RUBY
+
+    expect_correction(<<~RUBY)
+      x ?
+        do_something_else :
+        do_something # comment
+    RUBY
+  end
```
dvandersluis, line 66: This doesn't actually correct the comment properly but I wanted to add a test for this case just to document behaviour. Really no one should be writing multiline ternaries like this. 😏

lib/rubocop/cop/style/negated_if_else_condition.rb
```diff
@@ -98,21 +97,38 @@ def swap_branches(corrector, node)
...
+        # Collect the entire if branch, including whitespace and comments
+        def if_range(node)
+          if node.ternary?
+            node.if_branch
+          else
+            Parser::Source::Range.new(
+              processed_source.buffer,
+              node.condition.loc.expression.end_pos,
+              node.loc.else.begin_pos
+            )
```
koic, line 117: That's just my two cents :-)

```suggestion
            range_between(node.condition.loc.expression.end_pos, node.loc.else.begin_pos)
```

lib/rubocop/cop/style/negated_if_else_condition.rb
```diff
@@ -98,21 +97,38 @@ def swap_branches(corrector, node)
...
+        # Collect the entire else branch, including whitespace and comments
+        def else_range(node)
+          if node.ternary?
+            node.else_branch
+          else
+            Parser::Source::Range.new(
+              processed_source.buffer,
+              node.loc.else.end_pos,
+              node.loc.end.begin_pos
+            )
```
koic, line 130:
```suggestion
            range_between(node.loc.else.end_pos, node.loc.end.begin_pos)
```

dvandersluis:
Thanks for the review @koic!

PR #9765, opened 2021-05-03
[Fix #9761] Fix `Style/ClassAndModuleChildren` false negative for `compact` style when a class/module is partially nested

### LINKED ISSUE #9761: ClassAndModuleChildren compact syntax false negative when outer module is already nested
When using the `ClassAndModuleChildren` rule with `EnforcedStyle: compact`, it does not report violations when the outer module is already nested:

```rb
module Foo::Bar
  class Baz
  end
end
```

--------

## Expected behavior

The above would report a violation. It should require:

```rb
class Foo::Bar::Baz
end
```

## Actual behavior

It does not report a violation.

See above.

## RuboCop version

```
1.13.0 (using Parser 3.0.1.0, rubocop-ast 1.4.1, running on ruby 3.0.1 x86_64-linux)
  - rubocop-rails 2.9.1
  - rubocop-rspec 2.3.0
```

dvandersluis:
This fix makes `Style/ClassAndModuleChildren` register an offense for the following type of code with the `compact` style:

```ruby
class Foo::Bar
  class Baz
  end
end
```

Fixes #9761.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_fix_styleclassandmodulechildren_false.md
+++ b/changelog/fix_fix_styleclassandmodulechildren_false.md
@@ -1,0 +1,1 @@
+* [#9761](https://github.com/rubocop/rubocop/issues/9761): Fix `Style/ClassAndModuleChildren` false negative for `compact` style when a class/module is partially nested. ([@dvandersluis][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Nice!

PR #9772, opened 2021-05-05
[Fix #9771] Fix a false positive for `Style/TrivialAccessors`

### LINKED ISSUE #9771: Style/TrivialAccessors make unwanted changes
## Expected behavior

no action

## Actual behavior

Code:

```ruby
  def foo(foo)
    @foo = foo
  end
```

is suggested to be replaced with `attr_wrirer`, but it's false suggestion (attention to `=` sign in `def`)

## RuboCop version

```
1.12.1 (using Parser 3.0.1.0, rubocop-ast 1.4.1, running on ruby 2.7.2 x86_64-darwin19)
```

koic:
Fixes #9771.

This PR fixes a false positive for `Style/TrivialAccessors` when defining a method that is not a writer.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positive_for_style_trivial_accessors.md
+++ b/changelog/fix_false_positive_for_style_trivial_accessors.md
@@ -1,0 +1,1 @@
+* [#9771](https://github.com/rubocop/rubocop/issues/9771): Change `AllowDSLWriters` to true by default for `Style/TrivialAccessors`. ([@koic][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -4745,7 +4745,7 @@ Style/TrivialAccessors:
   StyleGuide: '#attr_family'
   Enabled: true
   VersionAdded: '0.9'
-  VersionChanged: '0.77'
+  VersionChanged: '<<next>>'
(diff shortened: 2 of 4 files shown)


bbatsov:
I have the vague memory that this was by design, so people would uncover "accidental" accessors, regardless of the name. I also thought we had some configuration about the exactness of the name match, but I don't see it. I wonder if I'm thinking of some similar cop. :-)

bbatsov:
Ah, actually it seems we didn't include examples of the config options in the docs, which is why I didn't see them at first. I was thinking of `ExactNameMatch`.

koic:
Actually the issue was because the default value of `AllowDSLWriters` option.
I referred to https://github.com/rubocop/rubocop/pull/963 and it looks like the default value was maybe set in opposite.

I've updated PR by changing the default value from false to true for `AllowDSLWriters` option adding missing examples.

bbatsov:
Great work with the examples! I had forgotten how many options this cop had, but now they will be a lot easier to comprehend/discover.

PR #9774, opened 2021-05-05
[Fix #9773] Fix `Style/EmptyLiteral` to not register offenses for `String.new` when `Style/FrozenStringLiteral` is enabled

### LINKED ISSUE #9773: Style/EmptyLiteral and frozen_string_literal generate broken code
--------

## Expected behavior

Example made up program showing the issue:

```ruby
result = String.new
amount = gets.chomp.to_i
amount.times do
  result << "hello world\n"
end

puts result
```

Usage:

```
ruby foo.rb
2
hello world
hello world
```

## Actual behavior

The `Style/EmptyLiteral` rule combined with the `frozen_string_literal` generates an invalid program:

```
# frozen_string_literal: true

result = ''
amount = gets.chomp.to_i
amount.times do
  result << "hello world\n"
end

puts result
```

Program error:
```
$ ruby foo.rb 
5
	2: from foo.rb:5:in `<main>'
	1: from foo.rb:5:in `times'
foo.rb:6:in `block in <main>': can't modify frozen String: "" (FrozenError)
```

Example rules:
```
foo.rb:1:1: C: [Corrected] Style/FrozenStringLiteralComment: Missing frozen string literal comment.
result = String.new
^
foo.rb:1:10: C: [Corrected] Style/EmptyLiteral: Use string literal '' instead of String.new.
result = String.new
         ^^^^^^^^^^
foo.rb:2:1: C: [Corrected] Layout/EmptyLineAfterMagicComment: Add an empty line after magic comments.
result = ''
^
```

Interestingly this problem only arises if the file doesn't already have the `frozen_string_literal` rule, so having the frozen string literal comment beforehand doesn't actually break the program as the Style/EmptyLiteral rule doesn't run.

```
# frozen_string_literal: true

result = String.new
amount = gets.chomp.to_i
amount.times do
  result << "hello world\n"
end

puts result
```

Create a file with the example text above, and run `rubocop -A foo.rb`

## RuboCop version

```
$ rubocop -V
warning: parser/current is loading parser/ruby27, which recognizes
warning: 2.7.3-compliant syntax, but you are running 2.7.2.
1.12.1 (using Parser 3.0.1.0, rubocop-ast 1.4.1, running on ruby 2.7.2 x86_64-darwin19)
```

dvandersluis:
`Style/EmptyLiteral` replaces `String.new` with `''`, unless the file has a `frozen-string-literal: true` magic comment. However, if `Style/FrozenStringLiteral` is also enabled, this will cause broken code, because the magic comment will be added afterwards.

This change makes string behaviour the same if `Style/FrozenStringLiteral` is enabled as if the magic comment is found. However, if `Style/FrozenStringLiteral` is enabled but there is a `frozen-string-literal: false` magic comment, `Style/EmptyLiteral` will continue to replace the literal.

Fixes #9773.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/frozen_string_literal.rb
+++ b/lib/rubocop/cop/mixin/frozen_string_literal.rb
@@ -35,6 +35,12 @@ module RuboCop
         leading_comment_lines.any? { |line| MagicComment.parse(line).frozen_string_literal? }
       end
 
+      def frozen_string_literals_disabled?
+        leading_comment_lines.any? do |line|
+          MagicComment.parse(line).frozen_string_literal == false
+        end
+      end
+
       def frozen_string_literal_specified?
         leading_comment_lines.any? do |line|
           MagicComment.parse(line).frozen_string_literal_specified?
--- a/changelog/fix_fix_styleemptyliteral_to_not_register.md
+++ b/changelog/fix_fix_styleemptyliteral_to_not_register.md
@@ -1,0 +1,1 @@
+* [#9773](https://github.com/rubocop/rubocop/issues/9773): Fix `Style/EmptyLiteral` to not register offenses for `String.new` when `Style/FrozenStringLiteral` is enabled. ([@dvandersluis][])
--- a/lib/rubocop/cop/style/empty_literal.rb
+++ b/lib/rubocop/cop/style/empty_literal.rb
@@ -62,7 +62,7 @@ module RuboCop
             ARR_MSG
           elsif offense_hash_node?(node)
             HASH_MSG
-          elsif str_node(node) && !frozen_string_literals_enabled?
+          elsif str_node(node) && !frozen_strings?
             format(STR_MSG, prefer: preferred_string_literal)
           end
         end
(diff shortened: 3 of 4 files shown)


lib/rubocop/cop/mixin/frozen_string_literal.rb
```diff
@@ -35,6 +35,10 @@ def frozen_string_literals_enabled?
         leading_comment_lines.any? { |line| MagicComment.parse(line).frozen_string_literal? }
       end
 
+      def frozen_string_literals_disabled?
```
AndreiEres, line 38: Could you please explain, why we're not able to use previous method `frozen_string_literals_enabled?`. In case we can't it even looks strange because we check leading comment for `MagicComment.parse(line).frozen_string_literal?` twice

dvandersluis: Because the `frozen_string_literal` value can be anything, and is only cast to `true` or `false` if it is literally one of those strings. I probably should change the `disabled?` method to check `frozen_string_literal == false`.

adfoster-r7:
Ran this against our codebase and it worked great - thanks for looking into this! :+1:

koic:
Cool👍 This solution approach to the issue seems to make sense!

PR #9775, opened 2021-05-05
[Fix #9770] Update `Lint/EmptyBlock` to handle procs the same way as lambdas

### LINKED ISSUE #9770: `Lint/EmptyBlock` should not apply to empty procs
Similar to #9015 / #9016, empty procs (e.g. `Proc.new {}` or `proc {}`) are also useful as callable placeholders when the callable may receive arguments.

_Originally posted by @gPrado in https://github.com/rubocop/rubocop/issues/9015#issuecomment-832127048_

dvandersluis:
`Lint/EmptyBlock` now treats procs the same way as lambdas (governed by the `AllowLambdas` configuration).

Fixes #9770.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/empty_block.rb
+++ b/lib/rubocop/cop/lint/empty_block.rb
@@ -7,7 +7,11 @@ module RuboCop
       # Such empty blocks are typically an oversight or we should provide a comment
       # be clearer what we're aiming for.
       #
-      # Empty lambdas are ignored by default.
+      # Empty lambdas and procs are ignored by default.
+      #
+      # NOTE: For backwards compatibility, the configuration that allows/disallows
+      # empty lambdas and procs is called `AllowEmptyLambdas`, even though it also
+      # applies to procs.
       #
       # @example
       #   # bad
@@ -40,6 +44,10 @@ module RuboCop
       #   end
       #   (callable || placeholder).call
       #
+      #   proc { }
+      #
+      #   Proc.new { }
+      #
       # @example AllowEmptyLambdas: false
       #   # bad
       #   allow(subject).to receive(:callable).and_return(-> {})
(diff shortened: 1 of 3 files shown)


dvandersluis:
@bbatsov it'd probably be "better" to replace the `AllowLambdas` configuration with `AllowLambdasAndProcs` or something, but this would break existing configurations. I was thinking about deprecating `AllowLambdas` with a warning (like we did in #9098) but it's probably "good enough" to just overload the configuration.

If you want me to change it, let me know.

koic (approved):
> it's probably "good enough" to just overload the configuration.

I agree with this opinion. It seems better to me to update the `AllowEmptyLambdas` configuration.

lib/rubocop/cop/lint/empty_block.rb
```diff
@@ -7,7 +7,11 @@ module Lint
       # Such empty blocks are typically an oversight or we should provide a comment
       # be clearer what we're aiming for.
       #
-      # Empty lambdas are ignored by default.
+      # Empty lambdas and procs are ignored by default.
+      #
+      # NOTE: For backwards compatibility, the configuration that allows/disallows
+      # empty lambdas and procs is called `AllowLambdas`, even though it also applies
```
koic, line 13: Is this type?

```suggestion
      # empty lambdas and procs is called `AllowEmptyLambdas`, even though it also applies
```

dvandersluis: Yes thanks @koic!

dvandersluis: Updated, thank you ❤️

bbatsov:
The change is fine in it's current form. Thanks!

PR #9778, opened 2021-05-06
Add `disabled_all?` to DirectiveComment

AndreiEres:
Because we have `DirectiveComment#enabled_all?` I also have added `disabled_all?` checking and I'm going to use it in further PR.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/directive_comment.rb
+++ b/lib/rubocop/directive_comment.rb
@@ -64,6 +64,11 @@ module RuboCop
       !disabled? && all_cops?
     end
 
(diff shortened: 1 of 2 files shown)


koic:
Could you update this PR to the code that actually uses the `disabled_all?` method and bundle into one commit, instead of the code that is not referenced from anywhere?
This clarifies use case, goal, change reason and supports maintenance tracking.

AndreiEres:
Of course :-) So I'll close the PR

PR #9786, opened 2021-05-07
Clarify documentation about the `Enabled` and `EnabledByDefault`/`DisabledByDefault`

dvandersluis:
I rewrote this section a bit because it was not worded very clearly. I also wanted to make sure it's explicit that `EnabledByDefault` will also enable cops that are disabled by RuboCop.

### DIFF
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -452,8 +452,12 @@ Most cops are enabled by default. Cops, introduced or significantly updated
 between major versions, are in a special pending status (read more in
 xref:versioning.adoc["Versioning"]). Some cops, configured the above `Enabled: false`
 in https://github.com/rubocop/rubocop/blob/master/config/default.yml[config/default.yml],
-are disabled by default. The cop enabling process can be altered by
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -452,8 +452,12 @@ Most cops are enabled by default. Cops, introduced or significantly updated
 between major versions, are in a special pending status (read more in
 xref:versioning.adoc["Versioning"]). Some cops, configured the above `Enabled: false`
 in https://github.com/rubocop/rubocop/blob/master/config/default.yml[config/default.yml],
-are disabled by default. The cop enabling process can be altered by
-setting `DisabledByDefault` or `EnabledByDefault` (but not both) to `true`.
+are disabled by default.
+
+The cop enabling process can be altered by setting `DisabledByDefault` or
+`EnabledByDefault` (but not both) to `true`. These settings override the default *all*
```
bbatsov, line 458: the default or by default? Something definitely seems wrong about this sentence. :-)

dvandersluis: 🤦 thanks for catching! I've updated it.

PR #9795, opened 2021-05-13
[Change #9744] Auto disable parallel when combined with cache false

### LINKED ISSUE #9744: How to make --parallel and --cache false work together?
I'm using the [linter-rubocop](https://github.com/AtomLinter/linter-rubocop) package in Atom, that sets `--cache false` for calls to rubocop. Since activation of `--parallel` via `.rubocop` file in my project, I get the error ```-P/--parallel uses caching to speed up execution, so combining with --cache false is not allowed.```

--------

## Expected behavior

I see two options:
- Deactivate `--parallel` if `--cache false` is set, similar to https://github.com/rubocop/rubocop/blob/74377b0502a211036713afca3c69a1198b0a6dc7/lib/rubocop/options.rb#L349-L354
- Allow `--no-parallel` CLI option to override the `--parallel` from `.rubocop` file.

Eventually `bundle exec rubocop --cache false [--no-parallel]` should run rubocop and not output
`-P/--parallel uses caching to speed up execution, so combining with --cache false is not allowed.`

## Actual behavior
Rubocop prints `-P/--parallel uses caching to speed up execution, so combining with --cache false is not allowed.`

Add `--parallel` to `.rubocop` file and run `bundle exec rubocop --cache false [--no-parallel]`

## RuboCop version
```
$ [bundle exec] rubocop -V
1.13.0 (using Parser 3.0.1.0, rubocop-ast 1.4.1, running on ruby 3.0.1 x86_64-darwin20)
  - rubocop-performance 1.10.2
  - rubocop-rails 2.9.1
  - rubocop-rspec 2.2.0
```

rrosenblum:
This closes #9744. 

Modify the behavior of `--parallel` combined with `--cache false` to emit a warning and disable parallel. It previously would raise an error and halt execution. This is consistent with the changes in #9647

### DIFF
--- a/changelog/change_disable_parallel_with_cache_false.md
+++ b/changelog/change_disable_parallel_with_cache_false.md
@@ -1,0 +1,1 @@
+* [#9744](https://github.com/rubocop/rubocop/pull/9744): The parallel flag will now be automatically ignored when combined with `--cache false`. Previously, an error was raised and execution stopped. ([@rrosenblum][])
--- a/lib/rubocop/options.rb
+++ b/lib/rubocop/options.rb
@@ -297,7 +297,7 @@ module RuboCop
       validate_auto_gen_config
       validate_auto_correct
       validate_display_only_failed
-      validate_parallel
+      disable_parallel_when_invalid_option_combo
 
       return if incompatible_options.size <= 1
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #9796, opened 2021-05-13
[Fix #9792] Fix false positive for `Lint/Void` cop

### LINKED ISSUE #9792: Lint/Void false positive result
## Expected behavior

```ruby
class Foo
  def foo
    1..100.times do |_n|
      break unless rand > 0.5
    end
    raise
  end
end
```

be ok

## Actual behavior

```bash
$ be rubocop
foo.rb:5:5: W: Lint/Void: Literal 1..100.times do |_n|
      break unless rand > 0.5
    end used in void context.
    1..100.times do |_n| ...
    ^^^^^^^^^^^^^^^^^^^^
```

If replace to `(1..100).times` then everything works as it should

## RuboCop version

```
1.14.0 (using Parser 3.0.1.1, rubocop-ast 1.5.0, running on ruby 2.7.2 x86_64-darwin19)
```

tejasbubane:
Closes #9792

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_lint_void_false_positive.md
+++ b/changelog/fix_lint_void_false_positive.md
@@ -1,0 +1,1 @@
+* [#9792](https://github.com/rubocop/rubocop/issues/9792): Fix false positive for `Lint/Void` cop with ranges. ([@tejasbubane][])
--- a/lib/rubocop/cop/lint/void.rb
+++ b/lib/rubocop/cop/lint/void.rb
@@ -104,7 +104,7 @@ module RuboCop
         end
 
         def check_literal(node)
-          return if !node.literal? || node.xstr_type?
+          return if !node.literal? || node.xstr_type? || node.range_type?
 
           add_offense(node, message: format(LIT_MSG, lit: node.source))
         end
--- a/spec/rubocop/cop/lint/void_spec.rb
+++ b/spec/rubocop/cop/lint/void_spec.rb
@@ -254,4 +254,22 @@ RSpec.describe RuboCop::Cop::Lint::Void, :config do
       nil
     RUBY
   end
+
+  it 'accepts method with irange block' do
+    expect_no_offenses(<<~RUBY)
+      def foo
+        1..100.times.each { puts 1 }
+        do_something
+      end
+    RUBY
+  end
+
+  it 'accepts method with erange block' do
+    expect_no_offenses(<<~RUBY)
+      def foo
+        1...100.times.each { puts 1 }
+        do_something
+      end
+    RUBY
+  end
 end


changelog/fix_lint_void_false_positive.md
```diff
@@ -0,0 +1 @@
+* [#9792](https://github.com/rubocop/rubocop/issues/9792): Fix false positive for `Lint/Void` cop. ([@tejasbubane][])
```
koic, line 1: Can you explain a little more about the fix?

lib/rubocop/cop/lint/void.rb
```diff
@@ -104,7 +104,7 @@ def check_var(node)
         end
 
         def check_literal(node)
-          return if !node.literal? || node.xstr_type?
+          return if !node.literal? || node.xstr_type? || node.irange_type?
```
koic, line 107:
```suggestion
          return if !node.literal? || node.xstr_type? || node.range_type?
```

spec/rubocop/cop/lint/void_spec.rb
```diff
@@ -254,4 +254,13 @@ def foo=(rhs)
       nil
     RUBY
   end
+
+  it 'handles method with irange block' do
```
koic, line 258: It looks like `accepts` instead of `handles`.

spec/rubocop/cop/lint/void_spec.rb
```diff
@@ -254,4 +254,13 @@ def foo=(rhs)
...
+    expect_no_offenses(<<~RUBY)
+      def foo
+        1..100.times.each { puts 1 }
+        raise
```
koic, line 262: That's just my two cents. `do_something` may be better as an example than `raise`.

```suggestion
        do_something
```

spec/rubocop/cop/lint/void_spec.rb
```diff
@@ -254,4 +254,13 @@ def foo=(rhs)
...
+      end
+    RUBY
+  end
```
koic, line 265: Can you add a test for `erange`?

tejasbubane:
@koic Changed as per your comments.

spec/rubocop/cop/lint/void_spec.rb
```diff
@@ -254,4 +254,14 @@ def foo=(rhs)
       nil
     RUBY
   end
+
+  it 'accepts method with range block' do
+    expect_no_offenses(<<~RUBY)
+      def foo
+        1..100.times.each { puts 1 }
+        1...100.times.each { puts 1 }
```
koic, line 262: Can you split it into irange and erange tests?

tejasbubane: Done.

lib/rubocop/cop/lint/void.rb
```diff
@@ -104,7 +104,7 @@ def check_var(node)
         end
 
         def check_literal(node)
-          return if !node.literal? || node.xstr_type?
+          return if !node.literal? || node.xstr_type? || node.range_type?
```
bbatsov, line 107: I wonder why `literal?` doesn't cover them. I think they weren't immutable or something like this, but my memory on the subject is fuzzy.

PR #9807, opened 2021-05-17
[Fix #9803] Fix `Bundler/GemVersion` cop not respecting git tags

### LINKED ISSUE #9803: Bundler/GemVersion not respecting tags
## Expected behavior

I expect the `Bundler/GemVersion` to respect tags on private repositories as "correct", since they lock the version.
This was working fine in the previous versions

## Actual behavior

The Cop is currently flagging all private gems as "wrong"

```
Gemfile:78:3: E: Bundler/GemVersion: Gem version specification is required.
  gem 'inkycop', github: 'customink/inkycop', tag: '3.4.1'
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

```
# Base gem for all Custom Ink Clients, pinned to this version to handle compatibility issues.
gem 'client_client', github: 'customink/client_client', tag: 'v0.2.1'
# Client to obtain design information from Custom Ink
gem 'design_client', github: 'customink/design_client', tag: 'v1.4.5'
# GDPR Implementation for Ruby on Rails
gem 'gdpr_rails', github: 'customink/gdpr_rails', tag: '0.7.0'
# Gem to obtain information about ink-colors in designs
gem 'ink_colors', '>=1.7.4', git: 'git@github.com:customink/ink_colors.git', branch: 'master'
# Gem to determine whether the application is up and running
gem 'is_it_up', github: 'customink/is_it_up', tag: '1.0.0'
# Gem to communicate with Custom Ink MMS
gem 'mms_client', github: 'customink/mms_client', tag: 'v2.2.1'
# Gem to communicate with Custom Ink OFS
gem 'order_fulfillment_client', github: 'customink/order_fulfillment_client', tag: 'v0.4.12'
# Gem to communicate with Custom Ink OIS
gem 'order_infrastructure', git: 'git@github.com:customink/order_infrastructure.git', tag: 'v0.31.0'
# Gem to communicate with Custom Ink Mint Service
gem 'payment_client', github: 'customink/payment_client', tag: 'v1.4.0'
# Gem to communicate with Custom Ink Quote Service
gem 'quote_client', github: 'customink/quote_client', tag: 'v0.0.9'
```

- Create a Gemfile with github references to repo's
- Set the version through tags
- Run RuboCop

## RuboCop version

```
➜  customink_international git:(coding-bunny/bump_inkycop) ✗ bundle exec rubocop -V
1.14.0 (using Parser 3.0.1.1, rubocop-ast 1.5.0, running on ruby 3.0.0 x86_64-linux)
  - rubocop-performance 1.11.3
  - rubocop-rails 2.10.1
  - rubocop-rake 0.5.1
  - rubocop-rspec 2.3.0

```

tejasbubane:
Closes #9803 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/bundler/gem_version.rb
+++ b/lib/rubocop/cop/bundler/gem_version.rb
@@ -45,6 +45,21 @@ module RuboCop
           (send nil? :gem <(str #version_specification?) ...>)
         PATTERN
 
+        # @!method with_git_ref?(node)
+        def_node_matcher :with_git_ref?, <<~PATTERN
+          (send nil? :gem <(hash <#git? #tag_ref? ...>) ...>)
+        PATTERN
+
+        # @!method tag_ref?(node)
+        def_node_matcher :tag_ref?, <<~PATTERN
+          (pair (sym {:tag :ref}) (str _))
+        PATTERN
+
+        # @!method git?(node)
+        def_node_matcher :git?, <<~PATTERN
+          (pair (sym {:git :github :bitbucket}) (str _))
+        PATTERN
+
         def on_send(node)
           return unless gem_declaration?(node)
           return if allowed_gem?(node)
@@ -78,8 +93,15 @@ module RuboCop
         end
 
         def offense?(node)
-          (required_style? && !includes_version_specification?(node)) ||
-            (forbidden_style? && includes_version_specification?(node))
+          required_offense?(node) || forbidden_offense?(node)
+        end
+
+        def required_offense?(node)
+          required_style? && !includes_version_specification?(node) && !with_git_ref?(node)
+        end
+
+        def forbidden_offense?(node)
+          forbidden_style? && includes_version_specification?(node)
         end
 
         def forbidden_style?
--- a/spec/rubocop/cop/bundler/gem_version_spec.rb
+++ b/spec/rubocop/cop/bundler/gem_version_spec.rb
@@ -15,6 +15,8 @@ RSpec.describe RuboCop::Cop::Bundler::GemVersion, :config do
         ^^^^^^^^^^^^^ Gem version specification is required.
         gem 'rubocop', require: false
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Gem version specification is required.
+        gem 'rubocop', tag: '1.2.0'
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Gem version specification is required.
       RUBY
     end
 
@@ -24,6 +26,9 @@ RSpec.describe RuboCop::Cop::Bundler::GemVersion, :config do
         gem 'rubocop', '~> 1'
         gem 'rubocop', '~> 1.12', require: false
         gem 'rubocop', '>= 1.5.0', '< 1.10.0', git: 'https://github.com/rubocop/rubocop'
+        gem 'rubocop', github: 'rubocop/rubocop', tag: 'v1'
+        gem 'rubocop', git: 'https://github.com/rubocop/rubocop', ref: 'b3f37bc7f'
+        gem 'foobar', bitbucket: 'foo/bar', tag: 'v1'
       RUBY
     end
 
@@ -65,6 +70,14 @@ RSpec.describe RuboCop::Cop::Bundler::GemVersion, :config do
     it 'does not flag gems included in AllowedGems metadata' do
       expect_no_offenses(<<~RUBY)
         gem 'rspec', '~> 3.10'
+      RUBY
+    end
+
+    it 'does not flag gems using git source with tag or ref' do
+      expect_no_offenses(<<~RUBY)
+        gem 'rubocop', github: 'rubocop/rubocop', tag: 'v1'
+        gem 'rubocop', git: 'https://github.com/rubocop/rubocop', ref: 'b3f37bc7f'
+        gem 'foobar', bitbucket: 'foo/bar', tag: 'v1'
       RUBY
     end
   end
--- a/changelog/fix_gem_version_not_respecting_tags.md
+++ b/changelog/fix_gem_version_not_respecting_tags.md
@@ -1,0 +1,1 @@
+* [#9803](https://github.com/rubocop/rubocop/pull/9803): Fix `Bundler/GemVersion` cop not respecting git tags. ([@tejasbubane][])


dvandersluis:
Can you add a test that uses `git:` instead of github?

Also I wonder if `git`/`github` with a `branch` key should be accepted too?

tejasbubane:
> Can you add a test that uses `git:` instead of github?

Added specs with `git` and `github` both.

> Also I wonder if `git`/`github` with a `branch` key should be accepted too?

I don't think so. Tags are usually fixed. Branch reference keeps changing so it is not really `pinning` a version.

lib/rubocop/cop/bundler/gem_version.rb
```diff
@@ -42,7 +42,9 @@ class GemVersion < Base
 
         # @!method includes_version_specification?(node)
         def_node_matcher :includes_version_specification?, <<~PATTERN
-          (send nil? :gem <(str #version_specification?) ...>)
+          (send nil? :gem
+            <{(str #version_specification?)
+              (hash <(pair (sym :tag) (str _)) (pair (sym {:git :github}) (str _)) ...>)} ...>)
```
timlkelly, line 47: Is matching `git` or `github` required? Would `(sym :tag)` be enough to indicate tag presence? That way the cop doesn't need to be concerned about the gem source.

If it is needed, it looks like [`bitbucket`](https://bundler.io/v2.2/man/gemfile.5.html#BITBUCKET) is also a valid option to provide.

spec/rubocop/cop/bundler/gem_version_spec.rb
```diff
@@ -52,6 +56,10 @@
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Gem version specification is forbidden.
         gem 'rubocop', '>= 1.5.0', '< 1.10.0', git: 'https://github.com/rubocop/rubocop'
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Gem version specification is forbidden.
+        gem 'rubocop', github: 'rubocop/rubocop', tag: 'v1'
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Gem version specification is forbidden.
+        gem 'rubocop', git: 'https://github.com/rubocop/rubocop', tag: 'v1'
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Gem version specification is forbidden.
```
koic, line 62: These look strange. If `tag` or `ref` is specified with `git` or `github`, it's likely that a meaningful version of commit is being used. So, shouldn't these cases always be accepted?

tejasbubane: This is for `'EnforcedStyle' => 'forbidden'` which forbids adding a gem version (or in this case an equivalent git tag).

lib/rubocop/cop/bundler/gem_version.rb
tejasbubane: I did not want this to report false positives - hence the extra check. 

Thanks I was not aware of the `bitbucket` option. Looks like there is `gist` as well. I'll add them both.

tejasbubane: Added `bitbucket` only, not sure if `gist` allows versions.

dvandersluis: Users can also add their own arbitrary git sources using `git_source` so we should probably handle that as well.

spec/rubocop/cop/bundler/gem_version_spec.rb
koic: Practically, if `:tag` or `:ref` with `:git` or `:github` is used, the pinned value should be respected. AFAIK, even `'EnforcedStyle' => 'forbidden'` should not be forbidden, as it is used for intentional forks and the like.

spec/rubocop/cop/bundler/gem_version_spec.rb
```diff
@@ -24,6 +26,9 @@
         gem 'rubocop', '~> 1'
         gem 'rubocop', '~> 1.12', require: false
         gem 'rubocop', '>= 1.5.0', '< 1.10.0', git: 'https://github.com/rubocop/rubocop'
+        gem 'rubocop', github: 'rubocop/rubocop', tag: 'v1'
+        gem 'rubocop', git: 'https://github.com/rubocop/rubocop', tag: 'v1'
+        gem 'rubocop', bitbucket: 'https://bitbucket.com/foo/bar', tag: 'v1'
```
koic, line 31: `:ref` can be accepted as well as `:tag`.

spec/rubocop/cop/bundler/gem_version_spec.rb
tejasbubane: changed

spec/rubocop/cop/bundler/gem_version_spec.rb
tejasbubane: changed

lib/rubocop/cop/bundler/gem_version.rb
tejasbubane: @dvandersluis I think this might be tricky given that we can define custom sources:

```ruby
git_source(:stash) { |repo_name| "https://stash.corp.acme.pl/#{repo_name}.git" }
...
gem 'rails', :stash => 'forks/rails'
```

dvandersluis: I agree it's tricky, but if we don't handle it someway it leads to false reporting (which is maybe another argument for just looking for `tag`/`ref`/etc. and not caring about the other keys).

bbatsov:
Two bits of feedback from me:

* Do we really need so many matchers? After all it seems to me we only care for the presence of `:tag` and `:ref`, as it's unlikely someone would use them without some VCS reference. 
* The examples need to be updated to reflect the support for `:tag` and `:ref`.

lib/rubocop/cop/bundler/gem_version.rb
```diff
@@ -45,6 +45,21 @@ class GemVersion < Base
           (send nil? :gem <(str #version_specification?) ...>)
         PATTERN
 
+        # @!method with_git_ref?(node)
+        def_node_matcher :with_git_ref?, <<~PATTERN
```
bbatsov, line 49: I'd use a more neutral name here - e.g. `commit_ref?`.

lib/rubocop/cop/bundler/gem_version.rb
```diff
@@ -45,6 +45,21 @@ class GemVersion < Base
...
+          (send nil? :gem <(hash <#git? #tag_ref? ...>) ...>)
+        PATTERN
+
+        # @!method tag_ref?(node)
+        def_node_matcher :tag_ref?, <<~PATTERN
+          (pair (sym {:tag :ref}) (str _))
+        PATTERN
+
+        # @!method git?(node)
+        def_node_matcher :git?, <<~PATTERN
```
bbatsov, line 59: Do we really need to change for the VCS? I doubt someone would use `:tag` or `:ref` without them and this will probably result in some mistake. Haven't tried it, though.

bbatsov:
@tejasbubane ping :-)

bbatsov:
Superseded by #9885.

PR #9825, opened 2021-05-26
Add new `Lint/EmptyInPattern` cop

koic:
This PR adds new `Lint/EmptyInPattern` cop for Ruby 2.7's pattern matching.
It checks for the presence of `in` branches without a body.

```ruby
# bad
case condition
in [a]
  do_something
in [a, b]
end

# good
case condition
in [a]
  do_something
in [a, b]
  nil
end

# good - AllowComments: true (default)
case condition
in [a]
  do_something
in [a, b]
  # noop
end

# bad - AllowComments: false
case condition
in [a]
  do_something
in [a, b]
  # noop
end
```

This cop is similar to `Lint/EmptyWhen`, but with different supported syntax and Ruby version (requires 2.7 or higher).

And this PR use rubocop/rubocop-ast#183 feature, so it requires RuboCop AST 1.6.0 or higher.

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/empty_in_pattern.rb
```diff
@@ -0,0 +1,62 @@
...
+      class EmptyInPattern < Base
+        extend TargetRubyVersion
+
+        MSG = 'Avoid `in` branches without a body.'
+
+        minimum_target_ruby_version 2.7
+
+        def on_case_match(node)
+          node.in_pattern_branches.each do |branch|
+            next if branch.node_parts[-1] || cop_config['AllowComments'] && comment_lines?(node)
```
koic, line 54: When https://github.com/rubocop/rubocop-ast/pull/183 can be used, `branch.node_parts[-1]` can be updated to `branch.body`.

koic: I updated it.

bbatsov:
Looks good!

PR #9828, opened 2021-05-26
[Fix #9827] Add basic auth support to download raw yml config from private repo

### LINKED ISSUE #9827: Add support for token auth to download raw yml config from private repo
## Is your feature request related to a problem? 

We are requiring auth for all internal Github repos, we will need a way to inherit and cache the raw yaml shared configs with auth so that RuboCop's existing download and cache logic would work the same way with minimal overall impact.

Since `inherit_from` supports ERB pre-precessing, we can set `.rubocop.yml`

``` yml
# .rubocop.yml

inherit_from:
  - <%= "https://#{ENV['GITHUB_PERSONAL_ACCESS_TOKEN']}@raw.github.com/AirWick219/../rubocop-main.yml" %>
```
and the token is in the `uri.user`
Screen Shot 2021-05-26 at 12 20 50 PM

Seems like we can do this [here](https://github.com/rubocop/rubocop/blob/bd27fcb0840ef3a6ae5975ab708c015d29f21fa1/lib/rubocop/remote_config.rb#L58)

``` ruby
request['Authorization'] = "token #{uri.user}" if uri.user
```

The above solution would be the simplest way to tap into existing logic. if not we will have to.
1. hack it with a new rake task to first download the shared config with curl and `inherit_from` that file
    - we lost the cache supports and dev will need to run a separate rake task just for rubocop
2. hack a binstub to download the shared config with curl and replace the `bin/rubocop`
    - still hacky but easier for dev to integrate

AirWick219:
This PR adds Token auth support so that users can inherit shared config from the internal private repo.  

This addresses https://github.com/rubocop/rubocop/issues/9827

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_support_token_auth_to_download_shared_config.md
+++ b/changelog/new_support_token_auth_to_download_shared_config.md
@@ -1,0 +1,1 @@
+* [#9827](https://github.com/rubocop/rubocop/issues/9827): Add Token auth support to download raw yml config from private repo. ([@AirWick219][])
--- a/changelog/new_support_token_auth_to_download_shared_config.md
+++ b/changelog/new_support_token_auth_to_download_shared_config.md
@@ -1,1 +1,0 @@
-* [#9827](https://github.com/rubocop/rubocop/issues/9827): Add Token auth support to download raw yml config from private repo. ([@AirWick219][])
--- a/lib/rubocop/remote_config.rb
+++ b/lib/rubocop/remote_config.rb
@@ -55,6 +55,7 @@ module RuboCop
     def generate_request(uri)
       request = Net::HTTP::Get.new(uri.request_uri)
 
+      request['Authorization'] = "token #{uri.user}" if uri.user
       request['If-Modified-Since'] = File.stat(cache_path).mtime.rfc2822 if cache_path_exists?
 
       yield request
@@ -70,7 +71,7 @@ module RuboCop
         begin
           response.error!
         rescue StandardError => e
-          message = "#{e.message} while downloading remote config file #{uri}"
+          message = "#{e.message} while downloading remote config file #{cloned_url}"
           raise e, message
         end
       end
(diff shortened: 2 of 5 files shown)


dvandersluis:
Can you add some tests for this feature please?

changelog/new_support_token_auth_to_download_shared_config.md
```diff
@@ -0,0 +1 @@
+* [#9828](https://github.com/rubocop/rubocop/pull/9828): Add Token auth support to download raw yml config from private repo. ([@AirWick219][])
```
dvandersluis, line 1:
```suggestion
* [#9827](https://github.com/rubocop/rubocop/issue/9827): Add Token auth support to download raw yml config from private repo. ([@AirWick219][])
```

Since this PR has a related issue, the changelog item should point to it.

dvandersluis (changes requested):
Can you add some tests for this feature please?

dvandersluis (approved):

koic (approved):
This is a new feature. So it will be merged when we can confirm that the next release is not a bugfix release. Thank you.

lib/rubocop/remote_config.rb
```diff
@@ -55,6 +55,7 @@ def request(uri = @uri, limit = 10, &block)
     def generate_request(uri)
       request = Net::HTTP::Get.new(uri.request_uri)
 
+      request['Authorization'] = "token #{uri.user}" if uri.user
```
nickpith, line 58: I'm not sure this really should be Github specific. A Personal Access Token (PAT) can be used in Github using basic auth - which is the syntax suggested here where the user field of the URL is the PAT. Should this line be removed and set the user on the `request` object like this?

```
request.basic_auth(uri.user, uri.password) if uri.user
```

I tested some sample code and this works for Github using a PAT and should work for any URL that supports basic auth.

AirWick219: O cool. That works. Didn't know PAT works with basic auth. Thanks

nickpith (approved):

AirWick219:
@dvandersluis and @koic Let me know if you guys are good with the changes

PR #9841, opened 2021-05-31
Support Ruby 2.7's pattern matching for `Layout/SpaceAroundKeyword`

koic:
This PR supports guard `if` and `unless` syntax keywords of Ruby 2.7's pattern matching for `Layout/SpaceAroundKeyword`.

```ruby
case 42
in 42 if true
  puts 'hello'
end
```

These nodes are called `if-guard` and `unless-guard` when using guard `if` and `unless` in pattern matching.

```ruby
% ruby-parse example.rb
(case-match
  (int 42)
  (in-pattern
    (int 42)
    (if-guard
      (true))
    (send nil :puts
      (str "hello"))) nil)
```

See below for this guard semantics:
https://github.com/whitequark/parser/pull/574#issuecomment-498365609

-----------------

* [x] Added tests.


bbatsov:
Thanks!

PR #9847, opened 2021-06-02
[Fix #9843] Fix `Style/RedundantSelf` to allow conditional nodes to use `self` in the condition when a variable named is shadowed inside

### LINKED ISSUE #9843: wrong detection of Style/RedundantSelf
Rubocop incorrectly detects a case of `Style/RedundantSelf` in the following example:
~~~ruby
def foo
  bar = self.bar if self.bar
  bar
end
~~~

--------

## Expected behavior

Should not falsely detect `Style/RedundantSelf` in the provided example.

## Actual behavior

When running with `rubocop -a` it "fixes" it to the following:

~~~ruby
def foo
  bar = self.bar if bar
  bar
end
~~~

This is incorrect, as the value of `bar` in this case will always be `nil` regardless of `self.bar`

~~~
For /private/tmp/rubocop: Default configuration from /Users/vitaly/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/rubocop-1.16.0/config/default.yml
Inspecting 1 file
Scanning /private/tmp/rubocop/bar.rb
C

Offenses:

bar.rb:1:1: C: [Correctable] Style/FrozenStringLiteralComment: Missing frozen string literal comment.
def foo
^
bar.rb:2:21: C: [Corrected] Style/RedundantSelf: Redundant self detected.
  bar = self.bar if self.bar
                    ^^^^^^^^

1 file inspected, 2 offenses detected, 1 offense corrected, 1 more offense can be corrected with `rubocop -A`
Finished in 0.12974799999210518 seconds
~~~

Create a file `foo.rb` with the following content:
~~~ruby
def foo
  bar = self.bar if self.bar
  bar
end
~~~

Run `rubocop -a foo.rb`

It should not remove `self` after `if`

## RuboCop version

```
$ rubocop -V
1.16.0 (using Parser 3.0.1.1, rubocop-ast 1.7.0, running on ruby 2.6.5 x86_64-darwin19)
```

dvandersluis:
Previously, `Style/RedundantSelf` registered false positives when a condition used `self` to avoid shadowing a local variable:

```ruby
a = self.a if self.a
```

This was because the condition was not correlated to the lvar assign because they happen at different times. This is now fixed so `if`, `unless`, `while` and `until` nodes are covered.

Fixes #9843.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_fix_styleredundantself_to_allow.md
+++ b/changelog/fix_fix_styleredundantself_to_allow.md
@@ -1,0 +1,1 @@
+* [#9843](https://github.com/rubocop/rubocop/issues/9843): Fix `Style/RedundantSelf` to allow conditional nodes to use `self` in the condition when a variable named is shadowed inside. ([@dvandersluis][])
--- a/lib/rubocop/cop/style/redundant_self.rb
+++ b/lib/rubocop/cop/style/redundant_self.rb
@@ -92,7 +92,7 @@ module RuboCop
 
         def on_masgn(node)
           lhs, rhs = *node
-          lhs.children.each { |child| add_lhs_to_local_variables_scopes(rhs, child.to_a.first) }
+          add_masgn_lhs_variables(rhs, lhs)
         end
 
         def on_lvasgn(node)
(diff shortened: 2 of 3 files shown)


dvandersluis:
Failing CI seems like a false negative.

koic:
This looks good! Thank you for catching up with `while`, `until`, and `masgn` cases!

PR #9854, opened 2021-06-05
Allow braced numeric block calls in Style/MethodCallWithArgsParentheses

gsamokovarov:
This is yet another fix for the `omit_parentheses` style of the
`Style/MethodCallWithArgsParentheses` cop. We're issuing offenses for
braced `numblock` calls:

```ruby
updated_ids = Set.new(items_to_update) { _1.remote_id.to_s }
                     ^^^^^^^^^^^^^^^^^ Omit parentheses for method calls with arguments.
```

Removing the parens leads to code that doesn't compile so this is not good.

Numeric block have their own AST node type:

```
-> ruby-parse -e "Set.new(items_to_update) { _1.remote_id }"
(numblock
  (send
    (const nil :Set) :new
    (send nil :items_to_update)) 1
  (send
    (lvar :_1) :remote_id))
```

vs

```
-> ruby-parse -e "Set.new(items_to_update) { |it| it.remote_id }"
(block
  (send
    (const nil :Set) :new
    (send nil :items_to_update))
  (args
    (procarg0
      (arg :it)))
  (send
    (lvar :it) :remote_id))
```

Send nodes can be queried for their block if given with `#block_node`,
however, the `numblock`s don't get recognized as such. Our check for braced
block calls used to rely on it:

https://github.com/rubocop/rubocop-ast/blob/master/lib/rubocop/ast/node/mixin/method_dispatch_node.rb#L35-L37
https://github.com/rubocop/rubocop-ast/blob/master/lib/rubocop/ast/node/mixin/method_dispatch_node.rb#L156-L158

I'm changing the implementation to work around this issue and recognize
`numblock` braced calls, but we may consider moving the fix to
`rubocop-ast` as well.

### DIFF
--- a/spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
+++ b/spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
@@ -404,6 +404,14 @@ RSpec.describe RuboCop::Cop::Style::MethodCallWithArgsParentheses, :config do
       end
     end
 
+    context 'numbered parameters in 2.7', :ruby27 do
+      it 'accepts parens for braced numeric block calls' do
+        expect_no_offenses(<<~RUBY)
+          numblock.call(:arg) { _1 }
+        RUBY
+      end
+    end
+
     it 'register an offense for parens in method call without args' do
       trailing_whitespace = ' '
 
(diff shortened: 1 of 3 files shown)


dvandersluis (approved):

spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
```diff
@@ -404,6 +404,14 @@ def method_missing(name, ...)
       end
     end
 
+    context 'numeric blocks in 2.7', :ruby27 do
```
koic, line 407: That's my two cents. The canonical name "numbered parameters" instead of "numeric blocks" may be better.

```suggestion
    context 'numbered parameters in 2.7', :ruby27 do
```

koic:
Thank you, Genadi!

PR #9859, opened 2021-06-08
configuration: * seems to bind before `; escape it

danbernier:
Text rendered, before:

> This affects cops that have customisable paths: if the default is db/migrate/.rb, and the cop is enabled in db/migrate/.rubocop.yml, the path will need to be explicitly set as .rb, as the default will look for db/migrate/db/migrate/*.rb. This is unlikely to be what you wanted.

image

After:

> This affects cops that have customisable paths: if the default is db/migrate/*.rb, and the cop is enabled in db/migrate/.rubocop.yml, the path will need to be explicitly set as *.rb, as the default will look for db/migrate/db/migrate/*.rb. This is unlikely to be what you wanted.


dvandersluis (approved):
Great, thanks!

PR #9863, opened 2021-06-09
Fix orphaned changelog file

dvandersluis:
This changelog file ended up in the root directory instead of `changelog/` so it didn't get merged in.


bbatsov:
Nice catch!

PR #9883, opened 2021-06-20
[Fix #7592] Line end concatenation indentation cop

### LINKED ISSUE #7592: Layout/MultilineOperationIndentation is missing some cases
Layout/MultilineOperationIndentation is missing cases when `\` is used as string concatenation.

```ruby
# 1.rb
puts 'a' +
'b'

puts 'a' \
'b'

puts "#{1}" \
'b'

puts 'a' \
"#{2}"

puts "#{1}" \
"#{2}"
```

--------

## Expected behavior

All the offences are detected.

## Actual behavior

Only the first statement is detected as offensive.

```
1.rb:2:1: C: Layout/MultilineOperationIndentation: Align the operands of an expression spanning multiple lines.
'b'
^^^
```

```sh
rubocop --only Layout/MultilineOperationIndentation 1.rb
```

## RuboCop version

```
$ [bundle exec] rubocop -V
0.78.0 (using Parser 2.6.5.0, running on ruby 2.5.5 x86_64-darwin18)
```

jonas054:
The cop `Layout/MultilineOperationIndentation` doesn't view the backslash as an operator, which is correct. Therefore, a new cop is needed to inspect character literals broken up into multiple lines with backslashes at the ends of the lines.
    
The new cop deals exclusively with string literals, not any other constructs where backslashes might be used to "glue" lines together.


bbatsov:
Good work! One can argue that often a more suitable correction would be a heredoc, but the cop's definitely a step in the right direction (consistency).

bquorning:
This PR fixes part 2 of #6420, thank you 🥇 

I actually mentioned @jonas054 on that issue, because I knew he’d be a good candidate to tackle those problems. I don’t know if the other parts of that issue are possible to fix, given the limitations of Parser.

PR #9893, opened 2021-06-29
Allow the term "Mastercard"

coorasse:
The new cop marks the term "Mastercard" as offensive.

Although they should rename the company to Maincard, that's not going to happen soon.

This PR allows the term Mastercard by default.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2561,6 +2561,7 @@ Naming/InclusiveLanguage:
       AllowedRegex:
         - 'Master of None'
         - 'Master Boot Record'
+        - 'Mastercard'        
     slave:
       Suggestions: ['replica', 'secondary', 'follower']
 


darmenise:
image

😱

schmijos:
2021-06-29 at 09 18

bbatsov:
Hmm, this seems like a bug in the cop as I don't think that it should target `master` as part of a word. And yeah - the cop's a bit controversial, so we may end up disabling it by default, based on the feedback we receive from our users.

bbatsov:
//cc @tjwp - is looking for the problematic terms as part of another word a feature or bug from your perspective? To me it seems like a bug.

ShockwaveNN:
> And yeah - the cop's a bit controversial, so we may end up disabling it by default, based on the feedback we receive from our users.

You got mine :)

https://github.com/rubocop/rubocop/issues/9895

tjwp:
> //cc @tjwp - is looking for the problematic terms as part of another word a feature or bug from your perspective? To me it seems like a bug.

For my part this was intentional behavior and there was a test included to document that matching part of a word is expected: https://github.com/rubocop/rubocop/blob/master/spec/rubocop/cop/naming/inclusive_language_spec.rb#L24-L30

I think it makes sense to add an exclusion for `Mastercard`.

I'd also be onboard with disabling this cop by default and letting users provide the configuration that works for them.

bbatsov:
@tjwp Got it. Might be a good idea to add some option like `ExactMatch` or something along those lines.

bbatsov:
We've cut a new release with a more permissive configuration for the cop. It shouldn't be noisy for anyone at this point.

PR #9896, opened 2021-06-29
[Fix #9895] Disable `Naming/InclusiveLanguage` by default

### LINKED ISSUE #9895: Do we really need `Naming/InclusiveLanguage` enabled by default?
I think this cop should be disable by default, because there is a lot of github repos that still using `master` as default branch and any code that works with those repo will be flagged

## Actual behavior

1. Run rubocop on this pseudo-code

```
# frozen_string_literal: true

def repo_fetcher(repo, branch)
  p "I fetch #{repo} from #{branch}"
end

repo_fetcher('github.com/org/repo-with-master', 'master')
```
2. rubocop test.rb

```
Inspecting 1 file
C

Offenses:

test.rb:7:40: C: Naming/InclusiveLanguage: Consider replacing problematic term 'master' with 'main', 'primary', or 'leader'.
repo_fetcher('github.com/org/repo-with-master', 'master')
                                       ^^^^^^
test.rb:7:50: C: Naming/InclusiveLanguage: Consider replacing problematic term 'master' with 'main', 'primary', or 'leader'.
repo_fetcher('github.com/org/repo-with-master', 'master')
```

## RuboCop version

```
$ [bundle exec] rubocop -V
1.18.0 (using Parser 3.0.1.1, rubocop-ast 1.7.0, running on ruby 2.7.3 x86_64-linux)
```

koic:
Fixes #9895 and follow up https://github.com/rubocop/rubocop/pull/9893#issuecomment-870343243.

This PR disables `Naming/InclusiveLanguage` by default because it has an unexpectedly impact for many users who give feedback.

-----------------

* [ ] Added tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -100,19 +100,6 @@ Metrics/ModuleLength:
 Naming/InclusiveLanguage:
   Exclude:
     - lib/rubocop/cop/naming/inclusive_language.rb
-    - spec/rubocop/cop/naming/inclusive_language_spec.rb
-  FlaggedTerms:
-    whitelist:
-      Suggestions:
-        - allowlist
-    blacklist:
-      Suggestions:
-        - denylist
-    master:
-      AllowedRegex:
-        - 'blob/master/'
-        - 'master \(unreleased\)'
-        - !ruby/regexp "/ENV\\['RUBOCOP_VERSION'\\] == 'master'/"
 
 RSpec/FilePath:
   Exclude:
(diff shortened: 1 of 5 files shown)


bbatsov:
Either this or just disabling the checking of string literals, which I think is the real problem. I'm reasonably certain this won't generate much noise for identifier names and file names. 

In hindsight - it doesn't make much sense for a cop focused on naming to check string literals, as their contents are text, not names. I should have thought of this a bit earlier. :D

bbatsov:
We might also remove `master` from the defaults, although I doubt anyone names identiers `master`. :D

tjwp:
As the author of this cop, I want to endorse that I'm in support of disabling by default. I think this is a useful tool to put in user's hands but don't want to force it or a specific configuration on anyone.

koic:
@bbatsov Sure! I agree with your suggestions! I updated this PR to set `CheckStrings: false` and remove `master` from `FlaggedTerms` for `Naming/InclusiveLanguage`. I think these updates can reduce an unexpectedly impact for users much smaller. So, I reverted the cop's status to pending.

dvandersluis:
I agree with making the default configuration more permissive if this cop is going to be enabled by default. It'd be good to have an exhaustive example of how to configure the cop in the docs though so that people could, for example, just apply the original configuration if they are so inclined.

PR #9899, opened 2021-06-30
[Fix #9890] Make Colon After Comment Annotation Configurable

### LINKED ISSUE #9890: Make enforcement of colon after comment annotation configurable
If the Style/CommentAnnotation rule is enabled, RuboCop requires that the annotation name be followed by a colon. In both the IntelliJ IDEs and the Eclipse IDE, the convention is to omit the colon. (See https://www.jetbrains.com/help/idea/using-todo.html and https://www.vogella.com/tutorials/Eclipse/article.html#task-management). It's fair to say that omitting the colon is also a well-established convention (especially in mixed programming language teams). However, if we write comment annotations (aka task tags) this way, this RuboCop rule is not usable since it gets stuck on the missing colon (rather than the annotation names).

I wish to see an enforced style (or something of the sort) that allows the colon to be omitted. The rule can then still fulfill its primary function while accommodating this style variation.

What we are talking about is being able to reverse the bad and good cases in the cited section of the Ruby style guide: https://rubystyle.guide/#annotate-keywords

gregfletch:
[Fix #9890]

Make the separator for the Style/CommentAnnotation cop configurable. By default, the enforced style will be the current behaviour of annotation keyword followed by a colon (`:`), and a space. A new supported style is added to only require a space following the annotation keyword.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -3150,10 +3150,7 @@ Style/CommentAnnotation:
     - HACK
     - REVIEW
     - NOTE
-  EnforcedStyle: colon
-  SupportedStyles:
-    - colon
-    - space
+  RequireColon: true
 
 Style/CommentedKeyword:
   Description: 'Do not place comments on the same line as certain keywords.'
--- a/config/default.yml
+++ b/config/default.yml
@@ -3150,6 +3150,10 @@ Style/CommentAnnotation:
     - HACK
     - REVIEW
     - NOTE
+  EnforcedStyle: colon
+  SupportedStyles:
+    - colon
+    - space
 
 Style/CommentedKeyword:
   Description: 'Do not place comments on the same line as certain keywords.'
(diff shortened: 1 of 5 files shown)


config/default.yml
```diff
@@ -3150,6 +3150,10 @@ Style/CommentAnnotation:
     - HACK
     - REVIEW
     - NOTE
+  EnforcedStyle: colon
```
bbatsov, line 3153: I think we don't need an enforced style for this - we can just have an boolean options like `RequireColon: true`.

gregfletch: OK, my initial thinking here was to leave it a bit more open with enforced style/supported styles in case somebody wanted to have a different separator, like a dash for example (i.e. `TODO- some todo comment here`). But I'm fine changing it to a simple boolean value indicating if a colon is required if you think that is better.

bbatsov: Sometimes we should just keep things simple, so let's go with the boolean value.

bbatsov:
Thanks!

PR #9903, opened 2021-07-02
[Fix #9580] Add New Cop to Enforce Bundler Gem filename

### LINKED ISSUE #9580: A cop that enforces `Gemfile` vs `gems.rb`

I'm kicking the tires on changing our `Gemfile` to `gems.rb`. However, we have many internal gems without our monorepo and it just occurred to me that it would be cool to have a configurable cop that enforces one or the other.

A cop that checks to make sure you have a `Gemfile` or a `gems.rb` (and it counterpart lockfiles) depending on what you've configured. Maybe we default to the newer `gems.rb`?

gregfletch:
Add a new cop which enforces which bundler gem filename to use. By default, enforces Gemfile and its related Gemfile.lock file. Alternatively, setting RequiresGemfile to false enforces gems.rb and its related gems.locked file.

-----------------

* [x] Added tests.


lib/rubocop/cop/bundler/gem_filename.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Bundler
+      # This cop verifies that a project contains Gemfile or gems.rb file and correct
+      # associated lock file based on the configuration.
+      #
+      # @example RequiresGemfile: true (default)
```
bbatsov, line 9: I'd suggest "EnforcedStyle" with the options "Gemfile" and "gems.rb". To me at least, that's much easier to understand.

lib/rubocop/cop/bundler/gem_filename.rb
```diff
@@ -0,0 +1,90 @@
...
+      #
+      #   # bad
+      #   Project contains gems.rb and Gemfile.lock file
+      #
+      #   # good
+      #   Project contains gems.rb and gems.locked files
+      class GemFilename < Base
+        include RangeHelp
+
+        MSG_GEMFILE_REQUIRED = 'gems.rb file was found but Gemfile is required.'
```
bbatsov, line 31: Please add backticks around the names of the files, so they are properly colored in the terminal.

lib/rubocop/cop/bundler/gem_filename.rb
```diff
@@ -0,0 +1,96 @@
...
+        MSG_GEMS_RB_MISMATCHED = 'Expected a `gems.locked` file with `gems.rb` but found '\
+                                 '`Gemfile.lock`.'
+        GEMFILE_FILES = %w[Gemfile Gemfile.lock].freeze
+        GEMS_RB_FILES = %w[gems.rb gems.locked].freeze
+
+        def on_new_investigation
+          file_path = processed_source.file_path
+          return if expected_gemfile?(file_path)
+
+          register_offense(processed_source, file_path)
```
bbatsov, line 45: I see this function uses only the underlying buffer, so I guess you can pass it down directly.

lib/rubocop/cop/bundler/gem_filename.rb
```diff
@@ -0,0 +1,96 @@
...
+
+        def register_gems_rb_offense(processed_source, file_path)
+          message = case file_path
+                    when 'Gemfile'
+                      MSG_GEMS_RB_REQUIRED
+                    when 'Gemfile.lock'
+                      MSG_GEMS_RB_MISMATCHED
+                    end
+
+          return if message.nil?
```
bbatsov, line 76: I think it's a bit weird that this function might register an offense or not. Perhaps it has to be separated in two for clarity.

lib/rubocop/cop/bundler/gem_filename.rb
dvandersluis: Absolutely agree, and it's more consistent with other cops that way.

lib/rubocop/cop/bundler/gem_filename.rb
```diff
@@ -0,0 +1,100 @@
...
+
+        def register_gemfile_offense(source_buffer, file_path)
+          message = case file_path
+                    when 'gems.rb'
+                      MSG_GEMFILE_REQUIRED
+                    when 'gems.locked'
+                      MSG_GEMFILE_MISMATCHED
+                    end
+
+          add_offense(source_range(source_buffer, 1, 0), message: message)
```
dvandersluis, line 63: I believe you should be able to use `add_global_offense` instead of the two `add_offense` calls since the offense isn't really related to a line.

https://github.com/rubocop/rubocop/blob/07415559a9249e0230cb9d7382cbf148ba9227aa/lib/rubocop/cop/base.rb#L105-L112

config/default.yml
```diff
@@ -174,6 +174,15 @@ Bundler/GemComment:
...
   OnlyFor: []
 
+Bundler/GemFilename:
+  Description: 'Enforces the filename for managing gems.'
+  Enabled: true
+  VersionAdded: '<<next>>'
+  EnforcedStyle: 'Gemfile'
+  SupportedStyles:
+    - 'Gemfile'
+    - 'gems.rb'
```
dvandersluis, line 184: This configuration section should have an `Include` key to limit what's scanned by the cop (otherwise, it'll scan every single file, which is unperformant and might end up with false positives, for instance if someone has a `lib/gems.rb` file).

lib/rubocop/cop/bundler/gem_filename.rb
```diff
@@ -0,0 +1,100 @@
...
+                    when 'Gemfile.lock'
+                      MSG_GEMS_RB_MISMATCHED
+                    end
+
+          add_offense(source_range(source_buffer, 1, 0), message: message)
+        end
+
+        def gemfile_offense?(file_path)
+          gemfile_required? && %w[gems.rb gems.locked].include?(file_path)
+        end
```
dvandersluis, line 79:
```suggestion
        def gemfile_offense?(file_path)
          gemfile_required? && GEMS_RB_FILES.include?(file_path)
        end
```

lib/rubocop/cop/bundler/gem_filename.rb
```diff
@@ -0,0 +1,100 @@
...
+
+        def gems_rb_offense?(file_path)
+          gems_rb_required? && %w[Gemfile Gemfile.lock].include?(file_path)
+        end
```
dvandersluis, line 83:
```suggestion
        def gems_rb_offense?(file_path)
          gems_rb_required? && GEMFILE_FILES .include?(file_path)
        end
```

dvandersluis:
The other bundler cops handle files that aren't in the root directory, but this doesn't appear like it'll do so given how the file names are matched.

bbatsov:
@gregfletch Ping :-)

gregfletch:
> @gregfletch Ping :-)

Oops, completely forgot about this. Should have time by this weekend to fix up the remaining comments and finish this off. Sorry about that!

PR #9906, opened 2021-07-03
[Fix #8469] Add cop Layout/SpaceAroundSingletonClassOperator

### LINKED ISSUE #8469: New cop: Layout/SpaceAroundSingletonInheritence

One could write

```ruby
class<<self
```

which isn't as readable as adding spaces around the `<<` operator:

```ruby
class << self
```

There should be a layout cop that checks whether there are spaces around the `<<` operator for singleton class inheritence.

-

## Additional context

-

jonas054:
I chose the name `SpaceAroundSingletonClassOperator` rather than `SpaceAroundSingletonInheritence`, which was suggested in the feature request. Although I'm not sure that `<<` is actually an operator, I am pretty sure that inheritance is not involved. The operator (for lack of a better term) opens up the singleton class, or eigenclass, so methods can be added to it.


koic:
IMO, it would be better for `Layout/SpaceAroundOperators` to be aware of sclass `<<` operator rather than adding a new cop. Also, it seems that there is no need to make new `no_space` option for the operator.
https://github.com/rubocop/rubocop/blob/v1.18.2/lib/rubocop/cop/layout/space_around_operators.rb

jonas054:
I guess you're right. I'll add it to the existing cop.

jonas054:
I'm not sure what I need to do in order to stick to the semantic versioning rules. A new cop is marked as `pending` until the next major release of RuboCop, but what about breaking changes like this one, where we want to add more inspection duties to an existing cop? Ideally we'd want those changes to not come into effect until the next major release.

I'm not aware of any mechanism to make the feature pending. Is it enough to mark it as Breaking in the change log?

koic:
`Layout/SpaceAroundOperators`'s role is to check that operators have space around them. Therefore I think it is not a breaking change because it can be considered a false negative of `Layout/SpaceAroundOperators`.

jonas054:
Closing this PR to open a new one with the bug fix approach when I'm done.

PR #9910, opened 2021-07-04
[Fix #8469] Add singleton class operator inspection

### LINKED ISSUE #8469: New cop: Layout/SpaceAroundSingletonInheritence

jonas054:
We classify the `<<` operator used for the singleton class construct `class << self` as an operator like any other. This means that not inspecting it was a bug, and this change fixes it.


koic:
Looks good to me. Thank you Jonas!

PR #9911, opened 2021-07-04
[Fix #9905] Fix single line concatenation false positive

### LINKED ISSUE #9905: `Layout/LineEndStringConcatenationIndentation`: false positive offense when no backslashes used
The `Layout/LineEndStringConcatenationIndentation` cop shows a false positive offense even when no backslashes are used to concatenate strings, such as in the example below. (`rubocop -d` output is provided below in the template.)

```
➜ bat test.rb
───────┬──────────────────────────────────────────────────────────────────────────────────
       │ File: test.rb
───────┼──────────────────────────────────────────────────────────────────────────────────
   1   │ text = 'offense'
   2   │ puts 'This probably should not be '"an #{text}"
───────┴──────────────────────────────────────────────────────────────────────────────────
```

--------

## Expected behavior

I would expect the example provided above to not be an offense (with respect to the `Layout/LineEndStringConcatenationIndentation` cop).

## Actual behavior

An offense was shown.

```
➜ rubocop --only Layout/LineEndStringConcatenationIndentation -d -- test.rb
For /Users/username/Downloads: Default configuration from /Users/username/.gem/ruby/3.0.0/gems/rubocop-1.18.2/config/default.yml
Inspecting 1 file
Scanning /Users/username/Downloads/test.rb
Loading cache from /Users/username/.cache/rubocop_cache/df3a019f8c749a80df8ba46ae487e69edc99342c/43966e1a2abd4a0e4fde78de3cedb433c66f58dd/4c712b4f334e7a6b8113a19a6ecffeed0f23db43
C

Offenses:

test.rb:2:36: C: Layout/LineEndStringConcatenationIndentation: Align parts of a string concatenated with backslash.
puts 'This probably should not be '"an #{text}"
                                   ^^^^^^^^^^^^

Finished in 0.10224999999627471 seconds
```

1. Create a `test .rb` file similar to the example provided (with string concatenation that does not use backslashes to span multiple lines).
2. Run `rubocop --only Layout/LineEndStringConcatenationIndentation -d -- test.rb`.

## RuboCop version

```
➜ rubocop -V
1.18.2 (using Parser 3.0.1.1, rubocop-ast 1.7.0, running on ruby 3.0.1 x86_64-darwin20)
```

jonas054:
By checking that the inspected node is multiline, we should get rid of this false positive. Happens for code like
```ruby
    puts 'a'"#{b}"
```
We're also checking that none of the children is multiline. This is done to skip weird corner cases that are impossible to have a consistent alignment rule for.
    
It turns out that the new multiline checks make it possible to remove a couple of old conditions when checking if it's a concatenation with backslashes we're inspecting. It is perhaps possible to search for backslash and newline in the code, but I'd like to avoid that, because I'm afraid of other weird corner cases if we take the regexp search approach.

### DIFF
--- a/changelog/fix_line_end_string_concatenation_indentation.md
+++ b/changelog/fix_line_end_string_concatenation_indentation.md
@@ -1,0 +1,1 @@
+* [#9905](https://github.com/rubocop/rubocop/issues/9905): Fix false positive for single line concatenation in `Layout/LineEndStringConcatenationIndentation`. ([@jonas054][])
--- a/lib/rubocop/cop/layout/line_end_string_concatenation_indentation.rb
+++ b/lib/rubocop/cop/layout/line_end_string_concatenation_indentation.rb
@@ -85,14 +85,9 @@ module RuboCop
         private
 
         def strings_concatenated_with_backslash?(dstr_node)
-          !dstr_node.heredoc? &&
-            !single_string_literal?(dstr_node) &&
-            dstr_node.children.length > 1 &&
-            dstr_node.children.all? { |c| c.str_type? || c.dstr_type? }
-        end
-
-        def single_string_literal?(dstr_node)
-          dstr_node.loc.respond_to?(:begin) && dstr_node.loc.begin
+          dstr_node.multiline? &&
+            dstr_node.children.all? { |c| c.str_type? || c.dstr_type? } &&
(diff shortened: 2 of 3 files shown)


koic:
Thanks!

PR #9923, opened 2021-07-10
[Fix #9434] Fix false positive in FirstArgumentIndentation

### LINKED ISSUE #9434: Layout/AssignmentIndentation vs Layout/FirstArgumentIndentation confuse each other
We have (a bit unusual, I agree) this setting:
```yaml
Layout/AssignmentIndentation:
  IndentationWidth: 4
```
(The reasoning of person who introduced this is "not to mix assignments with new block starts").

Now, considering these two cases:
```ruby
# 1
foo =
    bar
# 2
foo.baz =
    bar
```
(1) is OK (4 spaces indentation). (2) is OK with this cop, but ...
```
C: [Correctable] Layout/FirstArgumentIndentation: Indent the first argument one step more than the start of the previous line.
    bar
    ^^^
```

And if you try to auto-correct it... 
```
rbcp.rb:10:3: C: [Corrected] Layout/AssignmentIndentation: Indent the first line of the right-hand-side of a multi-line assignment.
  bar
  ^^^
rbcp.rb:10:5: C: [Corrected] Layout/FirstArgumentIndentation: Indent the first argument one step more than the start of the previous line.
    bar
    ^^^

0 files inspected, 2 offenses detected, 2 offenses corrected
Infinite loop detected in /home/zverok/playground/ruby/rbcp.rb and caused by Layout/FirstArgumentIndentation -> Layout/AssignmentIndentation
```

I understand where it comes from (this statement is kinda both "assignment" and "method call"), but it seems to me that the "assignment" should take a priority here... And in any case, infinite loop seems the sub-optimal behavior :)

jonas054:
Setter calls are a kind of method invocation that `Layout/FirstArgumentIndentation` should just skip.

### DIFF
--- a/changelog/fix_first_argument_indentation_false_pos.md
+++ b/changelog/fix_first_argument_indentation_false_pos.md
@@ -1,0 +1,1 @@
+* [#9434](https://github.com/rubocop/rubocop/issues/9434): Fix false positive for setter calls in `Layout/FirstArgumentIndentation`. ([@jonas054][])
--- a/lib/rubocop/cop/layout/first_argument_indentation.rb
+++ b/lib/rubocop/cop/layout/first_argument_indentation.rb
@@ -154,7 +154,7 @@ module RuboCop
 
         def on_send(node)
           return if style != :consistent && enforce_first_argument_with_fixed_indentation?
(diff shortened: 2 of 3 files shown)


koic:
Thanks!

PR #9410, opened 2021-01-23
[Fix #9328] Recognize shareable_constant_value magic comment in Style/MutableConstant

caalberts:
Ruby 3.0 adds `shareable_constant_value` magic comment. This PR changes `Style/MutableConstant` to be aware of `shareable_constant_value` magic comment.

Fixes #9328.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop-hq/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

Issue #9328, opened 2021-01-03
`Style/MutableConstant` should be aware of `# shareable_constant_value` (Ruby 3.0)


marcandre:
The following should not generate an offense:

```ruby
# shareable_constant_value: literal (or experimental_everything)

MY_CONSTANT = %w[anything here is fine even non-literals]
```

rrosenblum:
Here's a link to some documentation on the feature, https://docs.ruby-lang.org/en/3.0.0/doc/syntax/comments_rdoc.html#label-shareable_constant_value+Directive. 

I hadn't heard of this being added before so I found it to be an interesting feature. It seems like the intent is to be used with Ractors

marcandre:
>  It seems like the intent is to be used with Ractors

That's one way to view it. The other way is the general position that constants should be immutable, and clearly when we say that we never mean actual classes. So in effect that is pretty much what "Ractor-shareable" means.

caalberts:
I created https://github.com/rubocop-hq/rubocop/pull/9410 to add shareable_constant_value methods to the `MagicComment` class.

PR #9929, opened 2021-07-13
[Fix rubocop#9328] Honour shareable_constant_value magic comment

thearjunmdas:
Ruby 3.0 supports [shareable_constant_value](https://docs.ruby-lang.org/en/3.0.0/doc/syntax/comments_rdoc.html#label-shareable_constant_value+Directive)  magic comment. This PR updates Style/MutableConstant to honour shareable_constant_value magic comment.

Fixes [#9328](https://github.com/rubocop/rubocop/issues/9328)

Took PR [9410](https://github.com/rubocop/rubocop/pull/9410) as reference. Tried to add @caalberts as co-author in changelog, but changelog spec fails.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_recognize_shareable_constant_value_magic.md
+++ b/changelog/change_recognize_shareable_constant_value_magic.md
@@ -1,0 +1,1 @@
+* [#9328](https://github.com/rubocop/rubocop/issues/9328): Recognize shareable_constant_value magic comment. ([@thearjunmdas][])
--- a/changelog/change_recognize_shareable_constant_value_magic.md
+++ b/changelog/change_recognize_shareable_constant_value_magic.md
@@ -1,1 +1,1 @@
-* [#9328](https://github.com/rubocop/rubocop/issues/9328): Recognize shareable_constant_value magic comment. ([@thearjunmdas][])
+* [#9328](https://github.com/rubocop/rubocop/issues/9328): Recognize shareable_constant_value magic comment. ([@thearjunmdas][], [@caalberts][])
--- a/lib/rubocop/cop/style/mutable_constant.rb
+++ b/lib/rubocop/cop/style/mutable_constant.rb
@@ -53,6 +53,43 @@ module RuboCop
       #     end
       #   end.freeze
       class MutableConstant < Base
+        # Handles magic comment shareable_constant_value with O(n ^ 2) complexity
+        # n - number of lines in the source
+        # Iterates over all lines before a CONSTANT
+        # until it reaches shareable_constant_value
+        module ShareableConstantValue
+          module_function
+
+          def recent_shareable_value?(node)
+            shareable_constant_comment = magic_comment_in_scope node
+            return false if shareable_constant_comment.nil?
+
+            shareable_constant_value = MagicComment.parse(shareable_constant_comment)
+                                                   .shareable_constant_value
+            shareable_constant_value_enabled? shareable_constant_value
+          end
+
+          # Identifies the most recent magic comment with valid shareable constant values
+          # thats in scope for this node
+          def magic_comment_in_scope(node)
+            processed_source_till_node(node).reverse_each.find do |line|
+              MagicComment.parse(line).valid_shareable_constant_value?
+            end
+          end
+
+          private
+
+          def processed_source_till_node(node)
+            processed_source.lines[0..(node.last_line - 1)]
+          end
+
+          def shareable_constant_value_enabled?(value)
+            %w[literal experimental_everything experimental_copy].include? value
+          end
+        end
+        private_constant :ShareableConstantValue
+
+        include ShareableConstantValue
         include FrozenStringLiteral
         include ConfigurableEnforcedStyle
         extend AutoCorrector
@@ -85,18 +122,18 @@ module RuboCop
           return if immutable_literal?(value)
           return if operation_produces_immutable_object?(value)
           return if frozen_string_literal?(value)
+          return if shareable_constant_value?(value)
 
           add_offense(value) { |corrector| autocorrect(corrector, value) }
         end
 
         def check(value)
           range_enclosed_in_parentheses = range_enclosed_in_parentheses?(value)
-
           return unless mutable_literal?(value) ||
                         target_ruby_version <= 2.7 && range_enclosed_in_parentheses
-
           return if FROZEN_STRING_LITERAL_TYPES.include?(value.type) &&
                     frozen_string_literals_enabled?
+          return if shareable_constant_value?(value)
 
           add_offense(value) { |corrector| autocorrect(corrector, value) }
         end
@@ -128,6 +165,12 @@ module RuboCop
 
         def frozen_string_literal?(node)
           FROZEN_STRING_LITERAL_TYPES.include?(node.type) && frozen_string_literals_enabled?
+        end
+
+        def shareable_constant_value?(node)
+          return false if target_ruby_version < 3.0
+
+          recent_shareable_value? node
         end
 
         def frozen_regexp_or_range_literals?(node)
--- a/lib/rubocop/cop/style/mutable_constant.rb
+++ b/lib/rubocop/cop/style/mutable_constant.rb
@@ -14,6 +14,11 @@ module RuboCop
       # positives. Luckily, there is no harm in freezing an already
       # frozen object.
       #
+      # From Ruby 3.0, this cop honours the magic comment
+      # 'shareable_constant_value'. When this magic comment is set to any
+      # acceptable value other than none, it will suppress the offenses
+      # raised by this cop. It enforces frozen state.
+      #
       # NOTE: Regexp and Range literals are frozen objects since Ruby 3.0.
       #
       # @example EnforcedStyle: literals (default)
--- a/lib/rubocop/cop/style/mutable_constant.rb
+++ b/lib/rubocop/cop/style/mutable_constant.rb
@@ -57,6 +57,21 @@ module RuboCop
       #       puts 1
       #     end
       #   end.freeze
+      #
+      # @example
+      #   # Magic comment - shareable_constant_value: literal
+      #
+      #   # bad
+      #   CONST = [1, 2, 3]
+      #
+      #   # good
+      #   # shareable_constant_value: literal
+      #   CONST = [1, 2, 3]
+      #
+      # NOTE: This special directive helps to create constants
+      # that hold only immutable objects, or Ractor-shareable
+      # constants. - ruby docs
+      #
       class MutableConstant < Base
         # Handles magic comment shareable_constant_value with O(n ^ 2) complexity
         # n - number of lines in the source
--- a/spec/rubocop/cop/style/mutable_constant_spec.rb
+++ b/spec/rubocop/cop/style/mutable_constant_spec.rb
@@ -43,6 +43,43 @@ RSpec.describe RuboCop::Cop::Style::MutableConstant, :config do
     end
   end
 
+  shared_examples 'literals that are frozen' do |o|
+    let(:prefix) { o }
+
+    it_behaves_like 'immutable objects', '[1, 2, 3]'
+    it_behaves_like 'immutable objects', '%w(a b c)'
+    it_behaves_like 'immutable objects', '{ a: 1, b: 2 }'
+    it_behaves_like 'immutable objects', "'str'"
+    it_behaves_like 'immutable objects', '"top#{1 + 2}"'
+    it_behaves_like 'immutable objects', '1'
+    it_behaves_like 'immutable objects', '1.5'
+    it_behaves_like 'immutable objects', ':sym'
+    it_behaves_like 'immutable objects', 'FOO + BAR'
+    it_behaves_like 'immutable objects', 'FOO - BAR'
+    it_behaves_like 'immutable objects', "'foo' + 'bar'"
+    it_behaves_like 'immutable objects', "ENV['foo']"
+    it_behaves_like 'immutable objects', "::ENV['foo']"
+  end
+
+  shared_examples 'literals that are not frozen' do |o|
+    let(:prefix) { o }
+
+    it_behaves_like 'mutable objects', '[1, 2, 3]'
+    it_behaves_like 'mutable objects', '%w(a b c)'
+    it_behaves_like 'mutable objects', '{ a: 1, b: 2 }'
+    it_behaves_like 'mutable objects', "'str'"
+    it_behaves_like 'mutable objects', '"top#{1 + 2}"'
+
+    it_behaves_like 'immutable objects', '1'
+    it_behaves_like 'immutable objects', '1.5'
+    it_behaves_like 'immutable objects', ':sym'
+    it_behaves_like 'immutable objects', 'FOO + BAR'
+    it_behaves_like 'immutable objects', 'FOO - BAR'
+    it_behaves_like 'immutable objects', "'foo' + 'bar'"
+    it_behaves_like 'immutable objects', "ENV['foo']"
+    it_behaves_like 'immutable objects', "::ENV['foo']"
+  end
+
   context 'Strict: false' do
     let(:cop_config) { { 'EnforcedStyle' => 'literals' } }
 
@@ -157,6 +194,34 @@ RSpec.describe RuboCop::Cop::Style::MutableConstant, :config do
           RUBY
         end
       end
+
+      context 'when using shareable_constant_value' do
+        it_behaves_like 'literals that are frozen', '# shareable_constant_value: literal'
+        it_behaves_like 'literals that are frozen', '# shareable_constant_value: experimental_everything'
+        it_behaves_like 'literals that are frozen', '# shareable_constant_value: experimental_copy'
+        it_behaves_like 'literals that are not frozen', '# shareable_constant_value: none'
+      end
+
+      it 'raises offense when shareable_constant_value is specified as an inline comment' do
+        expect_offense(<<~RUBY)
+          X = [1, 2, 3] # shareable_constant_value: literal
+              ^^^^^^^^^ Freeze mutable objects assigned to constants.
+          Y = [4, 5, 6]
+              ^^^^^^^^^ Freeze mutable objects assigned to constants.
+        RUBY
+      end
+
+      it 'raises offense only for shareable_constant_value as none when set in the order of: literal, none and experimental_everything' do
+        expect_offense(<<~RUBY)
+          # shareable_constant_value: literal
+          X = [1, 2, 3]
+          # shareable_constant_value: none
+          Y = [4, 5, 6]
+              ^^^^^^^^^ Freeze mutable objects assigned to constants.
+          # shareable_constant_value: experimental_everything
+          Z = [7, 8, 9]
+        RUBY
+      end
     end
 
     context 'Ruby 2.7 or lower', :ruby27 do
@@ -219,6 +284,13 @@ RSpec.describe RuboCop::Cop::Style::MutableConstant, :config do
             XXX = (1...99).freeze
           RUBY
         end
+      end
+
+      context 'when using shareable_constant_values' do
+        it_behaves_like 'literals that are not frozen', '# shareable_constant_value: literal'
+        it_behaves_like 'literals that are not frozen', '# shareable_constant_value: experimental_everything'
+        it_behaves_like 'literals that are not frozen', '# shareable_constant_value: experimental_copy'
+        it_behaves_like 'literals that are not frozen', '# shareable_constant_value: none'
       end
     end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5812,3 +5812,4 @@
 [@AirWick219]: https://github.com/AirWick219
 [@markburns]: https://github.com/markburns
 [@gregfletch]: https://github.com/gregfletch
+[@thearjunmdas]: https://github.com/thearjunmdas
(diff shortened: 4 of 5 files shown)


thearjunmdas:
@marcandre  - Could you please take a look when free?

spec/rubocop/cop/style/mutable_constant_spec.rb
```diff
@@ -157,6 +157,123 @@
...
+        it_behaves_like 'immutable objects', "'str'"
+        it_behaves_like 'immutable objects', '"top#{1 + 2}"'
+        it_behaves_like 'immutable objects', '1'
+        it_behaves_like 'immutable objects', '1.5'
+        it_behaves_like 'immutable objects', ':sym'
+        it_behaves_like 'immutable objects', 'FOO + BAR'
+        it_behaves_like 'immutable objects', 'FOO - BAR'
+        it_behaves_like 'immutable objects', "'foo' + 'bar'"
+        it_behaves_like 'immutable objects', "ENV['foo']"
+        it_behaves_like 'immutable objects', "::ENV['foo']"
```
marcandre, line 176: Could these be grouped together and reused instead of copied?

spec/rubocop/cop/style/mutable_constant_spec.rb
```diff
@@ -157,6 +157,123 @@
...
+      it 'raises offense only for shareable_constant_value as none when set in the order of: literal, none and experimental_everything' do
+        expect_offense(<<~RUBY)
+          # shareable_constant_value: literal
+          X = [1, 2, 3]
+          # shareable_constant_value: none
+          Y = [4, 5, 6]
+              ^^^^^^^^^ Freeze mutable objects assigned to constants.
+          # shareable_constant_value: experimental_everything
+          Z = [7, 8, 9]
+        RUBY
```
marcandre, line 244: This test seems sufficient to me, all the others look like variations.

lib/rubocop/cop/mixin/shareable_constant_value.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # Common functionality for dealing with Shareable Constant Value
+    module ShareableConstantValue
+      module_function
+
+      def recent_shareable_value?(node)
```
marcandre, line 9: This methods has bad performance (`O(n)`) so overall this cop could be `O(n^2)`.
Ideally fix the performance of this method, or else make sure that it is called *last*, i.e. only when it really looks like there's a `.freeze` missing.

lib/rubocop/cop/style/mutable_constant.rb
```diff
@@ -132,6 +134,13 @@ def frozen_string_literal?(node)
           FROZEN_STRING_LITERAL_TYPES.include?(node.type) && frozen_string_literals_enabled?
         end
 
+        def shareable_constant_value?(node)
+          return false if target_ruby_version < 3.0
+          return false if node.nil?
```
marcandre, line 139: Is there a valid reason for this method to ever be called with `node == nil`?

marcandre (commented):
Tests that could be added

1)
```ruby
# inline comments:
X = "x" # shareable_constant_value: literal
```

2)
```ruby
# shareable_constant_value: literal
X = Set[1, 2]
```

3)
```ruby
# blabla shareable_constant_value: literal blabla
```

4) variations with hyphen

marcandre:
Good PR @thearjunmdas ! 👍
I left a few comments for you

thearjunmdas:
> Good PR @thearjunmdas ! 👍
> I left a few comments for you

Thanks a lot for the quick review. Will address the comments and get back.

lib/rubocop/cop/mixin/shareable_constant_value.rb
thearjunmdas: For now called this as last line to make the best effort on performance.

I believe the only way to make this method performant is to memoize, rather store the last instance shareable_constant_value was provided. Is there any place in the codebase where such memoization is already being done? I could refer that raise a future PR to improve the performance of this method

lib/rubocop/cop/style/mutable_constant.rb
thearjunmdas: All the scenarios like the following https://github.com/rubocop/rubocop/blob/master/spec/rubocop/cop/style/mutable_constant_spec.rb#L28 passes a nil node to this

spec/rubocop/cop/style/mutable_constant_spec.rb
thearjunmdas: Agreed. Refactored.

spec/rubocop/cop/style/mutable_constant_spec.rb
thearjunmdas: Thanks for the comment. Refactored.

thearjunmdas:
https://github.com/rubocop/rubocop/pull/9929#pullrequestreview-705204492 - Added all the test cases mentioned here.
Addressed all other review comments @marcandre. I am new to RSpec, please check if the way I have refactored is okay.

Please take a look and feel free to add more comments if we could make it better.

lib/rubocop/cop/mixin/shareable_constant_value.rb
marcandre: You can use normal memoization (`@cache ||= ...`). To be clean, you can clear the cache in `def on_new_investigation`, but it's not necessary unless your cop `support_multiple_source?`.

spec/rubocop/cop/style/mutable_constant_spec.rb
```diff
@@ -43,6 +43,45 @@
     end
   end
 
+  shared_examples 'freezing all constants' do |o|
```
marcandre, line 46: Using `shared_examples` is the right way to go 👍
OTOH, the label or grouping could be improved. "not freezing all constants" is not clear. Something like "enforces mutable literals are frozen" seems more precise, for example.

spec/rubocop/cop/style/mutable_constant_spec.rb
```diff
@@ -157,6 +196,32 @@
...
+        it_behaves_like 'freezing all constants', '# shareable_constant_value: experimental_everything'
+        it_behaves_like 'freezing all constants', '# shareable_constant_value: experimental_copy'
+        it_behaves_like 'not freezing all constants', '# shareable_constant_value: none'
+      end
+
+      it 'raises offense when shareable_constant_value is specified as an inline comment' do
+        expect_offense(<<~RUBY)
+          X = [1, 2, 3] # shareable_constant_value: literal
+              ^^^^^^^^^ Freeze mutable objects assigned to constants.
+        RUBY
```
marcandre, line 211: Did you verify what Ruby does for inline comments? How does it affect subsequent constants? I'm not sure what it does, and [the doc](https://docs.ruby-lang.org/en/3.0.0/doc/syntax/comments_rdoc.html#label-Scope) doesn't say. Do you feel like opening a bug report or proposing a PR for Ruby about this, or should I do it?
Also, the comments are actually scoped; current code does not take that into account. This limitation should be documented.

lib/rubocop/cop/style/mutable_constant.rb
marcandre: Oh, I see. That's a mistake, fixed in #9931

spec/rubocop/cop/style/mutable_constant_spec.rb
thearjunmdas: yes @marcandre - verified in my local. The above test is  the observed behaviour.
``` ruby
A = %w[anything here is fine even non-literals] # shareable_constant_value: literal 
puts A.frozen? # prints false
```

I can open a pr in ruby for the docs. And yes, that was a good catch, i first went to the docs after your review comment but not help there.

marcandre: Could you test how it affects constants that follow please (both in Ruby and in this PR)?

thearjunmdas: Verified it in ruby 3.0 and updated the test in PR. 
**inline comment is not recognised.**

spec/rubocop/cop/style/mutable_constant_spec.rb
```diff
@@ -43,6 +43,43 @@
...
+    it_behaves_like 'immutable objects', '1'
+    it_behaves_like 'immutable objects', '1.5'
+    it_behaves_like 'immutable objects', ':sym'
+    it_behaves_like 'immutable objects', 'FOO + BAR'
+    it_behaves_like 'immutable objects', 'FOO - BAR'
+    it_behaves_like 'immutable objects', "'foo' + 'bar'"
+    it_behaves_like 'immutable objects', "ENV['foo']"
+    it_behaves_like 'immutable objects', "::ENV['foo']"
+  end
+
```
thearjunmdas, line 82: @marcandre - Made a change here. Was reviewing code and found that 
```ruby
# shareable_constant_value: literal
C = Set[2, 3] # Raises this error "cannot assign unshareable object to C (Ractor::IsolationError)"
```

I think we would need to write a separate cop for shareable_constant_value to handle the real purpose of shareable constant - ractor safe objects.

spec/rubocop/cop/style/mutable_constant_spec.rb
thearjunmdas: I was trying to make it a readable sentence with **it_behaves_like**. Please do check if this makes sense
_it_behaves_like 'literals that are frozen'_
Also mentioning literals since i would like to include both mutable and immutable literals.

spec/rubocop/cop/style/mutable_constant_spec.rb
thearjunmdas: In addition, I was trying to raise PR for this in ruby documentation - any tips on which repo and where I should go about this? https://github.com/ruby/ruby/blob/master/doc/contributing.rdoc is not very clear where to contribute for docs. https://github.com/ruby/docs.ruby-lang.org - this led me no where as well.

marcandre: I guess it's ok; magic comments need always be alone on a line

spec/rubocop/cop/style/mutable_constant_spec.rb
marcandre: Yes, `Set[]` is not (yet) considered like a "builtin" literal, so it is normal it behaves differently from Array/Hash/...

spec/rubocop/cop/style/mutable_constant_spec.rb
```diff
@@ -43,6 +43,43 @@
...
+    it_behaves_like 'immutable objects', "'str'"
+    it_behaves_like 'immutable objects', '"top#{1 + 2}"'
+    it_behaves_like 'immutable objects', '1'
+    it_behaves_like 'immutable objects', '1.5'
+    it_behaves_like 'immutable objects', ':sym'
+    it_behaves_like 'immutable objects', 'FOO + BAR'
+    it_behaves_like 'immutable objects', 'FOO - BAR'
+    it_behaves_like 'immutable objects', "'foo' + 'bar'"
+    it_behaves_like 'immutable objects', "ENV['foo']"
+    it_behaves_like 'immutable objects', "::ENV['foo']"
```
marcandre, line 61: I don't understand these. `X = FOO + BAR` should raise an offense (unless `experimental_*`), as they won't be frozen in general, e.g.:

```ruby
FOO = [1, 2, 3].freeze
BAR = [1, 2, 3].freeze
X = FOO + BAR
X.frozen? # => false
```

spec/rubocop/cop/style/mutable_constant_spec.rb
marcandre: There are three categories of values assigned to constants:
1) immutable literals (e.g. `42`, `:symbol`)
2) mutable builtin literals (e.g. `[1, 2, 3]`)
3) general expressions (e.g. `FOO + BAR`, `Set[1, 2, 3]`)

The cop should raise for:
1) never
2) for `none`
3) for `literal` or `none`

I'd like the cop and the specs to reflect this.

Maybe something like "This cop behaves_like:
1) "it_considers_immutable_literals_frozen"
2) "it_considers_builtin_literals_frozen"
3) "it_considers_expressions_in_constants_frozen"

thearjunmdas: Thanks for the detailed comments. I guessed I rushed through the tests a bit, will take a deeper look into this and come back.

spec/rubocop/cop/style/mutable_constant_spec.rb
thearjunmdas: Yes @marcandre - I guess you are right,  tt should raise an offense. Looks like, in the current behaviour it does not. Can you check the following tests:
https://github.com/rubocop/rubocop/blob/master/spec/rubocop/cop/style/mutable_constant_spec.rb#L58

spec/rubocop/cop/style/mutable_constant_spec.rb
thearjunmdas:
```ruby
# When EnforcedStyle: literal
# +-------------------+--------------------------------+----------------+
# |                   | Cop offense raised for shareable_constant_value |
# |   Values          |------+---------+---------------+----------------+
# |                   | none | literal | experimental_ | experimental_  |
# |                   |      |         | everything    | copy           |
# +-------------------+------+---------+---------------+----------------+
# |Immutable literals |false |false    |false          |false           |
# |Examples:          |      |         |               |                |
# |42, :symbol        |      |         |               |                |
# +-------------------+------+---------+---------------+----------------+
# |Mutable literals   |true  |false    |false          |false           |
# |Examples:          |      |         |               |                |
# |[1, 2, 3, 4]       |      |         |               |                |
# +-------------------+------+---------+---------------+----------------+
# |General expressions|true  |true     |false          |false           |
# |Examples:          |      |         |               |                |
# |FOO + BAR          |      |         |               |                |
# +-------------------+------+---------+---------------+----------------+
# |Frozen expressions |      |         |               |                |
# |Examples:          |false | false   | false         | false          |
# |(anything).freeze  |      |         |               |                |
# +-------------------+------+---------+---------------+----------------+

# When EnforcedStyle: Strict, Only the following row changes
# +-------------------+------+---------+---------------+----------------+
# |General Expressions|true  |false    |false          |false           |
# |                   |      |         |               |                |
# |Examples: FOO + BAR|      |         |               |                |
# +-------------------+------+---------+---------------+----------------+
```

I am planning to add this comment - Did i summarise correctly? I will go ahead once you are okay with above @marcandre.

spec/rubocop/cop/style/mutable_constant_spec.rb
thearjunmdas: Found the issue - 
The mutable constant cop relies on [unless value.mutable_literal?](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/style/mutable_constant.rb#L122) whose definition is [here](https://github.com/rubocop/rubocop-ast/blob/master/lib/rubocop/ast/node.rb#L361). This is actually not used properly. 

@marcandre  - I think we should create another bug for this and solve it for this cop.

thearjunmdas:
Have added some comments, please review them - then I can proceed with the changes.

spec/rubocop/cop/style/mutable_constant_spec.rb
marcandre: Oh, right. It's ok to not change the current behaviour then, and yes, it could be discussed elsewhere 👍

spec/rubocop/cop/style/mutable_constant_spec.rb
marcandre: My observation was more about naming your shared behaviours. If it's not clear to you, we can go ahead as is and eventually refactor.

lib/rubocop/cop/mixin/shareable_constant_value.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # Common functionality for dealing with Shareable Constant Value
+    module ShareableConstantValue
```
marcandre, line 6: Could you please:
1) Add a comment about performance being bad
2) Move this the the cop, in a private constant? I would rather not expose this code as it is.

lib/rubocop/cop/style/mutable_constant.rb
```diff
@@ -53,6 +53,43 @@ module Style
...
+
+          def processed_source_till_node(node)
+            processed_source.lines[0..(node.last_line - 1)]
+          end
+
+          def shareable_constant_value_enabled?(value)
+            %w[literal experimental_everything experimental_copy].include? value
+          end
+        end
+        private_constant :ShareableConstantValue
```
thearjunmdas, line 90: @marcandre  - Please check this moved the non-performant module internal to the class and marked as private_constant.

spec/rubocop/cop/style/mutable_constant_spec.rb
thearjunmdas: I tried refactoring by breaking it upto smaller chunks - Dint work out.

For now changed the title a bit, will try to refactor this over time.

lib/rubocop/cop/mixin/shareable_constant_value.rb
thearjunmdas: Moved this to the cop.

thearjunmdas:
Please do take a look, @marcandre

thearjunmdas:
ping @marcandre

thearjunmdas:
@koic @marcandre  - can you folks please take a look? lets try to merge this before it becomes stale?

changelog/change_recognize_shareable_constant_value_magic.md
```diff
@@ -0,0 +1 @@
+* [#9328](https://github.com/rubocop/rubocop/issues/9328): Recognize shareable_constant_value magic comment. ([@thearjunmdas][])
```
dvandersluis, line 1: If you wanted to add a second person here, you could do something like this:

```suggestion
* [#9328](https://github.com/rubocop/rubocop/issues/9328): Recognize shareable_constant_value magic comment. ([@thearjunmdas][], [@caalberts][])
```

bbatsov:
@thearjunmdas Please, also document the support for this magic comment in the cop's description, so people would know what to expect.

bbatsov:
You'll also have to rebase on top of `master`.

marcandre:
Oh, so sorry I missed the pings 😅

lib/rubocop/cop/style/mutable_constant.rb
marcandre: 👍

marcandre (approved):

thearjunmdas:
Thanks for reviving the PR @marcandre and @bbatsov. I will update  on the comments.

P.S. Cleaning the commit history - squashing my commits, rebasing - please ignore intermittent commit history change. Will clean it up.

changelog/change_recognize_shareable_constant_value_magic.md
thearjunmdas: @dvandersluis thanks for the tip. Worked perfectly.

thearjunmdas:
@bbatsov @marcandre rebased on top of master, added documentation. Please let me know, if anything else needs to be done.

lib/rubocop/cop/style/mutable_constant.rb
```diff
@@ -14,6 +14,11 @@ module Style
       # positives. Luckily, there is no harm in freezing an already
       # frozen object.
       #
+      # From Ruby 3.0, this cop honours the magic comment
```
bbatsov, line 17: Please, also add a couple of code snippets with `shareable_constant_value` in the cop examples.

spec/rubocop/cop/style/mutable_constant_spec.rb
```diff
@@ -157,6 +194,34 @@
...
+      context 'when using shareable_constant_value' do
+        it_behaves_like 'literals that are frozen', '# shareable_constant_value: literal'
+        it_behaves_like 'literals that are frozen', '# shareable_constant_value: experimental_everything'
+        it_behaves_like 'literals that are frozen', '# shareable_constant_value: experimental_copy'
+        it_behaves_like 'literals that are not frozen', '# shareable_constant_value: none'
+      end
+
+      it 'raises offense when shareable_constant_value is specified as an inline comment' do
+        expect_offense(<<~RUBY)
+          X = [1, 2, 3] # shareable_constant_value: literal
```
bbatsov, line 207: At some point we'll need a lint cop to check for this. :D

lib/rubocop/cop/style/mutable_constant.rb
```diff
@@ -52,7 +57,59 @@ module Style
...
+      #   CONST = [1, 2, 3]
+      #
+      #   # good
+      #   # shareable_constant_value: literal
+      #   CONST = [1, 2, 3]
+      #
+      # NOTE: This special directive helps to create constants
+      # that hold only immutable objects, or Ractor-shareable
+      # constants. - ruby docs
+      #
```
thearjunmdas, line 74: @bbatsov - Added some examples. Also added an excerpt from ruby docs so that shareable_constant_value doesn't turn out to be an alternative to **.freeze**

spec/rubocop/cop/style/mutable_constant_spec.rb
thearjunmdas: Makes, sense. Since it's an experimental feature, maybe we need not do it right away.

thearjunmdas:
@bbatsov  - please take  a look.

bbatsov:
Looks good. Thanks!

PR #9933, opened 2021-07-15
Fixup GitHub Actions formatter when run in non-default directory

ojab:
In some cases (specifically when running inside the container with
non-default `working-directory`) GitHub cannot link absolute paths
produced by formatter to the code, while relative [to the project root] 
paths works.

Output relative paths if they're inside current directory, so this
usecase is covered.

I can craft a public repo with reproducer, since I see this issue in private repo, but hopefully you believe me ._.
I doubt that we should care about usecases outside of GHA, so this PR could break reporting only if project is checked out into non-default directory and `rubocop` runs from some outer directory (i. e. not from the project root), but such cases should be trivially fixable by changing `working-directory`. 

In the case in question it's unfixable or non-trivially fixable because GHA message<->code matching rules are non-public.

-----------------

* n/a Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Added tests.

### DIFF
--- a/changelog/fix_github_actions_in_non_default_directory.md
+++ b/changelog/fix_github_actions_in_non_default_directory.md
@@ -1,0 +1,1 @@
+* [#9933](https://github.com/rubocop/rubocop/pull/9933): Fix GitHub Actions formatter when running in non-default directory. ([@ojab][])
--- a/lib/rubocop/formatter/git_hub_actions_formatter.rb
+++ b/lib/rubocop/formatter/git_hub_actions_formatter.rb
@@ -33,7 +33,7 @@ module RuboCop
         output.printf(
           "\n::%<severity>s file=%<file>s,line=%<line>d,col=%<column>d::%<message>s\n",
           severity: github_severity(offense),
-          file: file,
+          file: PathUtil.smart_path(file),
           line: offense.line,
           column: offense.real_column,
           message: github_escape(offense.message)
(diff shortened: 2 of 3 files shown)


ojab:
And CircleCI forces me to login and give RW access to all repos (or only private repos, if I click on non-default login option)
image
to see the CI failures :/
Not really nice.

ojab:
GHA failures seems unrelated, hopefully CircleCI are the same, sending to review.

changelog/fix_github_actions_in_non_default_directory.md
```diff
@@ -0,0 +1 @@
+* [#9933](https://github.com/rubocop/rubocop/pull/) Fix GitHub Actions formatter when running in non-default directory ([@ojab][])
```
koic, line 1: `% bundle exec rspec spec/project_spec.rb` has failure. Can you update to the below?

```diff
-* [#9933](https://github.com/rubocop/rubocop/pull/) Fix GitHub Actions formatter when running in non-default directory ([@ojab][])
+* [#9933](https://github.com/rubocop/rubocop/pull/9933): Fix GitHub Actions formatter when running in non-default directory. ([@ojab][])
```

ojab: :facepalm: pushed.

ojab:
Ok, CI is green, but I swear that failures looked unrelated.

koic:
Thanks!

PR #9936, opened 2021-07-16
[Fix #9752] Improve error message for top level department

### LINKED ISSUE #9752: department-level enable/disable does not work
**UPDATE**: On further investigation, department-level enable/disable just doesn't work.

The docs [say this should work works](https://docs.rubocop.org/rubocop/1.13/configuration.html):

```
All cops are then disabled by default. Only cops appearing in user configuration files are enabled. Enabled: true does not have to be set for cops in user configuration. They will be enabled anyway. It is also possible to enable entire departments by adding for example

Style:
  Enabled: true
```

But it turns out this doesn't work at all. This is because `Style` is seen as an "invalid" config in Config::Validator.validate() and thus ignored. :/

--------

# Original ticket (for context)

This is probably not a rubocop bug, but I dunno where else to ask.

In the [cookstyle](https://github.com/chef/cookstyle/) project, we just sit ontop of the awesomeness that is Rubocop, and define a bunch of our own Cops and departments.

However, departments defined by our own cops don't seem to be enable-able the way, say `Style` is. So I can do this:

```yaml
AllCops:
  DisabledByDefault: true

Style:
  Enabled: true
```

But what I can't do is (`Chef/Correctness` here is a department, you can see code for cops in that department [here](https://github.com/chef/cookstyle/tree/master/lib/rubocop/cop/chef/correctness)).

```yaml
AllCops:
  DisabledByDefault: true

Chef/Correctness:
  Enabled: true
```

or even:
```yaml
AllCops:
  DisabledByDefault: true

Chef:
  Enabled: true
```

What's really strange is that in robocop I don't see any special setup for the `Style` or `Layout` etc _modules_, so I would think this would "just work"... but it doesn't seem to. When I we try we get:

```
Warning: Chef/Correctness does not support Enabled parameter.

Supported parameters are: 

  - StyleGuideBaseURL
```

If I do just `Chef`, then I get:

```
Error: unrecognized cop Chef found in .cookstyle.yml
```

I'm sure is some minor dumb thing we're doing or not doing that's preventing this from working like the rubocop-built-in departments.

Any help would be greatly appreciated.

jonas054:
This is a smaller change that the one suggested in #9752. Instead of allowing to use top level departments in configuration, I just made a better error message. I think this change has a lower risk of causing new problems down the line.

When there are cop departments that seem to have a structure, like in the cookstyle gem where departments are named `Chef/Style`, `Chef/Correctness`, etc., it's easy to assume that there is a `Chef` department that can disable all departments under it. This is not the case, but the error message `Error: unrecognized cop Chef found in .cookstyle.yml` is not informative enough.

If an unrecognized cop or department is the first part of existing departments, we list the existing departments. This should help the user realize what needs to be changed. Example:
```
Error: unrecognized cop or department Chef found in .rubocop.yml
Chef is not a department. Use `Chef/Style`, `Chef/Deprecations`, `Chef/Sharing`, `Chef/Correctness`, `Chef/Modernize`, `Chef/Effortless`, `Chef/RedundantCode`.
```

Also, we change the wording of `unrecognized cop` to `unrecognized cop or department`, and extend the did-you-mean functionality to also look for similar department names.

### DIFF
--- a/changelog/fix_error_message_for_dept.md
+++ b/changelog/fix_error_message_for_dept.md
@@ -1,0 +1,1 @@
+* [#9752](https://github.com/rubocop/rubocop/issues/9752): Improve error message for top level department used in configuration. ([@jonas054][])
--- a/lib/rubocop/config_validator.rb
+++ b/lib/rubocop/config_validator.rb
@@ -104,17 +104,30 @@ module RuboCop
         # to do so than to pass the value around to various methods.
         next if name == 'inherit_mode'
 
-        suggestions = NameSimilarity.find_similar_names(name, Cop::Registry.global.map(&:cop_name))
-        suggestion = "Did you mean `#{suggestions.join('`, `')}`?" if suggestions.any?
-
         message = <<~MESSAGE.rstrip
-          unrecognized cop #{name} found in #{smart_loaded_path}
-          #{suggestion}
+          unrecognized cop or department #{name} found in #{smart_loaded_path}
+          #{suggestion(name)}
         MESSAGE
 
         unknown_cops << message
       end
       raise ValidationError, unknown_cops.join("\n") if unknown_cops.any?
+    end
+
+    def suggestion(name)
+      registry = Cop::Registry.global
+      departments = registry.departments.map(&:to_s)
+      suggestions = NameSimilarity.find_similar_names(name, departments + registry.map(&:cop_name))
+      if suggestions.any?
+        "Did you mean `#{suggestions.join('`, `')}`?"
+      else
+        # Department names can contain slashes, e.g. Chef/Correctness, but there's no support for
+        # the concept of higher level departments in RuboCop. It's a flat structure. So if the user
+        # tries to configure a "top level department", we hint that it's the bottom level
+        # departments that should be configured.
+        suggestions = departments.select { |department| department.start_with?("#{name}/") }
+        "#{name} is not a department. Use `#{suggestions.join('`, `')}`." if suggestions.any?
+      end
     end
 
     def validate_syntax_cop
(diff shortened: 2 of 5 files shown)


bbatsov:
I love this change! Simple, but effective!

jaymzh:
This is great, thanks!

However, I really want to be able to disable all chef departments sometimes without having to keep up with the list of all of them, which change over time. So I can think of two approaches:

1. Enable nested departments so that users could disable either `Chef` or `Chef/Style`
2. Enable Regex disabling so I can do `^Chef/.*`

I think the former is cleaner and provides an even better base for people to continue building powerful tools on the amazing Rubocop base, but the latter would do as well.

jonas054:
I considered alternative 1, but decided on the approach in this PR, because I wasn't sure what the consequences could be if we introduce actual department hierarchies.

There is a variant of alternative 2 that you can use today without any changes to RuboCop. It uses ERB pre-processing.
```yaml
<% RuboCop::Cop::Registry.global.departments.grep(%r{\AChef/}).each do |dept| %>
<%= dept %>:
  Enabled: true
<% end %>
```
I'm not 100% sure that the APIs used on the first line are considered public APIs. I guess @bbatsov of @koic would know.

jaymzh:
Ohhhhhhh that's neat.  I didn't know the config file was passed through ERB! I can do all sorts of neat things with that (that I shouldn't).

That'll be a really useful short-term solution.

What do you think about #1, but doing it at a major version in case it's a breaking change?

bbatsov:
You also have to keep in mind the added complexity of such a change. If there's not a lot of demand for such a change, I'd rather avoid it. The codebase related to configuration is already very complex and we're quite wary of making it even more complex.

PR #9937, opened 2021-07-16
[Fix #9922] Replace extra disable directives with comma

### LINKED ISSUE #9922: Many infinite loop issues when using `--disable-uncorrectable` flag
## Expected behavior

No errors or skip errors and continue running.

## Actual behavior

Using `rubocop -a --disable-uncorrectable` a big project with no rubocop in the past results in many files running into infinite loops. Most often around `Metrics/AbcSize` and `Metrics/MethodLength`.

Using `rubocop -a` works.

I'll see if I can share code, but for now, I'm just curious to know if this is a known issue.

```
$ [bundle exec] rubocop -V
1.18.3
```

jonas054:
Just removing cop names from disable directives is not a good solution. It's bound to cause infinite loop errors when those cops try to add the disable directives back. The natural solution is to put in commas instead.


koic:
Looks good to me! Thanks!

PR #9949, opened 2021-07-23
[Fix #9946] Avoid slow regexp match in Style/CommentedKeyword

### LINKED ISSUE #9946: --auto-gen-config runs forever
## Expected behavior

--auto-gen-config should finish eventually.

## Actual behavior

--auto-gen-config ran at 100% CPU for 25 minutes before I gave up.

```
top -pid 27995
PID    COMMAND      %CPU  TIME     #TH  #WQ  #POR MEM  PURG CMPR PGRP  PPID  STATE   BOOSTS    %CPU_ME %CPU_OTHRS UID  FAULT COW  MSGSEN MSGR SYSBSD SYSM CSW
27995  ruby         99.8  24:03.73 2/1  0    22   103M 0B   0B   27995 93817 running *0[1]     0.00000 0.00000    redacted  40598 717  13972  85   217280 1309 264006+
```

With `--debug`, the last line printed before it hangs is:

```
Scanning /Users/jaredbeck/redacted/lib/tasks/redacted.rake
```

## Workaround

1. Deleted the .rake file 
2. Ran --auto-gen-config
3. Restored the .rake file

## Why can't --auto-gen-config handle this .rake file?

I don't know. `--debug` doesn't provide any clues, so I can only speculate. The file (below, redacted) has a very long string. Perhaps that's the culprit?

```ruby
desc 'Redacted'
task :redacted => :environment do
  Dir.glob("#{Rails.root}/app/redacted.rb").each { |file| require file }
  redacted1 = 'redacted
300 line
28 kilobyte
string'
  # redacted 338 char long comment
  label = OpenStruct.new(redacted2: redacted1)
  redacted6 = Redacted.new(redacted3: 999).generate_pssl(redacted4: true, redacted5: label)
  ff = File.open("redacted.pdf", "w")
  ff.write(redacted6.force_encoding('utf-8'))
  ff.close
end
```

Try creating a file with a 28 kilobyte string?

## RuboCop version

```
bundle exec rubocop -V
1.18.3 (using Parser 3.0.2.0, rubocop-ast 1.8.0, running on ruby 2.6.3 x86_64-darwin20)
```

jonas054:
The regular expression we use for finding the keyword on the commented line is very slow if the comment is long. We need to reverse the order of the two conditions so we only do the regexp matching after we've discovered that the line begins with one of the keywords we want to check.


koic:
Cool! 👍

Issue #277, opened 2013-06-14
"Missing space after #" warning is not compatible with RDoc markups #++ and #-- .


badal:
From Rdoc's doc :

Rdoc stops processing comments if it finds a comment line starting with -- right after the # character... Commenting can be turned back on with a line that starts with ++.

So lines starting by #++ or #-- should be accepted.

PR #9964, opened 2021-07-30
Make `Layout/LeadingCommentSpace` aware of `#:nodoc`

koic:
This PR makes `Layout/LeadingCommentSpace` aware of `#:nodoc`.

The following mention is the trigger.
https://twitter.com/kamipo/status/1421031611588485121

It seems that there was no definitive comment on whether to allow `#:nodoc:`.

- https://github.com/rubocop/rubocop/issues/277
- https://github.com/rubocop/rubocop/commit/db79146

I've look RDoc doc's example of Ruby 3.0 and it looks to include leading comment space.

```ruby
module MyModule # :nodoc:
  class Input
  end
end

module OtherModule # :nodoc: all
  class Output
  end
end
```

https://docs.ruby-lang.org/en/3.0.0/RDoc/Markup.html#class-RDoc::Markup-label-Controlling+what+is+documented

I also considered making it an option for the presence or absence of leading comment space in `#:nodoc:`, but I think it would be better to consistent it that space is included when possible as the role of cop.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_make_layout_leading_comment_space_aware_of_nodoc.md
+++ b/changelog/change_make_layout_leading_comment_space_aware_of_nodoc.md
@@ -1,0 +1,1 @@
+* [#9964](https://github.com/rubocop/rubocop/pull/9964): Make `Layout/LeadingCommentSpace` aware of `#:nodoc`. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
I like the change, even if we'll likely get some complaints for it. Thanks!

PR #9965, opened 2021-07-30
[Fix #9957] Add `WholeWord` configuration to `Naming/InclusiveLanguage`'s `FlaggedTerms` config

### LINKED ISSUE #9957: False positives for inclusive language
`RuboCop::Cop::Naming::InclusiveLanguage` flags symbols that happen to contain the string "slave" as offences, even when they are contained in composited words. For example "**SLAVe**rsion" or "Te**slaVe**hicle" are seen as offences, but "sla_version" or "tesla_vehicle" are not. I think these 'false positives' point to a bug.

--------

## Expected behavior
The inclusive language cop should mark "slave", "Slave" and "SLAVE" as offences, but not "SLAVe" or "slaVe" or other casings that are the result of compositing words. I would expect the casing of characters to have an influence on the determination of an offence.

## Actual behavior
Symbols containing "SLAVe" or "slaVe" are marked as offences by the inclusive language cop, with the following output:

```
rubocop -A --debug /Users/xxxx/app/models/concerns/xxxx.rb
For /Users/xxxx: configuration from /Users/xxxx/.rubocop.yml
configuration from /Users/xxxx/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/rubocop-rails-2.11.3/config/default.yml
configuration from /Users/xxxx/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/rubocop-rails-2.11.3/config/default.yml
Default configuration from /Users/xxxx/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/rubocop-1.18.4/config/default.yml
Inheriting configuration from /Users/xxxx/.rubocop_todo.yml
.rubocop.yml: Lint/UselessAssignment:Exclude overrides the same parameter in .rubocop_todo.yml
.rubocop.yml: Metrics/CyclomaticComplexity:Max overrides the same parameter in .rubocop_todo.yml
.rubocop.yml: Metrics/PerceivedComplexity:Max overrides the same parameter in .rubocop_todo.yml
Inspecting 1 file
Scanning /Users/xxxx/app/models/concerns/xxxx.rb
Loading cache from /Users/xxxx/.cache/rubocop_cache/f5e32fae37ef606b02bd1b4d287d15320b11827e/32a823ae5c5ceea585bb78727fd983b60ae7a358/7d463336a8c621971967fd1aa6370fda630363bf
C

Offenses:

app/models/concerns/xxxx.rb:52:10: C: Naming/InclusiveLanguage: Consider replacing problematic term 'slaVe' with 'replica', 'secondary', or 'follower'.
class VizslaVet; end
         ^^^^^
app/models/concerns/xxxx.rb:54:9: C: Naming/InclusiveLanguage: Consider replacing problematic term 'slaVe' with 'replica', 'secondary', or 'follower'.
class TeslaVehicle; end
        ^^^^^
app/models/concerns/xxxx.rb:56:7: C: Naming/InclusiveLanguage: Consider replacing problematic term 'SLAVe' with 'replica', 'secondary', or 'follower'.
class SLAVersion; end
      ^^^^^

1 file inspected, 3 offenses detected
Finished in 0.8737279999913881 seconds

Process finished with exit code 1
```

Put any one of these classes in a file inspected by rubocop with the inclusive language cop enabled:

- `class SLAVersion; end`
- `class TeslaVehicle; end`
- `class VizslaVet; end`

## RuboCop version
```
$ [bundle exec] rubocop -V -A /Users/****.rb
1.18.4 (using Parser 3.0.2.0, rubocop-ast 1.8.0, running on ruby 2.7.3 arm64-darwin20)
  - rubocop-rails 2.11.3

Process finished with exit code 0
```

dvandersluis:
This is a *naive* fix for #9957.

Allows `FlaggedTerms` to be set up to only match against a whole word (and enables it for `slave`). This allows false positives like `SLAVersion` and `TeslaVehicle` to not be flagged by the cop.

However, we don't have an inflector for identifiers, and this cop is explicitly case insensitive, which means that enabling this will also stop flagging partial word matches like `DatabaseSlave`. This really only affects class names since `database_slave` will still be flagged.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2557,6 +2557,7 @@ Naming/InclusiveLanguage:
         - denylist
         - block
     slave:
+      WholeWord: true
       Suggestions: ['replica', 'secondary', 'follower']
 
 Naming/MemoizedInstanceVariableName:
--- a/changelog/fix_add_wholeword_configuration_to.md
+++ b/changelog/fix_add_wholeword_configuration_to.md
@@ -1,0 +1,1 @@
+* [#9957](https://github.com/rubocop/rubocop/issues/9957): Add `WholeWord` configuration to `Naming/InclusiveLanguage`'s `FlaggedTerms` config. ([@dvandersluis][])
--- a/lib/rubocop/cop/naming/inclusive_language.rb
+++ b/lib/rubocop/cop/naming/inclusive_language.rb
@@ -19,6 +19,8 @@ module RuboCop
       # Regex can be specified to identify offenses. Suggestions for replacing a flagged term can
       # be configured and will be displayed as part of the offense message.
       # An AllowedRegex can be specified for a flagged term to exempt allowed uses of the term.
+      # `WholeWord: true` can be set on a flagged term to indicate the cop should only match when
+      # a term matches the whole word (partial matches will not be offenses).
       #
       # @example FlaggedTerms: { whitelist: { Suggestions: ['allowlist'] } }
       #   # Suggest replacing identifier whitelist with allowlist
@@ -56,6 +58,14 @@ module RuboCop
       #   # good
       #   # They had a master's degree
       #
+      # @example FlaggedTerms: { slave: { WholeWord: true } }
+      #   # Specify that only terms that are full matches will be flagged.
+      #
+      #   # bad
+      #   Slave
+      #
+      #   # good (won't be flagged despite containing `slave`)
+      #   TeslaVehicle
       class InclusiveLanguage < Base
         include RangeHelp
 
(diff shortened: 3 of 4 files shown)


config/default.yml
```diff
@@ -2557,6 +2557,7 @@ Naming/InclusiveLanguage:
         - denylist
         - block
     slave:
+      WholeWord: true
```
bbatsov, line 2560: This can also be named "PreciseMatch" or something like this. Sounds a bit more descriptive to me.

bbatsov: Or "ExactMatch". I think we use this somewhere else in the config. It'd be best to check, so we stay consistent with ourselves.

dvandersluis: We have `ExactNameMatch` in `Style/TrivialAccessors` but I don't think it exactly fits here. `PreciseMatch` would be ok with me but also I wanted to indicate that `foo_slave` *would* be caught, because it's a different "word". If you like `PreciseMatch` I'm happy to change it.

bbatsov: Fair point. Let's stick with `WholeWord` then. Naming is hard.

PR #9972, opened 2021-08-02
Try fix CI errors on Windows

deivid-rodriguez:
Not sure why the failure was specific to Windows, but I think the reason is that these specs were parsing an invalid lockfile. Since the `specs:` section didn't exist, bundler was failing to parse the lockfile source,and then trying to use it while parsing specs, resulting in a method being called on `nil`.

-----------------

* [ ] Added tests.


koic:
Thank you so much @deivid-rodriguez!

PR #9978, opened 2021-08-03
[Fix #9977] Update `Layout/EmptyLineAfterGuardClause` to not register an offense if there is another expression following the guard clause on the same line

### LINKED ISSUE #9977: Layout/EmptyLineAfterGuardClause inserts empty line outside of block, breaking method chaining
The Layout/EmptyLineAfterGuardClause cop misbehaves in a situation such as this (very contrived) example:

```ruby
# frozen_string_literal: true

def weird_sum(my_array)
  my_array
    .map { |item| next unless item.positive?; item * 2 }
    .compact
    .sum
end
```

## Expected behavior

No empty line inserted, **or** an empty line is inserted inside the block.

## Actual behavior

An empty line is inserted before the `.compact` line, resulting in a syntax error.

```
$ rubocop --debug --auto-correct test.rb
For /home/afn/work: Default configuration from /home/afn/.rvm/gems/ruby-3.0.1/gems/rubocop-1.18.4/config/default.yml
Inspecting 1 file
Scanning /home/afn/work/test.rb
E

Offenses:

test.rb:5:19: C: [Corrected] Layout/EmptyLineAfterGuardClause: Add empty line after guard clause.
    .map { |item| next unless item.positive?; item * 2 }
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
test.rb:7:5: E: Lint/Syntax: unexpected token tDOT
(Using Ruby 3.0 parser; configure using TargetRubyVersion parameter, under AllCops)
    .compact
    ^

Finished in 0.11396609200164676 seconds
```

Resulting file:

```ruby
# frozen_string_literal: true

def weird_sum(my_array)
  my_array
    .map { |item| next unless item.positive?; item * 2 }

    .compact
    .sum
end
```

See above.

## RuboCop version

1.18.4 (using Parser 3.0.2.0, rubocop-ast 1.8.0, running on ruby 3.0.1 x86_64-linux)

dvandersluis:
`Layout/EmptyLineAfterGuardClause` corrects by adding a blank line after the clause, assuming that it is on its own line (or followed by a comment). However, if there is another statement on the same line, it's not taken into account.

This change does not attempt to correct if the clause is not the last node on the line. It'll leave autocorrection to `Style/Semicolon` (which currently doesn't do this autocorrection either).

Fixes #9977.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/layout/empty_line_after_guard_clause_spec.rb
+++ b/spec/rubocop/cop/layout/empty_line_after_guard_clause_spec.rb
@@ -362,7 +362,7 @@ RSpec.describe RuboCop::Cop::Layout::EmptyLineAfterGuardClause, :config do
     RUBY
   end
 
-  it 'accpets a guard clause when the next line is `ensure`' do
+  it 'accepts a guard clause when the next line is `ensure`' do
     expect_no_offenses(<<~RUBY)
       def foo
         begin
@@ -374,7 +374,7 @@ RSpec.describe RuboCop::Cop::Layout::EmptyLineAfterGuardClause, :config do
     RUBY
   end
 
-  it 'accpets a guard clause when the next line is `rescue`-`else`' do
+  it 'accepts a guard clause when the next line is `rescue`-`else`' do
     expect_no_offenses(<<~RUBY)
       def foo
         begin
@@ -400,7 +400,7 @@ RSpec.describe RuboCop::Cop::Layout::EmptyLineAfterGuardClause, :config do
     RUBY
   end
 
-  it 'accpets a guard clause when the next line is `elsif`' do
+  it 'accepts a guard clause when the next line is `elsif`' do
     expect_no_offenses(<<~RUBY)
       def foo
         if cond
@@ -438,7 +438,7 @@ RSpec.describe RuboCop::Cop::Layout::EmptyLineAfterGuardClause, :config do
     RUBY
   end
 
-  it 'accpets a guard clause that is after a multiline heredoc with chained calls' do
+  it 'accepts a guard clause that is after a multiline heredoc with chained calls' do
     expect_no_offenses(<<~RUBY)
       def foo
         raise ArgumentError, <<~END.squish.it.good unless guard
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/layout/empty_line_after_guard_clause_spec.rb
```diff
@@ -560,4 +560,15 @@ def foo
       end
     RUBY
   end
+
+  it 'registers an offense but does not correct when the clause is ended with a semicolon' do
+    expect_offense(<<~RUBY)
+      def foo(item)
+        return unless item.positive?; item * 2
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Add empty line after guard clause.
```
koic, line 568: I think this case is acceptable because it is a false positive. So the cop can respect the will of programmers who want to use semicolons.

dvandersluis: @koic I'm not sure I follow, that offense isn't new, this PR just doesn't try to autocorrect the offense.

Are you saying this should not be an offense?

koic: Yes. I think `expect_no_offences` should be used in this test and pass for cop role’s design.

dvandersluis: @koic updated, it'll now not consider guard clauses when there's another statement on the same line. I added some additional tests as well.

spec/rubocop/cop/layout/empty_line_after_guard_clause_spec.rb
```diff
@@ -28,7 +34,7 @@ def foo
       end
     RUBY
 
-    expect_correction(<<~RUBY)
+    expect_correction(<<~RUBY )
```
koic, line 37: That's just my two cents.

```suggestion
    expect_correction(<<~RUBY)
```

spec/rubocop/cop/layout/empty_line_after_guard_clause_spec.rb
koic: Great! Thanks!

spec/rubocop/cop/layout/empty_line_after_guard_clause_spec.rb
dvandersluis: 🤦 thanks @koic!

PR #9981, opened 2021-08-04
[Fix #9962] Update `Style/WordArray` to register an offense in `percent` style if any values contain spaces

### LINKED ISSUE #9962: WordArray should enforce bracket style for strings that don't match WordRegex

Given that WordArray won't suggest changing `['string with space', 'string']` to `%w[string\ with\ space string]`, should it suggest changing `%w[string\ with\ space string]` to `['string with space', 'string']`?

WordArray should enforce bracket style for arrays that contain strings that don't match WordRegex

dvandersluis:
Updates `Style/WordArray` to mark `%w()` arrays as offenses if they contain any values with spaces.

Fixes #9962.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/word_array.rb
+++ b/lib/rubocop/cop/style/word_array.rb
@@ -9,6 +9,9 @@ module RuboCop
       # Alternatively, it can check for uses of the %w() syntax, in projects
       # which do not want to include that syntax.
       #
+      # NOTE: When using the `percent` style, %w() arrays containing a space
+      # will be registered as offenses.
+      #
       # Configuration option: MinSize
       # If set, arrays with fewer elements than this value will not trigger the
       # cop. For example, a `MinSize` of `3` will not enforce a style on an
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/style/word_array.rb
```diff
@@ -53,13 +62,22 @@ def on_array(node)
...
+            !string.valid_encoding? ||
+              (word_regex && !word_regex.match?(string)) ||
+              / /.match?(string)
           end
         end
 
+        def invalid_percent_array_contents?(node)
+          # Disallow %w() arrays that contain invalid encoding or spaces
+          complex_content?(node.values, word_regex: false)
+        end
```
dvandersluis, line 79: Originally this used the same `complex_content?` check, taking the `WordRegex` configuration into account, however, this caused 206 offenses on rubocop itself, disallowing a lot of `%w()` arrays that are probably good to keep (ie. they contained non word characters). For instance: https://github.com/rubocop/rubocop/blob/281591717f37dab297358f695eb6e67c38352169/spec/rubocop/cop/lint/literal_as_condition_spec.rb#L4

As a result, I made the check only look for space issues. If desired we could add in another configuration to specify a regex for this check?

bbatsov:
Good work!

PR #9982, opened 2021-08-05
[Fix #9980] Fix a false positive for `Style/IdenticalConditionalBranches`

### LINKED ISSUE #9980: Style/IdenticalConditionalBranches should be marked as unsafe
## Expected behavior

I expect safe autocorrection not to break my existing ruby code.

## Actual behavior

Running a cop marked as safe breaks my existing ruby code.

Given `test.rb`
```ruby
x = 0

if x == 0
  x += 1
  puts "if"
else
  x += 1
  puts "else"
end

# output when executed:
# if
```

When I run `rubocop -a --only Style/IdenticalConditionalBranches`, it autocorrects my file to:
```ruby
x = 0

x += 1
if x == 0
  puts "if"
else
  puts "else"
end

# output when executed:
# else
```

## RuboCop version

```
$ rubocop -V
1.18.4 (using Parser 3.0.2.0, rubocop-ast 1.8.0, running on ruby 2.7.3 x86_64-darwin20)
```

koic:
Fixes #9980.

This PR fixes a false positive for `Style/IdenticalConditionalBranches` when assigning to a variable used in a condition.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positive_for_style_identical_conditional_branches.md
+++ b/changelog/fix_false_positive_for_style_identical_conditional_branches.md
@@ -1,0 +1,1 @@
+* [#9980](https://github.com/rubocop/rubocop/issues/9980): Fix a false positive for `Style/IdenticalConditionalBranches` when assigning to a variable used in a condition. ([@koic][])
--- a/lib/rubocop/cop/style/identical_conditional_branches.rb
+++ b/lib/rubocop/cop/style/identical_conditional_branches.rb
@@ -124,14 +124,21 @@ module RuboCop
           return if branches.any?(&:nil?)
 
           tails = branches.map { |branch| tail(branch) }
-          check_expressions(node, tails, :after_condition) if duplicated_expressions?(tails)
+          check_expressions(node, tails, :after_condition) if duplicated_expressions?(node, tails)
 
           heads = branches.map { |branch| head(branch) }
-          check_expressions(node, heads, :before_condition) if duplicated_expressions?(heads)
+          check_expressions(node, heads, :before_condition) if duplicated_expressions?(node, heads)
         end
 
-        def duplicated_expressions?(expressions)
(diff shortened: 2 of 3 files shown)


Drowze:
One case that I had issues with autocorrection for this particular cop on our codebase is specifically as follows:
```ruby
if shipment.assignments.find_by(employee_id: employee_id)
  shipment.assignments.destroy_all
else
  shipment.assignments.destroy_all

  # do other stuff
end
```

When running this cop's autocorrection, that excerpt was refactored to:
```ruby
shipment.assignments.destroy_all

if shipment.assignments.find_by(employee_id: employee_id)  
else
  # do other stuff
end
```

After rubocop's autocorrection, the code always falls to the `else` block. I think this PR wouldn't solve that particular case, right? I'm also not sure if there's a way for this cop to be safe on such cases 🤔

vlad-pisanov:
I believe this autocorrect is inherently unsafe. Consider:

```ruby
if method_that_modifies_global_state
  method_that_relies_on_global_state
  foo
else
  method_that_relies_on_global_state
  goo
end
```

There's no way to safely factor out `method_that_relies_on_global_state` _before_ the `method_that_modifies_global_state` is executed. `method_that_modifies_global_state` could modify all sorts of things like global vars, mocks, etc. that `method_that_relies_on_global_state` needs to function properly.

koic:
@vlad-pisanov Ah... you're right! This cop can try to reduce false positives, but I understand that it cannot be safe because a side effects of method invocation.

koic:
I opened a separated PR #9985 because I think this should be marked as unsafe for auto-correction.
https://docs.rubocop.org/rubocop/1.18/usage/auto_correct.html#safe-auto-correct

bbatsov:
Thanks for tackling this!

PR #9986, opened 2021-08-06
[Fix #4182] Add `Lint/AmbiguousRange` cop to check for ranges with ambiguous boundaries

### LINKED ISSUE #4182: New cop for `..` operator precedence
The following code has a bug.

```ruby
# test.rb
def foo(x = nil)
  x || 1..2
end

p foo
p foo(1..3)
```

```sh
$ ruby test.rb
1..2
test.rb:3:in `foo': bad value for range (ArgumentError)
	from test.rb:7:in `<main>'
```

The code is interpreted as `(x || 1)..2`, but it looks like `x || (1..2)` to me.
I think it is confusing, so, we need a cop for this problem. However we don't have the cop.

The cause is `..` operator's precedence. https://ruby-doc.org/core-2.2.0/doc/syntax/precedence_rdoc.html

--------

## Expected behavior

```ruby
x || 1..2
```

RuboCop adds an offence for the above code.

## Actual behavior

The cop for this problem does not exist.

## RuboCop version

```
$ rubocop -V
0.48.0 (using Parser 2.4.0.0, running on ruby 2.4.1 x86_64-linux)
```

dvandersluis:
Adds a new cop `Lint/AmbiguousRange` to detect range boundaries that are ambiguous. An ambiguous boundary is one that is not a basic literal, single variable/constant/method reference, or method chain (depending on the value of `RequireParenthesesForMethodChains`, default `false`). Originally I was intending to have it only accept unparenthesized basic literals, but it would require too many ranges in rubocop's code to be changed that I re-evaluated.

This change will allow `1..2` without parens but require the parens in `(x || 1)..10` or `(x || 1)..(y || 10)`. Having the cop act on each boundary individually is the clearest approach IMO because `(x || 1..y || 10)` is just as ambiguous as without the parentheses.

In order to avoid the ambiguity of `1..2.to_a`, method calls with a basic literal receiver always require parentheses, regardless of `RequireParenthesesForMethodChains`.

Autocorrection is straightforward, wrapping any unacceptable boundaries in parens. However, I have marked this cop as `SafeAutocorrect: false` because it is likely that the intended code is not the same as the actual behaviour of the code. In other words, `1..2.to_a` is equivalent to `1..(2.to_a)`, which is what the cop will correct to, but the programmer almost definitely intended on `(1..2).to_a`. The autocorrection will make it clear how ruby will actually evaluate this code, however.

This seemed like it was going to be simple until it became complicated :grin:

Fixes #4182.

-----------------

* [x] Added tests.


bbatsov:
Looks pretty solid to me. Great work!

PR #9994, opened 2021-08-10
[Fix #5022] Fix false negative for `Lint/DuplicateMethods` inside `class << self`

### LINKED ISSUE #5022: Lint/DuplicateMethods for classes inside `class << self` of parent module
As in #4758, but with `class << self`.

Real-life case: https://github.com/panthomakos/timezone/blob/b70e1d8/lib/timezone/lookup.rb

--------

## Expected behavior

No warnings.

## Actual behavior

```
Inspecting 1 file
W

Offenses:

.test.rb:10:7: W: Method Foo.config is defined at both .test.rb:4 and .test.rb:10.
      attr_reader :config
      ^^^^^^^^^^^

```

```ruby
# Module with class-methods
module Foo
  class << self
    def config
      'Foo.config'
    end

    # Private class for module
    class Bar
      attr_reader :config
    end
  end
end
```

## RuboCop version

```
$ rubocop -V
0.51.0 (using Parser 2.4.0.0, running on ruby 2.4.2 x86_64-linux)
```

dvandersluis:
Requires rubocop/rubocop-ast#197 in order to fix `Node#parent_module_name`.

Properly handles the updated format of `parent_module_name` and adds tests to ensure the case from #5022 works.

Also reverts the pending tests in #9995.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/duplicate_methods_spec.rb
+++ b/spec/rubocop/cop/lint/duplicate_methods_spec.rb
@@ -196,7 +196,7 @@ RSpec.describe RuboCop::Cop::Lint::DuplicateMethods, :config do
       RUBY
     end
 
-    xit 'understands class << self' do
+    it 'understands class << self' do
       expect_offense(<<~RUBY, 'test.rb')
         #{opening_line}
           class << self
@@ -424,8 +424,8 @@ RSpec.describe RuboCop::Cop::Lint::DuplicateMethods, :config do
       RUBY
     end
 
-    xit 'registers an offense for duplicate class methods with `<<` and named ' \
-        "receiver in #{type}" do
+    it 'registers an offense for duplicate class methods with `<<` and named ' \
+       "receiver in #{type}" do
       expect_offense(<<~RUBY, 'test.rb')
         #{type} A
           class << self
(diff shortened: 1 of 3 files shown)


dvandersluis:
There's a bit of a catch 22 between this PR and rubocop/rubocop-ast#197. Each PR needs the other to get the tests passing.

spec/rubocop/cop/lint/duplicate_methods_spec.rb
```diff
@@ -233,16 +233,14 @@ def self.another
...
         #{opening_line}
           class << blah
             def some_method
               implement 1
             end
             def some_method
-            ^^^^^^^^^^^^^^^ Method `A#some_method` is defined at both test.rb:3 and test.rb:6.
               implement 2
             end
           end
```
dvandersluis, line 246: This test will not be able to ever pass, because `Node#parent_module_name` will return `nil` because it cannot evaluate `class << blah`, so I have removed the `pending` and marked it as `expect_no_offenses`.

marcandre: I would mark it as `pending`, as ideally the cop would detect it.

dvandersluis: Ok I'll drop that commit then.

PR #9995, opened 2021-08-10
Temporarily mark some `Lint/DuplicateMethods` specs as pending

dvandersluis:
See https://github.com/rubocop/rubocop-ast/pull/197#issuecomment-896167768.

This will be reverted once rubocop/rubocop-ast#197 is released.


dvandersluis:
RSpec is particular about how to mark a test pending if it passes 😅

PR #4537, opened 2017-06-21
Use `--parallel` option for internal_investigation task

pocke:
This change parallelizes internal investigation in local computer and CI.

checklist:

- [x] It works with JRuby? -> Memory error
- [x] It works on Windows? -> `Process.fork` is not supported

So, parallelizing is disabled in the above environments.

Note:

I'm trying parallelizing our tests by using [test-queue](https://github.com/tmm1/test-queue) also.

PR #10000, opened 2021-08-10
Parallel static analysis by default

koic:
Modern PCs have multi-core. RuboCop has `--parallel` option, but it is not used by default. This means that there is extra core resources for static analysis because multi-core is not being used effectively.

This PR will be parallel processing by default, so RuboCop will run faster in proportion to number of cores. Options that can be parallel processing, such as `--only`, `--except`, and others have the
same default behavior.
However, there are options that cannot be parallel processing, such as auto-correction and others, so they are serial processing as before.

`--parallel` can be a hidden option for some users. This change is expected to reduce user's wait time for running RuboCop.

Below is an example on my development PC.

Before:

```console
% cd path/to/rubocop
./exe/rubocop --display-time
(snip)

1293 files inspected, no offenses detected
Finished in 61.09493000002112 seconds
```

After:

```console
% cd path/to/rubocop
./exe/rubocop --display-time
(snip)

1293 files inspected, no offenses detected
Finished in 10.13813699997263 seconds
```

If user's PC has a lot of cores, memory and RuboCop targets a lot of files to inspect, it will be effective.

One concern is that parallel processing uses more memory. For example, Ruby's MJIT is off by default as an example of concerning about memory consumption in a low memory production runtime (e.g. Heroku's minimum Dyno).
However, RuboCop is a development tool, I assume that it will be fine in most cases.

If user want to use it in a low memory production runtime, user can specify `--no-parallel`.

Note: JRuby and Windows are not supported based on #4537.

-----------------

* [x] Added tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -22,12 +22,7 @@ require 'rubocop/rake_task'
 Dir['tasks/**/*.rake'].each { |t| load t }
 
 desc 'Run RuboCop over itself'
-RuboCop::RakeTask.new(:internal_investigation).tap do |task|
-  if RUBY_ENGINE == 'ruby' &&
-     !/mswin|msys|mingw|cygwin|bccwin|wince|emc/.match?(RbConfig::CONFIG['host_os'])
-    task.options = %w[--parallel]
-  end
-end
+RuboCop::RakeTask.new(:internal_investigation)
 
 task default: %i[documentation_syntax_check spec ascii_spec internal_investigation]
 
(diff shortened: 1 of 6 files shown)


dvandersluis:
The failures on this and your other PRs are resolved by #9995 (and will be fixed by #9994), so it looks like you need to rebase.

bbatsov:
I think that's a good move! The main reason why we didn't do this earlier is that for a while the parallel analysis was rough around the edges, but given the massive time savings, I think it makes for a a better default.

P.S. `10000` is the quite the magic number! 🚀 🏅

PR #10012, opened 2021-08-13
[Fix #10011] Fix a false positive for `Style/RedundantSelfAssignmentBranch`

### LINKED ISSUE #10011: Autocorrecting a Style/RedundantSelfAssignmentBranch offence breaks existing code
While upgrading our codebase to RuboCop 1.19.0 and running `bundle exec rubocop -a`, I noticed some failing specs. I believe the [Style/RedundantSelfAssignmentBranch](https://docs.rubocop.org/rubocop/cops_style.html#styleredundantselfassignmentbranch) cop autocorrections should be marked unsafe. See the example below.

--------

## Expected behavior

`bundle exec rubocop -a` should not break existing code.

## Actual behavior

`bundle exec rubocop -a` broke some existing code.

Let's say we have the following (rather odd) piece of code in `test.rb`:
```ruby
class Test
  def foo
    @foo = if @foo_set
      @foo
    else
      @foo_set = true
      @foo = 'bar'
    end
  end
end

test = Test.new
puts test.foo == 'bar' # should be true
puts test.foo == 'bar' # should be true
```

Running this file will give the following output:
```
true
true
```

RuboCop will display the following message on line 4: 
```
Style/RedundantSelfAssignmentBranch: Remove the self-assignment branch.
```

Executing `bundle exec rubocop -a test.rb` will update the code to this:
```ruby
class Test
  def foo
    @foo = unless @foo_set
      @foo_set = true
      @foo = 'bar'
    end
  end
end

test = Test.new
puts test.foo == 'bar' # should be true
puts test.foo == 'bar' # should be true
```

Running this file will give a different output:
```
true
false
```

## RuboCop version

```
1.19.0 (using Parser 3.0.2.0, rubocop-ast 1.10.0, running on ruby 2.7.4 x86_64-darwin20)
  - rubocop-performance 1.11.4
  - rubocop-rails 2.11.3
  - rubocop-rspec 2.4.0
```

koic:
Fixes #10011.

This PR fixes a false positive for `Style/RedundantSelfAssignmentBranch` when using instance variable, class variable, and global variable.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positive_for_style_redundant_self_assignment_branch.md
+++ b/changelog/fix_false_positive_for_style_redundant_self_assignment_branch.md
@@ -1,0 +1,1 @@
+* [#10011](https://github.com/rubocop/rubocop/issues/10011): Fix a false positive for `Style/RedundantSelfAssignmentBranch` when using instance variable, class variable, and global variable. ([@koic][])
--- a/lib/rubocop/cop/style/redundant_self_assignment_branch.rb
+++ b/lib/rubocop/cop/style/redundant_self_assignment_branch.rb
@@ -5,6 +5,9 @@ module RuboCop
     module Style
       # This cop checks for places where conditional branch makes redundant self-assignment.
       #
+      # It only detects local variable because it may replace state of instance variable,
+      # class variable, and global variable that have state across methods with `nil`.
+      #
       # @example
       #
       #   # bad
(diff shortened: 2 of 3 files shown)


bbatsov:
Perhaps we should have made the extended checks optional, as they can still yield meaningful results from time to time, but I'm fine with checking only locals. Thanks!

Issue #4811, opened 2017-09-29
Wanted: Layout/SpaceInsideArrayLiteralBrackets


LandonSchropp:
#I would like to be enforce a Ruby array spacing style like this:

``` ruby
array = [ 1, 2, 3, 4 ]
array = [ "hello", "hola", "bonjour" ]
```

I've searched the documentation and GitHub issues for something like this, but I haven't been able to find anything. Rubocop includes [Layout/SpaceInsideBrackets](https://rubocop.readthedocs.io/en/latest/cops_layout/#layoutspaceinsidebrackets), but that seems to only enforce *removing* spaces, not including them. Furthermore, I only want to include spaces inside arrays, not accessor brackets or other bracket operators.

This would be the array equivilent of [Layout/SpaceInsideHashLiteralBraces](https://rubocop.readthedocs.io/en/latest/cops_layout/#layoutspaceinsidehashliteralbraces). It's also equivilent to [array-bracket-spacing](https://eslint.org/docs/rules/array-bracket-spacing) in ESLint.

Rubocop version: `0.50.0 (using Parser 2.4.0.0, running on ruby 2.4.2 x86_64-darwin17)`

mvastola:
I would like to +1 this. Having spaces really improves readability IMHO, if you're listing elements.
Going off the convention rubocop seems employ though, I imagine this wouldn't be it's own layout cop but rather a couple of `SupportedStyles` for the existing cop.

Drenmi:
Fair enough. How would you like to treat multiline arrays under the new setting?

mvastola:
Hm. I personally use the style:
```ruby
arr = [
  :a,
  :b,
  :c
]
```
So this wouldn't apply to me, but I would imagine if you added spaces inside brackets you would do so everywhere. (Including the first/last line of a multi-line array if there is an element on it.) 

I have trouble imagining someone constructing any other sort of convention, though if others actually employ such a coding style or else envision it actually being used, then perhaps it merits its own `SupportedStyle`.

LandonSchropp:
Ditto to the multi-line style @mvastola mentioned. I think the array brackets should be configurable in the same way that hashes are. IMO, each one of the following cops should have an array equivalent:

* `Layout/MultilineHashBraceLayout` → `Layout/MultilineArrayBraceLayout` (exists)
* `Layout/SpaceInsideHashLiteralBraces` → `Layout/SpaceInsideArrayLiteralBrackets` (doesn't exist)
[...]

PR #10025, opened 2021-08-18
Implemented Layout/SpaceInsideParens with EnforcedStyle: compact

itay-grudev:
```yml
Layout/SpaceInsideParens:
  EnforcedStyle: compact
```

```ruby
@example 
# The `compact` style enforces that parentheses have a space at the
# beginning with the exception that successive right parentheses are allowed.
# Note: Empty parentheses should not have spaces.

# bad
f(3)
g = (a + 3)
y( )
g( f( x ) )
g( f( x( 3 ) ), 5 )

# good
f( 3 )
g = ( a + 3 )
y()
g( f( x ))
g( f( x( 3 )), 5 )
```

-----------------

* [X] Added tests.

### DIFF
--- a/changelog/new_space_inside_parens_compact.md
+++ b/changelog/new_space_inside_parens_compact.md
@@ -1,0 +1,1 @@
+* [#10025](https://github.com/rubocop/rubocop/pull/10025): Changed cop `SpaceInsideParens` to include a `compact` style.  ([@itay-grudev][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -1347,10 +1347,11 @@ Layout/SpaceInsideParens:
   StyleGuide: '#spaces-braces'
   Enabled: true
   VersionAdded: '0.49'
-  VersionChanged: '0.55'
+  VersionChanged: '<<next>>'
   EnforcedStyle: no_space
   SupportedStyles:
     - space
+    - compact
     - no_space
 
 Layout/SpaceInsidePercentLiteralDelimiters:
--- a/lib/rubocop/cop/layout/space_inside_parens.rb
+++ b/lib/rubocop/cop/layout/space_inside_parens.rb
@@ -33,6 +33,25 @@ module RuboCop
       #   g = ( a + 3 )
       #   y()
       #
+      # @example EnforcedStyle: compact
+      #   # The `compact` style enforces that parentheses have a space at the
+      #   # beginning with the exception that successive right parentheses are allowed.
+      #   # Note: Empty parentheses should not have spaces.
+      #
+      #   # bad
+      #   f(3)
+      #   g = (a + 3)
+      #   y( )
+      #   g( f( x ) )
+      #   g( f( x( 3 ) ), 5 )
+      #
+      #   # good
+      #   f( 3 )
+      #   g = ( a + 3 )
+      #   y()
+      #   g( f( x ))
+      #   g( f( x( 3 )), 5 )
+      #
       class SpaceInsideParens < Base
         include SurroundingSpace
         include RangeHelp
@@ -45,8 +64,11 @@ module RuboCop
         def on_new_investigation
           tokens = processed_source.sorted_tokens
 
-          if style == :space
+          case style
+          when :space
             process_with_space_style(tokens)
+          when :compact
+            process_with_compact_style(tokens)
           else
             each_extraneous_space(tokens) do |range|
               add_offense(range) do |corrector|
@@ -73,6 +95,17 @@ module RuboCop
           end
         end
 
+        def process_with_compact_style(tokens)
+          process_with_space_style(tokens)
+          tokens.each_cons(2) do |token1, token2|
+            each_extaneus_space_between_right_parents(token1, token2) do |range|
+              add_offense(range) do |corrector|
+                corrector.remove(range)
+              end
+            end
+          end
+        end
+
         def each_extraneous_space(tokens)
           tokens.each_cons(2) do |token1, token2|
             next unless parens?(token1, token2)
@@ -86,6 +119,14 @@ module RuboCop
           end
         end
 
+        def each_extaneus_space_between_right_parents(token1, token2)
+          return unless right_parens?(token1, token2)
+
+          return if range_between(token1.end_pos, token2.begin_pos).source != ' '
+
+          yield range_between(token1.end_pos, token2.begin_pos)
+        end
+
         def each_extraneous_space_in_empty_parens(token1, token2)
           return unless token1.left_parens? && token2.right_parens?
 
@@ -96,6 +137,8 @@ module RuboCop
 
         def each_missing_space(token1, token2)
           return if can_be_ignored?(token1, token2)
+
+          return if style == :compact && token1.right_parens? && token2.right_parens?
 
           if token1.left_parens?
             yield range_between(token2.begin_pos, token2.begin_pos + 1)
@@ -110,6 +153,10 @@ module RuboCop
 
         def parens?(token1, token2)
           token1.left_parens? || token2.right_parens?
+        end
+
+        def right_parens?(token1, token2)
+          token1.right_parens? && token2.right_parens?
         end
 
         def can_be_ignored?(token1, token2)
--- a/lib/rubocop/cop/layout/space_inside_parens.rb
+++ b/lib/rubocop/cop/layout/space_inside_parens.rb
@@ -35,7 +35,7 @@ module RuboCop
       #
       # @example EnforcedStyle: compact
       #   # The `compact` style enforces that parentheses have a space at the
-      #   # beginning with the exception that successive right parentheses are allowed.
+      #   # beginning with the exception that successive parentheses are allowed.
       #   # Note: Empty parentheses should not have spaces.
       #
       #   # bad
--- a/lib/rubocop/cop/layout/space_inside_parens.rb
+++ b/lib/rubocop/cop/layout/space_inside_parens.rb
@@ -44,6 +44,7 @@ module RuboCop
       #   y( )
       #   g( f( x ) )
       #   g( f( x( 3 ) ), 5 )
+      #   g( ( ( 3 + 5 ) * f) ** x, 5 )
       #
       #   # good
       #   f( 3 )
@@ -51,6 +52,7 @@ module RuboCop
       #   y()
       #   g( f( x ))
       #   g( f( x( 3 )), 5 )
+      #   g((( 3 + 5 ) * f ) ** x, 5 )
       #
       class SpaceInsideParens < Base
         include SurroundingSpace
@@ -70,11 +72,7 @@ module RuboCop
           when :compact
             process_with_compact_style(tokens)
           else
-            each_extraneous_space(tokens) do |range|
-              add_offense(range) do |corrector|
-                corrector.remove(range)
-              end
-            end
+            correct_extraneous_space(tokens)
           end
         end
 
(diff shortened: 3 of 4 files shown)


lib/rubocop/cop/layout/space_inside_parens.rb
```diff
@@ -86,6 +119,14 @@ def each_extraneous_space(tokens)
           end
         end
 
+        def each_extaneus_space_between_right_parents(token1, token2)
```
dvandersluis, line 122:
```suggestion
        def each_extraneous_space_between_right_parents(token1, token2)
```

Or just use `extra`?

lib/rubocop/cop/layout/space_inside_parens.rb
```diff
@@ -86,6 +119,14 @@ def each_extraneous_space(tokens)
...
+          return unless right_parens?(token1, token2)
+
+          return if range_between(token1.end_pos, token2.begin_pos).source != ' '
```
dvandersluis, line 125: You should extract the `range_between` call into a variable rather than call it twice.

changelog/new_space_inside_parens_compact.md
```diff
@@ -0,0 +1 @@
+* Changed cop `SpaceInsideParens` to include a `compact` style.  ([@itay-grudev][])
```
dvandersluis, line 1: This needs to link to this PR.
```suggestion
* [#10025](https://github.com/rubocop/rubocop/pull/10025) Changed cop `SpaceInsideParens` to include a `compact` style.  ([@itay-grudev][])
```

dvandersluis:
I'm not sure I understand why it'd be desired to compact only right parens? If this is a style that people use, should it not apply on both sides? Or else, the style should probably be called `compact_right` or something similar. I'm not sure if "compact" even makes sense here when most of the parentheses have white space around them, which makes them not compact 😄 Maybe `collapse`/`collapse_right`?

Also you'd need to update `VersionChanged` in default.yml.

config/default.yml
```diff
@@ -1347,10 +1347,11 @@ Layout/SpaceInsideParens:
   StyleGuide: '#spaces-braces'
   Enabled: true
   VersionAdded: '0.49'
-  VersionChanged: '0.55'
+  VersionChanged: '1.19'
```
dvandersluis, line 1350:
```suggestion
  VersionChanged: '<<next>>'
```

Sorry should have specified this.

itay-grudev: Out of curiosity, how is this replaced afterwards? Job that runs on pull_requests?

dvandersluis: There's a `cut_release` task that replaces all the `<<next>>`s with whatever the next version ends up being.

dvandersluis:
I'm not convinced that this is a style that makes sense for RuboCop as a whole. You could implement it as a custom cop for your codebase/company and just disable `Layout/SpaceInsideParens`.

@bbatsov what do you think?

itay-grudev:
RubCop is a collection of rules, configurable to suit different projects, different teams with different views. The objective of the project is internal consistency, with a healthy set of defaults. If there was one style that was suitable for everyone - there wouldn't be any need for configuration. But that's not the case and there is configuration.

All is good until you miss the one option you need, in which case you either disable that corresponding cop, or if you want to be better then that - you spend the time to reverse-engineer how Rubocop works to modify it to include your use case. And I did exactly that, I wrote tests, documentation, and followed your particular style and recommendations as close as I can.

Saying that this style doesn't makes sense for RuboCop as a whole, kind of violates the spirit of Rubocop, doesn't it? It adds too much personal opinion, in a project where personal opinion should be no more than the default, albeit configurable option.

dvandersluis:
@itay-grudev I'm not suggesting that there should be a single allowed style -- in fact this cop is already configurable as you have seen. That being said, we cannot guarantee that every single style that a company uses makes sense for the RuboCop userbase as a whole. I'm also not saying that this _doesn't_ fit, I'm interested what others on the team think.

We often recommend niche rules be implemented as custom cops (in fact, we have our own custom cops at my workplace that don't necessarily make sense for the tool as a whole). It is simple to create your own cop, and the effort you have put in for this PR would not go to waste, because it could be repurposed as such.

bbatsov:
Indeed. We can't possibly support any style that exists (that would mean an epic overhead in the codebase), that's why we focused only on those that are common within the community. I'm leaning towards rejecting the proposed addition.

itay-grudev:
@bbatsov @dvandersluis We have compact for the following:

`Layout/SpaceInsideArrayLiteralBrackets `
```ruby
array = [ a, [ b, c ]]
array = [[ a ],
  [ b, c ]]
  ```
  
`Layout/SpaceInsideHashLiteralBraces`
```ruby
h = { a: { b: 2 }}
foo = {{ a: 1 } => { b: { c: 2 }}}
```

The `Layout/SpaceInsideParens` `compact` version just adds consistency.

bbatsov:
Hmm, I don't recall when those were added. I'll have to dig up the relevant tickets. 

> The Layout/SpaceInsideParens compact version just adds consistency.

I guess now I'm more receptive to accepting this, but I first have to recall what exactly was I thinking back in the day for those other cops. Perhaps back then I still wasn't as afraid of complexity. :D

bbatsov:
Looking at [this ticket](https://github.com/rubocop/rubocop/issues/4811) things started from the `SpaceInsideHashLiteralBraces` cop and were just copied everywhere else. More digging is needed!

itay-grudev:
@bbatsov In any way - while not as popular, this pattern actually improves readability. Here is an excerpt from our internal guidelines:

> Numerous neuroscience researchers showed that the way the human brain processes text is by looking at a words as a whole rather than individual letters. By improving contrast between different words we help our brain read faster. For example, reading the following sentence is significantly harder when written like this:
> 
> Whitespace(is)oneofthe$most%important#tools@in!our+arsenal;to/increase>code>readability, and it all comes to increasing contrast between different words in a sentence. For example, code highlighting was invented just for that purpose. And while highlighting is usually sufficient to add contrast, adding whitespace correctly further improves it. By employing both syntax highlighting and correct whitespace usage we can help our brains to read code diagonally much faster.
> 
> So the real question then is - where do we add whitespace? What do we add contrast to? What is important?
> 
> Like every language, programming languages have grammar. Grammar is important for compilers or interpreters to understand our code correctly, but it's not that important for humans. We can still understand a sentence with a grammatical error, while a computer won't. For this reason, grammar in a programming language is more important for computers than for humans for reading and understanding a piece of code, hence why grammar is considered low priority when it comes to contrast.
> 
> Instead we should put contrast to part of our code that have meaning to us - like method and variable names, expression operators, values, etc. and not brackets, flow control operators, etc.
> 
> Across the app (_Front-end_ and _Back-end_), we believe spaces are key to code readability. Think of your code as important words or tokens with some grammar in-between. The main rule is that grammar is not as important as content and we add spaces to highlight words, not grammatical structures. Consider the following:
> 

> 
> ```ruby
> #   ╔ language grammar (surrounded by spaces, but note - the whole expression including the opening bracket)
> #   ║       ╔ variable name (surrounded by spaces)
> #   ║       ║    ╔ expression operator (surrounded by spaces as it is important for understanding the context of the expression)
> #   ║       ║    ║       ╔ string value (surrounded by spaces)
> #   ▽       ▽    ▽       ▽
>     if( variable == "expression" ){
> #     △                           △
> #     ╚ a bracket by itself is not a word and adds very little meaning. It's just part of the language's grammar, it is expected there - so no spaces.
> ```

flanger001:
> RubCop is a collection of rules, configurable to suit different projects and different teams with different views.

Exactly, and it is possible to write custom cops for your team for exactly this reason. People submit custom cops here all the time. Sometimes they get added, sometimes they don't.

bbatsov:
I'd suggest taking a look here https://docs.rubocop.org/rubocop/1.19/index.html#philosophy

> The early feedback we got lead us to adopt of philosophy of (extreme) configurability and flexibility, and trying to account for every common style of programming in Ruby.

"common" is the key word here. Anyways, I'll admit I've not always thought enough about some of the things I've accepted, plus over time, as the project grew I became significantly more concerned about the overhead of maintaining some niche styles. 

Again, I'm not rejecting this particular proposal (I'll likely merge it if this has existed in similar cops), I just want everyone to know what we're aiming for in general.

koic:
This is a perspective on the keyword "common". IMHO, I've never encountered this style in well-known programming books and OSS of Ruby. (I may have seen it in Perl...?, but I don't remember it well.) Of course this is my personal perspective. I know that each programmer encounters different styles.

OTOH, I agree with @dvandersluis as a maintainer. I also have several niche custom cops at my workplace.

bbatsov:
@koic I think we're all in agreement here, but as @itay-grudev pointed out somehow this style got added to two other existing cops (`Layout/SpaceInsideArrayLiteralBrackets` and `Layout/SpaceInsideHashLiteralBraces`). For the sake of consistency we either have to add it here as well or remove/deprecate it from the other cops. I have no idea why we felt those braces needed special handling in literals, but not in general, back in the day.

koic:
@bbatsov Thank you for your explanation. My understanding has caught up. But this is certainly a difficult issue to choose. I haven't come up with a better answer yet.
This is a maintenance perspective. `compact` style may be relatively accepted if the alternative option doesn't increase in the future. However, the tendency cannot be confirmed.

bbatsov:
True. At some point we'll have to evaluate all cops/supported styles and remove uncommon stuff to improve the long-term maintainability of the project.

dvandersluis:
I just want to point out that the style on the other cops are symmetrical (applies to both left and right parens) whereas this one is not (only applying to right parens). If we add this I still think it should also be symmetrical to match?

itay-grudev:
@dvandersluis Well, if you agree to merge it then I'll spend the extra time to make those changes and test them, so it's consistent with the other two cops.

dvandersluis:
Up to @bbatsov if we want to continue with this or deprecate the style from the other cops I guess.

bbatsov:
I've decided to continue with adding `compact` here, as the less disruptive change for the users. And to be more careful about the styles we accept going forward as well. :D

dvandersluis:
Alright @itay-grudev please update to make this symmetrical then!

itay-grudev:
@dvandersluis I will finish it later this week and I will ping you when it's ready. Thanks a lot guys!

itay-grudev:
@dvandersluis Implemented left side compaction. I also moved the correction code inside each respective style processor because I started tripping `Metrics/MethodLength` and needed to shorten the code.

lib/rubocop/cop/layout/space_inside_parens.rb
```diff
@@ -33,6 +33,27 @@ module Layout
       #   g = ( a + 3 )
       #   y()
       #
+      # @example EnforcedStyle: compact
+      #   # The `compact` style enforces that parentheses have a space at the
+      #   # beginning with the exception that successive right parentheses are allowed.
```
bbatsov, line 38: I think you should remove "right" here. It reads better to me without it.

bbatsov (commented):
Looks good, apart from my small remark.

lib/rubocop/cop/layout/space_inside_parens.rb
itay-grudev: On it.

itay-grudev: Done.

bbatsov (approved):

PR #10027, opened 2021-08-19
[Fix #10024] Fix an incorrect auto-correct for `Style/RedundantSelfAssignmentBranch`

### LINKED ISSUE #10024: Bug with `Style/RedundantSelfAssignmentBranch`
Multiline conditional assignments are firing the `Style/RedundantSelfAssignmentBranch` rule. If this rule is applied, it modifies the conditional. In the `falsey` case, those conditionals evaluate to `nil`, and then `nil` is assigned (instead of no assignment).

--------

## Expected behavior

The `Style/RedundantSelfAssignmentBranch` does not fire for multiline if/else conditional assignments. Basically the cases covered by 

https://github.com/rubocop/rubocop/commit/07c8ef8dd1c0c2f0b368c8010c17989b3f47883b#diff-9aa30a608c9b3f7d4c969285ada36ce789273d243e71262eeac97ae5a93ae1adR26-R34

## Actual behavior

The rules _do_ fire, leading to incorrect code

Using the configuration

```yaml
Style/RedundantSelfAssignmentBranch:
  Enabled: true

Style/IfUnlessModifier:
  Enabled: false
```

to isolate the behavior to only this cop,

1. Create test file:

```ruby
# frozen_string_literal: true
# join_if.rb

require 'test/unit/assertions'
include Test::Unit::Assertions

url = 'foo'
path = nil

url = if path
        [url, path].join('/')
      else
        url
      end

assert_equal('foo', url)
```
2. run the file `ruby join_if.rb`. Test passes

2. Run `bundle exec rubocop -a` with `Style/RedundantSelfAssignmentBranch` enabled. Cop reports 
```bash
join_if.rb:12:9: C: [Corrected] Style/RedundantSelfAssignmentBranch: Remove the self-assignment branch.
        url
        ^^^

2 files inspected, 2 offenses detected, 1 offense corrected
```
3.  Code is modified to 
```ruby
# frozen_string_literal: true
# join_if.rb

require 'test/unit/assertions'
include Test::Unit::Assertions

url = 'foo'
path = nil

url = if path
        [url, path].join('/')
      end

assert_equal('foo', url)
```
4. Run test again
5. Test fails

The modification fails because the multiline conditional expression returns `nil` and that's what is assigned to the variable.

If the change recommended were instead (which perhaps is what `Style/IfUnlessModifier` does)

```
url = [url, path].join('/') if path
```

it would be correct, but as recommended / auto-corrected, it is not.

## RuboCop version

```
$ bundle exec rubocop -V
1.19.0 (using Parser 3.0.2.0, rubocop-ast 1.10.0, running on ruby 2.7.1 x86_64-darwin20)
```

koic:
Fixes #10024.

This PR fixes an incorrect auto-correct for `Style/RedundantSelfAssignmentBranch` when using multiline `if` / `else` conditional assignment.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_incorrect_autocorrect_for_redundant_self_assignment_branch.md
+++ b/changelog/fix_incorrect_autocorrect_for_redundant_self_assignment_branch.md
@@ -1,0 +1,1 @@
+* [#10024](https://github.com/rubocop/rubocop/issues/10024): Fix an incorrect auto-correct for `Style/RedundantSelfAssignmentBranch` when using multiline `if` / `else` conditional assignment. ([@koic][])
--- a/lib/rubocop/cop/style/redundant_self_assignment_branch.rb
+++ b/lib/rubocop/cop/style/redundant_self_assignment_branch.rb
@@ -35,11 +35,11 @@ module RuboCop
 
         def on_lvasgn(node)
           variable, expression = *node
-          return unless expression&.if_type?
-          return unless expression.ternary? || expression.else?
+          return unless use_if_and_else_branch?(expression)
 
           if_branch = expression.if_branch
           else_branch = expression.else_branch
+          return if inconvertible_to_modifier?(if_branch, else_branch)
 
           if self_assign?(variable, if_branch)
             register_offense(expression, if_branch, else_branch, 'unless')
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #10047, opened 2021-08-26
[Fix #7849] Add new `Lint/AmbiguousOperatorPrecedence` cop

### LINKED ISSUE #7849: New Cop Idea: Explicit operator precedence

This is a request for new cop based on @bbatsov's comment [here](https://github.com/rubocop-hq/rubocop/pull/7847#discussion_r403443811).

The cop should check for explicit parens when operators with varying precedence are used in a statement.

```ruby
# bad
a && b || c
10 + 20 * 3

# good
(a && b) || c
10 + (20 * 3)
```

NA

## Additional context

We need to decide if the cop should consider all operators or only commonly used ones.

dvandersluis:
As per @bbatsov's request a year ago (#7849), adds a cop that requires parentheses around operations within an expression with multiple operations of different precedence. It does not consider comparison or unary operations.

Please take a look at the second commit to see how rubocop itself is changed by this cop (it looks pretty good to me but if anything stands out we can customize).

-----------------

* [x] Added tests.


bbatsov:
Let's go ahead with this cop and see how it goes. We can still change the name before the next release if someone comes up with a better name.

PR #10054, opened 2021-08-27
Fix a false positive for `Layout/SpaceAroundOperators`

koic:
This PR fixes a false positive for `Layout/SpaceAroundOperators` when match operators between `<<` and `+=`.

This change makes the following behavior consistent:

## before

Does not register an offense when aligned between `<<`.

```consle
% cat example.rb
# It does not register an offense.
x  << foo
yz << bar
```

An offense is registered when either is updated to `+=`.

```console
% cat example.rb
# It does not register an offense.
x  << foo
yz += bar

% bundle exec rubocop --only Layout/SpaceAroundOperators
(snip)

example.rb:2:4: C: [Correctable] Layout/SpaceAroundOperators: Operator
<< should be surrounded by a single space.
x  << foo
^^

```

## After

Both does not register an offense.

-----------------

* [x] Added tests.


bbatsov:
Looks good. Thanks!

PR #10062, opened 2021-08-31
[Fix #10061] Fix a false positive for `Style/RedundantSort`

### LINKED ISSUE #10061: RedundantSort rule not working properly in case of max_by(&:size)
The goal:

`return the last of most sized words in the string`

A possible way to solve it:

```ruby
"hello world".split(/\W+/).sort_by(&:size).last
# => "world"
```

This code raise `rubocop` offense:

```
RuboCop: Use `max_by` instead of `sort_by...last`. [Style/RedundantSort]
```

If we apply the suggested replacement, then we got a different result:

```ruby
"hello world".split(/\W+/).max_by(&:size)
# => "hello"
```

--------

## Expected behavior

No suggest replacement when the result going to change.

## Actual behavior

Suggest replacement according to `Style/RedundantSort` rule with result changes.

## RuboCop version

```
1.18.4 (using Parser 3.0.2.0, rubocop-ast 1.9.0, running on ruby 3.0.2 x86_64-linux)
  - rubocop-rails 2.11.3
  - rubocop-rspec 2.4.0
```

koic:
Fixes #10061.

This PR fixes a false positive for `Style/RedundantSort` when using `size` method in the block.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_style_redundant_sort.md
+++ b/changelog/fix_a_false_positive_for_style_redundant_sort.md
@@ -1,0 +1,1 @@
+* [#10061](https://github.com/rubocop/rubocop/issues/10061): Fix a false positive for `Style/RedundantSort` when using `size` method in the block. ([@koic][])
--- a/lib/rubocop/cop/style/redundant_sort.rb
+++ b/lib/rubocop/cop/style/redundant_sort.rb
@@ -76,21 +76,36 @@ module RuboCop
 
         def on_send(node)
           if (sort_node, sorter, accessor = redundant_sort?(node.parent))
+            return if use_size_method_in_block?(sort_node)
+
             ancestor = node.parent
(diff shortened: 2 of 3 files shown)


bbatsov:
Hmm, why is this behaving differently for strings?

koic:
Oops! That's my mistake. In fact, it is a false positive because using `size` method in the block. I've update this PR.

bbatsov:
Thanks!

Mifrill:
@koic Thanks! It was so quick!

koic:
@Mifrill You're welcome!

Drenmi:
Is this the correct approach? If I understand it correctly, this is due to how a certain Ruby version handles ties differently between two different sorting methods. It seems a bit dangerous to rely on this tie-breaking result in our programs to begin with, as it can change between versions (or platforms, or implementations) based Ruby on internals. 😅

PR #10064, opened 2021-09-05
Style/ExplicitBlockArgument: do not assume the block parameter name

casperisfine:
The autocorector would generate invalid code if the old code already defined the block parameter with any other name than `&block`.


dvandersluis:
Can you please add a changelog entry?

casperisfine:
@dvandersluis done!

dvandersluis:
Thanks but we actually don't update `CHANGELOG.md` manually. Please see https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format

casperisfine:
Done.

koic:
Thanks!

PR #10077, opened 2021-09-13
[Fix #10066] Fix how `MinDigits` is calculated for `Style/NumericLiterals` when generating a configuration file

### LINKED ISSUE #10066: Inline disabling of Style/NumericLiterals is ignored when generating MaxLength in the todo file
Using `# rubocop:disable Style/NumericLiterals` inline, works fine when running `rubocop`.

However, the length of the disabled literal is used by the logic that creates `.rubocop_todo.yml` file. This is a problem when the literal that has been excluded for the cop is longer than the existing value of `MinDigits` in the `todo` file. This means that the `MinDigits` parameter can get set to something larger than is desired, which eventually means it's impossible to use `# rubocop:disable Style/NumericLiterals` on numbers that are longer than `MinDigits` as I go on to describe.

--------

## Expected behavior

Running `rubocop --regenerate-todo` should not change the `MaxLength` value.

## Actual behavior

Running `rubocop --regenerate-todo` increases the value of `MinDigits` to be one more than the length of the literal that we have instructed the `Style/NumericLiterals` to ignore.

This then leads to follow-up problems - the next run of `rubocop` will not be green due to an un-necessary `disable` comment being detected, and removing the `disable` comment is undesirable as we intended to mark that specific literal is being OK.

1. Create a 10 digit constant somewhere `FIRST_VIOLATION = 1234567890`
2. Run `rubocop --regenerate-todo` which should generate something like
```
# Offense count: 1
# Cop supports --auto-correct.
# Configuration parameters: Strict.
Style/NumericLiterals:
  MinDigits: 11
```
3. Create a 20 digit constant somewhere, but add an inline disable.
```ruby
SECOND_VIOLATION = 12345678901234567890`
```
4. Run `rubocop` and everything is green. It's all good and as desired up to this point.
5. Now run `rubocop --regenerate-todo` again. It will get updated to
```
# Offense count: 2
# Cop supports --auto-correct.
# Configuration parameters: Strict.
Style/NumericLiterals:
  MinDigits: 21
```
6. Run `rubocop` and it doesn't run succesfully - it tells you the `disable` is no longer required.
7. Remove the inline `disable` and `rubocop` will return green. However, now the special case which we want to allow, is no longer explicitly allowed and is now part of the `todo` file.

## RuboCop version

```
1.18.4 (using Parser 3.0.2.0, rubocop-ast 1.8.0, running on ruby 2.7.1 x86_64-darwin19)
  - rubocop-rails 2.11.3
```

dvandersluis:
Previously, `MinDigits` is incremented regardless of if the offense is disabled or not.

Fixes #10066.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_fix_how_mindigits_is_calculated_for.md
+++ b/changelog/fix_fix_how_mindigits_is_calculated_for.md
@@ -1,0 +1,1 @@
+* [#10066](https://github.com/rubocop/rubocop/issues/10066): Fix how `MinDigits` is calculated for `Style/NumericLiterals` when generating a configuration file. ([@dvandersluis][])
--- a/lib/rubocop/rspec/cop_helper.rb
+++ b/lib/rubocop/rspec/cop_helper.rb
@@ -49,7 +49,7 @@ module CopHelper
     team = RuboCop::Cop::Team.new([cop], nil, raise_error: true)
     report = team.investigate(processed_source)
     @last_corrector = report.correctors.first || RuboCop::Cop::Corrector.new(processed_source)
-    report.offenses
+    report.offenses.reject(&:disabled?)
   end
 end
 
--- a/lib/rubocop/cop/style/numeric_literals.rb
+++ b/lib/rubocop/cop/style/numeric_literals.rb
@@ -58,18 +58,17 @@ module RuboCop
 
           case int
           when /^\d+$/
-            return unless (self.min_digits = int.size + 1)
-
-            register_offense(node)
+            register_offense(node) { self.min_digits = int.size + 1 }
           when /\d{4}/, short_group_regex
-            return unless (self.config_to_allow_offenses = { 'Enabled' => false })
-
-            register_offense(node)
+            register_offense(node) { self.config_to_allow_offenses = { 'Enabled' => false } }
           end
         end
 
-        def register_offense(node)
-          add_offense(node) { |corrector| corrector.replace(node, format_number(node)) }
+        def register_offense(node, &_block)
+          add_offense(node) do |corrector|
+            yield
+            corrector.replace(node, format_number(node))
+          end
         end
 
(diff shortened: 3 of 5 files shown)


lib/rubocop/rspec/cop_helper.rb
```diff
@@ -49,7 +49,7 @@ def _investigate(cop, processed_source)
     team = RuboCop::Cop::Team.new([cop], nil, raise_error: true)
     report = team.investigate(processed_source)
     @last_corrector = report.correctors.first || RuboCop::Cop::Corrector.new(processed_source)
-    report.offenses
+    report.offenses.reject(&:disabled?)
```
dvandersluis, line 52: In order to test this, I needed to update `expect_offense` and `expect_no_offenses` to not consider disabled offenses. There was only one test (in `spec/rubocop/cop/layout/line_length_spec.rb`) that was affected by this change.

changelog/fix_fix_how_mindigits_is_calculated_for.md
```diff
@@ -0,0 +1 @@
+* [#10066](https://github.com/rubocop/rubocop/pull/10066): Fix how `MinDigits` is calculated for `Style/NumericLiterals` when generating a configuration file. ([@dvandersluis][])
```
koic, line 1: That's just my two cents :-)

```suggestion
* [#10066](https://github.com/rubocop/rubocop/issues/10066): Fix how `MinDigits` is calculated for `Style/NumericLiterals` when generating a configuration file. ([@dvandersluis][])
```

dvandersluis: Good catch, thanks. Updated now!

bbatsov:
Nice!

PR #10085, opened 2021-09-15
[Fix #10084] Add new `Lint/RequireRelativeSelfPath` cop

### LINKED ISSUE #10084: Cop idea: mark self-`require_relative` as bad code

This code in file `my_class.rb`
```
# frozen_string_literal: true

require_relative 'my_class'

# Class which puts foo
class MyClass
  def foo
    puts 'foo'
  end
end

```

`require_relative 'my_class'` is completely useless (require itself) - rubocop should mark it as bad smell

## Additional context

```
# rubocop -V
1.21.0 (using Parser 3.0.2.0, rubocop-ast 1.11.0, running on ruby 3.0.2 x86_64-linux)

```

koic:
Fixes #10084.

This PR adds new `Lint/RequireRelativeSelfPath` cop.
It checks for uses of `require_relative` with self file path argument.

```ruby
# bad

# foo.rb
require_relative 'foo'
require_relative 'bar'

# good

# foo.rb
require_relative 'bar'
```

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -2045,6 +2045,11 @@ Lint/RequireParentheses:
   Enabled: true
   VersionAdded: '0.18'
 
+Lint/RequireRelativeSelfPath:
+  Description: 'Checks for uses of `require_relative` with self file path argument.'
```
dvandersluis, line 2049:
```suggestion
  Description: 'Checks for uses a file requiring itself with `require_relative`.'
```

koic: I've updated it. Thank you!

dvandersluis (approved):
This looks good to me! I'm wondering if there are more edge cases we should care about (since `require_relative` can have paths with `.` and `..` in them) but it might make this overly complicated for no tangible benefit.

koic:
> I'm wondering if there are more edge cases we should care about (since `require_relative` can have paths with `.` and `..` in them) but it might make this overly complicated for no tangible benefit.

Yeah, I have the same idea. It doesn't complicate the implementation because of the edge cases that user probably rarely encounter.

lib/rubocop/cop/lint/require_relative_self_path.rb
```diff
@@ -0,0 +1,45 @@
...
+    module Lint
+      # Checks for uses a file requiring itself with `require_relative`.
+      #
+      # @example
+      #
+      #   # bad
+      #
+      #   # foo.rb
+      #   require_relative 'foo'
+      #   require_relative 'bar'
```
tejasbubane, line 14:
```suggestion

```
I would suggest removing this because `bar` does not raise offense, only `foo` does.

spec/rubocop/cop/lint/require_relative_self_path_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  it 'does not register an offense when using `require_relative` without self file path argument' do
+    expect_no_offenses(<<~RUBY, 'foo.rb')
+      require_relative 'bar'
+    RUBY
+  end
+
+  it 'does not register an offense when using `require_relative` without argument' do
+    expect_no_offenses(<<~RUBY, 'foo.rb')
+      require_relative
```
tejasbubane, line 36: This will always raise `ArgumentError` and hence should be detected by rubocop. Maybe we change the cop name to `Lint/RequireRelativePath` and accomodate this case too?

lib/rubocop/cop/lint/require_relative_self_path.rb
koic: It would not be correct for `foo` to be `bar`.  I'd like to keep this for contrasted with the good case.

spec/rubocop/cop/lint/require_relative_self_path_spec.rb
koic: This test case has confirmed that it does not cause an error with a valid grammar for Ruby. And I think it should do method argument checking with RBS instead of RuboCop, so the cop name and role will remain the same.

bbatsov:
Nicely done!

ShockwaveNN:
Thank you for fastest implementation of idea )

koic:
@ShockwaveNN You're welcome!

PR #10086, opened 2021-09-15
Guard against `expect_corrections` with no changes

dvandersluis:
If autocorrection doesn't change anything, `expect_no_corrections` is clearer about what will happen (vs having to figure out if there is some small change that is not immediately evident). This change also fixes the existing specs that use the wrong method.


koic:
Cool! 👍

PR #10088, opened 2021-09-15
`Lint/BooleanSymbol` should be marked as `SafeAutoCorrect: false` rather than `Safe: false`

dvandersluis:
Although the correction is unsafe, the cop itself will not find false positives.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1491,9 +1491,9 @@ Lint/BinaryOperatorWithIdenticalOperands:
 Lint/BooleanSymbol:
   Description: 'Check for `:true` and `:false` symbols.'
   Enabled: true
-  Safe: false
+  SafeAutoCorrect: false
   VersionAdded: '0.50'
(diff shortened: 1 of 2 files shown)


config/default.yml
```diff
@@ -1491,7 +1491,7 @@ Lint/BinaryOperatorWithIdenticalOperands:
 Lint/BooleanSymbol:
   Description: 'Check for `:true` and `:false` symbols.'
   Enabled: true
-  Safe: false
+  SafeAutoCorrect: false
   VersionAdded: '0.50'
   VersionChanged: '0.83'
```
koic, line 1496: Can you update it and add a changelog entry?

```suggestion
  VersionChanged: '<<next>>'
```

bbatsov:
This deserves a changelog entry IMO.

config/default.yml
dvandersluis: Updated.

PR #10091, opened 2021-09-16
[Fix #10090] Fix a false negative for `Style/ArgumentsForwarding`

### LINKED ISSUE #10090: Forward all arguments with `Style/ArgumentsForwarding`
We have uses of `*args` and `**args` being forwarded in our codebase. They occur like

```ruby
def self.call(*args)
  new(*args).call
end
```

or 

```ruby
def self.call(**args)
  new(**args).call
end
```

I ideally want them to both be collapsed into

```ruby
def self.call(...)
  new(...).call
end
```

I wanted to enable `Style/ArgumentsForwarding` to prefer using `...` instead, however, I found that enabling `Style/ArgumentsForwarding` did not catch replacing `*args` with `...`. I then tried it with the configuration option `AllowOnlyRestArgument: false` on the cop and then `*args` are suggested to be replaced with `...`.

I could not find an option that suggests replacing `**args` with `...`.

koic:
Fixes #10090.

This PR fixes a false negative for `Style/ArgumentsForwarding` when using only kwrest arg.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_negative_for_style_arguments_forwarding.md
+++ b/changelog/fix_false_negative_for_style_arguments_forwarding.md
@@ -1,0 +1,1 @@
+* [#10090](https://github.com/rubocop/rubocop/issues/10090): Fix a false negative for `Style/ArgumentsForwarding` when using only kwrest arg. ([@koic][])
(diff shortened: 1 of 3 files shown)


dvandersluis (approved):

koic:
ruby-head CI matrix is failing, but it is not related to this PR change.
https://app.circleci.com/pipelines/github/rubocop/rubocop/5211/workflows/706c6297-bc62-420a-a61f-712274df787d/jobs/197194

I have reported this error to https://bugs.ruby-lang.org/issues/18177.

PR #10094, opened 2021-09-16
[Fix #8431] Add `Safety` section to documentation for all cops that are `Safe: false` or `SafeAutoCorrect: false`

### LINKED ISSUE #8431: Document safety issues
It would be nice that Cops that are not safe of have a not safe auto-correction have a section in their doc about this

E.g. for FrozenStringLiteral

```ruby
=== Safety
Unsafe since any code that mutates these literals will fail:
    # bad, not safely corrected:
    str = 'hello'
    str << 'world' # => Fails if magic comment is present.

    # good
    # frozen_string_literal: true

    str = +'hello'  # Notice the '+' to obtain unfrozen litteral here
    puts str << ', world'
```

dvandersluis:
Documents safety concerns for every cop that is either `Safe: false` or `SafeAutoCorrect: false`. This documentation will show up as a separate subsection in each cop's documentation once the .adoc files are generated.

For example: 
image

In order to make this happen, I added a custom `@safety` YARD tag that is used to process the documentation, which should make it simple when creating a new unsafe cop (I also updated the `rake new_cop` template) or changing a cop to be unsafe.

The text for the safety section comes either from the text that already existed in the cop (🙌 to everyone who added a reason when marking a cop unsafe); looking through many old PRs in order to try to determine why the cop was made unsafe; or using my best guess if I could not find a better reason. While I was going through documentation I took the opportunity to make some minor tweaks as well.

I tried writing a `project_spec` test to ensure a `@safety` block is added for cops that are unsafe, but could not find a way to get the yard code object for a class (the documentation generator relies on yard's generation rake task, which I did not want to run in a spec). If anyone has any suggestions please let me know!

Altogether, 66 cops were updated by this change.

Fixes #8431.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/string_hash_keys.rb
+++ b/lib/rubocop/cop/style/string_hash_keys.rb
@@ -6,6 +6,10 @@ module RuboCop
       # This cop checks for the use of strings as keys in hashes. The use of
       # symbols is preferred instead.
       #
+      # @safety
+      #   This cop is unsafe because while symbols are preferred for hash keys,
+      #   there are instances when string keys are required.
+      #
       # @example
       #   # bad
       #   { 'one' => 1, 'two' => 2, 'three' => 3 }
--- a/.yardopts
+++ b/.yardopts
@@ -1,2 +1,3 @@
 --markup markdown
 --hide-void-return
+--tag safety:"Cop Safety Information"
--- a/changelog/change_add_safety_section_to_all_cops_that_are.md
+++ b/changelog/change_add_safety_section_to_all_cops_that_are.md
@@ -1,0 +1,1 @@
+* [#8431](https://github.com/rubocop/rubocop/issues/8431): Add `Safety` section to documentation for all cops that are `Safe: false` or `SafeAutoCorrect: false`. ([@dvandersluis][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -3708,7 +3708,7 @@ Style/InPatternThen:
 Style/InfiniteLoop:
   Description: >-
                  Use Kernel#loop for infinite loops.
-                 This cop is unsafe in the body may raise a `StopIteration` exception.
+                 This cop is unsafe if the body may raise a `StopIteration` exception.
   Safe: false
   StyleGuide: '#infinite-loop'
   Enabled: true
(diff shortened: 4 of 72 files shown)


lib/rubocop/cop/style/string_hash_keys.rb
```diff
@@ -6,6 +6,10 @@ module Style
       # This cop checks for the use of strings as keys in hashes. The use of
       # symbols is preferred instead.
       #
+      # @safety
+      #   This cop is unsafe because while symbols are prefered for hash keys,
```
koic, line 10:
```suggestion
      #   This cop is unsafe because while symbols are preferred for hash keys,
```

koic (approved):
Fantastic work! This looks very user friendly! IMO, it's better to have a project_spec for this, but this PR changes themself are well worth!

lib/rubocop/cop/style/string_hash_keys.rb
dvandersluis: Fixed, thanks!

dvandersluis:
> IMO, it's better to have a project_spec for this

So as far as I can tell, we need YARD to have regenerated the `.yardoc` folder with the latest changes, and then we can load it in order to check the project. This _should_ work for `rake default` because it calls `documentation_syntax_check` which calls `yard_for_generate_documentation`. It should be reasonably fast too.

However, if `rake spec` is run without the yardoc being generated first, tests will fail. I'm not sure if we should do this as a separate test run as part of `rake default`? Or somehow exclude it when running `rake spec`? And then we'd have to figure out how to handle it in CI as well.

koic:
> IMO, it's better to have a project_spec for this

Thank you for investigating. Unfortunately I'm not sure about it right away 💦 (and so far I don't have a strong opinion yet.)

> might be worthwhile to do all this as a separate PR, once we have decided on a solution

Yeah, I think we can add that spec later when we are really in trouble. First of all, I think the content of the current PR is sufficient 😃

dvandersluis:
Sounds good to me!

bbatsov (approved):
Fantastic change!

bbatsov:
> If we can solve these, here is a scaffold for a test: https://gist.github.com/dvandersluis/0d692eafd4696f542f5bd6abd1840043. We can conceivably extend this to check for missing @example tags too, etc. (might be worthwhile to do all this as a separate PR, once we have decided on a solution).

I agreed. I've long wanted to automate checking the docs for some required structure and it'd be great to eventually get there. Step by step... :-)

PR #10100, opened 2021-09-20
Add new Style/NumberedParameters cop

Hugo-Hache:
Ruby 2.7 introduced numbered parameters. They come in handy for single line block where naming a variable was done only to please the parser but did not bring any value. However for longer blocks they can obfuscate the object manipulated.

This cop allows to restrict their use to single line block. 

-----------------

* [ ] ~~Commit message starts with `[Fix #issue-number]` (if the related issue exists).~~
* [x] Added tests.


Hugo-Hache:
👋 the Rubocop team, after discussing with my team the way we wanted to use the numbered parameters introduced by Ruby 2.7, I noticed there was no related cop.

Being new to this repo, I checked the docs you wrote for contributors, and tried to base the (really simple) code of this cop on existing ones. However I certainly made rookie mistakes, so please let me know what they are, and I'll do my best to fix them.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#10100](https://github.com/rubocop/rubocop/issues/10100): Add new `Style/NumberedParameters` cop. ([@Hugo-Hache][])
+
```
dvandersluis, line 8: You already created the `changelog/` file, so this is unneeded!

dvandersluis:
This cop is very straightforward, and looks like a good addition. I can see some additional styles being useful (for instance, disallowing numbered params altogether).

My only question is: does having this cop operate on the number of lines the block is vs the number of statements the block contains valid? (FWIW, the cop could still be pretty simple checking for more than one statement, you'd just need to check if the child of the block node is a `begin` node).

This also could probably autocorrect pretty easily if desired, see `Style/BlockDelimiters` for a similar example.

koic:
I think that the cop name `Style/NumberedParameters` makes users imagine denying numbered parameter.

The cop will probably have an option of always denying numbered parameter.

So, I'm not sure whether it's better to allow only single line by default. And it is doubtful that single lines will be auto-corrected to multi lines instead of being denied. It does not resolve nameless, so this cop may be `Naming` dept instead of `Style` dept.

IMO, use of numbered parameter is understandable to me when numbered parameter is used in disposable script, but I'm very worried about using numbered parameter in maintenance code. OTOH, I've met several developers who like parameters.

So, this rule probably needs discussion in the Ruby Style Guide first.
https://github.com/rubocop/ruby-style-guide

CHANGELOG.md
Hugo-Hache: Indeed, fixed

lib/rubocop/cop/style/numbered_parameters.rb
```diff
@@ -0,0 +1,36 @@
...
+module RuboCop
+  module Cop
+    module Style
+      # Use numbered parameters only for single-line block.
+      #
+      # @safety
+      #   Delete this section if the cop is not unsafe (`Safe: false` or
+      #   `SafeAutoCorrect: false`), or use it to explain how the cop is
+      #   unsafe.
+      #
```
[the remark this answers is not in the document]
Hugo-Hache: Remove

Hugo-Hache:
Thanks so much for your reactivity. 

Answers for @dvandersluis:
* Additional style disallow: great idea, just added it
* Number of lines vs statements: I initially chose lines over statements (or complexity) as my concern was focused on readability. Single line put beside the array name and numbered parameter, so it's very easy to follow what the numbered parameter is refering to.   
* Autocorrect: not sur how we could auto-correct numbered parameters, adding parameters with default name like `a`, `b`, etc? Not sure it would put the code in a much better state than the initial numbered parameters

@koic based on your advice I just opened [an issue on the subject](https://github.com/rubocop/ruby-style-guide/issues/885) in the style guide

dvandersluis:
> IMO, use of numbered parameter is understandable to me when numbered parameter is used in disposable script, but I'm very worried about using numbered parameter in maintenance code. OTOH, I've met several developers who like numbered parameter.

I agree, I would definitely disable numbered parameters in my work codebase it it was possible.

dvandersluis:
> Autocorrect: not sur how we could auto-correct numbered parameters, adding parameters with default name like a, b, etc? Not sure it would put the code in a much better state than the initial numbered parameters

My suggestion for autocorrect was only for changing multiple lines to a single line if we're going that way. Definitely don't change variables, that wouldn't be safe, and wouldn't work well anyways.

lib/rubocop/cop/style/numbered_parameters.rb
```diff
@@ -0,0 +1,41 @@
...
+
+        MSG = 'Avoid using numbered parameters.'
+        MSG_MULTI_LINE = 'Avoid using numbered parameters for multi-line blocks.'
+
+        minimum_target_ruby_version 2.7
+
+        def on_numblock(node)
+          return unless style == :disallow || node.multiline?
+
+          add_offense(node, message: style == :disallow ? MSG : MSG_MULTI_LINE)
```
dvandersluis, line 36: I tend to avoid `unless` with boolean logic, how about this?

```suggestion
          if style == :disallow
            add_offense(node, message: MSG)
          elsif node.multiline?
            add_offense(node, message: MSG_MULTILINE)
          end
```

dvandersluis (approved):
I left one minor nitpick comment, but this looks good to me. Once all of @koic's concerns are resolved then I think this is good to go.

lib/rubocop/cop/style/numbered_parameters.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Use numbered parameters only for single-line block.
```
bbatsov, line 6: I guess this needs to be updated to reflect the more general nature of the cop today.

bbatsov:
> My only question is: does having this cop operate on the number of lines the block is vs the number of statements the block contains valid? (FWIW, the cop could still be pretty simple checking for more than one statement, you'd just need to check if the child of the block node is a begin node).

I was also thinking about this, but we've already targeted single-line blocks in the past, so I guess the current approach is fine. 

I think this cop should also have some limit on the number of numbered params in a block (regardless of the length of the block), as things can get very dicey with 2+ numbered params IMO. This, however, would imply an "allow" style as well, so perhaps a separate cop is in order down the road (e.g. `NumberOfNumberedParams` or something like this).

dvandersluis:
> This, however, would imply an "allow" style as well, so perhaps a separate cop is in order down the road (e.g. NumberOfNumberedParams or something like this).

I opened #10111 to add a cop for this :grin:

Hugo-Hache:
I'm on the road this week, I'll incorporate your comments on Monday morning. Thanks again for your quick feedbacks!

lib/rubocop/cop/style/numbered_parameters.rb
Hugo-Hache: Updated, ok for you? (I updated default.yml as well)

lib/rubocop/cop/style/numbered_parameters.rb
Hugo-Hache: Great suggestion, amended

bbatsov:
Thanks!

dvandersluis:
I love how simple this cop is, great job @Hugo-Hache!

PR #10102, opened 2021-09-20
Add new `Security/IoMethods` cop

koic:
## Summary

Follow up to https://github.com/rubocop/rubocop/pull/9695#issuecomment-820151271.

This PR adds new `Security/IoMethods` cop.

It checks for the first argument to `IO.read`, `IO.binread`, `IO.write`, `IO.binwrite`, `IO.foreach`, and `IO.readlines`.

If argument starts with a pipe character (`'|'`) and the receiver is the `IO` class, a subprocess is created in the same way as `Kernel#open`, and its output is returned.
Consider to use `File.read` to disable the behavior of subprocess invocation.

This cop is unsafe because false positive will occur if the variable passed as the first argument is a command that is not a file path.

```ruby
# bad
IO.read(path)
IO.read('path')

# good
File.read(path)
File.read('path')
IO.read('| command') # Allow intentional command invocation.
```

## Other Information

Below are links related to this cop.

- https://github.com/rubygems/rubygems/pull/4530
- https://github.com/ruby/ruby/pull/4579

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_add_new_security_io_read_cop.md
+++ b/changelog/new_add_new_security_io_read_cop.md
@@ -1,0 +1,1 @@
+* [#10102](https://github.com/rubocop/rubocop/pull/10102): Add new `Security/IoMethods` cop. ([@koic][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2739,6 +2739,14 @@ Security/Eval:
   Enabled: true
   VersionAdded: '0.47'
 
+Security/IoMethods:
+  Description: >-
+                 Checks for the first argument to `IO.read`, `IO.binread`, `IO.write`, `IO.binwrite`,
+                 `IO.foreach`, and `IO.readlines`.
+  Enabled: pending
+  Safe: false
+  VersionAdded: '<<next>>'
+
 Security/JSONLoad:
   Description: >-
                  Prefer usage of `JSON.parse` over `JSON.load` due to potential
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -639,6 +639,7 @@ require_relative 'rubocop/cop/style/yoda_condition'
 require_relative 'rubocop/cop/style/zero_length_predicate'
 
 require_relative 'rubocop/cop/security/eval'
+require_relative 'rubocop/cop/security/io_methods'
 require_relative 'rubocop/cop/security/json_load'
 require_relative 'rubocop/cop/security/marshal_load'
 require_relative 'rubocop/cop/security/open'
--- a/lib/rubocop/cop/security/io_methods.rb
+++ b/lib/rubocop/cop/security/io_methods.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Security
+      # Checks for the first argument to `IO.read`, `IO.binread`, `IO.write`, `IO.binwrite`,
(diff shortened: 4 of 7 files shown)


spec/rubocop/cop/security/io_read_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      expect_no_offenses(<<~RUBY)
+        #{code}
+      RUBY
+    end
+
+    it "does not register an offense when using `#{code}`" do
+      expect_no_offenses(<<~RUBY)
+        #{code}
+      RUBY
+    end
```
dvandersluis, line 28: I think this is probably here twice by accident?

koic: Oops! This is a mistake and fixed it. Thanks for your review!

lib/rubocop/cop/security/io_read.rb
```diff
@@ -0,0 +1,47 @@
...
+      #
+      #   # good
+      #   File.read(path)
+      #   File.read('path')
+      #   IO.read('| command') # Allow intentional command invocation.
+      #
+      class IoRead < Base
+        extend AutoCorrector
+
+        MSG = 'Use `File` instead of `IO`.'
```
dvandersluis, line 31: I'm wondering if this can be more descriptive. Something like <code>`File.read(path)` is safer than `IO.read(path)</code>. Up to you, but just my thoughts to make this more clear.

koic: Great! I like this suggested message ❤️

spec/rubocop/cop/security/io_read_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    it_behaves_like 'offense', 'readlines(path)'
+  end
+
+  context 'when using `IO` receiver and string argument' do
+    it_behaves_like 'offense', 'read("command")'
+    it_behaves_like 'offense', 'write("command", "hi")'
+    it_behaves_like 'offense', 'binwrite("command", "hi")'
+    it_behaves_like 'offense', 'foreach("command") { |x| puts x }'
+    it_behaves_like 'offense', 'readlines("command")'
+  end
```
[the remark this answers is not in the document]
koic: > These all should be preceded by IO. right?

Yep, I updated these tests.

bbatsov:
The cop name is a bit misleading, as it checks for some writes as well. I'm not sure if `IoMethods` is not a better name, or just something more generic.

lib/rubocop/cop/security/io_read.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Security
+      # Checks for the first argument to `IO.read`, `IO.binread`, `IO.write`, `IO.binwrite`,
+      # `IO.foreach`, and `IO.readlines`.
+      #
+      # If argument starts with a pipe character (`'|'`) and the receiver is the `IO` class,
```
bbatsov, line 9: Probably you should mention somewhere that `Kernel#open` may allow unintentional command injection, which is the reason IO methods are a security risk.

bbatsov:
@koic ping :-)

bbatsov:
Thanks!

koic:
Thank you too!

PR #10105, opened 2021-09-21
[Fix #10065] Add new `Gemspec/TestFilesAssignment` cop

### LINKED ISSUE #10065: Cop idea: Gemspec/TestFiles

I don't think it's necessary to use the `test_files` attribute in a gemspec in most cases, since it has been removed from the [gemspec's official document](https://guides.rubygems.org/specification-reference/) and seems to be no longer recommended.

Also, if removing `test_files`, then we could receive smaller packed gems.

See also the discussions:
- rubygems/guides#90
- rubygems/rubygems#735

I suggest adding a cop like `Gemspec/TestFiles` that warns the use of `test_files` in a gemspec.
(I'm not sure about this naming... 😅 )

I have no alternatives.

## Additional context

Many gems including Rails have removed `test_files` to reduce a gem size.
(see also rails/rails#22272)

koic:
Fixes #10065.

This PR adds new `Gemspec/TestFilesAssignment` cop.

This cop checks that `test_files =` is not used in gemspec file.
if removing `test_files =`, then user could receive smaller packed gems.

```ruby
# bad
  spec.name = 'your_cool_gem_name'
  spec.test_files = Dir.glob('test/**/*')
end

# bad
  spec.name = 'your_cool_gem_name'
  spec.test_files += Dir.glob('test/**/*')
end

# good
  spec.name = 'your_cool_gem_name'
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -273,6 +273,13 @@ Gemspec/RubyVersionGlobalsUsage:
   Include:
     - '**/*.gemspec'
 
+Gemspec/TestFilesAssignment:
+  Description: Checks that the `test_files` attribute is not set in a gemspec file.
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  Include:
+    - '**/*.gemspec'
+
 #################### Layout ###########################
 
 Layout/AccessModifierIndentation:
--- a/lib/rubocop/cop/gemspec/test_files_assignment.rb
+++ b/lib/rubocop/cop/gemspec/test_files_assignment.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Gemspec
+      # This cop checks that the `test_files` attribute is not set in a gemspec file.
+      # Removing it allows the user to receive smaller packed gems.
+      #
+      # @example
+      #
+      #   # bad
+      #   Gem::Specification.new do |spec|
+      #     spec.name = 'your_cool_gem_name'
+      #     spec.test_files = Dir.glob('test/**/*')
+      #   end
+      #
+      #   # bad
+      #   Gem::Specification.new do |spec|
+      #     spec.name = 'your_cool_gem_name'
+      #     spec.test_files += Dir.glob('test/**/*')
+      #   end
+      #
+      #   # good
+      #   Gem::Specification.new do |spec|
+      #     spec.name = 'your_cool_gem_name'
+      #   end
+      #
+      class TestFilesAssignment < Base
+        include RangeHelp
+        extend AutoCorrector
+
+        MSG = 'Do not set `test_files` in gemspec.'
+
+        # @!method gem_specification(node)
+        def_node_matcher :gem_specification, <<~PATTERN
+          (block
+            (send
+              (const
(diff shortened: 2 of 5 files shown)


config/default.yml
```diff
@@ -273,6 +273,11 @@ Gemspec/RubyVersionGlobalsUsage:
   Include:
     - '**/*.gemspec'
 
+Gemspec/TestFilesAssignment:
+  Description: Checks that `test_files =` is not used in gemspec file.
+  Enabled: pending
+  VersionAdded: '<<next>>'
```
dvandersluis, line 279: Should this have

```yaml
Include:
  - '**/*.gemspec'
```

lib/rubocop/cop/gemspec/test_files_assignment.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Gemspec
+      # This cop checks that `test_files =` is not used in gemspec file.
+      # If removing `test_files =`, then user could receive smaller packed gems.
```
dvandersluis, line 7:
```suggestion
      # This cop checks that the `test_files` attribute is not set in a gemspec file.
      # Removing it allows the user to receive smaller packed gems.
```

lib/rubocop/cop/gemspec/test_files_assignment.rb
```diff
@@ -0,0 +1,77 @@
...
+      #   # good
+      #   Gem::Specification.new do |spec|
+      #     spec.name = 'your_cool_gem_name'
+      #   end
+      #
+      class TestFilesAssignment < Base
+        include RangeHelp
+        extend AutoCorrector
+
+        MSG = 'Do not use `test_files =` in gemspec.'
```
dvandersluis, line 32:
```suggestion
        MSG = 'Do not set `test_files` in gemspec.'
```

dvandersluis (approved):
Looks great! I added a couple wording suggestions but do with them what you'd like 😄

koic:
I updated this PR. Thanks a lot!

lib/rubocop/cop/gemspec/test_files_assignment.rb
```diff
@@ -0,0 +1,77 @@
...
+      #     spec.name = 'your_cool_gem_name'
+      #     spec.test_files += Dir.glob('test/**/*')
+      #   end
+      #
+      #   # good
+      #   Gem::Specification.new do |spec|
+      #     spec.name = 'your_cool_gem_name'
+      #   end
+      #
+      class TestFilesAssignment < Base
```
bbatsov, line 28: Why don't we simply name this "TestFiles"?

bbatsov:
The cop looks good to me as well. I'm just wondering if it doesn't make sense to have one cop that checks for multiple forbidden attributes in a spec. Then we can just have a hash mapping attribute names to related messages and this can be one of them.

bbatsov:
@koic ping :-)

bbatsov:
Seems we completely forgot about this cop.

PR #10111, opened 2021-09-22
Add new `Style/NumberedParametersLimit` cop

dvandersluis:
Follows https://github.com/rubocop/rubocop/pull/10100#issuecomment-924605125.

> @bbatsov: I think this cop should also have some limit on the number of numbered params in a block (regardless of the length of the block), as things can get very dicey with 2+ numbered params IMO. This, however, would imply an "allow" style as well, so perhaps a separate cop is in order down the road (e.g. NumberOfNumberedParams or something like this).

This adds a new `Style/ExcessiveNumberedParameters` cop that allows a maximum number of numbered params in a single block. The `Max` is configurable.

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -3432,6 +3432,12 @@ Style/EvenOdd:
   VersionAdded: '0.12'
   VersionChanged: '0.29'
 
+Style/ExcessiveNumberedParameters:
```
Drenmi, line 3435: I think we can probably drop the `Excessive` prefix, as (as far as I know) all cops are named objectively by what they check, e.g. `LineLength` and not `ExcessiveLineLength`. Perhaps (if needed) replaced by a suffix like in `NumberedParameterCount`. WDYT?

dvandersluis: `NumberedParameterCount` works for me!

bbatsov: Or NumberedParamatersLimit. Not sure what names we used in the past in similar situations.

dvandersluis: I have renamed the cop to `Style/NumberedParametersLimit`.

config/default.yml
```diff
@@ -4135,6 +4135,12 @@ Style/Not:
   VersionAdded: '0.9'
   VersionChanged: '0.20'
 
+Style/NumberedParametersLimit:
+  Description: 'Avoid excessive numbered params in a single block.'
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  Max: 2
```
koic, line 4142: In my opinion, it becomes unclear what it means when it over `_1`.
```suggestion
  Max: 1
```

dvandersluis: Works for me!

dvandersluis:
The default value for `Max` has been changed to 1 now.

lib/rubocop/cop/style/numbered_parameters_limit.rb
```diff
@@ -0,0 +1,43 @@
...
+          parameter = max_count > 1 ? 'parameters' : 'parameter'
+          message = format(MSG, max: max_count, parameter: parameter, count: param_count)
+          add_offense(node, message: message) { self.max = param_count }
+        end
+
+        private
+
+        def max_count
+          # There cannot be more than 9 numbered parameters
+          [cop_config['Max'], 9].min
```
koic, line 38: It may be safer to use `cop_config.fetch('Max', DEFAULT_MAX_VALUE)`.

```diff
+          DEFAULT_MAX_VALUE = 1

...

-          # There cannot be more than 9 numbered parameters
-          [cop_config['Max'], 9].min
+          max = cop_config.fetch('Max', DEFAULT_MAX_VALUE)
+          
+          # There cannot be more than 9 numbered parameters.
+          [max, 9].min
```

dvandersluis: Updated, thanks!

lib/rubocop/cop/style/numbered_parameters_limit.rb
```diff
@@ -0,0 +1,50 @@
...
+  module Cop
+    module Style
+      # This cop detects use of an excessive amount of numbered parameters in a
+      # single block. Having too many numbered parameters can make code too
+      # cryptic and hard to read.
+      #
+      # The cop defaults to registering an offense if there is more than 1 numbered
+      # parameter but this maximum can be configured by setting `Max`.
+      #
+      # @example
```
koic, line 13: Adding the config parameter to the following example may help the user understand :-)

```suggestion
      # @example Max: 1 (default)
```

dvandersluis: Another good suggestion! :grin:

dvandersluis:
@Drenmi @bbatsov @koic any other changes needed here?

bbatsov:
Nope, looks good to me. Thanks!

PR #10119, opened 2021-09-23
[Fix #10118] Fix crash with `Style/RedundantSort` when the block doesn't only contain a single `send` node

dvandersluis:
The check added in #10062 for `size` wasn't handling blocks that aren't just a single send node. Fixes #10118.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #10122, opened 2021-09-23
Update `Style/RedundantSort` to be unsafe

dvandersluis:
In an array where there are multiple elements where `a <=> b == 0`, or where the transformation done by the block has the same result, there will be a different element returned between `sort.last` and `max` (or `sort_by.last` and `max_by`), because `sort.last` will take the last element but `max` will return the first element. As such, this is not a safe change, as the specific element returned will change.

Honestly, we should probably revert #10062 as well because that adds a special case for `size` that doesn't really exist in reality. In that issue, `size` was the catalyst, but the exact same issue would happen with `length`, or any other transformation that results in multiple elements having the same value. If we revert #10062, then #10119 can be closed as well. If this is desired, I'll update this PR to include the revert and add the test case from #10119 which is still valueable.

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/change_update_styleredundantsort_to_be_unsafe.md
+++ b/changelog/change_update_styleredundantsort_to_be_unsafe.md
@@ -1,0 +1,1 @@
+* [#10122](https://github.com/rubocop/rubocop/pull/10122): Update `Style/RedundantSort` to be unsafe, and revert the special case for `size` from [#10061](https://github.com/rubocop/rubocop/pull/10061). ([@dvandersluis][])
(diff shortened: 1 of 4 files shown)


dvandersluis:
@koic are you ok with reverting your change like I proposed?

dvandersluis:
@bbatsov any thoughts?

bbatsov:
I'm fine with your proposal and reverting the extra commit that attempted to fix this before.

dvandersluis:
@bbatsov the special case for `size` is now reverted in addition to the cop being made unsafe.

bbatsov:
Thanks!

PR #10124, opened 2021-09-24
Fix `Layout/RedundantLineBreak` adding extra space within method chains

dvandersluis:
When `Layout/RedundantLineBreak` finds a multiline method chain that can be collapsed to a single line, it adds an extra space between each method. Although the spaces could also be detected by `Layout/SpaceAroundMethodCallOperator`, this change allows the cop to calculate if a correction will fit on the line better (as demonstrated by the fixes in the 2nd commit).

```ruby
foo.
  bar.
  baz
```

Before:
```ruby
foo. bar. baz
```

After:
```ruby
foo.bar.baz
```

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/layout/redundant_line_break_spec.rb
+++ b/spec/rubocop/cop/layout/redundant_line_break_spec.rb
@@ -272,17 +272,58 @@ RSpec.describe RuboCop::Cop::Layout::RedundantLineBreak, :config do
         RUBY
       end
 
-      it 'properly corrects a method chain on multiple lines' do
-        expect_offense(<<~RUBY)
-          foo
-          ^^^ Redundant line break detected.
-            .bar
-            .baz
-        RUBY
+      context 'method chains' do
+        it 'properly corrects a method chain on multiple lines' do
+          expect_offense(<<~RUBY)
+            foo(' .x')
+            ^^^^^^^^^^ Redundant line break detected.
+              .bar
+              .baz
+          RUBY
 
-        expect_correction(<<~RUBY)
-          foo.bar.baz
-        RUBY
+          expect_correction(<<~RUBY)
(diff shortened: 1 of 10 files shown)


spec/rubocop/cop/layout/redundant_line_break_spec.rb
```diff
@@ -271,6 +271,19 @@ def resolve_inheritance_from_gems(hash)
             m(7 + 8 + 9)
         RUBY
       end
+
+      it 'properly corrects a method chain on multiple lines' do
```
jonas054, line 275: Since the newly added `gsub` has a regexp that doesn't include a line break, it's a bit more error prone than the already existing ones. So if you change the test case to
```ruby
      it 'properly corrects a method chain on multiple lines' do
        expect_offense(<<~RUBY)
          foo(' .x')
          ^^^^^^^^^^ Redundant line break detected.
            .bar
            .baz
        RUBY

        expect_correction(<<~RUBY)
          foo(' .x').bar.baz
        RUBY
      end
```
it fails because whitespace is removed by autocorrect inside the string.

jonas054 (commented):
Found a potential problem. Looks fine otherwise.

spec/rubocop/cop/layout/redundant_line_break_spec.rb
dvandersluis: Oh good catch! Thank you, I’ll fix it up!

dvandersluis: @jonas054 I fixed that up and added more tests cases. Thank you!

koic:
Thanks!

jonas054:
Nice!

PR #10127, opened 2021-09-27
Replace `Cop` with `Base` in tests to prepare for deprecating `Cop`

dvandersluis:
This should replace be the last official uses of `Cop` in RuboCop itself. Follows #7868.


bbatsov:
One step closer to victory!

PR #10128, opened 2021-09-27
Add spell checking GitHub Actions workflow

koic:
This PR adds spell checking GitHub Actions workflow and fixes typos.

The following two tools are used for spell checking.

- misspell (requires Golang) ... https://github.com/client9/misspell
- codespell (requires Python) ... https://github.com/codespell-project/codespell

For contributors who don't have Golang and Python installed, it is only CONTRIBUTING.md and not run with `rake` task.

This GitHub Actions configuration is based on the Rails's spell checking workflow.
https://github.com/rails/rails/blob/v7.0.0.alpha2/.github/workflows/lint.yml

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/numbered_parameter_assignment.rb
+++ b/lib/rubocop/cop/lint/numbered_parameter_assignment.rb
@@ -10,7 +10,7 @@ module RuboCop
       #   ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin19]
       #   -e:1: warning: `_1' is reserved for numbered parameter; consider another name
       #
-      # Assiging to numbered parameter (from `_1` to `_9`) cause an error in Ruby 3.0.
+      # Assigning to a numbered parameter (from `_1` to `_9`) causes an error in Ruby 3.0.
       #
       #   % ruby -ve '_1 = :value'
       #   ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin19]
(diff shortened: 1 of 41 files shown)


lib/rubocop/cop/lint/numbered_parameter_assignment.rb
```diff
@@ -10,7 +10,7 @@ module Lint
       #   ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin19]
       #   -e:1: warning: `_1' is reserved for numbered parameter; consider another name
       #
-      # Assiging to numbered parameter (from `_1` to `_9`) cause an error in Ruby 3.0.
+      # Assigning to numbered parameter (from `_1` to `_9`) cause an error in Ruby 3.0.
```
dvandersluis, line 13:
```suggestion
      # Assigning to numbered parameter (from `_1` to `_9`) causes an error in Ruby 3.0.
```

😄

dvandersluis (approved):
This is really awesome! Great job @koic

spec/rubocop/cop/style/negated_if_else_condition_spec.rb
```diff
@@ -323,7 +323,7 @@
     RUBY
   end
 
-  it 'does not register an offense when not whe whole condition is negated' do
+  it 'does not register an offense when not the whole condition is negated' do
```
dvandersluis, line 326:
```suggestion
  it 'does not register an offense when the whole condition is negated' do
```

I'm a bit confused about this test statement though because the whole condition is not negated...

koic: I get it. I've updated this with your suggestion. Thank you!

dvandersluis: Maybe `it 'does not register an offense when only part of the condition is negated'` would be clearer.

bbatsov:
Lovely! Every project should have something like this in their CI!

pirj:
We urgently need this for style guides, too :heart:

bbatsov:
True that!

PR #10129, opened 2021-09-27
Update the message for `Style/OptionalBooleanParameter`

dvandersluis:
I noticed on stackoverflow that some people have confusion about how to correct `Style/OptionalBooleanParameter`, so this change makes the messaging more explicit.

### DIFF
--- a/spec/rubocop/cop/style/optional_boolean_parameter_spec.rb
+++ b/spec/rubocop/cop/style/optional_boolean_parameter_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe RuboCop::Cop::Style::OptionalBooleanParameter, :config do
   it 'registers an offense when defining method with optional boolean arg' do
     expect_offense(<<~RUBY)
       def some_method(bar = false)
-                      ^^^^^^^^^^^ Use keyword arguments when defining method with boolean argument.
+                      ^^^^^^^^^^^ Prefer keyword arguments for arguments with a boolean default value; use `bar: false` instead of `bar = false`.
       end
     RUBY
   end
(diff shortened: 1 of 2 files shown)


spec/rubocop/cop/style/optional_boolean_parameter_spec.rb
```diff
@@ -6,24 +6,24 @@
   it 'registers an offense when defining method with optional boolean arg' do
     expect_offense(<<~RUBY)
       def some_method(bar = false)
-                      ^^^^^^^^^^^ Use keyword arguments when defining method with boolean argument.
+                      ^^^^^^^^^^^ Prefer keyword arguments for arguments with a boolean default value; use `bar: false` instead of `bar = false`.
```
koic, line 9: IMHO, long message may not be preferred. 
How about implementing this message as an unsafe autocorrection, using up to a semicolon? So, autocorrection may solve Stack Overflow's issue :-)

```suggestion
                      ^^^^^^^^^^^ Prefer keyword arguments for arguments with a boolean default value.
```

dvandersluis: Yeah I was trying to figure out how to do this in a short way. I think the problem from SO was that the user didn't know what keyword arguments were (beginner) so I thought the example would be helpful.

dvandersluis: @bbatsov what do you think?

bbatsov: Like @koic I don't like long messages, but I'm OK with your suggestion. Two additional points from me:

- Is `Prefer` really better here? I'm mostly wondering as it seems that cop messages are all over the place ("Prefer", "Use", "Do not"). Might be good to analyze them and agree when to use what.
- It's a parameter, not an argument. ;-)

dvandersluis: Yeah. So I thought we prefer (😆) "prefer" because I've seen it a lot but you're right we should evaluate what to use in general. And I agree about arguments vs parameters but I don't know that anyone calls they "keyword parameters" so I wasn't sure what to do there.

bbatsov: Oh, well - not a big deal.

PR #10138, opened 2021-09-29
[Fix #10135] Fix `Style/WordArray` to exclude files in `--auto-gen-config` when `percent` style is given but brackets are required

### LINKED ISSUE #10135: Symbol Arrays with spaces are not excluded by --auto-gen-config
If you run `rubocop --auto-gen-config` against a codebase that contains a symbol array with spaces, eg.:
```ruby
%w[Paris Sydney New\ York]
```
the offense is not added to `rubocop_todo.yml` . Subsequent `rubocop` runs will always have an offense on this array.

## Expected behavior
The file containing the array should be excluded under the `Style/WordArray` cop in `rubocop_todo.yml`

## Actual behavior
The file is not excluded.

This can be reproduced with this spec:

```diff
diff --git a/spec/rubocop/cli/auto_gen_config_spec.rb b/spec/rubocop/cli/auto_gen_config_spec.rb
index 0691669b4..240a0bfa2 100644
--- a/spec/rubocop/cli/auto_gen_config_spec.rb
+++ b/spec/rubocop/cli/auto_gen_config_spec.rb
@@ -300,6 +300,31 @@ RSpec.describe 'RuboCop::CLI --auto-gen-config', :isolated_environment do # rubo
       end
     end
 
+    it 'excludes the complex array when the style is mixed for Style/WordArray' do
+      create_file(
+        'example1.rb',
+        [
+          '# frozen_string_literal: true',
+          '',
+          '%w[France Australia USA] +',
+          '  %w[Paris Sydney New\ York]'
+        ]
+      )
+
+      expect(cli.run(['--auto-gen-config'])).to eq(0)
+      expect(File.readlines('.rubocop_todo.yml')[8..-1].join)
+        .to eq(<<~YAML)
+          # Offense count: 1
+          # Cop supports --auto-correct.
+          # Configuration parameters: MinSize, WordRegex.
+          # SupportedStyles: percent, brackets
+          Style/WordArray:
+            EnforcedStyle: percent
+            Exclude:
+              - example1.rb
+        YAML
+    end
+
     it 'overwrites an existing todo file' do
       create_file('example1.rb', ['# frozen_string_literal: true',
                                   '',
```
The spec should pass, but instead it fails with the diff showing that the following is missing:
```diff
-            Exclude:
-              - example1.rb
```

## RuboCop version

Issue was encountered in app with the following version:
```
➜  finbot (dependabot/bundler/rubocop-1.21.0) ✗ be rubocop -V
1.21.0 (using Parser 3.0.2.0, rubocop-ast 1.12.0, running on ruby 2.5.7 x86_64-darwin19)
```

Spec was run against rubocop/master. HEAD at the time was 8b5d234.

dvandersluis:
When `Style/WordArray` is in `percent` style, offenses are registered for percent arrays that contain "invalid" characters, in order to replace them with a bracketed array instead. However, when this is the case, `--auto-gen-config` doesn't handle the file properly. This updates it to exclude the file since the result is inconsistent.

Fixes #10135.

-----------------

* [x] Added tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -144,3 +144,7 @@ InternalAffairs/UndefinedConfig:
   Exclude:
     - 'lib/rubocop/cop/correctors/**/*.rb'
     - 'lib/rubocop/cop/mixin/**/*.rb'
+
+InternalAffairs/StyleDetectedApiUse:
+  Exclude:
+    - 'lib/rubocop/cop/mixin/percent_array.rb'
--- a/changelog/fix_fix_stylewordarray_to_exclude_files_in.md
+++ b/changelog/fix_fix_stylewordarray_to_exclude_files_in.md
@@ -1,0 +1,1 @@
+* [#10135](https://github.com/rubocop/rubocop/issues/10135): Fix `Style/WordArray` to exclude files in `--auto-gen-config` when `percent` style is given but brackets are required. ([@dvandersluis][])
--- a/lib/rubocop/cop/mixin/percent_array.rb
+++ b/lib/rubocop/cop/mixin/percent_array.rb
@@ -36,7 +36,12 @@ module RuboCop
       def check_percent_array(node)
         array_style_detected(:percent, node.values.size)
 
-        return unless style == :brackets || invalid_percent_array_contents?(node)
+        brackets_required = invalid_percent_array_contents?(node)
+        return unless style == :brackets || brackets_required
+
+        # If in percent style but brackets are required due to
+        # string content, the file should be excluded in auto-gen-config
+        no_acceptable_style! if brackets_required
 
         bracketed_array = build_bracketed_array(node)
         message = format(self.class::ARRAY_MSG, prefer: bracketed_array)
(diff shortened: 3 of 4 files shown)


.rubocop.yml
```diff
@@ -144,3 +144,7 @@ InternalAffairs/UndefinedConfig:
   Exclude:
     - 'lib/rubocop/cop/correctors/**/*.rb'
     - 'lib/rubocop/cop/mixin/**/*.rb'
+
+InternalAffairs/StyleDetectedApiUse:
+  Exclude:
+    - 'lib/rubocop/cop/mixin/percent_array.rb'
```
dvandersluis, line 150: `InternalAffairs/StyleDetectedApiUse` is registering a false positive here because it's detecting `no_acceptable_style` as a negative but the positive is within `array_style_detected`.

Interestingly, there does not seem to be any way to inline disable global offenses...

bbatsov:
Thanks!

PR #10006, opened 2021-08-12
Address `frozen_string_literal` no longer applying to interpolated strings in Ruby 3.0 or higher

splattael:
Note, this PR is strongly inspired by https://github.com/rubocop/rubocop/pull/8743. Kudos to @sambostock :heart: 

----

Since [Ruby 3.0 or higher](https://bugs.ruby-lang.org/issues/17104) interpolated strings are not frozen automatically anymore.

Therefore, two changes must happen to how we treat interpolated literals

* `Style/MutableConstant` should enforce freezing them
* `Style/RedundantFreeze` should not consider it redundant to freeze them

The behaviour for Ruby 2.7 or lower is kept.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [-] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #10139, opened 2021-09-29
[Fix #10134] Update `Style/MutableConstant` to not consider multiline uninterpolated strings as unfrozen in ruby 3.0

### LINKED ISSUE #10134: Style/MutableConstant false positive with frozen_string_literal: true and line continuation
When `# frozen_string_literal: true` is set, a constant string that includes a line continuation (`\`) is incorrectly flagged by the `Style/MutableConstant` cop, even though the constant is indeed frozen.

--------

## Expected behavior

RuboCop should not flag this as a `Style/MutableConstant` violation.

## Actual behavior

RuboCop wants a `.freeze` at the end of the string literal to appease the `Style/MutableConstant` cop.

```ruby
# frozen_string_literal: true

C = 'this is a string '\
    'wrapped onto another line'
```

## RuboCop version

```
$ bundle exec rubocop -V
1.21.0 (using Parser 3.0.2.0, rubocop-ast 1.11.0, running on ruby 3.0.2 x86_64-darwin20)
  - rubocop-performance 1.11.5
  - rubocop-rails 2.12.2
  - rubocop-rspec 2.4.0
```

dvandersluis:
In #10006 `Style/MutableConstant` was updated to consider `dstr` nodes as unfrozen in Ruby 3.0 even when `frozen_string_literal: true` is specified, and requires `.freeze` on those strings. However, a multiline string is considered a `dstr` node but does not contain interpolation so would still be frozen.

```ruby
# frozen_string_literal: true

CONST = 'foo' \
        'bar'

CONST.frozen? #=> true
```

Fixes #10134.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_update_stylemutableconstant_to_not.md
+++ b/changelog/fix_update_stylemutableconstant_to_not.md
@@ -1,0 +1,1 @@
+* [#10134](https://github.com/rubocop/rubocop/issues/10134): Update `Style/MutableConstant` to not consider multiline uninterpolated strings as unfrozen in ruby 3.0. ([@dvandersluis][])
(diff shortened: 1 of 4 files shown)


koic:
Nice 👍

PR #10152, opened 2021-10-01
[Fix #10148] Fix `Style/QuotedSymbols` handling escaped characters incorrectly

### LINKED ISSUE #10148: `Style/QuotedSymbols` endless loop caused by escaped quote character
On upgrading rubocop from version 1.1.0 to 1.22.0 there was a problem running the auto-correct command. The issue causes a file to be continuously filled with '/'s.
--------

## Expected behavior

Normally, rubocop would replace the double ticks with single ticks in this statement:

`:"Le produit avec le SKU \"%1\" n’existe pas" `

Would be replaced with
`:'Le produit avec le SKU \"%1\" n’existe pas'`

## Actual behavior
 Rubocop gets caught in an endless loop continuously generating '/'s like below:
 
 `:"Le produit avec le SKU \\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\"%1\" n’existe pas" `

The file with this line is never corrected and CLI does not progress.

Create a ruby file with the line above.

Run rubocop -a

The command line will hang and when forced to exit the file should contain many '/'s.

## RuboCop version

1.22.0
no output possible as the code hangs on an endless loop.

dvandersluis:
Two cases are improved here:
1. Escaped double quotes (`\"`) within a double-quoted string are handled correctly (replaced with an unescaped double quote) when converting to a single-quoted string.
2. Escaped backslashes (`\\`) are treated properly and not doubled for both single- and double-quoted strings.

Fixes #10148.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_fix_stylequotedsymbols_handling_escaped.md
+++ b/changelog/fix_fix_stylequotedsymbols_handling_escaped.md
@@ -1,0 +1,1 @@
+* [#10148](https://github.com/rubocop/rubocop/issues/10148): Fix `Style/QuotedSymbols` handling escaped characters incorrectly. ([@dvandersluis][])
--- a/lib/rubocop/cop/mixin/string_literals_help.rb
+++ b/lib/rubocop/cop/mixin/string_literals_help.rb
@@ -13,7 +13,11 @@ module RuboCop
         if style == :single_quotes
           !double_quotes_required?(src)
         else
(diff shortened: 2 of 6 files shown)


bbatsov:
Thanks!

PR #10155, opened 2021-10-01
[Fix #10145] Update `Style/SelectByRegexp` to ignore cases where the receiver appears to be a hash

### LINKED ISSUE #10145: Style/SelectByRegexp flags/autocorrects hash.select incorrectly
## Expected behavior

I expect this cop to only apply to Array `select` and not to Hash `select`.

## Actual behavior

Hash `select` is flagged if the arity is 1, even though the suggested autocorrect is wrong. i.e. `hash.select {|k| /foo/.match? k }` can't be replaced with `hash.grep(/foo/)`, since single-arg `grep` always returns `[]` with hashes.

```
$ echo '{ a: 1, aa: 2 }.select { |k| /aa/.match? k }' | rubocop -sx -A
Inspecting 1 file
C

Offenses:

x:1:1: C: [Corrected] Style/SelectByRegexp: Prefer grep to select with a regexp match.
{ a: 1, aa: 2 }.select { |k| /aa/.match? k }
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

====================
{ a: 1, aa: 2 }.grep(/aa/)
```

## RuboCop version

1.22.0

dvandersluis:
Since `hash.grep` does not work in the same way as other enumerables, this update looks for receivers that look like hashes and does not register `Style/SelectByRegexp` offenses on them. This includes literal hashes (`{}`), `Hash.new`, `Hash[]`, `to_h` and `to_hash` (the latter two could potentially cause a false negative if they're overridden to not return hashes, but I can live with that).

I added explicit tests for receivers that I have confirmed works as expected (array, range, set). In case we ever extend this with more exceptions those will be covered.

Fixes #10145.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_update_styleselectbyregexp_to_ignore.md
+++ b/changelog/fix_update_styleselectbyregexp_to_ignore.md
@@ -1,0 +1,1 @@
+* [#10145](https://github.com/rubocop/rubocop/issues/10145): Update `Style/SelectByRegexp` to ignore cases where the receiver appears to be a hash. ([@dvandersluis][])
--- a/lib/rubocop/cop/style/select_by_regexp.rb
+++ b/lib/rubocop/cop/style/select_by_regexp.rb
@@ -3,9 +3,15 @@
 module RuboCop
   module Cop
     module Style
-      # This cop looks for places where an subset of an array
-      # is calculated based on a `Regexp` match, and suggests `grep` or
-      # `grep_v` instead.
+      # This cop looks for places where an subset of an Enumerable (array,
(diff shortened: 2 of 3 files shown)


bbatsov:
Looks good!

ghiculescu:
@dvandersluis could you also exclude `ENV` here? I think is guaranteed to be a Hash

koic:
@ghiculescu That makes sense. I've opened https://github.com/rubocop/rubocop/pull/10457. Thank you for the feedback.

PR #10157, opened 2021-10-01
Updated `Gemspec/RequiredRubyVersion` to detect more issues

dvandersluis:
Adds detection for empty versions and `required_ruby_version` being set more than once. Also did a minor refactor to use `on_send` instead of having all the logic in `on_new_investigation`.

-----------------

* [x] Added tests.


lib/rubocop/cop/gemspec/required_ruby_version.rb
```diff
@@ -49,14 +57,15 @@ module Gemspec
...
                         '(%<target_ruby_version>s, which may be specified in ' \
                         '.rubocop.yml) should be equal.'
         MISSING_MSG = '`required_ruby_version` should be specified.'
+        EXTRA_MSG = '`required_ruby_version` is set more than once.'
 
-        # @!method required_ruby_version(node)
-        def_node_search :required_ruby_version, <<~PATTERN
+        # @!method required_ruby_version?(node)
+        def_node_search :required_ruby_version?, <<~PATTERN
           (send _ :required_ruby_version= $_)
```
koic, line 69: This capture looks unnecessary.

```suggestion
          (send _ :required_ruby_version= _)
```

dvandersluis: Good call, it was left over from the previous version but I've removed it now 👍

lib/rubocop/cop/gemspec/required_ruby_version.rb
```diff
@@ -26,6 +28,12 @@ module Gemspec
       #     spec.required_ruby_version = '>= 2.6.0'
       #   end
       #
+      #   # bad (`required_ruby_version` set twice)
+      #   Gem::Specification.new do |spec|
+      #     spec.required_ruby_version = '>= 2.5.0'
+      #     spec.required_ruby_version = '>= 2.5.0'
+      #   end
```
koic, line 35: I think about it. It may be better to create a new cop like `Gemspec/DuplicateSpecificationAttribute` cop instead of this cop role extension. It seems preferable to keep one cop role simple based on Single Responsibility Principle :-) What do you think?

dvandersluis: I think it makes sense all together because the single responsibility here is checking `required_ruby_version`, and it basically was already doing this, but just for 0 and 1 calls, so now extended to 2+ as well. 

However, if you feel strongly about this I would be fine splitting it (and I’d move the check for 0 calls to that new cop too).

koic: IMHO, this cop is responsible for checking whether the values for `spec.required_ruby_version` of gemspec and `TargetRubyVersion` of .rubocop.yml match.

it seems simpler for another new cop to have the role of checking for duplicate attributes such as `required_ruby_version` for `Gem::Specification`. The new cop can check that various attributes such as `spec.name`, `spec.version`, `spec.description`, and etc are redundantly assigned.

bbatsov: I agree with Koichi. Seems a separate cop looking for duplicated/empty properties would be a better idea.

dvandersluis: So there actually is already another cop that checks for duplicate assignments: `Gemspec/DuplicatedAssignment` 🤦

dvandersluis:
I've updated this to just keep the added version checks (for `''` and `[]`, etc.) and removed the other logic. I've kept the refactoring to using `on_send` because I think that's preferable to doing everything in `on_new_investigation`

PR #10160, opened 2021-10-03
[Fix #7900] Fix `Style/FormatStringToken` false positive with formatted input and `template` style enforced, and add autocorrection

### LINKED ISSUE #7900: Style/FormatStringToken false positive with formatted input and template style enforced
## Expected behavior

When enforcing the template style: 
```
Style/FormatStringToken:
  EnforcedStyle: template
```

and with the following code:
```
format('%<number>0.02f', value) }
```

Formatting input doesn't work with template style - [source](https://ruby-doc.org/core-2.7.0/Kernel.html#method-i-format)

> %<name>s style uses format style, but %{name} style doesn’t.

The cop should not report offenses or accept an new configuration option for formatted input.

I guess this is a similar issue reported in [this comment](https://github.com/rubocop-hq/rubocop/issues/5630#issuecomment-546450054)
 

## Actual behavior

It returns the offense :

```
Style/FormatStringToken: Prefer template tokens (like %{foo}) over annotated tokens (like %<foo>s).
  format('%<number>0.02f', value)
```

See examples above

## RuboCop version

```
$ [bundle exec] rubocop -V
0.79.0
```

FnControlOption:
Fixes #7900.

Types of format strings:
- `annotated`: `'%<name>s'` (where `s` is the format type)
- `template`: `'%{name}'`
- `unannotated`: `'%s'`

The format string `'%{name}'` is only equivalent to `'%<name>s'`, so when enforcing the `template` style, no offenses should be reported if the format type is not `s` (for example, `'%<name>f'`).

The autocorrection from `template` style `%[flags][width][.precision]{name}` to `annotated` style is `%<name>[flags][width][.precision]s`, and vice versa.

Because `unannotated` has no names, it can't be autocorrected to `annotated` or `template`.

Autocorrecting from either `annotated` or `template` to `unannotated` is possible, but I haven't attempted adding this yet. The method arguments must be updated too, so `def on_send` would need to be added (maybe as a new cop). For example:

```rb
format('%<greeting>s', greeting: 'Hello')
format('%{foo} %{bar} %{foo}', foo: 'foo', bar: 'bar')
# becomes
format('%s', 'Hello')
format('%1$s %2$s %1$s', 'foo', 'bar')
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/format_string_token.rb
+++ b/lib/rubocop/cop/style/format_string_token.rb
@@ -69,6 +69,7 @@ module RuboCop
       class FormatStringToken < Base
         include ConfigurableEnforcedStyle
         include IgnoredMethods
+        extend AutoCorrector
 
         def on_str(node)
           return if format_string_token?(node) || use_ignored_method?(node)
@@ -77,13 +78,8 @@ module RuboCop
           return if detections.empty?
           return if allowed_unannotated?(detections)
 
-          detections.each do |detected_style, token_range|
-            if detected_style == style
-              correct_style_detected
-            else
-              style_detected(detected_style)
-              add_offense(token_range, message: message(detected_style))
-            end
+          detections.each do |detected_sequence, token_range|
+            check_sequence(detected_sequence, token_range)
           end
         end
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/format_string_token.rb
```diff
@@ -105,6 +101,38 @@ def use_ignored_method?(node)
...
+        def autocorrect_sequence(corrector, detected_sequence, token_range)
+          return if style == :unannotated
+
+          name = detected_sequence.name
+          return if name.nil?
+
+          flags = detected_sequence.flags
+          width = detected_sequence.width
+          precision = detected_sequence.precision
+          type = detected_sequence.style == :template ? 's' : detected_sequence.type
```
FnControlOption, line 128: I realized that these lines can be simplified to:

```rb
return if style == :unannotated || detected_sequence.style == :unannotated

name = detected_sequence.name
flags = detected_sequence.flags
width = detected_sequence.width
precision = detected_sequence.precision
type = case detected_sequence.style
       when :annotated then detected_sequence.type
       when :template then 's'
       end
```

marcandre (approved):
Looks good to me 👍

PR #10163, opened 2021-10-04
[Fix #8101] Reformat `rake spec` output to amplify signal and reduce noise

### LINKED ISSUE #8101: `rake spec` needs love
(at least when running in parallel:)

1) Rake spec lists all the pending specs. I, for one, never care when running rake spec. By default I wish it didn't. 

2) It does not give the usual summary with path to reproduce, e.g. `rspec spec/rest/of/path.rb:42`

3) The parallel summary is really not helpful. It lists tested / errors / pending, for each processor. Not only is this not helpful, but the main column is right aligned with "x pending" sometimes appearing which screws up everything. We should only have a Total line anyways.

dvandersluis:
I was working on this and then realized that there was already an issue for it, so this fixes #8101.

`rake spec` runs tests via `test-queue` by default in order to parallelize the work. When there are no failing tests, this is fine, because it just outputs one line per worker with a summary. However, when there are failures, a ton of noise (raw output from rspec, pending tests, deprecation warnings, etc.) is spewed out and it makes it really hard to find what the actual failures are in.

This change removes all the noise and makes it clearer what went wrong:
* Pending spec output is hidden
* Deprecation warnings are fixed (they were coming from the `test-queue` gem which is out of date)
* The summary line was reordered so that failures always show up on the right edge allowing for easier scanning
* Output is now colorized (red for failure, green for good, pending status does not change coloring)
* Added a new section at the bottom that contains only the "rerun" commands to be able to quickly rerun each failure, across all parallelization workers

This was not easy to accomplish because there is a lot of stuff we don't need or want baked into `TestQueue::Runner` in a not very friendly for change way. Unfortunately, the `test-queue` gem has stagnated overall and a new version has not been released since January 2017. 

Thus, in order to accomplish this, it is a mix between overriding methods from `test-queue` and adding a custom RSpec formatter. It's not the cleanest solution, but it works. I tried to keep things as clean as possible.

Successful run:
image

Failed run:
image

You'll notice the failure numbering doesn't really work right. This is because it's just concatenating together multiple workers' output, and each worker has its own counting. I didn't try to resolve this.


dvandersluis:
Decided that I might as well make the numbering work too, so now failures are numbered properly in the final output. :grin:

bbatsov:
Fantastic work!

PR #10164, opened 2021-10-04
Removed stray `puts` from changelog.rb

dvandersluis:
I noticed that the changelog specs were outputting text by accident, there was a leftover debugging puts.


koic:
Good catch!

PR #10166, opened 2021-10-05
Fix a false positive for `Style/StringLiterals`

koic:
Follow up to https://github.com/rubocop/rubocop/commit/a7ef72aa8c35336b169c5d78e2925fc27d739fcb#r57503462.

This PR fixes a false positive for `Style/StringLiterals` when using `'\s'` with `EnforcedStyle: double_quotes`

-----------------

* [x] Added tests.


Bo98:
Makes sense.

I take the intention here is to limit this to recognised control character escape sequences and quote escaping?

Because it's worth nothing even things like `'\z'`, for example, is not the same as `"\z"` (which equates to `"z"`) - you still need `"\\z"`.

nvasilevski:
Hey folks, I believe there are more examples of false-positives apart from the `'\s'` like `'\S'` or `'\d'` or `'\p'`
The autocorrector turns these into `"\\S"` and `"\\d"` and satisfies rubocop by keeping the behaviour, however, from my point of view, in large expressions it makes the code harder to read just to follow the rule which seems a bit unreasonable 
Should we consider ignoring such cases in general and avoid overcomplicating expressions from readability point of view?
Please feel free to let me know if I'm missing something, thanks!

mvz:
With this change, `'No mutations found \o/ \o/'` is now corrected to `"No mutations found \\o/ \\o/"`. I'm not sure if that was the intention @koic?

jdelStrother:
Yeah, I'm seeing a lot of similar problems across our codebase. I'm not entirely convinced about the rationale behind this change - we use double quoted strings everywhere, unless the string needs literal backslashes in it (where we use single-quotes to avoid the double-backslash).

> To clarify my point here: this applies to any character. Or at least I can't find any that behave differently. Am I missing something?

Correct - in all your examples, none of the double-quoted strings have a backslash in their 'output', and all the single-quoted ones do.

PR #10177, opened 2021-10-08
Add documentation about `rubocop-daemon`

### LINKED ISSUE #10172: RuboCop is slow - why not mention rubocop-daemon?
I'm always hesitant to use RuboCop as it's too slow and it adds up if it's run too often.
Just running `time rubocop` in one project (~200 files), takes 6 seconds. (after multiple runs)
In another project, with ~400 files, it takes 10 seconds.

It would be great if RuboCop could be run on file save and it would run fast.
I'd love to be able to use RuboCop the same way I use Prettier for my JS/TS files.

Today I discovered an incredible tool called `rubocop-daemon`.
I managed to set it up and it runs RuboCop so fast, it's crazy. :)

I saw this issue where `rubocop-daemon` was mentioned before (3.5 years ago!): [rubocop -a is slow](https://github.com/rubocop/rubocop/issues/6492)
The conclusion of the discussion there was that, in @jonas054's words:
> It sounds like `rubocop-daemon` is the way forward.

I used RuboCop in the meantime and even [contributed a bit](https://github.com/rubocop/rubocop/pulls?q=author%3Avfonic), but I've never heard of `rubocop-daemon` which is a life saver when it comes to "format on save" functionality.

Do you think we should add `rubocop-daemon` somewhere in the documentation and if so, where do you suggest would be the best place to mention it?

/cc @gurgeous @ndbroadbent

dvandersluis:
I added documentation to the Integrations documenation page. Please let me know if I should add a note elsewhere as well.

Closes #10172.

### DIFF
--- a/docs/modules/ROOT/pages/integration_with_other_tools.adoc
+++ b/docs/modules/ROOT/pages/integration_with_other_tools.adoc
@@ -1,5 +1,22 @@
 = Integration with Other Tools
 
+== Speeding up integrations
+
+RuboCop integrates with quite a number of other tools, including editors which may attempt
+to do auto-correction for you. In these cases, `rubocop` ends up getting called repeatedly,
+which may result in some slowness, as `rubocop` has to require its entire environment on
+each call.
+
+You can alleviate some of that boot time by using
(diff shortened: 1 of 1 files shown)


koic:
Nice 👍

PR #10178, opened 2021-10-09
Add rubocop-graphql to suggested extensions

DmitryTsepelev:
This PR adds [rubocop-graphql](https://github.com/DmitryTsepelev/rubocop-graphql) to the `SuggestExtensions` list.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -150,6 +150,7 @@ AllCops:
     rubocop-minitest: [minitest]
     rubocop-sequel: [sequel]
     rubocop-rake: [rake]
+    rubocop-graphql: [graphql]
 
 #################### Bundler ###############################
 


bbatsov:
Thanks!

codebender:
@DmitryTsepelev  It looks like this suggests a rubocop org gem that doesn't exist instead of your 3rd party gem:

```
Tip: Based on detected gems, the following RuboCop extension libraries might be helpful:
  * rubocop-graphql (https://github.com/rubocop/rubocop-graphql)
```

`https://github.com/rubocop/rubocop-graphql` doesn't exit

DmitryTsepelev:
Oh, I missed that it does `https://github.com/rubocop/#{extension}`. Looks like we have options:

**When `extension` contains `"http"` — do not interpolate it (so we can add custom extensions which are not in rubocop org)**

```
extensions.sort.each do |extension|
  name, link = if extension.include?("http")
    [extension.split("/").last, extension]
  else
    [extension, "https://github.com/rubocop/#{extension}"]
  end

  puts "  * #{name} (#{link})"
end
```

**Check if `extension` is a hash and allow configuring name and link**

```
SuggestExtensions:
  rubocop-graphql:
    graphql:
      name: graphql
      link: https://github.com/DmitryTsepelev/rubocop-graphql
```

**Transfer the gem to the rubocop organization**

@bbatsov do you have a preference how to do it better?

dvandersluis:
I like the latter suggestion but I think your proposed shape is a bit overly complicated. The array it currently accepts is all the gems that will trigger the suggestion for the given extension.

How about:

```yaml
SuggestExtensions:
  rubocop-graphql:
    gems: [graphql]
    url: https://github.com/DmitryTsepelev/rubocop-graphql
```

And maintain allowing just an array to be passed rather than requiring a hash (we need this for backwards compatibility anyways).

koic:
A simple solution would be to suggest rubygems.org URL instead of github.com URL. So, RubyGems URL is better than GitHub repository URL because it is likely to match `require` extension gem name.
https://guides.rubygems.org/name-your-gem/

I've opened PR #10275.

PR #10205, opened 2021-10-21
[Fix #10202] Add new `Lint/Ruby2Keywords` cop

### LINKED ISSUE #10202: Cop idea: useless `ruby2_keywords`
If method doesn't have kwargs, `ruby2_keywords` is useless, but sometime is used due to overlook. Rubocop should warn about
```
def xxx(foo)
end
ruby2_keywords :xxx
```
and alike to reduce useless code amount.

dvandersluis:
Looks for instances of `ruby2_keywords` for methods that do not need it. 

`ruby2_keywords` is only needed for methods that have a splat (`*args`) and no keyword args (regular, optional or keyword splat).

Can handle three forms:

```ruby
ruby2_keywords def foo(arg)
end
```

```ruby
def foo(arg)
end
ruby2_keywords :foo
```

```ruby
define_method(:foo) { |arg| }
ruby2_keywords :foo
```

Fixes #10202.

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/ruby2_keywords.rb
```diff
@@ -0,0 +1,111 @@
...
+module RuboCop
+  module Cop
+    module Lint
+      # This cop looks for `ruby2_keywords` calls for methods that do not need it.
+      #
+      # `ruby2_keywords` should only be called on methods that accept an argument splat
+      # (`*args`) but do not explicit keyword arguments (`k:` or `k: true`) or
+      # a keyword splat (`**kwargs`).
+      #
+      # @example `ruby2_keywords` given a `def`
```
koic, line 12: `@example` directive is used to represent `EnforcedStyle` (and similar style option). Can you use source comment for explanation to examples? :-)

lib/rubocop/cop/lint/ruby2_keywords.rb
```diff
@@ -0,0 +1,111 @@
...
+      #
+      #   # bad
+      #   def foo
+      #   end
+      #   ruby2_keywords :foo
+      #
+      #   # good
+      #   def foo(*args)
+      #   end
+      #   ruby2_keywords :foo
```
koic, line 47: A good case compatible with a bad case would be to remove `ruby2_keywords`.

```suggestion
      #   def foo
      #   end
```

lib/rubocop/cop/lint/ruby2_keywords.rb
```diff
@@ -0,0 +1,111 @@
...
+      # @example `ruby2_keywords` with a dynamically defined method
+      #   # The same argument formats as above apply
+      #
+      #   # bad
+      #   define_method(:foo) { |arg| }
+      #   ruby2_keywords :foo
+      #
+      #   # good
+      #   define_method(:foo) { |*args| }
+      #   ruby2_keywords :foo
```
koic, line 58:
```suggestion
      #   define_method(:foo) { |arg| }
```

lib/rubocop/cop/lint/ruby2_keywords.rb
koic: And I think it will be easier to understand whether this case is represented at the top of example.

koic:
I think the cop name `UselessRuby2Keywords` cop is better.

dvandersluis:
@koic I renamed the cop and changed the examples. I still think it's not a bad thing to have separate @example blocks for different use cases (I understand that we usually do this for configuration examples, but still), but I rewrote it to be more like you were suggesting.

PR #10223, opened 2021-10-29
[#10221] Update `Naming::FileName` to recognize `Struct`s as classes that satisfy the `ExpectMatchingDefinition` requirement

### LINKED ISSUE #10221: Naming/FileName: Detect constant assignment

Currently, `Naming/FileName` always flag the following code:

`lib/something/bar.rb`

```ruby
module Something
  Bar = Struct.new(:bar)
end
```

which produces

```
Offenses:

lib/something/bar.rb:1:1: C: Naming/FileName: bar.rb should define a class or module called Something::Bar.
module Something
^
```

It'd be great to support such constant assignments.

In this example, one could use `class Bar < Struct.new(:bar)` but those workarounds are not always feasible.

## Additional context

See https://gitlab.com/gitlab-org/gitlab/-/merge_requests/73252#note_717026413

dvandersluis:
Allows `Struct`s assigned to constants to be used for `ExpectMatchingDefinition` of `Naming::FileName`.

Allow refactors the cop's tests to use `expect_offense` and adds some documentation for options that aren't explained.

Fixes #10221.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_update_namingfilename_to_recognize.md
+++ b/changelog/change_update_namingfilename_to_recognize.md
@@ -1,0 +1,1 @@
+* [#10221](https://github.com/rubocop/rubocop/issues/10221): Update `Naming::FileName` to recognize `Struct`s as classes that satisfy the `ExpectMatchingDefinition` requirement. ([@dvandersluis][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2501,6 +2501,7 @@ Naming/FileName:
   StyleGuide: '#snake-case-files'
   Enabled: true
   VersionAdded: '0.50'
+  VersionChanged: '<<next>>'
   # Camel case file names listed in `AllCops:Include` and all file names listed
   # in `AllCops:Exclude` are excluded by default. Add extra excludes here.
   Exclude: []
--- a/lib/rubocop/cop/naming/file_name.rb
+++ b/lib/rubocop/cop/naming/file_name.rb
@@ -14,6 +14,18 @@ module RuboCop
       # (i.e. `bundler-console` becomes `Bundler::Console`). As such, the
       # gemspec is supposed to be named `bundler-console.gemspec`.
       #
+      # When `ExpectMatchingDefinition` (default: `false`) is `true`, the cop requires
+      # each file to have a class, module or `Struct` defined in it that matches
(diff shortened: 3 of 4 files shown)


splattael:
@dvandersluis Thanks for opening this PR so quickly :heart: 

I think this is a good start but I was wondering if we need to take care of non-`Struct` related assignments as well to support e.g. aliasing:

```ruby
module Something
  Bar = ::OtherNamespace::Bar
end
```

At GitLab we use such technique for aliasing in particular, for example: https://gitlab.com/gitlab-org/gitlab/-/blob/4e75a27af3add26b45c4dd05314946757a670578/ee/lib/elastic/v12p1/routing.rb#L6

dvandersluis:
I'm not sold on having a constant defined satisfying the FileName cop. For Struct it's one thing because it's actually defining a class, but a constant could be anything.

Would you think this should satisfy the `ExpectMatchingDefinition` requirement?

```ruby
module Foo
  Bar = 5
end
```

For your specific use case, I wonder if just excluding `gitlab/ee/lib/elastic/v12p1/**/*` from this cop would be a better solution.

splattael:
> I'm not sold on having a constant defined satisfying the FileName cop. For Struct it's one thing because it's actually defining a class, but a constant could be anything.
> 
> Would you think this should satisfy the `ExpectMatchingDefinition` requirement?
> 
> ```ruby
> module Foo
>   Bar = 5
> end
> ```

Yes, I think this cop should only check whether the defined constant matches the file name. The value of the constant can be anything - which is [`zeitwerk`'s](https://github.com/fxn/zeitwerk#the-idea-file-paths-match-constant-paths) approach as well.

Note that Ruby has many ways of creating a class/module so it don't have to be `Struct` specific at all:

```ruby
module Foo
  Bar = SomeClassFactory.create
end 
```

See for example, https://github.com/dry-rb/dry-types/blob/9562995e50f9dfba0b8628224b07d865d18335da/lib/dry/types/any.rb#L45.

> For your specific use case, I wonder if just excluding `gitlab/ee/lib/elastic/v12p1/**/*` from this cop would be a better solution.

We did exclude them for now but ideally we'd also support aliasing hence the issue :sweat_smile:

PR #10235, opened 2021-11-04
[Fix #10203] Fix `Style/FormatStringToken` to respect `IgnoredMethods` with nested structures

### LINKED ISSUE #10203: `Style/FormatStringToken` does not respect `IgnoredMethods` for values in nested structures
The `Style/FormatStringToken` cop does not respect `IgnoredMethods` for values which are nested, e.g. into an options hash.

### Suggested solution

I've investigated why that is, and the culprit seems to be `FormatStringToken#use_ignored_method?` where we we only check if the found node (the offending string) has a method parent (`send` type). In case of a value inside a hash, the parent is not the method itself.

Maybe it makes sense to allow climbing up the node tree as long as parents are `pair` or `hash` nodes (maybe others?), and only then check if the surrounding node is of type `send`.

Please let me know if that approach sounds reasonable and I can try implementing a fix.

--------

## Expected behavior

Given the following config,
```yaml
Style/FormatStringToken:
  EnforcedStyle: annotated
  IgnoredMethods:
    - example
```

and the following Ruby code,
```ruby
example("Hello %{name}")
example("Hello %{name}", bye: "Bye %{name}")
```

both calls to `example` should be ignored.

## Actual behavior

`example("Hello %{name}")` is ignored, but `example("Hello %{name}", bye: "Bye %{name}")` raises an offense.

```
Offenses:

test.rb:2:36: C: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over template tokens (like %{foo}).
example("Hello %{name}", bye: "Bye %{name}")
                                   ^^^^^^^
```

Run Rubocop with example config from above against a file with example Ruby code from above.

## RuboCop version

```
$ bundle exec rubocop -V
1.22.1 (using Parser 3.0.2.0, rubocop-ast 1.12.0, running on ruby 2.7.3 x86_64-linux)
  - rubocop-performance 1.11.5
  - rubocop-rails 2.12.4
  - rubocop-rspec 2.5.0
```

tejasbubane:
Closes #10203

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/format_string_token.rb
+++ b/lib/rubocop/cop/style/format_string_token.rb
@@ -102,7 +102,8 @@ module RuboCop
         end
 
         def use_ignored_method?(node)
-          (parent = node.parent) && parent.send_type? && ignored_method?(parent.method_name)
+          send_parent = node.each_ancestor(:send).first
+          send_parent && ignored_method?(send_parent.method_name)
         end
 
         def unannotated_format?(node, detected_style)
--- a/changelog/fix_format_string_token_nested_ignored_methods.md
+++ b/changelog/fix_format_string_token_nested_ignored_methods.md
@@ -1,0 +1,1 @@
+* [#10203](https://github.com/rubocop/rubocop/issues/10203): Fix `Style/FormatStringToken` to respect `IgnoredMethods` with nested structures. ([@tejasbubane][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/format_string_token.rb
```diff
@@ -102,7 +102,8 @@ def format_string_token?(node)
         end
 
         def use_ignored_method?(node)
-          (parent = node.parent) && parent.send_type? && ignored_method?(parent.method_name)
+          send_parent = node.ancestors.find(&:send_type?)
```
dvandersluis, line 105:
```suggestion
          send_parent = node.each_ancestor(:send).first
```

tejasbubane: @dvandersluis Thanks. Changed.

dvandersluis (approved):

dvandersluis:
Thanks @tejasbubane!

PR #10239, opened 2021-11-11
Require MFA for gem pushes

dvandersluis:
Ruby gems now allows MFA to be required for pushes: https://guides.rubygems.org/mfa-requirement-opt-in/
This allows users to have confidence that the actual authors were responsible for updates (obviously that doesn't mean that it can't be a malicious update, but at least it's not from someone who got access to a pusher's account).

I don't have push access but I hope all of us who do have MFA set up anyways (and if anyone doesn't it's easy to set up).

If this is acceptable, I'd like to create a cop to check for this as well.


bbatsov:
Yes and yes! Thanks!

PR #10242, opened 2021-11-12
[Fix #10241] Fix `last_column` value for `JSONFormatter`

koic:
Fixes #10241.

This PR fixes `last_column` value for `JSONFormatter`.

The minimum value of `start_column: real_column` is 1. So, the minimum value of `last_column` should be 1.
And non-zero value of `last_column` should be used as is.

This patch to `last_column` seems to be required only for `JSONFormatter` and not required for other formatters.

It solves a potential `JSONFormatter` issue for cops that use `add_global_offense`.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/formatter/json_formatter.rb
+++ b/lib/rubocop/formatter/json_formatter.rb
@@ -59,12 +59,15 @@ module RuboCop
       end
 
       # TODO: Consider better solution for Offense#real_column.
+      #       The minimum value of `start_column: real_column` is 1.
+      #       So, the minimum value of `last_column` should be 1.
+      #       And non-zero value of `last_column` should be used as is.
       def hash_for_location(offense)
         {
           start_line:   offense.line,
           start_column: offense.real_column,
           last_line:    offense.last_line,
-          last_column:  offense.last_column,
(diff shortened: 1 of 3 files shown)


bbatsov:
It's a bit ugly indeed, but it solves the problem. Thanks!

lib/rubocop/formatter/json_formatter.rb
```diff
@@ -59,12 +59,15 @@ def hash_for_offense(offense)
...
+      #       The minimum value of `start_column: real_column` is 1.
+      #       So, the minimum value of `last_column` should be 1.
+      #       And non-zero value of `last_column` should be used as is.
       def hash_for_location(offense)
         {
           start_line:   offense.line,
           start_column: offense.real_column,
           last_line:    offense.last_line,
-          last_column:  offense.last_column,
+          last_column:  offense.last_column.zero? ? 1 : offense.last_column,
```
radomirbosak, line 70: Sorry, I'm not familiar with the Rubocop codebase.

This change prevents a situation `start_column==1, last_column==0` from happening, but can something like `start_column==5, last_column==4` still happen?

radomirbosak:
Thank you very much for fixing this! Would it be possible to release this fix in a patch release?

PR #10243, opened 2021-11-12
Add new `Gemspec/RequireMFA` cop

dvandersluis:
Following #10239, adds a new `Gemspec` cop to ensure that the `rubygems_mfa_required` metadata is set.

By default operates in `Strict` mode which requires `rubygems_mfa_required` to be set to `true`. If given `Strict: false`, it just checks that it is given a value.

-----------------

* [x] Added tests.


bbatsov:
Looks good to me. I'm just wondering about the name `Strict`, as I might not have guessed what it does without reading your PR description.

dvandersluis:
Happy to change it if you have any ideas!

koic:
The `Strict` option looks unnecessary. I think it's better to make sure that MFA is always enabled because it's about security.
If there is any reason why MFA cannot be enabled, the user has the option of disabling cop instead of `Strict: false`. This will simplify the use and design of cop.

I've heard talks by RubyGems developer about enabling MFA at past conferences. I think it's good to be able to detect this :-)

bbatsov:
That would be fine by me as it solves the problem with the name. :D

dvandersluis:
@bbatsov @koic I absolutely agree, my rationale for creating this was that we can encourage every gem that uses RuboCop to require MFA for pushes, which is best for the community.

I've updated the PR to not have the `Strict` config anymore. I originally added it because the RubyGems docs state that you can set the value to `'false'`, but disabling the cop in that case is fine too.

koic:
Thank you for the update! This looks good to me. Can you squash your commits into one? As noted in Martin Fowler's "Refactoring 2nd edition", refactoring is closely tied to addition of new feature. Therefore once commits are separated, the context in which the refactoring took place is unknown.

dvandersluis:
Sure, I squashed the autocorrection commit in. I think the refactoring commit makes sense on its own because it's affecting multiple cops. I'm partial to [@marcandre's blog](https://blog.marc-andre.ca/2014/02/05/why-i-wont-squash-my-commits/) about it :grin:

koic:
As a most active contributor for several years, I like the pragmatic policies of Martin Fowler (No one can talk about refactoring better than he). Anyway that's it for this topic, as it's not the main subject of this PR and I'm grateful for your contribution :-)

config/default.yml
```diff
@@ -259,6 +259,15 @@ Gemspec/OrderedDependencies:
   Include:
     - '**/*.gemspec'
 
+Gemspec/RequireMFA:
+  Description: 'Checks that the gemspec has metadata to require MFA from Rubygems.'
```
koic, line 263: That's just my two cents.

```suggestion
  Description: 'Checks that the gemspec has metadata to require MFA from RubyGems.'
```

spec/rubocop/cop/gemspec/require_mfa_spec.rb
```diff
@@ -0,0 +1,175 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Gemspec::RequireMFA, :config do
+  context 'when the gemspec is blank' do
+    it 'registers an offense but does not correct' do
+      expect_offense(<<~RUBY, 'my.gemspec')
+        ^{} `metadata['rubygems_mfa_required']` must be set to `'true'`.
```
koic, line 7: If blank, this cop will be able to skip the process (In other words, it would be `expect_no_offenses`). Because there are many other things missing 😅

dvandersluis: Removed.

bbatsov (approved):

PR #10247, opened 2021-11-15
Fix docs for `Gemspec/RequireMFA`

dvandersluis:
Examples for `Gemspec/RequireMFA` are not showing up properly, so this should fix it.


ShockwaveNN:
Just a side note - maybe remove those examples all together and replace them with link to

https://guides.rubygems.org/mfa-requirement-opt-in/ ?

Because if rubygems for some reason add another command to that list - `rubocop` will need to update its own docs to

dvandersluis:
That page is already linked in the documentation, but that's a fair point, I'll make it clearer that it's not necessarily an exhaustive list.

Issue #8252, opened 2020-07-07
--safe-autocorrect typo?


ybiquitous:
Hi, I found the `--safe-auto-correct` option has been renamed to `--safe-autocorrect` since v0.87.0.
It looks like a typo to me.

## Expected behavior

`rubocop --safe-auto-correct` should succeed.

## Actual behavior

```console
$ rubocop --safe-auto-correct
invalid option: --safe-auto-correct
Did you mean?  safe-autocorrect
For usage information, use --help
```

## Steps to reproduce the problem

```console
$ gem install rubocop:0.87.0
...

$ rubocop --safe-auto-correct
invalid option: --safe-auto-correct
...
```

## RuboCop version

Include the output of `rubocop -V` or `bundle exec rubocop -V` if using Bundler. Here's an example:

```console
$ rubocop -V
0.87.0 (using Parser 2.7.1.4, rubocop-ast 0.1.0, running on ruby 2.7.1 x86_64-darwin19)
```

ybiquitous:
Thank you!

PR #10256, opened 2021-11-17
Fix --autocorrect typo in changelog

MaxLap:
Should have been --auto-correct and --auto-correct-all. The error is only in the changelog.

Sadly, `--safe-autocorrect`, which is the wrong name, was actually how this was implemented. This was fixed later in https://github.com/rubocop/rubocop/issues/8252. So that typo must remain.


bbatsov:
Thanks!

How did you even spot this??? :-)

MaxLap:
Haha, I wanted to know in which version --auto-correct-all was added, but I couldn't find it, so I investigated.

PR #10260, opened 2021-11-20
Add new `Style/FileWrite` cop

leoarnold:
In his 2016 talk ["Little Snippets"](https://youtu.be/mC9TyVeER_8?t=560) Xavier Noria (@fxn) points out that Ruby 1.8 had the idiom

```ruby
File.open(filename, 'w') do |f|
  f.write(content)
end
```

whereas starting with Ruby 1.9 this can be replaced with

```ruby
File.write(filename, content)
```

This commit introduces the `Style/FileWrite` cop which investigates
said offense and similar cases where the convenience methods
`File.write` or `File.binwrite` can be used.

-----------------

* [x] Added tests.


bbatsov:
I like the proposal. It'd be good to cover this and your other suggestion in the Ruby style guide as well.

leoarnold:
@bbatsov Done: https://github.com/rubocop/ruby-style-guide/pull/890

dvandersluis:
Thanks for making the changes I proposed on #10261 here as well! Now that there basically is the same structure to the two cops, we should probably extract the shared code out to a shared mixin. You could use arguments for the file mode sets and method name (see https://docs.rubocop.org/rubocop-ast/node_pattern.html#for-arguments).

leoarnold:
@dvandersluis Is it really worth it to extract the methods? So far, both cops are self-contained and I cannot envision future cops to make use of the extracted method. As far as I can see, it would just create unnecessary cohesion.

dvandersluis:
@leoarnold it's not mandatory, but it is a pattern we tend to follow when we would otherwise have duplicated code. `FileRead` and `FileWrite` here are very similar cops so I think it makes sense.

bbatsov:
The changes look good, but your branch has to be rebased on top of the current `master` branch due to merge conflicts. I'm ambivalent towards the duplicated code, as we can address this down the road as well.

bbatsov:
Thanks!

PR #10261, opened 2021-11-20
Add new `Style/FileRead` cop

leoarnold:
Complementing the `Style/FileWrite` cop proposed in #10260
we add a `Style/FileRead` cop which investigates file reads like

```ruby
File.open(filename, 'r') do |f|
  f.read
end
```
and similar cases where the convenience methods
`File.read` or `File.binread` can be used.

-----------------

* [x] Added tests.


bbatsov:
Btw, I assume we don't want to recommend this for multiple reads in the same block, right? Perhaps you should expand a bit the rationale section with something on the subject.

leoarnold:
@bbatsov I am a bit unclear about your "multiple read" scenario. Could you please give a code example?

lib/rubocop/cop/style/file_read.rb
```diff
@@ -0,0 +1,117 @@
...
+      #   File.binread(filename)
+      #
+      class FileRead < Base
+        extend AutoCorrector
+
+        MSG = 'Use `File.%<read_method>s`.'
+
+        RESTRICT_ON_SEND = %i[open].to_set.freeze
+
+        READ_FROM_BOF_MODES = %w[r rt rb r+ r+t r+b].to_set.freeze
```
pirj, line 45: I apologize for my ignorance, hat does BOF stand for?

lib/rubocop/cop/style/file_read.rb
```diff
@@ -0,0 +1,117 @@
...
+            )
+          }
+        PATTERN
+
+        # @!method file_open_read_with_mode(node)
+        def_node_matcher :file_open_read_with_mode, <<~PATTERN
+          {
+            (block
+              (send
+                $(const {nil? cbase} :File) :open $_ (str $%READ_FROM_BOF_MODES)
```
pirj, line 74: Wondering if it's possible to use `$[? %READ_FROM_BOF_MODES]` and collapse `file_open_read_with_mode` and `file_open_read` into one node pattern.

lib/rubocop/cop/style/file_read.rb
```diff
@@ -0,0 +1,117 @@
...
+        PATTERN
+
+        def on_send(node)
+          investigate(node.parent) || investigate(node)
+        end
+
+        private
+
+        def investigate(node)
+          return unless (file_const, filename, mode = evidence(node))
```
pirj, line 98:
```ruby
evidence(node) do |file_const, filename, mode|
```
can be handy.

lib/rubocop/result_cache.rb
```diff
@@ -179,7 +179,7 @@ def rubocop_checksum
             .select { |path| File.file?(path) }
             .sort!
             .each do |path|
-              content = File.open(path, 'rb', &:read)
+              content = File.binread(path)
```
pirj, line 182: ❤️

spec/rubocop/cop/style/file_read_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+      ::File.open(filename) do |f|
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `File.read`.
+        f.read
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      ::File.read(filename)
+    RUBY
+  end
```
pirj, line 94: Subjectively, combining cbase check with every other example seems excessive.

dvandersluis:
Your two matchers can be combined into one:

```ruby
def_node_matcher :file_open_read, <<~PATTERN
  {
    (block
      (send
        $(const {nil? cbase} :File) :open $_ (str $%READ_FROM_BOF_MODES)?
      )
      (args _)
      (send _ :read)
    )
    (send
      (send
        $(const {nil? cbase} :File) :open $_ (str $%READ_FROM_BOF_MODES)?
      )
      :read
    )
    (send
      $(const {nil? cbase} :File) :open $_ (str $%READ_FROM_BOF_MODES)? (block-pass (sym :read))
    )
  }
PATTERN
```

How you use this would have to change slightly, because the optional capture would return in an array, but otherwise, everything would still work.

dvandersluis (changes requested):
The patterns here are overly complex and I think unnecessarily so. They can be basically boiled down to two questions:
1. Is the send node calling `open` relevant for this cop?
2. Does that node have a parent or a block-pass that calls read.

You can write pattern matchers to answer these two questions and collect all the data you need:

```ruby
# @!method file_open?(node)
def_node_matcher :file_open?, <<~PATTERN
  (send
    (const {nil? cbase} :File)
    :open
    $_
    (str $%READ_FROM_BOF_MODES)?
    $(block-pass (sym :read))?
  )
PATTERN

# @!method file_open_read?(node)
def_node_matcher :file_open_read?, <<~PATTERN
  {
    (block _ _ (send _ :read))
    (send _ :read)
  }
PATTERN
```

And then use them in investigate (where `get_read_node` is a helper method to either return the node containing the block pass, or the parent `block` or `send` node)

```ruby
return unless (filename, mode, block_pass = file_open?(node))
return unless read_node = get_read_node(node, block_pass)

def get_read_node(node, block_pass)
  if block_pass.any?
    node
  elsif file_open_read?(node.parent)
    node.parent
  end
end
```

In my example I'm also capturing less than you are because you can simplify your corrector too by not including code that won't change:

```ruby
include RangeHelp

# ...

add_offense(read_node, message: format(MSG, read_method: read_method(mode))) do |corrector|
  range = range_between(node.loc.selector.begin_pos, read_node.loc.expression.end_pos)
  replacement = "#{read_method(mode)}(#{filename.source})"

  corrector.replace(range, replacement)
end
```

lib/rubocop/cop/style/file_read.rb
dvandersluis: This is a cool pattern that I don't go to enough!

lib/rubocop/cop/style/file_read.rb
leoarnold: Beginning of file, the opposite of EOF

lib/rubocop/cop/style/file_read.rb
```diff
@@ -0,0 +1,117 @@
...
+      #   File.open(filename) { |f| f.read }
+      #   File.open(filename) do |f|
+      #     f.read
+      #   end
+      #   File.open(filename, 'r').read
+      #   File.open(filename, 'r', &:read)
+      #   File.open(filename, 'r') { |f| f.read }
+      #   File.open(filename, 'r') do |f|
+      #     f.read
+      #   end
```
dvandersluis, line 21: These two examples are fundamentally the same so we don't really need both (same for lines 12-15)

lib/rubocop/cop/style/file_read.rb
leoarnold: It works with `(str $%READ_FROM_BOF_MODES)?`. Will be in the next push.

lib/rubocop/cop/style/file_read.rb
leoarnold: Will rename constant to `READ_FILE_START_TO_FINISH_MODES` because it also matters that we are reading all of it.

spec/rubocop/cop/style/file_read_spec.rb
leoarnold: There's 4 cbase checks. Three are needed to cover the 3 branches of the node matcher. Shall I delete one?

leoarnold: After adding @dvandersluis' suggestion, we now only have one branch, so I deleted all but one test.

dvandersluis (commented):
Thanks for making my requested changes!

dvandersluis (approved):

dvandersluis:
One consideration to think about: in the block form, there's no check that the block arg is the receiver of `read`; ie. this would be a false positive:

```ruby
File.open('foo.txt') do |f|
  something_else.read
end
```

If we don't fix this the cop should be marked unsafe most likely (and a `@safety` section added to the class comment block to explain why the cop is unsafe).

leoarnold:
@dvandersluis Addressed. I added your snippet as a spec.

lib/rubocop/cop/style/file_read.rb
```diff
@@ -0,0 +1,110 @@
...
+            end
+          end
+        end
+
+        private
+
+        def evidence(node)
+          file_open?(node) do |filename, mode_array, block_pass|
+            read_node?(node, block_pass) do |read_node|
+              yield(filename, mode_array.first || 'r', read_node)
```
pirj, line 85: Just out of curiosity - what are the rest of the elements of this array?

lib/rubocop/cop/style/file_read.rb
```diff
@@ -0,0 +1,110 @@
...
+      #   # good
+      #   File.binread(filename)
+      #
+      class FileRead < Base
+        extend AutoCorrector
+        include RangeHelp
+
+        MSG = 'Use `File.%<read_method>s`.'
+
+        RESTRICT_ON_SEND = %i[open].to_set.freeze
```
pirj, line 42: Cosmetic `to_set` is redundant.

pirj (approved):
Thank you!

lib/rubocop/cop/style/file_read.rb
leoarnold: I added it because of https://github.com/rubocop/rubocop-performance/issues/131

lib/rubocop/cop/style/file_read.rb
leoarnold: `(str $%READ_FILE_START_TO_FINISH_MODES)?` will either return `[]` or `['r']` (i.e. an array with just a single element).

lib/rubocop/cop/style/file_read.rb
pirj: I was unaware of that, but it turns out the set doesn't help much https://github.com/rubocop/rubocop/blob/de4f7cdd266999bf2cfbe78bd7fd8a0d4580863d/lib/rubocop/cop/base.rb#L305

lib/rubocop/cop/style/file_read.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Favor `File.(bin)read` convenience methods.
+      #
+      # @example text mode
```
koic, line 8: `@example` is used for user configurable options such as `EnforcedStyle`. Annotations should be expressed using comments. The following is an example.

```suggestion
      # @example
      #
      # Examples for text mode.
```

lib/rubocop/cop/style/file_read.rb
```diff
@@ -0,0 +1,110 @@
...
+      #   File.open(filename, 'r').read
+      #   File.open(filename, 'r', &:read)
+      #   File.open(filename, 'r') do |f|
+      #     f.read
+      #   end
+      #
+      #   # good
+      #   File.read(filename)
+      #
+      # @example binary mode
```
koic, line 25: Ditto.

PR #10271, opened 2021-11-26
Add logic to ensure badge information (Department/Name) in CamelCase

timtoronto634:
This PR enforces the badge information (Department/Name) is stored in CamelCase.

Before, the budge information can be stored in snake_case, and so I could run, for example, ` bundle exec rake 'new_cop[ex_snake/case]' ` .
This, however, creates a file `lib/rubocop/cop/ex_snake/case.rb` with class name in lowercase, and results in syntax error.
```
 class/module name must be CONSTANT (SyntaxError)
```

The expected behavior would be either `A. return error when task is called`  or  `B. create files with CamelCase`
This PR implemented `B`

I first fixed how the template file is prepared, but noticed that this PR is more straightforward.  I recognize that the potential influence can be larger than just fixing template, but I did not find any problem so far. Also this is better solution in a sense that the badge does not expect the infos to be in snake_case.

-----------------

* [x] Added tests.


bbatsov:
I think the change is fine, but we also need some tests that show the problem you mentioned has been solved.

PR #10283, opened 2021-12-04
[Fix #10282] Fix an incorrect autocorrect for `Style/EmptyCaseCondition`

### LINKED ISSUE #10282: `Style/EmptyCaseCondition`autofix breaks code when returning from case with inline `when..then` syntax
Autofix for `Style/EmptyCaseConditional` breaks code when the original code returns from a case statement that has an inline `when..then`.

--------

## Expected behavior

```ruby
# example.rb

def example
  return case
    when foo then :foo
    else :bar
  end
end
```

Autofix to working format (or leave as is).

## Actual behavior
```sh
rubocop example.rb -a --only Style/EmptyCaseConditional
```

generates

```ruby
def example
  return if foo then :foo
    else :bar
  end
end
```

Running it fails:

```
$ ruby example.rb

example.rb:2: syntax error, unexpected keyword_then, expecting keyword_end
  return if expression.foo then :foo
                           ^~~~
example.rb:4: warning: else without rescue is useless
example.rb:5: syntax error, unexpected keyword_end, expecting end-of-input
```

## RuboCop version

```
$ rubocop -V
1.23.0 (using Parser 3.0.3.1, rubocop-ast 1.14.0, running on ruby 2.5.8 x86_64-darwin20)
```

koic:
Fixes #10282.

This PR fixes an incorrect autocorrect for `Style/EmptyCaseCondition` when using `when ... then` in `case` in `return`.

This issue occurs not only with `return`, also with method calls.

And indentation of auto-corrected code will be entrusted to another existing cop.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_incorrect_autocorrect_for_style_empty_case_condition.md
+++ b/changelog/fix_incorrect_autocorrect_for_style_empty_case_condition.md
@@ -1,0 +1,1 @@
+* [#10282](https://github.com/rubocop/rubocop/issues/10282): Fix an incorrect autocorrect for `Style/EmptyCaseCondition` when using `when ... then` in `case` in a method call. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #10289, opened 2021-12-11
Support Ruby 3.1's hash value omission for `Style/HashSyntax`

koic:
## Summary

Ruby 3.1 will be introduced hash shorthand syntax:
https://bugs.ruby-lang.org/issues/17292.

The `EnforcedShorthandSyntax` option checks hash literal value.
It can enforce either the use of the explicit hash value syntax or the use of Ruby 3.1 hash value shorthand syntax (when applicable).

A separate offense is registered for each problematic hash literal value.

The option values are:

* always - forces use of the 3.1 syntax (e.g. `{foo:}`) when the hash key and value are the same.
* never - forces use of explicit hash literal value

```ruby
# EnforcedShorthandSyntax: always (default)

# bad
{foo: foo, bar: bar}

# good
{foo:, bar:}

# EnforcedShorthandSyntax: never

# bad
{foo:, bar:}

# good
{foo: foo, bar: bar}
```

## Additional Information

This cop is based on the upgrade from ES5 to ES6 with ESLint.

The ES6 shorthand syntax is chosen by default for ESLint shorthands:

> "always" (default) expects that the shorthand will be used whenever possible.

https://eslint.org/docs/rules/object-shorthand

The cop will also choose the new Ruby 3.1 shorthand syntax by default.

-----------------

* [x] Added tests.


koic:
And I opened a PR to The Ruby Style Guide.
https://github.com/rubocop/ruby-style-guide/pull/892

bbatsov:
Won't it be better to just add a boolean `use_shorthand_syntax` option (or something like this) to the existing `HashSyntax` cop? It seems like an overkill to add a separate cop for this.

koic:
Ah, that makes sense to me! I updated this PR.

And I added `EnforcedShorthandSyntax` to `Style/HashSyntax` instead of `UseShorthandSyntax`.
This is because if the option requires a trivalue like `mixed` (string value type) in the future, the type will be unmatched with `UseShorthandSyntax` which accepts `true` or` false` (boolean value type).

bbatsov:
Good idea! Thanks!

PR #10290, opened 2021-12-11
Add new `Naming/BlockForwarding` cop

koic:
This PR adds new `Naming/BlockForwarding` cop.

Ruby 3.1 will be introduced anonymous block forwarding syntax:
https://bugs.ruby-lang.org/issues/11256

This cop identifies places where `do_something(&block)` can be replaced by `do_something(&)`.

```ruby
# bad
def foo(&block)
  bar(&block)
end

# good
def foo(&)
  bar(&)
end
```

AFAIK, in most cases, block argument is given name similar to `&block` or `&proc`. Their names have no information and `&` will be sufficient for syntactic meaning.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_add_new_naming_block_forwarding_cop.md
+++ b/changelog/new_add_new_naming_block_forwarding_cop.md
@@ -1,0 +1,1 @@
+* [#10290](https://github.com/rubocop/rubocop/pull/10290): Add new `Naming/BlockForwarding` cop. ([@koic][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2487,6 +2487,15 @@ Naming/BinaryOperatorParameterName:
   VersionAdded: '0.50'
   VersionChanged: '1.2'
 
+Naming/BlockForwarding:
+  Description: 'Use anonymous block forwarding.'
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  EnforcedStyle: anonymous
+  SupportedStyles:
+    - anonymous
+    - explicit
+
 Naming/BlockParameterName:
   Description: >-
                  Checks for block parameter names that contain capital letters,
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -407,6 +407,7 @@ require_relative 'rubocop/cop/metrics/perceived_complexity'
 
 require_relative 'rubocop/cop/naming/accessor_method_name'
 require_relative 'rubocop/cop/naming/ascii_identifiers'
+require_relative 'rubocop/cop/naming/block_forwarding'
 require_relative 'rubocop/cop/naming/block_parameter_name'
 require_relative 'rubocop/cop/naming/class_and_module_camel_case'
 require_relative 'rubocop/cop/naming/constant_name'
--- a/lib/rubocop/cop/naming/block_forwarding.rb
+++ b/lib/rubocop/cop/naming/block_forwarding.rb
@@ -1,0 +1,102 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Naming
+      # In Ruby 3.1, anonymous block forwarding has been added.
+      #
+      # This cop identifies places where `do_something(&block)` can be replaced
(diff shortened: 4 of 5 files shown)


bbatsov:
Perhaps `Naming/AnonymousBlockParam` is a better name here? I don't really see any forwarding going on, so I find the proposed name a bit confusing in the context of naming. Or maybe a configurable `Naming/BlockParam` that can be anonymous or explicit? I think I like the second options more, as it's more in line with our spirit of extremely flexibility.

bbatsov:
I guess this should go to the style guide as well.

bbatsov:
@koic ping :-)

koic:
Thank you for your feedback. Let's dig into my point of view.

Anonymous block argument can only be used for forwarding. For example, the following syntax error will occur outside of method definition scope.

```console
% ruby -ce 'foo(&)'
-e:1: no anonymous block parameter
```

As mentioned above, the use of anonymous block argument can "only be used in the limited context".

So, I don't have a strong opinion to the `Naming/AnonymousBlockForwarding`, but I'm not sure if the generic `Naming/AnonymousBlockParam` (or `Naming/BlockParam`) is better 💦 

On the other hand, I think the cop can provide an option to force block naming. However, I can't predict whether there will really be user requests that wants to force the clear syntax `foo(&)` to `foo(&block)` (or `foo(&proc)`, and similar names).

bbatsov:
Fair enough. I'm now find with having "forwarding" in the name, I'm mostly wondering whether it's a good name to have to cop be so specific. Probably `Naming/BlockForwarding` with styles like `anonymous` and `explicit` and maybe some extra option "BlockForwardingName" (or something along those lines). Perhaps I'm overthinking this, though, but it seems that this will address all the potential things you've mentioned.

koic:
And I opened https://github.com/rubocop/ruby-style-guide/pull/893 to the style guide.

dvandersluis:
>Implemented onymous alternative style. But if explicit alternative style name is better, I will update it.

`explicit` would be better, `onymous` is not really a common English word (and also technically means that something has a _specific_ name rather than is named at all).

koic:
I get it and updated to `explicit`. Thank you!

bbatsov:
Looks great now! Thanks!

whiteleaf7:
@koic 
オートコレクトで構文エラーな変換がされるようです
第２引数以降かつ後置 if がある状態で、カッコがない不正な式に変換されます

```rb
# frozen_string_literal: true

def foo _opt
  yield
end

def hoge(&block)
  foo 'option', &block if true
end

hoge { puts 'hoge' }
```

↓

```rb
# frozen_string_literal: true

def foo(_opt)
  yield
end

def hoge(&)
  foo 'option', & if true # here
end

hoge { puts 'hoge' }
```

```
rubocop tmp/t.rb -a
Running via Spring preloader in process 775
Inspecting 1 file
C

Offenses:

tmp/t.rb:3:9: C: [Corrected] Style/MethodDefParentheses: Use def with parentheses when there are parameters.
def foo _opt
        ^^^^
tmp/t.rb:7:10: C: [Corrected] Naming/BlockForwarding: Use anonymous block forwarding.
def hoge(&block)
         ^^^^^^
tmp/t.rb:8:17: C: [Corrected] Naming/BlockForwarding: Use anonymous block forwarding.
  foo 'option', &block
                ^^^^^^

1 file inspected, 3 offenses detected, 3 offenses corrected
```

```
ruby tmp/t.rb
tmp/t.rb:11: syntax error, unexpected end-of-input, expecting `end'
```

koic:
@whiteleaf7 This repository is communicated in English. Please send feedback in Japanese to https://github.com/rubocop/rubocop-jp/issues from the next opportunity. This time I will take the issue. Thank you.

koic:
@whiteleaf7 The issue has been resolved by https://github.com/rubocop/rubocop/pull/10333. Can you upgrade RuboCop to 1.24.1?

PR #10328, opened 2021-12-28
[Fix #10317] Fix a false positive for `Style/MethodCallWithArgsParentheses`

### LINKED ISSUE #10317: Fix Style/MethodCallWithArgsParentheses inconsistency with hash punning
## Expected behavior

With Ruby 3.1.0's new hash punning syntax, if you try to use punning on *more than one line without parentheses*, you'll get a Rubocop error. Here's an inelegant and contrived example where the problem is with the use of parentheses in the constructor:

``` ruby 
Example = Struct.new :one, :two, keyword_init: true

class ExampleWrapper
  def initialize one:, two:
    @first = Example.new(one:, two:)
    @second = Example.new(one:, two:)
  end
end

puts ExampleWrapper.new(one: 1, two: 2).inspect
```

I would expect the above to not throw an error and be valid code since the Ruby parser requires it.

## Actual behavior

When you run the above code as:

``` 
rubocop snippet.rb
```

...you'll get the following errors:

``` 
Offenses:

/Users/bkuhlmann/Engineering/Misc/snippet.rb:19:25: C: [Correctable] Style/MethodCallWithArgsParentheses: Omit parentheses for method calls with arguments.
    @first = Example.new(one:, two:)
                        ^^^^^^^^^^^^
/Users/bkuhlmann/Engineering/Misc/snippet.rb:20:26: C: [Correctable] Style/MethodCallWithArgsParentheses: Omit parentheses for method calls with arguments.
    @second = Example.new(one:, two:)
                         ^^^^^^^^^^^^
```

Should you allow Rubocop to auto-correct the above code, you'll end up with the following implementation (truncated for brevity):

``` ruby
  def initialize one:, two:
    @first = Example.new one:, two:
    @second = Example.new one:, two:
  end
```

Without the parenthesis, you'll then have `Lint/Syntax` errors which you can confirm when attempting to run the code snippet:

``` 
/Users/bkuhlmann/Engineering/Misc/snippet.rb:20: syntax error, unexpected label, expecting `end'
    @second = Example.new one:, two:
/Users/bkuhlmann/Engineering/Misc/snippet.rb:24: syntax error, unexpected end-of-input, expecting `end'
...per.new(one: 1, two: 2).inspect
```

The problem appears when you are using hash punning on more than one line in your implementation. If you were to write code like this:

``` ruby
class ExampleWrapper
  def initialize one:, two:
    @first = Example.new one:, two:
  end
end
```

This would be syntactically correct for *both* Rubocop and Ruby. You only need the extra parentheses when using punning on more than one line. Which seems weird for Ruby too? Should this be logged as in issue with Ruby core as well? 

1. Save the above code snippet as `snippet.rb`.
2. Enable the `Style/MethodCallWithArgsParentheses` cop.
3. Run as `rubocop snippet.rb`.

## RuboCop version

```
$ [bundle exec] rubocop -V
1.24.0 (using Parser 3.0.3.2, rubocop-ast 1.15.0, running on ruby 3.1.0 arm64-darwin21.2.0)
  - rubocop-performance 1.13.0
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.6.0
```

koic:
Fixes #10317.

This PR fixes a false positive for `Style/MethodCallWithArgsParentheses` when using hash value omission.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positive_for_style_method_call_with_args_parentheses.md
+++ b/changelog/fix_false_positive_for_style_method_call_with_args_parentheses.md
@@ -1,0 +1,1 @@
+* [#10317](https://github.com/rubocop/rubocop/issues/10317): Fix a false positive for `Style/MethodCallWithArgsParentheses` when using hash value omission. ([@koic][])
--- a/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
+++ b/lib/rubocop/cop/style/method_call_with_args_parentheses.rb
@@ -43,9 +43,11 @@ module RuboCop
       # NOTE: Parentheses are still allowed in cases where omitting them
       # results in ambiguous or syntactically incorrect code. For example,
       # parentheses are required around a method with arguments when inside an
-      # endless method definition introduced in Ruby 3.0.  Parentheses are also
+      # endless method definition introduced in Ruby 3.0. Parentheses are also
       # allowed when forwarding arguments with the triple-dot syntax introduced
       # in Ruby 2.7 as omitting them starts an endless range.
+      # And Ruby 3.1's hash omission syntax has a case that requires parentheses
+      # because the issue https://bugs.ruby-lang.org/issues/18396.
       #
       # @example EnforcedStyle: require_parentheses (default)
       #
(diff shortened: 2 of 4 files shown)


bbatsov:
I think that this exception should also be explained in the cop's examples.

koic:
I've added to `NOTE` as well as similar exception cases.

bbatsov:
Thanks!

PR #10332, opened 2021-12-29
Fix an incorrect autocorrect for `Gemspec/RequireMFA` when .gemspec file contains `metadata` keys assignments

### LINKED ISSUE #10251: Gemspec/RequireMFA autocorrect does not detect some existing setting of metadata
## Expected behavior

Given a gemspec containing lines like the following:
```ruby
  spec.metadata["source_code_uri"] = some_uri
```

Gemspec/RequireMFA should autocorrect by setting just `spec.metadata["rubygems_mfa_required"]`.

## Actual behavior

The autocorrect adds the following lines:
```ruby
  spec.metadata = {
    "rubygems_mfa_required" => "true"
  }
```
Which means the existing metadata entries are removed.

1. Create a file `foo.gemspec` containing:
```ruby
# frozen_string_literal: true

  spec.metadata['source_code_uri'] = 'https://foo.example.com/'
end
```
2. Create a file `.rubocop.yml` containing:
```yml
AllCops:
  NewCops: enable
```
3. Run `rubocop -a`

`foo.gemspec` now contains:
```ruby
# frozen_string_literal: true

  spec.metadata['source_code_uri'] = 'https://foo.example.com/'
  spec.metadata = {
    'rubygems_mfa_required' => 'true'
  }
end
```

## RuboCop version

```
$ [bundle exec] rubocop -V
1.23.0 (using Parser 3.0.2.0, rubocop-ast 1.13.0, running on ruby 3.0.2 x86_64-linux)
```

fatkodima:
When we have `spec.metadata = {...}`, insert `rubygems_mfa_required` setting there. 
Otherwise, just always add `spec.metadata["rubygems_mfa_required"] = "true"`. 

Closes #10251

### DIFF
--- a/changelog/fix_incorrect_autocorrect_for_gemspec_require_mfa.md
+++ b/changelog/fix_incorrect_autocorrect_for_gemspec_require_mfa.md
@@ -1,0 +1,1 @@
+* [#10251](https://github.com/rubocop/rubocop/issues/10251): Fix an incorrect autocorrect for `Gemspec/RequireMFA` when .gemspec file contains `metadata` keys assignments. ([@fatkodima][])
--- a/lib/rubocop/cop/gemspec/require_mfa.rb
+++ b/lib/rubocop/cop/gemspec/require_mfa.rb
@@ -117,7 +117,7 @@ module RuboCop
 
             correct_metadata(corrector, metadata)
           else
-            correct_missing_metadata(corrector, node, block_var)
+            insert_mfa_required(corrector, node, block_var)
           end
         end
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Looking at the code and the tests the auto-correction is always the same whether there is metadata hash or not.

fatkodima:
https://github.com/rubocop/rubocop/pull/10332/files#diff-25aba207f4aafaea62f802eb10fa835d19ae4a61ebaaa4a79f21e902febd49fdR118-R121

When there was a hash, it inserts in it:
https://github.com/rubocop/rubocop/blob/6821569eeff14926a776850aaf74273eef6d923b/spec/rubocop/cop/gemspec/require_mfa_spec.rb#L73-L95

Otherwise in adds a new line:
https://github.com/rubocop/rubocop/pull/10332/files#diff-ed5e79cdcd0fdfbd4adf880cea63c284a45bcda07f0b4bd8d5396a8b29db0be0R117-R122

Or I missed something?

bbatsov:
My bad. Thanks!

PR #10333, opened 2021-12-30
Fix an incorrect autocorrect for `Naming/BlockForwarding`

koic:
This PR fixes an incorrect autocorrect for `Naming/BlockForwarding` using explicit block forwarding without method definition parentheses.

First, specify named block variable explicitly.

```ruby
% cat example.rb
def foo(&blk)
  bar
end
```

```console
% ruby-parse example.rb
(def :foo
  (args
    (blockarg :blk))
  (send nil :bar))
```

Next, specify anonymous block forwarding.

```ruby
% cat example.rb
def foo(&)
  bar
end
```

The structure of AST is the same.

```console
% ruby-parse example.rb
(def :foo
  (args
    (blockarg nil))
  (send nil :bar))
```

Finally, omit the method definition parentheses.

```ruby
% cat example.rb
def foo &
  bar
end
```

The structure of the AST will breaking change.

```console
% ruby-parse example.rb
(def :foo
  (args
    (blockarg :bar)) nil)
```

This PR enforces parentheses to maintain syntax compatibility for autocorrection.

-----------------

* [] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Added tests.

### DIFF
--- a/changelog/fix_incorrect_autocorrect_for_naming_block_forwarding.md
+++ b/changelog/fix_incorrect_autocorrect_for_naming_block_forwarding.md
@@ -1,0 +1,1 @@
+* [#10333](https://github.com/rubocop/rubocop/pull/10333): Fix an incorrect autocorrect for `Naming/BlockForwarding` using explicit block forwarding without method definition parentheses. ([@koic][])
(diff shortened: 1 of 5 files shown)


bbatsov:
Thanks!

PR #10351, opened 2022-01-08
Support `EnforcedShorthandSyntax: either` for `Style/HashSyntax`

koic:
Follow up to https://github.com/testdouble/standard/issues/375#issuecomment-1007505392.

This PR supports `EnforcedShorthandSyntax: either` option for `Style/HashSyntax`.
It accepts both shorthand and explicit use of hash literal value.

```ruby
# good
{foo: foo, bar: bar}

# good
{foo:, bar:}
```

I think that we can change the default enforced style in response to future feedback and discussion. Anyway at this point it only adds the new style.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_support_enforced_shorthand_syntax_either_option_for_style_hash_syntax.md
+++ b/changelog/new_support_enforced_shorthand_syntax_either_option_for_style_hash_syntax.md
@@ -1,0 +1,1 @@
+* [#10351](https://github.com/rubocop/rubocop/pull/10351): Support `EnforcedShorthandSyntax: either` option for `Style/HashSyntax`. ([@koic][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -3699,6 +3699,8 @@ Style/HashSyntax:
     - always
     # forces use of explicit hash literal value.
     - never
+    # accepts both shorthand and explicit use of hash literal value.
+    - either
   # Force hashes that have a symbol value to use hash rockets
   UseHashRocketsWithSymbolValues: false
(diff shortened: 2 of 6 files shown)


searls:
Thanks!

bbatsov:
> I think that we can change the default enforced style in response to future feedback and discussion. Anyway at this point it only adds the new style.

Agreed. Perhaps `either` is a better default, but let's wait a bit.

jibiel:
@koic Thanks for the third option. :neckbeard:

I find it that the new short syntax impairs readability a bit in the case when not all of the values are omitted and the hash is split into lines:

```ruby
{
  attribute: 'some_attribute_value',
  code:,
  message:
}
```
vs
```ruby
{
  attribute: 'some_attribute_value',
  code: code,
  message: message
}
```

Feels like something was missed unintentionally. Maybe it's a part of getting used to the new confusing syntax. 🙂 

Anyways, judging by the [original discussion](https://bugs.ruby-lang.org/issues/14579), I'd say that the new short syntax was introduced mostly with a 'one-liner' and a 'symmetric key/value hash' cases in mind.

So that's a 👍🏾 for `either` as a default option value.

P.S.

I would shortened the option to `EnforcedValueSyntax: full || short || mixed`. Although it may introduce some naming clashes with `EnforcedStyle`. :neckbeard:

jcraigk:
After applying the new shorthand syntax to a few projects and getting feedback from other engineers, I think `either` will be the best default. I'm not sure the loud dissenters of shorthand syntax will ever be assuaged, and perhaps that's for the best.

PR #10352, opened 2022-01-09
[Fix #8032] Improve argument alignment correction for kwargs

### LINKED ISSUE #8032: Layout/ArgumentAlignment vs Layout/HashAlignment
I am not sure it is a bug per se, but it is sure perceived as one.

In our codebase, we have `Layout/ArgumentAlignment` turned on, and `Layout/HashAlignment` turned **off** (the reason of the latter is like I describe in #7880: our preferred style is `EnforcedColonStyle: key`, but from time to time it is necessary to have a hash with `EnforcedColonStyle: table` or `EnforcedColonStyle: separator` -- especially in constants, where "how the constant looks" is important for grasping the meaning).

Now, (after autofixing `Layout/LineLength`, but the same could've happened on manual editing) I have code elements looking like this:
```ruby
            create(:edit_rating, transcription_job: transcription_job, work_type: 'review', user: reviewer, stars: 5,
created_at: 1.month.ago)
```
(note the second line).

Maybe it is just me, but I honestly expected `Layout/ArgumentAlignment` to fix this, and only after debugging into Rubocop's sources I understood why it does not (because all arguments to the method except `:edit_rating` considered one hash by the parser, and their alignment to be `HashAlignment`'s responsibility).

But even if I turn `HashAlignment` on, the proposed alignment would be this:
```ruby
            create(:edit_rating, transcription_job: transcription_job, work_type: 'review', user: reviewer, stars: 5,
                                 created_at: 1.month.ago)
```
...while looking at them as a separate arguments, one might expect those (depending on `ArgumentAlignment` settings):

```ruby
            create(:edit_rating, transcription_job: transcription_job, work_type: 'review', user: reviewer, stars: 5,
                   created_at: 1.month.ago)
# or
            create(:edit_rating, transcription_job: transcription_job, work_type: 'review', user: reviewer, stars: 5,
              created_at: 1.month.ago)
```

Could it be reasonable to complicate `ArgumentAlignment` a bit to treat no-braces hash as a separate arguments?

```
$ [bundle exec] rubocop -V
0.84.0 (using Parser 2.7.1.2, rubocop-ast 0.0.3, running on ruby 2.6.3 x86_64-linux)
```

mvz:
This fixes detection of incorrect indentation for the case where
- ArgumentAlignment is configured with EnforcedStyle `with_fixed_indentation`
- A method call has both regular arguments and keyword arguments
- The first keyword argument is on the same line as at least one of the regular arguments

In the original code, the keyword arguments would be seen as one single argument, and because it starts on the same line as the previous argument, alignment adjustment would be skipped.

Also, because of the fixed indentation style, HashAlignment always skips adjustment of the keyword argument block. This means neither cop would correct the alignment.

This change adjusts this behaviour by treating each keyword argument as a separate argument if EnforcedStyle is `with_fixed_indentation`.

This has an additional effect in that each keyword argument will be separately annotated.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_improve_argument_alignment_correction.md
+++ b/changelog/fix_improve_argument_alignment_correction.md
@@ -1,0 +1,1 @@
+* [#8032](https://github.com/rubocop/rubocop/issues/8032): Improve ArgumentAlignment detection and correction for keyword arguments. ([@mvz][])
--- a/lib/rubocop/cop/layout/argument_alignment.rb
+++ b/lib/rubocop/cop/layout/argument_alignment.rb
@@ -53,23 +53,50 @@ module RuboCop
                            'following the first line of a multi-line method call.'
 
         def on_send(node)
-          first_arg = node.first_argument
-          return if !multiple_arguments?(node, first_arg) || (node.send_type? && node.method?(:[]=))
+          return if !multiple_arguments?(node) || (node.send_type? && node.method?(:[]=))
 
-          if first_arg.hash_type? && !first_arg.braces?
-            pairs = first_arg.pairs
-            check_alignment(pairs, base_column(node, pairs.first))
-          else
-            check_alignment(node.arguments, base_column(node, first_arg))
-          end
+          items = flattened_arguments(node)
+
+          check_alignment(items, base_column(node, items.first))
         end
+
         alias on_csend on_send
 
         private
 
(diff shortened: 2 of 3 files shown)


koic:
Thanks!

PR #10353, opened 2022-01-10
Use `:ambiguous_regexp` to detect ambiguous Regexp in Ruby 3

danieldiekmeier:
At work, we recently started using Rubocop via [Standard](https://github.com/testdouble/standard), and we still have a long `.rubocop_todo.yml`. Luckily, this helped to uncover a bug in RuboCop! :D

After upgrading a Rails app to Ruby 3.0 and setting `TargetRubyVersion: 3.0`, I noticed that all todos relating to `Lint/AmbiguousRegexpLiteral` disappeared. But we hadn't changed the files, and the "bad" code was still in there.

```diff
- # Offense count: 7
- # Cop supports --auto-correct.
- Lint/AmbiguousRegexpLiteral:
-   Exclude:
-     - 'spec/features/example/example_spec.rb'
-     - 'spec/models/example.rb'
-     - 'spec/views/admin/example/example.haml_spec.rb'
```

Digging further, a colleague and I noticed that this _only_ happened with `TargetRubyVersion: 3.0`, and that the offenses showed up when setting `TargetRubyVersion: 2.7`.

It took a while, but we finally found this code in the parser: https://github.com/whitequark/parser/blob/e160eeb50bf505453c5a9b1538451d2d62e6144a/lib/parser/lexer.rl#L1571-L1578, added in this commit: https://github.com/whitequark/parser/pull/768/files

It adds a new symbol alongside `:ambiguous_literal`, namely `:ambiguous_regexp`.

RuboCop only checks for  `:ambiguous_literal`, so the warning didn't propagate and was ultimately not displayed.

This PR uses the new `:ambiguous_regexp` if `target_ruby_version >= 3.0`.

Thanks for your consideration!

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_use_different_symbol_in_ruby_3.md
+++ b/changelog/fix_use_different_symbol_in_ruby_3.md
@@ -1,0 +1,1 @@
+* [#10353](https://github.com/rubocop/rubocop/pull/10353): Use `:ambiguous_regexp` to detect ambiguous Regexp in Ruby 3. ([@danieldiekmeier][], [@joergschiller][])
--- a/lib/rubocop/cop/lint/ambiguous_regexp_literal.rb
+++ b/lib/rubocop/cop/lint/ambiguous_regexp_literal.rb
@@ -30,7 +30,11 @@ module RuboCop
 
         def on_new_investigation
           processed_source.diagnostics.each do |diagnostic|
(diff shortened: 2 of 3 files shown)


koic:
Good catch! Thank you!

PR #10357, opened 2022-01-13
Fix "Style/HashSyntax: Omit the hash value cases" that breaks syntax for specific cases

berkos:
👋  Looking for feedback as my current PR could be affecting other cases that I might have not noticed.  🙏 
The PR is more on the safe side of not doing code changes that break the syntax although I understand this might not be the desired direction.

### Summary

There are certain cases that suggestion of Style/HashSyntax: Omit the hash value and then the autocorrect both (`-a` or `-A`) will produce invalid syntax. 

Here is a scenario of a specific code block:

```ruby
RSpec.shared_examples 'Examples' do |option_a:|
  include_examples 'Example A', option_a: option_a if option_a

  include_examples 'Example B', option_a: option_a

  describe '.describe' do
    it { true }
  end

  include_examples 'Example C', option_a: option_a
end
```

When we run the autocorrect currently will it produce the following result:

```ruby
RSpec.shared_examples 'Examples' do |option_a:|
  include_examples 'Example A', option_a: if option_a # Omitting the option_a value would produce invalid syntax

  include_examples 'Example B', option_a: # Omitting the option_a value would produce invalid syntax

  describe '.describe' do
    it { true }
  end

  include_examples 'Example C', option_a: # Valid omission
end
```

### Suggested change
This change acts conservative and does not change/suggest a change for those cases. More specifically it will make a change for the `Example C` and won't suggest something for the rest.

Code changes with the new version:

```ruby
RSpec.shared_examples 'Examples' do |option_a:|
  include_examples 'Example A', option_a: option_a if option_a

  include_examples 'Example B', option_a: option_a

  describe '.describe' do
    it { true }
  end

  include_examples 'Example C', option_a:
end
```

Only `Example C` omits the value.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
+++ b/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
@@ -46,10 +46,22 @@ module RuboCop
 
       def without_parentheses_call_expr_follows?(node)
         return false unless (ancestor = node.parent.parent)
-        return false unless (right_sibling = ancestor.right_sibling)
 
-        ancestor.respond_to?(:parenthesized?) && !ancestor.parenthesized? &&
-          right_sibling.respond_to?(:parenthesized?) && !right_sibling.parenthesized?
+        right_sibling = ancestor.right_sibling
+
+        return true if right_sibling.nil? && without_parentheses?(ancestor)
+        return false unless right_sibling
+        return true if node_with_block_and_arguments?(right_sibling)
+
+        without_parentheses?(ancestor) && without_parentheses?(right_sibling)
+      end
+
+      def without_parentheses?(node)
+        node.respond_to?(:parenthesized?) && !node.parenthesized?
+      end
+
+      def node_with_block_and_arguments?(node)
+        node.respond_to?(:block_type?) && node.block_type? && node.children&.first&.arguments?
       end
     end
   end
--- a/spec/rubocop/cop/style/hash_syntax_spec.rb
+++ b/spec/rubocop/cop/style/hash_syntax_spec.rb
@@ -819,7 +819,7 @@ RSpec.describe RuboCop::Cop::Style::HashSyntax, :config do
         RUBY
       end
 
-      it 'does not register an offense when hash valuees are omitted' do
+      it 'does not register an offense when hash values are omitted' do
         expect_no_offenses(<<~RUBY)
           {foo:, bar:}
         RUBY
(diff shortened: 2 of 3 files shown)


spec/rubocop/cop/style/hash_syntax_spec.rb
```diff
@@ -891,6 +891,21 @@
         RUBY
       end
 
+      it 'does not register an offense when one line condition follows' do
+        expect_no_offenses(<<~RUBY)
+          foo value: value if boo
```
koic, line 896: Good catch! This is trivia, `bar` is more common than `boo`.

```suggestion
          foo value: value if bar
```

There is the RFC 3092 developed in April Fool's Day :-)
https://datatracker.ietf.org/doc/html/rfc3092

ErikSchierboom:
We've had some issue with the hash value cases rule too.

```ruby
track = create :track
exercise = create :concept_exercise, track:
create :concept_solution, exercise:
```

leads to:

```
my_file.rb:3:8: E: Lint/Syntax: unexpected token tSYMBOL
(Using Ruby 3.1 parser; configure using TargetRubyVersion parameter, under AllCops)
create :concept_solution, exercise:
       ^^^^^^^^^^^^^^^^^
```

Convert `create` to a regular method call fixes the problem:

```ruby
track = create(:track)
exercise = create(:concept_exercise, track:)
create(:concept_solution, exercise:)
```

The original syntax _is_ valid Ruby code though (and works). 
It would be great if this PR fixes that (or a separate one).

spec/rubocop/cop/style/hash_syntax_spec.rb
berkos: I did not know about the RFC 😅😀.

spec/rubocop/cop/style/hash_syntax_spec.rb
```diff
@@ -891,6 +891,21 @@
...
+        RUBY
+      end
+
+      it 'does not register an offense when block follows' do
+        expect_no_offenses(<<~RUBY)
+          foo value: value
+          boo do
+            value
+          end
```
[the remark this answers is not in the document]
berkos: Thanks for the review @koic . 🙏  Good point about the valid syntax. I've added an extra use case. Can you please let me know how it looks and I can squash the commits after that.

lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
```diff
@@ -46,10 +46,22 @@ def require_hash_value?(hash_key_source, node)
...
+        right_sibling = ancestor.right_sibling
+
+        return true if right_sibling.nil? && not_parenthesized_node?(ancestor)
+        return false unless right_sibling
+        return true if node_with_block_and_arguments?(right_sibling)
+
+        not_parenthesized_node?(ancestor) && not_parenthesized_node?(right_sibling)
+      end
+
+      def not_parenthesized_node?(node)
```
koic, line 59: How about method name like the following?

```suggestion
      def without_parentheses?(node)
```

koic:
@berkos I left one comment, but others look good to me. Can you add a changelog entry and squash your commits into one?

> * [ ] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

@ErikSchierboom Yeah, that problem will be solved by this PR.

koic:
Thanks!

ErikSchierboom:
Great! Thanks @koic

spec/rubocop/cop/style/hash_syntax_spec.rb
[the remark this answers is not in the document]
koic: Besides the validity of the syntax, https://bugs.ruby-lang.org/issues/18396 should have been considered. I've opened #10363 to fix these.

PR #10363, opened 2022-01-19
[Fix #10359] Fix a false positive and negative for `Style/HashSyntax`

koic:
Fixes #10359.

This PR fixes a false positive and negative for `Style/HashSyntax` when using hash value omission.

It resolves the lack of consideration for https://bugs.ruby-lang.org/issues/18396 and false positives for modifier forms.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #10368, opened 2022-01-20
Fix a false positive for `Style/HashSyntax`

koic:
Follow up to https://github.com/rubocop/rubocop/pull/10363#issuecomment-1016233254.

This PR fixes a false positive for `Style/HashSyntax` when using hash value omission.
It does not have a changelog entry because a regression of unreleased #10363.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
+++ b/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
@@ -65,7 +65,9 @@ module RuboCop
       end
 
       def without_parentheses_call_expr_follows?(ancestor)
-        return false unless (right_sibling = ancestor.right_sibling)
+        right_sibling = ancestor.right_sibling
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

tycooon:
Nice job! There still are some cases where rubocop doesn't detect issues that were detected in 1.24. Here are examples that I found:

```ruby
build :user, token: token
hello # Removing this line fixes the issue
```

```ruby
def attrs
  { **super, token: token }
end

hello # Removing this line fixes the issue
```

```ruby
User[token: token]
hello # Removing this line fixes the issue
```

I guess that's all the same issue.

koic:
The first case is an intended change to prevent https://bugs.ruby-lang.org/issues/18396 issue.
The second and third cases are enclosed in `{}` or `[]`, so the issue does not occur. I've opened #10371 to resolve the second and third cases. Thank you again for your feedback!

PR #10370, opened 2022-01-20
[Fix #10366] Fix a false positive for `Style/MethodCallWithArgsParentheses`

### LINKED ISSUE #10366: Fix Style/MethodCallWithArgsParentheses issue with guard statement
## Expected behavior

I ran into an issue where, if using a guard statement, RuboCop will attempt to auto-correct the code which then leads to a Ruby 3.1.0 syntax error.

## Actual behavior

Here's the error I see:

``` 
/Users/bkuhlmann/Engineering/Misc/snippet.rb:18:18: C: [Correctable] Style/MethodCallWithArgsParentheses: Omit parentheses for method calls with arguments. (https://rubystyle.guide#method-invocation-parens)
point = Point.new(x: 1, y:) if toggle
```

You can use the following code snippet to recreate the error:

``` ruby
# Save as `snippet.rb` and run as `rubocop  --only Style/MethodCallWithArgsParentheses snippet.rb`

class Point
  attr_reader :x, :y

  def initialize x:, y:
    @x, @y = x, y
  end
end

y = 2
toggle = true

point = Point.new(x: 1, y:) if toggle
```

If you were to change the last line to the following (no parenthesis -- as auto-corrected by RuboCop):

``` ruby
point = Point.new x: 1, y: if toggle
```

You'd then end up with the following syntax error:

``` 
/Users/bkuhlmann/Engineering/Misc/snippet.rb:17: syntax error, unexpected end-of-input
... = Point.new x: 1, y: if toggle
```

This almost feels as if the Ruby parser should be fine with the guard statement in this case but am not sure. Anyway, allowing Style/MethodCallWithArgsParentheses to ignore parenthesis in this case would make it so auto-correct works without causing a syntax error in Ruby.

## RuboCop version

```
1.25.0 (using Parser 3.1.0.0, rubocop-ast 1.15.1, running on ruby 3.1.0 arm64-darwin21.2.0)
  - rubocop-performance 1.13.2
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.7.0
```

koic:
Fixes #10366.

This PR fixes a false positive for `Style/MethodCallWithArgsParentheses` when setting `EnforcedStyle: omit_parentheses` and using hash value omission with modifier from.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positive_for_style_method_call_with_args_parentheses.md
+++ b/changelog/fix_false_positive_for_style_method_call_with_args_parentheses.md
@@ -1,0 +1,1 @@
+* [#10366](https://github.com/rubocop/rubocop/issues/10366): Fix a false positive for `Style/MethodCallWithArgsParentheses` when setting `EnforcedStyle: omit_parentheses` and using hash value omission with modifier from. ([@koic][])
--- a/lib/rubocop/cop/style/method_call_with_args_parentheses/omit_parentheses.rb
+++ b/lib/rubocop/cop/style/method_call_with_args_parentheses/omit_parentheses.rb
@@ -48,15 +48,21 @@ module RuboCop
             node.each_ancestor(:def, :defs).any?(&:endless?) && node.arguments.any?
           end
 
-          # Require hash value omission be enclosed in parentheses to prevent the following issue:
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #10371, opened 2022-01-20
Fix a false negative for `Style/HashSyntax`

koic:
Follow up to https://github.com/rubocop/rubocop/pull/10368#issuecomment-1017488540.

This PR fixes a false negative for `Style/HashSyntax` when `Hash[foo: foo]` or `{foo: foo}` is followed by a next expression.
https://bugs.ruby-lang.org/issues/18396 issue does not occur when hash syntax is enclosed in `[]`, `{}`, or `()`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_negative_for_style_hash_syntax.md
+++ b/changelog/fix_false_negative_for_style_hash_syntax.md
@@ -1,0 +1,1 @@
+* [#10371](https://github.com/rubocop/rubocop/pull/10371): Fix a false negative for `Style/HashSyntax` when `Hash[foo: foo]` or `{foo: foo}` is followed by a next expression. ([@koic][])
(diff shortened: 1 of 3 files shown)


tycooon:
I tested this version and I get errors with new `case` syntax:

```ruby
case foo
in a: { a: 1 } # Generates `An error occurred while Style/HashSyntax cop was inspecting` error.
  1
else
  2
end
```

tycooon:
Here is the error btw:
```
undefined method `braces?' for s(:hash_pattern,
  s(:pair,
    s(:sym, :a),
    s(:int, 1))):RuboCop::AST::Node

        !node.parent.braces? && !use_element_of_hash_literal_as_receiver?(ancestor, node.parent) &&
                    ^^^^^^^^
```

koic:
Good catch! I was unaware of that case. The hash pattern matching can be ignored because its different semantic. I've updated this PR.

tycooon:
Seems to work perfectly now 👍

bbatsov:
Looks good! Thanks!

PR #3427, opened 2016-08-18
Restrict inheriting from other subclasses of Exception

metcalf:
The Lint/InheritException cop currently prevents inheriting from Exception and suggests inheriting from RuntimeError or StandardError instead. This updates it to restrict inheriting from the other subclasses of Exception in the standard library that aren't themselves subclasses of StandardError. I did not add a separate test for this behavior since it would be largely duplicative of the existing test logic.

PR #10406, opened 2022-02-09
Fix a false positive for `Lint/InheritException`

koic:
Fixes https://github.com/testdouble/standard/issues/388 and reverts https://github.com/rubocop/rubocop/pull/3427.

This PR prevents the following false positive and auto-correction.

```diff
 class TestClass
-  FatalError = Class.new(Interrupt)
+  FatalError = Class.new(RuntimeError)
 end
```

Below is an inheritance tree for the exception classes.

```
        ---------------
        | BasicObject |
        ---------------
               ^
               |
        ---------------
        |   Kernel    |
        ---------------
               ^
               |
        ---------------
        |   Object    |
        ---------------
               ^
               |
        ---------------
        |  Exception  |
        ---------------
               ^
               |
      |------------------|
---------------  -----------------
|StandardError|  |SignalException|
---------------  -----------------
      ^                  ^
      |                  |
---------------  -----------------
|RuntimeError |  |   Interrupt   |
---------------  -----------------
```

As the inheritance tree shows, `Interrupt` is not `is_a?` in either `StandardError` or `RuntimeError`.

```ruby
RuntimeError.new.is_a?(StandardError) #=> true
Interrupt.new.is_a?(StandardError)    #=> false
Interrupt.new.is_a?(RuntimeError)     #=> false
```

This goes against The Liskov Substitution Principle. `Interup` should not be replaced with `StandardError` or its subclasses.
Also, according to The Open/Closed Principle, `Interrup` class should be open to inheritance.

The same is true for `SystemStackError`, `NoMemoryError`, `SecurityError`, `NotImplementedError`, `LoadError`, `SyntaxError`, `ScriptError`, `SignalException`, and `SystemExit` classes.

This PR allows for exception handling according to Object-oriented principles and Ruby exception handling mechanism.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_lint_inherit_exception.md
+++ b/changelog/fix_a_false_positive_for_lint_inherit_exception.md
@@ -1,0 +1,1 @@
+* [#10406](https://github.com/rubocop/rubocop/pull/10406): Fix a false positive for `Lint/InheritException` when inheriting a standard lib exception class that is not a subclass of `StandardError`. ([@koic][])
--- a/lib/rubocop/cop/lint/inherit_exception.rb
+++ b/lib/rubocop/cop/lint/inherit_exception.rb
@@ -3,10 +3,9 @@
 module RuboCop
   module Cop
     module Lint
-      # This cop looks for error classes inheriting from `Exception`
-      # and its standard library subclasses, excluding subclasses of
-      # `StandardError`. It is configurable to suggest using either
-      # `StandardError` (default) or `RuntimeError` instead.
+      # This cop looks for error classes inheriting from `Exception`.
+      # It is configurable to suggest using either `StandardError` (default) or
+      # `RuntimeError` instead.
       #
       # @safety
       #   This cop's autocorrection is unsafe because `rescue` that omit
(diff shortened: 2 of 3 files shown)


bbatsov:
> The same is true for SystemStackError, NoMemoryError, SecurityError, NotImplementedError, LoadError, SyntaxError, ScriptError, SignalException, and SystemExit classes.

I was the impression that those classes should not be extended by application code, as they are internal to Ruby, plus usually they can't be handled in a meaningful way. I think this was also the main point of the special check for them - if you end up using those most likely you're doing something wrong.

koic:
I think this `Lint/InheritException` cop has the role of protecting users from accidentally extending `Exception` class where they should extend `StandardError` or `RuntimeError` classes. This difference between `Exception`, `StandardError`, and `RuntimeError` may be mistaken by new Ruby user. I've seen some misunderstandings, especially among programmers skill transfering from Java to Ruby. This is because Java extends `Exception` class when creating custom exceptions.

On the other hand, I agree that it is rare for these exception classes to be extended at user layer, but these exception classes (e.g. `LoadError`) may be intentionally extended. In other words, it's possible that user intended it, not a mistake. So I think it's better to register an offense only with `Exception`.

bbatsov:
Fair enough, let's simplify this.

PR #10407, opened 2022-02-09
Change `EnforcedStyle` to `standard_error` for `Lint/InheritException`

koic:
This PR change `EnforcedStyle` from `runtime_error` to `standard_error` for `Lint/InheritException`.

I noticed this while investigating #10406.

Below is the inheritance tree for the exception classes.

```
---------------
|  Exception  |
---------------
      ^
      |
---------------
|StandardError| (default for `rescue`)
---------------
      ^
      |
---------------
|RuntimeError | (default for `raise`)
---------------
```

AFAIK, it seems that custom exceptions generally inherit from `StandardError`.
Subclasses of `StandardError` are handled with `rescue` by default.

`raise` creates a `RuntimeError` when the exception class is omitted.
If custom exception inherits from `RuntimeError`, the custom exception and `RuntimeError` will be is-a.
If custom exception inherits from `StandardError`, the custom exception and `RuntimeError` will not be is-a.
In other words, inheriting `StandardError` will not be mistakenly handled as `raise` which omits the exception class.

So, inheriting `StandardError` rather than `RuntimeError` does not include unnecessary inheritance for `rescue` handling.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_enforced_style_for_lint_inherit_exception.md
+++ b/changelog/change_enforced_style_for_lint_inherit_exception.md
@@ -1,0 +1,1 @@
+* [#10407](https://github.com/rubocop/rubocop/pull/10407): Change `EnforcedStyle` from `runtime_error` to `standard_error` for `Lint/InheritException`. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Good catch!

PR #10413, opened 2022-02-13
[Fix #10373] Support top level departments in config

### LINKED ISSUE #10373: Supported Hierarchical Departments

It would be very useful for Rubocop to be able to supported hierarchical/nested departments.

For example, if I have a department MyExtension/Foo and MyExtension/Bar it would be useful to be able to enable/disable MyExtension as a whole,

In particular this comes up with things like [Cookstyle](https://github.com/chef/cookstyle/) which is a tool built on top of rubocop to do Chef correctness linting. They have many departments all of the form Chef/<whatever>. It would be useful to be able to turn all of those on or off.

In my particular case since for standard ruby-correctness, I tend to stick much closer to Rubocop defaults than Cookstyle defaults, but Cookstyle includes a ton of its own defaults for rubocop rules, for my cookstyle runs, I actually want to disable all rules, then enable all Chef/* rules, ala something like this:

```yaml
AllCops:
  DisabledByDefault: true

Chef:
  Enabled: true
```

Unfortunately today I have to list every single department manually and if they add one, I have to add it to my config.

## Additional context

See further discussion of this in https://github.com/rubocop/rubocop/issues/9752

jonas054:
Support for top level departments is added, but there's no support for arbitrary levels of departments at this point. I.e., custom cops can be under various departments, which in turn are placed under a common top level department.

After this change it's possible to, for example, disable all cops in the cookstyle gem by adding to user configuration:

```
Chef:
  Enabled: false
```

Note! Only the `Enabled` parameter is supported. Setting other parameters such as `Severity` on the top level (when there are two department levels) will silently fail. No warning is shown.

-----------------

### Comments

Adding support for `Enabled` was relatively easy. Supporting other parameters on the top level seems harder, and it makes me question how useful this change would be and it it's worth the effort. There's currently only one custom cop project that has its own top level, and that's cookstyle, right? And the problem has been worked around by setting default `Enabled` parameters on the subdepartments there, plus a warning in RuboCop for settings on the top level.


bbatsov:
It's a step in the right direction for sure. Before merging this we'll definitely need some warnings and proper documentation, though. 

> There's currently only one custom cop project that has its own top level, and that's cookstyle, right?

One that I know of. :-)

bbatsov:

PR #10418, opened 2022-02-15
[Fix #9816] Add a ShortCircuit option to Lint/SafeNavigationConsistency

### LINKED ISSUE #9816: Request: Lint/SafeNavigationConsistency short-circuit option for &&
When using `Lint/SafeNavigationConsistency` in an `&&` / `and` experission, I'd like if `variable&.` is only required the first time that variable is used. In terms of boolean logic, any subsequent usages of `variable&.` are redundant because `&&` are short-circuit evaluated.

(Note that for `||` / `or` and other operators it is still needed on each instance of that variable)

I would suggest to make this an option such as `ShortCircuitAndOperator`

**Lint/SafeNavigationConsistency - Desired behavior**

```ruby
@foo&.bar? && @foo.baz?
```

**Lint/SafeNavigationConsistency - Currently enforced behavior**

```ruby
@foo&.bar? && @foo&.baz?
```

#### RuboCop version

```
bundle exec rubocop -V
1.15.0 (using Parser 3.0.1.1, rubocop-ast 1.5.0, running on ruby 2.6.6 x64-mingw32)
  - rubocop-rails 2.10.1
```

rrosenblum:
This closes #9816. Add a ShortCircuit option to Lint/SafeNavigationConsistency. `ShortCircuit` does not support auto-correct. I found a few edge case conditions that could make auto-correct with `ShortCircuit` unsafe.

example:
```ruby
foo + 1 && foo&.even?
```
non-dot method needing safe nav. We don't auto-correct safe nav onto non-dot methods so removing safe nav `foo&.even?` would change the code and make this cop unable to flag it on subsequent passes. I think we can detect this an not auto-correct this use case, however I also discovered that this cop doesn't currently support methods with a block or non-dot methods. 

There's some sizeable work that can be done to this cop. I think it would be beneficial to change the entry point from looking for safe nav to looking for `&&` and `||`. We should be able to more reliable always start at the beginning of a condition rather than starting anywhere and going up the chain. I think this will make the variable scope easier to work with and make auto-correction easier.

### DIFF
--- a/lib/rubocop/cop/lint/safe_navigation_consistency.rb
+++ b/lib/rubocop/cop/lint/safe_navigation_consistency.rb
@@ -7,6 +7,11 @@ module RuboCop
       # call in an `&&` or `||` condition that safe navigation is used for all
       # method calls on that same object.
       #
+      # Options:
+      # `ShortCircuit` (default `false`) - Enforces short circuit syntax for `&&` conditions. Only
+      # the first method call is required to use safe navigation. All other method calls do not
+      # use safe navigation. There is no impact on `||` conditions.
+      #
       # @example
       #   # bad
       #   foo&.bar && foo.baz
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/lint/safe_navigation_consistency.rb
```diff
@@ -26,48 +31,139 @@ module Lint
...
+      #   # good
+      #   foo&.bar || foo&.bar && foo.qux
+      #
       class SafeNavigationConsistency < Base
         include IgnoredNode
         include NilMethods
         extend AutoCorrector
 
         MSG = 'Ensure that safe navigation is used consistently inside of `&&` and `||`.'
+        SHORT_CIRCUIT_MSG = 'Only use safe navigation for the first conidition of `&&`.'
```
bbatsov, line 65: `&&` doesn't have conditions, but rather "operands".

bbatsov:
> There's some sizeable work that can be done to this cop. I think it would be beneficial to change the entry point from looking for safe nav to looking for && and ||. We should be able to more reliable always start at the beginning of a condition rather than starting anywhere and going up the chain. I think this will make the variable scope easier to work with and make auto-correction easier.

Yeah, probably that makes sense. As the proposed change is not something particularly important there's no rush to merge this and you can take your time to make deeper changes if you see fit to do so.

bbatsov:
One more thing - it'd be nice if we explained it more details the rationale for this short-circuit behavior in the docs.

dvandersluis:
@rrosenblum ping :)

rrosenblum:
This work fell on my back burner. I have not made much of any progress on it. It is something that I would like to complete, however I would be perfectly fine if someone else wants to champion it. I will try to dedicate some time to finishing this up

bbatsov:
@rrosenblum No rush.

koic:
The revised #13312 has been merged. This resolves the issue in #9816 by default.

PR #10419, opened 2022-02-15
Add new `Style/FileDirnameLevel` cop

koic:
## Summary

This cop checks for nested `File.dirname`. It replaces repeats with the level argument introduced in Ruby 3.1.

```ruby
# bad
File.dirname(File.dirname(path))

# good
File.dirname(path, 2)
```

See: https://bugs.ruby-lang.org/issues/12194

## Other Information

For `File.expand_path('../..', path)`, there is a partial conflict with `Style/ExpandPathArguments` cop. So this PR does not support it.
https://docs.rubocop.org/rubocop/1.25/cops_style.html#styleexpandpatharguments

-----------------

* [x] Added tests.


lib/rubocop/cop/style/file_dirname_level.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # This cop checks for repeated `File.dirname`.
```
bbatsov, line 6: repeated -> nested

lib/rubocop/cop/style/file_dirname_level.rb
```diff
@@ -0,0 +1,66 @@
...
+      #
+      # @example
+      #
+      #   # bad
+      #   File.dirname(File.dirname(path))
+      #
+      #   # good
+      #   File.dirname(path, 2)
+      #
+      class FileDirnameLevel < Base
```
bbatsov, line 17: I wonder if `NestedFileDirname` is not a better name, but I'm fine with the proposed name.

spec/rubocop/cop/style/file_dirname_level_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Style::FileDirnameLevel, :config do
+  context 'Ruby >= 3.1', :ruby31 do
+    it 'registers and corrects an offense when using `File.dirname(path)` repeated two times' do
```
bbatsov, line 5: nested two times

And so on for the other example descriptions.

lib/rubocop/cop/style/file_dirname_level.rb
koic: Yeah, I was wondering whether to use that name. So, I've updated to `NestedFileDirname` to be consistent with the doc.

bbatsov:
Thanks!

PR #10425, opened 2022-02-17
[Fix #10421] Make `Style/DefWithParentheses` aware of endless method definition

### LINKED ISSUE #10421: Flag offence for empty parentheses in single line method definitions
## Problem

Rubocop allows single line method definitions to use empty parentheses.
Eg.
```rb
def method_1() 'foo'; end
def method_2() = 'bar'
```
Though this behaviour has been part of Rubocop for many years, as far as I can see it is contrary to the style guide which expressly instructs against using parentheses in a method definition when there are no parameters:

> ### Method Definition Parentheses
> Use def with parentheses when there are parameters. Omit the parentheses when the method doesn’t accept any parameters.

Link: [Method Definition Parentheses](https://rubystyle.guide/#method-parens)

I have been unable to find an exception to this rule for single line methods anywhere in the guide (though I might be missing it).

## Solution
`Style/DefWithParentheses` flags offences for single line method definitions with parentheses which do not accept parameters

## Alternatives

I have been unable to discover the rationale behind originally allowing empty parentheses in [this commit](https://github.com/rubocop/rubocop/pull/8/commits/c973550a9a9b90ca2347344e3c0b58d1e36e230a). If it is behaviour we want to retain, perhaps there should be a mention of it in the style guide (assuming there isn't one already which I have been unable to find).

koic:
Fixes #10421.

This PR makes `Style/DefWithParentheses` aware of endless method definition.

`def foo() do_something end` will continue to allow because `def foo do_something end` is a syntax error.
On the other hand, be aware that the endless method definition's parentheses can be omitted.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/def_with_parentheses.rb
+++ b/lib/rubocop/cop/style/def_with_parentheses.rb
@@ -11,27 +11,33 @@ module RuboCop
       #
       #   # bad
       #   def foo()
-      #     # does a thing
+      #     do_something
       #   end
       #
       #   # good
       #   def foo
-      #     # does a thing
+      #     do_something
       #   end
       #
-      #   # also good
-      #   def foo() does_a_thing end
+      #   # bad
+      #   def foo() = do_something
+      #
+      #   # good
+      #   def foo = do_something
+      #
+      #   # good (without parentheses it's a syntax error)
+      #   def foo() do_something end
       #
       # @example
       #
       #   # bad
       #   def Baz.foo()
-      #     # does a thing
+      #     do_something
       #   end
       #
       #   # good
       #   def Baz.foo
-      #     # does a thing
+      #     do_something
       #   end
       class DefWithParentheses < Base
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/def_with_parentheses.rb
```diff
@@ -11,35 +11,41 @@ module Style
...
       #
-      #   # also good
-      #   def foo() does_a_thing end
+      #   # bad
+      #   def foo() = do_something
+      #
+      #   # good
+      #   def foo = do_something
+      #
+      #   # good
```
jonas054, line 28: Should we add an explanation here? E.g., `# good (without parentheses it's a syntax error)`.

koic: Good point! I've updated it!

bbatsov:
Thanks!

PR #10428, opened 2022-02-19
[Fix #10345] Update docs for MultilineOperationIndentation

### LINKED ISSUE #10345: MultilineOperationIndentation not working properly on implicit return
## disclamer

I'm not sure if it's a bug or not.

--------

Per the [docs](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/MultilineOperationIndentation):
>The `aligned` style checks that operators are aligned if they are part of an `if` or `while` condition, a `return` statement, etc. In other contexts, the second operand should be indented regardless of enforced style.

That's not always true, at least for explicit returns

Let's have the following in config file:

```yml
# just for the sake of this example
Style/RedundantReturn:
  Enabled: false

Layout/MultilineOperationIndentation:
  EnforcedStyle: aligned
```

And the following code
```rb
def my_test
  my_var >>
  my_function >>
  my_other_function
end

def another_test
  return my_var >>
         my_function >>
         my_other_function
end
```

## Expected behavior
The 2 examples should (?) be ok

## Actual behavior

In reality, the first one is failing, as (my guess) it's not "detecting" that this really is a return statement.

## RuboCop version

```
1.24.0 (using Parser 3.0.3.2, rubocop-ast 1.15.0, running on ruby 2.7.5 x86_64-darwin20)
  - rubocop-faker 1.1.0
  - rubocop-performance 1.12.0
  - rubocop-rails 2.12.4
```

jonas054:
It was unclear that the special rule for `return` statements (and other cases where an operation follows a keyword) did not include implicit returns.

Also, the examples given for the `aligned` style did not reflect was was said about the second operand always being indented except after a keyword.

### DIFF
--- a/lib/rubocop/cop/layout/multiline_operation_indentation.rb
+++ b/lib/rubocop/cop/layout/multiline_operation_indentation.rb
@@ -3,27 +3,26 @@
 module RuboCop
   module Cop
     module Layout
-      # This cop checks the indentation of the right hand side operand in
-      # binary operations that span more than one line.
+      # This cop checks the indentation of the right hand side operand in binary operations that
+      # span more than one line.
       #
-      # The `aligned` style checks that operators are aligned if they are part
-      # of an `if` or `while` condition, a `return` statement, etc. In other
-      # contexts, the second operand should be indented regardless of enforced
-      # style.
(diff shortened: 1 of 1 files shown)


koic:
Thanks!

PR #10441, opened 2022-03-03
Add `Security/CompoundHash` cop

sambostock:
# Context

Hash codes for Ruby objects are a potential security issue. Bad hash codes may make us vulnerable to denial-of-service-attacks, where users are able to craft objects that trigger hash collisions. Hash collisions may cause linear instead of constant-time behaviour, or they may cause thrashing of caches. Let Ruby's implementation manage the construction of hash codes from multiple objects using `Array#hash`, via `[...].hash`, rather than rolling your own hash.

# Summary

This cop checks implementations of the `hash` method which combine value using the XOR operator instead of delegating to `Array#hash`.
    
`Array#hash` is

- faster
- clearer
- standardized
    
Ruby has an existing way to combine multiple hash values, so users should not implement their own.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/security/compound_hash.rb
+++ b/lib/rubocop/cop/security/compound_hash.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Security
+      # This cop checks for implementations of the `hash` method which combine
+      # values using custom logic instead of delegating to `Array#hash`.
+      #
+      # Manually combining hashes is error prone and hard to follow, especially
+      # when there are many values. Poor implementations may also introduce
+      # performance or security concerns if they are prone to collisions.
+      # Delegating to `Array#hash` is clearer, faster, and safer.
+      #
+      # @safety
+      #   This cop may be unsafe if the application logic depends on the hash
+      #   value, however this is inadvisable anyway.
+      #
+      # @example
+      #
+      #   # bad
+      #   def hash
+      #     @foo ^ @bar
+      #   end
+      #
+      #   # good
+      #   def hash
+      #     [@foo, @bar].hash
+      #   end
+      class CompoundHash < Base
+        COMBINATOR_IN_HASH_MSG = 'Use `[...].hash` instead of combining hash values manually.'
+        MONUPLE_HASH_MSG =
+          'Delegate hash directly without wrapping in an array when only using a single value'
+        REDUNDANT_HASH_MSG = 'Calling .hash on elements of a hashed array is redundant'
+
+        # @!method hash_method_definition?(node)
+        def_node_matcher :hash_method_definition?, <<~PATTERN
+          {#static_hash_method_definition? | #dynamic_hash_method_definition?}
+        PATTERN
+
+        # @!method dynamic_hash_method_definition?(node)
+        def_node_matcher :dynamic_hash_method_definition?, <<~PATTERN
+          (block
+            (send _ {:define_method | :define_singleton_method}
+              (sym :hash))
+            (args)
+            _)
+        PATTERN
+
+        # @!method static_hash_method_definition?(node)
+        def_node_matcher :static_hash_method_definition?, <<~PATTERN
+          ({def | defs _} :hash
+            (args)
+            _)
+        PATTERN
+
+        # @!method bad_hash_combinator?(node)
+        def_node_matcher :bad_hash_combinator?, <<~PATTERN
+          ({send | op-asgn} _ {:^ | :+ | :* | :|} _)
+        PATTERN
+
+        # @!method monuple_hash?(node)
+        def_node_matcher :monuple_hash?, <<~PATTERN
+          (send (array _) :hash)
+        PATTERN
+
+        # @!method redundant_hash?(node)
+        def_node_matcher :redundant_hash?, <<~PATTERN
+          (
+            ^^(send array ... :hash)
+            _ :hash
+          )
+        PATTERN
+
+        def contained_in_hash_method?(node, &block)
+          node.each_ancestor.any? do |ancestor|
+            hash_method_definition?(ancestor, &block)
+          end
+        end
+
+        def outer_bad_hash_combinator?(node)
+          bad_hash_combinator?(node) do
+            yield true if node.each_ancestor.none? { |ancestor| bad_hash_combinator?(ancestor) }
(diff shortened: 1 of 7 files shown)


bbatsov:
I like this proposal.

sambostock (commented):
Based on some searches to find implementations of `hash` "in the wild", I have added various examples, and update the cop to target the use of various operators that wouldn't appear in a basic implementation delegating to `Array#hash`.

However, I am not an expert in hashing, so I am looking for feedback on if all of these cases are appropriate to flag. I suspect our stance should be

> Always use `Array#hash`, unless you have a very good reason not to

and leave it up to implementers to decide if their use case is special.

lib/rubocop/cop/security/compound_hash.rb
```diff
@@ -0,0 +1,105 @@
...
+          node.each_ancestor.any? do |ancestor|
+            hash_method_definition?(ancestor, &block)
+          end
+        end
+
+        def outer_bad_hash_combinator?(node)
+          bad_hash_combinator?(node) do
+            yield true if node.each_ancestor.none? { |ancestor| bad_hash_combinator?(ancestor) }
+          end
+        end
```
sambostock, line 84: This ensures that **we do not generate redundant offenses** like this:

```ruby
def hash)
  1.hash ^ 2.hash ^ 3.hash
# ^^^^^^^^^^^^^^^ Use `[...].hash` instead of combining hash values manually.
# ^^^^^^^^^^^^^^^^^^^^^^^^ Use `[...].hash` instead of combining hash values manually.
end
```

lib/rubocop/cop/security/compound_hash.rb
```diff
@@ -0,0 +1,105 @@
...
+            ^^(send array ... :hash)
+            _ :hash
+          )
+        PATTERN
+
+        def contained_in_hash_method?(node, &block)
+          node.each_ancestor.any? do |ancestor|
+            hash_method_definition?(ancestor, &block)
+          end
+        end
```
sambostock, line 78: I originally looked for `def` and `defs`, but given I wanted to also detect `define_method` and `define_singleton_method`, which would have needed to use `on_send` anyway, it was simpler to go upwards from the potentially "bad" code to check if it was in a `hash` implementation, than it was to start from a `hash` method definition, and look for "bad" code.

lib/rubocop/cop/security/compound_hash.rb
```diff
@@ -0,0 +1,105 @@
...
+
+        # @!method monuple_hash?(node)
+        def_node_matcher :monuple_hash?, <<~PATTERN
+          (send (array _) :hash)
+        PATTERN
+
+        # @!method redundant_hash?(node)
+        def_node_matcher :redundant_hash?, <<~PATTERN
+          (
+            ^^(send array ... :hash)
```
sambostock, line 69: The [documentation for matching a node's parent](https://docs.rubocop.org/rubocop-ast/node_pattern.html#for-parent) could use expanding, and more examples for those not intimately familiar with the AST. It took me a very long time, lots of refactoring, and lots of experimenting to figure out how to properly match what I wanted.

I would have been lost if not for https://nodepattern.herokuapp.com  🙇

bbatsov: Admittedly our documentation there is not great indeed, so PRs are always welcome!

spec/rubocop/cop/security/compound_hash_spec.rb
```diff
@@ -0,0 +1,203 @@
...
+        [red, blue, green, alpha].hash ^ self.class.hash
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `[...].hash` instead of combining hash values manually.
+      end
+    RUBY
+  end
+
+  it 'registers an offense when using XOR involving super' do
+    expect_offense(<<~RUBY)
+      def hash
+        foo.hash ^ super ^ bar.hash
```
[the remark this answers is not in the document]
bbatsov: Not sure myself to be honest. I guess we can leave it as is and see what kind of feedback we'll get.

bbatsov:
Great PR, I loved all the comments you've inserted here and there! 🙇 

Might be a good idea to add some section about this in the Ruby style guide as well at some point.

sambostock:
> Might be a good idea to add some section about this in the Ruby style guide as well at some point. 

@chrisseaton and I are planning on submitting improvements to the Ruby hashing docs to recommend the use of `Array#hash`, so we could probably reuse some of that work or link to it for details.

spec/rubocop/cop/security/compound_hash_spec.rb
```diff
@@ -0,0 +1,203 @@
...
+      end
+    RUBY
+  end
+
+  it 'registers an offence when using bitshift and OR' do
+    expect_offense(<<~RUBY)
+      def hash
+        ([@addr, @mask_addr, @zone_id].hash << 1) | (ipv4? ? 0 : 1)
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `[...].hash` instead of combining hash values manually.
+      end
```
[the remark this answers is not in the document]
chrisseaton: Not preferable I'm sure, even without benchmarking. A conditional and an or vs just another value in the array literal.

spec/rubocop/cop/security/compound_hash_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+    expect_offense(<<~RUBY)
+      def hash
+        @hash ||= begin
+          hash_ = [@factory, geometry_type].hash
+          (0...num_geometries).inject(hash_) do |h_, i_|
+            (1664525 * h_ + geometry_n(i_).hash).hash
+             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `[...].hash` instead of combining hash values manually.
+          end
+        end
+      end
```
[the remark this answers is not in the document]
chrisseaton: Should we document something about when it makes sense to memoize and when not?

sambostock:
Just to close the loop, the [documentation PR in Ruby/ruby](https://github.com/ruby/ruby/pull/5805) has been merged.

PR #10459, opened 2022-03-18
[Fix #10374] Fix a false positive for `Layout/LineLength` when long URIs in yardoc comments to have titles

### LINKED ISSUE #10374: Layout/LineLength AllowURI does not allow long URLs in yardoc comments to have titles
## Expected behavior

YARD comments with [linked URLs](https://rubydoc.info/gems/yard/file/docs/GettingStarted.md#linking-urls-http) of the form `{<url> <title>}` should not be offenses against `Layout/LineLength` when `AllowURI` is true.

## Actual behavior

URLs of the form `{<url>}` are allowed, but including a space-separated `<title>` causes the check to fail.

Breaking the line, as:

```rb
{<url>
<title}
```

allows the test to pass, but breaks doc generation.

Create a Ruby file `mms_id.rb` containing the following code:

```ruby
# frozen_string_literal: true

# {RecordId} subclass representing an Alma MMS ID. Note that only
# bibliographic records (prefix `99`) are supported.
#
# See {https://knowledge.exlibrisgroup.com/Alma/Product_Documentation/010Alma_Online_Help_(English)/Metadata_Management/005Introduction_to_Metadata_Management/020Record_Numbers Record Numbers}
# in the Alma documentation.
class MMSID; end
```

Run RuboCop with default configuration:

```none
$ rubocop --debug mms_id.rb
For /private/tmp: Default configuration from /Users/david/.rvm/gems/ruby-3.0.2/gems/rubocop-1.25.0/config/default.yml
Use parallel by default.
Running parallel inspection
Inspecting 1 file
Scanning /private/tmp/mms_id.rb
Loading cache from /Users/david/.cache/rubocop_cache/7ccfad009a5b251caa148252ae8bbc361b628b36/6d7a3b621ca1730e04accd938619e4bdab66cfb1/7cd4b602c9667aa65e1b35069c3ea22b4b5228fa
C

Offenses:

mms_id.rb:6:177: C: Layout/LineLength: Line is too long. [192/120]
# See {https://knowledge.exlibrisgroup.com/Alma/Product_Documentation/010Alma_Online_Help_(English)/Metadata_Management/005Introduction_to_Metadata_Management/020Record_Numbers Record Numbers}
                                                                                                                                                                                ^^^^^^^^^^^^^^^^

Finished in 0.11044700001366436 seconds
```

## RuboCop version

```
1.25.0 (using Parser 3.1.0.0, rubocop-ast 1.15.1, running on ruby 3.0.2 arm64-darwin21)
```

### Use Markdown syntax in YARD

When passing `-m markdown` as an argument to `yard` (or in `.yardopts`), it's possible to write URLs in the form `[<title>](<url>)`. This passes `Layout/LineLength`, so long as the start of the URL is within the length limit, and so long as the (parenthesized) URL is the last text on the line.

### Using `rubocop:disable`, awkwardly

It's possible to wrap documentation comments in `rubocop:disable` / `rubocop:enable` without YARD picking them up and including them in the generated docs, but it's awkward:

```ruby
# rubocop:disable Layout/LineLength

# Class documentation goes here, after newline,
# but if we don't want the `enable` comment to be
# included in the docs, we have to move it past
# the first line of the code we're documenting
# 
# See {http://some-really-really-really-long-url.example.org/really-really-really-really-really/long/long/long/thing/with/a title}
class Foo
  # rubocop:enable Layout/LineLength

  # ...
end
```

## Discussion

I think I understand the intent behind the [`find_excessive_uri_range`](https://github.com/rubocop/rubocop/blob/v1.25.0/lib/rubocop/cop/mixin/line_length_help.rb#L34-L52) and [`allowed_uri_position?`](https://github.com/rubocop/rubocop/blob/v1.25.0/lib/rubocop/cop/mixin/line_length_help.rb#L26-L28) calculations in `Rubocop::Cop::LineLengthHelp`, but I wonder if it's not trying to be too clever. I suspect a cop that simply ignored the first URL in a line of code when calculating the line length would be simpler and would still catch the vast majority of cases.

ydah:
Fixes https://github.com/rubocop/rubocop/issues/10374

This PR fixes a false positive for `Layout/LineLength` when when long URIs in yardoc comments to have titles.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_layout_line_length.md
+++ b/changelog/fix_a_false_positive_for_layout_line_length.md
@@ -1,0 +1,1 @@
+* [#10459](https://github.com/rubocop/rubocop/pull/10459): Fix a false positive for `Layout/LineLength` when long URIs in yardoc comments to have titles. ([@ydah][])
--- a/lib/rubocop/cop/mixin/line_length_help.rb
+++ b/lib/rubocop/cop/mixin/line_length_help.rb
@@ -39,12 +39,7 @@ module RuboCop
           pos + indentation_difference(line)
         end
 
-        # Extend the end position until the start of the next word, if any.
-        # This allows for URIs that are wrapped in quotes or parens to be handled properly
-        # while not allowing additional words to be added after the URL.
-        if (match = line[end_position..line_length(line)]&.match(/^\S+(?=\s|$)/))
-          end_position += match.offset(0).last
-        end
+        end_position = extend_uri_end_position(line, end_position)
 
         return nil if begin_position < max_line_length && end_position < max_line_length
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Looks good. Thanks!

PR #10467, opened 2022-03-23
[Fix #10438] Add new `Style/RedundantInitialize` cop to check for unnecessary `initialize` methods

### LINKED ISSUE #10438: Suggested cop: Warn of unnecessary `initialize` methods
Sometimes a refactor can lead to a situation such as:

```ruby
class Foo < Bar
  def initialize(a, b)
    super
  end
end
```

dvandersluis:
Adds `Lint/RedundantInitialize` to look for `initialize` methods that are not useful.

The following are considered redundant:
* Methods that only call `super`, passing the same argument list.
* Methods that have no body, or only have a comment.

The following are not considered redundant:
* Methods that only call `super`, but with a different argument list.
* Methods that take a `*`, `**` or `...` in the argument list (as they can change the arguments required in a superclass)
* Methods that set default values to arguments (as they can change behaviour from the superclass).

This cop **does not** check for properties of the containing class (such as inheritence or module inclusion) for simplicity. To avoid false positives, it assumes that any initialize class _may_ be redefining a superclass's method.

Closes #10438.

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/redundant_initialize.rb
```diff
@@ -0,0 +1,115 @@
...
+      # an argument that accepts multiple values (`restarg`, `kwrestarg`, etc.) it
+      # will not register an offense, because it allows the initializer to take a different
+      # number of arguments as its superclass potentially does.
+      #
+      # NOTE: If an initializer argument has a default value, RuboCop assumes it
+      # to *not* be redundant.
+      #
+      # @example
+      #   # bad
+      #   def initialize
```
andyw8, line 19: I'm wondering if there is a situation where someone would intentionally define an empty `initialize` to prevent the parent's `initialize` being called (even though this is probably a terrible approach!).

andyw8: Maybe it should only flag an empty initialize if there's no superclass?

dvandersluis: @bbatsov any thoughts? Are you concerned about this edge case?

koic:
`Lint` is a department that represents a very strong warning level. This would be the cop belonging to `Style`.

lib/rubocop/cop/lint/redundant_initialize.rb
vlad-pisanov: This seems to spill over into the domain of the [Lint/MissingSuper](https://docs.rubocop.org/rubocop/cops_lint.html#lintmissingsuper) cop. I can imagine (rare) but legit use cases for omitting `super`, e.g. for abstract base classes:

```ruby
class Parent
  def initialize
    raise NotImplementedError, "I'm abstract!"
  end
end

class Child < Parent
  def initialize
     # we deliberately don't call super here
  end
end
```

bbatsov: @dvandersluis I'm not very concerned about this. I'd mention this case in the cop description, though. We can adjust this based on the feedback we get from our users once this is out.

bbatsov:
Yeah, I agree that's probably more of a Style cop.

dvandersluis:
Updated to Style namespace, added documentation note about empty initializers.

lib/rubocop/cop/lint/redundant_initialize.rb
dvandersluis: Updated as per @bbatsov.

bbatsov:
Thanks!

PR #10482, opened 2022-03-26
Changed default percent regex delimiters to singlequotes

Epigene:
Delimiters should be symbols that are unlikely (ideally, guaranteed!) to appear in the body. Curly braces have a meaning in regex, as do most other symbols, unfortunately, with the prominent exceptions of single and double quotes, so let's use either or both as delimiters for regexp literals!


dvandersluis:
Hi @Epigene you can configure this for your own .rubocop.yml if that's the style you want, but this isn't a change we're going to make for all of RuboCop.

PR #10483, opened 2022-03-26
Fix #10311 Layout/RedundantLineBreak inside block

### LINKED ISSUE #10311: Layout/RedundantLineBreak misses case inside block
## Expected behavior

Given the below code, I would expect the `Layout/RedundantLineBreak` rule to report that the line inside the block should not be wrapped. **Note:** our line length is set to 84 and `InspectBlocks` is set to `false`.

```rb
some_array.map do |something|
  my_slightly_long_but_not_that_long_method(
    something,
  )
end
```

The code should be enforced to not wrap:
```rb
some_array.map do |something|
  my_slightly_long_but_not_that_long_method(something)
end
```

## Actual behavior

Rubocop does not report an error. If I add *another* line to the block it does report an issue:

```rb
some_array.map do |something|
  foo
  my_slightly_long_but_not_that_long_method(
    something,
  )
end

# C: [Correctable] Layout/RedundantLineBreak: Redundant line break detected.                                                                                                
#  my_slightly_long_but_not_that_long_method( ...
#  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## RuboCop version

```
$ rubocop -V
1.23.0 (using Parser 3.0.3.2, rubocop-ast 1.15.0, running on ruby 2.7.5 x86_64-linux)
  - rubocop-performance 1.12.0
  - rubocop-rails 2.12.4
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.6.0
```

jonas054:
When `Layout/RedundantLineBreak` inspects a send node, it starts by expanding the node to inspect. It does this by traversing the AST upwards as long as the parent is a send node of a block node.
    
The problem was that if the inspected node is a single node inside a `do`..`end` block, we shouldn't move to the parent block and inspect it for redundant line breaks. It's only send nodes before the `do`..`end` that we want to expand in order to inspect "the whole expression". So we should check what kind of block/send relationship it is before moving upwards in the AST.

### DIFF
--- a/changelog/fix_RedundantLineBreak_inside_block.md
+++ b/changelog/fix_RedundantLineBreak_inside_block.md
@@ -1,0 +1,1 @@
+* [#10311](https://github.com/rubocop/rubocop/issues/10311): Fix false negative inside `do`..`end` for `Layout/RedundantLineBreak`. ([@jonas054][])
--- a/lib/rubocop/cop/bundler/duplicated_gem.rb
+++ b/lib/rubocop/cop/bundler/duplicated_gem.rb
@@ -46,11 +46,7 @@ module RuboCop
 
           duplicated_gem_nodes.each do |nodes|
             nodes[1..-1].each do |node|
-              register_offense(
-                node,
-                node.first_argument.to_a.first,
-                nodes.first.first_line
-              )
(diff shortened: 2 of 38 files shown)


bbatsov:
Nice catch!

PR #10485, opened 2022-03-27
[Fix #10310] Add InspectBlocksOnAllLevels config param

### LINKED ISSUE #10310: limit Layout/RedundantLineBreak block nesting

Not really a problem, just a code smell.

I really like the `InspectBlocks` setting for the `Layout/RedundantLineBreak` rule, but I would like to be able to limit the number of nested blocks that the linter reports. For example, if there are blocks inside of blocks, then I would want the outer blocks to be wrapped:

```ruby
# bad
respond_to { |format| format.json { render(json: feedbacks) } }

# bad
respond_to do |format|
  format.json do
    render(json: feedbacks)
  end
end

# good
respond_to do |format|
  format.json { render(json: feedbacks) }
end
```

Well, for now either accepting slightly more confusing single-line blocks, or keeping the rule disabled and having a lot of blocks that could be reasonably wrapped.

jonas054:
Add a configuration parameter to `Layout/RedundantLineBreak` to signal whether blocks on any level (depth) should be inspected. If `false`, only the innermost level of blocks within blocks will be inspected. If `true`, all levels will be inspected.

To achieve backwards compatibility, the default value for the new parameter is set to `true`.

### DIFF
--- a/changelog/change_InspectBlocksOnAllLevels_param.md
+++ b/changelog/change_InspectBlocksOnAllLevels_param.md
@@ -1,0 +1,1 @@
+* [#10310](https://github.com/rubocop/rubocop/issues/10310): Add `InspectBlocksOnAllLevels` parameter in `Layout/RedundantLineBreak`. ([@jonas054][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -1166,6 +1166,7 @@ Layout/RedundantLineBreak:
                  on a single line.
   Enabled: false
   InspectBlocks: false
+  InspectBlocksOnAllLevels: true
   VersionAdded: '1.13'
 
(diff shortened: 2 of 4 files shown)


bbatsov:
I get the idea behind this change, but I can't imagine who would ever want to set something like this. Essentially we're adding a config option to keep supporting a bug, which sets a dangerous precedent. I'm against going in this direction, but as usual I'm open to hear what the other @rubocop/rubocop-core members think as well.

koic:
I agree with @bbatsov opinion. I wonder whether it's a general style that should be enforced in exchange for complex configuration.

jonas054:
Fair enough. Added complexity for minimal gain.

No problem for me to close this one.

Issue #6228, opened 2018-08-29
Warning: "...is a symlink, which is not allowed"


chortels:
**Is your feature request related to a problem? Please describe.**

Hi, 

I'm frustrated when launching rubocop from my CI tool Jenkins. When executing `/bin/rubocop --format=json --out=rubocop-result.json`, Jenkins returns the following message: 

**Warning: <symlink_in_my_computer> is a symlink, which is not allowed.**

And this makes the execution finish with an error. Is there any way to solve this problem?

Regards

mikegee:
That warning is almost certainly coming from the `ResultCache` code here: https://github.com/rubocop-hq/rubocop/blob/d714e247f77457a593792d2ed50483622d3c3686/lib/rubocop/result_cache.rb#L126

I'm not too familiar with that code, but I skimmed it just now. Do you have a `rubocop_cache` directory committed into your project? If you do, try removing that from version control.

chortels:
Hi @mikegee . Thanks for your answer. This .cache/rubocop_cache directory is not committed in my project, but generated when executing bundle install. Is there any way to prevent this directory for being created?

Regards.

mikegee:
You can turn off Rubocop's cache at the command line with `--cache false` or with the `AllCops: UseCache` configuration setting.

koic:
Hi @chortels.
Did this issue be solved with the above `--cache false` option or `AllCops: UseCache` configuration?

PR #10496, opened 2022-03-31
Fix symlink warning always triggered on FreeBSD

gearnode:
This is a proposal not to have the symlink warning on the FreeBSD system.

ref: #6228

### DIFF
--- a/lib/rubocop/result_cache.rb
+++ b/lib/rubocop/result_cache.rb
@@ -73,7 +73,15 @@ module RuboCop
                  # access.
                  File.join(ENV['XDG_CACHE_HOME'], Process.uid.to_s)
                else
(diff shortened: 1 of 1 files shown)


bbatsov:
I don't quite understand why this is a problem for that env variable, but not for the others which are also some paths. I'm also wondering how to make sure this fix would stick, given that's an os-specific problem with no unit tests.

bbatsov:
> On FreeBSD the /home path is a symbolic link to /usr/home and the $HOME environment variable returns the /home path.

I guess you can just add this as a comment around the code in question and we can call it a day.

gearnode:
@bbatsov Patch updated with a comment to explain why we resolve the absolute path.

bbatsov:
Thanks!

PR #10500, opened 2022-04-03
Add new `Lint/RefinementImportMethods` cop

koic:
Follow up to https://github.com/ruby/ruby/commit/6606597.

This cop supports Ruby 3.1's `Refinement#import_methods` and checks if `include` or `prepend` is called in `refine` block.
These methods are deprecated and should be replaced with `import_methods`.

## safety

This cop's autocorrection is unsafe because `include M` will affect the included class if any changes are made to module `M`.
On the other hand, `import_methods M` uses a snapshot of method definitions, thus it will not be affected if module `M` changes.

## example

```ruby
# bad
refine Foo do
  include Bar
end

# bad
refine Foo do
  prepend Bar
end

# good
refine Foo do
 import_methods Bar
end
```

-----------------

* [x] Added tests.


bbatsov:
I think we should mention in the cop's description and the offense message that those methods were deprecated in Ruby 3.1, so it's a bit clearer to users why the cop exists.

koic:
That makes sense! I've updated this PR.

bbatsov:
Thanks!

PR #10502, opened 2022-04-04
Add new `Style/FetchEnvVar` cop

johnny-miyake:
Reading environment variables using brackets like `ENV['SOME_KEY']` may unexpectedly be evaluated as `nil` if the variable is unset.

Because ENV behaves like a Hash, I think the style guide for Hash can be applied to ENV as well.
https://rubystyle.guide/#hash-fetch-defaults

This cop suggests `ENV.fetch` for the replacement of `ENV[]`.

-----------------

* [x] Added tests.


johnny-miyake:
Seems this cop does not attract anyone; I’m closing this PR :)

bbatsov:
We've got a pretty big backlog, so we're moving relatively slowly through all PRs. Lack of feedback, doesn't mean we're not interested in some proposal. I hope to get to review this over the weekend.

bbatsov:
I just reviewed the cop and I like the idea overall. It seems, however, there's no cop for checking the same thing for hashes (likely because it's hard to infer what's a hash or not). I've been wondering if it makes sense to try to make a more generic cop that checks hashes and ENV (maybe with some configuration options to decide what exactly you want to check) or just limit this to what's currently being proposed. The current cop definitely has the advantage of being free from false positives which is nice. @rubocop/rubocop-core Any opinions?

johnny-miyake:
@bbatsov Thank you for reviewing the code. (and sorry for my impatiently closing the PR)
As you said, in a static analysis, it's hard to know whether an object is a hash except in the case it is a hash literal. Plus, developers can implement their `[]` and `[]=` methods; we could need to care about not having false positives for that case if we want to make a more generic cop.

bquorning:
In my opinion, guarding against `ENV[]` is much more useful than guarding against `Hash#[]`.

.rubocop.yml
```diff
@@ -42,6 +42,15 @@ Style/IpAddresses:
   Exclude:
     - spec/rubocop/cop/style/ip_addresses_spec.rb
 
+Style/ReadingEnvVarWithoutDefaultValue:
+  ExcludedEnvVars:
```
bbatsov, line 46: I'd name this `AllowedEnvVars` or just `AllowedVars`.

bbatsov:
> https://rubystyle.guide/#hash-fetch-defaults

You can also add this to the cop metadata as a reference.

johnny-miyake:
How about `FetchEnvVar`? There are cases where `fetch` without the default value fits more (e.g., raising an error against an unset environment variable as a circuit breaker.), so we can suggest using `fetch` while probably should leave the developers to decide to use the default value.

For the same reason, I'm wondering if the error message below was better, while the auto-correction probably should give `nil` for the default value not to break the software's current behavior.

```diff
- Use `ENV.fetch('X', nil)` instead of `ENV['X']`.
+ Use `ENV.fetch('X')` or `ENV.fetch('X', nil)` instead of `ENV['X']`.
```

What do you think?

bbatsov:
I like the proposed changes to the cop name and its message. 

> while the auto-correction probably should give nil for the default value not to break the software's original behavior.

Yeah, exactly.

johnny-miyake:
Thanks, I updated this PR.

.rubocop.yml
```diff
@@ -42,6 +42,15 @@ Style/IpAddresses:
...
     - spec/rubocop/cop/style/ip_addresses_spec.rb
 
+Style/FetchEnvVar:
+  AllowedVars:
+    - HOME
+    - RUBOCOP_CACHE_ROOT
+    - XDG_CACHE_HOME
+    - XDG_CONFIG_HOME
+    - CIRCLE_STAGE
+    - CI
```
bquorning, line 52: How were these 6 variable names chosen as exceptions to the new rule?

lib/rubocop/cop/style/fetch_env_var.rb
```diff
@@ -0,0 +1,113 @@
...
+
+        private
+
+        def allowed_var?(expression)
+          expression.str_type? && cop_config['AllowedVars'].include?(expression.value)
+        end
+
+        # rubocop:disable Metrics/CyclomaticComplexity
+        def offensive?(node)
+          only_node_of_expression?(node) ||
```
bbatsov, line 56: I think it'd be a good idea to add some explanatory comment here, as it's not very clear why all those checks are needed.

.rubocop.yml
johnny-miyake: Thanks for pointing it out. They are the environment variables used in this project. I listed them here to get opinions on how these should be. I think we can replace almost all `ENV[...]` in this project with `ENV.fetch(..., nil)`, except that [`ENV['XDG_CACHE_HOME']` and `ENV['HOME']` here](https://github.com/rubocop/rubocop/blob/a4f01948b7e83cec2678d359dfda426da89cce86/lib/rubocop/result_cache.rb#L74-L84) probably should be replaced with `ENV.fetch(...)` without the default values. Do you have any opinions?

lib/rubocop/cop/style/fetch_env_var.rb
johnny-miyake: Great point. I will write some comment for that.

BTW, this cop ignore some `ENV[]` in the below cases;
- when an `ENV[]` is no more than a flag for a condition
- when an `ENV[]` receives some message (This may be controversial though)

This feature is documented in the `also good` section of the documentation.
I’m going to mention about above as well, but would like to hear if this idea sounds good for others.

.rubocop.yml
johnny-miyake: I added a commit to show how the code will be.
https://github.com/rubocop/rubocop/pull/10502/commits/190111747122d83e49123b7e1279aeb41a329ebf

lib/rubocop/cop/style/fetch_env_var.rb
johnny-miyake: I added some comments and made the logic simpler.
https://github.com/rubocop/rubocop/pull/10502/commits/17a52efd5773c078698ed0d4bfa6dd4aad255f34

bbatsov:
Looks good. Just address the conflict in `result_cache.rb` and we're good to merge.

johnny-miyake:
I resolved the conflict and fixed a method so that it can cover another case.
https://github.com/rubocop/rubocop/pull/10502/commits/1270cc3cf6db96fd001605686dce63c8d63723dd

bbatsov:
Thanks!

benpickles:
The thing about ENV is that its values (and keys) will always be a (frozen) string so the Ruby Style Guide `Hash#fetch` recommendation isn't applicable because the `false` value case cannot occur. This means that `ENV['foo']` and `ENV.fetch('foo', nil)` have the same behaviour and don't seem to protect against any meaningful issue (correct me if I'm wrong).

johnny-miyake:
As you pointed out, the Style Guide `Hash#fetch` is to distinguish `nil` and `false`, so it's not applicable to ENV. The description of this PR may have been confusing because the actual purpose of this cop is different from it.

The purpose of this cop is to alert the developers that the code to read an environment variable is possibly `nil` on runtime. If the developer assumes that the environment variable is absolutely set, using `ENV.fetch` would be a better choice than `ENV[]` because `fetch` raises an error when the variable is not set and stops the execution immediately not expanding the incident. If they want to intentionally allow the case that the variable is `nil`, specifying the default value `nil` can help others know the developer's intention.
(This idea is based on my experience, but I know it depends on the situation or how the developers think.)

rmacklin:
Hey there, wanted to point out something I noticed when running this cop's autocorrection.

Code like
```rb
value = ENV['EXAMPLE_KEY'] || 'default_value'
```
gets autocorrected to
```rb
value = ENV.fetch('EXAMPLE_KEY', nil) || 'default_value'
```

I ended up manually changing that to
```rb
value = ENV.fetch('EXAMPLE_KEY', 'default_value')
```
in my codebase. It'd be neat if the autocorrection could handle that pattern.

ShockwaveNN:
@rmacklin I agree with you, I've replaced a lot of `a = ENV['a'] || 'default'` in my code base and I really like `a = ENV.fetch('a', 'default')` - it's look nicer and tider

johnny-miyake:
That's true. I opened an issue for this enhancement.
https://github.com/rubocop/rubocop/issues/10578

PR #10503, opened 2022-04-04
[FIx #10356] Add `AllowConsecutiveConditionals` option to `Style/GuardClause`

### LINKED ISSUE #10356: Feature Request: Style/GuardClause should have an option to skip in cases where there are multiple if statements
The `Style/GuardClause` cop will trigger in the below example. I'd prefer to keep both if statements if they are happening at the same level (non-nested)

```ruby
def my_method
  if condition_a
    do_something_a
  end

  if condition_b # <-- Style/GuardClause triggers here (undesirable)
    do_something_b
  end
end
```

Nested if statements should still be made into guard clauses, like so:

```ruby
def my_method
  if condition_a # <--
    do_something_a
    if condition_b # <--
      do_something_b
    end
  end
end

# this is always desirable for Style/GuardClause
def my_method
  return unless condition_a
  do_something_a
  return unless condition_b
  do_something_b
end
```

ydah:
Resolve #10356

This PR add `AllowConsecutiveConditionals` option to `Style/GuardClause` and the option is false by default.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_add_option_for_guard_clause.md
+++ b/changelog/change_add_option_for_guard_clause.md
@@ -1,0 +1,1 @@
+* [#10356](https://github.com/rubocop/rubocop/issues/10356): Add `AllowConsecutiveConditionals` option to `Style/GuardClause` and the option is false by default. ([@ydah][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -3634,10 +3634,11 @@ Style/GuardClause:
   StyleGuide: '#no-nested-conditionals'
   Enabled: true
   VersionAdded: '0.20'
-  VersionChanged: '0.22'
+  VersionChanged: <<next>>
   # `MinBodyLength` defines the number of lines of the a body of an `if` or `unless`
   # needs to have to trigger this cop
   MinBodyLength: 1
+  AllowConsecutiveConditionals: false
 
 Style/HashAsLastArrayItem:
   Description: >-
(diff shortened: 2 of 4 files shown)


bbatsov:
I'm fine with the proposal in principle, but I don't like very much the name of this variable. Something like `AllowMultipleSameDepthConditionals` sounds a bit better, even if it's longer. Naming is hard, maybe @rubocop/rubocop-core have better ideas.

ydah:
My naming of variables was certainly not good. I think `AllowMultipleSameDepthConditionals` looks better.
I updated this PR.

bbatsov:
Btw, there's one detail of this we need to clarify - do we want to apply it if there's *any* conditional at the same depth or only if the preceding expression is a conditional at the same depth. The implementation seems to imply the former, but at least to me it seems that that latter probably makes more sense. The OP (@johnnyshields) is welcome to share his thoughts as well.

johnnyshields:
@bbatsov thanks for pinging me. I'm somewhat agnostic to either approach. I think I might favor it to be the "only if the preceding expression is a conditional at the same depth", but probably either is fine.

ydah:
@bbatsov @johnnyshields Thanks so much for the feedback!
I too think it is "only good when there are conditions of the same depth immediately prior to the event".
Certainly, I'd like "only if the preceding expression is a conditional at the same depth" is better.

I updated this PR. Thank you so much!

bbatsov:
Good. Now that we've settled this I have a proposal how to simplify the config name further - "AllowConsecutiveConditionals". That's essentially the same as before, but a bit more specific and easier to parse mentally.

johnnyshields:
@bbatsov yep perfect name its 100% clear.

ydah:
@bbatsov Yeah perfect name! I updated this PR. Thank you so much!

bbatsov:

PR #10508, opened 2022-04-06
[Fix#10191] Add `MaxChainLength` option to `Style/SafeNavigation`

### LINKED ISSUE #10191: Restrict or configure`Style/SafeNavigation` to only apply when not chained

Since [this change](https://github.com/rrosenblum/rubocop/commit/a1013297f71349f775323367cc48e3863168c869#diff-23cac2ab7ea30e00b6179c2f3afe5f0da5ff67696318e6efaa8a0d121d41d9b6R102-R104) `Style/SafeNavigation` only does this if there are two methods chained. That was a very important improvement, in my opinion.

From the [examples](https://docs.rubocop.org/rubocop/1.0/cops_style.html#examples-153):

```ruby
foo && foo.bar # => foo&.bar (I like this behavior)
foo && foo.bar.baz # => foo&.bar&.baz (I do not like this behavior)
foo && foo.bar.baz.qux  # => foo && foo.bar.baz.qux (I like this behavior)
```

I would like to _only_ correct the single `foo && foo.bar` case into `foo&.bar`. This is generally helpful as it is equivalent and concise. On the other hand, `foo && foo.bar.baz` into `foo&.bar&.baz` is not exactly the same. It will always _work_ where it would have worked before, but it is more complicated, does not communicate the intent that I had before which is that if we have a `bar`, we will always have a `baz`, and it does not implicitly assert that at runtime (which could prevent surprising behavior if it's accidentally `nil` and shouldn't be).

As a motivating example, I had this code today to strip off a country prefix from a phone number. There are other ways we could write it (move the logic onto the object, etc.) but for this one of case this was a simple option:

```ruby
phone && phone.number.delete_prefix('+1')
```

If we have a phone object, we definitely have a number (this is even database enforced in this case).  But rubocop changed the code to:

```ruby
phone&.number&.delete_prefix('+1')
```

This still _works_ but it's clearly worse in my opinion because it doesn't preserve the knowledge (and implicit assertion) that a number will never be `nil` so long as we have a `phone`. This is basically embedding an extra `if` statement which seems less clear and I'm curious if it might even be a tiny bit less performant.

I would be very happy if the cop just stopped correcting chains OR nearly as happy if there was a configuration option for controlling the maximum chain length (or just turning chain correction on and off).

I have considered disabling the cop but it seems like a shame because it is helpful for simple `a && a.b` cases.

## Additional context

N/A

ydah:
Resolve #10191

This PR add `MaxChainLength` option to `Style/SafeNavigation` and the option is 2 by default.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_add_option_for_safe_navigation.md
+++ b/changelog/change_add_option_for_safe_navigation.md
@@ -1,0 +1,1 @@
+* [#10191](https://github.com/rubocop/rubocop/issues/10191): Add `MaxChainLength` option to `Style/SafeNavigation` and the option is 2 by default. ([@ydah][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -4656,7 +4656,7 @@ Style/SafeNavigation:
                   be `nil` or truthy, but never `false`.
   Enabled: true
   VersionAdded: '0.43'
-  VersionChanged: '0.77'
+  VersionChanged: <<next>>
   # Safe navigation may cause a statement to start returning `nil` in addition
   # to whatever it used to return.
   ConvertCodeThatCanStartToReturnNil: false
@@ -4667,6 +4667,8 @@ Style/SafeNavigation:
     - try
     - try!
   SafeAutoCorrect: false
+  # Maximum length of method chains for register an offense.
+  MaxChainLength: 2
 
 Style/Sample:
   Description: >-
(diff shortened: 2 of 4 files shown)


bbatsov:
I think "length" is a better word than "size" in this context.

ydah:
I certainly thought "length" was better too.

PR #10520, opened 2022-04-10
[Fix #10469] Fix code length calculation when kwargs written in single line

### LINKED ISSUE #10469: CodeLength incorrectly calculated when there are kwargs that should be counted as one
When there are keyword arguments, and the configuration says that hashes should be counted as one line, the code length module calculates the length incorrectly, by always subtracting 2.

I tracked the issue down to the `Metrics::Utils::CodeLengthCalculator` which includes this line
```
# Subtract 2 length of opening and closing brace if method argument omits hash braces.
length -= 2 if descendant.hash_type? && !descendant.braces?
```

This is valid in this example:
```
foo(
              foo: :bar,
              baz: :quux
            )
```
But not this one:
```
foo(foo: :bar, baz: :quux)
```

As a reproduction spec:
```
it 'folds hashes as method kwargs if asked' do
        source = parse_source(<<~RUBY)
          def test
            foo(foo: :bar, baz: :quux)
          end
        RUBY

        length = described_class.new(source.ast, source, foldable_types: %i[hash]).calculate
        expect(length).to eq(1)
      end
```
This gives a result of -1

## RuboCop version

Tested on master

```
$ rubocop -V 
1.26.1 (using Parser 3.1.1.0, rubocop-ast 1.16.0, running on ruby 2.6.8 universal.arm64e-darwin21)
  - rubocop-performance 1.13.3
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.9.0
```

nobuyo:
Fixes #10469.

This PR fixes a wrong calculation for code length of kwargs at single line when asked folding.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_code_length_calculation_when_kwargs_written_in_single_line.md
+++ b/changelog/fix_code_length_calculation_when_kwargs_written_in_single_line.md
@@ -1,0 +1,1 @@
+* [#10469](https://github.com/rubocop/rubocop/issues/10469): Fix code length calculation when kwargs written in single line. ([@nobuyo][])
--- a/lib/rubocop/cop/metrics/utils/code_length_calculator.rb
+++ b/lib/rubocop/cop/metrics/utils/code_length_calculator.rb
@@ -31,7 +31,7 @@ module RuboCop
               descendant_length = code_length(descendant)
               length = length - descendant_length + 1
               # Subtract 2 length of opening and closing brace if method argument omits hash braces.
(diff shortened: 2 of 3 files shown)


nobuyo:
@koic Could you review this PR too?
According to the commit history, you committed same lines just before this change, so I’d be appriciated if take a look.

koic:
Thanks!

PR #10523, opened 2022-04-11
[Fix #10515] Add --json-schema generator

### LINKED ISSUE #10515: JSON Schema for Rubocop Config

It's kinda laborious to maintain my .rubocop.yml file - I spend a lot of time switching between my browser and editor, have to worry about misspelling the names of rules/options, etc. This should all happen in my editor, like it does for other sorts of config files (eslint, package.json, github actions config, etc.)

Could we generate a JSON schema for RuboCop config files? I believe we can get all the information on Cops we need from the `default.yml` files in RuboCop and extensions. Lacking that, we can also get at least the cop names via introspection.

I'd love to get a PR started for this if you're open to it.

A few ideas for making the schema available:
- Add an flag to the rubocop executable that generates a json schema file (using the user's project config.) The user is responsible for putting it somewhere and letting their [editor know about it.](https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings)
- Bundle it with the gem, user lets their editor know the filepath (every extension would need to do this too.)
- Generate it with the docs and make it publicly available at the docs site (user lets their editor know the URL, extensions would have to do this as well.)
- Generate and push it to the [JSON Schema Store](https://github.com/SchemaStore/schemastore) as part of the release process. Not sure on the specifics of this or whether it can be automated.

There's a gem by @hx that does this, might be able to use it as a starting point:
https://github.com/hx/rubocop-schema

It appears to accomplish this by downloading and parsing the adoc files from the RuboCop github repo. This seems kinda roundabout but maybe there are considerations I'm not aware of.

## Additional context

- [JSON Schema](https://json-schema.org/)
- [JSON Schema Store](https://github.com/SchemaStore/schemastore)

Editor config instructions:
- [VSCode](https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings)
- [JetBrains](https://www.jetbrains.com/help/idea/json.html#ws_json_schema_add_custom)
- [vim](https://github.com/neoclide/coc-json)

dleavitt:
Per #10515, adds a `--json-schema` flag to output a JSON Schema for your project's RuboCop configuration file.

This allows your IDE to autocomplete/validate the cop, department and global settings in your `.rubocop.yml`, including those of any extensions you've added via `require`.

Notes:
- As of now this just adds an option `--json-schema` that prints the schema to stdout. The user is on their own from there, though I did add examples for vscode and intellij to the docs.
- In the long-term it would be better to generate the schema and add it to the [Schema Store](https://github.com/SchemaStore/schemastore) as part of the release process. This will allow editors supporting JSON/YAML schemas to pick it up automatically, but would make it harder to incorporate cops/departments from extensions.
- The test coverage for this is fairly light - to go much further we'd need to need to incorporate a JSON schema reference resolver or add some helpers along these lines to the specs.
- To build the list of rules and their defaults, we want to look at the project config file but only use the `require` key from it. This is accomplished in a kinda hacky way as it seemed like the file loading was tightly coupled to the require resolution. Looking at it again there might be a cleaner way.

-----------------

* [x] Added tests.


bbatsov:
Seems like an useful addition. As I don't use VS Code I can't comment much on the implementation, but we'll need to get the CI build green before we can move forward with this PR.

dleavitt:
Thanks! Back to green.

lib/rubocop/cli/command/json_schema.rb
```diff
@@ -0,0 +1,187 @@
...
+            begin
+              # silence RuboCop's "pending" warning
+              old_verbose = $VERBOSE
+              $VERBOSE = nil
+              ConfigLoader.configuration_from_file(f.path)
+            ensure
+              $VERBOSE = old_verbose
+            end
+          end
+        end
```
dleavitt, line 73: All of this is just to load the default config for RuboCop and any required extensions.

Any ideas on better better ways to accomplish this?

dleavitt:
@hx - any interest in taking a look at this? Similar intent to [your repo](https://github.com/hx/rubocop-schema).

lib/rubocop/cli/command/json_schema.rb
bbatsov: @jonas054 might have some ideas. He's our master of configuration. :-)

jonas054: No, I can't think of any better solution. Your solution seems well researched, @dleavitt. Also, this is new code we're talking about, not changing anything in existing configuration loading logic, so I don't foresee any problems.

spec/rubocop/cli/options_spec.rb
```diff
@@ -1235,6 +1235,70 @@ def full_description_of_cop(cop)
...
+    it 'knows about cops' do
+      expect(properties.key?('Layout/MultilineAssignmentLayout')).to be(true)
+    end
+
+    it 'knows about lots of cops' do
+      cop_props = properties.keys.select { |k| k.include?('/') }
+      expect(cop_props.length).to be > 100
+    end
+
+    it 'knows about departments in other cities' do
```
jonas054, line 1283: I don't think we've used "city" before as part of RuboCop lingo. `InternalAffairs` is technically a _custom department,_ I believe. Its cops are loaded because of a `require` in RuboCop's root `.rubocop.yml`.

dleavitt: yeah probably too cute, I'll change it to "extension", which I think is the term used elsewhere.

I wanted to use the `Performance` rather than `InternalAffairs` as the test-case for this but had a little bit of trouble.

lib/rubocop/cli/command/json_schema.rb
dleavitt: @jonas054 Thanks! I've actually had some trouble getting it to play well with the frozen registry in the tests - specifically getting `ConfigLoader.load_file` to be deterministic.

The first time it is run in a test it seems to add all cops specified in the config `require` line, but calls in subsequent tests populate the registry _only_ with the default built-in Cops and not with additional `require`'d ones. It's specifically related to the order the tests are run in.

Doing `RuboCop::ConfigLoader.clear_options` and similar do not seem to help. Removing the `:restore_registry` meta and unfreezing the registry _does_ make the tests pass (though obvs isn't the right solution.)

Minimal repro:

```ruby
# frozen_string_literal: true

RSpec.describe RuboCop::CLI::Command::JSONSchema, :isolated_environment, :restore_registry do
  include FileHelper

  3.times do
    # contents of file are always the same but performance cops are only loaded on the first run
    it 'is deterministic' do
      create_file('.rubocop.yml', 'require: [rubocop-performance]')
      
      path = RuboCop::ConfigLoader.configuration_file_for(Dir.pwd)
      RuboCop::ConfigLoader.configuration_from_file(path)
      # RuboCop::ConfigStore.new.for_pwd, #load_file, etc give same result

      # 520 built in cops, 49 performance cops
      # always 569 on first run and 520 subsequently
      expect(RuboCop::Cop::Registry.global.length).to eq 569
    end
  end
end

```

jonas054: Yes, I know what you mean. I've tried out your minimal repro now, but was unable to find a solution. I've come across similar problems before, and something I concluded then was, that it seems impossible to make a spec example that requires something and then have that require "undone". It's just not possible in Ruby, I think.

So my suggestion is that you try to work around, or avoid, the problem, rather than solve it. Sorry I can't be more helpful.

dleavitt: No problem, that's helpful too.

bbatsov:
FYI - I'll be cutting a new release some time tomorrow, so we'll have to wrap this up today if you want it to make the release.

dleavitt:
Probably best to save it for the next release - I want to test and validate it a bit more.

dleavitt:
Hiya! Sorry to leave this branch hanging. I'm a worried there may be bugs left. 

@bbatsov would you be open to my adding a JSON schema validator as a development dependency in order to be able to test it a bit better?

bbatsov:
@dleavitt Ping :-)

dleavitt:
thanks for the reminder! I'll try to get some progress in this weekend.

bbatsov:

PR #10525, opened 2022-04-11
[Fix #10280] Add `AllowComments` option to `Style/SymbolProc`

### LINKED ISSUE #10280: Style/SymbolProc ignores comments
Style/SymbolProc ignores comments when linting, and destroys them when it auto-corrects.

--------

## Expected behavior

In the following code, I expect:
* no reported offense from Style/SymbolProc, because the block doesn't only call `x`, it also contains a comment.
* no auto-correction from Style/SymbolProc, because the block doesn't only call `x`, it also contains a comment.
```
hello do |r|
  # IMPORTANT MESSAGE
  r.x
end
```

## Actual behavior

It shows reports an offense, and auto-correct destroys the comment!

```
hello(&:x)
```

I mean, I know the cop is unsafe, but that may be a bit too unsafe? 😆 

Paste the first snipped in a file, and run `rubocop -A` on it.

## RuboCop version

Happens in 
0.79.0 (using Parser 3.0.2.0, running on ruby 2.6.6 x86_64-linux)
and 
1.23.0 (using Parser 3.0.2.0, rubocop-ast 1.13.0, running on ruby 2.6.6 x86_64-linux)

ydah:
Resolve: https://github.com/rubocop/rubocop/issues/10280

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/symbol_proc.rb
+++ b/lib/rubocop/cop/style/symbol_proc.rb
@@ -52,7 +52,24 @@ module RuboCop
       # @example AllowMethodsWithArguments: true
       #   # good
       #   something.do_something(foo) { |o| o.bar }
+      #
+      # @example AllowComments: false (default)
+      #   # bad
+      #   something.do_something do |s| # some comment
+      #     # some comment
+      #     s.upcase # some comment
+      #     # some comment
+      #   end
+      #
+      # @example AllowComments: true
+      #   # good  - if there are comment in either position
+      #   something.do_something do |s| # some comment
+      #     # some comment
+      #     s.upcase # some comment
+      #     # some comment
+      #   end
       class SymbolProc < Base
+        include CommentsHelp
         include RangeHelp
         include IgnoredMethods
         extend AutoCorrector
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/style/symbol_proc.rb
```diff
@@ -42,6 +42,13 @@ module Style
       #   # good
       #   something.map(&:upcase)
       #
+      #   # good  - if there are comment in either position
```
koic, line 45: Maybe this is a corner case and it would be better to provide `AllowComments: false (default)` option. if `AllowComments: false`, it will not be auto-corrected. And it would be user-friendly to mention it in the document.

ydah: @koic It's certainly better that way. Thank you so much.
I've made the correction, could you please review again?

lib/rubocop/cop/style/symbol_proc.rb
```diff
@@ -100,6 +119,11 @@ def destructuring_block_argument?(argument_node)
 
         private
 
+        def contains_comment?(node)
```
bbatsov, line 122: Why don't you use the `CommentsHelp` module instead? I think it provides exactly the same method there.

bbatsov:
The proposed changes look good to me, apart from the note about reusing `CommentsHelp` to remove some code duplication.

lib/rubocop/cop/style/symbol_proc.rb
ydah: I'd missed it. Thank you so much.

ydah:
@bbatsov Thank you so much. I updated this PR.

bbatsov:
Thanks!

PR #10526, opened 2022-04-11
[Fix #10522] Add new `Style/ObjectThen` cop

### LINKED ISSUE #10522: Cop idea: standardize on `#then` instead of `#yield_self`

It seems like it would be nice to have a consistent style for `then` v.s. `yield_self` like there are cops for some `Enumerable` methods.

A new cop with autocorrect (maybe configurable style?) that converts `#yield_self` to `#then`. It would be unsafe to autocorrect.

N/A

## Additional context

Not sure this is a high priority/huge need, but I thought I would share the idea. Thanks!

ydah:
Resolve: #10522.

This cop enforces the use of consistent method names `Object#yield_self` or `Object#then`.

## example

```ruby
# EnforcedStyle: then (default)

# bad
obj.yield_self { |x| x.do_something }

# good
obj.then { |x| x.do_something }
```

```ruby
# EnforcedStyle: yield_self

# bad
obj.then { |x| x.do_something }

# good
obj.yield_self { |x| x.do_something }
```

-----------------

* [x] Added tests.


lib/rubocop/cop/style/object_methods.rb
```diff
@@ -0,0 +1,59 @@
...
+      #       then: yield_self
+      #
+      # @example
+      #   # These examples are based on the default mapping for `PreferredMethods`.
+      #
+      #   # bad
+      #   obj.yield_self
+      #
+      #   # good
+      #   obj.then
```
koic, line 24: I see cases where block arguments are used.

```suggestion
      #   # bad
      #   obj.yield_self { |x| x.do_something }
      #
      #   # good
      #   obj.then { |x| x.do_something }
```

koic:
If this design is better, `InstanceMethods` would be better named than `ObjectMethods`.

However, it may be better to implement the cop called `YieldSelf` without over-abstraction. In that case, I think user can select `EnforcedStyle: yield_self` and `EnforcedStyle: then`. Also, which one is the preferred default can be discussed in the style guide.
https://github.com/rubocop/ruby-style-guide/issues

lib/rubocop/cop/style/object_methods.rb
ydah: Oops, Sorry. I’ll add it.

config/default.yml
```diff
@@ -5135,6 +5135,18 @@ Style/WordArray:
   # The regular expression `WordRegex` decides what is considered a word.
   WordRegex: !ruby/regexp '/\A(?:\p{Word}|\p{Word}-\p{Word}|\n|\t)+\z/'
 
+Style/YieldSelf:
+  Description: 'Preferred object methods.'
```
koic, line 5139: https://github.com/rubocop/ruby-style-guide/pull/895 has been merged. Can you insert the style guide link?
```suggestion
  Description: 'Preferred object methods.'
  StyleGuide: '#object-yield-self-vs-object-then'
```

koic: And can you update text of `Description`?

ydah: Thank you for your review. I've added a link to the style guide and updated the `Description`.

ydah:
Thank you so much. I updated this PR.

bbatsov:
I like the cop and I was about the press the "rebase" button when I realized that the name is a bit too specific in the sense that `YieldSelf` is in it. I've been wondering if there's a generic name for the abstraction that "yield_self" represents, but I can't think of anything. I can live with the current name, but if we can come up with something slightly more abstract that'd be better IMO.

ydah:
@bbatsov Thank you so much. I couldn't think of better name than `YieldSelf` either. It may not be optimal, but given the precedent of [`StringChars`](https://docs.rubocop.org/rubocop/cops_style.html#stylestringchars), how about `ObjectYieldSelf` for naming consistency?

ydah:
@bbatsov Following the precedent of [`Style/StringChars`](https://docs.rubocop.org/rubocop/cops_style.html#stylestringchars), I propose `Style/ObjectThen`. What do you think?

bbatsov:
Let's go with `ObjectThen`. Thanks for working on this!

PR #10527, opened 2022-04-11
[Fix #10524] Mark `Style/RedundantInitialize` as unsafe

### LINKED ISSUE #10524: Style/RedundantInitialize dislikes an empty initialize that changes the arity
```ruby
class A
  def initialize(a)
    @a = a
  end
end

class B < A
  def initialize
  end
end

B.new
```

## Expected behavior

All initialize methods accomplish something, so there should be no violation of `Style/RedundantInitialize`.

## Actual behavior

It considers `B#initialize` a violation of `Style/RedundantInitialize`, even though its not redundant. Removing it breaks the final line.

run `rubocop` with the `Style/RedundantInitialize` cop enabled on the code sample above.

## RuboCop version

```
1.27.0 (using Parser 3.1.1.0, rubocop-ast 1.17.0, running on ruby 2.6.8 arm64-darwin21)
  - rubocop-performance 1.13.3
```

koic:
Fixes #10524.

This PR marks `Style/RedundantInitialize` as unsafe because if subclass overrides `initialize` method with a different arity than superclass.

```ruby
class Base
  def initialize
  end
end

class Derived < Base
  def initialize(arg)
  end
end

Derived.new(arg) # Requires one argument.
```

Therefore, the following is a false positive.

```console
% bundle exec rubocop example.rb --only Style/RedundantInitialize
Inspecting 1 file
C

Offenses:

example.rb:8:3: C: Style/RedundantInitialize: Remove unnecessary empty initialize method.
  def initialize(a) ...
  ^^^^^^^^^^^^^^^^^

```

The same can be said for Module mix-in, which cannot be determined by static analysis.

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/change_mark_unsafe_for_style_redundant_initialize.md
+++ b/changelog/change_mark_unsafe_for_style_redundant_initialize.md
@@ -1,0 +1,1 @@
+* [#10524](https://github.com/rubocop/rubocop/issues/10524): Mark `Style/RedundantInitialize` as unsafe. ([@koic][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -4539,7 +4539,9 @@ Style/RedundantFreeze:
 Style/RedundantInitialize:
   Description: 'Checks for redundant `initialize` methods.'
   Enabled: pending
+  Safe: false
   VersionAdded: '1.27'
+  VersionChanged: '<<next>>'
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #10540, opened 2022-04-15
Add new MaxAllowedLettersBeforeNumber option to Naming/VariableNumber snake_case

### LINKED ISSUE #10539: Allow _v1 suffix in Naming/VariableNumber without having to specify every individual method

We've started suffixing some methods with `_v1`, `_v2`, etc. to indicate when we're effectively versioning them and introducing new versions, so we can do "soft" code releases where we start using the new method, and can still use the old method alongside to identify and upgrade existing things that were stored by the old method. E.g.
```diff
 def model_details_hasher(model)
+  NewHashingAlgorithm(model.to_json)
+end
+
+def model_details_hasher_v1(model)
   SomeHashingAlgorithm(model.to_s)
 end
```

We have `Naming/VariableNumber` set to `EnforcedStyle: snake_case` so the cop says it must be `_v_1`, `_v_2`, etc.
image
image

We prefer `_v1` to `_v_1`, so we'd like to be able to always allow these methods, whenever we add new ones. We also want to keep the rest of the rules under `EnforcedStyle: snake_case` the same.

Perhaps by allowing a regex to be configured, like `AllowIf: /_v\d$/`?

Or offering a specific option like `AllowVersionSuffixInSnakeCase` that does the same as the above, when the cop is set to `EnforcedStyle: snake_case`? This would be helpful if you want to control the speed of this feature (if allowing any regex could slow down rubocop and cause more bug reports/support issues as a result), and if other people would also want to use this combination of snake_case whilst allowing (imo) nicer-reading `_v1`.

Or even an option like `MaxAllowedLettersBeforeNumber` so you can also allow `_a1`, `_b2`, etc.
```yaml
# .rubocop.yml
Naming/VariableNumber:
  EnforcedStyle: snake_case
  MaxAllowedLettersBeforeNumber: 1 # default is 0
```

Disable the cop inline on every method, which then feels like that particular method is "going against style", when it's actually the style we want.
image

Or add each new method name to the `AllowedIdentifiers` array on the cop settings, which will become tiresome as we use this method of code versioning more and more into the future.
```yaml
# .rubocop.yml
Naming/VariableNumber:
  EnforcedStyle: snake_case
  AllowedIdentifiers:
    - model_details_hasher_v1
```

henrahmagix:
Fix #10539 

Adds a new integer option called `MaxAllowedLettersBeforeNumber` to
allow a positive number of non-digits before the digits in a snake_case
style that would otherwise require separation between them, e.g. `_v1`.

-----------------

I wasn't sure how to update [docs/modules/ROOT/pages/cops_naming.adoc](docs/modules/ROOT/pages/cops_naming.adoc) to include this new option – it looks like it should go there, but I'm not familiar with the format and I wasn't sure how to explain this option that only applies to `Naming/VariableNumber` and not the other naming cops that have share config.

Just made this to get the ball rolling on #10539 (and make it more likely to be actioned =D), but I don't mind if it's not wanted, it's no sweat =)

### DIFF
--- a/lib/rubocop/cop/naming/variable_number.rb
+++ b/lib/rubocop/cop/naming/variable_number.rb
@@ -102,6 +102,19 @@ module RuboCop
 
         MSG = 'Use %<style>s for %<identifier_type>s numbers.'
 
+        def max_allowed_letters_before_number
+          cop_config.fetch('MaxAllowedLettersBeforeNumber', 0)
+        end
+
+        def valid_name?(node, name, given_style = style)
+          return true if super
+
+          return false unless given_style == :snake_case
+          return false unless max_allowed_letters_before_number.to_i.positive?
+
+          name.match?(/(?:_|\b)\D{0,#{max_allowed_letters_before_number}}\d+\z/)
+        end
+
         def on_arg(node)
           @node = node
           name, = *node
--- a/changelog/new_add_new_option_to_namingvariablenumber.md
+++ b/changelog/new_add_new_option_to_namingvariablenumber.md
@@ -1,0 +1,1 @@
+* [#10539](https://github.com/rubocop/rubocop/issues/10539): Add new MaxAllowedLettersBeforeNumber option to Naming/VariableNumber snake_case. ([@henrahmagix][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2787,6 +2787,7 @@ Naming/VariableNumber:
     - rfc822       # Time#rfc822
     - rfc2822      # Time#rfc2822
     - rfc3339      # DateTime.rfc3339
+  MaxAllowedLettersBeforeNumber: 0
 
 #################### Security ##############################
 
--- a/spec/rubocop/cop/naming/variable_number_spec.rb
+++ b/spec/rubocop/cop/naming/variable_number_spec.rb
@@ -33,6 +33,14 @@ RSpec.describe RuboCop::Cop::Naming::VariableNumber, :config do
     end
   end
 
+  shared_examples 'accepts_array' do |style, variables|
+    it "accepts #{variables} in #{style}" do
+      lines = variables.map { |v| "#{v} = 1" }
+
+      expect_no_offenses(lines.join("\n"), first_variable: variables.first)
+    end
+  end
+
   shared_examples 'accepts integer symbols' do
     it 'accepts integer symbol' do
       expect_no_offenses(':"42"')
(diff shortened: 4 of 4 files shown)


lib/rubocop/cop/naming/variable_number.rb
```diff
@@ -102,6 +102,19 @@ class VariableNumber < Base
...
+          cop_config.fetch('MaxAllowedLettersBeforeNumber', 0)
+        end
+
+        def valid_name?(node, name, given_style = style)
+          return true if super
+
+          return false unless given_style == :snake_case
+          return false unless max_allowed_letters_before_number.to_i.positive?
+
+          name.match?(/(?:_|\b)\D{0,#{max_allowed_letters_before_number}}\d+\z/)
```
henrahmagix, line 115: For readability, this could also be:
```suggestion
          super || (
            given_style == :snake_case && max_allowed_letters_before_number.to_i.positive? &&
            name.match?(/(?:_|\b)\D{0,#{max_allowed_letters_before_number}}\d+\z/)
          )
```

dvandersluis:
I wonder if adding something like [`IgnoredPattern`](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/mixin/ignored_pattern.rb) -- would be better to allow a pattern to be ignored by the cop than special casing `vN`. We'd probably want to apply the pattern to the identifier rather than the whole line but it's a start.

> I wasn't sure how to update [docs/modules/ROOT/pages/cops_naming.adoc](https://github.com/rubocop/rubocop/pull/docs/modules/ROOT/pages/cops_naming.adoc) to include this new option

The docs are automatically recreated when a release is made using the top comments for each cop. To that end, you should add `@example` sections with good and bad examples for new configuration options.

henrahmagix:
Great idea @dvandersluis  =)

henrahmagix:
@dvandersluis In making this, it'll be an option alongside existing option `AllowedIdentifiers` (in both `Naming/VariableName` and `Naming/VariableNumber`).

Do you think that wording could cause confusion?
```yaml
AllowedIdentifiers:
  - capture3
IgnoredPatterns:
  - /v\d+\z/
```

Like just the fact that both `Allowed` and `Ignored` are doing the same thing, but sound like inverse logic when compared to each other.

This will be the first place where `Allowed*` is alongside `Ignored*` in the options. Other places have `Allow*` with `Ignored*`, but not `Allowed*`.

I'm going to build it anyway because `IgnoredPatterns` is an existing mixin and will be consistent with other modules, I just wanted to be sure there's not going to be some antagonistic effects of this wording =)

dvandersluis:
@henrahmagix yeah it is a bit confusing I guess. We could use `AllowedPatterns` I suppose but then it wouldn't be consistent with configuration for other cops. @bbatsov thoughts?

Since `AllowedIdentifiers` is more specific I would have that take effect first and if there's no match then check the patterns, if any.

bbatsov:
I used to use `IgnoredSomething` a lot in the early days of the project, but at some point I realized that `Allowed` reflects better how things work, plus it's easier to mentally parse. The mixin is a legacy-one, but it still makes sense to use it. At some point I guess we should make `IgnoredPatterns` and `AllowedPatterns` a synonym for every cop in an effort to make the configuration more uniform.

dvandersluis:
> At some point I guess we should make IgnoredPatterns and AllowedPatterns a synonym for every cop in an effort to make the configuration more uniform.

I opened https://github.com/rubocop/rubocop/pull/10555.

dvandersluis:
@henrahmagix you can use `AllowedPatterns` instead now 😁

PR #10543, opened 2022-04-16
Fix incorrect code length calculations for few more patterns

nobuyo:
Follow up #10469 and #10520.

This PR improve code length calculation for few more patterns like follows:

```rb
def test
  foo(
    foo,
    foo: :bar,
  )
end

def test
  foo(foo: :bar,
      baz: :quux)
end
```

Previously, it always subtracted 2 under certain conditions, but now it considers cases where hash begins/ends on the same line as the parentheses or there are other arguments.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/metrics/utils/code_length_calculator.rb
+++ b/lib/rubocop/cop/metrics/utils/code_length_calculator.rb
@@ -30,8 +30,8 @@ module RuboCop
 
               descendant_length = code_length(descendant)
               length = length - descendant_length + 1
-              # Subtract 2 length of opening and closing brace if method argument omits hash braces.
-              length -= 2 if descendant.hash_type? && !descendant.braces? && descendant.multiline?
+              # Subtract length of opening and closing brace if method argument omits hash braces.
+              length -= omit_length(descendant) if descendant.hash_type? && !descendant.braces?
             end
 
             length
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/metrics/utils/code_length_calculator.rb
```diff
@@ -153,6 +153,20 @@ def irrelevant_line?(source_line)
...
+            return 0 if another_args?(parent)
+
+            [
+              parent.loc.begin.end_pos != descendant.loc.expression.begin_pos,
+              parent.loc.end.begin_pos != descendant.loc.expression.end_pos
+            ].count(true)
+          end
+
+          def another_args?(node)
+            node.send_type? && node.children.count > 3
```
Darhazer, line 168:
```suggestion
            node.send_type? && node.arguments.count > 1
```

is easier to understand

Darhazer (approved):
Nice test coverage. Thank you :bow:

lib/rubocop/cop/metrics/utils/code_length_calculator.rb
Darhazer: Also better use `call_type?` as it also covers csend (&.) nodes

nobuyo: Thank you, I've updated `another_args?` and add a test case.

bbatsov:
Nicely done! Thanks!

PR #10547, opened 2022-04-18
[Fix #10095] Change "auto-correct" to "autocorrect"

### LINKED ISSUE #10095: Autocorrect vs auto-correct
```sh
~/code/rubocop$ grep -iro "autocorrect" . | wc -l
    1971
~/code/rubocop$ grep -iro "auto-correct" . | wc -l
    1551
```

Both `autocorrect` and `auto-correct` are used frequently in RuboCop. Should we be consistent? If so, which one?

Any change should only affect comments and other string content, not method names.

chris-hewitt:
Resolves #10095.

More detailed searches for existing usage:
```bash
[~/projects/rubocop ⌥ master]$ git grep -E 'autocorrect|Autocorrect' -- ':!relnotes/*' ':!CHANGELOG.md' | wc -l
1283
[~/projects/rubocop ⌥ master]$ (git grep -iE 'auto-correct|auto_correct|auto correct' -- ':!relnotes/*' ':!CHANGELOG.md'; git grep -E 'autoCorrect|AutoCorrect' -- ':!relnotes/*' ':!CHANGELOG.md') | wc -l
1160
```

This PR changes usage to "autocorrect" when possible without making breaking changes.

-----------------

* [x] Added tests.

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -3,7 +3,7 @@
 * xref:compatibility.adoc[Compatibility]
 * Usage
 ** xref:usage/basic_usage.adoc[Basic Usage]
-** xref:usage/auto_correct.adoc[Auto-correct]
+** xref:usage/autocorrect.adoc[Autocorrect]
 ** xref:usage/caching.adoc[Caching]
 * xref:configuration.adoc[Configuration]
 * xref:cops.adoc[Cops]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,7 +86,7 @@ Just make sure they are properly named.
 Here are a few examples:
 
 ```
-* [#716](https://github.com/rubocop/rubocop/issues/716): Fixed a regression in the auto-correction logic of `MethodDefParentheses`. ([@bbatsov][])
+* [#716](https://github.com/rubocop/rubocop/issues/716): Fixed a regression in the autocorrection logic of `MethodDefParentheses`. ([@bbatsov][])
 * New cop `ElseLayout` checks for odd arrangement of code in the `else` branch of a conditional expression. ([@bbatsov][])
 * [#7542](https://github.com/rubocop/rubocop/pull/7542): **(Breaking)** Move `LineLength` cop from `Metrics` department to `Layout` department. ([@koic][])
 ```
--- a/changelog/change_auto_correct_to_autocorrect.md
+++ b/changelog/change_auto_correct_to_autocorrect.md
@@ -1,0 +1,1 @@
+* [#10095](https://github.com/rubocop/rubocop/issues/10095): Change "auto-correct" to "autocorrect" in arguments, documentation, messages, comments, and specs. ([@chris-hewitt][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -928,7 +928,7 @@ Layout/IndentationStyle:
   VersionChanged: '0.82'
   # By default the indentation width from `Layout/IndentationWidth` is used,
   # but it can be overridden by setting this parameter.
-  # It is used during auto-correction to determine how many spaces should
+  # It is used during autocorrection to determine how many spaces should
   # replace each tab.
   IndentationWidth: ~
   EnforcedStyle: spaces
@@ -4706,7 +4706,7 @@ Style/SafeNavigation:
                   Transforms usages of a method call safeguarded by
                   a check for the existence of the object to
                   safe navigation (`&.`).
-                  Auto-correction is unsafe as it assumes the object will
+                  Autocorrection is unsafe as it assumes the object will
                   be `nil` or truthy, but never `false`.
   Enabled: true
   VersionAdded: '0.43'
--- a/docs/modules/ROOT/pages/automated_code_review.adoc
+++ b/docs/modules/ROOT/pages/automated_code_review.adoc
@@ -19,7 +19,7 @@ https://codeclimate.com/[Code Climate] provides automated code review for test c
 
 == CodeFactor
 
-https://www.codefactor.io[CodeFactor] reports various code metrics like duplication, churn, and problems for code style, performance, complexity, and many others. CodeFactor is free for open source. It supports analysis and auto-correction for RuboCop.
+https://www.codefactor.io[CodeFactor] reports various code metrics like duplication, churn, and problems for code style, performance, complexity, and many others. CodeFactor is free for open source. It supports analysis and autocorrection for RuboCop.
 
 == Hound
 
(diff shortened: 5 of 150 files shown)


chris-hewitt:
Things to possibly rename in the future (with deprecated aliases to the original):
```
CLI args and their corresponding @options value
  --auto-correct     to --autocorrect
  --auto-correct-all to --autocorrect-all
  (--safe-auto-correct stays as-is, since it's already deprecated)
config setting
  AutoCorrect to Autocorrect
mixin module
  AutoCorrector to Autocorrector
rake task
  rubocop:auto_correct to rubocop:autocorrect
internal code with possible external dependencies
  RuboCop::Cop::Style::MethodCallWithArgsParentheses::OmitParentheses#auto_correct
  RuboCop::Formatter::DisabledConfigFormatter#supports_safe_auto_correct
  :safe_auto_correct initialize() argument and @safe_auto_correct variable in RuboCop::Formatter::SimpleTextFormatter::Report
  RuboCop::OptionsValidator#validate_auto_correct
also: what about auto-generate?
  --[no-]auto-gen-only-exclude
  --[no-]auto-gen-timestamp
  --auto-gen-config
  AutoGenConfigFormatter
  AutoGenerateConfig
```

docs/modules/ROOT/nav.adoc
```diff
@@ -3,7 +3,7 @@
 * xref:compatibility.adoc[Compatibility]
 * Usage
 ** xref:usage/basic_usage.adoc[Basic Usage]
-** xref:usage/auto_correct.adoc[Auto-correct]
+** xref:usage/autocorrect.adoc[Autocorrect]
```
dvandersluis, line 6: I'm not sure how to get around this but this will break links.

dvandersluis: https://docs.antora.org/antora/latest/page/page-aliases/

dvandersluis:
Thanks for tackling this!

I wouldn't change any classes or method names at this point; all the extensions make use of things like `AutoCorrector` for instance, and that will definitely break them.

I would add aliases along with this to the command line arguments (update the default spelling but allow the old as an alias) and rake task. `safe-auto-correct` should be removed as of the next major version I assume.

chris-hewitt:
Sure thing. How would you like the help text to end up - like this?
```
...
Autocorrection:
    -a, --autocorrect                Autocorrect offenses (only when it's safe).
        --auto-correct               (same, deprecated)
        --safe-auto-correct          (same, deprecated)
    -A, --autocorrect-all            Autocorrect offenses (safe and unsafe)
        --auto-correct-all           (same, deprecated)
...
```

dvandersluis:
That'd probably be the best but might be too verbose. We could maybe just support them but not document them but I'd see what @bbatsov thinks here.

docs/modules/ROOT/nav.adoc
bbatsov: Yep, we'll need an alias here or we should just push this renaming for later.

bbatsov:
> That'd probably be the best but might be too verbose. We could maybe just support them but not document them but I'd see what @bbatsov thinks here.

I'm fine with @chris-hewitt's proposal. No need to make it more complicated than things need to be. And I agree with all comments made by @dvandersluis earlier.

bbatsov:
> Is this intentional? 

Likely an oversight. @dvandersluis or @jonas054 might know better.

> If not, should I fix it, although it could be a breaking change for someone? Could we eliminate usage of deprecated @options keys, or is that var used externally?

I think this shouldn't be used externally, so it should be safe to remove.

bbatsov:
You'll also have to rebase on top of the current `master` branch.

bbatsov:
@chris-hewitt Ping :-)

chris-hewitt:
Hi @bbatsov, I was away in the mountains but I'm back now. I'm working to have this ready for review soon.

dvandersluis:
> Also, I am not exactly sure how to do this, but adding auto-correct and its various forms to the invalid words list in the spell checks we do on CI would be useful to maintain this into the future.

@bbatsov @koic @chris-hewitt any idea how we do this?

bbatsov:
There are a couple of configuration files for the two spell-checking tools that we use.

chris-hewitt:
I've been in the weeds with `auto_correct` vs. `safe_auto_correct` from the existing codebase. My conclusion is: although those two _arguments_ are synonyms, their equivalent @options keys are _not_ synonyms. Here are the relevant @options values and the different ways they're set:
```
                                                       :auto_correct  :safe_auto_correct  :auto_correct_all 
-x, --fix-layout                                       true           -                   -
-a, --auto-correct                                     true           true                -
    --safe-auto-correct                                true           true                -
-A, --auto-correct-all                                 true           -                   true
CopHelper#autocorrect_source                           true           -                   -
RuboCop::RSpec::ExpectOffense#set_formatter_options    true           -                   -
spec/rubocop/cop/lint/script_permission_spec.rb:46     true           -                   -

                                                                      ^
                used by lib/rubocop/cop/autocorrect_logic.rb:37       |
          and lib/rubocop/formatter/simple_text_formatter.rb:64 ------+
```
Therefore, I think we still need a phantom option `:safe_autocorrect` to sufficiently capture these states. I tried to specifically highlight this behavior in `spec/rubocop/options_spec.rb` in case we want to change anything about in the future. I hope that makes sense and explains some of the updates in this PR.

chris-hewitt:
Regarding spellchecking -
It looks like both spellcheckers (`codespell` & `misspell`) have their own mistake dictionaries, but neither one seems set up to accept a supplemental reject list / mistakes to check for. I have an experimentation branch [here](https://github.com/chris-hewitt/rubocop/pull/2) with one possible solution. But anyway, I think it's non-trivial and might be worth its own PR.

bbatsov:
Yeah, we can address this separately down the road.

chris-hewitt:
@bbatsov @dvandersluis Do you want me to change any of these as well? I'm not sure which ones, if any, would be non-breaking:
- `auto_correct` method in `RuboCop::Cop::Style::MethodCallWithArgsParentheses::OmitParentheses`
- `supports_safe_auto_correct?` method in `RuboCop::Formatter::DisabledConfigFormatter`
- `:safe_auto_correct` parameter in `RuboCop::Formatter::SimpleTextFormatter::Report#initialize()`
- `@safe_auto_correct` instance variable in `RuboCop::Formatter::SimpleTextFormatter::Report`
- `supports_safe_auto_correct?` method in `RuboCop::Runner::InfiniteCorrectionLoop`
- `AutoCorrect` cop parameter
- `SafeAutoCorrect` cop parameter
- `AutoCorrector` module
- `rubocop:auto_correct` rake task

bbatsov:
> Do you want me to change any of these as well? I'm not sure which ones, if any, would be non-breaking:

We can leave this for later, as most of this would require aliases and some deprecation warnings for the transition to go smoothly.

bbatsov (approved):

bbatsov:

dvandersluis:
Great job @chris-hewitt thanks!

PR #10562, opened 2022-04-21
[Fix #10472] Add `AllowPercentArray` option to `Layout/MultilineArrayLineBreaks`

### LINKED ISSUE #10472: Add configuration options to `Layout/MultilineArrayLineBreaks`

I have `Layout/MultilineArrayLineBreaks` enabled which I like for arrays in the form of `[1, 2, 3, 4]`. I want the array on a single line or each element getting its own line. However, I would like for `%` literal arrays to allow multiple elements per line. 

```ruby
%w[1 2 3 4]
# and
%w[
  1 2
  3 4
]
# are both valid
```

Add a configuration flag like `IgnorePercentArrays`

ydah:
Resolve: #10472

This PR add `AllowPercentArray` to `Layout/MultilineArrayLineBreaks` and the option is false by default.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_add_option_for_multiline_array_line_breaks.md
+++ b/changelog/new_add_option_for_multiline_array_line_breaks.md
@@ -1,0 +1,1 @@
+* [#10472](https://github.com/rubocop/rubocop/issues/10472): Add `AllowPercentArray` option to `Layout/MultilineArrayLineBreaks` and the option is false by default. ([@ydah][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -1014,6 +1014,8 @@ Layout/MultilineArrayLineBreaks:
                  starts on a separate line.
   Enabled: false
   VersionAdded: '0.67'
+  VersionChanged: <<next>>
+  AllowPercentArray: false
 
 Layout/MultilineAssignmentLayout:
   Description: 'Check for a newline after the assignment operator in multi-line assignments.'
--- a/lib/rubocop/cop/layout/multiline_array_line_breaks.rb
+++ b/lib/rubocop/cop/layout/multiline_array_line_breaks.rb
@@ -20,6 +20,22 @@ module RuboCop
       #     b,
       #     c
       #   ]
+      #
+      # @example AllowPercentArray: false (default)
+      #
+      #   # bad
+      #   %w[
+      #     1 2
+      #     3 4
+      #   ]
+      #
+      # @example AllowPercentArray: true
+      #
+      #   # good
+      #   %w[
+      #     1 2
+      #     3 4
+      #   ]
       class MultilineArrayLineBreaks < Base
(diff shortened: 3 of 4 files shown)


bbatsov:
I have to admit it's still not clear to my why someone would want to treat this syntax differently, but I'm not opposed to the proposal in principle.

ydah:
@bbatsov It’s discussed in the following: https://github.com/rubocop/rubocop/issues/10472#issuecomment-1106444619

Certainly, this case I thought it would be nice to be able to choose whether allow or not to be a `%` literal array. What do you think?

@mlwyatt Thank you so much.

bbatsov:
So, reading the comment you mentioned it seems to me that the real problem we want to solve has nothing to do with the delimiters, but it's rather about the length of an array - if an array is long it might be fine to have it in a more "compact" form (without every element being on a line of its own). @mlwyatt feel free to correct me if I'm wrong. So, perhaps there should be some "MaxLength" option (or something like this) after which the cop allows the array the to stay as it was. 

@rubocop/rubocop-core Any thoughts on this?

bbatsov:
> Would this "MaxLength" option be for ignoring single elements per line once an array reaches that length? Or a maximum number of elements per line?

I was thinking about the total number of items in the array (based on the original example you gave). I guess some number of elements per line also makes some sense as a configuration option. I'm just trying to understand the use-cases better, as experience has thought me that often the first solutions that get suggested are not the optimal solutions.

bbatsov:
@rubocop/rubocop-core Any thoughts on how to proceed here?

dvandersluis:
I'm not really sure why there's a distinction here based on array type. Like why should an array be required to be one per row if it's delimited with braces but the same array contents can be multiple on a line if in a % array.

If anything, I think what we need here is more flexibility about how to layout long arrays but I'm not sure what the solution necessarily is. The common argument for one item per line is that it makes git diffs "cleaner" but that argument doesn't really matter much to me personally.

I wonder if there are some configuration that would be helpful in an ideal world (more so than being easy to describe or implement):
* 1 item per line up to n items
* max number of items per line (does this differ based on size of array?)
* allow up to LineLength per line (the LineLength cop might already do this though)

bquorning:
Looking at the example given in https://github.com/rubocop/rubocop/issues/10472#issuecomment-1106444619, I am not sure there is any way to make a 96-element array readable.

If we made a histogram of array sizes in https://github.com/eliotsykes/real-world-rails and https://github.com/jeromedalbert/real-world-ruby-apps it would be easier to decide whether we are trying to solve a common problem, or whether this is a special case that could be solved with an `Exclude: ['app/controllers/']`,

ydah:
Sorry for leaving it unattended this PR. I am going to close this PR as I certainly don't think this approach looks good.

Issue #10557, opened 2022-04-21
[1.28.0] Problem with `ENV.fetch`


afuno:
Hello. Today updated from version 1.27.0 to version 1.28.0. Then I ran `bundle exec rubocop -A` and got this change:

```ruby
ENV.fetch("BUNDLE_GEMFILE", nil) ||= File.expand_path("../Gemfile", __dir__)
```

Before rubocop, this line looked like this:

```ruby
ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../Gemfile", __dir__)
```

This code cannot work. But this code is fixed in this way only after the update.

```ruby
1.28.0 (using Parser 3.1.2.0, rubocop-ast 1.17.0, running on ruby 3.1.2 aarch64-linux-musl)
  - rubocop-performance 1.13.3
  - rubocop-rails 2.14.2
  - rubocop-rspec 2.10.0
```

PR #10565, opened 2022-04-22
Fix a false positive and a true negative for `Style/FetchEnvVar`

ydah:
This PR improves the edge case of `Style/FetchEnvVar`.
There are cases that are false positive and true negative, and it improves them.

false positive cases are as follows:
```ruby
ENV['X'] &&= y
```

true negative cases are as follows:
```ruby
y ||= ENV['X']
y &&= ENV['X']
```

This PR follows up on the same issue as the Issue raised at https://github.com/rubocop/rubocop/issues/10557.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/fetch_env_var.rb
+++ b/lib/rubocop/cop/style/fetch_env_var.rb
@@ -69,9 +69,20 @@ module RuboCop
         # - Used as a flag (e.g., `if ENV['X']` or `!ENV['X']`) because
         #   it simply checks whether the variable is set.
         # - Receiving a message with dot syntax, e.g. `ENV['X'].nil?`.
-        # - `ENV['key']` is a receiver of `||=`, e.g. `ENV['X'] ||= y`.
+        # - `ENV['key']` assigned by logical AND/OR assignment.
         def allowable_use?(node)
-          used_as_flag?(node) || message_chained_with_dot?(node) || node.parent&.or_asgn_type?
+          used_as_flag?(node) || message_chained_with_dot?(node) || assigned?(node)
+        end
+
+        # The following are allowed cases:
+        #
+        # - `ENV['key']` is a receiver of `||=`, e.g. `ENV['X'] ||= y`.
+        # - `ENV['key']` is a receiver of `&&=`, e.g. `ENV['X'] &&= y`.
+        def assigned?(node)
+          return false unless (parent = node.parent)&.assignment?
+
+          lhs, _method, _rhs = *parent
+          node == lhs
         end
       end
     end
--- a/spec/rubocop/cop/style/fetch_env_var_spec.rb
+++ b/spec/rubocop/cop/style/fetch_env_var_spec.rb
@@ -86,9 +86,37 @@ RSpec.describe RuboCop::Cop::Style::FetchEnvVar, :config do
   end
 
   context 'when the node is a receiver of `||=`' do
-    it 'does not register an offense with `||`' do
+    it 'does not register an offense' do
       expect_no_offenses(<<~RUBY)
         ENV['X'] ||= y
+        x ||= ENV['X'] ||= y
+      RUBY
+    end
+  end
+
+  context 'when the node is a receiver of `&&=`' do
+    it 'does not register an offense' do
+      expect_no_offenses(<<~RUBY)
+        ENV['X'] &&= y
+        x &&= ENV['X'] ||= y
+      RUBY
+    end
+  end
+
+  context 'when the node is a assigned by `||=`' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        y ||= ENV['X']
+              ^^^^^^^^ Use `ENV.fetch('X')` or `ENV.fetch('X', nil)` instead of `ENV['X']`.
+      RUBY
+    end
+  end
+
+  context 'when the node is a assigned by `&&=`' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        y &&= ENV['X']
+              ^^^^^^^^ Use `ENV.fetch('X')` or `ENV.fetch('X', nil)` instead of `ENV['X']`.
       RUBY
     end
   end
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/fetch_env_var.rb
```diff
@@ -69,9 +69,22 @@ def message_chained_with_dot?(node)
...
+
+        # The following are allowed cases:
+        #
+        # - `ENV['key']` is a receiver of `||=`, e.g. `ENV['X'] ||= y`.
+        # - `ENV['key']` is a receiver of `&&=`, e.g. `ENV['X'] &&= y`.
+        def assigned?(node)
+          return false unless node.parent&.assignment?
+
+          node.parent.each_child_node.inject(true) do |last_not_env_node, child_node|
+            break true unless last_not_env_node
```
koic, line 84: The specs seems to pass without this `break`. What about the following logic?

```diff
diff --git a/lib/rubocop/cop/style/fetch_env_var.rb b/lib/rubocop/cop/style/fetch_env_var.rb
index 64f5b81d8..e9c5e375b 100644
--- a/lib/rubocop/cop/style/fetch_env_var.rb
+++ b/lib/rubocop/cop/style/fetch_env_var.rb
@@ -78,13 +78,11 @@ module RuboCop
         # - `ENV['key']` is a receiver of `||=`, e.g. `ENV['X'] ||= y`.
         # - `ENV['key']` is a receiver of `&&=`, e.g. `ENV['X'] &&= y`.
         def assigned?(node)
-          return false unless node.parent&.assignment?
+          return false unless (parent = node.parent)&.assignment?

-          node.parent.each_child_node.inject(true) do |last_not_env_node, child_node|
-            break true unless last_not_env_node
+          lhs, _method, _rhs = *parent

-            child_node != node
-          end
+          node == lhs
         end
       end
     end
```

deepj:
I have something like this in the code:

```
if Rails.env.production? && ENV['DISABLE_DOTENV'] != '1'
```

This is marked as (RuboCop 1.28.1)

```
C: [Correctable] Style/FetchEnvVar: Use ENV.fetch('DISABLE_DOTENV') or ENV.fetch('DISABLE_DOTENV', nil) instead of ENV['DISABLE_DOTENV'].
if Rails.env.production? && ENV['DISABLE_DOTENV'] != '1'
```

I guess this missing to be fixed in this PR, right?

lib/rubocop/cop/style/fetch_env_var.rb
ydah: Thank you so much! It is very simple.

ydah:
@deepj I think it looks a little different case than what this PR is trying to cover. Sorry.
From what I have seen of that code, it looks like `Style/FetchEnvVar` is what you want to detect.

ydah:
@koic Thank you for your review. I updated this PR.

lib/rubocop/cop/style/fetch_env_var.rb
```diff
@@ -69,9 +69,19 @@ def message_chained_with_dot?(node)
         # - Used as a flag (e.g., `if ENV['X']` or `!ENV['X']`) because
         #   it simply checks whether the variable is set.
         # - Receiving a message with dot syntax, e.g. `ENV['X'].nil?`.
-        # - `ENV['key']` is a receiver of `||=`, e.g. `ENV['X'] ||= y`.
```
koic, line 72: It looks better to replace the deletion with a simple description of the updated condition.

ydah: Thank you. I’ll write a brief description.

ydah:
I updated this PR. Thank you so much!

lib/rubocop/cop/style/fetch_env_var.rb
```diff
@@ -69,9 +69,20 @@ def message_chained_with_dot?(node)
...
+        # - `ENV['key']` assigned by  Logical AND/OR assignment.
```
koic, line 72: That's just my two cents :-)

```suggestion
        # - `ENV['key']` assigned by logical AND/OR assignment.
```

ydah: Sorry for spending your time on my trivial mistake...💦

spec/rubocop/cop/style/fetch_env_var_spec.rb
```diff
@@ -89,6 +89,34 @@
     it 'does not register an offense with `||`' do
       expect_no_offenses(<<~RUBY)
         ENV['X'] ||= y
+        x ||= ENV['X'] ||= y
+      RUBY
+    end
+  end
+
+  context 'when the node is a receiver of &&=`' do
```
koic, line 97:
```suggestion
  context 'when the node is a receiver of `&&=`' do
```

spec/rubocop/cop/style/fetch_env_var_spec.rb
```diff
@@ -89,6 +89,34 @@
...
+  context 'when the node is a assigned by `||=`' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        y ||= ENV['X']
+              ^^^^^^^^ Use `ENV.fetch('X')` or `ENV.fetch('X', nil)` instead of `ENV['X']`.
+      RUBY
+    end
+  end
+
+  context 'when the node is a assigned by &&=`' do
```
koic, line 115:
```suggestion
  context 'when the node is a assigned by `&&=`' do
```

spec/rubocop/cop/style/fetch_env_var_spec.rb
```diff
@@ -89,6 +89,34 @@
     it 'does not register an offense with `||`' do
       expect_no_offenses(<<~RUBY)
         ENV['X'] ||= y
+        x ||= ENV['X'] ||= y
+      RUBY
+    end
+  end
+
+  context 'when the node is a receiver of `&&=`' do
+    it 'does not register an offense with `&&`' do
```
koic, line 98: It may be redundant (The same is true for L89).

```suggestion
    it 'does not register an offense' do
```

ydah:
@koic Sorry for taking up your valuable time.

koic:
Never mind! Thank you, too!

PR #10571, opened 2022-04-22
Add AllowedPatterns to Naming/VariableNumber,VariableName

### LINKED ISSUE #10539: Allow _v1 suffix in Naming/VariableNumber without having to specify every individual method

henrahmagix:
Fix #10539, redo of #10540

Also fix global variables not being matched correctly, discovered by
filling in the tests with global variable examples and having failures.

### DIFF
--- a/changelog/new_add_new_option_to_namingvariablenumber.md
+++ b/changelog/new_add_new_option_to_namingvariablenumber.md
@@ -1,0 +1,1 @@
+* [#10539](https://github.com/rubocop/rubocop/issues/10539): Add AllowedPatterns option to Naming/VariableNumber and Naming/VariableName. ([@henrahmagix][])
--- a/changelog/new_add_new_option_to_namingvariablenumber.md
+++ b/changelog/new_add_new_option_to_namingvariablenumber.md
@@ -1,1 +1,1 @@
-* [#10539](https://github.com/rubocop/rubocop/issues/10539): Add AllowedPatterns option to Naming/VariableNumber and Naming/VariableName. ([@henrahmagix][])
+* [#10539](https://github.com/rubocop/rubocop/issues/10539): Add `AllowedPatterns` configuration option to `Naming/VariableNumber` and `Naming/VariableName`. ([@henrahmagix][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2772,6 +2772,7 @@ Naming/VariableName:
     - snake_case
     - camelCase
   AllowedIdentifiers: []
+  AllowedPatterns: []
 
 Naming/VariableNumber:
   Description: 'Use the configured style when numbering symbols, methods and variables.'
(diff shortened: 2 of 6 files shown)


changelog/new_add_new_option_to_namingvariablenumber.md
```diff
@@ -0,0 +1 @@
+* [#10539](https://github.com/rubocop/rubocop/issues/10539): Add AllowedPatterns option to Naming/VariableNumber and Naming/VariableName. ([@henrahmagix][])
```
dvandersluis, line 1:
```suggestion
* [#10539](https://github.com/rubocop/rubocop/issues/10539): Add `AllowedPatterns` configuration option to `Naming/VariableNumber` and `Naming/VariableName`. ([@henrahmagix][])
```

dvandersluis (commented):
Other than one my suggestion this looks good now! Thanks!

dvandersluis:
Please squash your commits.

bbatsov:
Looks good to me. Thanks!

henrahmagix:
Thanks all!

PR #10572, opened 2022-04-23
[Fix #10570] Add new `Gemspec/DependencyVersion` cop

### LINKED ISSUE #10570: Feature idea: `Gemspec/GemVersion`

There is already a `Bundler/GemVersion` cop that enforces gems in the Gemfile to include a version specification.
But this cop doesn't work for gemspec files.

I would like either `Bundler/GemVersion`  to also work for gemspec files or to have a new `Gemspec/GemVersion` cop that would enforce version specifications in gemspec files.

## Examples

```
spec.add_dependency 'redis' # bad
spec.add_dependency 'redis', '~> 4.5.1' # good

spec.add_development_dependency 'faker' # bad
spec.add_development_dependency 'faker', '~> 2.19.0' # good
```

nobuyo:
Resolves #10570.

This cop requires/forbids version specifications or a commit reference for gem dependency in gemspec.

The name for cop suggested in the issue was `GemVersion`, but I'm renaming it because cop is interested in the version of the dependency, not the gem owning gemspec.

## example

```ruby
# EnforcedStyle: required (default)

# bad
  spec.add_dependency 'rubocop'
end

# good
  spec.add_dependency 'rubocop', '~> 1.28'
end
```

```ruby
# EnforcedStyle: forbidden

# bad
  spec.add_dependency 'rubocop', '~> 1.28'
end

# good
  spec.add_dependency 'rubocop'
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -242,6 +242,18 @@ Gemspec/DateAssignment:
   Include:
     - '**/*.gemspec'
 
+Gemspec/DependencyVersion:
+  Description: 'Requires or forbids specifying gem dependency versions.'
+  Enabled: false
+  VersionAdded: '<<next>>'
+  EnforcedStyle: 'required'
+  SupportedStyles:
+    - 'required'
+    - 'forbidden'
+  Include:
+    - '**/*.gemspec'
+  AllowedGems: []
+
 Gemspec/DuplicatedAssignment:
   Description: 'An attribute assignment method calls should be listed only once in a gemspec.'
   Enabled: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -149,3 +149,6 @@ InternalAffairs/UndefinedConfig:
 InternalAffairs/StyleDetectedApiUse:
   Exclude:
     - 'lib/rubocop/cop/mixin/percent_array.rb'
+
+Gemspec/DependencyVersion:
+  Enabled: true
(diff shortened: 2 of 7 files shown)


config/default.yml
```diff
@@ -242,6 +242,18 @@ Gemspec/DateAssignment:
   Include:
     - '**/*.gemspec'
 
+Gemspec/DependencyVersion:
+  Description: 'Requires or forbids specifying gem dependency versions.'
+  Enabled: pending
```
koic, line 247: There is a similar argument in #7669, I think it should be disabled by default.

```suggestion
  Enabled: false
```

nobuyo: Thanks, I think .rubocop.yml should be revert back together too, so I'll fix this directly on my local branch.

bbatsov:
I think you can enable this for RuboCop itself. I'm a fan of explicit versions everywhere.

bbatsov:
Thanks!

koic:
> Althrough a lot of guides mention that you should not store Gemfile.lock in gem repos

I also have this idea. Unlike applications, the `runtime_dependency` of gemspec affects user's library-dependent version management. And I think SCM should not store gem repo's Gemfile.lock. In my maintenance experience, bot updates haven't helped much. Well, it may depend on the developer :-)

PR #10575, opened 2022-04-23
[Fix #10573] Fix a false positive for `Layout/SpaceBeforeBrackets`

### LINKED ISSUE #10573: `Layout/SpaceBeforeBrackets`: false positive for arguments forwarding to `Hash#[]`
## Expected behavior

It does not detect offense of `Layout/SpaceBeforeBrackets`.

## Actual behavior

```
$ rubocop --debug --cache false --only Layout/SpaceBeforeBrackets foo.rb
For /Users/yuji_developer/foo: configuration from /Users/yuji_developer/foo/.rubocop.yml
Default configuration from /Users/yuji_developer/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/rubocop-1.28.1/config/default.yml
Inspecting 1 file
Scanning /Users/yuji_developer/foo/foo.rb
C

Offenses:

foo.rb:7:10: C: [Correctable] Layout/SpaceBeforeBrackets: Remove the space before the opening brackets.
    @hash.[](...)
         ^

Finished in 0.06047500018030405 seconds
```

Run `rubocop --only Layout/SpaceBeforeBrackets foo.rb` on the following config and code.

```yaml
# .rubocop.yml
AllCops:
  TargetRubyVersion: 2.7
  NewCops: disable
```
```ruby
# foo.rb
class Foo
  def initialize
    @hash = {}
  end

  def [](...)
    @hash.[](...)
  end
end
```

## RuboCop version

```
$ rubocop -V
1.28.1 (using Parser 3.1.2.0, rubocop-ast 1.17.0, running on ruby 2.7.6 arm64-darwin21)
```

nobuyo:
…when using desugared `Hash#[]` method.

Fixes #10573.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/space_before_brackets.rb
+++ b/lib/rubocop/cop/layout/space_before_brackets.rb
@@ -37,11 +37,18 @@ module RuboCop
             receiver_end_pos = node.receiver.source_range.end_pos
             selector_begin_pos = node.loc.selector.begin_pos
             return if receiver_end_pos >= selector_begin_pos
+            return if dot_before_brackets?(node, receiver_end_pos, selector_begin_pos)
 
             range_between(receiver_end_pos, selector_begin_pos)
           elsif node.method?(:[]=)
             offense_range_for_assignment(node, begin_pos)
           end
+        end
+
+        def dot_before_brackets?(node, receiver_end_pos, selector_begin_pos)
+          return false unless node.loc.respond_to?(:dot) && (dot = node.loc.dot)
+
+          dot.begin_pos == receiver_end_pos && dot.end_pos == selector_begin_pos
         end
 
         def offense_range_for_assignment(node, begin_pos)
--- a/changelog/fix_a_false_positive_for_layout_space_before_brackets.md
+++ b/changelog/fix_a_false_positive_for_layout_space_before_brackets.md
@@ -1,0 +1,1 @@
+* [#10573](https://github.com/rubocop/rubocop/issues/10573): Fix a false positive for `Layout/SpaceBeforeBrackets` when there is a dot before brackets. ([@nobuyo][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/layout/space_before_brackets.rb
```diff
@@ -37,13 +37,20 @@ def offense_range(node, begin_pos)
...
             return if receiver_end_pos >= selector_begin_pos
+            return if brackets_without_syntax_sugar?(node, receiver_end_pos, selector_begin_pos)
 
             range_between(receiver_end_pos, selector_begin_pos)
           elsif node.method?(:[]=)
             offense_range_for_assignment(node, begin_pos)
           end
         end
 
+        def brackets_without_syntax_sugar?(node, receiver_end_pos, selector_begin_pos)
```
koic, line 48: Does this mean something like this?

```suggestion
        def dot_before_brackets?(node, receiver_end_pos, selector_begin_pos)
```

nobuyo: Exactly, that was a littie weired naming :)

lib/rubocop/cop/layout/space_before_brackets.rb
```diff
@@ -37,13 +37,20 @@ def offense_range(node, begin_pos)
...
             range_between(receiver_end_pos, selector_begin_pos)
           elsif node.method?(:[]=)
             offense_range_for_assignment(node, begin_pos)
           end
         end
 
+        def dot_before_brackets?(node, receiver_end_pos, selector_begin_pos)
+          return false unless node.loc.respond_to?(:dot)
+
+          node.loc.dot&.begin_pos == receiver_end_pos && node.loc.dot&.end_pos == selector_begin_pos
```
koic, line 51: I'm not sure about the best, but the following can reduce method calls and eliminate safe navigation.

```suggestion
          return false unless node.loc.respond_to?(:dot) && (dot = node.loc.dot)

          dot.begin_pos == receiver_end_pos && dot.end_pos == selector_begin_pos
```

nobuyo: Thank you, I've committed your suggestion.

koic:
Thanks!

PR #10582, opened 2022-04-26
[Fix #10581] Fix a false positive for `Style/FetchEnvVar`

### LINKED ISSUE #10581: Style/FetchEnvVar - false positive cases with (not) equal operators
## Expected behavior

No issue while using ENVs and (not) equal operators

```ruby
if ENV['DISABLE_DOTENV'] != '1'
  # Some Ruby code
end
```

## Actual behavior

```
Style/FetchEnvVar: Use ENV.fetch('DISABLE_DOTENV') or ENV.fetch('DISABLE_DOTENV', nil) instead of ENV['DISABLE_DOTENV'].
if ENV['DISABLE_DOTENV'] != '1'
   ^^^^^^^^^^^^^^^^^^^^^
```

Code snippet

```ruby
if ENV['DISABLE_DOTENV'] != '1'
  # Some Ruby code
end
```

Run

```
rubocop --only Style/FetchEnvVar
```

## RuboCop version

```
$ rubocop -V
1.28.2 (using Parser 3.1.2.0, rubocop-ast 1.17.0, running on ruby 3.1.2 arm64-darwin21)
  - rubocop-performance 1.13.3
  - rubocop-rails 2.14.2
```

koic:
Fixes #10581.

This PR fixes a false positive for `Style/FetchEnvVar` when comparing with `ENV['TERM']`.
`ENV['DISABLE_DOTENV'] != '1'` could be treated the same as the already allowed in `used_as_flag?` (that is, used as a flag).

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_style_fetch_env_var_cop.md
+++ b/changelog/fix_a_false_positive_for_style_fetch_env_var_cop.md
@@ -1,0 +1,1 @@
+* [#10581](https://github.com/rubocop/rubocop/issues/10581): Fix a false positive for `Style/FetchEnvVar` when comparing with `ENV['TERM']`. ([@koic][])
--- a/lib/rubocop/cop/style/fetch_env_var.rb
+++ b/lib/rubocop/cop/style/fetch_env_var.rb
@@ -53,8 +53,9 @@ module RuboCop
 
         def used_as_flag?(node)
           return false if node.root?
+          return true if node.parent.if_type?
 
-          node.parent.if_type? || (node.parent.send_type? && node.parent.prefix_bang?)
+          node.parent.send_type? && (node.parent.prefix_bang? || node.parent.comparison_method?)
         end
 
         # Check if the node is a receiver and receives a message with dot syntax.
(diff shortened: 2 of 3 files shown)


johnny-miyake:
I think there can be a case like the below where it isn't a flag use. (Not sure if this is a good example, though).
What do you think?
```ruby
if ENV['SERVER_ROLE'] == 'mailing'
  # Disable some features that must not be used in 'mailing' role.
else
  # Expect the development environment
end
```

koic:
Hm, it may be possible to allow comparisons only to compare with commonly used values as flags. e.g. `'1'`, `'true'`. Thought?

johnny-miyake:
Yes, that would be a possible option, or we could suggest these allowable use.
```ruby
if ENV['DISABLE_DOTENV'].nil?
  ...
end
```

```ruby
unless ENV['DISABLE_DOTENV']
  ...
end
```

johnny-miyake:
But it is obvious that the developer knows how the logic branches by the variable in the condition, so detecting the code like my example as an offense may not make sense.

koic:
> it is obvious that the developer knows how the logic branches by the variable in the condition

Yeah, probably so. I'm going to merge this PR. Thank you for your feedback.

Issue #10578, opened 2022-04-24
Cop idea:  Set the default value of `ENV.fetch` by autocorrect


johnny-miyake:
## Is your feature request related to a problem? Please describe.

From: https://github.com/rubocop/rubocop/pull/10502#issuecomment-1106950555

With `Style/FetchEnvVar`, code like
```ruby
value = ENV['EXAMPLE_KEY'] || 'default_value'
```

gets autocorrected to
```ruby
value = ENV.fetch('EXAMPLE_KEY', nil) || 'default_value'
```

Autocorrecting like below helps people reduce their manual changings like above.
```ruby
value = ENV.fetch('EXAMPLE_KEY', 'default_value')

# This version would be better on performance if the default value is calculated by an expensive computation.
value = ENV.fetch('EXAMPLE_KEY'){'default_value'}
```

## Describe the solution you'd like

- When `ENV[]` is the left-hand of `||`, the autocorrect can move the right-hand of `||` to the default value of `ENV.fetch`.
- In some cases, the right hand of `||` can't be the default value (e.g., `next`), so we need to carefully handle them.

bbatsov:
Seems like a good idea to me. 

As for the auto-correct - I guess if the default is some literal or a local value we can use the first form and if it's a method call we can go for the block style, although I don't have a strong preference. The performance gains would be notable only if the computation of the default value is expensive.

ShockwaveNN:
Please note that if you got code like this:

```ruby
return ENV['foo'] if ENV.key?('foo')

File.read('file_if_foo_not_exists')
```
Code like this will work fine if `ENV['foo']` exists and File `file_if_foo_not_exists` not exists

But if you by similarity with proposal from this issue change it to

```ruby
ENV.fetch('foo', File.read('file_if_foo_not_exists'))
```

If will fail in any case if `file_if_foo_not_exists` do not exists (even if `ENV['foo']` is initialized)

johnny-miyake:
Great point. That would be one of the cases we should use the block style.
```ruby
ENV.fetch('foo') { File.read('file_if_foo_not_exists') }
```
In my opinion, in most cases always using the block style is safer and reduces the rules. Any thoughts?

johnny-miyake:
I created a PR for this.
https://github.com/rubocop/rubocop/pull/10585
I made the default value style configurable so that users can choose their preferred way.
Any help in finding missing test cases or edge cases is very appreciated.

PR #10585, opened 2022-04-28
Enhance the autocorrect for `Style/FetchEnvVar`

johnny-miyake:
https://github.com/rubocop/rubocop/issues/10578

This PR enhances the autocorrect for `Style/FetchEnvVar`.

When an `ENV[]` is the LHS of `||`, the autocorrect makes the RHS the default value of `ENV.fetch`.

Examples

```ruby
# from
ENV['X'] || 'y'

# to
ENV.fetch('X', 'y')
```

```ruby
# from
ENV['X'] || ENV['Y'] || a

# to
ENV.fetch('X') { ENV.fetch('Y') { a } }
```

```ruby
# from
z || ENV['X'] || y || ENV['Z'] || a || ENV['B']

# to
z || ENV.fetch('X') { y } || ENV.fetch('Z') { a } || ENV.fetch('B', nil)
```

```ruby
# from
ENV['X'] || y.map do |a|
  a * 2
end

# to
ENV.fetch('X') do
  y.map do |a|
    a * 2
  end
end
```

Please refer to the RSpec file for more examples.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_enhance_the_autocorrect_for_fetch_env_var.md
+++ b/changelog/change_enhance_the_autocorrect_for_fetch_env_var.md
@@ -1,0 +1,1 @@
+* [#10585](https://github.com/rubocop/rubocop/pull/10585): Enhance the autocorrect for `RSpec/FetchEnvVar`. ([@johnny-miyake][])
--- a/changelog/change_enhance_the_autocorrect_for_fetch_env_var.md
+++ b/changelog/change_enhance_the_autocorrect_for_fetch_env_var.md
@@ -1,1 +1,1 @@
-* [#10585](https://github.com/rubocop/rubocop/pull/10585): Enhance the autocorrect for `RSpec/FetchEnvVar`. ([@johnny-miyake][])
+* [#10585](https://github.com/rubocop/rubocop/pull/10585): Enhance the autocorrect for `Style/FetchEnvVar`. ([@johnny-miyake][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -3538,6 +3538,9 @@ Style/FetchEnvVar:
   VersionAdded: '1.28'
   # Environment variables to be excluded from the inspection.
   AllowedVars: []
+  # - `SecondArgOfFetch` puts the RHS to the second argument of `ENV.fetch`. (default)
+  # - `BlockContent` puts the RHS into a block.
+  BasicLiteralRhs: 'SecondArgOfFetch'
 
 Style/FileRead:
   Description: 'Favor `File.(bin)read` convenience methods.'
--- a/config/default.yml
+++ b/config/default.yml
@@ -3538,9 +3538,6 @@ Style/FetchEnvVar:
   VersionAdded: '1.28'
   # Environment variables to be excluded from the inspection.
   AllowedVars: []
-  # - `SecondArgOfFetch` puts the RHS to the second argument of `ENV.fetch`. (default)
-  # - `BlockContent` puts the RHS into a block.
-  BasicLiteralRhs: 'SecondArgOfFetch'
 
 Style/FileRead:
   Description: 'Favor `File.(bin)read` convenience methods.'
--- a/spec/rubocop/cop/style/fetch_env_var_spec.rb
+++ b/spec/rubocop/cop/style/fetch_env_var_spec.rb
@@ -62,37 +62,6 @@ RSpec.describe RuboCop::Cop::Style::FetchEnvVar, :config do
     end
   end
 
-  context 'when the node is an operand of `||`' do
-    it 'registers no offenses with `||`' do
-      expect_offense(<<~RUBY)
-        ENV['X'] || y
-        ^^^^^^^^ Use `ENV.fetch('X')` or `ENV.fetch('X', nil)` instead of `ENV['X']`.
-      RUBY
-
-      expect_correction(<<~RUBY)
-        ENV.fetch('X', nil) || y
-      RUBY
-
-      expect_offense(<<~RUBY)
-        y || ENV['X']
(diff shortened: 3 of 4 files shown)


changelog/change_enhance_the_autocorrect_for_fetch_env_var.md
```diff
@@ -0,0 +1 @@
+* [#10585](https://github.com/rubocop/rubocop/pull/10585): Enhance the autocorrect for `RSpec/FetchEnvVar`. ([@johnny-miyake][])
```
bbatsov, line 1: RSpec? :-)

config/default.yml
```diff
@@ -3538,6 +3538,9 @@ Style/FetchEnvVar:
   VersionAdded: '1.28'
   # Environment variables to be excluded from the inspection.
   AllowedVars: []
+  # - `SecondArgOfFetch` puts the RHS to the second argument of `ENV.fetch`. (default)
+  # - `BlockContent` puts the RHS into a block.
+  BasicLiteralRhs: 'SecondArgOfFetch'
```
bbatsov, line 3543: Seems to me that adds extra complexity for the end users - I generally don't see why someone would want to put basic literals in a block, as there are no performance gains for this.

changelog/change_enhance_the_autocorrect_for_fetch_env_var.md
johnny-miyake: Oops, thanks!

spec/rubocop/cop/style/fetch_env_var_spec.rb
```diff
@@ -336,4 +304,343 @@
...
+              a * 2
+            end
+          RUBY
+
+          expect_correction(<<~RUBY)
+            ENV.fetch('X') do
+              y.map do |a|
+                a * 2
+              end
+            end
```
johnny-miyake, line 621: Autocorrect doesn't indent properly when `ENV` is indented.

```ruby
# With no indentation
ENV['X'] || y.map do |a|
  a * 2
end

ENV.fetch('X') do
  y.map do |a|
    a * 2
  end
end

# With an indentation
def foo
  ENV['X'] || y.map do |a|
    a * 2
  end
end

def foo
  ENV.fetch('X') do
    y.map do |a|
        a * 2
      end
end
end
```

johnny-miyake: Fixed: https://github.com/rubocop/rubocop/pull/10585/commits/214f6f0b5e5cafa148aa4426e5ecba2a9d504ab9

config/default.yml
johnny-miyake: That's true. I removed the option. https://github.com/rubocop/rubocop/pull/10585/commits/37979149a76054a4879fc8121f74ab92c33bc1a3

johnny-miyake:
I added some additional test cases and fixed the code to make them pass.

koic:
As I noticed in this PR, I think most `ENV['TERM'] || something` should be allowed. Rewriting this to `ENV.fetch('TERM', something)` or `ENV.fetch('TERM') { something }` has no of the following benefits:

```
      # This cop suggests `ENV.fetch` for the replacement of `ENV[]`.
      # `ENV[]` silently fails and returns `nil` when the environment variable is unset,
      # which may cause unexpected behaviors when the developer forgets to set it.
      # On the other hand, `ENV.fetch` raises KeyError or returns the explicitly
      # specified default value.
```

johnny-miyake:
Yeah, given `ENV['TERM'] || something` is allowed in https://github.com/rubocop/rubocop/pull/10582 , this PR may be no longer needed.

johnny-miyake:
> Rewriting this to ENV.fetch('TERM', something) or ENV.fetch('TERM') { something } does not have much of the following benefits:

Yeah, that sounds correct. This problem somehow looks silmilar to https://github.com/rubocop/rubocop/pull/10582 .

bbatsov:
> As I noticed in this PR, I think most ENV['TERM'] || something should be allowed. Rewriting this to ENV.fetch('TERM', something) or ENV.fetch('TERM') { something } does not have much of the following benefits:

Still, to me this seems like the type of thing that can be optimized as the `||` check is essentially the same and hash has this API exactly to handle this common scenario (default value for a missing key). I do agree, however, that the optimization is not critical, so perhaps we should just have a config for the cop wether to detect and fix usages of `||` and `or` and to fix them. I, myself, would be fine if we don't and we just flag and fix all of them, as I prefer `fetch` + default value all of the time.  Am I missing some case in which using `||` is actually more useful?

johnny-miyake:
> Rewriting this to ENV.fetch('TERM', something) or ENV.fetch('TERM') { something } does not have much of the following benefits:
>
> User can get the KeyError, but the fact that the default value is used would mean that it supplements that condition.

I understand the above that both codes below are read as “Get 'TERM' environment variable if it's set, otherwise, get the return value of `something`", so correcting the first one doesn't make any changes to the behavior when the environment variable isn't set unexpectedly, plus,  that diverges from what this cop is originally addressing, correct?

```ruby
ENV['TERM'] || something

ENV.fetch('TERM', something)
```

I think using an option is a way, while there’s another way that we can see how it goes allowing `ENV[]` with `||` for now to move forward little by little, although I’m not sure which is the best for users.

bbatsov:
It seems to me that this is what @koic meant, but he has to clarify himself. We can always tackle this separately in a different cop, although perhaps that'd be an overkill.

koic:
> the `||` check is essentially the same and hash has this API exactly to handle this common scenario (default value for a missing key). 

I'm sorry, I was misunderstanding 💦  The above comment makes sense. So, there is no problem correcting it with compatible API.

PR #10593, opened 2022-04-30
[Fix #10325] Enhance `Style/RedundantCondition` for the case that variable assignments in each branch

### LINKED ISSUE #10325: New cop to replace conditional assignment with or-assignment
I've written the following code
```ruby
def lock_retrier=(value)
  if value.nil?
    @lock_retrier = NullLockRetrier.new
  else
    @lock_retrier = value
  end
end
```

But then realized that this should be really written as
```ruby
def lock_retrier=(value)
  @lock_retrier = value || NullLockRetrier.new
end
```

and wondering why rubocop hasn't suggested such change.

Quick search on the rails codebase via regexp revealed this: 
https://github.com/rails/rails/blob/296ef7a17221e81881e38b51aa2b014d7a28bac5/actionpack/lib/action_dispatch/journey/gtg/transition_table.rb#L75-L79

So I suggest implementing a new cop (not sure about its name).

This should also check for cases like
```ruby
if value
  @lock_retrier = value
else
  @lock_retrier =  NullLockRetrier.new
end
```

etc.

nobuyo:
Fixes #10325.

The example using `nil?` given in the issue is not supported here, as it would destroy the code whose intent is to execute else_branch when `value  == false`.

```ruby
# not supported
def lock_retrier=(value)
  if value.nil?
    @lock_retrier = NullLockRetrier.new
  else
    @lock_retrier = value
  end
end

# supported
def lock_retrier=(value)
  if value
    @lock_retrier = NullLockRetrier.new
  else
    @lock_retrier = value
  end
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_enhance_style_redundantcondition.md
+++ b/changelog/fix_enhance_style_redundantcondition.md
@@ -1,0 +1,1 @@
+* [#10325](https://github.com/rubocop/rubocop/issues/10325): Enhance `Style/RedundantCondition` by considering the case that variable assignments in each branch. ([@nobuyo][])
--- a/lib/rubocop/cop/style/redundant_condition.rb
+++ b/lib/rubocop/cop/style/redundant_condition.rb
@@ -73,13 +73,12 @@ module RuboCop
         end
 
         def offense?(node)
-          condition, if_branch, else_branch = *node
+          _condition, _if_branch, else_branch = *node
 
           return false if use_if_branch?(else_branch) || use_hash_key_assignment?(else_branch)
 
-          condition == if_branch && !node.elsif? && (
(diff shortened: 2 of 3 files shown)


nobuyo:
Should the following patterns be covered as well?

```rb
if foo
  test.bar foo
else
  test.bar 'baz'
end
# => test.bar foo || 'baz'

if foo
  test.bar = foo
else
  test.bar = 'baz'
end
# => test.bar = foo || 'baz'
```

nobuyo:
@rubocop/rubocop-core I pushed af27f01 with the split as I dared. If you give me your thoughts, I'd like to squash or revert it.

bbatsov:
The changes look reasonable to me. Thanks!

PR #10596, opened 2022-05-02
Remove an old reference to version 1.0

skanev:
Version 1.0 is already out, so this note feels unnecessary.

I just edited this on GitHub. Apologies for the poor branch name, let me know if you need a better one :)

### DIFF
--- a/docs/modules/ROOT/pages/installation.adoc
+++ b/docs/modules/ROOT/pages/installation.adoc
@@ -25,9 +25,6 @@ in your `Gemfile`:
 gem 'rubocop', '~> 1.28', require: false
 ----
 
-NOTE: You can check out our progress on the road to version 1.0 https://github.com/rubocop/rubocop/milestone/4[here].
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/installation.adoc
```diff
@@ -25,9 +25,6 @@ in your `Gemfile`:
 gem 'rubocop', '~> 1.28', require: false
 ----
 
-NOTE: You can check out our progress on the road to version 1.0 https://github.com/rubocop/rubocop/milestone/4[here].
-You can also help us get there faster! :-)
```
koic, line 29: Good catch! Can you replace it with the following instead of deleting it?

```
See xref:versioning.adoc[our versioning policy] for further details.
```

cf: https://github.com/rubocop/rubocop/commit/0bb3303d662c0d456f0d81033ac196bb6be5e309

bbatsov:
Thanks for tackling this!

PR #10598, opened 2022-05-02
[Fix #10568] Add new `Style/EnvHome` cop

### LINKED ISSUE #10568: Cop idea: Replace `ENV['HOME']` with `Dir.home`

While fixing issues from latest rubocop update with `Style/FetchEnvVar` cop I found a lot of placed in my code where construction like this used:

```ruby
File.read("#{ENV['HOME']}/foo.file"
```
I got two problems:

1. Autofix from `Style/FetchEnvVar` advice to use
    ```ruby
    File.read("#{ENV.fetch('HOME'), nil]}/foo.file"
    ```
    Which seem rather lengthy for me and do not solve any real problems
2. Not sure that my original code ever worked on `Windows` system (but cannot check it)

Cop that replace all `ENV['HOME']` to `Dir.home` wich will be pure ruby solution to get home dir
But maybe there is some downside in this, but I'm not aware of such

koic:
Fixes #10568.

This cop checks for consistent usage of `ENV['HOME']`. If `nil` is used as the second argument of `ENV.fetch`, it is treated as a bad case like `ENV[]`. This avoids conflicts with `Style/FetchEnvVar`'s autocorrection.

## Safety

The cop is unsafe because The result when `nil` is assigned to `ENV['HOME']` changes:

```ruby
ENV['HOME'] = nil
ENV['HOME'] # => nil
Dir.home    # => '/home/foo'
```

## Example

```ruby
# bad
ENV['HOME']
ENV.fetch('HOME', nil)

# good
Dir.home

# good
ENV.fetch('HOME', default)
```

-----------------

* [x] Added tests.


bbatsov:
Thanks!

ShockwaveNN:
Thanks for implementing this ) Glad my idea was usefull

PR #10603, opened 2022-05-06
[Fix #10569] Fix a false positive for `Style/FetchEnvVar`

### LINKED ISSUE #10569: `Style/FetchEnvVar` check for `if` condition
Not sure that `rubocop` can check it or dynamic check is required for such checks, but I'll report it anyway

1. Create `test.rb` with content:
```ruby
# frozen_string_literal: true

if ENV['foo']
  puts 'some code'
  puts ENV['foo']
  puts 'other code'
end
```

2. run `rubocop -A` check
result code will be:
```ruby
# frozen_string_literal: true
    
if ENV['foo']
  puts 'some code'
  puts ENV.fetch('foo', nil)
  puts 'other code'
end
```

So fetch is literally useless in code like this

## RuboCop version

```
$ [bundle exec] rubocop -V
1.28.1 (using Parser 3.1.2.0, rubocop-ast 1.17.0, running on ruby 3.1.2 x86_64-linux)
  - rubocop-performance 1.13.3
```

koic:
Fixes #10569.

This PR fixes a false positive for `Style/FetchEnvVar` when using the same value of `ENV` as `if` condition in the body.

The `ENV` var as a `if` condition is currently allowed.
So the `ENV` var already used in the condition will be accepted as well.

And the following (maybe) corner case is accepted in this PR.

```ruby
if ENV['foo']
  ENV.delete('foo')
  puts 'some code'
  puts ENV['foo']
  puts 'other code'
end
```

https://github.com/rubocop/rubocop/issues/10569#issuecomment-1106213970

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_for_style_fetch_env_var.md
+++ b/changelog/fix_a_false_for_style_fetch_env_var.md
@@ -1,0 +1,1 @@
+* [#10569](https://github.com/rubocop/rubocop/issues/10569): Fix a false positive for `Style/FetchEnvVar` when using the same `ENV` var as `if` condition in the body. ([@koic][])
--- a/lib/rubocop/cop/style/fetch_env_var.rb
+++ b/lib/rubocop/cop/style/fetch_env_var.rb
@@ -97,7 +97,9 @@ module RuboCop
 
         def used_as_flag?(node)
           return false if node.root?
-          return true if node.parent.if_type?
+
+          if_node = node.ancestors.find(&:if_type?)
+          return true if if_node&.condition == node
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

inkstak:
Yes, that may be another issue : the false positive is in the `if` statement and not the `if` block.

```
# if condition || ENV["FOO"]
#                 ^^^^^^^^^
```

It's definitely because the `condition` might be truthy I used the fetch method in `puts ENV.fetch("FOO")`

PR #10606, opened 2022-05-07
[Fix #10605] Fix autocorrect for `Style/RedundantCondition`

### LINKED ISSUE #10605: Style/RedundantCondition making invalid corrections
## Expected behavior

Running `rubocop --auto-correct` should not result in broken code.

## Actual behavior

Running rubocop 1.29.0 triggers a correction from `Style/RedundantCondition` on this code:

```
if referer
  redirect_to referer
else
  redirect_to :action => :inbox
end
```

changing it to this:

```
redirect_to referer || :action => :inbox
```

which is incorrect and is rejected by ruby requiring it to be changed to:

```
redirect_to referer || { :action => :inbox }
```

in order to work.

See above.

## RuboCop version

```
1.29.0 (using Parser 3.1.2.0, rubocop-ast 1.17.0, running on ruby 3.1.2 x86_64-linux)
  - rubocop-minitest 0.19.1
  - rubocop-performance 1.13.3
  - rubocop-rails 2.14.2
  - rubocop-rake 0.6.0
```

nobuyo:
...if argument for method in else branch is hash without braces.

Fixes #10605.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_condition.rb
+++ b/lib/rubocop/cop/style/redundant_condition.rb
@@ -157,6 +157,8 @@ module RuboCop
         def else_source_if_has_method(else_branch)
           if require_parentheses?(else_branch.first_argument)
             "(#{else_branch.first_argument.source})"
+          elsif require_braces?(else_branch.first_argument)
+            "{ #{else_branch.first_argument.source} }"
           else
             else_branch.first_argument.source
           end
@@ -165,6 +167,8 @@ module RuboCop
         def else_source_if_has_assignment(else_branch)
           if require_parentheses?(else_branch.expression)
             "(#{else_branch.expression.source})"
+          elsif require_braces?(else_branch.expression)
+            "{ #{else_branch.expression.source} }"
           else
             else_branch.expression.source
           end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/redundant_condition.rb
```diff
@@ -196,6 +200,10 @@ def require_parentheses?(node)
             (node.respond_to?(:semantic_operator?) && node.semantic_operator?)
         end
 
+        def require_braces?(node)
+          node.hash_type?
```
koic, line 204: It will get a syntax error in the following already braced case:

```ruby
           if foo
           ^^^^^^ Use double pipes `||` instead.
             bar foo
           else
             bar({ baz => quux })
           end
```

Auto-corrected code is `bar foo || { { baz => quux } }`.

```ruby
% ruby -ce "bar foo || { { baz => quux } }"
-e:1: syntax error, unexpected '}', expecting =>
bar foo || { { baz => quux } }
```

Can you add the above test case and fix it?

nobuyo: Thank you, fixed in 97c6273.

spec/rubocop/cop/style/redundant_condition_spec.rb
```diff
@@ -292,6 +292,36 @@ def test
         RUBY
       end
 
+      it 'registers an offense and corrects when the branches contains method call with hash' do
```
koic, line 295: That's just my two cents. It will be symmetric with the `non-braced` of L310 :-)

```suggestion
      it 'registers an offense and corrects when the branches contains method call with braced hash' do
```

koic:
Thanks!

PR #10608, opened 2022-05-07
Add new `Style/MapCompactWithConditionalBlock` cop

nobuyo:
This cop checks for usage of `select` or `reject`
instead of `map { ... }.compact`.

```ruby
# bad
array.map { |e| some_condition? ? e : next }.compact

# bad
array.map do |e|
  if some_condition?
    e
  else
    next
  end
end.compact

# bad
array.map do |e|
  next if some_condition?

  e
end.compact

# bad
array.map do |e|
  e if some_condition?
end.compact

# good
array.select { |e| some_condition? }

# good
array.reject { |e| some_condition? }
```

-----------------

* [x] Added tests.


lib/rubocop/cop/style/redundant_map_compact.rb
```diff
@@ -0,0 +1,137 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # This cop checks for usage of `select` or `reject`
```
koic, line 6: When a block has `if` condition, it cannot be replaced with `reject`.

```suggestion
      # This cop checks for usage of `select`
```

Instead, it could add text to the description, such as "replace with `reject` when the block has `unless` condition".

lib/rubocop/cop/style/redundant_map_compact.rb
```diff
@@ -0,0 +1,137 @@
...
+      #   # bad
+      #   array.map do |e|
+      #     e if some_condition?
+      #   end.compact
+      #
+      #   # good
+      #   array.select { |e| some_condition? }
+      #
+      #   # good
+      #   array.reject { |e| some_condition? }
```
koic, line 39: It is compatible when the block has `unless` condition, but bad examples are only `if` condition. Maybe it's better to delete it.

lib/rubocop/cop/style/redundant_map_compact.rb
```diff
@@ -0,0 +1,137 @@
...
+      #
+      class RedundantMapCompact < Base
```
koic, line 41: I have a feeling of strangeness in the name `Redundant`.  ``Style/MapCompactWithConditionalBlock`` (or `Style/MapWithConditionalBlock`) may be close to the behavior of this cop. Of course there seems to be a better name.

spec/rubocop/cop/style/redundant_map_compact_spec.rb
```diff
@@ -0,0 +1,389 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Style::RedundantMapCompact, :config do
+  context 'With multiline block' do
+    it 'registers an offense and corrects to `select` with `if` condition' do
+      expect_offense(<<~RUBY)
+        foo.map do |item|
+            ^^^^^^^^^^^^^ Use `select { ... }` instead of `map { ... }.compact`.
```
koic, line 8: The message looks simple as follows.

```suggestion
            ^^^^^^^^^^^^^ Use `select` instead`.
```

bbatsov: I'm fine with the more descriptive message, as some people might forget to remove the `compact` otherwise. "Replace X with Y." is another wording that I like.

lib/rubocop/cop/style/redundant_map_compact.rb
nobuyo: I think such an example using `if` can be rewritten as `reject`.

```ruby
# bad
[1, 2, 3, 4].map do |e|
  next if e.odd?

  e
end.compact
# => [2, 4]

# good
[1, 2, 3, 4].reject { |e| e.odd? }
# => [2, 4]
```

lib/rubocop/cop/style/redundant_map_compact.rb
nobuyo: Thank you. Actually, I wasn't sure about this name.

lib/rubocop/cop/style/redundant_map_compact.rb
nobuyo: The "instead of `map { |value| simple_condition? ? value : next }.compact`." at L7 could be bad.
Would modifying it as "instead of `map { ... }.compact`." solve the problem?

nobuyo:
Thank you for your review. I've renamed the cop and tweaked the message.

spec/rubocop/cop/style/map_compact_with_conditional_block_spec.rb
```diff
@@ -0,0 +1,389 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Style::MapCompactWithConditionalBlock, :config do
+  context 'With multiline block' do
+    it 'registers an offense and corrects to `select` with `if` condition' do
+      expect_offense(<<~RUBY)
+        foo.map do |item|
+            ^^^^^^^^^^^^^ Replace `map { ... }.compact` with `select`.
```
koic, line 8: OK, maybe the fix suggestion will be clearer if the block is shown.

```suggestion
            ^^^^^^^^^^^^^ Replace `map { ... }.compact` with `select { ... }`.
```

nobuyo: I agree. Updated :)

lib/rubocop/cop/style/map_compact_with_conditional_block.rb
```diff
@@ -0,0 +1,136 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for usage of `select` or `reject` instead of `map { ... }.compact`.
```
bbatsov, line 6: Btw, shouldn't the description be inverted? I think it most cops the bad case is the first one, right?

nobuyo: You're right. How is wording like "Prefer `select` or `reject` over `map { ... }.compact`."?

bbatsov: Sounds good to me.

config/default.yml
```diff
@@ -3925,6 +3925,13 @@ Style/LineEndConcatenation:
   VersionAdded: '0.18'
   VersionChanged: '0.64'
 
+Style/MapCompactWithConditionalBlock:
+  Description: >-
+                  Checks for usage of `select` or `reject`
```
bbatsov, line 3930: Same here.

lib/rubocop/cop/style/map_compact_with_conditional_block.rb
nobuyo: Thanks, I've updated the PR :)

bbatsov:
One more thing regarding the name - how about simply `Style/MapCompact`. I think the part about the block in the name is redundant and slightly confusing.

nobuyo:
I'm fine with that.

@koic How do you think?

koic:
It may not be ambiguous to user because the department is different between `Performance` and `Style`, but I was concerned about the contrast with `Performance/MapCompact` cop. This cop seems to be more specific to a particular block than `Performance/MapCompact` cop.
https://docs.rubocop.org/rubocop-performance/1.13/cops_performance.html#performancemapcompact

That may be over-thinking, though...

nobuyo:
So... Which should we name? I feel like they're both reasonable, but a little too geeky about the difference in `Performance` and `Style`.

bbatsov (approved):
Naming is hard. :-) Let's just go with the current name.

PR #10577, opened 2022-04-24
Drop Ruby 2.5 support

koic:
This includes the following changes:

## Require JRuby 9.3 or higher

CRuby 2.5 compatible JRuby 9.2 will be dropped and CRuby 2.6 compatible JRuby 9.3 or higher is required.

## Retire `Lint/UselessElse` cop

Ruby 2.6 does not support `begin` ... `else` syntax. So `Lint/UselessElse` can't work because it causes a parsing error.

NOTE: Some bugs found in RuboCop 1.28.1 have been fixed in the master branch. If the next version is a bug fix (i.e. 1.28.2), this PR would wait to be merged until the next minor release.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #10626, opened 2022-05-12
[Fix #10625] Restore the specification to `TargetRubyVersion: 2.5`

### LINKED ISSUE #10625: Decouple minimum TargetRubyVersion from supported runtime platform

For many of the gems I maintain, I've set up a combination of the following:

- A dev dependency on RuboCop with rather liberal version specification (e.g., "~> 1.27"
- A GitHub Action to run (among others) a RuboCop check, set up to run on pull requests, of course, but also once a month. This Action runs on Ruby 2.7 at the moment.

This combination is intended to catch any new offenses without having to update the gemspec.

However, this plan is no longer working since the Action for some of my projects now fails because latest RubyCop dropped support for Ruby 2.5 as a TargetRubyVersion.

Allow TargetRubyVersion 2.5, even when RuboCop itself will only run on Ruby 2.6 and up. This should be possible since Parser supports 2.5 just fine when running on a later Ruby version.

I could use a tighter version specification for RuboCop. However this means churn in the repositiories, and/or potentially not keeping my dependencies up-to-date with the latest versions. Both of these I'd like to avoid, to support my scheme to make maintaining my gems as light-weight as possible.

## Additional context

There was a very short but heated discussion about this in the semver repository starting here: https://github.com/semver/semver/issues/716#issuecomment-1115128186. I'd like to avoid any discussion of what semver really means and focus on this specific feature.

koic:
Fixes #10625 and reverts part of #10577.

Only the Ruby version (2.5) to runtime should have been dropped, not code analysis.
This PR the specification to `TargetRubyVersion: 2.5` So, this keeps it compatible with `TargetRubyVersion` up to 1.28.2.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_restore_target_ruby_version_ruby_25.md
+++ b/changelog/fix_restore_target_ruby_version_ruby_25.md
@@ -1,0 +1,1 @@
+* [#10625](https://github.com/rubocop/rubocop/issues/10625): Restore the specification to `TargetRubyVersion: 2.5`. ([@koic][])
--- a/lib/rubocop/cop/lint/erb_new_arguments.rb
+++ b/lib/rubocop/cop/lint/erb_new_arguments.rb
@@ -62,6 +62,9 @@ module RuboCop
       class ErbNewArguments < Base
         include RangeHelp
         extend AutoCorrector
+        extend TargetRubyVersion
+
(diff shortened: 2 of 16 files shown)


bbatsov:
Thanks!

PR #10630, opened 2022-05-13
[Fix #10627] Add command-line option `--ignore-unrecognized-cops`

### LINKED ISSUE #10627: Option to ignore unrecognized cop / department

We are using githubs superlinter (https://github.com/github/super-linter) for linting in our projects. The superlinter sometimes lags behind the current rubocop version for a few weeks (eg. current latest release is from April 1st - there have been 6 new rubocop versions in that time).

However, developers update rubocop more often locally, and run into cops we might not want to enforce, so we update our central rubocop config to disable / tweak rules. However, now github superlinter fails because we include configuration which it doesn't understand yet due to being on an older version of rubocop.

An option to opt-out of errors in case of unrecognized cops. The default could stay as is.

I though about disabling `NewCops` - but that's not really something we want to do, we just want to opt-out of cops which don't fit our use cases.

nobuyo:
…to ignore any unknown cops or departments in .rubocop.yml.

Fixes #10627.

This PR implements new command-line option `--ignore-unrecognized-cops`.

```rb
# .rubocop.yml
AllCops:
  NewCops: enable

Foo:
  Enabled: true

Bar/Baz:
  Enabled: true
```

```sh
# run without option
$ rubocop
Error: unrecognized cop or department Foo found in .rubocop.yml
unrecognized cop or department Bar/Baz found in .rubocop.yml

# with option
$ bundle exec rubocop --ignore-unrecognized-cops
The following cops or departments are not recognized and will be ignored:
unrecognized cop or department Foo found in .rubocop.yml
unrecognized cop or department Bar/Baz found in .rubocop.yml
Inspecting 3 files
...

#...
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/config_validator.rb
+++ b/lib/rubocop/config_validator.rb
@@ -102,6 +102,18 @@ module RuboCop
     end
 
     def alert_about_unrecognized_cops(invalid_cop_names)
+      unknown_cops = list_unknown_cops(invalid_cop_names)
+
+      if ConfigLoader.ignore_unrecognized_cops
+        warn Rainbow('The following cops or departments are not '\
+                     'recognized and will be ignored:').yellow
+        warn unknown_cops.join("\n")
+      elsif unknown_cops.any?
+        raise ValidationError, unknown_cops.join("\n")
+      end
+    end
+
+    def list_unknown_cops(invalid_cop_names)
       unknown_cops = []
       invalid_cop_names.each do |name|
         # There could be a custom cop with this name. If so, don't warn
@@ -119,7 +131,8 @@ module RuboCop
 
         unknown_cops << message
       end
-      raise ValidationError, unknown_cops.join("\n") if unknown_cops.any?
+
+      unknown_cops
     end
 
     def suggestion(name)
--- a/changelog/new_add_commandline_option_ignore_unrecognized_cops.md
+++ b/changelog/new_add_commandline_option_ignore_unrecognized_cops.md
@@ -1,0 +1,1 @@
+* [#10627](https://github.com/rubocop/rubocop/issues/10627): Add command-line option `--ignore-unrecognized-cops` to ignore any unknown cops or departments in .rubocop.yml. ([@nobuyo][])
--- a/docs/modules/ROOT/pages/usage/basic_usage.adoc
+++ b/docs/modules/ROOT/pages/usage/basic_usage.adoc
@@ -210,6 +210,9 @@ $ rubocop --only Rails/Blank,Layout/HeredocIndentation,Naming/FileName
 | `--ignore-parent-exclusion`
 | Ignores all Exclude: settings from all .rubocop.yml files present in parent folders. This is useful when you are importing submodules when you want to test them without being affected by the parent module's rubocop settings.
 
+| `--ignore-unrecognized-cops`
+| Ignore unrecognized cops or departments in the config.
+
 | `--init`
 | Generate a .rubocop.yml file in the current directory.
 
--- a/lib/rubocop/cli.rb
+++ b/lib/rubocop/cli.rb
@@ -124,6 +124,7 @@ module RuboCop
       ConfigLoader.disable_pending_cops = @options[:disable_pending_cops]
       ConfigLoader.enable_pending_cops = @options[:enable_pending_cops]
       ConfigLoader.ignore_parent_exclusion = @options[:ignore_parent_exclusion]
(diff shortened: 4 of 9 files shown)


lib/rubocop/config_validator.rb
```diff
@@ -102,11 +102,21 @@ def check_target_ruby
...
+        raise ValidationError, unknown_cops.join("\n")
+      end
+    end
+
+    def list_unknown_cops(invalid_cop_names)
       unknown_cops = []
       invalid_cop_names.each do |name|
         # There could be a custom cop with this name. If so, don't warn
         next if Cop::Registry.global.contains_cop_matching?([name])
-
```
koic, line 109: This looks like an intentional blank line.

nobuyo: Oh, sorry that is a mistake.

lib/rubocop/config_validator.rb
```diff
@@ -102,6 +102,17 @@ def check_target_ruby
     end
 
     def alert_about_unrecognized_cops(invalid_cop_names)
+      unknown_cops = list_unknown_cops(invalid_cop_names)
+
+      if ConfigLoader.ignore_unrecognized_cops
+        warn Rainbow('Ignoring unrecognized cops or departments by option.').yellow
```
bbatsov, line 108: I'd use wording like "The following cops or departments are not recognized and will be ignored:".

bbatsov:
Apart from my small remark the PR looks good to me.

lib/rubocop/config_validator.rb
nobuyo: Thank you! I've updated the message.

bbatsov:
Thanks!

pirj:
@nobuyo Thanks for adding this option!

 It's a solid first step to resolving another use case, checking new cops using a large set of real-world ([Ruby](https://github.com/pirj/real-world-ruby-apps), [Rails](https://github.com/eliotsykes/real-world-rails) and [RSpec](https://github.com/pirj/real-world-rspec)) repositories.

However, I've run into an issue:
```
$ rubocop --only Rails/StripHeredoc --ignore-unrecognized-cops /Users/pirj/source/real-world-rspec/
The following cops or departments are not recognized and will be ignored:

The following cops or departments are not recognized and will be ignored:

/Users/pirj/source/real-world-rspec/24pullrequests/.rubocop.yml: Metrics/LineLength has the wrong namespace - should be Layout
Error: The `Lint/HandleExceptions` cop has been renamed to `Lint/SuppressedException`.
(obsolete configuration found in /Users/pirj/source/real-world-rspec/24pullrequests/.rubocop.yml, please update it)
```

Does it make sense to ignore obsoletions with this flag, too?

Another strange thing is that there's nothing in the output after "will be ignored".

nobuyo:
Hi @pirj.

> Does it make sense to ignore obsoletions with this flag, too?

Yeah, it would be nice if obsoletions as a warning instead of an error. I'll take look at about that.

> Another strange thing is that there's nothing in the output after "will be ignored".

I tried with `real-world-rspec` repository, but no "will be ignored" outputs on my environment... 🤔 

```
$ rubocop --only Rails/StripHeredoc --ignore-unrecognized-cops /Users/nobuyo/dev/real-world-rspec
real-world-rspec/24pullrequests/.rubocop.yml: Metrics/LineLength has the wrong namespace - should be Layout
Error: The `Lint/HandleExceptions` cop has been renamed to `Lint/SuppressedException`.
(obsolete configuration found in real-world-rspec/24pullrequests/.rubocop.yml, please update it)
```

dvandersluis:
Ignoring obsoletions will be tricky. I added the ability to add obsoletions as warnings last year but there’s only a couple that use it. The rest fail because either the cop has changed/been removed or some of its configuration has changed, and without there being actual code in each cop to fallback (the obsoletions that warn generally maintain backwards compatibility with the previous configuration) the cop will fail. 

I’m not sure that adding fallbacks for every obsoletions at this point when some are many years old makes sense either tbh.

dvandersluis:
BTW, if I understand correctly, `ignore-unrecognized-cops` should handle rename/delete obsoletions so it’s really the config changes that will be the problem here.

pirj:
> no "will be ignored" outputs on my environment

Not a big deal since it's a just warning. Probably my env is slightly broken.

nobuyo:
> Another strange thing is that there's nothing in the output after "will be ignored".

Sorry I've confirmed reproducing it. And will be fixed by https://github.com/rubocop/rubocop/pull/10676.
Thanks @pirj!

PR #10637, opened 2022-05-13
[Fix #10636] Fix false positive in `Style/RedundantCondition` when the branches call the same method on different receivers

### LINKED ISSUE #10636: Style/RedundantCondition autofixes incorrectly (rubocop 1.29.1)
This code

```ruby
class Pippo
  def foo
    result =
      if x
        X.find(x)
      else
        Y.find(y)
      end

    result.to_s
  end
end
```

generates this warning:

```bash
app/pippo.rb:4:7: C: [Correctable] Style/RedundantCondition: Use double pipes || instead.
      if x ...
```

If I run `rubocop -a` it gets autofixed with:

```ruby
class Pippo
  def foo
    result =
      X.find(x || y)

    result.to_s
  end
end
```

but the behavior now it's changed.

## Expected behavior

I think the cop should not raise a warning, because it's not a redundant condition 🤔 

## RuboCop version

```
$ bundle exec rubocop -V
 ↳ $ bundle exec rubocop -V
1.29.1 (using Parser 3.1.2.0, rubocop-ast 1.17.0, running on ruby 2.7.6 x86_64-darwin21)
  - rubocop-performance 1.13.3
  - rubocop-rails 2.14.2
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.10.0
```

dvandersluis:
Fixes #10636.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/redundant_condition_spec.rb
+++ b/spec/rubocop/cop/style/redundant_condition_spec.rb
@@ -376,6 +376,31 @@ RSpec.describe RuboCop::Cop::Style::RedundantCondition, :config do
           end
         RUBY
       end
+
+      it 'registers an offense and correct when the branches are the same with the same receivers' do
+        expect_offense(<<~RUBY)
+          if x
+          ^^^^ Use double pipes `||` instead.
+            X.find(x)
+          else
+            X.find(y)
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          X.find(x || y)
+        RUBY
+      end
+
+      it 'does not register an offense when the branches are the same with different receivers' do
+        expect_no_offenses(<<~RUBY)
+          if x
+            X.find(x)
+          else
+            Y.find(y)
+          end
+        RUBY
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/redundant_condition_spec.rb
```diff
@@ -376,6 +376,34 @@ def test
           end
         RUBY
       end
+
+      it 'registers an offense and correct when the branches are the same with the same receivers' do
+        expect_offense(<<~RUBY)
+          result =
```
koic, line 382: That's just my two cents. Can you remove it? Because this assignment doesn't seem to matter.

```suggestion
```

spec/rubocop/cop/style/redundant_condition_spec.rb
```diff
@@ -376,6 +376,34 @@ def test
...
+
+        expect_correction(<<~RUBY)
+          result =
+            X.find(x || y)
+        RUBY
+      end
+
+      it 'does not register an offense when the branches are the same with different receivers' do
+        expect_no_offenses(<<~RUBY)
+        result =
```
koic, line 399: Ditto.

```suggestion
```

dvandersluis:
@koic thanks for the suggestion, updated.

koic:
Perfect 👍 Thanks!

PR #10638, opened 2022-05-15
[Fix #10618] Fix `LineBreakCorrector` so that it won't remove a semicolon in the class/module body

### LINKED ISSUE #10618: Autocorrect for `Style/TrailingBodyOnClass` breaks the code when body trails after singleton class definition
Related with https://github.com/rubocop/rubocop/pull/10617

The autocorrect for [Style/TrailingBodyOnClass](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/TrailingBodyOnClass) breaks this code
```ruby
class Foo bar # <------------ `bar` is the actual first body line of this class.
  class << self; baz  # <------------ `baz` is the actual first body line of this singleton class.
    def func
      puts 1
    end
  end
end
```
to
```ruby
class Foo 
  bar
  class << self baz # Syntax error
    def func
      puts 1
    end
  end
end
```

## Expected behavior

```ruby
class Foo 
  bar    
  class << self
    baz
    def func
      puts 1
    end
  end
end
```

Save the example code to example.rb, and run
```sh
exe/rubocop example.rb -a --only Style/TrailingBodyOnClass 
```

## RuboCop version

```sh
$ bundle exec rubocop -V
1.29.0 (using Parser 3.1.2.0, rubocop-ast 1.17.0, running on ruby 2.7.5 x86_64-darwin20)
  - rubocop-performance 1.13.3
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.10.0
```

johnny-miyake:
Fixes #10618

This PR fixes `LineBreakCorrector` (used by `Style/TrailingBodyOnClass` and `Style/TrailingBodyOnModule`).
It currently breaks the code by removing a semicolon in the class/module body when the class body is trailing class/module definition without a semicolon after the class/module name.

For example,
```ruby
class Foo def bar; end
end
class Foo bar
  a; b
end
```
is wrongly corrected into
```ruby
class Foo 
  def bar end # Unexpected correction
end
class Foo 
  bar
  a b # Unexpected correction
end
```
-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_linebreakcorrector_so_that_it_wont_remove_semicolon_in_the_class_module_body.md
+++ b/changelog/fix_linebreakcorrector_so_that_it_wont_remove_semicolon_in_the_class_module_body.md
@@ -1,0 +1,1 @@
+* [#10618](https://github.com/rubocop/rubocop/issues/10618): Fix `LineBreakCorrector` so that it won't remove a semicolon in the class/module body. ([@johnny-miyake][])
--- a/lib/rubocop/cop/correctors/line_break_corrector.rb
+++ b/lib/rubocop/cop/correctors/line_break_corrector.rb
@@ -50,7 +50,13 @@ module RuboCop
 
         def semicolon(node)
           @semicolon ||= {}.compare_by_identity
-          @semicolon[node] ||= processed_source.tokens_within(node).find(&:semicolon?)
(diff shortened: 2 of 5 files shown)


bbatsov:
The changes look good to me. Thanks!

PR #10632, opened 2022-05-13
[Fix #10258] Recover Ruby 2.4 code analysis using `TargetRubyVersion: 2.4`

koic:
Fixes #10258 and reverts part of #9648.

Only the Ruby version (2.4) to runtime should have been dropped, not code analysis.
This PR makes Ruby 2.4 code analysis with `TargetRubyVersion: 2.4`.
So, this keeps it compatible with `TargetRubyVersion` up to RuboCop 1.12.1 as semver.

It aims to solve essentially the same problem as #10626.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #7030, opened 2019-05-09
Follow up "Drop support for Ruby 2.2"

koic:
Follow up #7026.

This PR updates the following deficiencies.

- The default `EnforcedStyle` of `Style/FrozenStringLiteralComment is `always`,  so duplicate settings are removed from .rubocop.yml.
- Remove unnecessary logic because `SupportedStyles: auto_detection` has been removed from `Layout/IndentHeredoc`.
- Remove the check for warn introduced in #6227.
- Remove specification of `let(:ruby_version) { 2.3 }` written in spec.
- Remove unnecessary code using Ruby version
- Updated the README to make JRuby support 9.1 or higher because JRuby  9.0 compatible with Ruby 2.2.

```console
% ruby -v
jruby 9.0.5.0 (2.2.3) 2016-01-26 7bee00d Java HotSpot(TM) 64-Bit Server
VM 25.5-b02 on 1.8.0_05-b13 +jit [darwin-x86_64]
% ruby -v
jruby 9.1.17.0 (2.3.3) 2018-04-20 d8b1ff9 Java HotSpot(TM) 64-Bit Server
VM 25.5-b02 on 1.8.0_05-b13 +jit [darwin-x86_64]
% ruby -v
jruby 9.2.7.0 (2.5.3) 2019-04-09 8a269e3 Java HotSpot(TM) 64-Bit Server
VM 25.5-b02 on 1.8.0_05-b13 +jit [darwin-x86_64]
```

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [ ] Added tests.
* [ ] Added an entry to the [Changelog](https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format).
* [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.
* [x] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: https://chris.beams.io/posts/git-commit/

PR #10644, opened 2022-05-17
Recover Ruby 2.2 code analysis using `TargetRubyVersion: 2.2`

koic:
Follow up https://github.com/rubocop/rubocop/pull/10632#issuecomment-1125909580.

Reverts part of #6766, #7026, and #7030.

Only the Ruby version (2.2) to runtime should have been dropped, not code analysis.
This PR makes Ruby 2.2 code analysis with `TargetRubyVersion: 2.2`.
It aims to solve essentially the same problem as #10626, #10632, and #10640.

Previously, there was the following default enforced style `when_needed` for `Style/FrozenStringLiteralComment` cop.

```ruby
# @example EnforcedStyle: when_needed (default)
#   # The `when_needed` style will add the frozen string literal
#   # to files only when the `TargetRubyVersion` is set to 2.3+.
#   # bad
#   module Foo
#     # ...
#   end
#
#   # good
#   # frozen_string_literal: true
#
#   module Foo
#     # ...
#   end
```

This PR does not restore that option, but sets the `minimum_target_ruby_version 2.3` to make `always (default)` apply by default. It is a simple solution that does not handle frozen literal magic comment added in Ruby 2.3 when `TargetRubyVersion` is Ruby 2.2 or lower.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/changelog/fix_recover_ruby_22_code_analysis.md
+++ b/changelog/changelog/fix_recover_ruby_22_code_analysis.md
@@ -1,0 +1,1 @@
+* [#10644](https://github.com/rubocop/rubocop/pull/10644): Recover Ruby 2.2 code analysis using `TargetRubyVersion: 2.2`. ([@koic][])
--- a/lib/rubocop/cop/lint/safe_navigation_chain.rb
+++ b/lib/rubocop/cop/lint/safe_navigation_chain.rb
@@ -25,6 +25,9 @@ module RuboCop
       #   x&.foo || bar
       class SafeNavigationChain < Base
         include NilMethods
+        extend TargetRubyVersion
+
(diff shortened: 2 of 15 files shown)


bbatsov:
Not sure if we need to go back so far, as I doubt there are many projects that still target such an old Ruby release. I'll leave this up to you to decide. Personally I was happy when we got back to 2.5. :D

koic:
I found the following backlink in PR #10632 that recovered Ruby 2.4 code analysis.
https://github.com/brianmario/mysql2/pull/1256

So, there seems to be use case with gems that support older Ruby versions. Perhaps RuboCop's analysis can recover up to Ruby 1.9 code. I'm trying it, but I'm not sure if it's the best 😅

PR #10653, opened 2022-05-19
[Fix #10652] Fix a false positive for `Style/FetchEnvVar` in conditions

### LINKED ISSUE #10652: False positives with `Style/FetchEnvVar` in conditions
## Actual behavior

@koic added in #10603:

> The ENV var as a if condition is currently allowed.
> So the ENV var already used in the condition will be accepted as well.

````ruby
if ENV["SOME_VAR"]
  do_something ENV["SOME_VAR"]
end
````

Nothing to say about this snippet.  
But it won't work with more complex conditions:

````ruby
if some_condition || ENV["FOO"] 
  do_something ENV.fetch("FOO")
end

if ENV["BAR"] || some_condition
  do_something ENV.fetch("BAR")
end

# Style/FetchEnvVar: Use ENV.fetch("FOO") or ENV.fetch("FOO", nil) instead of ENV["FOO"].
#   if some_condition || ENV["FOO"]
#                        ^^^^^^^^^^^
# Style/FetchEnvVar: Use ENV.fetch("BAR") { some_condition } instead of ENV["BAR"] || some_condition.
#   if ENV["BAR"] || some_condition
#      ^^^^^^^^^^^
````

I don't think the cop should suggest to use `fetch` in the first condition `if some_condition || ENV["FOO"]` just as `if ENV["FOO"]` is allowed.

The second condition might be subjective: `ENV.fetch("BAR") { some_condition }` is correct but just because the first example should not report any offenses, this one should not suggest to use `fetch` in conditions. 

There no issue with the actual behavior inside the `if` blocks:
as `some_condition` might be true, we should use `do_something ENV.fetch("VAR1")`

````ruby
if ENV["ITEM_1"] == 1
  do_something ENV["ITEM_1"]
end

if ENV["ITEM_2"].present?
  do_something ENV["ITEM_2"]
end

if ENV["ITEM_3"].in?(%w[A B C])
  do_something_else ENV["ITEM_3"]
end

if ENV.key?("ITEM_4")
  do_something ENV["ITEM_4"]
end

# Style/FetchEnvVar: Use ENV.fetch("ITEM_1") or ENV.fetch("ITEM_1", nil) instead of ENV["ITEM_1"].
#     do_something ENV["ITEM_1"]
#                  ^^^^^^^^^^^^^
# Style/FetchEnvVar: Use ENV.fetch("ITEM_2") or ENV.fetch("ITEM_2", nil) instead of ENV["ITEM_2"].
#     do_something ENV["ITEM_2"]
#                  ^^^^^^^^^^^^^
# Style/FetchEnvVar: Use ENV.fetch("ITEM_3") or ENV.fetch("ITEM_3", nil) instead of ENV["ITEM_3"].
#     do_something ENV["ITEM_3"]
#                  ^^^^^^^^^^^^^
# Style/FetchEnvVar: Use ENV.fetch("ITEM_4") or ENV.fetch("ITEM_3", nil) instead of ENV["ITEM_3"].
#     do_something ENV["ITEM_4"]
#                  ^^^^^^^^^^^^^
````

Is it possible to accept these situation as the ENV var has been "tested" in conditions ?
I guess it is complicated to tell if the methods used in conditions are testing or not the variable.
Allowing equallity and predicate methods could be a start ?

````ruby
if %w[A B C].include?(ENV["OPTION"])
  do_something ENV["OPTION"]
end

# Style/FetchEnvVar: Use ENV.fetch("OPTION") or ENV.fetch("OPTION", nil) instead of ENV["OPTION"].
#   if %w[A B C].include?(ENV["OPTION"])
#                         ^^^^^^^^^^^^^
# Style/FetchEnvVar: Use ENV.fetch("OPTION") or ENV.fetch("OPTION", nil) instead of ENV["OPTION"].
#     do_something ENV["OPTION"]
#                  ^^^^^^^^^^^^^
````

This last one might be more complicated.
Is it possible to catch and exclude this situation ?
There is no point in using fetch in the `if` condition nor inside the `id` block.

## RuboCop version

```
$ [bundle exec] rubocop -V
1.29.1 (using Parser 3.1.2.0, rubocop-ast 1.18.0, running on ruby 2.7.6 x86_64-darwin21)
  - rubocop-performance 1.13.3
  - rubocop-rails 2.14.2
  - rubocop-rspec 2.11.1
```

### PR DESCRIPTION
**Pull request by ydah**

-----------------

* [ ] Added tests.

### CODE REVIEW

File: lib/rubocop/cop/style/fetch_env_var.rb [hunk 1]
```diff
@@ -97,11 +97,33 @@ def allowed_var?(node)
...
 
+          node.parent.send_type? && (node.parent.prefix_bang? || node.parent.comparison_method?)
+        end
+
+        def used_if_condition_in_body(node) # rubocop:disable Metrics/CyclomaticComplexity
           if_node = node.ancestors.find(&:if_type?)
-          return true if if_node&.condition == node
+          if_node&.condition&.child_nodes == node.child_nodes ||
+            used_with_other_condition?(node, if_node&.condition) ||
+            if_node&.condition&.comparison_method? ||
```
johnny-miyake, line 109: I found this code gets an error.
```ruby
if a || b
  puts ENV['C']
end
```

```bash
$ exe/rubocop -d --only "Style/FetchEnvVar" test.rb 
...
An error occurred while Style/FetchEnvVar cop was inspecting /Users/johnny/development/rubocop/test.rb:2:7.
undefined method `comparison_method?' for s(:or,
  s(:send, nil, :a),
  s(:send, nil, :b)):RuboCop::AST::OrNode
/Users/johnny/development/rubocop/lib/rubocop/cop/style/fetch_env_var.rb:109:in `used_if_condition_in_body'
```

ydah: Thank you so much. I'll fixed it.

ydah:
I updated this PR! Thank you so much.

lib/rubocop/cop/style/fetch_env_var.rb
```diff
@@ -97,11 +97,41 @@ def allowed_var?(node)
...
           return false if node.root?
+          return true if used_if_condition_in_body(node)
 
+          node.parent.send_type? && (node.parent.prefix_bang? || node.parent.comparison_method?)
+        end
+
+        def used_if_condition_in_body(node)
           if_node = node.ancestors.find(&:if_type?)
-          return true if if_node&.condition == node
+          if_node&.condition&.child_nodes == node.child_nodes ||
```
johnny-miyake, line 107:
```ruby
if ENV && x
  ENV[x]
end
```
The cop in this PR doesn't report that this code is offensive, while the current version of rubocop does. (I'm not sure if it's worth it to handle this edge case, though.)

lib/rubocop/cop/style/fetch_env_var.rb
```diff
@@ -97,11 +97,41 @@ def allowed_var?(node)
...
+          if condition&.or_type? || condition&.and_type?
+            condition&.child_nodes&.any?(node)
+          else
+            false
+          end
+        end
+
+        def used_with_comparison_method?(condition)
+          if condition&.send_type?
+            condition&.comparison_method?
```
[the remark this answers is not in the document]
ydah: Oops... I didn't realize it. Thank you so much! I'll fix it.

lib/rubocop/cop/style/fetch_env_var.rb
ydah: I see. I hadn't even envisioned such a case. Thanks for your review! I'll fix it.

ydah:
I updated this PR! Thank you so much.

lib/rubocop/cop/style/fetch_env_var.rb
```diff
@@ -97,11 +97,47 @@ def allowed_var?(node)
...
+            false
+          end
+        end
+
+        def used_with_predicate_method?(node, condition)
+          if condition&.send_type? && condition&.predicate_method?
+            condition&.child_nodes&.any?(node)
+          else
+            false
+          end
```
koic, line 140: I'm worried about the amount of save navigation and duplication of code. How about the following updates?

```diff
diff --git a/lib/rubocop/cop/style/fetch_env_var.rb b/lib/rubocop/cop/style/fetch_env_var.rb
index 69d24b626..c997b67cb 100644
--- a/lib/rubocop/cop/style/fetch_env_var.rb
+++ b/lib/rubocop/cop/style/fetch_env_var.rb
@@ -104,40 +104,21 @@ module RuboCop

         def used_if_condition_in_body(node)
           if_node = node.ancestors.find(&:if_type?)
-          used_if_condition?(node, if_node&.condition) ||
-            used_with_other_condition?(node, if_node&.condition) ||
-            used_with_comparison_method?(node, if_node&.condition) ||
-            used_with_predicate_method?(node, if_node&.condition)
-        end

-        def used_if_condition?(node, condition)
-          return false unless condition&.send_type?
+          return false unless (condition = if_node&.condition)
+          return true if condition.send_type? && (condition.child_nodes == node.child_nodes)

-          condition&.child_nodes == node.child_nodes
+          used_in_condition?(node, condition)
         end

-        def used_with_other_condition?(node, condition)
-          if condition&.or_type? || condition&.and_type?
-            condition&.child_nodes&.any?(node)
-          else
-            false
+        def used_in_condition?(node, condition)
+          if condition.send_type?
+            return false if !condition.comparison_method? && !condition.predicate_method?
+          elsif !condition.or_type? && !condition.and_type?
+            return false
           end
-        end

-        def used_with_comparison_method?(node, condition)
-          if condition&.send_type?
-            condition&.comparison_method? && condition&.child_nodes&.any?(node)
-          else
-            false
-          end
-        end
-
-        def used_with_predicate_method?(node, condition)
-          if condition&.send_type? && condition&.predicate_method?
-            condition&.child_nodes&.any?(node)
-          else
-            false
-          end
+          condition.child_nodes&.any?(node)
         end

         def offensive?(node)
```

koic: The updated code is below.

```ruby
         def used_if_condition_in_body(node)
           if_node = node.ancestors.find(&:if_type?)

           return false unless (condition = if_node&.condition)
           return true if condition.send_type? && (condition.child_nodes == node.child_nodes)

           used_in_condition?(node, condition)
         end

         def used_in_condition?(node, condition)
           if condition.send_type?
             return false if !condition.comparison_method? && !condition.predicate_method?
           elsif !condition.or_type? && !condition.and_type?
             return false
           end

           condition.child_nodes&.any?(node)
         end
```

lib/rubocop/cop/style/fetch_env_var.rb
```diff
@@ -39,7 +39,7 @@ module Style
       #   !ENV['X']
       #   ENV['X'].some_method # (e.g. `.nil?`)
       #
-      class FetchEnvVar < Base
+      class FetchEnvVar < Base # rubocop:disable Metrics/ClassLength
```
koic, line 42: Can you rebase with the latest master branch to remove this disable comment?

```suggestion
      class FetchEnvVar < Base
```

spec/rubocop/cop/style/fetch_env_var_spec.rb
```diff
@@ -305,16 +305,127 @@
       RUBY
     end
 
-    it 'registers no offenses when using an `ENV` var that is different from `if` condition in the body' do
+    it 'registers no offenses when using the same `ENV` var as `if` condition in the body with other conditon' do
```
johnny-miyake, line 308: typo
```suggestion
    it 'registers no offenses when using the same `ENV` var as `if` condition in the body with other conditions do
```

ydah: Oops, fixed that. Thank you so much!

lib/rubocop/cop/style/fetch_env_var.rb
```diff
@@ -97,11 +97,28 @@ def allowed_var?(node)
...
+          return true if condition.send_type? && (condition.child_nodes == node.child_nodes)
+
+          used_in_condition?(node, condition)
+        end
+
+        def used_in_condition?(node, condition)
+          if condition.send_type?
+            return false if !condition.comparison_method? && !condition.predicate_method?
+          elsif !condition.or_type? && !condition.and_type?
+            return false
```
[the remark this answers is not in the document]
ydah: I think you're right. It's an unnecessary `elsif`, so I'll remove it.

johnny-miyake:
This case is also the same kind of false positive, I think.
```ruby
if ENV['X'] = x
  puts ENV['X']
end

# The above is corrected into,

if ENV['X'] = x
  puts ENV.fetch('X', nil)
end
```

bbatsov:
I'm about to cut a release today, so I'll merge this as is. The additional false positive can be addressed separately.

ydah:
Sorry for the delay in addressing this issue.
I've created a corresponding PR: https://github.com/rubocop/rubocop/pull/10670.

PR #10663, opened 2022-05-23
[Fix #10656] Mark `Style/RedundantInterpolation` as unsafe autocorrection

### LINKED ISSUE #10656: `Style/RedundantInterpolation` auto-correct should not be marked safe or should be implemented differently
The `Style/RedundantInterpolation` cop's auto-correct should not be marked safe because it cannot guarantee that subsequent code wasn't relying on the "redundant" interpolation essentially dup'ing the string.

I've put together a brief demo of two scenarios where replacing the interpolation with `to_s` fails to produce equivalent code: https://github.com/jimryan/rubocop-redundant-interpolation-unsafe-autocorrect-demo

An alternative solution to not marking as safe would be to replace redundant string interpolations with `.to_s.dup` which is more precisely what the redundant interpolation was doing. E.g. `"#{foo}"` becomes `foo.to_s.dup`. I believe that would be safe.

--------

## Expected behavior

Autocorrection should not be marked "safe" for this cop. I would not expect the potential for an error or change in behavior of any kind after running a safe auto-correction.

Alternative behavior could be to replace redundant string interpolations with `to_s.dup`.

## Actual behavior

It is marked safe, yet safe autocorrection can break/change code in some scenarios.

Steps are in the `README` of the [demo repo](https://github.com/jimryan/rubocop-redundant-interpolation-unsafe-autocorrect-demo), but in general:
1. Write code that interpolates a string and stores it in a new variable
2. Mutate that string
3. `Style/RedundantInterpolation` will replace that interpolation with `.to_s` on the original string, which just returns the same String object. Now the mutation is mutating the _original_ string (the one that was interpolated), and not a copy of it. **When a string is the subject of the redundant interpolation, the interpolation is equivalent to `dup`, not `to_s`.** 

## RuboCop version

```
$ [bundle exec] rubocop -V
1.29.1 (using Parser 3.1.2.0, rubocop-ast 1.18.0, running on ruby 2.7.6 arm64-darwin21)
```

koic:
Fixes #10656.

This PR marks `Style/RedundantInterpolation` as unsafe autocorrection because when calling a destructive method to string, the resulting string may have different behavior or raise `FrozenError`.

```ruby
x = 'a'
y = "#{x}"
y << 'b'   # return 'ab'
x          # return 'a'
y = x.to_s
y << 'b'   # return 'ab'
x          # return 'ab'

x = 'a'.freeze
y = "#{x}"
y << 'b'   # return 'ab'.
y = x.to_s
y << 'b'   # raise `FrozenError`.
```

Updating the auto-corrected code from `'a'.to_s` to `'a'.to_s.dup` will get safe, but it will make `dup` redundant when unneeded. And This would go against the cop's name "redundant". That means replacing it with a different redundancy that cannot be removed by static analysis. So I decided to mark it as unsafe.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_mark_unsafe_autocorrection_for_style_redundant_interpolation.md
+++ b/changelog/change_mark_unsafe_autocorrection_for_style_redundant_interpolation.md
@@ -1,0 +1,1 @@
+* [#10656](https://github.com/rubocop/rubocop/issues/10656): Mark `Style/RedundantInterpolation` as unsafe autocorrection. ([@koic][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -4599,7 +4599,9 @@ Style/RedundantInitialize:
 Style/RedundantInterpolation:
   Description: 'Checks for strings that are just an interpolated expression.'
   Enabled: true
+  SafeAutoCorrect: false
   VersionAdded: '0.76'
+  VersionChanged: '<<next>>'
 
 Style/RedundantParentheses:
   Description: "Checks for parentheses that seem not to serve any purpose."
--- a/lib/rubocop/cop/style/redundant_interpolation.rb
+++ b/lib/rubocop/cop/style/redundant_interpolation.rb
@@ -5,6 +5,27 @@ module RuboCop
     module Style
       # Checks for strings that are just an interpolated expression.
       #
+      # @safety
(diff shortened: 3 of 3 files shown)


bbatsov:
Thanks!

PR #10673, opened 2022-05-26
Update auto-gen-config's comment re auto-correct for `SafeAutoCorrect: false`

ydah:

-----------------

* [ ] Added tests.


lib/rubocop/formatter/disabled_config_formatter.rb
```diff
@@ -242,6 +243,22 @@ def output_exclude_path(output_buffer, exclude_path, parent)
...
+      def unsafe?(config)
+        config['Safe'] == false
+      end
+
+      def safe_auto_correct?(cofig)
+        cofig['SafeAutoCorrect'] || cofig['SafeAutoCorrect'].nil?
+      end
+
+      def unsafe_auto_correct?(cofig)
+        cofig['SafeAutoCorrect'] == false
```
nobuyo, line 260: typo `cofig` → `config`

ydah: Thank you very much. I fixed it.

ydah:
Indeed it is. I've standardized on `autocorrect`.

lib/rubocop/formatter/disabled_config_formatter.rb
```diff
@@ -242,6 +243,22 @@ def output_exclude_path(output_buffer, exclude_path, parent)
...
+        config['Safe'] == false
+      end
+
+      def safe_autocorrect?(config)
+        config['SafeAutoCorrect'] || config['SafeAutoCorrect'].nil?
+      end
+
+      def unsafe_autocorrect?(config)
+        config['SafeAutoCorrect'] == false
+      end
```
nobuyo, line 261: Looks like they can be combined into one method.
Actually, it's defined in autocorrect_logic, but namespace considerations will force us to redefine it here.
https://github.com/rubocop/rubocop/blob/00e7ac78fb83383ef537e9e08ecf2301a1427578/lib/rubocop/cop/autocorrect_logic.rb#L27-L29

```suggestion
      def safe_autocorrect?(config)
        config.fetch('Safe', true) && config.fetch('SafeAutoCorrect', true)
      end
```

lib/rubocop/formatter/disabled_config_formatter.rb
```diff
@@ -137,13 +137,14 @@ def output_cop_comments(output_buffer, cfg, cop_name, offense_count)
         output_cop_param_comments(output_buffer, params, default_cfg)
       end
 
-      def supports_safe_auto_correct?(cop_class, default_cfg)
+      def supports_safe_autocorrect?(cop_class, default_cfg)
         cop_class&.support_autocorrect? &&
-          (default_cfg.nil? || default_cfg['Safe'] || default_cfg['Safe'].nil?)
+          (default_cfg.nil? || (safe?(default_cfg) && safe_autocorrect?(default_cfg)))
```
nobuyo, line 142: Combined with the other comments, I think the condition can be written like this.

```diff
       def supports_safe_autocorrect?(cop_class, default_cfg)
-        cop_class&.support_autocorrect? &&
-          (default_cfg.nil? || (safe?(default_cfg) && safe_autocorrect?(default_cfg)))
+        cop_class&.support_autocorrect? && safe_autocorrect?(default_cfg)
       end
```

lib/rubocop/formatter/disabled_config_formatter.rb
```diff
@@ -137,13 +137,14 @@ def output_cop_comments(output_buffer, cfg, cop_name, offense_count)
...
       end
 
       def supports_unsafe_autocorrect?(cop_class, default_cfg)
-        cop_class&.support_autocorrect? && !default_cfg.nil? && default_cfg['Safe'] == false
+        cop_class&.support_autocorrect? &&
+          !default_cfg.nil? && (unsafe?(default_cfg) || unsafe_autocorrect?(default_cfg))
```
nobuyo, line 147: This line can also be expressed by inverting safe_autocorrect, right?

```diff
       def supports_unsafe_autocorrect?(cop_class, default_cfg)
-        cop_class&.support_autocorrect? &&
-          !default_cfg.nil? && (unsafe?(default_cfg) || unsafe_autocorrect?(default_cfg))
+        cop_class&.support_autocorrect? && !safe_autocorrect?(default_cfg)
       end
```

lib/rubocop/formatter/disabled_config_formatter.rb
```diff
@@ -242,6 +241,10 @@ def output_exclude_path(output_buffer, exclude_path, parent)
         regexp = exclude_path
         output_buffer.puts "    - !ruby/regexp /#{regexp.source}/"
       end
+
+      def safe_autocorrect?(config)
+        config&.fetch('Safe', true) && config&.fetch('SafeAutoCorrect', true)
```
nobuyo, line 246: Sorry my example didn't take into account the case where config is nil. But wouldn't using save_navigation_operator here have the opposite result of what originally intended?

I'm not sure whether there is a real-world case that config is nil though...

```rb
config = {}
config&.fetch('Safe', true) && config&.fetch('SafeAutoCorrect', true) # => true

config = nil
config&.fetch('Safe', true) && config&.fetch('SafeAutoCorrect', true) # => nil
```

ydah: Indeed. I fixed it to take that into account as well. WDYT?

nobuyo: Looks good to me, thank you!
At first I thought it could be written like following, but yeah, this would invert the meaning when unsafe side.

Maybe the maintainers can give you a better pointer.

```rb
def safe_autocorrect?(config)
  return true if config.nil?

  config.fetch('Safe', true) && config.fetch('SafeAutoCorrect', true)
end
```

ydah: Thanks for taking the time to brush up on yours! I appreciate it.

Issue #10574, opened 2022-04-23
`Style/SingleArgumentDig`: false positive for arguments forwarding


yujideveloper:
## Expected behavior

It does not detect offense of `Style/SingleArgumentDig`.

## Actual behavior

```
$ rubocop --debug --cache false --only Style/SingleArgumentDig foo.rb
For /Users/yuji_developer/foo: configuration from /Users/yuji_developer/foo/.rubocop.yml
Default configuration from /Users/yuji_developer/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/rubocop-1.28.1/config/default.yml
Inspecting 1 file
Scanning /Users/yuji_developer/foo/foo.rb
C

Offenses:

foo.rb:7:5: C: [Correctable] Style/SingleArgumentDig: Use @hash[...] instead of @hash.dig(...).
    @hash.dig(...)
    ^^^^^^^^^^^^^^

1 file inspected, 1 offense detected, 1 offense auto-correctable
Finished in 0.05598199996165931 seconds

```

## Steps to reproduce the problem

Run `rubocop --only Style/SingleArgumentDig foo.rb` on the following config and code.

```yaml
# .rubocop.yml
AllCops:
  TargetRubyVersion: 2.7
  NewCops: disable
```
```ruby
# foo.rb
class Foo
  def initialize
    @hash = {}
  end

  def dig(...)
    @hash.dig(...)
  end
end

```

## RuboCop version

```
$ rubocop -V
1.28.1 (using Parser 3.1.2.0, rubocop-ast 1.17.0, running on ruby 2.7.6 arm64-darwin21)
```

PR #10674, opened 2022-05-27
Fix a false positive for `Naming/AccessorMethodName` with type of the first argument is other than `arg`

ydah:
This PR is fixed based on a similar problem: https://github.com/rubocop/rubocop/issues/10574

`Naming/AccessorMethodName` should only inform offense
if the type of the first argument is `arg`.

```ruby
# Bad
def set_something(arg)
  # ...
end

# Good
def set_something(arg = :default)
  # ...
end

def set_something(*args)
  # ...
end

def set_something(k: v)
  # ...
end

def set_something(k:)
  # ...
end

def set_something(**options)
  # ...
end

def set_something(&block)
  # ...
end

def set_something(...)
  # ...
end
```

-----------------

~* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).~
* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/naming/accessor_method_name.rb
+++ b/lib/rubocop/cop/naming/accessor_method_name.rb
@@ -63,7 +63,9 @@ module RuboCop
         end
 
         def bad_writer_name?(node)
-          node.method_name.to_s.start_with?('set_') && node.arguments.one?
+          node.method_name.to_s.start_with?('set_') &&
+            node.arguments.one? &&
+            node.first_argument.arg_type?
         end
       end
     end
--- a/spec/rubocop/cop/naming/accessor_method_name_spec.rb
+++ b/spec/rubocop/cop/naming/accessor_method_name_spec.rb
@@ -1,19 +1,19 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::Naming::AccessorMethodName, :config do
-  it 'registers an offense for method get_... with no args' do
+  it 'registers an offense for method get_something with no args' do
     expect_offense(<<~RUBY)
-      def get_attr
-          ^^^^^^^^ Do not prefix reader method names with `get_`.
+      def get_something
+          ^^^^^^^^^^^^^ Do not prefix reader method names with `get_`.
         # ...
       end
     RUBY
   end
 
-  it 'registers an offense for singleton method get_... with no args' do
+  it 'registers an offense for singleton method get_something with no args' do
     expect_offense(<<~RUBY)
-      def self.get_attr
-               ^^^^^^^^ Do not prefix reader method names with `get_`.
+      def self.get_something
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/naming/accessor_method_name.rb
```diff
@@ -63,7 +63,9 @@ def bad_reader_name?(node)
         end
 
         def bad_writer_name?(node)
-          node.method_name.to_s.start_with?('set_') && node.arguments.one?
+          node.method_name.to_s.start_with?('set_') &&
+            node.arguments.one? &&
+            !node.first_argument.forward_arg_type?
```
koic, line 68: This seems like it should probably only match the first `arg` type. Can you add tests and update the condition?

```
% ruby-parse -e "def foo(arg); end"
(def :foo
  (args
    (arg :arg)) nil)
```

```
% ruby-parse -e "def foo(arg = :default); end"
(def :foo
  (args
    (optarg :arg
      (sym :default))) nil)
```

```
% ruby-parse -e "def foo(*args); end"
(def :foo
  (args
    (restarg :args)) nil)
```

```
% ruby-parse -e "def foo(k:); end"
(def :foo
  (args
    (kwarg :k)) nil)
```

```
% ruby-parse -e "def foo(**options); end"
(def :foo
  (args
    (kwrestarg :options)) nil)
```

```
% ruby-parse -e "def foo(&block); end"
(def :foo
  (args
    (blockarg :block)) nil)
```

```
% ruby-parse -e "def foo(...); end"
(def :foo
  (args
    (forward-arg)) nil)
```

ydah: Yes, that sounds very good to me. I'll fix it.

spec/rubocop/cop/naming/accessor_method_name_spec.rb
```diff
@@ -3,81 +3,139 @@
...
-               ^^^^^^^^ Do not prefix reader method names with `get_`.
+      def self.get_something
+               ^^^^^^^^^^^^^ Do not prefix reader method names with `get_`.
         # ...
       end
     RUBY
   end
 
-  it 'accepts method get_something with args' do
+  it 'accepts method get_... with args' do
```
koic, line 22: IMO, `get_...` looks weird, so it looks better to be consistent with `get_something`.

spec/rubocop/cop/naming/accessor_method_name_spec.rb
```diff
@@ -3,81 +3,139 @@
...
+  it 'accepts singleton method get_... with args' do
     expect_no_offenses(<<~RUBY)
       def self.get_something(arg)
         # ...
       end
     RUBY
   end
 
-  it 'registers an offense for method set_something with one arg' do
+  it 'registers an offense for method set_... with one arg' do
```
koic, line 38: Ditto.

koic:
Thanks!

PR #10676, opened 2022-05-28
Fix `--ignore-unrecognized-cops` option always showing empty warning even if there was no problem

nobuyo:
The internal condition was wrong, and the message `The following cops or departments are ...` was always displayed with just the option is specified.

Reported at: https://github.com/rubocop/rubocop/pull/10630#issuecomment-1140263726

```rb
$ bundle exec rubocop --ignore-unrecognized-cops
The following cops or departments are not recognized and will be ignored: # <--- 👀 

Inspecting 2 files
.W

...
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/config_validator.rb
+++ b/lib/rubocop/config_validator.rb
@@ -104,13 +104,17 @@ module RuboCop
     def alert_about_unrecognized_cops(invalid_cop_names)
       unknown_cops = list_unknown_cops(invalid_cop_names)
 
+      return if unknown_cops.empty?
+
       if ConfigLoader.ignore_unrecognized_cops
         warn Rainbow('The following cops or departments are not '\
                      'recognized and will be ignored:').yellow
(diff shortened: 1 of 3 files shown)


pirj (approved):
Thanks!

lib/rubocop/config_validator.rb
```diff
@@ -104,13 +104,17 @@ def check_target_ruby
     def alert_about_unrecognized_cops(invalid_cop_names)
       unknown_cops = list_unknown_cops(invalid_cop_names)
 
+      return unless unknown_cops.any?
```
koic, line 107: Is it something like this? It may be a little simpler In this case.

```suggestion
      return if unknown_cops.empty?
```

nobuyo: Thank you, I've updated.

PR #10690, opened 2022-06-03
Fix: rubocop-performance SafeMode deprecation version 

santir489:
Rubocop v0.74.0 [changelog](https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#0740-2019-07-31) says:
```
Deprecate the SafeMode option. Users will need to upgrade rubocop-performance to 1.15.0+ before the SafeMode module is removed. ([@rrosenblum](https://github.com/rrosenblum))
```

But rubocop-performance version 1.15.0 does not exist. This was introduced in this [PR](https://github.com/rubocop/rubocop/pull/7249/files) 
It should say 1.5.0 instead of 1.15.0. 

According to [rubocop-performance releases](https://github.com/rubocop/rubocop-performance/releases/tag/v1.5.0), SafeMode was removed in version 1.5.0


bbatsov:
Thanks!

PR #10696, opened 2022-06-06
Add new `Lint/NonAtomicFileOperation` cop

ydah:
This PR is add new `Lint/NonAtomicFileOperation` cop

Checks for non-atomic file operations.
And then replace it with a nearly equivalent and atomic method.

These can cause problems that are difficult to reproduce, especially in cases of frequent file operations in parallel, such as test runs with parallel_rspec.

For examples: creating a directory if there is none, has the following problems

An exception occurs when the directory didn't exist at the time of `exist?`, but someone else created it before `mkdir` was executed.

Subsequent processes are executed without the directory that should be there when the directory existed at the time of `exist?`, but someone else deleted it shortly afterwards.

## @safety
This cop is unsafe, because autocorrection change to atomic processing.
The atomic processing of the replacement destination is not guaranteed to be strictly equivalent to that before the replacement.

## @example
```ruby
# bad
unless FileTest.exist?(path)
  FileUtils.makedirs(path)
end

if FileTest.exist?(path)
  FileUtils.remove(path)
end

# good
FileUtils.mkdir_p(path)

FileUtils.rm_rf(path)
```
-----------------

* [x] Added tests.


bbatsov:
In this case the `Redundant` check is actually `Dangerous` if you ask me, so perhaps we can rename the cop accordingly. Something like `NonAtomicFileOperation` might also be appropriate IMO. @rubocop/rubocop-core what do you think? 

P.S. Naming is still hard! :D

ydah:
Naming always annoys me for a long time...
I feel that `NonAtomicFileOperation` looks good and I changed it. What do you think?

bbatsov:
It's fine by me, but I'll leave this open for a couple of days to see if some of the other maintainers comes up with a better name. 

It'd be nice if we also mention this in the Ruby style guide, as I doubt many people are aware of this problem.

config/default.yml
```diff
@@ -1956,6 +1956,13 @@ Lint/NoReturnInBeginEndBlocks:
   Enabled: pending
   VersionAdded: '1.2'
 
+Lint/NonAtomicFileOperation:
+  Description: Checks for non-atomic file operations.
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  Safe: false
```
koic, line 1963: I think this `Safe: false` can be removed because it seems likely that false positives will not occur.

```suggestion
```

mathieujobin:
this new cop makes my rubocop crash...
on both 1.30.0 and 1.30.1

image

koic:
@mathieujobin Can you open a new issue with the bug report template?
https://github.com/rubocop/rubocop/issues/new/choose

PR #10699, opened 2022-06-07
[Fix #10698] Enhance `Style/HashExcept` to support array inclusion checks

### LINKED ISSUE #10698: Support array inclusion checks for Style/HashExcept

Given the following code:
```
{foo: 1, bar: 2, baz: 3}.reject { |k, _v| k.in?(%I[bar baz]) }
```

Style/HashExcept does not trigger an offense, as currently it only support equality checks

It should suggest:
```
{foo: 1, bar: 2, baz: 3}.except(:bar, :baz)
```

Here are some other usages that could be detected:
```
{foo: 1, bar: 2, baz: 3}.reject { |k, _v| k.in?(SOME_CONSTANT) }
{foo: 1, bar: 2, baz: 3}.reject { |k, _v| k.in?(some_var) }
{foo: 1, bar: 2, baz: 3}.reject { |k, _v| %I[bar baz].include?(k) }
{foo: 1, bar: 2, baz: 3}.select { |k, _v| %I[bar baz].exclude?(k) }
```

nobuyo:
Fixes #10698.

```ruby
# Note the actual behaviors because it is confusing.

require 'active_support/core_ext/enumerable'
require 'active_support/core_ext/object'

# except
{ foo: 1, bar: 2, baz: 3 }.except(:foo, :bar)
# => {:baz=>3}

# reject
{ foo: 1, bar: 2, baz: 3 }.reject { |k, v| k.in?(%i[foo bar]) }
{ foo: 1, bar: 2, baz: 3 }.reject { |k, v| %i[foo bar].include?(k) }
{ foo: 1, bar: 2, baz: 3 }.reject { |k, v| !%i[foo bar].exclude?(k) }
# => {:baz=>3}

# select
{ foo: 1, bar: 2, baz: 3 }.select { |k, v| !k.in?(%i[foo bar]) }
{ foo: 1, bar: 2, baz: 3 }.select { |k, v| !%i[foo bar].include?(k) }
{ foo: 1, bar: 2, baz: 3 }.select { |k, v| %i[foo bar].exclude?(k) }
# => {:baz=>3}
```

-----------------

* [x] Added tests.


bbatsov:
> I'm not sure if in?, exclude? should be covered by RuboCop, since it is an extension by ActiveSupport.

That's a tricky question. Normally I'd say "no", but ActiveSupport extensions are kind of special and most people use Rails. On the other hand they might have disabled some AS extension...

I don't think there's a high chance for false positives, though, so perhaps you can add some check if `rubocop-rails` is present and check for `in?` an `exclude?` or put this behind a config option.

Darhazer:
Perhaps we need a configuration option whether rubocop cops to account for ActiveSupport methods, so people can just turn it on or off, and cops have common way to extend their own checks based on that.

bbatsov:
> Perhaps we need a configuration option whether rubocop cops to account for ActiveSupport methods, so people can just turn it on or off, and cops have common way to extend their own checks based on that.

Keep in mind that outside of Rails you can require those in a very granular way, though. But I guess a single option will be fine for most people, as I assume it's rarely used outside of Rails.

config/default.yml
```diff
@@ -3724,6 +3724,8 @@ Style/HashExcept:
                  that can be replaced with `Hash#except` method.
   Enabled: pending
   VersionAdded: '1.7'
+  VersionChanged: '<<next>>'
+  ActiveSupportEnabled: false
```
nobuyo, line 3728: I've added `ActiveSupportEnabled` option to switch checking methods extended by ActiveSupport.
The option name might have better alternatives 🤔

bbatsov: @nobuyo I like @Darhazer's idea to have some shared config option for this (perhaps a global config option?), so all the impacted cop can check for the same thing. The name of the option is fine, although if we want to be more specific I guess it should `ActiveSupportExtensionsEnabled` or however they call them these days. I'm fine with the current name as well.

@rubocop/rubocop-core any thoughts on this?

nobuyo: Sorry I was misunderstanding. I'll start over with a new way to create a new mixin.

config/default.yml
```diff
@@ -153,6 +153,8 @@ AllCops:
     rubocop-sequel: [sequel]
     rubocop-rake: [rake]
     rubocop-graphql: [graphql]
+  # Enable/Disable checking the methods extended by ActiveSupport.
```
koic, line 156: That's my two cents.

```suggestion
  # Enable/Disable checking the methods extended by Active Support.
```

> The proper names of Rails components have a space in between the words, like "Active Support". `ActiveRecord` is a Ruby module, whereas Active Record is an ORM. 

https://guides.rubyonrails.org/api_documentation_guidelines.html#wording

nobuyo: Thank you. I've corrected it along with other location.

bbatsov (approved):

bbatsov:
Looks good! We'll have to see which other cops can benefit from this setting and perhaps auto-set it to true in the presence of `rubocop-rails`, but overall we're on the right track.

Mbarnes2022:
All photos need to be linked #state #city #type #number ( 1,2,3,4,....)

PR #10702, opened 2022-06-08
Fix displaying version information for known extensions

Darhazer:
The rubocop extension namespaces not always match the gem name. We already had the mapping done for rubocop-rspec. This adds the mapping for rubocop-md, rubocop-thread_safety and rubocop-graphql.

Note that I also changed is so it would try to check the `RuboCop::<extension>::Version::STRING` constant if the load itself fails. This was needed in order to write the test, but since it also handles the constant not being defined gracefully, I decided there is no harm in the check still being performed.

rubocop-md was the worst offender, as the folder structure doesn't follow the gem name (the files are under `markdown` folder and not `md`). So it required one additional mapping - quite ugly at this point, but since it's the only exception, I hope it's acceptable 🙏 

Any suggestions for further improvement (or avoiding the degradation of the existing code) are welcome

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/version.rb
+++ b/lib/rubocop/version.rb
@@ -9,7 +9,9 @@ module RuboCop
           'rubocop-ast %<rubocop_ast_version>s, ' \
           'running on %<ruby_engine>s %<ruby_version>s %<ruby_platform>s)'
 
-    CANONICAL_FEATURE_NAMES = { 'Rspec' => 'RSpec' }.freeze
+    CANONICAL_FEATURE_NAMES = { 'Rspec' => 'RSpec', 'Graphql' => 'GraphQL', 'Md' => 'Markdown',
+                                'Thread_safety' => 'ThreadSafety' }.freeze
+    EXTENSION_PATH_NAMES = { 'rubocop-md' => 'markdown' }.freeze
 
     # @api private
     def self.version(debug: false, env: nil)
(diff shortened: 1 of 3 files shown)


lib/rubocop/version.rb
```diff
@@ -44,15 +45,17 @@ def self.extension_versions(env)
         next unless (match = loaded_feature.match(/rubocop-(?<feature>.*)/))
 
         feature = match[:feature]
+
+        feature = 'markdown' if feature == 'md'
```
bbatsov, line 49: I'd add a comment here explaining the need for this.

Darhazer: Sure. Btw I'd rather provide a mapping like `CANONICAL_FEATURE_NAMES`, but couldn't think of a good name 🙈 
CANONICAL_PATH_NAMES perhaps?

bbatsov: Or something like `EXTENSION_PATH_NAMES`?

bbatsov (approved):

bbatsov:
Thanks!

lib/rubocop/version.rb
Darhazer: Perhaps we didn't need to include detection of rubocop-md in the first place, as it's not really an extension to rubocop (e.g. unlike the rest doesn't define new cops) 🤔

PR #10715, opened 2022-06-15
Line continuations

bquorning:

-----------------

* [ ] Added tests.


bbatsov:
> Should we allow space at the beginning of line n+1 if line n ends with a space?

What's the reasoning for something like this?

> Who can come up with a better cop name?

Naming is hard. :D I'm okay with the proposed name, but @dvandersluis and @pirj might have better ideas. 

> Is Style the correct department?

I'm on the fence about this one, but I'm leaning towards `Layout`.

bbatsov:
I prefer trailing whitespace myself, but I'm wondering if we shouldn't make this configurable (leading or trailing).

pirj:
Since the cop is triggered by leading spaces, maybe `LineContinuationLeadingSpace`?

bquorning:
>> Should we allow space at the beginning of line n+1 if line n ends with a space?

> What's the reasoning for something like this?

I guess something like https://github.com/rubocop/rubocop/blob/master/spec/rubocop/cli/options_spec.rb#L1491-L1494. But I think it would make the cop too complex for too little benefit.

> configurable

I had a really hard time getting this cop to select correct ranges etc., and I think I would need some help in making it configurable. Perhaps first: @pirj can you help me get numbers on “real world offenses” with leading vs trailing spaces?

pirj:
Here you go:
```
﻿rg --no-ignore --hidden --multiline "['\"][ ]*\\\\[ ]*\\n[ ]+['\"][ ]+"
```
<details>
  <summary>Leading spaces on line continuation</summary>

```
rspec-rails/lib/rspec/rails/matchers/have_http_status.rb
105:            "expected the response to have status code #{expected.inspect}" \
106:              " but it was #{actual.inspect}"
158:            "expected the response to have status code #{pp_expected} but it" \
159:              " was #{pp_actual}"

rspec-mocks/spec/rspec/mocks_spec.rb
29:          pending "JRuby 9.1.7.0 currently generates a circular warning which" \
30:                  " is unrelated to our suite."

rspec-mocks/spec/rspec/mocks/verifying_doubles/naming_spec.rb
17:        "expected #{actual.inspect} to fail expectations as:\n" \
18:          "  #{expected.inspect}, but failed with:\n" \
19:          "  #{@error_message.inspect}"

rspec-mocks/spec/rspec/mocks/message_expectation_string_representation_spec.rb
30:          "expected string representation: #{expected_representation}\n" \
31:          " but got string representation: #{object.to_s}"

rspec-mocks/spec/rspec/mocks/diffing_spec.rb
38:        }.to fail_with("#<Double \"double\"> received :foo with unexpected arguments\n" \
39:          "  expected: (\"some string\\nline2\")\n       got: (\"this other string\")\n" \
49:        }.to fail_with("#<Double \"double\"> received :foo with unexpected arguments\n" \
50:          "  expected: (\"some string\\nline2\", \"some other string\")\n" \
51:          "       got: (\"this other string\")\nDiff:\n@@ -1,3 +1,2 @@\n-some string\\nline2\n-some other string\n+this other string\n")
94:              "#<Double \"double\"> received :foo with unexpected arguments\n" \
95:              "  expected: ({:baz=>:quz, :foo=>:bar}) (keyword arguments)\n" \
96:              "       got: ({:baz=>:quz, :foo=>:bar}) (options hash)"
110:              "#<Double \"double\"> received :foo with unexpected arguments\n" \
111:              "  expected: (:positional, {:keyword=>1}) (keyword arguments)\n" \
112:              "       got: (:positional, {:keyword=>1}) (options hash)"
153:          }.to fail_with("#<Double \"double\"> received :foo with unexpected arguments\n" \
154:            "  expected: (#{collab_inspect})\n" \
155:            "       got: ([])\nDiff:\n@@ #{one_line_header} @@\n-[#{collab_inspect}]\n+[[]]\n")
172:          }.to fail_with("#<Double \"double\"> received :foo with unexpected arguments\n" \
173:            "  expected: (#{collab_description})\n" \
174:            "       got: ([:a, :b])\nDiff:\n@@ #{one_line_header} @@\n-[\"#{collab_description}\"]\n+[[:a, :b]]\n")
200:          }.to fail_with("#<Double \"double\"> received :foo with unexpected arguments\n" \
201:            "  expected: (#{collab_inspect})\n" \
202:            "       got: ([:a, :b])\nDiff:\n@@ #{one_line_header} @@\n-[#{collab_pp}]\n+[[:a, :b]]\n")

puppet/lib/puppet/functions/eyaml_lookup_key.rb
31:        _("'eyaml_lookup_key': one of 'path', 'paths' 'glob', 'globs' or 'mapped_paths' must be declared in hiera.yaml"\
32:              " when using this lookup_key function")

puppet/lib/puppet/interface/action.rb
108:              _("The when_rendering method for the %{face} face %{name} action takes either just one argument,"\
109:                  " the result of when_invoked, or the result plus the %{arg_count} arguments passed to the"\
110:                  " when_invoked block, not a variable number") %
114:              _("The when_rendering method for the %{face} face %{name} action takes either just one argument,"\
115:                  " the result of when_invoked, or the result plus the %{arg_count} arguments passed to the"\
116:                  " when_invoked block, not %{string}") %

puppet/acceptance/lib/puppet/acceptance/windows_utils/package_installer.rb
50:        create_installer_exe = "\"Add-Type"\
51:          " -TypeDefinition (Get-Content #{source_path_on_host} | Out-String)"\
52:          " -Language CSharp"\
53:          " -OutputAssembly #{destination}"\
54:          " -OutputType ConsoleApplication\""

puppet/acceptance/lib/puppet/acceptance/windows_utils/service.rb
39:        create_service_exe = "\"Add-Type"\
40:          " -TypeDefinition (Get-Content #{code_path_win} | Out-String)"\
41:          " -Language CSharp"\
42:          " -OutputAssembly #{service_exe_path_win}"\
43:          " -OutputType ConsoleApplication"\
44:          " -ReferencedAssemblies 'System.ServiceProcess'\""
48:        register_service_with_scm = "\"New-Service"\
49:          " #{config[:name]}"\
50:          " #{service_exe_path_win}"\
51:          " -DisplayName '#{config[:display_name]}'"\
52:          " -Description '#{config[:description]}'"\
53:          " -StartupType Automatic\""
86:          get_property_value = "\"Get-WmiObject -Class Win32_Service"\
87:            " | Where-Object { \\$_.name -eq '#{name}' }"\
88:            " | ForEach-Object { \\$_.#{property} }\""
100:        get_delayed_service = "\"Get-ChildItem HKLM:\\SYSTEM\\CurrentControlSet\\Services"\
101:          " | Where-Object { \\$_.Property -Contains 'DelayedAutoStart' -And \\$_.PsChildName -Like '#{name}' }"\
102:          " | Select-Object -ExpandProperty PSChildName\""

cartodb/lib/carto/styles/presenters/cartocss.rb
35:              "#{cartocss_property[key].map { |v| "    #{v}" }.join("\n")}\n"\
36:              "  }"

cartodb/spec/lib/carto/styles/point_spec.rb
8:          "#layer {\n"\
9:          "  marker-width: 7;\n"\
10:          "  marker-fill: #EE4D5A;\n"\
11:          "  marker-fill-opacity: 0.9;\n"\
12:          "  marker-line-color: #FFFFFF;\n"\
13:          "  marker-line-width: 1;\n"\
14:          "  marker-line-opacity: 1;\n"\
15:          "  marker-placement: point;\n"\
16:          "  marker-type: ellipse;\n"\
17:          "  marker-allow-overlap: true;\n"\

cartodb/spec/lib/carto/styles/presenters/cartocss_spec.rb
23:          "#layer {\n"\
24:          "  color-camisa: blanco;\n"\
25:          "  talla-zapatos: 46;\n"\
26:          "  altura: 186cm;\n"\

cartodb/spec/lib/carto/styles/geometry_spec.rb
8:          "#layer['mapnik::geometry_type'=1] {\n"\
9:          "  marker-width: 7;\n"\
10:          "  marker-fill: #EE4D5A;\n"\
11:          "  marker-fill-opacity: 0.9;\n"\
12:          "  marker-line-color: #FFFFFF;\n"\
13:          "  marker-line-width: 1;\n"\
14:          "  marker-line-opacity: 1;\n"\
15:          "  marker-placement: point;\n"\
16:          "  marker-type: ellipse;\n"\
17:          "  marker-allow-overlap: true;\n"\
19:          "#layer['mapnik::geometry_type'=2] {\n"\
20:          "  line-color: #4CC8A3;\n"\
21:          "  line-width: 1.5;\n"\
22:          "  line-opacity: 1;\n"\
24:          "#layer['mapnik::geometry_type'=3] {\n"\
25:          "  polygon-fill: #826DBA;\n"\
26:          "  polygon-opacity: 0.9;\n"\
27:          "  ::outline {\n"\
28:          "    line-color: #FFFFFF;\n"\
29:          "    line-width: 1;\n"\
30:          "    line-opacity: 0.5;\n"\
31:          "  }\n"\

cartodb/spec/lib/carto/styles/polygon_spec.rb
8:          "#layer {\n"\
9:          "  polygon-fill: #826DBA;\n"\
10:          "  polygon-opacity: 0.9;\n"\
11:          "  ::outline {\n"\
12:          "    line-color: #FFFFFF;\n"\
13:          "    line-width: 1;\n"\
14:          "    line-opacity: 0.5;\n"\
15:          "  }\n"\

cartodb/spec/lib/carto/styles/line_spec.rb
8:          "#layer {\n"\
9:          "  line-color: #4CC8A3;\n"\
10:          "  line-width: 1.5;\n"\
11:          "  line-opacity: 1;\n"\

cartodb/spec/models/carto/helpers/billing_spec.rb
25:      it "returns the current month if period_end_date.day <= today.day" \
26:      " (today = #{today}, period_end_date = #{period_end_date}, expected = #{expected})" do
44:      it "returns the previous month if period_end_date.day > today.day" \
45:      " (today = #{today}, period_end_date = #{period_end_date}, expected = #{expected})" do
62:      it "returns the previous valid day when dealing with 28, 29 or 30-day month corner cases" \
63:      " (today = #{today}, period_end_date = #{period_end_date}, expected = #{expected})" do
91:      it "returns the current month if period_end_date.day > today.day" \
92:      " (today = #{today}, period_end_date = #{period_end_date}, expected = #{expected})" do
110:      it "returns the next month if period_end_date.day <= today.day" \
111:      " (today = #{today}, period_end_date = #{period_end_date}, expected = #{expected})" do
128:      it "returns the previous valid day when dealing with 28, 29 or 30-day month corner cases" \
129:      " (today = #{today}, period_end_date = #{period_end_date}, expected = #{expected})" do

cartodb/app/models/synchronization/member.rb
95:        "url:\"#{@url}\" error_code:\"#{@error_code}\" error_message:\"#{@error_message}\" modified_at:\"#{@modified_at}\" " \
96:        " user_id:\"#{@user_id}\" type_guessing:\"#{@type_guessing}\" " \

diaspora/lib/archive_validator/author_private_key_validator.rb
12:      logger.info "Archive author couldn't be fetched (old home pod is down?), will continue with data"\
13:        " import only"

diaspora/lib/diaspora/shareable.rb
30:          joins("LEFT OUTER JOIN aspect_visibilities ON aspect_visibilities.shareable_id = #{table_name}.id AND "\
31:          " aspect_visibilities.shareable_type = '#{base_class}'")

diaspora/script/diaspora-dev
14:      print_usage_header "setup [options]" \
15:        "    --force    Rebuild image without using Docker's cache;" \
16:        "               overwrite existing configuration" \
17:        "    --mysql    Use MySQL as database (PostgreSQL is default)"
30:      print_usage_header "restart [options]" \
31:        "    --full    Restart entire container"
44:      print_usage_header "clean [options]" \
45:        "    --config    Delete configuration files as well"
76:      print_usage_header "build [options]" \
77:        "    --no-cache    Rebuild image without using Docker's cache"
86:      print_usage_header "config [options]" \
87:        "    --mysql       Use MySQL as database (PostgreSQL is default)" \
88:        "    --overwrite   Overwrite existing configuration"

diaspora/config/load_config.rb
69:    warn "FATAL: Diaspora doesn't know where your certificate authorities are." \
70:         " Please ensure they are set to a valid path in diaspora.yml"

diaspora/app/controllers/users_controller.rb
237:      flash.now[:error] = "Your account migration could not be scheduled for the following reason:"\
238:                          " #{@user.errors.full_messages}"

diaspora/app/models/account_migration.rb
199:      .joins("INNER JOIN contacts as c2 ON (contacts.user_id = c2.user_id AND contacts.person_id=#{old_person.id} AND"\
200:        " c2.person_id=#{newest_person.id})")
207:        "AND likes.author_id=#{old_person.id} AND"\
208:        " l2.author_id=#{newest_person.id})")
215:        "AND participations.author_id=#{old_person.id} AND"\
216:        " p2.author_id=#{newest_person.id})")
222:        "AND poll_participations.author_id=#{old_person.id} AND"\
223:        " p2.author_id=#{newest_person.id})")
232:      .joins("INNER JOIN contacts as c2 ON (contacts.person_id = c2.person_id AND contacts.user_id=#{old_user.id} AND"\
233:        " c2.user_id=#{newest_user.id})")
236:      .joins("INNER JOIN tag_followings as t2 ON (tag_followings.tag_id = t2.tag_id AND"\
237:        " tag_followings.user_id=#{old_user.id} AND t2.user_id=#{newest_user.id})")

diaspora/app/mailers/notification_mailers/base.rb
55:      log_string = "event=mail mail_type=#{type} recipient_id=#{recipient_id} sender_id=#{sender_id} " \
56:                   " recipient_handle=#{@recipient.diaspora_handle}"

capybara/lib/capybara/queries/base_query.rb
82:          message << " between #{between.begin ? between.first : 1} and" \
83:                     " #{between.end ? between.last : 'infinite'} times"

capybara/spec/selenium_spec_chrome_remote.rb
24:            "You can run a selenium server easily with: \n" \
25:            '  $ docker-compose up -d selenium_chrome'

capybara/spec/selenium_spec_firefox_remote.rb
24:            "You can run a selenium server easily with: \n" \
25:            '  $ docker-compose up -d selenium_firefox'

discourse/lib/onebox/sanitize_config.rb
26:          'sandbox' => 'allow-same-origin allow-scripts allow-forms allow-popups allow-popups-to-escape-sandbox' \
27:                       ' allow-presentation',


rubocop/lib/rubocop/cop/corrector.rb
112:              "name:#{buffer.name.inspect}" \
113:              " is not current #{@source_buffer.object_id} " \

rubocop/lib/rubocop/cop/style/multiline_if_modifier.rb
22:        MSG = 'Favor a normal %<keyword>s-statement over a modifier' \
23:              ' clause in a multiline statement.'

rubocop/lib/rubocop/cop/style/implicit_runtime_error.rb
18:        MSG = 'Use `%<method>s` with an explicit exception class and message,' \
19:              ' rather than just a message.'

rubocop/lib/rubocop/cop/layout/multiline_method_call_indentation.rb
59:                'The `Layout/MultilineMethodCallIndentation`' \
60:                ' cop only accepts an `IndentationWidth` ' \

rubocop/lib/rubocop/cop/layout/first_array_element_indentation.rb
166:            'Indent the right bracket the same as the start of the line' \
167:              ' where the left bracket is.'

rubocop/lib/rubocop/cop/layout/multiline_operation_indentation.rb
60:          raise ValidationError, 'The `Layout/MultilineOperationIndentation`' \
61:                                 ' cop only accepts an `IndentationWidth` ' \

rubocop/lib/rubocop/cop/lint/struct_new_override.rb
25:        MSG = '`%<member_name>s` member overrides `Struct#%<method_name>s`' \
26:              ' and it may be unexpected.'

rubocop/lib/rubocop/cop/lint/regexp_as_condition.rb
22:        MSG = 'Do not use regexp literal as a condition.' \
23:              ' The regexp literal matches `$_` implicitly.'

rubocop/lib/rubocop/config_validator.rb
264:        "Property #{Rainbow(key).yellow} of cop #{Rainbow(parent).yellow}" \
265:        " is supposed to be a boolean and #{Rainbow(value).yellow} is not."

rubocop/lib/rubocop/formatter/simple_text_formatter.rb
127:              "#{files} inspected, #{offenses} detected, #{corrections} corrected,"\
128:                " #{correctable}"

rubocop/spec/rubocop/cop/corrector_spec.rb
97:                             'Corrector expected range source buffer to be'\
98:                             ' a Parser::Source::Buffer, but got String')

rubocop/spec/rubocop/cop/commissioner_spec.rb
121:      it 'passes the input params to all cops/forces that implement their own' \
122:         ' #investigate method' do

rubocop/spec/rubocop/cop/style/sole_nested_conditional_spec.rb
251:  it 'registers an offense and corrects when using `unless` and `||` and parens in the outer condition' \
252:     ' and nested modifier condition' do
267:  it 'registers an offense and corrects when using `unless` and `||` without parens in the outer condition' \
268:     ' and nested modifier condition' do
283:  it 'registers an offense and corrects when using `unless` and `&&` without parens in the outer condition' \
284:     ' and nested modifier condition' do

rubocop/spec/rubocop/cop/style/redundant_self_spec.rb
215:    it 'accepts a self receiver used to distinguish from an argument' \
216:       ' when an inner method is defined' do

rubocop/spec/rubocop/cop/style/trailing_comma_in_arguments_spec.rb
49:    it 'accepts method call without trailing comma with single element hash' \
50:       ' parameters at the end' do
85:      it 'registers an offense for trailing comma in a method call with hash' \
86:         ' parameters at the end' do
319:      it 'registers an offense for no trailing comma in a method call with' \
320:         ' hash parameters at the end' do
370:      it 'accepts no trailing comma in a method call with a multiline' \
371:         ' braceless hash at the end with more than one parameter on a line' do
448:      it 'registers an offense for no trailing comma in a method call with' \
449:         ' hash parameters at the end' do

rubocop/spec/rubocop/cop/layout/line_length_spec.rb
400:      it "registers an offense for a line that's including 2 tab with size 2" \
401:         ' and 28 other characters' do

rubocop/spec/rubocop/cop/lint/useless_assignment_spec.rb
572:  context 'when a variable is reassigned and unreferenced in a if branch' \
573:          ' while the variable is referenced in the paired else branch' do

rubocop/spec/rubocop/cop/lint/shadowed_exception_spec.rb
203:    it 'registers an offense when a higher level exception is rescued before' \
204:       ' a lower level exception' do

rubocop/spec/rubocop/cli/options_spec.rb
877:        .to eq(["For #{abs('')}:" \
878:                " Default configuration from #{home}/config/default.yml"])
1347:            '                                                   ' \
1348:            '                                                   ' \
1349:            '                  ^^^^^^^^^^',

rubocop/spec/rubocop/cli_spec.rb
816:                                          '(for `Layout/SpaceAroundOperators`)' \
817:                                          ' found')
850:          'Error: The `Layout/MultilineOperationIndentation` cop only accepts' \
851:          ' an `IndentationWidth` configuration parameter when ' \

lobsters/app/models/keystore.rb
99:    exception = ActiveRecord::ValueTooLong.new("#{MAX_KEY_LENGTH}" \
100:      " characters is the maximum allowed for key")

rspec-core/lib/rspec/core/configuration.rb
227:            RSpec.warning "Cannot set `RSpec.configuration.fail_fast`" \
228:              " to `#{value.inspect}`. Only `true`, `false`, `nil` and integers" \
229:              " are valid values."

rspec-core/spec/rspec/core/metadata_spec.rb
222:            "expected #{group_or_example.inspect}\n" \
223:            "   to have id: #{expected_id}\n" \
224:            "   but had id: #{group_or_example.id}\n" \
225:            "   and have scoped id: #{scoped_id}\n" \
226:            "   but had  scoped id: #{group_or_example.metadata[:scoped_id]}"

administrate/spec/lib/administrate/search_spec.rb
177:          'LOWER(CAST("roles"."name" AS CHAR(256))) LIKE ?'\
178:          ' OR LOWER(CAST("people"."first_name" AS CHAR(256))) LIKE ?'\
179:          ' OR LOWER(CAST("people"."last_name" AS CHAR(256))) LIKE ?'\
180:          ' OR LOWER(CAST("addresses"."street" AS CHAR(256))) LIKE ?',

brew/Library/Homebrew/vendor/bundle/ruby/2.6.0/gems/ruby-macho-3.0.0/lib/macho/exceptions.rb
86:      super "Unrecognized CPU sub-type: 0x%08<cpusubtype>x" \
87:            " (for CPU type: 0x%08<cputype>x" % { :cputype => cputype, :cpusubtype => cpusubtype }
122:      super "Expected #{expected_arity} arguments for #{cmd_sym} creation," \
123:            " got #{actual_arity}"
139:      super "Load command #{lc.type} at offset #{lc.view.offset} contains a" \
140:            " malformed string"
206:      super "Offset #{offset} exceeds the 32-bit width of a fat_arch offset." \
207:            " Consider merging with `fat64: true`"

brew/Library/Homebrew/vendor/bundle/ruby/2.6.0/gems/bindata-2.4.10/lib/bindata/base_primitive.rb
106:          "def #{symbol}(*args, &block);" \
107:          "  snapshot.#{symbol}(*args, &block);" \

brew/Library/Homebrew/vendor/bundle/ruby/2.6.0/gems/zeitwerk-2.5.4/lib/zeitwerk/loader.rb
482:                "loader\n\n#{pretty_inspect}\n\nwants to manage directory #{dir.chop}," \
483:                " which is already managed by\n\n#{loader.pretty_inspect}\n"

brew/Library/Homebrew/env_config.rb
81:                     "`~/.profile`, `~/.bash_profile`, or `~/.zshenv`:\n\n" \
82:                     '    `export HOMEBREW_CASK_OPTS="--appdir=~/Applications --fontdir=/Library/Fonts"`',
247:                     "installing, upgrading or reinstalling formulae. This will result in fewer dependents " \
248:                     " (and their dependencies) being upgraded or reinstalled but may result in more breakage " \

brew/Library/Homebrew/utils/github.rb
552:        ohai "git commit --no-edit --verbose --message='#{commit_message}'" \
553:             " -- #{changed_files.join(" ")}"

brew/Library/Homebrew/test/utils/curl_spec.rb
65:        "set-cookie"      => "__cf_bm=0123456789abcdef; path=/; expires=Wed, 31-Jan-20 01:23:45 GMT;" \
66:                             " domain=www.example.com; HttpOnly; Secure; SameSite=None",
80:      "__cf_bm=abcdef0123456789; path=/; expires=Thu, 28-Apr-22 18:38:40 GMT; domain=www.example.com; HttpOnly;" \
81:      " Secure; SameSite=None",

brew/Library/Homebrew/test/rubocops/cask/stanza_order_spec.rb
112:        source:   "uninstall :quit => 'com.example.foo',\n" \
113:                  "            :kext => 'com.example.foo.kext'",

brew/Library/Homebrew/rubocops/options.rb
30:              problem "Options should begin with with/without."\
31:                      " Migrate '--#{option}' with `deprecated_option`."
37:            problem "Use '--with#{Regexp.last_match(1)}-test' instead of '--#{option}'."\
38:                    " Migrate '--#{option}' with `deprecated_option`."

chef/lib/chef/runner.rb
164:      Chef::Log.info( "#{notification.notifying_resource} sending #{notification.action}"\
165:                      " action to #{notification.resource} (delayed)")

chef/lib/chef/resource/package.rb
27:      description "Use the **package** resource to manage packages. When the package is"\
28:                  " installed from a local file (such as with RubyGems, dpkg, or RPM"\
29:                  " Package Manager), the file must be added to the node using the remote_file"\
30:                  " or cookbook_file resources.\n\nThis resource is the base resource for"\
31:                  " several other resources used for package management on specific platforms."\
32:                  " While it is possible to use each of these specific resources, it is"\
33:                  " recommended to use the package resource as often as possible."

chef/lib/chef/resource/link.rb
31:                  "A symbolic link—sometimes referred to as a soft link—is a directory entry"\
32:                  " that associates a file name with a string that contains an absolute or"\
33:                  " relative path to a file on any file system. In other words, 'a file that"\
34:                  " contains a path that points to another file.' A symbolic link creates a new"\
35:                  " file with a new inode that points to the inode location of the original file.\n\n"\
36:                  "A hard link is a directory entry that associates a file with another file in the"\
37:                  " same file system. In other words, 'multiple directory entries to the same file.'"\
38:                  " A hard link creates a new file that points to the same inode as the original file."

chef/lib/chef/resource/log.rb
35:      description "Use the **log** resource to create log entries. The log resource behaves"\
36:                  " like any other resource: built into the resource collection during the"\
37:                  " compile phase, and then run during the execution phase. (To create a log"\
38:                  " entry that is not built into the resource collection, use Chef::Log instead"\
39:                  " of the log resource.)"

chef/lib/chef/resource/sudo.rb
30:      description "Use the **sudo** resource to add or remove individual sudo entries using sudoers.d files."\
31:                  " Sudo version 1.7.2 or newer is required to use the sudo resource, as it relies on the"\
32:                  " `#includedir` directive introduced in version 1.7.2. This resource does not enforce"\
33:                  " installation of the required sudo version. Chef-supported releases of Ubuntu, SuSE, Debian,"\
34:                  " and RHEL (6+) all support this feature."

chef/lib/chef/resource/perl.rb
32:      description "Use the **perl** resource to execute scripts using the Perl interpreter."\
33:                  " This resource may also use any of the actions and properties that are"\
34:                  " available to the **execute** resource. Commands that are executed with this"\
35:                  " resource are (by their nature) not idempotent, as they are typically"\
36:                  " unique to the environment in which they are run. Use `not_if` and `only_if`"\
37:                  " to guard this resource for idempotence."

chef/lib/chef/resource/csh.rb
27:      description "Use the **csh** resource to execute scripts using the csh interpreter."\
28:                  " This resource may also use any of the actions and properties that are"\
29:                  " available to the **execute** resource. Commands that are executed with this"\
30:                  " resource are (by their nature) not idempotent, as they are typically"\
31:                  " unique to the environment in which they are run. Use `not_if` and `only_if`"\
32:                  " to guard this resource for idempotence."

chef/lib/chef/resource/mdadm.rb
28:      description "Use the **mdadm** resource to manage RAID devices in a Linux environment using the mdadm utility. The mdadm resource"\
29:                  " will create and assemble an array, but it will not create the config file that is used to persist the array upon"\
30:                  " reboot. If the config file is required, it must be done by specifying a template with the correct array layout,"\
31:                  " and then by using the mount provider to create a file systems table (fstab) entry."

chef/lib/chef/resource/python.rb
31:      description "Use the **python** resource to execute scripts using the Python interpreter."\
32:                  " This resource may also use any of the actions and properties that are available"\
33:                  " to the **execute** resource. Commands that are executed with this resource are (by"\
34:                  " their nature) not idempotent, as they are typically unique to the environment in"\
35:                  " which they are run. Use `not_if` and `only_if` to guard this resource for idempotence."

chef/lib/chef/resource/reboot.rb
28:      description "Use the **reboot** resource to reboot a node, a necessary step with some"\
29:                  " installations on certain platforms. This resource is supported for use on"\
30:                  " the Microsoft Windows, macOS, and Linux platforms.\n"\
31:                  "In using this resource via notifications, it's important to *only* use"\
32:                  " immediate notifications. Delayed notifications produce unintuitive and"\
33:                  " probably undesired results."

chef/lib/chef/resource/script.rb
30:      description "Use the **script** resource to execute scripts using a specified interpreter, such as Bash, csh, Perl, Python, or Ruby."\
31:                  " This resource may also use any of the actions and properties that are available to the **execute** resource. Commands"\
32:                  " that are executed with this resource are (by their nature) not idempotent, as they are typically unique to the"\
33:                  " environment in which they are run. Use `not_if` and `only_if` to guard this resource for idempotence."

chef/lib/chef/resource/directory.rb
30:      description "Use the **directory** resource to manage a directory, which is a hierarchy"\
31:                  " of folders that comprises all of the information stored on a computer."\
32:                  " The root directory is the top-level, under which the rest of the directory"\
33:                  " is organized. The directory resource uses the name property to specify the"\
34:                  " path to a location in a directory. Typically, permission to access that"\
35:                  " location in the directory is required."

chef/lib/chef/resource/ksh.rb
27:      description "Use the **ksh** resource to execute scripts using the Korn shell (ksh)"\
28:                  " interpreter. This resource may also use any of the actions and properties"\
29:                  " that are available to the **execute** resource. Commands that are executed"\
30:                  " with this resource are (by their nature) not idempotent, as they are"\
31:                  " typically unique to the environment in which they are run. Use `not_if`"\
32:                  " and `only_if` to guard this resource for idempotence."

chef/lib/chef/application/exit_code.rb
142:          "#{ChefUtils::Dist::Infra::CLIENT} attempted to exit with a non-standard exit code of #{exit_code}." \
143:          " The #{ChefUtils::Dist::Infra::PRODUCT} Exit Codes design document (https://github.com/chef/chef/blob/main/docs/dev/design_documents/client_exit_codes.md)" \
144:          " defines the exit codes that should be used with #{ChefUtils::Dist::Infra::CLIENT}. Chef::Application::ExitCode defines"  \
145:          " valid exit codes Non-standard exit codes are redefined as GENERIC_FAILURE."

chef/lib/chef/run_context.rb
321:        logger.info( "#{notification.notifying_resource} not queuing delayed action #{notification.action} on #{notification.resource}"\
322:                       " (delayed), as it's already been queued")

chef/lib/chef/provider/dsc_resource.rb
59:          err = ["The LCM must have its RefreshMode set to Disabled for" \
60:                 " PowerShell versions before 5.0.10586.0."]
165:        switches = "-Method #{method} -Name #{new_resource.resource}"\
166:                   " -Property #{properties} -Module #{module_info_object} -Verbose"

chef/lib/chef/exceptions.rb
315:            super("The managed identity could not be found. This could mean one of the following things:\n\n" \
316:                  "  1. The VM has no system or user assigned identities.\n" \
317:                  "  2. The managed identity object_id or client_id that was specified is not assigned to the VM.\n")

publify/lib/publify_textfilter_flickr.rb
80:            "<img src=\"#{imageurl}\" width=\"#{width}\" height=\"#{height}\"" \
81:            " alt=\"#{alt}\" title=\"#{title}\"/></a>#{captioncode}</div>"
84:          "<div class='broken_flickr_link'>`#{img}' could not be displayed because:" \
85:            " <br />#{CGI.escapeHTML(e.message)}</div>"

publify/spec/controllers/articles_controller_spec.rb
231:              to match("in markdown format * we * use [ok](http://blog.ok.com)" \
232:                       " to define a link")

publify/spec/lib/publify_textfilter_lightbox_spec.rb
77:        filter_text('<publify:lightbox img="31366117" thumbsize="Thumbnail"' \
78:                    ' displaysize="Large" style="float:left"/>')
80:        '<a href="//photos23.flickr.com/31366117_b1a791d68e_b.jpg"' \
81:        ' data-toggle="lightbox" title="Matz">' \
82:        '<img src="//photos23.flickr.com/31366117_b1a791d68e_t.jpg"' \
83:        ' width="67" height="100" alt="Matz" title="Matz"/></a>' \
91:        '<a href="//photos23.flickr.com/31366117_b1a791d68e_b.jpg"' \
92:        ' data-toggle="lightbox" title="Matz">' \
93:        '<img src="//photos23.flickr.com/31366117_b1a791d68e_s.jpg"' \
94:        ' width="75" height="75" alt="Matz" title="Matz"/></a>' \
101:        '<a href="//photos23.flickr.com/31366117_b1a791d68e_o.jpg"' \
102:        ' data-toggle="lightbox" title="Matz">' \
103:        '<img src="//photos23.flickr.com/31366117_b1a791d68e_s.jpg"' \
104:        ' width="75" height="75" alt="Matz" title="Matz"/></a>' \
111:        '<a href="//photos23.flickr.com/31366117_b1a791d68e_o.jpg"' \
112:        ' data-toggle="lightbox" title="Matz">' \
113:        '<img src="//photos23.flickr.com/31366117_b1a791d68e_s.jpg"' \
114:        ' width="75" height="75" alt="Matz" title="Matz"/></a>'

publify/spec/lib/publify_textfilter_flickr_spec.rb
81:        '<img src="//photos23.flickr.com/31366117_b1a791d68e_s.jpg"' \
82:        ' width="75" height="75" alt="Matz" title="Matz"/></a>' \
92:        '<img src="//photos23.flickr.com/31366117_b1a791d68e_s.jpg"' \
93:        ' width="75" height="75" alt="Matz" title="Matz"/></a>' \
103:        '<img src="//photos23.flickr.com/31366117_b1a791d68e_s.jpg"' \
104:        ' width="75" height="75" alt="Matz" title="Matz"/></a></div>'
125:        '<img src="//photos23.flickr.com/31366117_b1a791d68e_s.jpg"' \
126:        ' width="75" height="75" alt="Matz" title="Matz"/></a>' \

publify/publify_textfilter_code/spec/lib/publify_app/textfilter_code_spec.rb
39:                    "foo-code</span></pre></div> blah blah" \
40:                    ' <div class="CodeRay"><pre><span class="CodeRay">' \

publify/app/models/authors_sidebar.rb
5:  description "Displays a list of authors ordered by name with links to their" \
6:              " articles and profile"

publify/publify_core/lib/publify_textfilter_markdown_smartquotes.rb
9:    plugin_description "Markdown markup language from" \
10:                       ' <a href="http://daringfireball.com/">Daring Fireball</a>' \
11:                       " with automatic use of typographically correct quotes and dashes"

publify/publify_core/lib/publify_textfilter_markdown.rb
11:      plugin_description "Markdown markup language from" \
12:                         ' <a href="http://daringfireball.com/">Daring Fireball</a>'

publify/publify_core/spec/views/feedback/index.rss.builder_spec.rb
45:          to eq "Trackback from #{trackback.blog_name}:" \
46:                " #{trackback.title} on #{article.title}"

publify/publify_core/spec/views/feedback/index.atom.builder_spec.rb
53:          to eq "Trackback from #{trackback.blog_name}:" \
54:                " #{trackback.title} on #{article.title}"

publify/publify_core/spec/views/articles/index_rss_feed_spec.rb
132:          to eq "<p>This article is password protected. Please" \
133:                " <a href='#{@article.permalink_url}'>fill in your password</a>" \
134:                " to read it</p>"
147:          to eq "<p>This article is password protected. Please" \
148:                " <a href='#{@article.permalink_url}'>fill in your password</a>" \
149:                " to read it</p>"

publify/publify_core/spec/views/articles/index_atom_feed_spec.rb
139:          to eq "<p>This article is password protected. Please" \
140:                " <a href='#{@article.permalink_url}'>fill in your password</a>" \
141:                " to read it</p>"
158:          to eq "<p>This article is password protected. Please" \
159:                " <a href='#{@article.permalink_url}'>fill in your password</a>" \
160:                " to read it</p>"

publify/publify_core/spec/lib/publify_textfilter_twitterfilter_spec.rb
35:              "&amp;src=tren&amp;mode=realtime\">#hashtag</a> and a" \
36:              " <a href=\"https://twitter.com/mention\">@mention</a>")

publify/publify_core/spec/models/note_spec.rb
186:          "A very big(10) message with lot of text (40)inside just to try the" \
187:            " shortener and (80)the new link that publify must create and add at the end"
190:          "A very big(10) message with lot of text (40)inside just to try the" \
191:            " shortener and (80)the new link that publify... (#{note.redirect.from_url})"
200:          "Le dojo de nantes, c'est comme au McDo, sans les odeurs, et en plus rigolo:" \
201:            " RT @abailly Ce midi c'est coding dojo à la Cantine #Nantes." \
202:            " Pour s'inscrire si vous voulez c'est ici:" \
203:            " http://cantine.atlantic2.org/evenements/coding-dojo-8/ …" \
204:            " Sinon venez comme vous êtes"
207:          "Le dojo de nantes, c'est comme au McDo, sans les odeurs, et en plus rigolo:" \
208:            " RT @abailly Ce midi c'est coding... (#{note.redirect.from_url})"
217:          '"JSFuck is an esoteric and educational programming style based on the' \
218:            " atomic parts of JavaScript. It uses only six different characters to" \
219:            ' write and execute code." http://www.jsfuck.com/ '
222:          "\"JSFuck is an esoteric and educational programming style based on the" \
223:            " atomic parts of JavaScript. It uses only... (#{note.redirect.from_url})"
232:          "Le #mobprogramming c'est un peu comme faire un dojo sur une journée entière" \
233:            " (ça permet sûrement de faire des petites journées ;-))"
236:          "Le #mobprogramming c'est un peu comme faire un dojo sur une journée entière" \
237:            " (ça permet sûrement de faire des... (#{note.redirect.from_url})"
246:          "RT @stephaneducasse http://pharocloud.com is so cool. I love love such idea" \
247:            " and I wish them success. Excellent work."
250:          "RT @stephaneducasse http://pharocloud.com is so cool. I love love such idea" \
251:            " and I wish them success. Excellent... (#{note.redirect.from_url})"

publify/publify_core/spec/models/text_filter_spec.rb
45:          "<p><a href=\"http://daringfireball.net/projects/markdown/\">Markdown</a>" \
46:          " is a simple")

publify/publify_core/spec/models/static_sidebar_spec.rb
18:      to eq "Static content, like links to other sites, advertisements," \
19:            " or blog meta-information"

publify/publify_core/app/models/meta_sidebar.rb
4:  description "This widget just displays links to Publify main site," \
5:              " this blog's admin and RSS."

publify/publify_core/app/models/static_sidebar.rb
15:  description "Static content, like links to other sites, advertisements," \
16:              " or blog meta-information"

canvas-lms/gems/tatl_tael/README.md
13:  puts "this will be printed if there are ruby additions or modifications,"\
14:       " but no spec additions or modifications."

canvas-lms/gems/plugins/academic_benchmark/spec_canvas/academic_benchmark_spec.rb
71:          description: "Describe the importance of historiography, which includes how historical knowledge is obtained" \
72:                       " and transmitted, when interpreting events in history."
77:          description: "Utilize a variety of primary and secondary sources to identify author, historical significance," \
78:                       " audience, and authenticity to understand a historical period."
88:          description: "Analyze how images, symbols, objects, cartoons, graphs, charts, maps, and artwork may be used" \
89:                       " to interpret the significance of time periods and events from the past."
114:          description: "Analyze the transformation of the American economy and the changing social and" \
115:                       " political conditions in response to the Industrial Revolution."
130:          description: "Understand the causes and course of World War II, the character of the war at home and abroad," \
131:                       " and its reshaping of the United States role in the post-war world."
136:          description: "Understand the rise and continuing international influence of the United States as a" \
137:                       " world leader and the impact of contemporary social and political movements on American life."

canvas-lms/gems/plugins/academic_benchmark/spec_canvas/academic_benchmark/converter_spec.rb
79:           "Locating on a map major battle sites of the American Revolution," \
80:           " including the battles of Lexington and Concord, Bunker Hill, Saratoga, and Yorktown" },

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/no_wait_for_no_such_element.rb
24:        MSG = "Avoid using wait_for_no_such_element. Instead, use"\
25:              " not_to contain_css/contain_link.\n"\

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/no_selenium_web_driver_wait.rb
25:              "Our finders (f/fj and ff/ffj) will wait up to the implicit wait" \
26:              " (just like find_element, etc), and will raise a" \
27:              " Selenium::WebDriver::Error::NoSuchElementError" \
28:              " (just like find_element, etc).\n" \
29:              "Look through custom_selenium_rspec_matchers.rb" \
30:              " and custom_wait_methods.rb."

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/no_before_once_stubs.rb
24:        MSG = "Stubs in a `before(:once)` block won't carry over"\
25:              " to the examples; you should move this to a `before(:each)`"

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/no_execute_script.rb
24:        MSG = "Avoid using execute_script. Instead, perform actual"\
25:              " user interactions such as click/keypress. If these"\
26:              " seem insufficient, consider converting your"\
27:              " integration spec into a JavaScript unit test."

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/ensure_spec_extension.rb
29:        MSG = "Spec files need to end with \"_spec.rb\""\
30:              " for rspec  to find and run them."

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/no_strftime.rb
25:        MSG = "Avoid using strftime." \
26:              " Use format_date_for_view or format_time_for_view instead."

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/no_skip_without_ticket.rb
26:        MSG = "Reference a ticket if skipping."\
27:              " Example: skip('time bomb on saturdays CNVS-123456')."

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/prefer_f_over_fj.rb
50:          "Prefer `#{alternative}` instead of `#{method}`; `#{method}`"\
51:            " should only be used if you are doing jquery-fake-css selectors"\
52:            " (e.g. `:visible`)"

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/no_disable_implicit_wait.rb
25:              "Look through custom_selenium_rspec_matchers.rb" \
26:              " and custom_wait_methods.rb."

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/no_no_such_element_error.rb
25:              "Our finders (f/fj and ff/ffj) will wait up to the implicit wait" \
26:              " (just like find_element, etc), and will raise a" \
27:              " Selenium::WebDriver::Error::NoSuchElementError" \
28:              " (just like find_element, etc).\n" \
29:              "Look through custom_selenium_rspec_matchers.rb, particularly" \
30:              " contain_css and contain_jqcss."

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/specs/scope_helper_modules.rb
24:        MSG = "Define all helper and factory methods within modules"\
25:              " (or `shared_context`). Otherwise they will live on Object"\
26:              " and potentially wreak havoc on other specs."

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/migration/delay.rb
26:        PREDEPLOY_MSG = "`delay` cannot be used in a"\
27:                        " predeploy migration, since job servers won't"\
28:                        " have the new code yet"

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/lint/no_file_utils_rm_rf.rb
24:        MSG = "In order to enable spec parallelization, avoid FileUtils.rm_rf"\
25:              " and making persistent files/directories. Instead use"\
26:              " Dir.mktmpdir. See https://gerrit.instructure.com/#/c/73834"\
27:              " for the pattern you should follow."

canvas-lms/gems/rubocop-canvas/lib/rubocop_canvas/cops/lint/no_sleep.rb
27:        SPEC_MSG = "Avoid using sleep. Depending on what you are trying to do,"\
28:                   " you should instead consider: Timecop,"\
29:                   " vanilla `f` calls (since they wait),"\
30:                   " the `become` matcher, `wait_for_ajaximations`, or `keep_trying_until`."

canvas-lms/lib/sis/user_importer.rb
534:        "Could not save the user with user_id: '#{options[:user_id]}'." \
535:          " #{reason}"

canvas-lms/spec/views/accounts/settings.html.erb_spec.rb
542:        "select#account_settings_smart_alerts_threshold" \
543:        "  option[value=\"#{value}\"][selected]"

canvas-lms/spec/lib/outcomes/learning_outcome_group_children_spec.rb
324:          description: "Explain anticipated meaning, recognize relationships, and draw conclusions; self-correct"\
325:                       " understanding using a variety of strategies [including rereading for story sense]."
331:          description: "apply their growing knowledge of root words, prefixes and suffixes (etymology and morphology)"\
332:                       " as listed in English Appendix 1, both to read aloud and to understand the meaning of new wor"\

canvas-lms/spec/models/submission_spec.rb
460:        it "does not change if an override is added for the student and the due date is earlier" \
461:           " than an existing override that applies to the student for the assignment" do
767:    it "returns seconds_late_override if the submission has a late_policy_status of 'late'" \
768:       " and a seconds_late_override" do
776:    it "is not adjusted if the student resubmits and the submission has a late_policy_status of 'late'" \
777:       " and a seconds_late_override" do
801:    it "is zero if it was turned in late but the teacher sets the late_policy_status to 'late'" \
802:       " and sets seconds_late_override to zero" do

canvas-lms/spec/selenium/quizzes/quizzes_restrictions_student_spec.rb
71:      expect(f("#content")).to include_text "This quiz is protected and is only available from certain locations."\
72:                                            " The computer you are currently using does not appear to be at a valid location for taking this quiz."

canvas-lms/spec/selenium/quizzes/quizzes_grading_teacher_spec.rb
58:          expect(fj(".ui-dialog:visible .alert")).to include_text "Choose a regrade option" \
59:                                                                  " for students who have already taken the quiz"
91:          expect(driver.switch_to.alert.text).to eq "Are you sure? Deleting answers from a question with" \
92:                                                    " submissions disables the option to regrade this question."
98:          expect(fj(".ui-dialog:visible .regrade_option_text")).to include_text "Regrading is not allowed" \
99:                                                                                " on this question because either an answer was removed or the" \
100:                                                                                " question type was changed after a student completed a submission."

canvas-lms/spec/selenium/outcomes/outcome_spec.rb
72:          expect(f("#calculation_int_example")).to include_text("Most recent result counts as 65%"\
73:                                                                " of mastery weight, average of all other results count"\
74:                                                                " as 35% of weight. If there is only one result, the single score"\
75:                                                                " will be returned.")
104:          expect(f("#calculation_int_example")).to include_text("Must achieve mastery at least 5 times."\
105:                                                                " Scores above mastery will be averaged"\
106:                                                                " to calculate final score")

canvas-lms/spec/selenium/rcs/discussion_group_submit_spec.rb
54:      expect(f("#discussion_container").text).to include("Since this is a group discussion,"\
55:                                                         " each group has its own conversation for this topic."\
56:                                                         " Here are the ones you have access to:\nsome group")
65:      expect(f("#discussion_container").text).to include("Since this is a group discussion,"\
66:                                                         " each group has its own conversation for this topic."\
67:                                                         " Here are the ones you have access to:\nsome group")

canvas-lms/spec/selenium/miscellaneous/content_security_policy_spec.rb
53:      @csp_course.wiki_pages.create!(title: "Page1", body: "<iframe width=\"560\" height=\"315\""\
54:                                                           " src=\"https://www.youtube.com/embed/dQw4w9WgXcQ\" frameborder=\"0\""\
55:                                                           " allow=\"accelerometer; autoplay; encrypted-media; gyroscope;"\
56:                                                           " picture-in-picture\" allowfullscreen></iframe>")

canvas-lms/spec/selenium/discussions/discussion_group_submit_spec.rb
66:      expect(f("#discussion_container").text).to include("Since this is a group discussion,"\
67:                                                         " each group has its own conversation for this topic."\
68:                                                         " Here are the ones you have access to:\nsome group")
77:      expect(f("#discussion_container").text).to include("Since this is a group discussion,"\
78:                                                         " each group has its own conversation for this topic."\
79:                                                         " Here are the ones you have access to:\nsome group")

canvas-lms/spec/selenium/discussions/rcs/discussions_show_page_specific_user_spec.rb
164:        expect(f("#discussion_container").text).to include("Since this is a group discussion,"\
165:                                                           " each group has its own conversation for this topic."\
166:                                                           " Here are the ones you have access to:\nsome group")
205:        expect(f("#discussion_container").text).to include("Since this is a group discussion,"\
206:                                                           " each group has its own conversation for this topic."\
207:                                                           " Here are the ones you have access to:\nsome group")

canvas-lms/spec/selenium/grades/speedgrader/speedgrader_spec.rb
112:        Rails.logger.debug "SpeedGrader for course #{@course.id} and assignment"\
113:                           " #{@quiz.assignment_id} loaded in #{page_load_time.real} seconds"

canvas-lms/app/views/authentication_providers/_ldap_settings_test.html.erb
36:                                       "Are all Canvas IP address allowed to"\
37:                                       " access your server?")
71:      "Verify that the user object in LDAP has search privileges for the"\
72:      " provided search base.")

canvas-lms/app/views/authentication_providers/_ldap_fields.html.erb
22:        "If Canvas times out too many times communicating with the LDAP server,"\
23:          " it will block further login attempts from using that LDAP "\

canvas-lms/app/models/assessment_question.rb
141:        logger.error("Error while cloning attachment during"\
142:                     " AssessmentQuestion#translate_links: "\

canvas-lms/app/models/learning_outcome_result.rb
149:      .where("(ra.id IS NULL AND qa.id IS NULL AND sa.id IS NULL)"\
150:             " OR submissions.posted_at IS NOT NULL"\
151:             " OR ra.grading_type = 'not_graded'"\
152:             " OR qa.grading_type = 'not_graded'"\
153:             " OR sa.grading_type = 'not_graded'"\
154:             " OR pc.id IS NULL"\
155:             " OR (pc.id IS NOT NULL AND pc.post_manually = False)")

refinerycms/bin/rails
9:       "To create one in spec/dummy, please run: \n\n" \
10:       "    rake refinery:testing:dummy_app\n\n" \

refinerycms/core/lib/refinery/extension_generation.rb
353:        exit_with_message! "You can't use '--extension #{options[:extension]}' because an" \
354:                           " extension with the name '#{options[:extension]}' doesn't exist."

hound/spec/models/job_failure_spec.rb
12:            "cd29f72ea924f5bfb3d08afa3b1ebe1f9afdfe86: 404 - Not Found // See:"\
13:            " https://developer.github.com/v3/repos/statuses/#create-a-status",
25:            "2888102937852b84f9fbce58def0f5132b4a5002: 404 - Not Found // See:"\
26:            " https://developer.github.com/v3/repos/statuses/#create-a-status",

gitlabhq/lib/gitlab/sidekiq_middleware/memory_killer.rb
31:          warn("Sidekiq worker PID-#{pid} current RSS #{current_rss}"\
32:               " exceeds maximum RSS #{MAX_RSS} after finishing job #{worker.class} JID-#{job['jid']}")
34:          warn("Sidekiq worker PID-#{pid} will stop fetching new jobs"\
35:               " in #{GRACE_TIME} seconds, and will be shut down #{SHUTDOWN_WAIT} seconds later")

gitlabhq/lib/gitlab/reference_counter.rb
54:          Gitlab::AppLogger.warn("Reference counter for #{gl_repository} decreased" \
55:            " when its value was less than 1. Resetting the counter.")

gitlabhq/lib/gitlab/import_export/project/import_task.rb
75:          logger.info "Importing GitLab export: #{file_path} into GitLab" \
76:            " #{full_path}" \
77:            " as #{current_user.name}"

gitlabhq/lib/gitlab/database/partitioning_migration_helpers/table_management_helpers.rb
285:            Gitlab::AppLogger.warn "Partitioned table not created because it already exists" \
286:              " (this may be due to an aborted migration or similar): table_name: #{partitioned_table_name} "
339:            Gitlab::AppLogger.warn "Partition not created because it already exists" \
340:              " (this may be due to an aborted migration or similar): partition_name: #{partition_name}"
357:            Gitlab::AppLogger.warn "Partitioning sync function not created because it already exists" \
358:              " (this may be due to an aborted migration or similar): function name: #{name}"
390:            Gitlab::AppLogger.warn "Partitioning sync trigger not created because it already exists" \
391:              " (this may be due to an aborted migration or similar): trigger name: #{trigger_name}"

gitlabhq/lib/gitlab/database/partitioning_migration_helpers/index_helpers.rb
33:            Gitlab::AppLogger.warn "Index not created because it already exists (this may be due to an aborted" \
34:              " migration or similar): table_name: #{table_name}, index_name: #{options[:name]}"

gitlabhq/lib/gitlab/database/load_balancing/sidekiq_server_middleware.rb
20:            raise JobReplicaNotUpToDate, "Sidekiq job #{worker_class} JID-#{job['jid']} couldn't use the replica."\
21:              "  Replica was not up to date."

gitlabhq/rubocop/cop/migration/update_column_in_batches.rb
13:        MSG = 'Migration running `update_column_in_batches` must have a spec file at' \
14:          ' `%s`.'

gitlabhq/rubocop/cop/migration/add_columns_to_wide_tables.rb
12:        MSG = '`%s` is a wide table with several columns, adding more should be avoided unless absolutely necessary.' \
13:              ' Consider storing the column in a different table or creating a new one.'

gitlabhq/qa/qa/specs/features/api/4_verify/api_variable_inheritance_with_forward_pipeline_variables_spec.rb
79:                            "Expected to find `{key: 'TEST_VAR', value: 'This is great!'}`" \
80:            " but got #{pipeline.pipeline_variables}"
86:                                "Did not expect to find `{key: 'TEST_VAR', value: 'This is great!'}`" \
87:            " but got #{pipeline.pipeline_variables}"

gitlabhq/spec/services/issues/build_service_spec.rb
70:        note_result = "    > This is a string\n"\
71:                      "    > \n"\
72:                      "    > > with a blockquote\n"\
73:                      "    > > > That has a quote\n"\
74:                      "    > \n"

gitlabhq/spec/services/feature_flags/update_service_spec.rb
111:          include("Updated description from \"\""\
112:                  " to \"new description\".")

gitlabhq/spec/lib/system_check/incoming_email/imap_authentication_check_spec.rb
6:  ":mailboxes:\n"\
7:  "  \n"\
8:  "    -\n"\
9:  "      :host: \"gitlab.example.com\"\n"\
10:  "      :port: 143\n"\

gitlabhq/spec/lib/gitlab/prometheus/queries/validate_query_spec.rb
47:        "URL 'http://192.168.1.1/api/v1/query?query=avg%28metric%29&time=#{Time.now.to_f}'" \
48:        " is blocked: Requests to the local network are not allowed"

gitlabhq/spec/lib/gitlab/reference_counter_spec.rb
24:      expect(Gitlab::AppLogger).to receive(:warn).with("Reference counter for project-1" \
25:        " decreased when its value was less than 1. Resetting the counter.")

gitlabhq/spec/lib/gitlab/incoming_email_spec.rb
113:      '<issue_1@localhost>' \
114:        ' <reply-59d8df8370b7e95c5a49fbf86aeb2c93@localhost>' \

gitlabhq/spec/lib/gitlab/usage/metrics/instrumentations/count_bulk_imports_entities_metric_spec.rb
34:        "SELECT COUNT(\"bulk_import_entities\".\"id\") FROM \"bulk_import_entities\""\
35:        " WHERE \"bulk_import_entities\".\"created_at\" BETWEEN '#{start}' AND '#{finish}'"
53:        "SELECT COUNT(\"bulk_import_entities\".\"id\") FROM \"bulk_import_entities\""\
54:        " WHERE \"bulk_import_entities\".\"source_type\" = 1"
67:        "SELECT COUNT(\"bulk_import_entities\".\"id\") FROM \"bulk_import_entities\""\
68:        " WHERE \"bulk_import_entities\".\"created_at\" BETWEEN '#{start}' AND '#{finish}'"\
69:        " AND \"bulk_import_entities\".\"source_type\" = 1"
82:        "SELECT COUNT(\"bulk_import_entities\".\"id\") FROM \"bulk_import_entities\""\
83:        " WHERE \"bulk_import_entities\".\"source_type\" = 0"
96:        "SELECT COUNT(\"bulk_import_entities\".\"id\") FROM \"bulk_import_entities\""\
97:        " WHERE \"bulk_import_entities\".\"created_at\" BETWEEN '#{start}' AND '#{finish}'"\
98:        " AND \"bulk_import_entities\".\"source_type\" = 0"

gitlabhq/spec/lib/gitlab/usage/metrics/instrumentations/count_imported_projects_metric_spec.rb
34:        "SELECT COUNT(\"projects\".\"id\") FROM \"projects\" WHERE \"projects\".\"created_at\""\
35:        " BETWEEN '#{start}' AND '#{finish}' AND \"projects\".\"import_type\" = 'gitea'"
61:        "SELECT COUNT(\"projects\".\"id\") FROM \"projects\" WHERE \"projects\".\"created_at\""\
62:        " BETWEEN '#{start}' AND '#{finish}' AND \"projects\".\"import_type\" = 'bitbucket'"

gitlabhq/spec/lib/gitlab/ci/ansi2html_spec.rb
216:      '<div class="section-start"' \
217:      " data-timestamp=\"#{section_start_time.to_i}\" data-section=\"#{class_name(section_name)}\"" \
218:      ' role="button"></div>'

gitlabhq/spec/support/shared_examples/features/sidebar_shared_examples.rb
112:          _('Only project members with at least' \
113:          ' Reporter role can view or be' \
114:          ' notified about this issue.')

gitlabhq/spec/support/matchers/have_gitlab_http_status.rb
17:    "expected the response to have status code #{expected.inspect}" \
18:    " but it was #{response_code}. The response was: #{actual.body}"

gitlabhq/spec/models/integrations/chat_message/pipeline_message_spec.rb
66:      "Pipeline [#123](http://example.gitlab.com/-/pipelines/123)" \
67:        " of branch [develop](http://example.gitlab.com/-/commits/develop)" \
68:        " by The Hacker (hacker) has passed"
79:        "Pipeline [#123](http://example.gitlab.com/-/pipelines/123)" \
80:          " of branch [develop](http://example.gitlab.com/-/commits/develop)" \
81:          " by The Hacker (hacker) has failed"
93:        "Pipeline [#123](http://example.gitlab.com/-/pipelines/123)" \
94:          " of branch [develop](http://example.gitlab.com/-/commits/develop)" \
95:          " by The Hacker (hacker) has passed with warnings"
107:        "Pipeline [#123](http://example.gitlab.com/-/pipelines/123)" \
108:          " of branch [develop](http://example.gitlab.com/-/commits/develop)" \
109:          " by API has passed"
138:      "<http://example.gitlab.com|project_name>:" \
139:        " Pipeline <http://example.gitlab.com/-/pipelines/123|#123>" \
140:        " of branch <http://example.gitlab.com/-/commits/develop|develop>" \
141:        " by The Hacker (hacker) has passed in 02:00:10"
373:        "[project_name](http://example.gitlab.com):" \
374:          " Pipeline [#123](http://example.gitlab.com/-/pipelines/123)" \
375:          " of branch [develop](http://example.gitlab.com/-/commits/develop)" \
376:          " by The Hacker (hacker) has passed in 02:00:10"

gitlabhq/spec/features/issues/create_issue_for_single_discussion_in_merge_request_spec.rb
82:      expect(page).to have_content("The thread at #{merge_request.to_reference}"\
83:                                   " (discussion #{discussion.first_note.id}) will stay unresolved."\
84:                                   " Ask someone with permission to resolve it.")

gitlabhq/config/initializers_before_autoloader/003_gc_compact.rb
15:      puts 'Note: GC compacting is currently disabled.'\
16:        ' Refer to `config/initializers_before_autoloader/003_gc_compact.rb` for details.'

gitlabhq/app/services/pod_logs/kubernetes_service.rb
45:        return error(_('pod_name cannot be larger than %{max_length}'\
46:          ' chars' % { max_length: K8S_NAME_MAX_LENGTH }))
72:        return error(_('container_name cannot be larger than'\
73:          ' %{max_length} chars' % { max_length: K8S_NAME_MAX_LENGTH }))

gitlabhq/app/helpers/application_settings_helper.rb
139:    s_("ExternalAuthorization|Access to projects is validated on an external service"\
140:        " using their classification label.")
170:    s_("ExternalAuthorization|Classification label to use when requesting authorization if no specific "\
171:      " label is defined on the project.")

gitlabhq/app/helpers/preferences_helper.rb
124:      raise "`User` defines #{user_dashboards.size} dashboard choices," \
125:        " but `localized_dashboard_choices` defined #{localized_dashboard_choices.size}."

gitlabhq/app/graphql/resolvers/work_items/types_resolver.rb
10:               description: 'If `true`, only taskable work item types will be returned.' \
11:                            ' Argument is experimental and can be removed in the future without notice.'

gitlabhq/app/graphql/types/query_type.rb
93:          description: 'Find a work item. Returns `null` if `work_items` feature flag is disabled.' \
94:                       ' The feature is experimental and is subject to change without notice.'

gitlabhq/app/graphql/types/group_type.rb
212:          description: 'Work item types available to the group.' \
213:                       ' Returns `null` if `work_items` feature flag is disabled.' \
214:                       ' This flag is disabled by default, because the feature is experimental and is subject to change without notice.'

gitlabhq/app/graphql/types/project_type.rb
415:          description: 'Work item types available to the project.' \
416:                       ' Returns `null` if `work_items` feature flag is disabled.' \
417:                       ' This flag is disabled by default, because the feature is experimental and is subject to change without notice.'

gitlabhq/app/graphql/mutations/work_items/update.rb
7:      description "Updates a work item by Global ID." \
8:                  " Available only when feature flag `work_items` is enabled. The feature is experimental and is subject to change without notice."

gitlabhq/app/graphql/mutations/work_items/delete.rb
7:      description "Deletes a work item." \
8:                  " Available only when feature flag `work_items` is enabled. The feature is experimental and is subject to change without notice."

gitlabhq/app/graphql/mutations/work_items/create_from_task.rb
10:      description "Creates a work item from a task in another work item's description." \
11:                  " Available only when feature flag `work_items` is enabled. This feature is experimental and is subject to change without notice."

gitlabhq/app/graphql/mutations/work_items/create.rb
11:      description "Creates a work item." \
12:                  " Available only when feature flag `work_items` is enabled. The feature is experimental and is subject to change without notice."

gitlabhq/app/graphql/mutations/work_items/update_task.rb
7:      description "Updates a work item's task by Global ID." \
8:                  " Available only when feature flag `work_items` is enabled. The feature is experimental and is" \
9:                  " subject to change without notice."

gitlabhq/app/graphql/mutations/work_items/delete_task.rb
8:      description "Deletes a task in a work item's description." \
9:                  ' Available only when feature flag `work_items` is enabled. This feature is experimental and' \
10:                  ' is subject to change without notice.'

rspec-expectations/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
73:            "#{diff_label}\n" \
74:              "  <The diff is empty, are your objects producing identical `#inspect` output?>"

rspec-expectations/lib/rspec/matchers/built_in/exist.rb
63:              " but `exist?` and `exists?` returned different values:\n\n"\
64:              " exist?: #{existence_values.first}\n"\

rspec-expectations/lib/rspec/matchers/built_in/include.rb
97:                  "expected #{description_of @actual} #{preposition}" \
98:                  " include#{readable_list_of @divergent_items}" \

rspec-expectations/lib/rspec/matchers/built_in/count_expectation.rb
127:            raise ArgumentError, "Expected a number, :once, :twice or :thrice," \
128:              " but got #{n}"
144:          "#{count_expectation_description}" \
145:          " but #{action}#{human_readable_count(@actual_count)}"

rspec-expectations/lib/rspec/matchers/built_in/be.rb
160:          "expected: #{@operator} #{expected_formatted}\n" \
161:          "     got: #{@operator.to_s.gsub(/./, ' ')} #{actual_formatted}"
```
</details>

Total 1019 lines (not occurrences!).
This was painfully slow, felt even slower than running a full corpus of cops on the same set of files 😄

Out of those, I'd say the ones with `\n` should be written as HEREDOC, not line continuation.
Some to give some visual indentation, e.g. for SQL code.

Personally, more leaning towards trailing spaces, I can't find many justifiable examples for the leading space style.

pirj:
```
rg --no-ignore --hidden --multiline "[ ]+['\"][ ]*\\\\[ ]*\\n"
```

Trailing spaces on continuation are in a majority with 4200 total lines (not occurrences!). Too long to list for a GitHub comment.

bquorning:
Thank you @pirj 🙏🏼 

I agree, where the leading whitespace serves as indentation, a heredoc would often be a better choice.

bbatsov:
> I agree, where the leading whitespace serves as indentation, a heredoc would often be a better choice.

Sounds like another cop idea to me. ;-) 

Let's keep the proposed cop simple and just agree on the final name & department. As mentioned before it seems to me that's mostly a layout cop, even if it's about the spacing inside multi-line string literals. I'm okay with the current name.

bquorning:
> I'm okay with the current name.

I like @pirj’s suggestion better :-) `Layout/LineContinuationLeadingSpace` then?

bbatsov:
Fine by me. Thanks for working on this!

PR #10717, opened 2022-06-16
Add Layout/LineContinuationSpacing cop

bquorning:
Fixes the 1st issue mentioned in #6420.

`Style/LineEndConcatenation` recommends that we should use `\` instead of `+` or `<<` to concatenate strings across multiple lines. But I am still seeing wildly inconsistent formatting of these multiline strings. This new `Layout/LineContinuationSpacing` cop checks that there is consistently _one_ space before a `\` at the end of a line. (Can be configured to _zero_ spaces)

```rb
# Bad
'this text is too '\
'long'

# Bad
'this text is too '  \
'long'

# Good
'this text is too ' \
'long'
```

-----------------

* [x] Added tests.


lib/rubocop/cop/style/line_continuation_spacing.rb
```diff
@@ -0,0 +1,122 @@
...
+        def autocorrect(corrector, range)
+          correction = if zero_space_style?
+                         '\\'
+                       elsif one_space_style?
+                         ' \\'
+                       end
+          corrector.replace(range, correction)
+        end
+
+        def string_literal_ranges(ast)
```
bquorning, line 85: ~Inspired~ Copied from https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/layout/indentation_style.rb

bbatsov:
I love this cop as the inconsistent spacing has been bugging me for a while. One small question, though - won't it be better to just make this some number config? 

If we stick to the styles I'd suggest naming them "single_space" and "no_space" as "zero" and "one" are kind of weird for style names. :-)

bquorning:
I also thought about the number config, but then thought that I don’t want to write specs for how it works if someone chooses e.g. 20 spaces 😆 I think really only 0 and 1 are valid values.

I personally would hate to use 0, but I have seen it used in the wild and added it up front instead of waiting for the inevitable feature request.

We have a lot of cops with `SupportedStyles: [space, no_space]`, so I wonder if I should adopt that naming instead of the suggested `single_space`/`no_space`?

bquorning:
I would actually like feedback on two specific things:

1. Please help me find a better cop name than LineContinuationSpacing.
2. Does this cop belong in the Style, Lint, or Layout department?

bbatsov:
> Please help me find a better cop name than LineContinuationSpacing.

I think the name if fine. I thought we might have just added this to the other cop, but I think in this case it's better to have 2 of them.

> Does this cop belong in the Style, Lint, or Layout department?

I think it should be a Layout cop.

bbatsov (approved):

bbatsov:
Looks good to me!

bquorning:
Hi @ngouy, I think you message was meant for #10715 where Layout/LineContinuationLeadingSpace was added. The two cops are quite similar, yet different :-)

We did talk about configurability in the PR, so please have a look.

ngouy:
Fair, I understand. I get that it would've significantly increase the complexity of the implementation
I can just disable it until further notice,  that's fine too. That's also how rubocop works ;)

Thanks again @bquorning (and sorry for wrong thread 😂 , I tried but somehow managed to miss the right one)

bquorning:
@ngouy See #10762 😅

ngouy:
lol

PR #10722, opened 2022-06-17
Add new `Lint/ConstantOverwrittenInRescue ` cop

ydah:
This PR is add a new cop.
Checks for overwriting an exception with an exception result by use `rescue =>`.

## @example

```ruby
# bad
begin
  something
rescue => StandardError
end

# good
begin
  something
rescue StandardError
end
```

-----------------

* [x] Added tests.


pirj:
Wow!
Just to show the scale of destruction:
```ruby
> E=Class.new(StandardError)
> begin; raise E; rescue => E; end

(pry):2: warning: already initialized constant E
(pry):1: warning: previous definition of E was here

> E
#<E: E>
```
😱 

Frankly, it looks like a Ruby bug or a deficiency.

In any case, thanks for the cop! 👏

lib/rubocop/cop/lint/overwrite_by_rescue.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Checks for overwriting an exception with an exception result by use `rescue =>`.
```
bbatsov, line 6: I'd expand on the documentation here and add some explanation that people wrote a `rescue` modifier form instead of rescue that actually catches exception, which they probably meant to write.

bbatsov:
I like the cop, we just need to expand a bit the docs and the examples.

lib/rubocop/cop/lint/overwrite_by_rescue.rb
```diff
@@ -0,0 +1,47 @@
...
+      #   rescue => StandardError
+      #   end
+      #
+      #   # good
+      #   begin
+      #     something
+      #   rescue StandardError
+      #   end
+      #
+      class OverwriteByRescue < Base
```
bbatsov, line 22: You've used `Overwrite` in the class name and `Override` in the message. :)

lib/rubocop/cop/lint/overwrite_by_rescue.rb
```diff
@@ -0,0 +1,47 @@
...
+        extend AutoCorrector
+        include RangeHelp
+
+        MSG = 'Override exception by `rescue =>`.'
```
bbatsov, line 26: I guess the message can be more specific about what we actually overwrote here.

pirj:
I'd go with `ConstantOverwrittenInRescue` for the cop name.

lib/rubocop/cop/lint/overwrite_by_rescue.rb
ydah: Documentation has been expanded to include additional explanations.

lib/rubocop/cop/lint/overwrite_by_rescue.rb
ydah: Oh. I had missed it💦 Thank you!

lib/rubocop/cop/lint/overwrite_by_rescue.rb
ydah: A message is now show to indicate what is being overwritten by `rescue =>`.

ydah:
@pirj `ConstantOverwrittenInRescue` I think it is a good Cop name. I changed the name :) Thank you so much.

bbatsov:
Looks good. Thanks!

Issue #10089, opened 2021-09-15
Inconsistent cop names `Lint/DisjunctiveAssignmentInConstructor` and `Lint/OrAssignmentToConstant`


dvandersluis:
Both of these cops relate to the `||=` operator, but have different names ("disjunctive assignment" vs "or assignment"). They should be made consistent.

dvandersluis:
@bbatsov I'm not sure if this is something that can be done before 2.0?

bbatsov:
Yep, this will have to wait unless we figure out a way to be able to have a primary cop name and some alias for the sake of not breaking user configuration when renaming stuff. I guess we should start some list of all the cops we'd like to rename for one reason or another.

bbatsov:
We should also agree which name to adopt as the canonical one. I assume that disjunctive assignment is the "proper" term, but I'm not sure about this.

dvandersluis:
Agreed. FWIW I've never seen "disjunctive assignment" elsewhere, but it sounds fancier. It might be confusing to end-users though (and it was raised when the cop was first created - https://github.com/rubocop/rubocop/pull/6608#issuecomment-450776354)

bbatsov:
Fair point - logical or is known as "disjunction", but I've heard this term used only in academic contexts. Perhaps we should stick to simpler terms.

dvandersluis:
I think it's fine to stick with disjunctive, it's just more academic, but that's not a bad thing :grin:

PR #10725, opened 2022-06-19
[Fix #7077] Rename cops to meet the cop naming guideline

### LINKED ISSUE #7077: Cop naming guidelines
With RuboCop 1.0 closer than ever it's time for us to start doing to final cleanup leading to the 1.0 release. One aspect of the cleanup is to make sure all cops follow consistent naming patterns (and we'll enforce those down the road). Here are a few examples of inconsistencies we have today:

* We use "Unneeded" and "Redundant" (we should use always "Redundant")
* Blank vs Empty (for lines)
* "Indent something" vs "something indentation"
* Mixing params and arguments here and there

I hope you get the idea. The goal of this ticket is to agree on how names should use in general, document our decisions, and update all cops to match those before 1.0. 

We should do the same for cop configuration keys as well.

pirj:
fixes #7077, with an exception, see below:

### Not Done

> We also need to put something in the docs on the subject and maybe introduce an internal affairs cop that checks for some naming anti-patterns.

#10089, will fix in a separate PR

-----------------

* [-] Added tests.


bbatsov:
Might also be a good idea to spell out the naming guidelines explicitly in the docs and add some internal affairs cop that looks for common problems (mostly problematic words appearing in the name or in the wrong order). I'm also wondering if we can come up with a good replacement for things like `PercentArray`, as I hate this terminology - after all those are just different array/string/etc literals.

Perhaps we should be spelling out `Def` fully as `Definition`. We have to see what has been the predominant practice so far.

pirj:
> spell out the naming guidelines explicitly

I've added [a small note to](https://github.com/rubocop/rubocop/pull/10725/files#diff-d215eaa340849c3e72defb154136bcb8e2238d47e34236eaabd7e09b9b334e1bR13) to the [Development](https://docs.rubocop.org/rubocop/1.30/development.html) section of our docs, as it seems to be the only one on the topic of creating new cops:
>  Name the cop to explain the offense it detects. See the https://github.com/rubocop/rubocop/blob/master/config/obsoletion.yml:["renamed" section of `config/obsoletion.yml] for good and bad examples (old name is on the left, new name on the right).

After a while, my old idea of writing it as:
> Name the cop to explain the offense it detects, not where or what the subject of the offense is.

does not seem very specific and guiding. There are some good practices behind good cop naming, but I miserably fail to extract them into concise statements.

pirj:
Done https://github.com/rubocop/rubocop/blob/dced8a2af865a9a04a90294c2a614e8c11b52b12/docs/modules/ROOT/pages/development.adoc#choose-a-name

docs/modules/ROOT/pages/development.adoc
```diff
@@ -29,9 +29,33 @@ Do 4 steps:
...
+
+Name the cop to explain the offense it detects, e.g. `ExtraSpacing`.
+
+Use a leading noun instead of a verb, e.g. use `ArrayAlignment` instead of `AlignArray`.
+
+Make it easy to understand, e.g. use `IndentationStyle` instead of just `Tab`.
+
+Be specific, e.g. use `DuplicateHashKey` instead of just `DuplicateKey`.
+
+Stick to commonly used terms, e.g. use `RedundantPercentI` instead of `RedundantPercentSymbolArray'.
```
dvandersluis, line 48:
```suggestion
Stick to commonly used terms, e.g. use `RedundantPercentI` instead of `RedundantPercentSymbolArray`.
```

docs/modules/ROOT/pages/development.adoc
```diff
@@ -29,9 +29,33 @@ Do 4 steps:
...
+
+Use correct terms, e.g. arguments in a method call, and parameters in a method signature.
+
+Refer to lines with no symbols as "empty", not "blank", e.g. `LeadingEmptyLines` instead of `LeadingBlankLines`.
+
+Prefer "redundant" to "unneeded", e.g. `RedundantSelf` instead of `UnneededSelf`.
+
+See the https://github.com/rubocop/rubocop/blob/master/config/obsoletion.yml:["renamed" section of `config/obsoletion.yml`]
+for good and bad examples (old name is on the left, new name on the right).
```
dvandersluis, line 57: The link on line 56 seems to be broken because it's including the colon (at least in the github link you added in your comment).

dvandersluis: Actually looks like the [colon shouldn't be there](https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#links)

```suggestion
See the https://github.com/rubocop/rubocop/blob/master/config/obsoletion.yml["renamed" section of `config/obsoletion.yml`]
for good and bad examples (old name is on the left, new name on the right).
```

pirj: My bad! Thanks for the pointer.

pirj:
Docs (with reference formatting fixes) for easier view https://github.com/rubocop/rubocop/blob/28d52cb0d2c730c7f854961170c8480c85529973/docs/modules/ROOT/pages/development.adoc#choose-a-name

bbatsov:
Just to confirm - right now that's not a breaking change, right? I recall @dvandersluis worked on this a while back.

docs/modules/ROOT/pages/development.adoc
```diff
@@ -29,9 +29,33 @@ Do 4 steps:
      e.g. "Add new `Department/Name` cop"
   4. Run `bundle exec rake changelog:new` to generate a changelog entry
      for your new cop.
-
 ----
 
+=== Choose a Name
+
+Pick a department. See the xref:cops.adoc[list of existing departments].
```
bbatsov, line 36: Why not make these bullet points?

docs/modules/ROOT/pages/development.adoc
```diff
@@ -29,9 +29,33 @@ Do 4 steps:
...
+
+Be specific, e.g. use `DuplicateHashKey` instead of just `DuplicateKey`.
+
+Stick to commonly used terms, e.g. use `RedundantPercentI` instead of `RedundantPercentSymbolArray`.
+
+Use correct terms, e.g. arguments in a method call, and parameters in a method signature.
+
+Refer to lines with no symbols as "empty", not "blank", e.g. `LeadingEmptyLines` instead of `LeadingBlankLines`.
+
+Prefer "redundant" to "unneeded", e.g. `RedundantSelf` instead of `UnneededSelf`.
```
bbatsov, line 54: I'd also add to the list to pick names that are neutral when possible and would accommodate multiple styles when feasible.

pirj:
> right now that's not a breaking change, right? I recall @dvandersluis worked on this a while back.

It is unfortunately:
```
Error: The `Layout/IndentationConsistency` cop has been renamed to `Layout/InconsistentIndentation`.
(obsolete configuration found in .rubocop.yml, please update it)
```

dvandersluis:
I added a way to set `severity: warning` for some types of obsoletions (which just outputs a warning instead of an error) but it is not set up for renames presently. I could add it if desired?

dvandersluis:
Actually I remember now why we didn't. I tried implementing warnings for renames, but we still get an error because there's some other checks, for instance `ConfigValidator` raises its own error if there is configuration for a cop that does not exist, so we end up with something like this:

image

and then RuboCop quits anyways. So there doesn't look like there's an easy way to make a warning happen here.

What we've done in some cases for renamed params is have the config handle both the old (deprecated) case and the new renamed case, and then update the cop to use both. We can't really easily do that with full cop renames though...

dvandersluis:
I have an idea about how we might be able to solve this problem, I'm going to play with it tomorrow.

In the meantime, as @pirj said, this would be a breaking change.

docs/modules/ROOT/pages/development.adoc
```diff
@@ -29,9 +29,27 @@ Do 4 steps:
      e.g. "Add new `Department/Name` cop"
   4. Run `bundle exec rake changelog:new` to generate a changelog entry
      for your new cop.
-
 ----
 
+=== Choose a Name
```
pirj, line 34: @bbatsov Does this guide look complete?

pirj:
@dvandersluis Any breakthrough on the idea to painlessly rename cops between major version updates?

dvandersluis:
@pirj I’ve been away from home but I’m hoping to get back to it in next week.

pirj:
@dvandersluis We came up with an idea to obsolete cops without actually removing them, see https://github.com/rubocop/rubocop-rspec/pull/1519. The idea is to:
 - rename the cop
 - [inherit the old cop from the new one](https://github.com/rubocop/rubocop-rspec/pull/1519/files#diff-1c2c7147a4824eb9e7b423ad762e5759b5d273481e7feb4829015649842064b4R32-R34)
 - keep the configuration in `config/default.yml` for both
 - introduce an entry in `config/obsoletion.yml`

That way, there's an obsoletion warning, but no error trying to load the config.

I can rework this PR in the same way. Would you consider this a non-breaking, mergeable change?

In 2.0 we'll have to remove cops with the old names along with their configs and that's it.

dvandersluis:
Hey @pirj, sorry that I obviously never had the time to work on this like I wanted to. I think your solution makes sense!

bbatsov:
I agree that the new proposal is a non-breaking one (at least in my book) and provides a nice path to complete removal down the road.

koic:
This looks like a good way. One thing, can we use assignment (`NewNameCop = ObsoleteNameCop`) instead of inheritance (`class NewNameCop < ObsoleteNameCop`) ?

pirj:
> can we use assignment?

That was [the initial approach](https://github.com/rubocop/rubocop-rspec/compare/c820abd1853c0c98b66644439c98eebc717ba76e..ab3b50f35c60ac7a167d48608b35df6ba99d5129), I can't recall why I've switched to inheriting, I'll check if it works.

pirj:
> can we use assignment (NewNameCop = ObsoleteNameCop) instead of inheritance (class NewNameCop < ObsoleteNameCop) ?

Probably, but the doc generation task breaks.
Our goal is to keep the documentation for both until the major version update, right?
We have to tell YARD to parse a const definition as a class definition:
```ruby
# Some cop docs
# !@parse class class NewNameCop < ::RuboCop::Cop::Base; end
NewNameCop = ObsoleteNameCop
```
But for some reason, documentation above the `!@parse` directive is not considered as the class doc. I'll investigate why.

It appears that this is the only way in any case, as obsoletion errors sometimes appear when inheritance is used.

pirj:
Luckily, there's a workaround:
```ruby
# !@parse
#   # Some cop docs
#   class NewNameCop < ::RuboCop::Cop::Base; end
NewNameCop = ObsoleteNameCop
```
See https://github.com/rubocop/rubocop-rspec/pull/1519 for more details.

koic:
As I left a comment on https://github.com/rubocop/rubocop-rspec/pull/1519#issuecomment-1379893876, the current implementation should have used inheritance (`class NewNameCop < ObsoleteNameCop`). I forgot to consider `Base.inherited` 💦  Please see link for details.

pirj:
My apologies, @koic, I'm not convinced enough that duplicate output from the old and the new cop that is an impact on user experience is a good tradeoff for not modifying the code that builds cops' documentation.

I may miss something important, but the duplicate output was my experience with the inheritance approach.

PR #10740, opened 2022-06-22
Make `Style/GuardClause` a bit more lenient when the replacement would make the code more verbose

dvandersluis:
We currently have an exception in `Style/GuardClause` when the suggested replacement would be too long.

```ruby
def test
  if something && something_that_makes_the_guard_clause_too_long_to_fit_on_one_line
    work
  end
end
```

RuboCop suggests replacing this with:

```ruby
def test
  unless something && something_that_makes_the_guard_clause_too_long_to_fit_on_one_line
    return
  end
  work
end
```

However, this makes the code more verbose (adds a line that doesn't really add any benefit) and harder to understand (since this only happens when the line is too long, it's a safe assumption that the condition is long, and flipping it with `unless` adds cognitive overhead).

This exception is not covered in the [style guide](https://rubystyle.guide/#no-nested-conditionals) either, so it should be "non-breaking".

My proposal here is that if the guard clause is too line but what is inside the `if` is trivial (a single statement, and not a nested if), it will not be registered as an offense.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/guard_clause.rb
+++ b/lib/rubocop/cop/style/guard_clause.rb
@@ -110,7 +110,7 @@ module RuboCop
           kw = if guard_clause_in_if
                  node.loc.keyword.source
                else
-                 opposite_keyword(node)
+                 node.inverse_keyword
                end
 
           register_offense(node, guard_clause_source(guard_clause), kw)
@@ -123,7 +123,7 @@ module RuboCop
           return if allowed_consecutive_conditionals? &&
                     consecutive_conditionals?(node.parent, node)
 
-          register_offense(node, 'return', opposite_keyword(node))
+          register_offense(node, 'return', node.inverse_keyword)
         end
 
         def consecutive_conditionals?(parent, node)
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/style/guard_clause.rb
```diff
@@ -134,14 +134,12 @@ def consecutive_conditionals?(parent, node)
           end
         end
 
-        def opposite_keyword(node)
-          node.if? ? 'unless' : 'if'
-        end
```
dvandersluis, line 139: This is not strictly related to my change but I noticed it could be replaced with `node.inverse_keyword`.

https://github.com/rubocop/rubocop-ast/blob/816dfe7f2ca4e92c7eda226a9e8b44aa9fa81e81/lib/rubocop/ast/node/if_node.rb#L61-L73

bbatsov:
That's an improvement for sure, but overall I think that's one of those cops that need more work and have to be relaxed a bit to focus on cases where the guard clause really improves something.

PR #10743, opened 2022-06-22
Add `RESTRICT_ON_SEND` to the Cop template generated by the `new_cop` task

ydah:
The generated Cop template uses `on_send`.
In many cases, `RESTRICT_ON_SEND` is used when `on_send` is to be used as is.

Also Its also possible to make Cop without using `RESTRICT_ON_SEND`, but it slows down the processing speed, so we want to add it here.

Also added a cop to check for useless `RESTRICT_ON_SEND` since `RESTRICT_ON_SEND` is defined by default.

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -302,6 +302,11 @@ Gemspec/RubyVersionGlobalsUsage:
 
 #################### Layout ###########################
 
+InternalAffairs/UselessRestrictOnSend:
+  Description: 'TODO: Write a description of the cop.'
+  Enabled: pending
+  VersionAdded: '<<next>>'
+
```
koic, line 309: Can you remove it?

```suggestion
```

lib/rubocop.rb
```diff
@@ -715,3 +715,4 @@
   RuboCop::ResultCache.rubocop_required_features = $LOADED_FEATURES - before_us
 end
 RuboCop::AST.rubocop_loaded if RuboCop::AST.respond_to?(:rubocop_loaded)
+require_relative 'rubocop/cop/internal_affairs/useless_restrict_on_send'
```
koic, line 718: Can you move it to the `lib/rubocop/cop/internal_affairs.rb`?
cf. https://github.com/rubocop/rubocop/pull/10412

changelog/new_add_restrict_on_send_for_cop_template.md
```diff
@@ -0,0 +1 @@
+* [#10743](https://github.com/rubocop/rubocop/pull/10743): Add `RESTRICT_ON_SEND` to the Cop template generated by the `new_cop` task. ([@ydah][])
```
koic, line 1: No change log is required because the change is an internal improvements and have no user impact. Can you remove this entry?

ydah: I thought I saw a PR that had added `InternalAffairs` cop to the changelog, so I had added it, but it is certainly unnecessary. I deleted it.

lib/rubocop.rb
ydah: I'd pushed commit in the middle of my work. Sorry. It has been moved.

config/default.yml
ydah: I have removed it.

lib/rubocop/cop/generator.rb
```diff
@@ -62,6 +62,10 @@ class %<cop_name>s < Base
                 # For example
                 MSG = 'Use `#good_method` instead of `#bad_method`.'
 
+                # optimization: don't call `on_send` unless the method name is in this list
```
koic, line 65: I think this is `TODO`.

```suggestion
                # TODO: Don't call `on_send` unless the method name is in this list
```

koic:
Thanks!

PR #10749, opened 2022-06-26
Fix Style/BlockDelimiters for blocks with numbered arguments

gsamokovarov:
With `EnforcedStyle` of `line_count_based`, the default, the following code **will** issue an offense, and will be auto-corrected:

```ruby
[1, 2, 3].map { |n| # [1, 2, 3].map do |n|
  n + 1             #   n + 1
}                   # end

```

While the code below won't be converted to a multiline `do ... end` block:

```ruby
[1, 2, 3].map {
  _1 + 1
}
```

This is yet another case of missing `on_numblock` implementation. I see a dozen of those in the codebase and I think whenever we handle `block` AST nodes we should also handle `numblock` nodes. I can go over the codebase and fix those problems on a per-cop basis, or I can try to fix it generally.

Currently, my best idea is to write an `InternalAffairs` cop that warns when we define `on_block` handlers, but we miss a `on_block` implementation. For the majority of the cases, an `on_numblock` alias of the `on_block` method would be enough to handle blocks with numbered arguments. I can then use the internal co and systematically go over the cases and add aliases or custom `numblock` handlers where required. I think this is better than solutions involving meta-programming or default `on_numblock` implementations delegating to `on_block`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_block_delimiters_with_numbered_arguments.md
+++ b/changelog/fix_block_delimiters_with_numbered_arguments.md
@@ -1,0 +1,1 @@
+* [#10749](https://github.com/rubocop/rubocop/pull/10749): Fix Style/BlockDelimiters for blocks with numbered arguments. ([@gsamokovarov][])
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

PR #10774, opened 2022-06-29
Fix false negatives when public method is defined after a private one

Darhazer:
[Similar to Rails/Delegate](https://github.com/rubocop/rubocop-rails/pull/736) this cop was just checking the previous lines for a `private` declaration. This fails when there is a private method in a nested class, or as the added spec illustrates - when private methods are declared before the public ones

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/def_node.rb
+++ b/lib/rubocop/cop/mixin/def_node.rb
@@ -5,8 +5,7 @@ module RuboCop
     # Common functionality for checking def nodes.
     module DefNode
       extend NodePattern::Macros
-
-      NON_PUBLIC_MODIFIERS = %w[private protected].freeze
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/mixin/def_node.rb
```diff
@@ -5,8 +5,7 @@ module Cop
     # Common functionality for checking def nodes.
     module DefNode
       extend NodePattern::Macros
-
-      NON_PUBLIC_MODIFIERS = %w[private protected].freeze
+      include ::RuboCop::Cop::VisibilityHelp
```
koic, line 8: Can you update it?

```suggestion
      include VisibilityHelp
```

koic:
Thanks!

PR #10776, opened 2022-06-29
Allow asking to avoid mixing hash shorthand styles in single hashes

h-lame:
For those who want to use hash shorthand style, but don't want to mix shorthand and explicit styles in the same hash we provide a new option for `EnforcedShorthandSyntax` called `consistent` that raises an offence if we've mixed styles.

E.g.

    # bad
    {foo: , bar: bar}

    # good
    {foo:, bar:}

    # bad
    {foo: , bar: baz}

    # good
    {foo: foo, bar: baz}

# Requesting help 🙋 

I don't particularly like the implementation.  Prior to this commit `HashSyntax` mixed in `HashShorthandSyntax` but was otherwise unaware of it.  With this change I've had to make `HashSyntax` aware of `HashShorthandSyntax` so it can call `on_hash_for_mixed_shorthand` in `on_hash`.  I don't think there's a way to enforce a no mixing style at the pair level meaning we can't do this in `on_pair` already present in `HashShorthandSyntax` and have to use `on_hash` to look at the whole hash in one go.  Would love some advice on this.

-----------------

* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists). (n/a)
* [x] Feature branch is up-to-date with `master` (if not - rebase it). (current rebase against - 0984886d4aacb9b1a6de5c35d5984fac7500d9d0)
* [x] Squashed related commits together. (n/a - single commit)
* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
+++ b/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
@@ -5,7 +5,7 @@ module RuboCop
     # This module checks for Ruby 3.1's hash value omission syntax.
     module HashShorthandSyntax
       OMIT_HASH_VALUE_MSG = 'Omit the hash value.'
-      EXPLICIT_HASH_VALUE_MSG = 'Explicit the hash value.'
+      EXPLICIT_HASH_VALUE_MSG = 'Include the hash value.'
 
       def on_pair(node)
         return if ignore_hash_shorthand_syntax?(node)
--- a/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
+++ b/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
@@ -6,6 +6,21 @@ module RuboCop
     module HashShorthandSyntax
       OMIT_HASH_VALUE_MSG = 'Omit the hash value.'
       EXPLICIT_HASH_VALUE_MSG = 'Include the hash value.'
+      DO_NOT_MIX_MSG_PREFIX = 'Do not mix explicit and implicit hash values.'
+      DO_NOT_MIX_OMIT_VALUE_MSG = "#{DO_NOT_MIX_MSG_PREFIX} #{OMIT_HASH_VALUE_MSG}"
+      DO_NOT_MIX_EXPLICIT_VALUE_MSG = "#{DO_NOT_MIX_MSG_PREFIX} #{EXPLICIT_HASH_VALUE_MSG}"
+
+      def on_hash_for_mixed_shorthand(hash_node)
+        return if ignore_mixed_hash_shorthand_syntax?(hash_node)
+
+        hash_value_type_breakdown = breakdown_value_types_of_hash(hash_node)
+
+        if hash_with_mixed_shorthand_syntax?(hash_value_type_breakdown)
+          mixed_shorthand_syntax_check(hash_value_type_breakdown)
+        else
+          no_mixed_shorthand_syntax_check(hash_value_type_breakdown)
+        end
+      end
 
       def on_pair(node)
         return if ignore_hash_shorthand_syntax?(node)
@@ -36,8 +51,14 @@ module RuboCop
         end
       end
 
+      def ignore_mixed_hash_shorthand_syntax?(hash_node)
+        target_ruby_version <= 3.0 || enforced_shorthand_syntax != 'consistent' ||
+          !hash_node.hash_type?
+      end
+
       def ignore_hash_shorthand_syntax?(pair_node)
         target_ruby_version <= 3.0 || enforced_shorthand_syntax == 'either' ||
+          enforced_shorthand_syntax == 'consistent' ||
           !pair_node.parent.hash_type?
       end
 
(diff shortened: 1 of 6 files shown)


lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
```diff
@@ -6,6 +6,21 @@ module Cop
...
+      DO_NOT_MIX_MSG_PREFIX = 'Do not mix explicit and implicit hash values.'
+      DO_NOT_MIX_OMIT_VALUE_MSG = "#{DO_NOT_MIX_MSG_PREFIX}  #{OMIT_HASH_VALUE_MSG}"
+      DO_NOT_MIX_EXPLICIT_VALUE_MSG = "#{DO_NOT_MIX_MSG_PREFIX}  #{EXPLICIT_HASH_VALUE_MSG}"
+
+      def on_hash_for_mixed_shorthand(hash_node)
+        return if ignore_mixed_hash_shorthand_syntax?(hash_node)
+
+        hash_value_type_breakdown = breakdown_value_types_of_hash(hash_node)
+
+        if hash_with_mixed_shorthand_syntax? hash_value_type_breakdown
```
jonas054, line 18: There should be parentheses here for the method call.

jonas054 (changes requested):
I don't have any problems with the dependency from HashSyntax to the mixin. It all looks good to me. But I'm missing the new supported style in `default.yml`.

lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
h-lame: Oops.  Thanks for pointing it out!

h-lame:
> I don't have any problems with the dependency from HashSyntax to the mixin. 

Thanks for looking - I'll leave it as-is then!

> But I'm missing the new supported style in `default.yml`.

My bad, I didn't realise I needed to update that too - all done now though!  Rebased again too.

jonas054 (approved):

bbatsov:
I'm OK with the proposal, but I don't like the name of the config option. I'm think of something like `(keep)_consistent`, but I'm open to other suggestions. Naming is hard. :-)

dvandersluis:
We use `consistent` as a style name elsewhere so it’d be consistent 😂

h-lame:
I admit I'm not sold on `prefer_short` but I thought the `no_mixing` gave some consistency with the `no_mixed_keys` and `ruby19_no_mixed_keys` values for `SupportedStyles`.  Admittedly, `no_mixed_keys` is clear _what_ it's not mixing, I'm not sure what we call this shorthand syntax - it's not keys, and values doesn't feel right to me.  Some alternates:

1. `prefer_short_no_mixing` - what I came up with originally
3. `always_no_mixing` - clear that it's like the value `always` in intent, but has a suffix to suggest a difference (things won't be mixed)
4. `always_keep_consistent` - again, clear that it's like `always` and uses a suffix to suggest a difference (however, it's less clear that there's an implied punctuation between `always` and `keep_consistent` (same problem if we drop `keep` here - `always_consistent` doesn't work either)
5. `always_no_mixed_omissions` - 
6. `for_entire_hashes_only` - going away from all the other options, this feels like it's clearer

I'll keep pondering it 🤔

bbatsov:
Just keep in mind that's not really a supported style, but rather the value of `EnforcedShorthandSyntax`, so the expectation here therefore my suggestion for `consistent`, as everything but "never" implies some usage of the shorthand syntax anyways. I don't like `always_something` as it's not really `always` then.

h-lame:
I don't really have a better suggestion than `consistent`.  What I think it lacks is a hint that it's going to prefer the shorthand syntax and re-write your hashes to use it if it can.  I don't think `consistent` is clear enough.  But, as I said, I don't have a better suggestion, and, as you say, it's kinda implied by the name of the option anyway: `EnforcedShorthandSyntax`.  The docs can explain it too.  I'll rebase and update to use `consistent` as the option value.

spec/rubocop/cop/style/hash_syntax_spec.rb
```diff
@@ -1268,4 +1268,76 @@ def do_something
...
+
+        expect_correction(<<~RUBY)
+          {foo:, bar:}
+        RUBY
+      end
+
+      it 'registers an offense when some hash values are omitted but they cannot all be omitted' do
+        expect_offense(<<~RUBY)
+          {foo:, bar: baz}
+           ^^^ Do not mix explicit and implicit hash values.  Explicit the hash value.
```
dvandersluis, line 1302: `Explicit the hash value` doesn't really make sense.
```suggestion
           ^^^ Do not mix explicit and implicit hash values.  Include the hash value.
```

h-lame: I agree that it doesn't make much sense, but I was reusing the existing error message for the `never` option.  Should I change that one too - are changes to existing error messages a breaking change?  

We can agree to disagree on the period double space formatting though 😉.  I will remove it though as it does seem more consistent with the rest of the error messages.

dvandersluis: We could change it as a separate PR if you'd prefer, but if you're willing, change it everywhere please! Changing messages aren't considered breaking.

h-lame: Added a commit to reword the existing message before adding the new feature.

dvandersluis:
Thanks!

koic:
@dvandersluis From next time onwards, please wait a few days after a minor version release for new features to merge. If there is an urgent bug, it will be a bugfix release. For example, I will wait for @bbatsov to merge new features before merging PR like this. For future reference. Anyway thank you for your activities :-)

h-lame:
@schmijos - I haven't explicitly tested that example, but it should be consistent and not flag for change because the implementation only looks at hashes, and in the "hash" part of that method call (`customer: , store:`) they're already consistent.  If you wanted it to do something else then I think you'd need to add a different implementation around argument lists.

PR #10783, opened 2022-07-03
Add new `InternalAffairs/RestrictOnSend` cop

ydah:
This cop will make sure that `RESTRICT_ON_SEND` is defined if `on_send` or `after_send` is defined.

This comment https://github.com/rubocop/rubocop-rspec/pull/1265#discussion_r861758342
is also mentioned in and leads to optimize the cop by `RESTRICT_ON_SEND` whenever possible.

However, this PR does not resolve existing cop.

If we included that much in this PR, the scale would be enormous,
and we thought it would be better to look at how to optimize each cop in detail.

-----------------

* [x] Added tests.


koic:
`RESTRICT_ON_SEND` cannot always be defined for `on_send`.
For example, if the method name to be detected is not decided, it can never be solved. 
Unfortunately, I think there is more noise than detection efficiency, so I will close this PR.

Issue #10455, opened 2022-03-16
`Style/{Symbol,Word}Array` should preserve multiline semantics when autocorrecting


sambostock:
## Is your feature request related to a problem? Please describe.

When applying an autocorrection, `Style/SymbolArray` and `Style/WordArray` collapse the array onto a single line.

```diff
-%w(
- a
- b
- c
-)
+["a", "b", "c"]
```

This is not ideal, as the original author may have chosen to specify the array over multiple lines for a reason, and the autocorrection ignores that.

## Describe the solution you'd like

Ideally, the autocorrection would preserve the same semantics when executing the replacement.

- arrays specified over multiple lines should stay specified over multiple lines
  ```diff
  -%w(
  -  a
  -  b
  -  c
  -)
  +[
  +  "a",
  +  "b",
  +  "c",
  +]
  ```
- arrays specified on a single line should stay specified over a single line
  ```diff
  -%w(a b c)
  +["a", "b", "c"]
  ```

If we were really fancy, we'd even preserve line breaks between the same elements
  ```diff
  -%w(a b
  -  c)
  +["a", "b",
  +  "c"]
  ```

PR #10784, opened 2022-07-04
Preserve multiline semantics on `Style/SymbolArray` and `Style/WordArray`

r7kamura:
To fix https://github.com/rubocop/rubocop/issues/10455.

Example:

```diff
# Before
-%w(
- foo
- bar
- baz
-)
+["foo", "bar", "baz"]

# After
-%w(
- foo
- bar
- baz
-)
+[
+  "foo",
+  "bar",
+  "baz"
+]
```

To mimic the original whitespace semantics, I decided to extract three types of whitespace and reuse them in autocorrection.

```ruby
# Before
"%w[\n  foo\n  bar\n  baz\n]\n"
#   ^^^^   ^^^^          ^^
#   |      |             `-- whitespace_trailing
#   |      `-- whitespace_between
#   `-- whitespace_leading

# After
"[\n  'foo',\n  'bar',\n  'baz'\n]\n"
# ^^^^      ^^^^      ^^^^     ^^
# |         |         |        `-- whitespace_trailing
# |         `------------ whitespace_between
# `-- whitespace_leading
```

Note that this commit also changes the offense message like this (on multiline case):

```ruby
# Before
%w(
^^^ Use `[
  'foo',
  'bar'
]` for an array of words.
  foo
  bar
)

# After
%w(
^^^ Use an array literal `[...]` for an array of words.
  foo
  bar
)
```

This means that the auto-corrected code no longer appears in the offense message (on multiline case). I think this is a reasonable change because this format is already used for percent styles: 

> 'Use `%w` or `%W` for an array of words.'`.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/symbol_array.rb
+++ b/lib/rubocop/cop/style/symbol_array.rb
@@ -39,7 +39,7 @@ module RuboCop
         minimum_target_ruby_version 2.0
 
         PERCENT_MSG = 'Use `%i` or `%I` for an array of symbols.'
-        ARRAY_MSG = 'Use `%<prefer>s` for an array of symbols.'
+        ARRAY_MSG = 'Use %<prefer>s for an array of symbols.'
 
         class << self
           attr_accessor :largest_brackets
@@ -74,8 +74,7 @@ module RuboCop
               to_symbol_literal(c.value.to_s)
             end
           end
-
-          "[#{syms.join(', ')}]"
+          build_bracketed_array_with_appropriate_whitespace(elements: syms, node: node)
         end
 
         def to_symbol_literal(string)
--- a/lib/rubocop/cop/style/word_array.rb
+++ b/lib/rubocop/cop/style/word_array.rb
@@ -44,7 +44,7 @@ module RuboCop
         extend AutoCorrector
 
         PERCENT_MSG = 'Use `%w` or `%W` for an array of words.'
-        ARRAY_MSG = 'Use `%<prefer>s` for an array of words.'
+        ARRAY_MSG = 'Use %<prefer>s for an array of words.'
 
         class << self
           attr_accessor :largest_brackets
@@ -92,8 +92,7 @@ module RuboCop
               to_string_literal(word.children[0])
             end
           end
-
-          "[#{words.join(', ')}]"
+          build_bracketed_array_with_appropriate_whitespace(elements: words, node: node)
         end
       end
     end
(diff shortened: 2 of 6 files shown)


lib/rubocop/cop/style/symbol_array.rb
```diff
@@ -39,7 +39,7 @@ class SymbolArray < Base
         minimum_target_ruby_version 2.0
 
         PERCENT_MSG = 'Use `%i` or `%I` for an array of symbols.'
-        ARRAY_MSG = 'Use `%<prefer>s` for an array of symbols.'
+        ARRAY_MSG = 'Use `[...]` for an array of symbols.'
```
koic, line 42: Can you tweak the message?

```suggestion
        ARRAY_MSG = 'Use array brackets `[...]` for an array of symbols.'
```

lib/rubocop/cop/style/word_array.rb
```diff
@@ -44,7 +44,7 @@ class WordArray < Base
         extend AutoCorrector
 
         PERCENT_MSG = 'Use `%w` or `%W` for an array of words.'
-        ARRAY_MSG = 'Use `%<prefer>s` for an array of words.'
+        ARRAY_MSG = 'Use `[...]` for an array of words.'
```
koic, line 47: Ditto.

```suggestion
        ARRAY_MSG = 'Use array brackets`[...]` for an array of words.'
```

lib/rubocop/cop/style/symbol_array.rb
r7kamura: Sure, thank you for pointing that out. I too was afraid that this `[...]` might be a bit confusing.

spec/rubocop/cop/style/symbol_array_spec.rb
```diff
@@ -201,7 +201,7 @@
     it 'registers an offense for array starting with %i' do
       expect_offense(<<~RUBY)
         %i(one two three)
-        ^^^^^^^^^^^^^^^^^ Use `[:one, :two, :three]` for an array of symbols.
+        ^^^^^^^^^^^^^^^^^ Use array brackets `[...]` for an array of symbols.
```
dvandersluis, line 204: I think the offense messages are less clear here than they used to be. Can we continue showing the full replacement if the array is on a single line?

Also, `Use an array literal […]` would probably be clearer wording for when we cannot show the whole array.

r7kamura: Indeed, that's true 👍 
I've made the changes exactly as they should be, so take another look at it.

dvandersluis: Thanks!

PR #10790, opened 2022-07-06
Support `EnforcedStyle: comment` option for `Style/EmptyElse`

ydah:
This PR is support `EnforcedStyle: comment` option for `Style/EmptyElse`.

It accepts same as `both`, but if there is a comment in `else`, it's acceptable

```ruby
if cond
  foo
else
  # some comment
end
```
I think that in the case of complex conditional statements, `else` is empty, but only the explanation of the `else` case is written in the comment.

`EnforcedStyle: comment` is didn't allow an empty `else`, but allow an `else` that is intentionally the comments.

-----------------

* [ ] Added tests.


config/default.yml
```diff
@@ -3491,10 +3491,12 @@ Style/EmptyElse:
   # empty - warn only on empty `else`
   # nil - warn on `else` with nil in it
   # both - warn on empty `else` and `else` with `nil` in it
+  # comment - same as `both`, but if there is a comment in `else`, it's acceptable
   SupportedStyles:
     - empty
     - nil
     - both
+    - comment
```
bbatsov, line 3499: Perhaps `allow_comment` would be a better name here? Thinking more about this - probably this shouldn't be a style but rather an option `AllowComments` that affects all styles.

bbatsov: (false by default)

ydah: Thank you very much. I changed to the `AllowComments` option. I think this is better too.

bbatsov:
Thanks!

Issue #10789, opened 2022-07-05
Lint/EmptyWhen: false positive with range


kazzix14:
## Expected behavior

RuboCop does not report an error.

## Actual behavior

RuboCop does report an error that seemed to be invalid.
```
root@5654c24326e8:/tmp/repro# bundle exec rubocop
The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file.

Please also note that you can opt-in to new cops by default by adding this to your config:
  AllCops:
    NewCops: enable

Gemspec/DeprecatedAttributeAssignment: # new in 1.30
  Enabled: true
Gemspec/RequireMFA: # new in 1.23
  Enabled: true
Layout/LineContinuationLeadingSpace: # new in 1.31
  Enabled: true
Layout/LineContinuationSpacing: # new in 1.31
  Enabled: true
Layout/LineEndStringConcatenationIndentation: # new in 1.18
  Enabled: true
Layout/SpaceBeforeBrackets: # new in 1.7
  Enabled: true
Lint/AmbiguousAssignment: # new in 1.7
  Enabled: true
Lint/AmbiguousOperatorPrecedence: # new in 1.21
  Enabled: true
Lint/AmbiguousRange: # new in 1.19
  Enabled: true
Lint/ConstantOverwrittenInRescue: # new in 1.31
  Enabled: true
Lint/DeprecatedConstants: # new in 1.8
  Enabled: true
Lint/DuplicateBranch: # new in 1.3
  Enabled: true
Lint/DuplicateRegexpCharacterClassElement: # new in 1.1
  Enabled: true
Lint/EmptyBlock: # new in 1.1
  Enabled: true
Lint/EmptyClass: # new in 1.3
  Enabled: true
Lint/EmptyInPattern: # new in 1.16
  Enabled: true
Lint/IncompatibleIoSelectWithFiberScheduler: # new in 1.21
  Enabled: true
Lint/LambdaWithoutLiteralBlock: # new in 1.8
  Enabled: true
Lint/NoReturnInBeginEndBlocks: # new in 1.2
  Enabled: true
Lint/NonAtomicFileOperation: # new in 1.31
  Enabled: true
Lint/NumberedParameterAssignment: # new in 1.9
  Enabled: true
Lint/OrAssignmentToConstant: # new in 1.9
  Enabled: true
Lint/RedundantDirGlobSort: # new in 1.8
  Enabled: true
[...]

PR #10792, opened 2022-07-06
Add new `Lint/RequireRangeParentheses` cop

koic:
## Context

It emulates the following Ruby warning.

```console
% cat example.rb
1...
2
```

```consle
% ruby example.rb
example.rb:1: warning: ... at EOL, should be parenthesized?
```

So, this cop will detect case like #10789.

## Summary

It checks that a range literal is enclosed in parentheses when the end of the range is at a line break.

### example

```ruby
# bad - Represents `(1..42)`, not endless range.
1..
42

# good - It's incompatible, but your intentions when using endless range may be:
(1..)
42

# good
1..42

# good
(1..42)

# good
(1..
42)
```

NOTE: The following is maybe intended for `(42..)`. But, compatible is `42..do_something`. So, this cop does not provide autocorrection because it is left to user.

```ruby
case condition
when 42..
  do_something
end
```

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/require_range_parentheses.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Checks that the range is enclosed in parentheses when the end of the range is line break.
+      #
+      # NOTE: The following is maybe intended for `(42..)`. But, compatible is `42..do_something`.
+      # So, this cop does not provide autocollection because it is left to user.
```
bbatsov, line 9: autocollection -> autocorrection

lib/rubocop/cop/lint/require_range_parentheses.rb
```diff
@@ -0,0 +1,54 @@
...
+      # ----
+      # case condition
+      # when 42..
+      #   do_something
+      # end
+      # ----
+      #
+      # @example
+      #
+      #   # bad
```
bbatsov, line 21: You might also add comments for the bad/good cases so it's clearer what they mean.

lib/rubocop/cop/lint/require_range_parentheses.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Checks that the range is enclosed in parentheses when the end of the range is line break.
```
bbatsov, line 6: that a range literal
is at a line break

lib/rubocop/cop/lint/require_range_parentheses.rb
```diff
@@ -0,0 +1,54 @@
...
+      #
+      #   # good
+      #   1..42
+      #
+      #   # good - It's incompatible, but your intentions when using endless range may be:
+      #   (1..)
+      #   42
+      #
+      class RequireRangeParentheses < Base
+        MSG = 'Use parentheses in the endless range to avoid confusion about precedence.'
```
bbatsov, line 40: Wrap the endless range literal %s to avoid precedence ambiguity.

koic:
I've updated this PR. Thank you!

PR #10797, opened 2022-07-08
[Fix #10263] Fix aliasing problem for Enabled in Config

### LINKED ISSUE #10263: configuration in child directory can corrupt configuration in parent directory
The general problem here is that rubocop is using different configuration for a single file depending on how it's run, and that configuration is wrong. I haven't completely nailed down everything that's affected, but it's at least a fairly simple repro.

Given the following four files:

`.rubocop.yml`:
```yml
require:
  - rubocop-rails

AllCops:
  TargetRubyVersion: 2.7
  NewCops: enable

Rails:
  Severity: error
```

`subdir/.rubocop.yml`:
```yml
inherit_from:
  - ../.rubocop.yml

Rails:
  Enabled: false
```

`test.rb`:
```ruby
# frozen_string_literal: true

User.where('name = ?', 'Gabe')
```

`subdir/test2.rb` (I'm not sure if this file is necessary in order for rubocop to process the configuration file in its directory):
```ruby
# frozen_string_literal: true

puts 'hi'
```

## Expected behavior

I would expect that rubocop will find an offense of `Rails/WhereEquals` regardless of where it's run. I.e. `rubocop`, `rubocop -A`, `rubocop -- test.rb`, `rubocop -A -- test.rb` should find a single offense each time.

## Actual behavior

If I run `rubocop` or `rubocop -A`, no offenses are found in either file (WRONG!). But if I run `rubocop -- test.rb`, the offense is found (RIGHT!). My supposition is that there's a bug in the inheritance mechanism that because the entire Rails department is disabled in the sub-directory that state leaks to the parent directory's copy of the configuration. But if you're running _just_ `test.rb`, the sub-directory's config is never loaded, not having a chance to corrupt the parent. I have reason to believe it's not as simple as that, though. In my full app, the only cops that seems to "regress" with this sort of configuration were all cops that are marked as pending. So it might be that Enabled is getting reset incorrectly somewhere. I'm also not sure if being a separate gem (rubocop-rails) is important at all - that just happened to be the department that I encountered this with.

## RuboCop version

```
1.22.3 (using Parser 3.0.2.0, rubocop-ast 1.13.0, running on ruby 2.7.2 x86_64-darwin20)
  - rubocop-rails 2.12.4
```

jonas054:
There was an aliasing problem in the `Config` class causing the `cop_options` local variable to reference the same object between different calls to `for_cop` even when the calls were to different `Config` objects with different `@hash` objects.

I have not been able to figure out the root cause. Only that the phenomenon occurs, and that duplicating the cop options before setting the `Enabled` property resolves it.


lib/rubocop/config.rb
```diff
@@ -25,7 +25,7 @@ def initialize(hash = {}, loaded_path = nil)
       @loaded_path = loaded_path
       @for_cop = Hash.new do |h, cop|
         qualified_cop_name = Cop::Registry.qualified_cop_name(cop, loaded_path)
-        cop_options = self[qualified_cop_name] || {}
+        cop_options = self[qualified_cop_name]&.dup || {}
```
koic, line 28: `&` seems unnecessary because `dup` is always callable.

```suggestion
        cop_options = self[qualified_cop_name].dup || {}
```

jonas054:
Thanks, @koic! Updated.

koic:
@jonas054 This looks good to me! Can you quash your commits into one?

jonas054:
@koic It's a single commit now.

koic:
Thanks!

PR #10801, opened 2022-07-09
[Fix #10613] Allow parallel inspection for autocorrect

### LINKED ISSUE #10613: make --parallel and --auto-correct work together
I can do it with xargs, so we can also do it internally 🤷 
... maybe it needs to say "rules xyz are not enforced because they would rename files" or whatever the reason was to disable this, but for 99% case they should just work! :)

jonas054:
Allow the `-P`/`--parallel` flag together with the autocorrection flags `--[safe-]auto-correct`/`-a` `--auto-correct-all`/`-A` and  `--fix-layout`/`-x`. Also make this the default, so that `--no-parallel` has to be given for a single core run.

The algorithm for normal parallel inspection is followed, so that a cache warm-up phase is run first. This phase only does caching of inspection results. It doesn't update any files. Then it does a single core run that takes advantage of the cached results and does the corrections. The speed gain from this approach is probably smaller than writing files in parallel, but it's a pretty small change that still outputs correction results in the same way as before.

Also fix the documentation for `--parallel`, which did not mention `--no-parallel`.

### DIFF
--- a/changelog/change_allow_parallel_autocorrect.md
+++ b/changelog/change_allow_parallel_autocorrect.md
@@ -1,0 +1,1 @@
+* [#10613](https://github.com/rubocop/rubocop/issues/10613): Allow autocorrecting with -P/--parallel and make it the default. ([@jonas054][])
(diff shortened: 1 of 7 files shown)


jonas054:
What gain in speed you will get from this change depends on two things:

1. the state of the cache
2. the percentage of files with no offenses

With a cold cache and offenses in 10% of the files, it can cause `rubocop -a` to execute in about half the time in a large project.

You can get the same effect (guarantee a warm cache) in the current version by running `rubocop; rubocop -a`.

bbatsov:
Looks like a nice improvement to me. Thanks!

P.S. Probably we should document somewhere how the autocorrect works, so that people are not confused it writes the corrections in parallel.

PR #10804, opened 2022-07-11
Support --server with JRuby

ccutrer:
by using spawn instead of fork when fork is unavailable. it's a bit
slower to startup since it has to clean-start a new jruby process,
but subsequent runs still get massive benefits.

### DIFF
--- a/docs/modules/ROOT/pages/usage/server.adoc
+++ b/docs/modules/ROOT/pages/usage/server.adoc
@@ -13,7 +13,11 @@ Normally RuboCop starts somewhat slowly because it needs to `require` a ton of f
 slow. With the RuboCop server we sidestep this nasty issue and make it much more pleasant to
 interact with RuboCop from text editors and IDEs.
 
-NOTE: The feature cannot be used on JRuby and Windows, as they do not support the `fork` system call.
+NOTE: As JRuby and Windows do not support the `fork` system call, they have a
+slightly different implementation that spawns a new process. This means the
+startup time for a new instance of the background server is even longer than
+just running RuboCop directly, but once booted they still experience
+significant improvements for subsequent runs.
 
 == Run with Server
 
@@ -48,6 +52,8 @@ $ ps aux | grep 'rubocop --server'
 user             16060   0.0  0.0  5078568   2264   ??  S     7:54AM   0:00.00 rubocop --server /Users/user/src/github.com/rubocop/rubocop
 user             16337   0.0  0.0  5331560   2396   ??  S    23:51PM   0:00.00 rubocop --server /Users/user/src/github.com/rubocop/rubocop-rails
 ```
+
+NOTE: For JRuby and Windows, it might be `rubocop --start-server`.
 
 When you update and run `rubocop`, the server process will restart automatically.
 
--- a/lib/rubocop/server/cli.rb
+++ b/lib/rubocop/server/cli.rb
@@ -31,12 +31,6 @@ module RuboCop
       end
 
       def run(argv = ARGV)
-        unless Server.support_server?
-          return error('RuboCop server is not supported by this Ruby.') if use_server_option?(argv)
-
-          return STATUS_SUCCESS
-        end
-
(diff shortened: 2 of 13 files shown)


lib/rubocop/server/cli.rb
```diff
@@ -22,7 +22,13 @@ class CLI
...
 
       SERVER_OPTIONS = %w[
-        --server --no-server --server-status --restart-server --start-server --stop-server
+        --server
+        --no-server
+        --server-status
+        --restart-server
+        --start-server
+        --stop-server
+        --spawn-server
```
bbatsov, line 31: Why do we need a separate command for this? Can't we just have a switch inside `--server/start-server` for this?

bbatsov:
Apart from my remark about the `--spawn-server` flag we'll also need to document the differences on JRuby in the docs.

ccutrer:
I've reworked it to re-use the --start-server command. I can't use additional arguments to differentiate that this is the child process, due to how the Server::CLI extracts pertinent arguments, but using an env var is easy enough. I've also updated the docs. I'm going to test on Windows shortly.

ccutrer:
I've got some issues on Windows I'm working through :(. I don't have time to fix them tonight though.

bbatsov:
No rush! 

@koic Please, take a look as well.

ccutrer:
🤞 it works well on Windows now. It definitely does on `ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x64-mingw32]`, but looking at the GH Action workflow it seems like there's a non-mingw version of ruby that I don't know how to install locally to test with.

koic:
The Windows CI seems to be failing. I don't have Windows so I can't verify it. Anyway, I would like to start the review after passing the CI. And If the CircleCI doesn't work after the next force push, can you contact CircleCI?

ccutrer:
@koic: was that directed at me? I have a windows machine, and tests are passing for me, but I don't know how to set up the ruby environments running in the tests (btw, it looks like the failing Windows runs are on GitHub Actions, not CircleCI). Do you have any suggestions on how to set up the proper Rubies? I've just used the RubyInstaller package.

koic:
@ccutrer 

> was that directed at me? 

Yes.

> Do you have any suggestions on how to set up the proper Rubies?

Unfortunately, I'm not familiar with Windows environment so it's hard for me to suggest it.

dvandersluis:
@ccutrer you should be able to use rbenv via WSL on windows I believe: https://medium.com/alterway/use-rbenv-ruby-on-windows-10-linux-wsl-a9bce8d97300

ccutrer:
> @ccutrer you should be able to use rbenv via WSL on windows I believe: https://medium.com/alterway/use-rbenv-ruby-on-windows-10-linux-wsl-a9bce8d97300

won't this give me a linux version of ruby though?

dvandersluis:
Looks like github actions installs each ruby normally and just uses ruby from a directory:

image

docs/modules/ROOT/pages/usage/server.adoc
```diff
@@ -49,6 +53,8 @@ user             16060   0.0  0.0  5078568   2264   ??  S     7:54AM   0:00.00 r
 user             16337   0.0  0.0  5331560   2396   ??  S    23:51PM   0:00.00 rubocop --server /Users/user/src/github.com/rubocop/rubocop-rails
 ```
 
+NOTE: For JRuby and Windows, it might be `rubocop --start-server`.
```
dvandersluis, line 56: What does this note mean?

ccutrer: I can reword that as "the process name might be"

dvandersluis: Ah that makes sense to me now, thanks

bbatsov:
Perhaps this will work if rebased on top of `master`? (after@tdeo's recent changes)

bbatsov:

PR #10805, opened 2022-07-11
Follow-up clean-up of Ruby < 2.6

pirj:
This complements #10577, considering #10625 an #10697.

As we don't run specs on CI for Ruby < 2.5, those specs that would only run on those Ruby versions are now useless.

Spec changes are better viewed [without whitespace](https://github.com/rubocop/rubocop/pull/10805/files?diff=split&w=1).

-----------------

* [-] Added tests.

### DIFF
--- a/lib/rubocop/cop/util.rb
+++ b/lib/rubocop/cop/util.rb
@@ -159,7 +159,7 @@ module RuboCop
       private
 
       def compatible_external_encoding_for?(src)
-        src = src.dup if RUBY_VERSION < '2.3' || RUBY_ENGINE == 'jruby'
+        src = src.dup if RUBY_ENGINE == 'jruby'
         src.force_encoding(Encoding.default_external).valid_encoding?
       end
     end
--- a/spec/rubocop/cop/lint/useless_else_without_rescue_spec.rb
+++ b/spec/rubocop/cop/lint/useless_else_without_rescue_spec.rb
@@ -1,19 +1,6 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::Lint::UselessElseWithoutRescue, :config do
-  context 'with `else` without `rescue`', :ruby25 do
-    it 'registers an offense' do
-      expect_offense(<<~RUBY)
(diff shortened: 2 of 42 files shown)


lib/rubocop/cop/util.rb
```diff
@@ -159,7 +159,7 @@ def to_supported_styles(enforced_style)
       private
 
       def compatible_external_encoding_for?(src)
-        src = src.dup if RUBY_VERSION < '2.3' || RUBY_ENGINE == 'jruby'
+        src = src.dup if RUBY_ENGINE == 'jruby'
```
pirj, line 162: I'm uncertain if this is still needed on JRuby > 9.2.

spec/rubocop/cop/lint/useless_else_without_rescue_spec.rb
```diff
@@ -1,19 +1,6 @@
...
-      expect_offense(<<~RUBY)
-        begin
-          do_something
-        else
-        ^^^^ `else` without `rescue` is useless.
-          handle_unknown_errors
-        end
-      RUBY
-    end
-  end
```
pirj, line 15: This fails on Ruby 2.6+, and we don't run specs on CI for them.
Is this spec still useful? Is it still possible to run RuboCop specs locally on Ruby 2.5?

dvandersluis:
FYI this will revert a series of older ruby support restoration PRs starting at #10626

koic:
Ruby 2.5 and below were dropped in runtime versions, not analysis versions. I'm going close this PR because these codes need to be kept as analysis versions. Thank you.

pirj:
> revert a series of older ruby support restoration PRs

I'd argue with that. My two points are:
 - the same specs are still run across the range of Ruby versions that we run on CI
 - we're properly setting the `minimum_target_ruby_version` even if it is 2.5 or 2.6, so RuboCop would be able to run the code that is targeted to earlier versions

Still, I might be wrong with those points, or might be missing something else, so I'll be happy to adjust/close the PR.

PR #10809, opened 2022-07-12
Enhanced documentation for "IgnoredMethods" option

ydah:
This PR is documentation was added according to the following policy for "IgnoredMethods" option

- Add code examples for cop that don't have code examples
- Add default values for cop that don't have default values

The code example also clearly shows the difference in behavior from the default when the default for "IgnoredMethods" is an empty array.

I thought We could flag "IgnoredMethods" as obsolete and merge it into "AllowedMethods".
Refs (case of "AllowedPatterns") : https://github.com/rubocop/rubocop/pull/10555

In the future, if it looks like it would be better to merge them, I would like to handle it other PR.

-----------------

* [-] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/ambiguous_block_association.rb
+++ b/lib/rubocop/cop/lint/ambiguous_block_association.rb
@@ -7,6 +7,7 @@ module RuboCop
       # when param passed without parentheses.
       #
       # This cop can customize ignored methods with `IgnoredMethods`.
+      # By default, there are no methods to ignored.
       #
       # @example
       #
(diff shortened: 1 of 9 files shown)


bbatsov:
> I thought We could flag "IgnoredMethods" as obsolete and merge it into "AllowedMethods".

@ydah Yes, `IgnoredMethods` should be made obsolete and replaced by `AllowedMethods` everywhere. More broadly speaking - we should make the configuration of individual cops consistent across RuboCop. Unfortunately some of the older cops use inconsistent/legacy names for some of their options. (back then we didn't think much about consistency)

ydah:
@bbatsov Thank you for your response!

> More broadly speaking - we should make the configuration of individual cops consistent across RuboCop. Unfortunately some of the older cops use inconsistent/legacy names for some of their options. (back then we didn't think much about consistency)

I see. If there is anything that can be improved for consistency in the future, I would like to also contribute on that :-)

PR #10811, opened 2022-07-13
Fix Style/ClassAndModuleChildren compaction for oneliner children

Morriar:
`Style/ClassAndModuleChildren` with `EnforcedStyle: Compact` doesn't handle definitions on a single line properly.

Consider this example:

```rb
class Foo
  class Bar; end
end
```

Before this PR, the autocorrect was producing code with a syntax error:

```rb
class Foo::Barend
```

After this PR, the autocorrect is producing the correct code:

```rb
class Foo::Bar; end
```

This also fixes an error that was happening when autocorrecting this code:

```rb
class Foo
  class Bar end
end
```

Which was raising a `Parser::ClobberingError: Parser::Source::TreeRewriter detected clobbering` error.

After this PR, the autocorrect works properly:

```rb
class Foo::Bar end
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_style_class_and_module_children.md
+++ b/changelog/fix_style_class_and_module_children.md
@@ -1,0 +1,1 @@
+* Fix Style/ClassAndModuleChildren compaction for oneliner children. ([@Morriar][])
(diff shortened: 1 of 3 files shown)


bbatsov:

koic:
The fix conflicted with #10826. Sorry for the late notice...

PR #10812, opened 2022-07-13
Introduce new AllowMultilineFinalElement option for all __LineBreak(s) cops

Korri:
This add a `AllowMultilineFinalElement` option, that when set to `true` allows the following cops to not consider a statement to be multiline even if the last (or only) element (value, argument or parameter) itself is multiline:

```
FirstArrayElementLineBreak
FirstHashElementLineBreak
FirstMethodArgumentLineBreak
FirstMethodParameterLineBreak
MultilineArrayLineBreaks
MultilineHashKeyLineBreaks
MultilineMethodArgumentLineBreaks
MultilineMethodParameterLineBreaks
```

The goal of this addition is to be able to use `Layout/LineLength` to format each elements on a new line, while not forcing some of the use cases that are considered "ok exceptions" by some people.

### Examples of exceptions that those changes make possible

```ruby
# Allow final element to be multiline

# good
foo(a, b, bar(
  c,
  d
))

# good
{ a: b, c: {
 foo: bar,
 baz: qux,
}}

# good
[ 1, 2, [
 foo,
 bar,
]]

# Single elements are ignored too

# good
foo({
 a: b,
})

# good
a = { foo: [
  bar,
  baz,
]}

# good
b = [{
  foo: bar
}]
```

### While still correcting the followings

```ruby
# bad
foo(bar, baz,
bat)

[1, 2
3]

# good
foo(
  bar,
  baz,
  bat,
)
[
  1,
  2,
  3,
]
```

-----------------

* [x] Added tests.

### DIFF
--- a/docs/modules/ROOT/pages/cops_layout.adoc
+++ b/docs/modules/ROOT/pages/cops_layout.adoc
@@ -2914,7 +2914,7 @@ multi-line array.
 |===
 | Name | Default value | Configurable values
 
-| AllowMultilineFinalElement
+| xref:cops_layout.adoc#allowmultilinefinalelement[AllowMultilineFinalElement]
 | `false`
 | Boolean
 |===
@@ -3144,7 +3144,7 @@ multi-line hash.
 |===
 | Name | Default value | Configurable values
 
-| AllowMultilineFinalElement
+| xref:cops_layout.adoc#allowmultilinefinalelement[AllowMultilineFinalElement]
 | `false`
 | Boolean
 |===
@@ -3245,7 +3245,7 @@ method foo, bar,
 |===
 | Name | Default value | Configurable values
 
-| AllowMultilineFinalElement
+| xref:cops_layout.adoc#allowmultilinefinalelement[AllowMultilineFinalElement]
 | `false`
 | Boolean
 |===
@@ -3342,7 +3342,7 @@ end
 |===
 | Name | Default value | Configurable values
 
-| AllowMultilineFinalElement
+| xref:cops_layout.adoc#allowmultilinefinalelement[AllowMultilineFinalElement]
 | `false`
 | Boolean
 |===
@@ -4795,7 +4795,7 @@ starts on a separate line.
 |===
 | Name | Default value | Configurable values
 
-| AllowMultilineFinalElement
+| xref:cops_layout.adoc#allowmultilinefinalelement[AllowMultilineFinalElement]
 | `false`
 | Boolean
 |===
@@ -5165,7 +5165,7 @@ starts on a separate line.
 |===
 | Name | Default value | Configurable values
 
-| AllowMultilineFinalElement
+| xref:cops_layout.adoc#allowmultilinefinalelement[AllowMultilineFinalElement]
 | `false`
 | Boolean
 |===
@@ -5263,7 +5263,7 @@ foo(
 |===
 | Name | Default value | Configurable values
 
-| AllowMultilineFinalElement
+| xref:cops_layout.adoc#allowmultilinefinalelement[AllowMultilineFinalElement]
 | `false`
 | Boolean
 |===
@@ -5709,7 +5709,7 @@ end
 |===
 | Name | Default value | Configurable values
 
-| AllowMultilineFinalElement
+| xref:cops_layout.adoc#allowmultilinefinalelement[AllowMultilineFinalElement]
 | `false`
 | Boolean
 |===
@@ -7615,3 +7615,5 @@ RUBY
 === References
 
 * https://rubystyle.guide#no-trailing-whitespace
+
+include::../partials/cops_layout_footer.adoc[]
--- a/docs/modules/ROOT/pages/cops_layout.adoc
+++ b/docs/modules/ROOT/pages/cops_layout.adoc
@@ -2859,6 +2859,36 @@ multi-line array.
 
 === Examples
 
+==== AllowMultilineFinalElement: false (default)
+
+[source,ruby]
+----
+# bad
+[ :a,
+  :b]
+
+# bad
+[ :a, {
+  :b => :c
+}]
+
+# good
+[:a, :b]
+
+# good
+[
+  :a,
+  :b]
+
+# good
+[
+  :a, {
+  :b => :c
+}]
+----
+
+==== AllowMultilineFinalElement: true
+
 [source,ruby]
 ----
 # bad
@@ -2866,10 +2896,28 @@ multi-line array.
   :b]
 
 # good
+[ :a, {
+  :b => :c
+}]
+
+# good
 [
   :a,
   :b]
+
+# good
+[:a, :b]
 ----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| AllowMultilineFinalElement
+| `false`
+| Boolean
+|===
 
 == Layout/FirstHashElementIndentation
 
@@ -3035,17 +3083,71 @@ multi-line hash.
 
 === Examples
 
+==== AllowMultilineFinalElement: false (default)
+
 [source,ruby]
 ----
 # bad
 { a: 1,
   b: 2}
 
+# bad
+{ a: 1, b: {
+  c: 3
+}}
+
 # good
 {
   a: 1,
   b: 2 }
+
+# good
+{
+  a: 1, b: {
+  c: 3
+}}
 ----
+
+==== AllowMultilineFinalElement: true
+
+[source,ruby]
+----
+# bad
+{ a: 1,
+  b: 2}
+
+# bad
+{ a: 1,
+  b: {
+  c: 3
+}}
+
+# good
+{ a: 1, b: {
+  c: 3
+}}
+
+# good
+{
+  a: 1,
+  b: 2 }
+
+# good
+{
+  a: 1, b: {
+  c: 3
+}}
+----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| AllowMultilineFinalElement
+| `false`
+| Boolean
+|===
 
 == Layout/FirstMethodArgumentLineBreak
 
@@ -3064,21 +3166,89 @@ multi-line method call.
 
 === Examples
 
+==== AllowMultilineFinalElement: false (default)
+
 [source,ruby]
 ----
 # bad
 method(foo, bar,
   baz)
 
+# bad
+method(foo, bar, {
+  baz: "a",
+  qux: "b",
+})
+
 # good
 method(
   foo, bar,
   baz)
 
+# good
+method(
+  foo, bar, {
+  baz: "a",
+  qux: "b",
+})
+
 # ignored
 method foo, bar,
   baz
 ----
+
+==== AllowMultilineFinalElement: true
+
+[source,ruby]
+----
+# bad
+method(foo, bar,
+  baz)
+
+# bad
+method(foo,
+  bar,
+  {
+    baz: "a",
+    qux: "b",
+  }
+)
+
+# good
+method(foo, bar, {
+  baz: "a",
+  qux: "b",
+})
+
+# good
+method(
+  foo, bar,
+  baz)
+
+# good
+method(
+  foo,
+  bar,
+  {
+    baz: "a",
+    qux: "b",
+  }
+)
+
+# ignored
+method foo, bar,
+  baz
+----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| AllowMultilineFinalElement
+| `false`
+| Boolean
+|===
 
 == Layout/FirstMethodParameterLineBreak
 
@@ -3097,11 +3267,59 @@ multi-line method parameter definition.
 
 === Examples
 
+==== AllowMultilineFinalElement: false (default)
+
 [source,ruby]
 ----
 # bad
 def method(foo, bar,
     baz)
+  do_something
+end
+
+# bad
+def method(foo, bar, baz = {
+  :a => "b",
+})
+  do_something
+end
+
+# good
+def method(
+    foo, bar,
+    baz)
+  do_something
+end
+
+# good
+def method(
+  foo, bar, baz = {
+  :a => "b",
+})
+  do_something
+end
+
+# ignored
+def method foo,
+    bar
+  do_something
+end
+----
+
+==== AllowMultilineFinalElement: true
+
+[source,ruby]
+----
+# bad
+def method(foo, bar,
+    baz)
+  do_something
+end
+
+# good
+def method(foo, bar, baz = {
+  :a => "b",
+})
   do_something
 end
 
@@ -3118,6 +3336,16 @@ def method foo,
   do_something
 end
 ----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| AllowMultilineFinalElement
+| `false`
+| Boolean
+|===
 
 == Layout/FirstParameterIndentation
 
@@ -4498,6 +4726,40 @@ starts on a separate line.
 
 === Examples
 
+==== AllowMultilineFinalElement: false (default)
+
+[source,ruby]
+----
+# bad
+[
+  a, b,
+  c
+]
+
+# bad
+[ a, b, foo(
+  bar
+)]
+
+# good
+[
+  a,
+  b,
+  c
+]
+
+# good
+[
+  a,
+  b,
+  foo(
+    bar
+  )
+]
+----
+
+==== AllowMultilineFinalElement: true
+
 [source,ruby]
 ----
 # bad
@@ -4507,12 +4769,36 @@ starts on a separate line.
 ]
 
 # good
+[ a, b, foo(
+  bar
+)]
+
+# good
 [
   a,
   b,
   c
 ]
+
+# good
+[
+  a,
+  b,
+  foo(
+    bar
+  )
+]
 ----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| AllowMultilineFinalElement
+| `false`
+| Boolean
+|===
 
 == Layout/MultilineAssignmentLayout
 
@@ -4812,6 +5098,39 @@ starts on a separate line.
 
 === Examples
 
+==== AllowMultilineFinalElement: false (default)
+
+[source,ruby]
+----
+# bad
+{
+  a: 1, b: 2,
+  c: 3
+}
+
+# bad
+{ a: 1, b: {
+  c: 3,
+}}
+
+# good
+{
+  a: 1,
+  b: 2,
+  c: 3
+}
+
+# good
+{
+  a: 1,
+  b: {
+    c: 3,
+  }
+}
+----
+
+==== AllowMultilineFinalElement: true
+
 [source,ruby]
 ----
 # bad
@@ -4821,12 +5140,35 @@ starts on a separate line.
 }
 
 # good
+{ a: 1, b: {
+  c: 3,
+}}
+
+# good
 {
   a: 1,
   b: 2,
   c: 3
 }
+
+# good
+{
+  a: 1,
+  b: {
+    c: 3,
+  }
+}
 ----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| AllowMultilineFinalElement
+| `false`
+| Boolean
+|===
 
 == Layout/MultilineMethodArgumentLineBreaks
 
@@ -4848,12 +5190,19 @@ be on a separate line, see `Layout/FirstMethodArgumentLineBreak`.
 
 === Examples
 
+==== AllowMultilineFinalElement: false (default)
+
 [source,ruby]
 ----
 # bad
 foo(a, b,
   c
 )
+
+# bad
+foo(a, b, {
+  foo: "bar",
+})
 
 # good
 foo(
@@ -4864,7 +5213,60 @@ foo(
 
 # good
 foo(a, b, c)
+
+# good
+foo(
+  a,
+  b,
+  {
+    foo: "bar",
+  }
+)
 ----
+
+==== AllowMultilineFinalElement: true
+
+[source,ruby]
+----
+# bad
+foo(a, b,
+  c
+)
+
+# good
+foo(a, b, {
+  foo: "bar",
+})
+
+# good
+foo(
+  a,
+  b,
+  c
+)
+
+# good
+foo(a, b, c)
+
+# good
+foo(
+  a,
+  b,
+  {
+    foo: "bar",
+  }
+)
+----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| AllowMultilineFinalElement
+| `false`
+| Boolean
+|===
 
 == Layout/MultilineMethodCallBraceLayout
 
@@ -5226,12 +5628,20 @@ be on a separate line, see `Layout/FirstMethodParameterLineBreak`.
 
 === Examples
 
+==== AllowMultilineFinalElement: false (default)
+
 [source,ruby]
 ----
 # bad
 def foo(a, b,
   c
 )
+end
+
+# bad
+def foo(a, b = {
+  foo: "bar",
+})
 end
 
 # good
@@ -5243,9 +5653,66 @@ def foo(
 end
 
 # good
+def foo(
+  a,
+  b = {
+    foo: "bar",
+  }
+)
+end
+
+# good
 def foo(a, b, c)
 end
 ----
+
+==== AllowMultilineFinalElement: true
+
+[source,ruby]
+----
+# bad
+def foo(a, b,
+  c
+)
+end
+
+# good
+def foo(a, b = {
+  foo: "bar",
+})
+end
+
+# good
+def foo(
+  a,
+  b,
+  c
+)
+end
+
+# good
+def foo(
+  a,
+  b = {
+    foo: "bar",
+  }
+)
+end
+
+# good
+def foo(a, b, c)
+end
+----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| AllowMultilineFinalElement
+| `false`
+| Boolean
+|===
 
 == Layout/MultilineOperationIndentation
 
--- a/lib/rubocop/cop/layout/first_array_element_line_break.rb
+++ b/lib/rubocop/cop/layout/first_array_element_line_break.rb
@@ -6,17 +6,49 @@ module RuboCop
       # Checks for a line break before the first element in a
       # multi-line array.
       #
-      # @example
+      # @example AllowMultilineFinalElement: false (default)
       #
       #     # bad
       #     [ :a,
       #       :b]
+      #
+      #     # bad
+      #     [ :a, {
+      #       :b => :c
+      #     }]
+      #
+      #     # good
+      #     [:a, :b]
       #
       #     # good
       #     [
       #       :a,
       #       :b]
       #
+      #     # good
+      #     [
+      #       :a, {
+      #       :b => :c
+      #     }]
+      #
+      # @example AllowMultilineFinalElement: true
+      #
+      #     # bad
+      #     [ :a,
+      #       :b]
+      #
+      #     # good
+      #     [ :a, {
+      #       :b => :c
+      #     }]
+      #
+      #     # good
+      #     [
+      #       :a,
+      #       :b]
+      #
+      #     # good
+      #     [:a, :b]
       class FirstArrayElementLineBreak < Base
         include FirstElementLineBreak
         extend AutoCorrector
@@ -26,7 +58,7 @@ module RuboCop
         def on_array(node)
           return if !node.loc.begin && !assignment_on_same_line?(node)
 
-          check_children_line_break(node, node.children)
+          check_children_line_break(node, node.children, ignore_last: ignore_last_element?)
         end
 
         private
@@ -34,6 +66,10 @@ module RuboCop
         def assignment_on_same_line?(node)
           source = node.source_range.source_line[0...node.loc.column]
           /\s*=\s*$/.match?(source)
+        end
+
+        def ignore_last_element?
+          !!cop_config['AllowMultilineFinalElement']
         end
       end
     end
--- a/lib/rubocop/cop/layout/first_method_argument_line_break.rb
+++ b/lib/rubocop/cop/layout/first_method_argument_line_break.rb
@@ -6,16 +6,69 @@ module RuboCop
       # Checks for a line break before the first argument in a
       # multi-line method call.
       #
-      # @example
+      # @example AllowMultilineFinalElement: false (default)
       #
       #     # bad
       #     method(foo, bar,
       #       baz)
       #
+      #     # bad
+      #     method(foo, bar, {
+      #       baz: "a",
+      #       qux: "b",
+      #     })
+      #
       #     # good
       #     method(
       #       foo, bar,
       #       baz)
+      #
+      #     # good
+      #     method(
+      #       foo, bar, {
+      #       baz: "a",
+      #       qux: "b",
+      #     })
+      #
+      #     # ignored
+      #     method foo, bar,
+      #       baz
+      #
+      # @example AllowMultilineFinalElement: true
+      #
+      #     # bad
+      #     method(foo, bar,
+      #       baz)
+      #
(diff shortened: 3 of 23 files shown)


bbatsov:
I'm definitely not very excited by this option, as I find such indentation to be somewhat questionable but I'll consider supporting it if it's popular in the wild. @rubocop/rubocop-core Any thoughts?

Also the name of the flag can be better - e.g. `AllowMultilineFinalElement` or something like this. (typically our options are named `AllowX` in such cases.

koic:
> I'm definitely not very excited by this option, as I find such indentation to be somewhat questionable

Me too. Introducing this configuration will increase maintenance costs across some cop. TBH, If this rule isn't very major, it's not aggressive to add. OTOH, I'm also interested in other developer opinions that this is useful.

Korri:
I looked into what ESlint does and it's interesting, looks like it's not a specific option, it's just "on" by default:

For method calls both single argument, and last argument are considered ok as multiline, they consider the following to be single line for [function-call-argument-newline](https://eslint.org/docs/latest/rules/function-call-argument-newline):
```javascript
# good
bar("one", "two", {
    one: 1,
    two: 2
});

# good
bar({
    one: 1,
    two: 2
});
```

For arrays though, they [only consider single element arrays to be ok on multiple lines](https://eslint.org/play#eyJ0ZXh0IjoiLyogZXNsaW50IHF1b3RlczogW1wiZXJyb3JcIiwgXCJkb3VibGVcIl0gKi9cblxuXG5jb25zdCBhID0gW3tcbiAgYmFyOiBcImJhdFwiXG59XVxuY29uc3QgYiA9IFtcImhlbGxvXCIsIHtcbiAgYmFyOiBcImJhdFwiXG59XVxuXG5hICsgYiIsIm9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hVmVyc2lvbiI6ImxhdGVzdCIsInNvdXJjZVR5cGUiOiJzY3JpcHQiLCJlY21hRmVhdHVyZXMiOnt9fSwicnVsZXMiOnsiY29uc3RydWN0b3Itc3VwZXIiOlsiZXJyb3IiXSwiZm9yLWRpcmVjdGlvbiI6WyJlcnJvciJdLCJnZXR0ZXItcmV0dXJuIjpbImVycm9yIl0sIm5vLWFzeW5jLXByb21pc2UtZXhlY3V0b3IiOlsiZXJyb3IiXSwibm8tY2FzZS1kZWNsYXJhdGlvbnMiOlsiZXJyb3IiXSwibm8tY2xhc3MtYXNzaWduIjpbImVycm9yIl0sIm5vLWNvbXBhcmUtbmVnLXplcm8iOlsiZXJyb3IiXSwibm8tY29uZC1hc3NpZ24iOlsiZXJyb3IiXSwibm8tY29uc3QtYXNzaWduIjpbImVycm9yIl0sIm5vLWNvbnN0YW50LWNvbmRpdGlvbiI6WyJlcnJvciJdLCJuby1jb250cm9sLXJlZ2V4IjpbImVycm9yIl0sIm5vLWRlYnVnZ2VyIjpbImVycm9yIl0sIm5vLWRlbGV0ZS12YXIiOlsiZXJyb3IiXSwibm8tZHVwZS1hcmdzIjpbImVycm9yIl0sIm5vLWR1cGUtY2xhc3MtbWVtYmVycyI6WyJlcnJvciJdLCJuby1kdXBlLWVsc2UtaWYiOlsiZXJyb3IiXSwibm8tZHVwZS1rZXlzIjpbImVycm9yIl0sIm5vLWR1cGxpY2F0ZS1jYXNlIjpbImVycm9yIl0sIm5vLWVtcHR5IjpbImVycm9yIl0sIm5vLWVtcHR5LWNoYXJhY3Rlci1jbGFzcyI6WyJlcnJvciJdLCJuby1lbXB0eS1wYXR0ZXJuIjpbImVycm9yIl0sIm5vLWV4LWFzc2lnbiI6WyJlcnJvciJdLCJuby1leHRyYS1ib29sZWFuLWNhc3QiOlsiZXJyb3IiXSwibm8tZXh0cmEtc2VtaSI6WyJlcnJvciJdLCJuby1mYWxsdGhyb3VnaCI6WyJlcnJvciJdLCJuby1mdW5jLWFzc2lnbiI6WyJlcnJvciJdLCJuby1nbG9iYWwtYXNzaWduIjpbImVycm9yIl0sIm5vLWltcG9ydC1hc3NpZ24iOlsiZXJyb3IiXSwibm8taW5uZXItZGVjbGFyYXRpb25zIjpbImVycm9yIl0sIm5vLWludmFsaWQtcmVnZXhwIjpbImVycm9yIl0sIm5vLWlycmVndWxhci13aGl0ZXNwYWNlIjpbImVycm9yIl0sIm5vLWxvc3Mtb2YtcHJlY2lzaW9uIjpbImVycm9yIl0sIm5vLW1pc2xlYWRpbmctY2hhcmFjdGVyLWNsYXNzIjpbImVycm9yIl0sIm5vLW1peGVkLXNwYWNlcy1hbmQtdGFicyI6WyJlcnJvciJdLCJuby1uZXctc3ltYm9sIjpbImVycm9yIl0sIm5vLW5vbm9jdGFsLWRlY2ltYWwtZXNjYXBlIjpbImVycm9yIl0sIm5vLW9iai1jYWxscyI6WyJlcnJvciJdLCJuby1vY3RhbCI6WyJlcnJvciJdLCJuby1wcm90b3R5cGUtYnVpbHRpbnMiOlsiZXJyb3IiXSwibm8tcmVkZWNsYXJlIjpbImVycm9yIl0sIm5vLXJlZ2V4LXNwYWNlcyI6WyJlcnJvciJdLCJuby1zZWxmLWFzc2lnbiI6WyJlcnJvciJdLCJuby1zZXR0ZXItcmV0dXJuIjpbImVycm9yIl0sIm5vLXNoYWRvdy1yZXN0cmljdGVkLW5hbWVzIjpbImVycm9yIl0sIm5vLXNwYXJzZS1hcnJheXMiOlsiZXJyb3IiXSwibm8tdGhpcy1iZWZvcmUtc3VwZXIiOlsiZXJyb3IiXSwibm8tdW5kZWYiOlsiZXJyb3IiXSwibm8tdW5leHBlY3RlZC1tdWx0aWxpbmUiOlsiZXJyb3IiXSwibm8tdW5yZWFjaGFibGUiOlsiZXJyb3IiXSwibm8tdW5zYWZlLWZpbmFsbHkiOlsiZXJyb3IiXSwibm8tdW5zYWZlLW5lZ2F0aW9uIjpbImVycm9yIl0sIm5vLXVuc2FmZS1vcHRpb25hbC1jaGFpbmluZyI6WyJlcnJvciJdLCJuby11bnVzZWQtbGFiZWxzIjpbImVycm9yIl0sIm5vLXVudXNlZC12YXJzIjpbImVycm9yIl0sIm5vLXVzZWxlc3MtYmFja3JlZmVyZW5jZSI6WyJlcnJvciJdLCJuby11c2VsZXNzLWNhdGNoIjpbImVycm9yIl0sIm5vLXVzZWxlc3MtZXNjYXBlIjpbImVycm9yIl0sIm5vLXdpdGgiOlsiZXJyb3IiXSwicmVxdWlyZS15aWVsZCI6WyJlcnJvciJdLCJ1c2UtaXNuYW4iOlsiZXJyb3IiXSwidmFsaWQtdHlwZW9mIjpbImVycm9yIl0sImFycmF5LWVsZW1lbnQtbmV3bGluZSI6WyJlcnJvciJdfSwiZW52Ijp7ImVzNiI6dHJ1ZX19fQ==):
```javascript
# good
const a = [{
  bar: "bat"
}]

# bad
const b = ["hello", {
  bar: "bat"
}]
```

Objects [behave the same as arrays](https://eslint.org/play#eyJ0ZXh0IjoiLyogZXNsaW50IHF1b3RlczogW1wiZXJyb3JcIiwgXCJkb3VibGVcIl0gKi9cblxuXG5jb25zdCBhID0ge2Zvbzoge1xuICBiYXI6IFwiYmF0XCJcbn19XG5jb25zdCBiID0ge2E6IFwiQVwiLCBmb286IHtcbiAgYmFyOiBcImJhdFwiXG59fVxuXG5hICsgYiIsIm9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hVmVyc2lvbiI6ImxhdGVzdCIsInNvdXJjZVR5cGUiOiJzY3JpcHQiLCJlY21hRmVhdHVyZXMiOnt9fSwicnVsZXMiOnsiY29uc3RydWN0b3Itc3VwZXIiOlsiZXJyb3IiXSwiZm9yLWRpcmVjdGlvbiI6WyJlcnJvciJdLCJnZXR0ZXItcmV0dXJuIjpbImVycm9yIl0sIm5vLWFzeW5jLXByb21pc2UtZXhlY3V0b3IiOlsiZXJyb3IiXSwibm8tY2FzZS1kZWNsYXJhdGlvbnMiOlsiZXJyb3IiXSwibm8tY2xhc3MtYXNzaWduIjpbImVycm9yIl0sIm5vLWNvbXBhcmUtbmVnLXplcm8iOlsiZXJyb3IiXSwibm8tY29uZC1hc3NpZ24iOlsiZXJyb3IiXSwibm8tY29uc3QtYXNzaWduIjpbImVycm9yIl0sIm5vLWNvbnN0YW50LWNvbmRpdGlvbiI6WyJlcnJvciJdLCJuby1jb250cm9sLXJlZ2V4IjpbImVycm9yIl0sIm5vLWRlYnVnZ2VyIjpbImVycm9yIl0sIm5vLWRlbGV0ZS12YXIiOlsiZXJyb3IiXSwibm8tZHVwZS1hcmdzIjpbImVycm9yIl0sIm5vLWR1cGUtY2xhc3MtbWVtYmVycyI6WyJlcnJvciJdLCJuby1kdXBlLWVsc2UtaWYiOlsiZXJyb3IiXSwibm8tZHVwZS1rZXlzIjpbImVycm9yIl0sIm5vLWR1cGxpY2F0ZS1jYXNlIjpbImVycm9yIl0sIm5vLWVtcHR5IjpbImVycm9yIl0sIm5vLWVtcHR5LWNoYXJhY3Rlci1jbGFzcyI6WyJlcnJvciJdLCJuby1lbXB0eS1wYXR0ZXJuIjpbImVycm9yIl0sIm5vLWV4LWFzc2lnbiI6WyJlcnJvciJdLCJuby1leHRyYS1ib29sZWFuLWNhc3QiOlsiZXJyb3IiXSwibm8tZXh0cmEtc2VtaSI6WyJlcnJvciJdLCJuby1mYWxsdGhyb3VnaCI6WyJlcnJvciJdLCJuby1mdW5jLWFzc2lnbiI6WyJlcnJvciJdLCJuby1nbG9iYWwtYXNzaWduIjpbImVycm9yIl0sIm5vLWltcG9ydC1hc3NpZ24iOlsiZXJyb3IiXSwibm8taW5uZXItZGVjbGFyYXRpb25zIjpbImVycm9yIl0sIm5vLWludmFsaWQtcmVnZXhwIjpbImVycm9yIl0sIm5vLWlycmVndWxhci13aGl0ZXNwYWNlIjpbImVycm9yIl0sIm5vLWxvc3Mtb2YtcHJlY2lzaW9uIjpbImVycm9yIl0sIm5vLW1pc2xlYWRpbmctY2hhcmFjdGVyLWNsYXNzIjpbImVycm9yIl0sIm5vLW1peGVkLXNwYWNlcy1hbmQtdGFicyI6WyJlcnJvciJdLCJuby1uZXctc3ltYm9sIjpbImVycm9yIl0sIm5vLW5vbm9jdGFsLWRlY2ltYWwtZXNjYXBlIjpbImVycm9yIl0sIm5vLW9iai1jYWxscyI6WyJlcnJvciJdLCJuby1vY3RhbCI6WyJlcnJvciJdLCJuby1wcm90b3R5cGUtYnVpbHRpbnMiOlsiZXJyb3IiXSwibm8tcmVkZWNsYXJlIjpbImVycm9yIl0sIm5vLXJlZ2V4LXNwYWNlcyI6WyJlcnJvciJdLCJuby1zZWxmLWFzc2lnbiI6WyJlcnJvciJdLCJuby1zZXR0ZXItcmV0dXJuIjpbImVycm9yIl0sIm5vLXNoYWRvdy1yZXN0cmljdGVkLW5hbWVzIjpbImVycm9yIl0sIm5vLXNwYXJzZS1hcnJheXMiOlsiZXJyb3IiXSwibm8tdGhpcy1iZWZvcmUtc3VwZXIiOlsiZXJyb3IiXSwibm8tdW5kZWYiOlsiZXJyb3IiXSwibm8tdW5leHBlY3RlZC1tdWx0aWxpbmUiOlsiZXJyb3IiXSwibm8tdW5yZWFjaGFibGUiOlsiZXJyb3IiXSwibm8tdW5zYWZlLWZpbmFsbHkiOlsiZXJyb3IiXSwibm8tdW5zYWZlLW5lZ2F0aW9uIjpbImVycm9yIl0sIm5vLXVuc2FmZS1vcHRpb25hbC1jaGFpbmluZyI6WyJlcnJvciJdLCJuby11bnVzZWQtbGFiZWxzIjpbImVycm9yIl0sIm5vLXVudXNlZC12YXJzIjpbImVycm9yIl0sIm5vLXVzZWxlc3MtYmFja3JlZmVyZW5jZSI6WyJlcnJvciJdLCJuby11c2VsZXNzLWNhdGNoIjpbImVycm9yIl0sIm5vLXVzZWxlc3MtZXNjYXBlIjpbImVycm9yIl0sIm5vLXdpdGgiOlsiZXJyb3IiXSwicmVxdWlyZS15aWVsZCI6WyJlcnJvciJdLCJ1c2UtaXNuYW4iOlsiZXJyb3IiXSwidmFsaWQtdHlwZW9mIjpbImVycm9yIl0sImFycmF5LWVsZW1lbnQtbmV3bGluZSI6WyJlcnJvciJdLCJvYmplY3QtcHJvcGVydHktbmV3bGluZSI6WyJlcnJvciJdfSwiZW52Ijp7ImVzNiI6dHJ1ZX19fQ==):

```javascript
# good
const a = {foo: {
  bar: "bat"
}}

# bad
const b = {a: "A", foo: {
  bar: "bat"
}}
```

bbatsov:
I'm still pondering about the best course of action here. Too many configuration options definitely seem like an overkill to me, so I'm wondering if this shouldn't be more or less a single toggle that all cops should respect. I'm not sure I want us to follow what ESLint is doing to the letter, given that the behavior so far has been different for years and it seems most people are fine with it. The indentation topics are harder than the naming topics! :-)

bbatsov:
@dvandersluis @pirj Any thoughts on this?

Korri:
> I'm wondering if this shouldn't be more or less a single toggle that all cops should respect.

😮 is that a pattern Rubocop uses? That does definitely sounds better then having all those options!

Korri:
The main downside with a global rule, is that people lose the flexibility to allow this only in certain constructs, I could see how some folks would prefer to allow this exclusively for method calls for example.
Having this allowed Hashes is probably the most controversial one (method definitions too maybe?).

I think a global rule would solve our use case though, so I'd be OK with that, we could also decide to make it more "opinionated" and not allow it for the more controversial constructs like Hashes and Arrays.

lib/rubocop/cop/layout/first_array_element_line_break.rb
```diff
@@ -6,17 +6,49 @@ module Layout
...
+      #     # good
+      #     [
+      #       :a,
+      #       :b]
+      #
+      #     # good
+      #     [
+      #       :a, {
+      #       :b => :c
+      #     }]
```
pirj, line 32: This looks awful. Is it really considered good by default now?

lib/rubocop/cop/layout/first_method_argument_line_break.rb
```diff
@@ -6,17 +6,63 @@ module Layout
...
       #
       #     # bad
       #     method(foo, bar,
       #       baz)
       #
+      #     # bad
+      #     method(foo, bar, {
+      #       baz: "a",
+      #       qux: "b",
+      #     })
```
pirj, line 19: What is a "good" counterpart for this?

spec/rubocop/cop/layout/first_method_parameter_line_break_spec.rb
```diff
@@ -83,4 +83,66 @@ def foo(
...
+      RUBY
+    end
+
+    it 'registers and corrects last parameter that starts on another line' do
+      expect_offense(<<~RUBY)
+        def foo(bar, baz,
+                ^^^ Add a line break before the first parameter of a multi-line method parameter list.
+          qux = false)
+          do_something
+        end
```
pirj, line 136: I may miss something with this example. It also offends `Layout/ParameterAlignment` which suggests changing to:
```ruby
def foo(bar, baz,
        qux = false)
  do_something
end
```

Does it make sense to use examples that don't trigger other cops?
The primary reason I'm asking is that I'm heavily triggered by this, too, and I feel an irresistible bias against this change in general as a result.

In any case,
```ruby
def foo(
bar, baz,
qux = false)
  do_something
end
```
or the "corrected" one below look horrible to my eye.

lib/rubocop/cop/layout/first_array_element_line_break.rb
Korri: Yeah this rule is only about having a line break before the first element in a multi line array, so it always considered this to be "good".

lib/rubocop/cop/layout/first_method_argument_line_break.rb
Korri: Good point will add it, it is:
```ruby
     # good
     method(
       foo, bar, {
       baz: "a",
       qux: "b",
     })
```

Korri: ✔️ done

spec/rubocop/cop/layout/first_method_parameter_line_break_spec.rb
Korri: > or the "corrected" one below look horrible to my eye.

🤔 yeah that rule is definitely not mean to be used in isolation. Not sure what the feedback is here though, this is the one test that is making sure that the existing behaviour is maintained for method definitions that are _actually_ considered multiline even with `AllowMultilineFinalElement` on. It looks a lot like the tests that already existed in this file.

dvandersluis: @Korri if you run the full correction suite on the original code (with your change) how does it end up looking?

In general, it's better to not have cops do other corrections, especially when it comes to Layout because it results in duplicating a lot of work and additionally causes situations that the correct raises `ClobberingErrors` (the same piece of code is corrected in incompatible ways).

Korri: > In general, it's better to not have cops do other corrections

What do you mean by "other corrections" the cop here is only adding a new line before the first parameter, it _is_ what it's main and only role is.

For reference, the example you said "looks horrible" differs from what the autocorrect actually does, it doesn't change the indentation of the `qux` keyword, so:
```ruby
# bad
def foo(bar, baz,
        qux = false)
  do_something
end

# get corrected to "good":
def foo(
bar, baz,
  qux = false)
  do_something
end
```

> @Korri if you run the full correction suite on the original code (with your change) how does it end up looking?

Exactly the same as it does today!

First `Layout/FirstMethodParameterLineBreak` is not on by default, and pretty weird to enable on it's own, as it's meant to be used at the very least in conjunction with `Layout/MultilineMethodParameterLineBreaks`, but let's give it a try:

```ruby
# Base example (made a few tweaks to satisfy other rules, but functionally the same)
def foo(bar, baz,
  qux: false)
  do_something(bar, baz, qux)
end

# Autocorrect with the default cops set
def foo(bar, baz,
        qux: false)
  do_something(bar, baz, qux)
end


# With `Layout/FirstMethodParameterLineBreak` enabled
def foo(
  bar, baz,
  qux: false
)
  do_something(bar, baz, qux)
end

# With `Layout/FirstMethodParameterLineBreak` and `Layout/MultilineMethodParameterLineBreaks` enabled:
def foo(
  bar,
  baz,
  qux: false
)
  do_something(bar, baz, qux)
end

# With `Layout/FirstMethodParameterLineBreak` and `Layout/MultilineMethodParameterLineBreaks`  and the `AllowMultilineFinalElement` true on both (no differences):
def foo(
  bar,
  baz,
  qux: false
)
  do_something(bar, baz, qux)
end

```

As the last point shows, there **is NO differences** in behaviour for that specific example, based on the changes in this Pull Request, I only added that unit test to make sure that the existing behaviour DID NOT CHANGE, if you look higher in the test file, you will see _basically the same exact test_.

The current pull request only affects method that have their last element multiline!

dvandersluis: Thanks for the examples. FYI it wasn't me who you were responding to before.

> What do you mean by "other corrections" the cop here is only adding a new line before the first parameter, it is what it's main and only role is.

Autocorrection for some cops tries to do layout reformatting as well. For instance, adding a new line indented to the correct level. This tends to cause issues in the long run, and is redundant because we have cops that are already present to reflow the layout when code is changed/added.

Korri: > Thanks for the examples. FYI it wasn't me who you were responding to before.

Hoo totally missed that, will edit to make it clearer!

> Autocorrection for some cops tries to do layout reformatting as well. For instance, a cop that adds a new line might try to indent it to the correct level. This tends to cause issues in the long run, and is redundant because we have cops that are already present to reflow the layout when code is changed/added.

That sounds like exactly what this rule is doing :) It's not trying to fix layout, it's focusing on its one job: Adding a new line.

Korri:
Hey folks. Would love some more input on this! I'm happy to work on a more opinionated/restricted PR too if that's how we prefer it! Or a global rule too!

bbatsov:
I'll try to make a decision about the next steps by the end of next week.

Korri:
@bbatsov so what's the decision on this? 😄

Korri:
@bbatsov 🥺 ?

bbatsov:
I spent much of September away from computers and I have a lot of catching up to do with my OSS project these days.

Korri:
@bbatsov any change we could get this moving? I'm happy to explore alternative approaches, like a global option, or a more limited, opinionated approach, just need your input on what direction to take from here!

bbatsov:
@Korri After much deliberation I'm leaning towards your original proposal - a separate configuration toggle for every relevant cop. And having this be disabled by default everywhere.

bbatsov:
> The main non-starter for us, that means Shopify is https://github.com/Shopify/ruby-style-guide/pull/387 is allowing single elements to be multi line:

Just to be clear - that's also controlled by the new configuration flags, right? 

I also think it'd be nice to add some section in the docs that explains how all those related cops have similar configuration and why people might want to have different settings in different LineBreak cops.

Korri:
> Just to be clear - that's also controlled by the new configuration flags, right?

yeah, the configuration allows it for the last element, which a single argument/element technically is, I think there might be a use case for folks that would want to allow it only for single element, but I did not address it here.

So to be clear, you are talking about merging this PR as-is? I can rebase it, and try to add the documentation you mentioned 👍

bbatsov:
> So to be clear, you are talking about merging this PR as-is? I can rebase it, and try to add the documentation you mentioned. 

Yep, I like the current version of the PR - I just want to make sure this is documented in enough details so people outside of Shopify would be able to make good use of it. I'm thinking we probably need to add some section of the manual about the common behavior in LineBreak cops and some of the reasoning for the new config.

Korri:
@bbatsov I took a stab at documenting this option inside it's own Layout cops section, let me know what you think!

docs/modules/ROOT/pages/cops_layout.adoc
```diff
@@ -7124,3 +7191,181 @@ RUBY
...
+  * FirstArrayElementLineBreak
+  * FirstHashElementLineBreak
+  * FirstMethodArgumentLineBreak
+  * FirstMethodParameterLineBreak
+  * MultilineArrayLineBreaks
+  * MultilineHashKeyLineBreaks
+  * MultilineMethodArgumentLineBreaks
+  * MultilineMethodParameterLineBreaks
+
+Those cops ignores their respective statements if all or the elements of
```
bbatsov, line 7209: ignores -> ignore

docs/modules/ROOT/pages/cops_layout.adoc
```diff
@@ -7124,3 +7191,181 @@ RUBY
...
+the statement are on the same line. If `AllowMultilineFinalElement` is
```
bbatsov, line 7210: I think that here "expression" is often a better term than "statement", so I'd probably use it.

docs/modules/ROOT/pages/cops_layout.adoc
```diff
@@ -7124,3 +7191,181 @@ RUBY
...
+because of their last element.
+
+Each cop can be configured independently allowing for more fine-grained
+control on what is considered ok in the codebase.
+
+=== Examples
+
+==== AllowMultilineFinalElement: true
+
+Those are examples of statements that would always get wrapped by their
```
bbatsov, line 7226: Here are some examples...

bbatsov:
@Korri The documentation looks good overall. I just left a few minor remarks.

Korri:
Addressed the comments 👍 One thing I'm not sure of, is the use of the `link:` the VSCode preview I used for the AsciiDoc doesn't handle links really other than making them clickable? Is there any way I can make sure that this will work on the live doc? I haven't seen those used that way anywhere else.

bbatsov:
The simplest thing to do to check that `link:` works is just to install Antora locally and build the docs yourself. Btw, how did you manage to add this to the cop doc given its auto-generated from the cop's source?

Drenmi:
> https://github.com/orgs/rubocop/teams/rubocop-core Any thoughts?

I have some concerns.

Firstly I find it pretty confusing to format code so that some arguments run horizontally and then some vertically in a sort of "inverted L" shape. Despite being marked as "seen often" I have never encountered this anywhere. (Can we run this on some popular code bases to see how common it actually is?)

Secondly I can't find a succinct way to explain to someone what this does, and doesn't, do without listing all the cases, which I take to be a sign that there might be a problem with the conceptual integrity of the option. However, the fact that this is relatively simple (in terms of lines of code) to implement, might suggest I'm wrong on this second note.

I guess my review is "not sure".

bbatsov:
@Drenmi For the record - I do share your concerns, that's part of the reason why progress here was pretty slow. It'd be nice to get some real-world usage metrics for this style, although at this point it's not a big deal for me. In the end I've opted to accept this proposal mostly because Shopify seems to need it and they have been a good supporter of the Ruby community. (and it doesn't seem to add that much complexity to our code)

I'm hoping that while working on the documentation we'll manage to crystallize better why someone would like to be arranging their code in this way.

Korri:
> The simplest thing to do to check that `link:` works is just to install Antora locally and build the docs yourself. Btw, how did you manage to add this to the cop doc given its auto-generated from the cop's source?

Well I didn't I added that manually, sounds like that's not the right approach, will dig into fixing this.

> @Drenmi For the record - I do share your concerns, that's part of the reason why progress here was pretty slow. It'd be nice to get some real-world usage metrics for this style, although at this point it's not a big deal for me. In the end I've opted to accept this proposal mostly because Shopify seems to need it and they have been a good supporter of the Ruby community. (and it doesn't seem to add that much complexity to our code)

To be fair, we don't have any kind or rules internally, I just noticed that we had a lot of current usage of that pattern, and implementation was the most straight forward. As I shared though, I'd be happy to go with a more opinionated approach if we have one. I do _not_ intent to turn that rule on for all the different cops though, the method deffiniton one for example seems like a bit of a stretch to me.

Let me do some digging, and share what we intent to actually use, maybe that could inform a more opinionated approach!

Korri:
After some digging here is the more opinionated that I will be pushing on our own rules. Let me know if that makes sense, and I can update the PR to follow this more opinionated approach, will also highlight the points where my opinions are not strong:

Here are examples from our codebase (tweaked to be more generic)
```ruby
# good (must have in my opinion)

# Last argument of a method call without parentheses

  # Rails controller before actin
  before_action :load_something, only [
    :show,
    :list,
    :do_some_update_logic,
    // ...
  ]

  # Rails model validation
  validate :name, presence: true, on: [:create, :activate], if: -> {
    active? && onboarded?
  }

  # same for model hooks, with some Sorbet binding
  after_commit :geolocate, unless: -> {
    T.bind(self, Address)
    archived? || invalid?
  }

# good (Nice to have in my opinion)

# last argument of a method call with parentheses

  # Logging
  write_log(:error,  {
     "scheduler" => job.class.name,
    "shop_id" => shop.id,
    "message" => "Something wrong happened here",
  })

  # Monorail logging
  Monorail.produce("some_monorail_event/1.0",  {
    shop_id: shop_id,
    reason: "the real reason",
    reason_code: reason_code,
    // ...
  })

# single argument method call

  # T.let from Sorbet, on a large hash
  SOME_CONSTANT = T.let({
    label: "value",
    label2: "value2",
    label3: "value3",
    label4: "value4",
    label5: "value5",
  }, T::Hash[Symbol, String])
  
  # Simple array of a single hash (I like this one as it makes it obvious that the array is single element)
  errors = [{
    error: "Something went wrong",
    error_code: error_code,
    // ...
  }]
  ```

I think that summarizes what we really need. I added all the rules here focusing on consistency, but I'd hate myself if we just merged those "because Shopify needs them" when what we need is really a subset of it.

I think the extreme opposite of this PR, would be a really opinionated set of rules that:
1. Allows last parameter to be multiline for `MultilineMethodArgumentLineBreaks` and `FirstMethodArgumentLineBreak` when there are no parentheses (by default, or as an `EnforcedStyle`)
2. Same as above, but for all method calls.
3. Add an `EnforcedStyle` to `Layout/FirstArrayElementLineBreak` that allows single value arrays to be multiline.
4. Alternatively, do the above, to only allow single argument method calls to be multiline without applying `MultilineMethodArgumentLineBreaks`

**Comment by Korri** [comment]
@bbatsov with that additional context, do you think a more opinionated approach is worth pursuing, or should I clean up the doc and ship this approach instead?

**Comment by bbatsov** [comment]
> @bbatsov with that additional context, do you think a more opinionated approach is worth pursuing, or should I clean up the doc and ship this approach instead?

I'm okay with the currently suggested approach, as I don't want to add too much complexity for this. We'll just need good examples in the docs (similar to what you've provided about the real-world use-cases), so that people can understand why such options exist in the first place.

**Comment by Korri** [comment]
@bbatsov updated the code with some of the real world examples, and tweaked `CopsDocumentationGenerator` to allow for the changes I had done manually before, confirmed that those work (including the `xref` links) generating the doc locally 👍

**Comment by Korri** [comment]
@bbatsov would love your 👁️  on this! Just rebased on `master`. Does someone with the right credentials need to run `CircleCI Pipeline` though? It keeps failing instantly.

**Comment by bbatsov** [comment]
Likely something went bad with the deploy key for CircleCI again... I've regened it just now, so you can try triggering a new build.

**Comment by Korri** [comment]
> Likely something went bad with the deploy key for CircleCI again... I've regened it just now, so you can try triggering a new build.

Still failing :(

**Comment by Korri** [comment]
@bbatsov not sure why this is still failing, but I think the PR is otherwise ready for a (hopefully) final review.

**Comment by Korri** [comment]
@bbatsov it's probably something related to my account and/or access right on Rubocop, as even an empty pull request (just an empty commit on top of `master`) fails for me: https://github.com/rubocop/rubocop/pull/11272

File: lib/rubocop/cops_documentation_generator.rb [hunk 7]
```diff
@@ -42,7 +42,7 @@ def cops_body(cop, description, examples_objects, safety_objects, pars) # ruboco
     content << "#{description}\n"
     content << safety_object(safety_objects) if safety_objects.any? { |s| !s.text.blank? }
     content << examples(examples_objects) if examples_objects.any?
-    content << configurations(pars)
+    content << configurations(cop.department, pars)
```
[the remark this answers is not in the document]
Korri: I see yeah, you mean [this commit right](https://github.com/rubocop/rubocop/commit/f2875b59523d16674512fb34e7cb3171040f1a3f), either way it has no impact here, as we always end up usng the department in the context of `department_to_basename(department)`

bbatsov:
Time to fly! 🚀

Issue #10813, opened 2022-07-13
False positive of `NonAtomicFileOperation` for `FileUtils.rm_r`


AlexWayfer:
From this discussion: https://github.com/rubocop/rubocop/issues/10760#issuecomment-1171707113

## Expected behavior

No offenses or no exceptions for `FileUtils.rm_r`.

## Actual behavior

If I have the code:

```ruby
FileUtils.rm_r gem_name if Dir.exist? gem_name
```

it causes offenses by `NonAtomicFileOperation` cop.

If I remove the condition — it causes the exception:

```ruby
     Failure/Error: FileUtils.rm_r gem_name
     
     Errno::ENOENT:
       No such file or directory @ apply2files - foo_bar
```

Even with `require 'fileutils'`.

## Steps to reproduce the problem

The code is above.

## RuboCop version

```
$ [bundle exec] rubocop -V
1.31.2 (using Parser 3.1.2.0, rubocop-ast 1.19.1, running on ruby 3.1.2 x86_64-linux)
  - rubocop-performance 1.14.2
  - rubocop-rspec 2.12.1
```

AlexWayfer:
@ydah thank you, it's a good change. But… `--force` option is not only for cases where files (or dirs) don't exist. It'll also delete all protected files, probably something else. It's not an equal I beleive.

ydah:
@AlexWayfer Yes, I also certainly don't think they are strictly equivalent. The reason for adding this cop is that these can cause problems that are difficult to reproduce, especially when frequent file operations are performed in parallel, as in test runs using `parallel_rspec`.  However, it might be nice to be able to options for cases allowed where atomic file deletion cannot be safely guaranteed.

PR #10818, opened 2022-07-14
[Fix rubocop#10813] Fix recursive deletion to suppression in `Lint/NonAtomicFileOperation`

ydah:
Fixes: rubocop#10813

This PR is suppress offense in case of
recursive deletion in `Lint/NonAtomicFileOperation`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_recursive_deletions_to_suppression_in_non_atomic_file_operation.md
+++ b/changelog/fix_recursive_deletions_to_suppression_in_non_atomic_file_operation.md
@@ -1,0 +1,1 @@
+* [#10813](https://github.com/rubocop/rubocop/issues/10813): Fix recursive deletion to suppression in `Lint/NonAtomicFileOperation`. ([@ydah][])
--- a/lib/rubocop/cop/lint/non_atomic_file_operation.rb
+++ b/lib/rubocop/cop/lint/non_atomic_file_operation.rb
@@ -51,12 +51,11 @@ module RuboCop
         MSG_CHANGE_FORCE_METHOD = 'Use atomic file operation method `FileUtils.%<method_name>s`.'
         MAKE_FORCE_METHODS = %i[makedirs mkdir_p mkpath].freeze
         MAKE_METHODS = %i[mkdir].freeze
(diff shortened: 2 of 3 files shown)


config/default.yml
```diff
@@ -1972,7 +1972,9 @@ Lint/NonAtomicFileOperation:
   Description: Checks for non-atomic file operations.
   Enabled: pending
   VersionAdded: '1.31'
+  VersionChanged: '<<next>>'
   SafeAutoCorrect: false
+  AllowUnsafeRecursiveRemove: true
```
bbatsov, line 1977: What's unsafe about it?

ydah: We were discussing it here: https://github.com/rubocop/rubocop/issues/10813#issuecomment-1184312234

`rm_r` with check file existence is expected to be rewritten to `rm_rf`. 
While this is certainly an atomic file operation and will not cause the problems that occur with frequent parallel file operations, on the other hand, there is a danger of recursively deleting protected files without warning, which we hope this added parameter will allow the user to adjust. What do you think?

koic: Hm... this cop is Lint, not Style. I doubt it makes sense to have such an option in Lint. I'm not sure, but always allowing is one option for `rm_r`.

ydah: > Hm... this cop is Lint, not Style. I doubt it makes sense to have such an option in Lint.

That is certainly a bit of a concern there. Surely it might be a good idea to always allow it.

ydah:
Thank you so much. I updated this PR.

config/default.yml
```diff
@@ -1972,6 +1972,7 @@ Lint/NonAtomicFileOperation:
   Description: Checks for non-atomic file operations.
   Enabled: pending
   VersionAdded: '1.31'
+  VersionChanged: '<<next>>'
```
koic, line 1975: I think this change is unnecessary.

```suggestion
```

ydah: Indeed it is. I removed that.

bbatsov (approved):

PR #10820, opened 2022-07-16
Add new `Style/EmptyHeredoc` cop

koic:
## Summary

Checks for using empty heredoc to reduce redundancy.

```ruby
# bad
<<~EOS
EOS

<<-EOS
EOS

<<EOS
EOS

# good
''

# bad
do_something(<<~EOS)
EOS

do_something(<<-EOS)
EOS

do_something(<<EOS)
EOS

# good
do_something('')
```

## Other Information

I noticed the following redundancy in the RuboCop test code:

```ruby
expect_correction(<<~RUBY)
RUBY
```

This cop solves it generically.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/empty_heredoc.rb
```diff
@@ -0,0 +1,59 @@
...
+      #
+      #   # good
+      #   do_something('')
+      #
+      class EmptyHeredoc < Base
+        include Heredoc
+        include RangeHelp
+        extend AutoCorrector
+
+        MSG = "Use `''` instead of heredoc."
```
bbatsov, line 41: I guess that will be tied to whether the user selected single or double quoted string literals in their config. You may cheap by saying "an empty string literal".

koic: Sure! I've updated!

natematykiewicz: Similarly, the autocorrector says `corrector.replace(node, "''")`. Though maybe the autocorrector will see the `''` and swap it for `""` because it just inserted an invalid rule. I'm not sure how the autocorrector works. Just, the code looks like it'll have the same problem as the `MSG` did.

PR #10829, opened 2022-07-24
Deprecate `IgnoredMethods` option in integrate to `AllowedMethods` and `AllowedPatterns` option

ydah:
This PR is change deprecate `IgnoredMethods` option in integrate to `AllowedMethods` and `AllowedPatterns` option.

In order to not break extensions, it adds aliases for the existing class, methods and configuration.
Also adds obsoletion warnings for configuration files.

Motivation for create this PR:  https://github.com/rubocop/rubocop/pull/10809#issuecomment-1182778154 and https://github.com/rubocop/rubocop/pull/10540#issuecomment-1103997757
Similarly, the case of eliminating `Ignore`: https://github.com/rubocop/rubocop/pull/10555

-----------------

* [x] Added tests.


lib/rubocop/cop/mixin/ignored_methods.rb
```diff
@@ -1,52 +0,0 @@
...
-      end
-
-      def ignored_method?(name)
-        ignored_methods.any? do |value|
-          case value
-          when Regexp
-            value.match? String(name)
-          else
-            value == String(name)
-          end
```
ydah, line 32: Because of the processing for regex here, we decided against simply replacing it with `AllowedMethods`.

koic:
What is the rationale for this change? It's good to have consistent name, but I'm wondering if it's good for users. I observe that such changes are often stressful for users.

ydah:
As I added in my comments on the PR, the motivation for creating this PR: https://github.com/rubocop/rubocop/pull/10809#issuecomment-1182778154 and https://github.com/rubocop/rubocop/pull/10540#issuecomment-1103997757.
Similarly case, remove `Ignore`: https://github.com/rubocop/rubocop/pull/10555

bbatsov:
> What is the rationale for this change? It's good to have consistent name, but I'm wondering if it's good for users. I observe that such changes are often stressful for users.

Inconsistent naming is also frustrating. :-) I welcome this change and I don't plan us to remove the old name any time soon, which should limit the frustration. 

In general I want us to settle on uniform naming for everything over the course of time, so people wouldn't be confused by different cops using different names for the same things.

pboling:
This is SUPER confusing.

The [documentation](https://www.rubydoc.info/gems/rubocop/1.22.1/RuboCop/Cop/IgnoredMethods) for `IgnoredMethods` was:

> This module encapsulates the ability to ignore certain methods when parsing

The [documentation](https://msp-greg.github.io/rubocop/RuboCop/Cop/AllowedMethods.html) for `AllowedMethods` is the opposite:

> This module encapsulates the ability to allow certain methods when parsing.

How are these interchangeable?

1. As documented, `AllowedMethods` is currently broken, as it does the opposite of what it says it should do.
2. As implemented `AlllowedMethods` is currently misleading, as it does the opposite of what its name implies.

Proof:  The settings are used interchangeably, in spite of the fact that they should be opposites:
https://github.com/rubocop/rubocop-rails/pull/902/files

I think this should be reverted.

ydah:
Both `IgnoredMethods` and `AllowedMethods` are used with the same meaning.
If these contain the name of a setting method, it should not be considered an offense.

### AllowedMethods

https://docs.rubocop.org/rubocop/cops_style.html#allowedmethods-default-2

```ruby
# bad - AllowedMethods: [] (default)
redirect('foo/%{bar_id}')
# good - AllowedMethods: [redirect] (default)
redirect('foo/%{bar_id}')
```

### IgnoredMethods

This is the same for IgnoredMethods.

https://docs.rubocop.org/rubocop/1.30/cops_style.html#ignoredmethods-redirect

```
# good
redirect('foo/%{bar_id}')
```

pboling:
The fact that they are the same is the problem.  They, quite literally, can't be the same, and thus it is a (very confusing) bug.

ydah:
Maybe I'm not quite getting the full picture of the issue yet.

> Proof: the settings are used interchangeably, in spite of the fact that they should be opposites:

These appear to be polar opposites, but from the time they first appeared, the settings had the same meaning. https://github.com/rubocop/rubocop/pull/8200 So we integrated them in this PR.

And I think the original issue https://github.com/rubocop/rubocop-rails/pull/902/files#r1086060912 that is occurring this time was copying the obsolete settings into the default.yml on the RuboCop extension side.

This is indeed confusing because those options are still in the default.yml of rubocop/rubocop, so I am wondering if the same fixes below will solve the problem.

https://github.com/rubocop/rubocop-rails/pull/927

pboling:
If they _ever_ had the same meaning, it can _only_ have been by mistake (a _huge_ mistake).  In English "Ignored" and "Allowed" have opposite meanings.  It is not possible that they could correctly be the same.  As I satated above, the documentation for them shows that they were intended, at some point, to have been opposites.

The [documentation](https://www.rubydoc.info/gems/rubocop/1.22.1/RuboCop/Cop/IgnoredMethods) for IgnoredMethods was:

> This module encapsulates the ability to **_ignore_** certain methods when parsing

The [documentation](https://msp-greg.github.io/rubocop/RuboCop/Cop/AllowedMethods.html) for AllowedMethods is the opposite:

> This module encapsulates the ability to **_allow_** certain methods when parsing.

ydah:
Perhaps, I think in terms of RuboCop's use of the words `Allowed` and `Ignored`, we can think of it as follows.

Ignored:
Code that is an offense, but it is ignored because it contains a method to be ignored., so it is ignored. (does not register an offense).

Allowed:
Code that is an offense, but it does not registered offense because it contains methods that are allowed , so it is allowed. (does not register an offense).

Also, the documentation appears to be confusing. `AllowedPattern` does the following

https://github.com/rubocop/rubocop/blob/7c9e611064b5cadc6071847dafe5681245d2f533/lib/rubocop/cop/mixin/allowed_pattern.rb#L5-L6

cc/@koic I think you extracted the AllowedMethods module, please point out if I am incorrect.

pboling:
I see.  So by reversing the context it makes sense.

In order to make it make sense, perhaps the documentation should more fully give that context, e.g.

> **IgnoredMethods**
> Code that contains an offense, but it is _ignored_ because it contains one of the specified methods to be ignored.

> **AllowedMethods**
> Code that contains an offense, but it is ignored because it contains one of the specified methods that are _allowed_ to contain an offense.

Sorry for the noise!  I found some code that had the pattern where both the new style and the old deprecated style were present, and it struck me odd, so I looked up the documentation, and that made it even more strange.

I makes sense now, but I expect people may continue to get confused unless more context if given in the documentation.

ydah:
@pboling Thank you for taking this up.
I have made a few changes to the documentation to add some explanations. If you find something incongruous, please let me know in the comments.

- https://github.com/rubocop/rubocop/pull/11502

bbatsov:
I'll take the blame for the current naming confusion. Originally I opted for the name "IgnoredSomething", because were ignoring offenses related to something, but as time passed I've started to think that "AllowedSomething" is a better term as aren't really ignoring offenses, but rather allowing exceptions. Older code used the ignored language, newer used the allowed language and at some point I've decided we should make this uniform. 

Even if the two names sound different, essentially they behave exactly the same. But I guess we failed to communicate this properly.

bbatsov:
P.S. Naming remains one of the hardest problems for me, even after 20 years on the job. :-)

koic:
Yeah, naming is really hard. It's one of the hardest in software development :-)
https://martinfowler.com/bliki/TwoHardThings.html

PR #10836, opened 2022-07-26
Support autocorrecton for `Style/CommentAnnotation` when note does not exist on the same line but exist on the next line

ydah:
This PR is support autocorrecton for `Style/CommentAnnotation` when note does not exist on the same line but exist on the next line

Even in cases where note does not exist on the same line, if there is a comment on the next line, it will be merged and automatically modified to the good case.

```ruby
# @example RequireColon: true (default)
# bad
# HACK:
# does not work

# good
# HACK: does not work

# @example RequireColon: false
# bad
# HACK
# does not work

# good
# HACK does not work
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/comment_annotation.rb
+++ b/lib/rubocop/cop/style/comment_annotation.rb
@@ -40,6 +40,13 @@ module RuboCop
       #   # good
       #   # OPTIMIZE: does not work
       #
+      #   # bad
+      #   # HACK:
+      #   # does not work
+      #
+      #   # good
+      #   # HACK: does not work
+      #
       # @example RequireColon: false
       #   # bad
       #   # TODO: make better
@@ -58,6 +65,14 @@ module RuboCop
       #
       #   # good
       #   # OPTIMIZE does not work
+      #
+      #   # bad
+      #   # HACK
+      #   # does not work
+      #
+      #   # good
+      #   # HACK does not work
+      #
       class CommentAnnotation < Base
         include RangeHelp
         extend AutoCorrector
@@ -71,20 +86,20 @@ module RuboCop
         MISSING_NOTE = 'Annotation comment, with keyword `%<keyword>s`, is missing a note.'
 
         def on_new_investigation
-          processed_source.comments.each_with_index do |comment, index|
-            next unless first_comment_line?(processed_source.comments, index) ||
-                        inline_comment?(comment)
+          comments = processed_source.comments
+          comments.each_with_index do |comment, index|
+            next unless first_comment_line?(comments, index) || inline_comment?(comment)
 
             annotation = AnnotationComment.new(comment, keywords)
             next unless annotation.annotation? && !annotation.correct?(colon: requires_colon?)
 
-            register_offense(annotation)
+            register_offense(annotation, comments, index)
           end
         end
 
         private
 
-        def register_offense(annotation)
+        def register_offense(annotation, comments, index)
           range = annotation_range(annotation)
           message = if annotation.note
                       requires_colon? ? MSG_COLON_STYLE : MSG_SPACE_STYLE
@@ -93,9 +108,11 @@ module RuboCop
                     end
 
           add_offense(range, message: format(message, keyword: annotation.keyword)) do |corrector|
-            next if annotation.note.nil?
+            if (has_note = annotation.note.nil?) && next_line_comment?(comments, index)
+              correct_offensess(corrector, range, annotation.keyword, comments[index + 1])
+            end
 
-            correct_offense(corrector, range, annotation.keyword)
+            correct_offense(corrector, range, annotation.keyword) unless has_note
           end
         end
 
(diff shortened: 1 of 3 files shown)


koic:
I'm curious about the following things.

1. I don't think the title "Support autocorrecton for `Style/CommentAnnotation`" is appropriate because `Style/CommentAnnotation` cop has auto-correction
2. `Style/CommentAnnotation` cop only mentions colon and space after annotation. So, this is a rule change and it's unclear why this is the right thing to do.

https://github.com/rubocop/ruby-style-guide#annotate-keywords

ydah:
> 1. I don't think the title "Support autocorrecton for Style/CommentAnnotation" is appropriate because Style/CommentAnnotation cop has autocorrection.

Indeed it is. Change this PR title and changelog.

lib/rubocop/cop/style/comment_annotation.rb
```diff
@@ -40,6 +40,13 @@ module Style
       #   # good
       #   # OPTIMIZE: does not work
       #
+      #   # bad
+      #   # HACK:
+      #   # does not work
+      #
+      #   # good
+      #   # HACK: does not work
```
koic, line 48: Hm, user may have intentionally wrote this, so this cop may need to respect user intent. There is also the possibility of avoiding `Layout/LineLength` offense for example. WDYT?

ydah:
2. `Style/CommentAnnotation` cop only mentions colon and space after annotation. So, this is a rule change and it's unclear why this is the right thing to do.

Yes. Indeed it is, but the behavior when there is a comment on the next line is originally offense.
The space is trailing because it is at the end of a line, and the space after annotation always be nothing.

Therefore, I was having trouble with the following comments not being autocorrected.

```ruby.
# NOTE:
# This is the text of the note
# :
````

**Comment by koic** [comment]
Perhaps I don't fully understand the use case. IMHO, it looks like an over-warning as this cop role. Anyway I don't have a strong opinion for this, so I'll leave it to @rubocop/rubocop-core.

File: lib/rubocop/cop/style/comment_annotation.rb (hunk 1 above)
Comment (ydah) [reply lib/rubocop/cop/style/comment_annotation.rb:48]: I mentioned in another comment https://github.com/rubocop/rubocop/pull/10836#issuecomment-1195043385 , it will always be offense and not autocorrection.

The following files
```ruby
# frozen_string_literal: true

# NOTE:
# This is the text of the note
# :
```

Run rubocop -A
```command
bundle exec rubocop -A test.rb               
Inspecting 1 file
C

Offenses:

test.rb:3:3: C: Style/CommentAnnotation: Annotation comment, with keyword NOTE, is missing a note.
# NOTE:
  ^^^^^

```

But, they are not autocorrected.

ydah: I found it very painful when there are many todo's to `.rubocop_todo.yml` and I wanted to deal with them all at once.
For example, I thought it might be a good idea to add another option that would toggle this behavior. WDYT?

ydah:
Either treat it as an offense as before and automatically correct it, or change it to a good case as an excessive offense in the first place, both of which are good outcomes for the use case I have.
The proposal was to match the existing behavior, but I think it would be acceptable to change to the other one.

dvandersluis:
I agree with @koic and I don't think we should be rewriting people's comments. It's also unclear to me from your tests what would happen in this case with `RequireColon: true`.

```ruby
# HACK
# something else
```

Also this slightly modified example from your tests would change comment semantics IMO:

```ruby
# HACK:
# - the API may not always return this
# - the response format may change

... becomes ...

# HACK - the API may not always return this
# - the response format may change
```

Ultimately we should leave it to developers to choose how to format their comments.

ydah:
I see. I certainly did not take that case into account.
I would like to change the approach a little and not make the space after the colon offense even if it is removed when followed by a comment as shown below. How about this?

The following cases change bad to good:
```ruby
# @example RequireColon: true (default)
# TODO:
# some comment

# @example RequireColon: false
# TODO
# some comment
```

However, this also moves away from the original style guide reference: https://rubystyle.guide/#annotate-keywords .
I will close this PR if it looks good as before.

lib/rubocop/cop/style/comment_annotation.rb
ydah: I was mistaken. As you pointed out, it is no good to always autocorrect, as it is an offense by `Layout/LineLength` in some cases. Thank you very much.

ydah:
I've reconsidered and certainly think this is an excessive change, so I think I'll close it.
Thanks for your time @koic @dvandersluis .

palexvs:
> I've reconsidered and certainly think this is an excessive change, so I think I'll close it.

:(

Sometimes, when I work on MVPs or refactoring I need multi string TODOs like:
```
# TODO:
# - fix ...
# - think about ...
# - refactor ....
```

ydah:
> > I've reconsidered and certainly think this is an excessive change, so I think I'll close it.
> 
> :(
> 
> Sometimes, when I work on MVPs or refactoring I need multi string TODOs like:
> ```
> # TODO:
> # - fix ...
> # - think about ...
> # - refactor ....
> ```

I didn't understand what your comment has to do with this PR. This PR simply adds auto-correction, and I don't think this PR solves your issue?

palexvs:
sorry, I misread it. I thought it adds multi lines TODOs support

Issue #2077, opened 2015-07-27
Can one use rubocop to correct ruby code in slim files?


AlexVPopov:
I want to replace all old-hash syntax occurances in my slim view files in a rails project. I tried running `rubocop --only HashSyntax --auto-correct`, but it didn't do any corrections in `*.html.slim` files. Is it possible to do this?

bbatsov:
See https://github.com/sds/slim-lint

AlexVPopov:
:+1:

Issue #1113, opened 2014-05-26
Support parsing erb files


ferrous26:
Right now when I try to run rubocop on an erb template it does not parse out the ruby code. It would be nice to run rubocop on every file that has ruby code in it.

gregblass:
I have to say that I was surprised that my view files were untouched after running this in my rails project. I wanted all my code converted, whether embedded in an html.erb, or just plain ruby in an .rb file. I was trying to update all of my old ruby hash syntax to the new syntax easily/automatically.

I don't even think it would be that hard to have it parse anything inside a <%= %>. I guess your response would be "then write it yourself!" Was there any other showstoppers that were keeping this out of the scope? Or are you guys just not Rails people, or just prefer to keep this project framework agnostic, which means skip .html.erb? (WAT IS YOUR REASONING?!...haha)

Is there anything else out there that could parse my .html.erb files and convert them to the new ruby hash syntax?

askl56:
@gregblass i have a perl script for this if you are still looking.

haihappen:
@askl56 please share :)

bxt:
There is a gem called [Ruumba](https://github.com/ericqweinstein/ruumba) which does that.

However, it just uses a regex to extract everything between `<%=`, `<%` and `%>`, copies the extracted code into temporary files, executes `rubocop` and then deletes those files again. This is suboptimal for error messages, automatic corrections and configurations, so I guess something better would be nice still.

pt-stripe:
We're looking at writing this.

 @bbatsov is this still out of scope of `rubocop` or would you accept a patch making rubocop depend on an `.erb` parser like https://rubygems.org/gems/erubis/versions/2.7.0.

gregblass:
Jeez I was a little aggressive two years ago, wasn't I? Apologies for that.

I just returned back here since prettier is all the rage these days and really wished I could have that type of thing work with Rails.

Cool to see that it's under consideration!

sshaw:
For those still looking there's [ERB Lint](https://github.com/Shopify/erb-lint) which integrates with [Better HTML](https://github.com/Shopify/better-html). 🌐

PR #10839, opened 2022-07-28
Add API for 3rd party template support

r7kamura:
## Background

As I discussed at the following thread, I would like to add 3rd party template language plugins support to RuboCop.

- https://github.com/rubocop/rubocop/discussions/10768

## Changes

To support template languages, I figured we need a new API to extract a set of Ruby code from one source code, so I added `RuboCop::Runner.ruby_extractors` in this pull request.

Ruby extractor is a callable object that takes a `RuboCop::ProcessedSource` and returns an `Array` of `Hash` (or returns nil for unrelated file) like this:

```ruby
ruby_extractor.call(processed_source)
```

```ruby
[
  {
    offset: 2,
    processed_source: #<RuboCop::ProcessedSource>
  },
  {
    offset: 10,
    processed_source: #<RuboCop::ProcessedSource>
  },
]
```

On the 3rd party plugin side, the following code would be written.

```ruby
# This is the newly added API.
RuboCop::Runner.ruby_extractors.unshift(
  ruby_extractor_for_some_template
)

# This is the same thing that other RuboCop extensions are doing.
RuboCop::ConfigLoader.instance_variable_set(
  :@default_configuration,
  default_config_for_some_template
)
```

## Examples

To demonstrate this change, I have prepared some 3rd party plugins:

- https://github.com/r7kamura/rubocop-erb
- https://github.com/r7kamura/rubocop-haml
- https://github.com/r7kamura/rubocop-slim
- https://github.com/r7kamura/rubocop-markdown

If you add these gems, `bundle exec rubocop` will automatically lint Ruby code in template files as well.

```
$ bundle exec rubocop
Inspecting 16 files
............WC..

Offenses:

ruby_offense_example.rb:1:1: W: Lint/Void: Literal "foo" used in void context.
"foo"
^^^^^
ruby_offense_example.rb:1:1: C: [Correctable] Style/FrozenStringLiteralComment: Missing frozen string literal comment.
"foo"
^
ruby_offense_example.rb:1:1: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
"foo"
^^^^^
ruby_offense_example.rb:2:1: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
"baar"
^^^^^^
spec/fixtures/dummy.slim:1:3: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
- "a"
  ^^^
spec/fixtures/dummy.slim:3:5: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
| #{"c"}
    ^^^
spec/fixtures/dummy.slim:5:8: C: [Correctable] Style/ZeroLengthPredicate: Use !empty? instead of size > 0.
- a if array.size > 0
       ^^^^^^^^^^^^^^
spec/fixtures/dummy.slim:6:3: C: [Correctable] Style/NegatedIf: Favor unless over if for negative conditions.
- a if !b
  ^^^^^^^
spec/fixtures/dummy.slim:8:6: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
- if "a"
     ^^^

16 files inspected, 9 offenses detected, 8 offenses autocorrectable
```

## Related issues

I know there are some related Issues on the template language support:

- https://github.com/rubocop/rubocop/issues/1113
- https://github.com/rubocop/rubocop/issues/2077

## Drawbacks of the current ruby extractors implementation

I think there are limitations in the design of this ruby extractors pattern.

When one ruby extractor extracts code from a file, subsequent ruby extractors are not executed. This is a good mechanism to prevent the default ruby extractor from accidentally processing unrelated template files, for example, but on the other hand, it prevents multiple ruby extractors from running.

For example, if you try to create a rubocop-yard_example gem that inspects the code in the `@example` of a YARD comment, this will not work because the system does not allow multiple ruby extractors to be applied to a single file. If you use this gem, you will not be able to inspect normal Ruby code parts.

A mechanism such as the Rack middleware stack would allow plug-ins to choose whether or not to continue processing, but it would also increase the complexity of the design.

-----------------

* [x] Added tests.

### DIFF
--- a/docs/modules/ROOT/pages/extensions.adoc
+++ b/docs/modules/ROOT/pages/extensions.adoc
@@ -200,3 +200,43 @@ For example, when you have defined `MyCustomFormatter` in
 ----
 $ rubocop --require ./path/to/my_custom_formatter --format MyCustomFormatter
 ----
+
+== Template support
+
+RuboCop has API for extensions to support templates such as ERB, Haml, Slim, etc.
+
+Normally, RuboCop extracts one Ruby code from one Ruby file, however there are multiple embedded Ruby codes in one template file. To solve this problem, RuboCop has a mechanism called `Rubocop::Runner.ruby_extractors`, to which any Ruby extractor can be added on the extension side.
+
+Ruby extractor must be a callable object that takes a `RuboCop::ProcessedSource` and returns an `Array` of `Hash` that contains Ruby source codes and their offsets from original source code, or returns `nil` for unrelated file.
+
+[source,ruby]
+---
+ruby_extractor.call(processed_source)
+---
+
+An example returned value from a Ruby extractor would be as follows:
+
+[source]
+---
+[
+  {
+    offset: 2,
+    processed_source: #<RuboCop::ProcessedSource>
+  },
+  {
+    offset: 10,
+    processed_source: #<RuboCop::ProcessedSource>
+  },
+]
+---
+
+On the extension side, the code would be something like this:
+
+[source,ruby]
+---
+RuboCop::Runner.ruby_extractors.unshift(ruby_extractor)
+---
+
+`RuboCop::Runners.ruby_extractors` is processed from the beginning and ends when one of them returns a non-nil value. By default, there is a Ruby extractor that returns the given Ruby source code with offset 0, so you can unshift any Ruby extractor before it.
+
+NOTE: This is still an experimental feature and may change in the future.
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -180,6 +180,10 @@ module RuboCop
 
         status, corrector = enabled_line?(range.line) ? correct(range, &block) : :disabled
 
+        # Since this range may be generated from Ruby code embedded in some
+        # template file, we convert it to location info in the original file.
+        range = range_for_original(range)
+
         current_offenses << Offense.new(severity, range, message, name, status, corrector)
       end
 
@@ -286,6 +290,21 @@ module RuboCop
       end
       # rubocop:enable Layout/ClassStructure
 
+      # Called before any investigation
+      # @api private
+      def begin_investigation(processed_source, offset: 0, original: processed_source)
+        @current_offenses = nil
+        @current_offense_locations = nil
+        @currently_disabled_lines = nil
+        @processed_source = processed_source
+        @current_corrector = nil
+
+        # We need to keep track of the original source and offset,
+        # because `processed_source` here may be an embedded code in it.
+        @current_offset = offset
+        @current_original = original
+      end
+
       private
 
       ### Reserved for Cop::Cop
@@ -318,15 +337,6 @@ module RuboCop
 
       private_class_method def self.restrict_on_send
         @restrict_on_send ||= self::RESTRICT_ON_SEND.to_a.freeze
-      end
-
-      # Called before any investigation
-      def begin_investigation(processed_source)
-        @current_offenses = nil
-        @current_offense_locations = nil
-        @currently_disabled_lines = nil
-        @processed_source = processed_source
-        @current_corrector = nil
       end
 
       EMPTY_OFFENSES = [].freeze
@@ -458,6 +468,14 @@ module RuboCop
                     "Valid severities are #{Severity::NAMES.join(', ')}."
           warn(Rainbow(message).red)
         end
+      end
+
+      def range_for_original(range)
+        ::Parser::Source::Range.new(
+          @current_original.buffer,
+          range.begin_pos + @current_offset,
+          range.end_pos + @current_offset
+        )
       end
     end
   end
--- a/lib/rubocop/cop/commissioner.rb
+++ b/lib/rubocop/cop/commissioner.rb
@@ -76,10 +76,10 @@ module RuboCop
       end
 
       # @return [InvestigationReport]
-      def investigate(processed_source)
+      def investigate(processed_source, offset: 0, original: processed_source)
         reset
 
-        @cops.each { |cop| cop.send :begin_investigation, processed_source }
+        begin_investigation(processed_source, offset: offset, original: original)
         if processed_source.valid_syntax?
           invoke(:on_new_investigation, @cops)
           invoke_with_argument(:investigate, @forces, processed_source)
@@ -94,6 +94,12 @@ module RuboCop
       end
 
       private
+
+      def begin_investigation(processed_source, offset:, original:)
+        @cops.each do |cop|
+          cop.begin_investigation(processed_source, offset: offset, original: original)
+        end
+      end
 
       def trigger_responding_cops(callback, node)
         @callbacks[callback]&.each do |cop|
--- a/changelog/new_add_api_for_3rd_party_template.md
+++ b/changelog/new_add_api_for_3rd_party_template.md
@@ -1,0 +1,1 @@
+* [#10839](https://github.com/rubocop/rubocop/pull/10839): Add API for 3rd party template support. ([@r7kamura][])
--- a/lib/rubocop/cop/team.rb
+++ b/lib/rubocop/cop/team.rb
@@ -77,7 +77,7 @@ module RuboCop
       end
 
       # @return [Commissioner::InvestigationReport]
-      def investigate(processed_source)
+      def investigate(processed_source, offset: 0, original: processed_source)
         be_ready
 
         # The autocorrection process may have to be repeated multiple times
@@ -87,14 +87,15 @@ module RuboCop
         on_duty = roundup_relevant_cops(processed_source.file_path)
 
         autocorrect_cops, other_cops = on_duty.partition(&:autocorrect?)
+        report = investigate_partial(autocorrect_cops, processed_source,
+                                     offset: offset, original: original)
 
-        report = investigate_partial(autocorrect_cops, processed_source)
-
-        unless autocorrect(processed_source, report)
+        unless autocorrect(processed_source, report, offset: offset, original: original)
           # If we corrected some errors, another round of inspection will be
           # done, and any other offenses will be caught then, so only need
           # to check other_cops if no correction was done
-          report = report.merge(investigate_partial(other_cops, processed_source))
+          report = report.merge(investigate_partial(other_cops, processed_source,
+                                                    offset: offset, original: original))
         end
 
         process_errors(processed_source.path, report.errors)
@@ -116,12 +117,12 @@ module RuboCop
 
       private
 
-      def autocorrect(processed_source, report)
+      def autocorrect(processed_source, report, original:, offset:)
         @updated_source_file = false
         return unless autocorrect?
         return if report.processed_source.parser_error
 
-        new_source = autocorrect_report(report)
+        new_source = autocorrect_report(report, original: original, offset: offset)
 
         return unless new_source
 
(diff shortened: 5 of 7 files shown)


r7kamura:
It looks like autocorrection is not working properly. I may need to make some more changes.

bbatsov:
Well done!

I like the simplicity of the proposed solution, despite the limitations you've mentioned. The extension API should be documented very well in the online docs and we should probably mark it as experimental for a while, so we can do breaking changes there if needed.

@rubocop/rubocop-core Any thoughts?

lib/rubocop/cop/base.rb
```diff
@@ -310,12 +311,15 @@ def currently_disabled_lines
...
       # Called before any investigation
-      def begin_investigation(processed_source)
+      def begin_investigation(processed_source, offset: 0,
+                              original: processed_source)
         @current_offenses = []
         @current_offense_locations = nil
         @currently_disabled_lines = nil
         @processed_source = processed_source
         @current_corrector = Corrector.new(@processed_source) if @processed_source.valid_syntax?
+        @current_offset = offset
```
bbatsov, line 321: I'd add some comments explaining the newly added attributes.

lib/rubocop/cop/base.rb
```diff
@@ -126,6 +126,7 @@ def add_offense(node_or_range, message: nil, severity: nil, &block)
 
         status, corrector = enabled_line?(range.line) ? correct(range, &block) : :disabled
 
+        range = range_for_original(range)
```
bbatsov, line 129: I think this warrants some comment.

lib/rubocop/cop/commissioner.rb
```diff
@@ -94,6 +94,12 @@ def investigate(processed_source)
 
       private
 
+      def begin_investigation(processed_source, offset:, original:)
+        @cops.each do |cop|
+          cop.send(:begin_investigation, processed_source, offset: offset, original: original)
```
koic, line 99: `send` is slow in general. Can you move the `begin_investigation` method from private and mark `# @api private`?

```suggestion
          cop.begin_investigation(processed_source, offset: offset, original: original)
```

This is the difference for reference.

```diff
diff --git a/lib/rubocop/cop/base.rb b/lib/rubocop/cop/base.rb
index 3dc069cee..fc51bc9ff 100644
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -287,6 +287,21 @@ module RuboCop
         end
       end

+      # Called before any investigation.
+      # @api private
+      def begin_investigation(processed_source, offset: 0, original: processed_source)
+        @current_offenses = []
+        @current_offense_locations = nil
+        @currently_disabled_lines = nil
+        @processed_source = processed_source
+        @current_corrector = Corrector.new(@processed_source) if @processed_source.valid_syntax?
+
+        # We need to keep track of the original source and offset,
+        # because `processed_source` here may be an embedded code in it.
+        @current_offset = offset
+        @current_original = original
+      end
+
       private

       ### Reserved for Cop::Cop
@@ -313,21 +328,6 @@ module RuboCop
         @restrict_on_send ||= self::RESTRICT_ON_SEND.to_a.freeze
       end

-      # Called before any investigation
-      def begin_investigation(processed_source, offset: 0,
-                              original: processed_source)
-        @current_offenses = []
-        @current_offense_locations = nil
-        @currently_disabled_lines = nil
-        @processed_source = processed_source
-        @current_corrector = Corrector.new(@processed_source) if @processed_source.valid_syntax?
-
-        # We need to keep track of the original source and offset,
-        # because `processed_source` here may be an embedded code in it.
-        @current_offset = offset
-        @current_original = original
-      end
-
       # Called to complete an investigation
       def complete_investigation
         InvestigationReport.new(self, processed_source, @current_offenses, @current_corrector)
diff --git a/lib/rubocop/cop/commissioner.rb b/lib/rubocop/cop/commissioner.rb
index 85820dea2..412db6f26 100644
--- a/lib/rubocop/cop/commissioner.rb
+++ b/lib/rubocop/cop/commissioner.rb
@@ -96,7 +96,7 @@ module RuboCop

       def begin_investigation(processed_source, offset:, original:)
         @cops.each do |cop|
-          cop.send(:begin_investigation, processed_source, offset: offset, original: original)
+          cop.begin_investigation(processed_source, offset: offset, original: original)
         end
       end
```

koic: This is required as part of this feature and not a change I want to see independently in maintenance. Can you squash your commits into one?

r7kamura:
OK, I squashed it 👍 

Maybe would it be easier to always squash and force push every time, even if we don't know if the review is done or still in progress? I am concerned that it will make the review more difficult because the reviewer will have no way of knowing what changes have been made to the raised points.

koic:
There may be differences between reviewers, I rarely see working commits. GitHub has a "Compare" button when I need it, and that's enough. And most of the time I re-review the entire PR.

docs/modules/ROOT/pages/extensions.adoc
```diff
@@ -196,3 +196,43 @@ For example, when you have defined `MyCustomFormatter` in
 ----
 $ rubocop --require ./path/to/my_custom_formatter --format MyCustomFormatter
 ----
+
+== Template language support
```
koic, line 200: I don't hear the term "Template language" very often. How about replacing it with the term "Template engine"?

r7kamura: Since mentioning the details may seem like a nuanced difference, what about "Template support"?

The term "engine" is a nuance that refers to the processor itself that converts the template and data model into artifacts, but in reality, this feature does not support any specific processor implementation, but supports some languages that embed Ruby. For example, there is a template language called eRuby, and there are processors such as erb, erubi, and erubis that use it. In this case, I think that this feature supports the eRuby language, but not specific processor implementations.

Examples of the term "template language" include Wikipedia's [Template processor](https://en.wikipedia.org/wiki/Template_processor) page and [Slim](https://github.com/slim-template/slim)'s description. However, it is true that we are more likely to come into contact with the term "template engine" because the language and processor are often combined, and we refer exclusively to the processor when writing programs.

bquorning:
Just wanted to ensure that you are familiar with the https://github.com/Shopify/erb-lint gem. You may want to consider opening an issue on that repository, or mention some of their maintainers to look at this PR.

r7kamura:
Thanks. I followed your advice and created the following Issue.

- https://github.com/Shopify/erb-lint/issues/274

Indeed, erb-lint provides the same functionality for eRuby, and rubocop-erb I created has almost the same implementation as erb-lint about its Rubocop Linter, although the details are a bit different. It uses the same better_html gem as a parser.

I know that there are other linters that use RuboCop, such as haml-lint and slim-lint. I would like to ask the author of this gem for his opinion if you would like. cc: @sbs

- https://github.com/sds/haml-lint
- https://github.com/sds/slim-lint

These gems did not seem to support autocorrect. Because the existing Haml and Slim parsers do not include location information in the AST, so it was not feasible as is.

In rubocop-haml and rubocop-slim I created dedicated Haml parser and Slim parser for this purpose. This allows us to retrieve the Ruby code with positional information from within the template and pass the appropriate values to RuboCop's template API.

- https://github.com/r7kamura/hamli
- https://github.com/r7kamura/slimi

docs/modules/ROOT/pages/extensions.adoc
koic: Developers probably think of "Engine" when referring to ERB, Haml, Slim. OTOH, I can understand your point of view. However, this heading term is included as an anchor in the URL, so I'm curious what the intuitive terminology is for users.

Earlopain:
> I know that there are other linters that use RuboCop, such as haml-lint and slim-lint. I would like to ask the author of this gem for his opinion if you would like. cc: `@sbs`

You seem to have pinged the wrong user, @sds is the correct handle.

r7kamura:
Wow, it's true. Thanks. And sbs, sorry about that.

bbatsov:
Any updates/progress here?

r7kamura:
The current agenda is as follows, and I don't think there are any major problems:

- What to call this function (template, template engine, template language, etc.)
    - I prefer to call it simply "template", though I think this is a trivial issue.
- Any opinions from the authors of the related linters? (erb-lint, haml-lint, slim-lint, etc.)
    - [A question was raised](https://github.com/Shopify/erb-lint/issues/274#issuecomment-1258853478) whether RuboCop's caching mechanism works for templates.
        - I think it works, but I am not very familiar with this mechanism and have not been able to verify it rigorously
    - No other opinions expressed for now.

bbatsov:
> I prefer to call it simply "template", though I think this is a trivial issue.

Fine by me.

> Any opinions from the authors of the related linters? (erb-lint, haml-lint, slim-lint, etc.) 

Might be good to open a ticket about this on each repo, as it'd be more likely that anyone would see this. Still, I'd like us to agree on some limit for collecting feedback as we can't wait for it forever. (e.g. by the end of November)

bbatsov:
@r7kamura Any updates here? Did you get any feedback from 3rd party gem authors?

r7kamura:
Sorry for my late response 🙇 

I solicited opinions by the end of November and received two feedback which are discussed below.

- https://github.com/sds/haml-lint/issues/379
- https://github.com/sds/slim-lint/issues/142
- https://github.com/Shopify/erb-lint/issues/274

### 1. Can template files also benefit from file caching?

I believe that template files can benefit in the same way as Ruby files.

### 2. Can successive Ruby lines be combined and applied to RuboCop?

This means that it would be nice to do the following autocorrection:

```haml
-# bad
- if foo
  - if bar
    - do_something

-# good
- if foo && bar
  - do_something
```

I have given up on this, as it seems quite difficult to implement...

There is a pull request on haml-lint that tries to challenge this.

- https://github.com/sds/haml-lint/pull/346

bbatsov:
@r7kamura So, essentially the PR is ready to be merged at this? (I guess we can flag the API as experimental, so we can make breaking changes to it if needed)

r7kamura:
Yes, I think it's ready to be merged. I will resolve the conflict and fix the minor wording issue about "template language" or "template".

bbatsov:
Thanks!

r7kamura:
I tested with [rubocop-slim](https://github.com/r7kamura/rubocop-slim) using rubocop in the master branch and successfully inspected the Ruby code in a Slim template 👍 

```
$ bundle exec rubocop spec/fixtures/dummy.slim 
Inspecting 1 file
C

Offenses:

spec/fixtures/dummy.slim:1:3: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
- "a"
  ^^^
spec/fixtures/dummy.slim:3:5: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
| #{"c"}
    ^^^
spec/fixtures/dummy.slim:5:8: C: [Correctable] Style/ZeroLengthPredicate: Use !empty? instead of size > 0.
- a if array.size > 0
       ^^^^^^^^^^^^^^
spec/fixtures/dummy.slim:6:3: C: [Correctable] Style/NegatedIf: Favor unless over if for negative conditions.
- a if !b
  ^^^^^^^
spec/fixtures/dummy.slim:8:6: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
- if "a"
     ^^^

1 file inspected, 5 offenses detected, 5 offenses autocorrectable
```

I will try to test it on a larger code base.

bbatsov:
Nice!

PR #10845, opened 2022-07-29
Support Bundler-like namespaced feature on require config

r7kamura:
As you probably know, when you write `gem 'foo-bar'` in Gemfile, `Bundler.require` will tries to `require 'foo/bar'` as well. 

```ruby
# Gemfile
gem 'foo-bar'
```

This behavior is implemented here in bundler gem:

- https://github.com/rubygems/rubygems/blob/446cc57a7ccdf1924deb291be9571219e7ba8523/bundler/lib/bundler/runtime.rb#L59-L73

It would be nice if RuboCop works in the same way when we set `require` config in .rubocop.yml:

```yaml
# .rubocop.yml
require:
  - foo-bar
```

Running `bundle gem foo-bar` will generate lib/foo/bar.rb, so we need to call `require 'foo/bar'` to use foo-bar gem. However, I think users will want to write `- foo-bar` as above.

We can avoid this problem by preparing lib/foo-bar.rb on the foo-bar gem side, but RuboCop will register an offense with `Naming/FileName` cop for such file name, so I think we should avoid it.

The change in this pull request would eliminate the need for files with non-standard name, such as

- [lib/rubocop-rspec.rb](https://github.com/rubocop/rubocop-rspec/blob/c5f3643dcf5dc2b3c63ac80d2d282fccd182aafc/lib/rubocop-rspec.rb)
- [lib/rubocop-rails.rb](https://github.com/rubocop/rubocop-rails/blob/8c40def7b0faefc975ad12a070d0e198f7b20960/lib/rubocop-rails.rb)
- etc.

and additional configuration to specifically allow such file names:

- https://github.com/rubocop/rubocop-rspec/blob/c5f3643dcf5dc2b3c63ac80d2d282fccd182aafc/.rubocop.yml#L54-L56
- https://github.com/rubocop/rubocop-rails/blob/8c40def7b0faefc975ad12a070d0e198f7b20960/.rubocop.yml#L117-L119

-----------------

    * No related open issue was found.
* [x] Added tests.


r7kamura:
I think this is a controversial change myself.  WDYT?

This is because unlike the Bundler case, we have explicitly used the vocabulary "require" for this config. This change means that there are two patterns of targets for which `require` is performed. This is not Simple, but Easy for users.
I decided to propose this change because I thought the advantages outweighed the disadvantages.

bbatsov:
I'm open to the proposal, but I think we'll have to document this behavior clearly in the configuration documentation. @jonas054 Any thoughts on this?

lib/rubocop/feature_loader.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+module RuboCop
+  # Load feature for "require" config.
+  # @api private
+  class FeatureLoader
```
bbatsov, line 6: This class can probably use more documentation itself.

r7kamura:
Sorry for the slow response.
I added some more documentation to the class doc comment and the documentation page. I was going to add it to the Configuration page, but it seemed to mention `require` in Extensions page, so I decided to add it there.

bbatsov:
Looks good. Thanks!

PR #10850, opened 2022-08-01
Fix `Style/ClassEqualityComparison` autocorrection within module

r7kamura:
In the situation where `Module.nesting` is not empty, `bar.class.name == 'Bar'` and `bar.instance_of?(Bar)` yield different results.

This is a simple example of the problematic case:
(In reality, it is more complicated because constant search is done by tracing back to ancestor inheritance tree)

```ruby
module Foo
  class << self
    def bar1?(value)
      value.instance_of?(Bar)
    end

    def bar2?(value)
      value.class.name == 'Bar'
    end
  end

  class Bar
  end
end

class Bar
end

p Foo.bar1?(Bar.new) #=> false
p Foo.bar1?(Foo::Bar.new) #=> true

p Foo.bar2?(Bar.new) #=> true
p Foo.bar2?(Foo::Bar.new) #=> false
```

So I changed it to prepend `::` in those situations.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_fix_styleclassequalitycomparison.md
+++ b/changelog/fix_fix_styleclassequalitycomparison.md
@@ -1,0 +1,1 @@
+* [#10850](https://github.com/rubocop/rubocop/pull/10850): Fix `Style/ClassEqualityComparison` autocorrection within module. ([@r7kamura][])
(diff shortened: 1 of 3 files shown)


koic:
Good catch! Thanks!

PR #10851, opened 2022-08-01
Clarify the API stability wording

jeffwidman:
Previously the wording implied that between major versions the API won't change.
But it's pretty clear from the surrounding wording and general semvar concept
that the API _will_ change between major versions... it's only between minor/patch
versions that it won't change.


bbatsov:
Ouch! Thanks for spotting and fixing this!

PR #10856, opened 2022-08-03
Add `DisableCopForEntireFileInSourceCode` cop

sambostock:
Enforces that disabling cops for an entire file happens in a config file, rather than by wrapping the entire file in `rubocop:disable` and `rubocop:enable` comments.
```ruby
# bad
# rubocop:disable Department/CopName
entire_file
# rubocop:enabled Department/CopName
```
```ruby
# bad
# rubocop:todo Department/CopName
entire_file
# rubocop:enabled Department/CopName
```
```ruby
# good
# Department/CopName disabled in config, such as .rubocop.yml or .rubocop_todo.yml
entire_file
```
```ruby
# good
code
# rubocop:disable Department/CopName
subset_of_code
# rubocop:enabled Department/CopName
more_code
```

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/redundant_cop_disable_directive.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
-# The Lint/RedundantCopDisableDirective cop needs to be disabled so as
-# to be able to provide a (bad) example of a redundant disable.
-# rubocop:disable Lint/RedundantCopDisableDirective
 module RuboCop
   module Cop
+    # The Lint/RedundantCopDisableDirective cop needs to be disabled so as
+    # to be able to provide a (bad) example of a redundant disable.
+    # rubocop:disable Lint/RedundantCopDisableDirective
```
sambostock, line 7: Need to confirm that moving these types of comments doesn't result in accidentally generating unintentional documentation for the module defined below it (here `Lint`).

lib/rubocop/cop/lint/redundant_cop_disable_directive.rb
```diff
@@ -26,6 +26,7 @@ module Lint
       #   # good
       #   x += 1
       class RedundantCopDisableDirective < Base # rubocop:todo Metrics/ClassLength
+        # rubocop:enable Lint/RedundantCopDisableDirective
```
sambostock, line 29: Ideally, we'd be able to have this comment appear in same block as above, but we'd need some way to strip it out, and I'm not sure if that exists.

lib/rubocop/cop/style/disable_cop_for_entire_file_in_source_code.rb
```diff
@@ -0,0 +1,82 @@
...
+      #   code
+      #   # rubocop:disable Department/CopName
+      #   subset_of_code
+      #   # rubocop:enable Department/CopName
+      #   more_code
+      class DisableCopForEntireFileInSourceCode < Base
+        # rubocop:enable Lint/RedundantCopDisableDirective
+        MSG = 'When disabling cops for an entire file, prefer excluding the file in `.rubocop.yml`.'
+        TODO_MSG = 'When marking cops as TODO for an entire file, ' \
+                   'prefer excluding the file in `.rubocop_todo.yml`.'
```
sambostock, line 36: I should update these messages to also suggest approaches in descending order of preference:
1. Disable the cop(s) per line
  _See the change made to `lib/rubocop/cop/style/block_delimiters.rb`._
3. Disable the cop(s) for only the affected range
  _See the change made to `lib/rubocop/cop/lint/redundant_cop_enable_directive.rb`._
5. Disable the cop in a config file (ideally only if to remove friction from expecting to add future offenses)
  _I considered this for `spec/rubocop/cop/lint/script_permission_spec.rb`, but ended up using in-line disabling, since there were only two occurrences._

lib/rubocop/cop/style/disable_cop_for_entire_file_in_source_code.rb
```diff
@@ -0,0 +1,82 @@
...
+            next unless disabled_range.cover?(code_range)
+
+            on_offending_comment(range_disabling_comment(disabled_range))
+          end
+        end
+
+        private
+
+        def on_offending_comment(comment)
+          return add_offense(comment, message: TODO_MSG) if comment.text.include?('rubocop:todo')
```
sambostock, line 51: Not sure if there's some constant I should refer to, or helper method, rather than hard code it. Perhaps something in `CommentConfig`? 🤔

spec/rubocop/cop/style/disable_cop_for_entire_file_in_source_code_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+
+  it 'registers no offenses when disabling all cops' do
+    # Because `rubocop:disable all` disables all cops, it disables this one too.
+    # Ideally, it would probably still apply, but then this cop would need to be a special case.
+    expect_no_offenses(<<~RUBY)
+      # rubocop:disable all
+      false
+      # rubocop:enable all
+    RUBY
+  end
```
sambostock, line 123: I think changing this would require either making this cop a special case, or changing block directives to not apply to the lines that control them, which would necessitate a change to `rubocop-ast`. I'm not sure of the implication of making such a change.

config/default.yml
```diff
@@ -3443,6 +3443,14 @@ Style/Dir:
   Enabled: true
   VersionAdded: '0.50'
 
+Style/DisableCopForEntireFileInSourceCode:
```
sambostock, line 3446: I'm not sure about this name. Open to suggestions.

changelog/new_add_disable_cop_for_entire_file_in_source_code.md
```diff
@@ -0,0 +1 @@
+* [#10856](https://github.com/rubocop/rubocop/pull/10856): Add DisableCopForEntireFileInSourceCode. ([@sambostock][])
```
dvandersluis, line 1: Pending cop name changes, it should contain the department and be surround by code marks in the changelog.

```suggestion
* [#10856](https://github.com/rubocop/rubocop/pull/10856): Add `Style/DisableCopForEntireFileInSourceCode`. ([@sambostock][])
```

dvandersluis:
I think `Lint` probably makes more sense here than `Style`, so that it stays with `RedundantCopDisableDirective` and `RedundantCopEnableDirective`.

lib/rubocop/cop/lint/redundant_cop_disable_directive.rb
dvandersluis: I don't believe it should, I'm pretty sure that only comments directly on the class are used. You could run `rake update_cops_documentation` to check (but make sure you don't add adoc changes to this PR please).

dvandersluis: Alternately you could disable your cop for this file in `.rubocop.yml`.

lib/rubocop/cop/style/block_delimiters.rb
```diff
@@ -165,7 +164,7 @@ module Style
       #     process(something)
       #   }
       #
-      class BlockDelimiters < Base
+      class BlockDelimiters < Base # rubocop:disable Metrics/ClassLength
```
dvandersluis, line 167: Based on the rationale for the cop here, shouldn't this comment not be present and an exception added to `.rubocop.yml` instead? If this cop is a good idea for users, then it should be for RuboCop itself too.

lib/rubocop/cop/style/disable_cop_for_entire_file_in_source_code.rb
dvandersluis: How would you determine which option to show, or are you suggesting showing all of them? If so, it'd probably be better served in the documentation for the cop and keeping the message more terse.

lib/rubocop/cop/style/disable_cop_for_entire_file_in_source_code.rb
dvandersluis: You could probably add a `todo?` method to `DirectiveComment` and then call `DirectiveComment.new(comment).todo?`

spec/rubocop/cop/style/disable_cop_for_entire_file_in_source_code_spec.rb
dvandersluis: This is an interesting edge case. I think we should just leave it because if you want to disable all cops, that shouldn't have special cased cops that are not included.

config/default.yml
dvandersluis: `Lint/CopDisabledForEntireFile` is probably simpler IMO.

dvandersluis:
Thanks @sambostock! I think this is a good idea in general, but left some comments about things I'd like us to change her. Particularly, we should use the cop on RuboCop if we're going to add it :smile:

config/default.yml
sambostock: Yes, I like that better. Thanks!

lib/rubocop/cop/style/block_delimiters.rb
sambostock: The purpose of this cop is not to recommend against disabling cops in source code entirely (that's covered by [`Style/DisableCopsWithinSourceCodeDirective`](https://docs.rubocop.org/rubocop/cops_style.html#styledisablecopswithinsourcecodedirective)). It is only meant to flag the use of `rubocop:disable` around the entire file.

In this case, addressing the cop could happen in three ways:
- Move the config to `.rubocop.yml`
  - This is okay, but it means the cop is effectively still disabled across the entire file
- Move the `rubocop:disable`/`rubocop:enable` to just be around the offending class
  - This is better, but it means the cop is still disabled for any nested classes that might be defined
- Target this specific class when disabling, so the cop still applies to any inner classes
  - This is ideal, because it conveys exactly what we want: "We're okay with the length of the `BlockDelimiters`."

lib/rubocop/cop/style/disable_cop_for_entire_file_in_source_code.rb
sambostock: Hmm... Good point. I guess my main thing is that I think authors should first ask themselves "can I be more specific when disabling this cop?" before just moving it to the config file.

sambostock (commented):
My gut feeling actually was to put this cop in `Lint`, but I put it in `Style` because I saw `Style/DisableCopsWithinSourceCodeDirective`.

On that topic, I'm wondering if perhaps it makes more sense for this cop to become a supported style in `Style/DisableCopsWithinSourceCodeDirective` instead.

Currently, that cop only supports the following:
- Forbidding all `rubocop:disable` comments
- Being disabled

We could add a style to say:
- Allow `rubocop:disable` comments, only if they target a subset of the file.

What do you think?

config/default.yml
koic: `CopDisabledForEntireFile` is a good name 👍  But this seems to favor `Style` over `Lint`. `Lint` is the highest warning level, emulating Ruby warnings, preventing errors and deprecated APIs (e.g. that will be removed in the future) . For example, specifying `Exclude` in .rubocop.yml is practically equivalent but not as bad as `Lint` would detect it. It will depend on the usage preference of user (`Style`) .

lib/rubocop/cop/style/disable_cop_for_entire_file_in_source_code.rb
```diff
@@ -0,0 +1,82 @@
...
+      # Enforces that disabling cops for an entire file happens in a config
+      # file, rather than by wrapping the entire file in `rubocop:disable` and
+      # `rubocop:enable` comments.
+      #
+      # @example
+      #   # bad
+      #   # rubocop:disable Department/CopName
+      #   entire_file
+      #   # rubocop:enable Department/CopName
+
```
koic, line 16: Can you padding with `#`? Because missing the `#` breaks the documentation.

config/default.yml
dvandersluis: Honestly, there probably should be a separate `Rubocop`, etc. department for the cops that deal with rubocop directives, because `Style` is about code style to me, not about all preferences, and `Lint` is not currently limited to ruby warnings, plus has other directive cops as mentioned earlier.

I think it makes sense to be grouped with the related cops, but it doesn't really matter at the end of the day.

lib/rubocop/cop/style/block_delimiters.rb
dvandersluis: I understand, and I'm not necessarily suggesting that its detection should change at all. But since almost every ruby file contains a single class, the following are practically the same:

```ruby
# rubocop:disable Foo/Bar
class X
  # ...
end
# rubocop:enable Foo/Bar
```

```ruby
class X
# rubocop:disable Foo/Bar
  # ...
# rubocop:enable Foo/Bar
end
```

The latter won't register the cop though obviously. So in the instances where you have replaced the former with the latter, IMO it'd be better to just put the disable into .rubocop.yml because otherwise it's just shuffling code to satisfy a cop. Does that make sense?

lib/rubocop/cop/style/disable_cop_for_entire_file_in_source_code.rb
dvandersluis: Perhaps something along these lines...

```ruby
MSG = 'Prefer using directives on smaller sections of code, or if you need to disable the entire file, do it in your configuration file.'
```

config/default.yml
koic: Ah, I see. But the fact that `Style` and `Lint` have different warning levels is a point for users. This is a tough one...

lib/rubocop/cop/style/block_delimiters.rb
sambostock: I agree given your example, but note that the code actual code switches to an inline directive, not a block one,

```diff
-# rubocop:disable Foo/Bar
-class X
+class X # rubocop:disable Foo/Bar
   # ...
 end
-# rubocop:enable Foo/Bar
```

which applies to only **that** class and **not to any nested classes**,

```ruby
class ExceptionallyLongClass # rubocop:disable Metrics/ClassLength
  class MustBeShortClass
  end
end
```

whereas the block comment or config comment would apply to the entire range or file.

```ruby
# rubocop:disable Metrics/ClassLength
class ExceptionallyLongClass
  class CanAlsoBeLongClass
  end
end
# rubocop:enable Metrics/ClassLength
```

config/default.yml
sambostock: 🤔 I see there is `InternalAffairs` for Cops that apply to Rubocop itself, although this is a cop that applies to Rubocop usage by consumers. Is there a name that would keep the "cop" theme? 😄 Maybe `Dispatch` or `Patrol` or something?

Either that, or we should decide on the criteria for if a cop is `Style` or `Lint` (if that isn't already documented already), and apply that to all the Rubocop cops.

Yet another option would be to follow the pattern of extracting a gem (`rubocop-rubocop`?), although this may be a special case as some of the cops catch actual mistakes (e.g. forgetting to re-enable a cop).

lib/rubocop/cop/style/disable_cop_for_entire_file_in_source_code.rb
sambostock: I implemented this, but later switched to your suggested message, which doesn't require specialization. However, I'll see if it turns out to be useful anyway, and add it in another PR.

bbatsov:
> Allow rubocop:disable comments, only if they target a subset of the file.

@sambostock Sorry for the radio silence for months! I do like the above suggestion, as the functionality of the new cop seems pretty related to the one of the existing one. 

And I have to say that I love your suggestion in general - those directives were definitely meant for more granular use. Probably it worth to re-iterate this in the docs as well.

bbatsov:
Btw, now we can do the opposite as well (https://docs.rubocop.org/rubocop/1.38/configuration.html#temporarily-enabling-cops-in-source-code), so perhaps we should account for it.

sambostock:
Oh yes, agreed, thanks for the heads up on that new feature. If we encourage developers to exclude a file from a rule in the config rather than disable comments around the entire file, then similarly we should encourage developers to enable a rule for a specific file from config, rather than wrap the entire file in enable comments.

bbatsov:
Yeah, exactly.

bbatsov:
@sambostock can you rebase your work on top of `master`?

Issue #10858, opened 2022-08-03
Auto-correction for Style/IdenticalConditionalBranches cop breaks behavior


DmitryBarskov:
Description: suggested auto-correction for `Style/IdenticalConditionalBranches` offence breaks an `if` statement if it is the last statement in the method with no explicit `return` keyword.

--------

## Expected behavior

Rubocop does not suggest auto-correction for `Style/IdenticalConditionalBranches` rule.

## Actual behavior

Rubocop suggests an auto-correction which breaks the last `if` statement in a method.

Running with options `rubocop test.rb --debug --cache false`
```
For $PWD: Default configuration from $HOME/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-1.32.0/config/default.yml
Inspecting 1 file
Scanning $PWD/test.rb
For $PWD: configuration from $PWD/.rubocop.yml
configuration from $HOME/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-rails-2.13.2/config/default.yml
configuration from $HOME/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-rails-2.13.2/config/default.yml
configuration from $HOME/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-rspec-2.10.0/config/default.yml
configuration from $HOME/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-rspec-2.10.0/config/default.yml
configuration from $HOME/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-performance-1.13.2/config/default.yml
configuration from $HOME/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-performance-1.13.2/config/default.yml
configuration from $HOME/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-thread_safety-0.4.4/config/default.yml
configuration from #{path}
C

Offenses:

xpense-tracker-api/test.rb:12:7: C: [Correctable] Style/IdenticalConditionalBranches: Move statement_a out of the conditional.
      statement_a
      ^^^^^^^^^^^
xpense-tracker-api/test.rb:14:7: C: [Correctable] Style/IdenticalConditionalBranches: Move statement_a out of the conditional.
      statement_a
      ^^^^^^^^^^^

[...]

PR #10862, opened 2022-08-03
Add autocorrection to `Lint/EmptyConditionalBody`

dvandersluis:
I noticed in https://github.com/rubocop/rubocop/issues/10858#issuecomment-1204175590 that `Lint/EmptyConditionalBody` does not have autocorrection, so I added it. It does the following:

* Removes the empty branch, including any comments if `AllowComments` is false.
* Handles subsequent branches, including changing `elsif` to `if` and `else` to `if`/`unless` appropriately, if the topmost branch is the one being removed.
* Can handle the `if` statement being preceded by another node.

I believe I've covered the edge cases here but if there's one you can think of let me know and I'll add it in. I also was unsure if we should mark this as `SafeAutoCorrect: false` but am happy to do so.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/empty_conditional_body_spec.rb
+++ b/spec/rubocop/cop/lint/empty_conditional_body_spec.rb
@@ -9,6 +9,8 @@ RSpec.describe RuboCop::Cop::Lint::EmptyConditionalBody, :config do
       ^^^^^^^^^^^^ Avoid `if` branches without a body.
       end
     RUBY
+
+    expect_correction('')
   end
 
   it 'does not register an offense for missing `if` body with a comment' do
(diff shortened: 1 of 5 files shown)


spec/rubocop/config_validator_spec.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::ConfigValidator, :isolated_environment do
+  include FileHelper
+
+  # before do
+  #   allow(default_config).to receive(:internal?).and_return(true)
+  # end
```
koic, line 8: Did you forget to remove this?

dvandersluis: Yes sorry this was something else I was working on and shouldn't have been checked in 🤦‍♂️

dvandersluis: Removed now, thanks @koic!

spec/rubocop/cop/lint/empty_conditional_body_spec.rb
```diff
@@ -27,6 +29,64 @@
...
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      if other_condition
+        do_something
+      elsif another_condition
+        do_something_else
+      end
+    RUBY
```
BrianHawley, line 89: @dvandersluis this correction doesn't have the same semantics. To get the same meaning, it should be:
```ruby
if !condition && other_condition
  do_something
elsif !condition && another_condition
  do_something_else
end
```
I don't know if this autocorrection was considered safe before, but it definitely isn't safe or correct now.

spec/rubocop/cop/lint/empty_conditional_body_spec.rb
```diff
@@ -90,6 +168,22 @@
...
+      elsif bar
+        5
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      x = if bar
+        5
+      end
+    RUBY
```
BrianHawley, line 184: Same, this should be:
```ruby
x = if !foo && bar
  5
end
```

PR #10868, opened 2022-08-04
[Fixes #9227] Corrected the misleading example for `Style/GuardClause`

### LINKED ISSUE #9227: Misleading example for Style/GuardClause
The example in the documentation for `Style/GuardClause` is misleading because mentioned bad and good version don#t match execution.

It says

```
# bad
if something
  foo || raise('exception')
else
  ok
end

# good
foo || raise('exception') if something
ok
```
However both versions behave differently and _bad_ can't be replaced by _good_.

Imagine `something` being `true` and `foo` returning a truthy value.
Bad (original) behavior: `foo` gets called.
Good (suggested) behavior: `foo` gets called and `ok` gets called.

I suggest to remove this example because I don't see a new case that would work which hasn't been mentioned before.

v1nayv:
Corrected the misleading example for Stlye/GuardClause reported in #9227
Added the return statement to be more correct with the bad example.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_add_return_statement_to_guard_clause.md
+++ b/changelog/fix_add_return_statement_to_guard_clause.md
@@ -1,0 +1,1 @@
+* [#9227](https://github.com/rubocop/rubocop/issues/9227): Add return statement to guard clause example to make it more correct. ([@v1nayv][])
--- a/docs/modules/ROOT/pages/cops_style.adoc
+++ b/docs/modules/ROOT/pages/cops_style.adoc
@@ -1637,7 +1637,7 @@ items.include?
 
 | PreferredMethods
 | `{"collect"=>"map", "collect!"=>"map!", "inject"=>"reduce", "detect"=>"find", "find_all"=>"select", "member?"=>"include?"}`
-| 
+|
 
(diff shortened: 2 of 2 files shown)


dvandersluis:
If you want to make documentation changes, you need to change the comment in the cop instead of the adoc file directly. The adoc files are automatically generated on release.

v1nayv:
Thanks @dvandersluis , i was not aware of it. I will change it.

bbatsov:
@v1nayv ping :-)

KessaPassa:
I think you also have to change generated example in warning message.
https://github.com/rubocop/rubocop/blob/554fd197b28dc8823faca607bc46d3dd2ad9ec10/lib/rubocop/cop/style/guard_clause.rb#L139-L144

But you should not add 'return' for this case.
`raise 'exception' if something`

This code looks need to be updated.
https://github.com/rubocop/rubocop/blob/554fd197b28dc8823faca607bc46d3dd2ad9ec10/lib/rubocop/cop/style/guard_clause.rb#L152-L156

bbatsov:

PR #10873, opened 2022-08-05
[Fix #10857] Add `AllowedPatterns` to `Style/NumericLiterals`.

### LINKED ISSUE #10857: Add list of exception patterns to Style/NumericLiterals

In a current project, I need to handle identification numbers, which follow the format `<2 digit classification>` `<2 digit region id>` `<4 digit running number>` `<1 digit checksum>` ("[Institutionskennzeichen](https://de.wikipedia.org/wiki/Institutionskennzeichen)", "IK number"), for example `112233334`

I'd like to format them in code as `11_22_3333_4`, but the Style/NumericLiterals cop wants me to format them as `112_233_334`, which is nonsensical.

I also don't want to lose the warnings for ordinary numbers, where grouping the digits in triplets is desirable.

- `Strict: false` (default) takes offense with `11_22_3333_4`
- `AllowedNumbers: [...]` does not scale
- `Enabled: false` is my current workaround

## Additional context

It would be nice to define (a list of) allowed patterns, for example:

```yaml
Style/NumericLiterals:
  AllowedPatterns:     # list of exceptions, like AllowedNumbers
  - xx_xx_xxxx_x       # either using placeholders
  - \d\d_\d\d_\d{4}_\d # or using regexp
```

I'm not sure if there's precedent for either placeholders or regular expressions in other cops. Maybe there are better alternatives to declare such exceptions.

dvandersluis:
Allows additional allowed patterns to be specified for `Style/NumericLiterals` so that special patterns will not register offenses. The patterns are forced to be anchored (so `\d{4}_\d{4}` will only match exactly that, and not substrings), and this does not affect autocorrection, which will continue to add underscores between groups of 3 digits.

Fixes #10857.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -4405,6 +4405,7 @@ Style/NumericLiterals:
   Strict: false
   # You can specify allowed numbers. (e.g. port number 3000, 8080, and etc)
   AllowedNumbers: []
+  AllowedPatterns: []
 
 Style/NumericPredicate:
   Description: >-
--- a/changelog/change_add_allowedpatterns_to.md
+++ b/changelog/change_add_allowedpatterns_to.md
@@ -1,0 +1,1 @@
+* [#10857](https://github.com/rubocop/rubocop/issues/10857): Add `AllowedPatterns` to `Style/NumericLiterals`. ([@dvandersluis][])
--- a/lib/rubocop/cop/style/numeric_literals.rb
+++ b/lib/rubocop/cop/style/numeric_literals.rb
@@ -3,8 +3,16 @@
 module RuboCop
   module Cop
     module Style
-      # Checks for big numeric literals without _ between groups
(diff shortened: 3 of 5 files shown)


config/default.yml
```diff
@@ -4400,11 +4400,12 @@ Style/NumericLiterals:
   StyleGuide: '#underscores-in-numerics'
   Enabled: true
   VersionAdded: '0.9'
-  VersionChanged: '0.48'
+  VersionChanged: '<<next>>'
```
koic, line 4403: It looks like update is unneeded. Because adding this `AllowedPatterns` configuration does not change the default behavior.

```suggestion
  VersionChanged: '0.48'
```

dvandersluis: Sure, updated.

koic:
Thanks!

PR #10874, opened 2022-08-05
[Fix #10648] Allow `Style/TernaryParentheses` to take priority over `Style/RedundantParentheses` for requiring parentheses

### LINKED ISSUE #10648: Allow `Style/TernaryParentheses` to override `Style/RedundantParentheses` when parentheses are required

`Style/TernaryParentheses` has infinite loop protection to prevent it from endlessly undoing corrections for `Style/RedundantParentheses` when both cops are enabled. This makes sense, but it has an unfortunate side effect: if `Style/TernaryParentheses` is set to require parentheses (via `EnforcedStyle: require_parentheses` or `require_parentheses_when_complex`), that preference is overridden and no offenses are flagged on ternary expressions that lack parentheses. To illustrate:

```ruby
# rubocop.yml
Style/RedundantParentheses:
  Enabled: true

Style/TernaryParentheses:
  Enabled: true
  EnforcedStyle: require_parentheses_when_complex

# foo.rb
foo = bar.empty? || baz.empty? ? "a" : "b" # no Style/TernaryParentheses offense is registered
```

I figured this out by reading the source for both cops - I don't think it's documented anywhere.

I propose moving the infinite loop protection to `Style/RedundantParentheses`: have it skip flagging offenses on ternary expressions when `Style/TernaryParentheses` is set to require parentheses. That way, rubocop can properly enforce the configurations for both cops.

The only alternative I'm aware of is to disable `Style/RedundantParentheses`, but it's a good cop to have and I'd prefer to have it enabled.

## Additional context

I've been a rubocop user for years and very much appreciate the work the maintainers do. I'd be glad to put a PR together if you folks are onboard with this idea but don't have the bandwidth.

dvandersluis:
`Style/TernaryParentheses` allows RuboCop to require ternary conditions are wrapped in parentheses, but this behaviour doesn't take effect when `Style/RedundantParentheses` is also enabled, in order to prevent an infinite loop.

However, it makes more sense for `TernaryParentheses` to take priority here and have `RedudantParentheses` allow the parentheses instead of creating an infinite autocorrection loop.

This change moves the infinite loop protection into `RedundantParentheses` instead.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_allow_styleternaryparentheses_to_take.md
+++ b/changelog/change_allow_styleternaryparentheses_to_take.md
@@ -1,0 +1,1 @@
+* [#10648](https://github.com/rubocop/rubocop/issues/10648): Allow `Style/TernaryParentheses` to take priority over `Style/RedundantParentheses` when parentheses are enforced. ([@dvandersluis][])
--- a/lib/rubocop/cop/style/redundant_parentheses.rb
+++ b/lib/rubocop/cop/style/redundant_parentheses.rb
@@ -58,7 +58,8 @@ module RuboCop
           allowed_ancestor?(node) ||
             allowed_method_call?(node) ||
             allowed_array_or_hash_element?(node) ||
-            allowed_multiple_expression?(node)
+            allowed_multiple_expression?(node) ||
+            allowed_ternary?(node)
         end
 
         def allowed_ancestor?(node)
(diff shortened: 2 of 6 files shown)


bbatsov:
Looks good to me. Thanks!

PR #10876, opened 2022-08-06
Change an option from `AllowedMethods` to `ForbiddenMethods` for `Lint/RedundantSafeNavigation` cop

ydah:
The `AllowedMethods` indicates the method names that you want to allow in the bad case.
However, for `Lint/RedundantSafeNavigation`, it indicates the method names that should be considered as bad cases.

e.g.
- https://docs.rubocop.org/rubocop/cops_style.html#allowedmethods-lambda-proc-it-default
- https://docs.rubocop.org/rubocop/cops_style.html#allowedmethods-nonzero-default

Since the usage is different, we propose to change the option names.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_change_an_option_from_allowedmethods_to.md
+++ b/changelog/change_change_an_option_from_allowedmethods_to.md
@@ -1,0 +1,1 @@
+* [#10876](https://github.com/rubocop/rubocop/pull/10876): Change an option from `AllowedMethods` to `ForbiddenMethods` for `Lint/RedundantSafeNavigation` cop. ([@ydah][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2097,13 +2097,15 @@ Lint/RedundantSafeNavigation:
   Description: 'Checks for redundant safe navigation calls.'
   Enabled: true
   VersionAdded: '0.93'
-  AllowedMethods:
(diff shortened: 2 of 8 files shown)


dvandersluis:
I'm not really sure `ForbiddenMethods` makes sense here, because when one of these methods is used, it's the _operator_ that needs to change, not the method.

I'm not sold that this change is necessary, but if we do want to change it, it's really just a list of methods that are available on `nil` and thus don't need the safe-nav operator. Perhaps `DetectedMethods`. What do you think @bbatsov?

bbatsov:
Yeah, that's definitely not the typical forbidden methods logic and I think the allowed methods name here makes more sense as those are the methods where you're allowed to skip safe navigation for, as they `nil` safe. Probably not the best naming indeed, but perhaps extending the documentation would be enough.

ydah:
Aah! Now that you explained it, I understand.

When I was writing the https://github.com/rubocop/rubocop/pull/10807 .
This is the only place where the behavior looked different, so I felt it was strange, but I guess I was mistaken.

Thank you so much @dvandersluis @bbatsov . I'll close this PR.

bbatsov:
We can still improve the documentation, though. If you got confused, we can assume other people got/will get confused as well.

PR #10879, opened 2022-08-07
Add ArrayAlignment to list of recommended Layout cops for LineLength

jdufresne:
Both ArgumentAlignment and HashAlignment are listed so it looks like
ArrayAlignment should also be included.

### DIFF
--- a/lib/rubocop/cop/layout/line_length.rb
+++ b/lib/rubocop/cop/layout/line_length.rb
@@ -22,6 +22,7 @@ module RuboCop
       # (Many of these are enabled by default.)
       #
       # * ArgumentAlignment
+      # * ArrayAlignment
       # * BlockAlignment
       # * BlockDelimiters
       # * BlockEndNewline


koic:
Can you update https://github.com/rubocop/rubocop/blob/v1.33.0/lib/rubocop/cop/layout/line_length.rb#L8-L61 instead of updating docs/modules/ROOT/pages/cops_layout.adoc directly. docs/modules/ROOT/pages/cops_layout.adoc is auto-generated, so don't change it.

koic:
@jdufresne ping.

jdufresne:
> Can you update https://github.com/rubocop/rubocop/blob/v1.33.0/lib/rubocop/cop/layout/line_length.rb#L8-L61 instead of updating docs/modules/ROOT/pages/cops_layout.adoc directly. docs/modules/ROOT/pages/cops_layout.adoc is auto-generated, so don't change it.

Thanks for the review. I've made this change in the latest revision.

koic:
Thanks!

PR #3078, opened 2016-04-24
[Fix #2874] Accept parentheses in hash and array

lumeet:
Before submitting a PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [x] Used the same coding conventions as the rest of the project.
* [x] Feature branch is up-to-date with `master` (if not - rebase it)
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Added an entry to the [Changelog](CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](../CONTRIBUTING.md#changelog-entry-format).
* [x] All tests are passing.
* [x] The new code doesn't generate RuboCop offenses.
* [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.

[1]: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

`Style/RedundantParentheses` doesn't register an offense for
parentheses in array and hash literals when the removal would cause
invalid syntax. For example, the following is accepted:

    array = [(
             <<-EOF
               something
             EOF
             ),
             'something']

Issue #2874, opened 2016-02-22
heredoc as hash value raises Style/RedundantParentheses


SaltwaterC:
I'll let the command line do the talking:

```bash
$ rubocop -V # supplied by ChefDK 0.11.0
0.37.2 (using Parser 2.3.0.5, running on ruby 2.1.6 x86_64-darwin12.0)
```

```bash
$ ruby code.rb
{:foo=>"heredoc\nmore heredoc\n", :bar=>"baz"}
```

```bash
$ cat code.rb
foo = {
  foo: (
  <<-EOF.gsub(/^\s*\|/, '')
    |heredoc
    |more heredoc
  EOF
  ),
  bar: 'baz'
}

p foo
```

```bash
$ rubocop
Inspecting 1 file
C

Offenses:

code.rb:2:8: C: Don't use parentheses around a method call.
  foo: (
       ^

1 file inspected, 1 offense detected
```

```bash
$ rubocop -a
Inspecting 1 file
E

Offenses:

code.rb:2:8: C: [Corrected] Don't use parentheses around a method call.
  foo: (
       ^
code.rb:7:3: E: unexpected token tCOMMA
(Using Ruby 2.0 parser; configure using TargetRubyVersion parameter, under AllCops)
  ,
  ^

1 file inspected, 2 offenses detected, 1 offense corrected
```

```bash
$ ruby code.rb
code.rb:7: syntax error, unexpected ',', expecting '}'
code.rb:9: syntax error, unexpected '}', expecting end-of-input
```

```bash
$ cat code.rb
foo = {
  foo:
  <<-EOF.gsub(/^\s*\|/, '')
    |heredoc
    |more heredoc
  EOF
  ,
  bar: 'baz'
}

p foo
```

Don't know whether this issue is related to #2654 and/or #2845. This is the closest I got before raising a new issue.

alexdowad:
This doesn't specifically have anything to do with heredocs.

```
$ cat test.rb
hash = {
  foo: 1
  ,
  bar: 2
}
$ ruby-parse test.rb
test4.rb:3:3: error: unexpected token tCOMMA
test4.rb:3:   ,
test4.rb:3:   ^
```

alexdowad:
From `parser`'s Ruby grammar:

```
          assocs: assoc
                    {
                      result = [ val[0] ]
                    }
                | assocs tCOMMA assoc
                    {
                      result = val[0] << val[2]
                    }
```

alexdowad:
[...]

PR #10885, opened 2022-08-08
[Fix #10883] Fix `Style/RedundantParentheses` to be able to detect offenses and properly correct when the end parentheses and comma are on their own line

### LINKED ISSUE #10883: Style/RedundantParentheses multiline list detection problem
Consider

```ruby
[
  (
    1
  ),
  2
]
```

`Style/RedundantParentheses` doesn't detect a problem in this example. If we take out the second list element, the cop triggers:

```ruby
[
  (
    1
  )
]
```

```
[Correctable] Style/RedundantParentheses: Don't use parentheses around a literal
```

## Expected behavior

Both examples should trigger `Style/RedundantParentheses`

## RuboCop version

```
1.33.0 (using Parser 3.1.2.0, rubocop-ast 1.19.1, running on ruby 2.6.9 x86_64-darwin21)
  - rubocop-minitest 0.21.0
  - rubocop-performance 1.14.3
  - rubocop-rails 2.15.2
```

dvandersluis:
Previously in #2874 / #3078, `Style/RedundantParentheses` was updated to ignore parentheses in arrays when the comma is on a separate line with the end parenthesis, as that would leave a comma on a line by itself, which is a syntax error.

Additionally, this trailing comma problem shows up in method argument lists, hashes, etc.

This change removes the false negative for this situation, and instead updates `ParenthesesCorrector` to be able to handle commas that are orphaned onto their own line after removing parentheses. Note that this corrector is also used by `Style/ParenthesesAroundCondition` and `Style/RescueModifier` but all tests are passing without change.

Fixes #10883.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/correctors/parentheses_corrector.rb
+++ b/lib/rubocop/cop/correctors/parentheses_corrector.rb
@@ -5,9 +5,13 @@ module RuboCop
     # This autocorrects parentheses
     class ParenthesesCorrector
       class << self
+        include RangeHelp
+
         def correct(corrector, node)
           corrector.remove(node.loc.begin)
           corrector.remove(node.loc.end)
+          handle_orphaned_comma(corrector, node)
+
           return unless ternary_condition?(node) && next_char_is_question_mark?(node)
 
           corrector.insert_after(node.loc.end, ' ')
@@ -21,6 +25,30 @@ module RuboCop
 
         def next_char_is_question_mark?(node)
           node.loc.last_column == node.parent.loc.question.column
+        end
+
+        def only_closing_paren_before_comma?(node)
+          source_buffer = node.source_range.source_buffer
+          line_range = source_buffer.line_range(node.loc.end.line)
+
+          line_range.source.start_with?(/\s*\)\s*,/)
+        end
+
+        # If removing parentheses leaves a comma on its own line, remove all the whitespace
+        # preceding it to prevent a syntax error.
+        def handle_orphaned_comma(corrector, node)
+          return unless only_closing_paren_before_comma?(node)
+
+          range = range_with_surrounding_space(
+            range: node.loc.end,
+            buffer: node.source_range.source_buffer,
+            side: :left,
+            newlines: true,
+            whitespace: true,
+            continuations: true
+          )
+
+          corrector.remove(range)
         end
       end
     end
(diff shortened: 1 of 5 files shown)


config/default.yml
```diff
@@ -4701,6 +4701,7 @@ Style/RedundantParentheses:
   Description: "Checks for parentheses that seem not to serve any purpose."
   Enabled: true
   VersionAdded: '0.36'
+  VersionChanged: '<<next>>'
```
koic, line 4704: This seems like a bugfix for the cop purposes, so it's unnecessary.

```suggestion
```

lib/rubocop/cop/correctors/parentheses_corrector.rb
```diff
@@ -22,6 +26,30 @@ def ternary_condition?(node)
         def next_char_is_question_mark?(node)
           node.loc.last_column == node.parent.loc.question.column
         end
+
+        def only_closing_paren_before_comma?(node)
+          source_buffer = node.source_range.source_buffer
+          line_range = source_buffer.line_range(node.loc.end.line)
+
+          /^\s*\)\s*,/.match?(line_range.source)
```
koic, line 34: Can you use `start_with?`?

```suggestion
          line_range.source.start_with?(/\s*\)\s*,/)
```

config/default.yml
dvandersluis: The behavior of the cop is changing away from something that was hardcoded and tested (which will mean code that did not register an offense for 6 years will now). If you want me to still remove this I will but that was my rationale 😄

lib/rubocop/cop/correctors/parentheses_corrector.rb
dvandersluis: Updated!

config/default.yml
koic: I see, but it looks that `VersionChanged` is unnecessary as it is what end user wants it to be :-)

dvandersluis: Removed it.

koic:
Awesome! Thanks!

justisb:
FYI this update causes invalid syntax for a heredoc string used in an array or hash, as the autocorrect adds the comma to the end of the heredoc end marker, breaking the marker. In that scenario we'll want to leave the dangling parentheses + comma.

Example:
```
    [
      (
        <<-'HEREDOC'
          Foo
        HEREDOC
      ),
      (
        <<-'HEREDOC'
          Bar
        HEREDOC
      ),
    ]
```

Is improperly autocorrected to:
```
    [
      
        <<-'HEREDOC'
          Foo
        HEREDOC,
      
        <<-'HEREDOC'
          Bar
        HEREDOC,
    ]
```

PR #10886, opened 2022-08-08
[Fix #10881] Fix `Style/SoleNestedConditional` to properly wrap `block` and `csend` nodes when necessary

### LINKED ISSUE #10881: [Style/SoleNestedConditional] Syntax error when autocorrecting nested if conditions with a block
If there are nested `if` statesments, where the second `if` condition is a function with an arugment and a block, autocorrecting causes a syntax error. Given snippet `test.rb`

```ruby
def foo
  if cond
    if validate :strict do
        "hello"
      end
    end
  end
end
```

and running 

```
rubocop test.rb --force-default-config --only Style/SoleNestedConditional -a
```

there is a syntax error reported by `ruby` (see below). Prior to autocorrect, there was no syntax error.

--------

## Expected behavior

`ruby test.rb` do not report syntax error after autocorrect. Proposed solution is to add parenthesis around arguments.

```ruby
def foo
  if cond && validate(:strict) do
        "hello"
      end
    end
end
```

## Actual behavior

The source code is autocorrected to

```ruby
def foo
  if cond && validate :strict do
        puts "hello"
      end
    end
end
```

`ruby test.rb` reports syntax error:

```
/tmp/test.rb:2: syntax error, unexpected symbol literal, expecting `do' or '{' or '('
  if cond && validate :strict do
/tmp/test.rb:5: syntax error, unexpected `end', expecting end-of-input
    end
```

## RuboCop version

```
$ [bundle exec] rubocop -V
1.33.0 (using Parser 3.1.2.0, rubocop-ast 1.19.1, running on ruby 2.7.4 x86_64-linux)
```

dvandersluis:
`Style/SoleNestedConditional` adds parentheses around `send` nodes that have unparenthesized arguments, but this wasn't being done for `csend` or `block` nodes, so this change fixes that.

Fixes #10881.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/sole_nested_conditional_spec.rb
+++ b/spec/rubocop/cop/style/sole_nested_conditional_spec.rb
@@ -663,6 +663,46 @@ RSpec.describe RuboCop::Cop::Style::SoleNestedConditional, :config do
             end
         RUBY
       end
+
+      context 'with a `csend` node' do
+        it 'registers an offense and corrects' do
+          expect_offense(<<~RUBY)
+            if foo
+              if bar&.baz quux
+              ^^ Consider merging nested conditions into outer `if` conditions.
+                do_something
+              end
+            end
+          RUBY
+
+          expect_correction(<<~RUBY)
+            if foo && (bar&.baz quux)
+                do_something
+              end
+          RUBY
+        end
+      end
+
+      context 'with a block' do
+        it 'registers an offense and corrects' do
+          expect_offense(<<~RUBY)
+            if foo
+              if ok? bar do
+              ^^ Consider merging nested conditions into outer `if` conditions.
+                  do_something
+                end
+              end
+            end
+          RUBY
+
+          expect_correction(<<~RUBY)
+            if foo && (ok? bar do
+                  do_something
+                end)
+              end
+          RUBY
+        end
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/sole_nested_conditional_spec.rb
```diff
@@ -663,6 +663,46 @@ def foo
...
+            if foo
+              if bar&.baz quux
+              ^^ Consider merging nested conditions into outer `if` conditions.
+                do_something
+              end
+            end
+          RUBY
+
+          expect_correction(<<~RUBY)
+            if foo && (bar&.baz quux)
```
koic, line 679: It would be more better to correct as follows.

```suggestion
            if foo && bar&.baz(quux)
```

spec/rubocop/cop/style/sole_nested_conditional_spec.rb
```diff
@@ -663,6 +663,46 @@ def foo
...
+                end
+              end
+            end
+          RUBY
+
+          expect_correction(<<~RUBY)
+            if foo && (ok? bar do
+                  do_something
+                end)
+              end
```
koic, line 702: Ditto :-)

```suggestion
            if foo && ok?(bar) do
                  do_something
                end
              end
```

spec/rubocop/cop/style/sole_nested_conditional_spec.rb
dvandersluis: I agree, but this is the existing autocorrection for `send` nodes I'm just extending it to `csend` and `block` nodes as well to prevent syntax errors from autocorrection.

Would you be good with accepting this as-is and improving autocorrection in a future PR?

koic: OK, I think we can do it separately. It would be useful to leave a `FIXME` comment.

dvandersluis: Added!

koic:
Thanks!

PR #10889, opened 2022-08-08
Do not fork off extra processes if only a single file needs inspection

WJWH:
This MR makes it so that if there is only a single file to be inspected, we do not split off extra worker processes to do the work. This is because doing the work in a separate process can never be faster than doing it inline. I benchmarked this change on my dev machine (with 6 physical cores) and it saves 40-50 ms per run (only if only a single file is inspected of course, otherwise it will take as long as before).

I did not know what to update the test with, so I just made it expect the new text. Let me know if this should be different.

-----------------

* [x] Updated tests.


WJWH:
Sigh, only commit [39a942d](https://github.com/rubocop/rubocop/pull/10889/commits/39a942d3d0a3aaa88af143825a081b9925fcfa06) contains the changes I made. The rest comes from rebasing where I think I did something wrong when adding this repo as a remote :thinking:

dvandersluis:
I like this idea but you have to clean up the commits. It looks like you somehow got a merge of master in between your two commits a653a0e and 39a942d. You should be able to do a `git rebase —interactive` against master and remove the other commits, but if you can’t figure that out you could create a new branch off of master and cherry pick those two commits (and then squash them together)

```sh
git checkout master
git pull master

# interactive rebase (it’ll open an editor with instructions)
git  rebase master —interactive

# or, create a new branch and cherry pick
git checkout -b [branch name]
git cherry pick a653a0e
git cherry pick 39a942d
```

bbatsov:
The proposed change looks good to me. Don't forget to add a changelog entry about it.

WJWH:
I made https://github.com/rubocop/rubocop/pull/10903, which only contains the relevant changes.

PR #10892, opened 2022-08-09
Add new `Lint/NoopRescue` cop

ydah:
This PR is add new `Lint/NoopRescue` cop.
This cop check for suppress or ignore checked exception.

Suppressing or ignoring exceptions hides the problem, makes it invisible, and makes it behave as if it were not a problem.
The test code becomes less reliable and the possibility of problems increases.

-----------------

* [x] Added tests.


dvandersluis (changes requested):
If we're going to add this, it needs to be way more permissive than it currently is.

First off `foo rescue nil` is a very common pattern. Moreso, there are a lot of returned values in a rescue that are not `send` nodes but are not no-ops. For instance:

```ruby
def get_api_value
  MyAPI.request(foo, bar)
rescue APIError
  nil
end
```

```ruby
def resource_available?
  Resource.connect
rescue Connection::TimedOut
  false
end
```

These are completely valid, non-noop responses that will fail this cop.

IMO, this cop should maybe just catch rescues with no body but leave everything else as allowed.

config/default.yml
```diff
@@ -1994,6 +1994,11 @@ Lint/NonLocalExitFromIterator:
   Enabled: true
   VersionAdded: '0.30'
 
+Lint/NoopRescue:
+  Description: 'Check for suppress or ignore checked exception.'
```
dvandersluis, line 1998:
```suggestion
  Description: 'Check for suppressed or ignored rescued exceptions.'
```

lib/rubocop/cop/lint/noop_rescue.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Check for suppress or ignore checked exception.
```
dvandersluis, line 6: Same here.

lib/rubocop/cop/lint/noop_rescue.rb
```diff
@@ -0,0 +1,51 @@
...
+      #
+      #   # good
+      #   def foo
+      #     do_something
+      #   rescue StandardError => e
+      #     do_something
+      #   end
+      #
+      class NoopRescue < Base
+        MSG = "Don't suppress or ignore checked exception."
```
dvandersluis, line 30:
```suggestion
        MSG = "Don't suppress or ignore rescued exceptions."
```

spec/rubocop/cop/lint/noop_rescue_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      expect_no_offenses(<<~RUBY)
+        begin
+          do_something
+        rescue StandardError
+          do_something
+        end
+      RUBY
+    end
+
+    it 'does not registers an offense when catching multiple exceptions' do
```
dvandersluis, line 70:
```suggestion
    it 'does not register an offense when catching multiple exceptions' do
```

spec/rubocop/cop/lint/noop_rescue_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    it 'does not register an offense when inside the method definition' do
+      expect_no_offenses(<<~RUBY)
+        def foo
+        rescue
+          do_something
+        end
+      RUBY
+    end
+
+    it 'does not registers an offense when begin-rescue' do
```
dvandersluis, line 60:
```suggestion
    it 'does not register an offense with begin-rescue' do
```

dvandersluis:
Actually, see `Lint/SuppressedExceptions`. We already have a cop to do this, I'm afraid 😅

ydah:
If I thought it was not pointed out, I could detect it by setting the default: false option to true. (I hadn't noticed. Thank you so much.)

```yaml
Lint/SuppressedException:
  AllowComments: false
  AllowNil: false
````

PR #10897, opened 2022-08-10
Skip a flaky test for RSpec 4

koic:
This PR skips the following flaky test for RSpec 4 GitHub Actions job.

```console
  1) rubocop --server when using `--server` option after updating
  RuboCop displays a restart information message
     Failure/Error:
       expect(`ruby -I . "#{rubocop}" #{options}`).to start_with(
         'RuboCop version incompatibility found, RuboCop server restarting...'
       )

       expected "RuboCop server starting on 127.0.0.1:36753.\nInspecting
       1 file\n\e[32m.\e[0m\n\n1 file inspected, \e[32mno offenses\e[0m
       detected\n" to start with "RuboCop version incompatibility found,
       RuboCop server restarting...
```

Unfortunately, the root cause is unknown yet. It may be related that test-queue is not available.
In any case, this test is run on a CI other than RSpec 4 (beta) job, so there's nothing I've overlooked.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [ ] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #10903, opened 2022-08-10
Single file skip parallel

WJWH:
This MR makes it so that if there is only a single file to be inspected, we do not split off extra worker processes to do the work. This is because doing the work in a separate process can never be faster than doing it inline. I benchmarked this change on my dev machine (with 6 physical cores) and it saves 40-50 ms per run (only if only a single file is inspected of course, otherwise it will take as long as before).

I did not know what to update the test with, so I just made it expect the new text. Let me know if this should be different.

(This is a fixed version of https://github.com/rubocop/rubocop/pull/10889, where I had messed up the rebasing)

-----------------

* [x] Updated tests.

### DIFF
--- a/changelog/fix_speed_up_single_file_inspection.md
+++ b/changelog/fix_speed_up_single_file_inspection.md
@@ -1,0 +1,1 @@
+* [#10903](https://github.com/rubocop/rubocop/pull/10903): Skip forking off extra processes for parallel inspection when only a single file needs to be inspected. ([@wjwh][])
--- a/lib/rubocop/runner.rb
+++ b/lib/rubocop/runner.rb
@@ -64,6 +64,10 @@ module RuboCop
     # instances that each inspects its allotted group of files.
     def warm_cache(target_files)
       saved_options = @options.dup
+      if target_files.length <= 1
+        puts 'Skipping parallel inspection: only a single file needs inspection' if @options[:debug]
+        return
+      end
       puts 'Running parallel inspection' if @options[:debug]
       %i[autocorrect safe_autocorrect].each { |opt| @options[opt] = false }
       Parallel.each(target_files) { |target_file| file_offenses(target_file) }
--- a/spec/rubocop/cli/options_spec.rb
+++ b/spec/rubocop/cli/options_spec.rb
@@ -14,6 +14,11 @@ RSpec.describe 'RuboCop::CLI options', :isolated_environment do # rubocop:disabl
   describe '--parallel' do
     if RuboCop::Platform.windows?
       context 'on Windows' do
+        before do
+          create_file('test_1.rb', ['puts "hello world"'])
+          create_file('test_2.rb', ['puts "what a lovely day"'])
+        end
+
         it 'prints a warning' do
           cli.run ['-P']
           expect($stderr.string).to include('Process.fork is not supported by this Ruby')
@@ -35,7 +40,9 @@ RSpec.describe 'RuboCop::CLI options', :isolated_environment do # rubocop:disabl
       context 'on Unix-like systems' do
         it 'prints a message if --debug is specified' do
           cli.run ['--parallel', '--debug']
-          expect($stdout.string).to match(/Running parallel inspection/)
+          expect($stdout.string).to match(
+            /Skipping parallel inspection: only a single file needs inspection/
+          )
         end
 
         it 'does not print a message if --debug is not specified' do


changelog/fix_speed_up_single_file_inspection.md
```diff
@@ -0,0 +1 @@
+* [#10889](https://github.com/rubocop/rubocop/pull/10889): Skip forking off extra processes for parallel inspection when only a single file needs to be inspected. ([@wjwh][])
```
dvandersluis, line 1: You need to add a trailing newline to this file please.

WJWH: Done :+1:

lib/rubocop/runner.rb
```diff
@@ -64,6 +64,10 @@ def aborting?
     # instances that each inspects its allotted group of files.
     def warm_cache(target_files)
       saved_options = @options.dup
+      if target_files.length <= 1
```
bbatsov, line 67: Btw, shouldn't this be the first line of the method?

dvandersluis:
Can you take a look at this test failure:

```
1) RuboCop::CLI options when supporting server --server-status when server is runnning displays server status
     Failure/Error: expect(output).to match(/RuboCop server \(\d+\) is running./)
     
       expected "RuboCop server is not running.\n" to match /RuboCop server \(\d+\) is running./
       Diff:
       @@ -1 +1 @@
       -/RuboCop server \(\d+\) is running./
       +RuboCop server is not running.
       
     # ./spec/rubocop/cli/options_spec.rb:189:in `block (5 levels) in <top (required)>'
```

it's different than the flaky test failure from #10897.

lib/rubocop/runner.rb
WJWH: If we make this the first line of the method then the `ensure` clause needs to have a separate `begin` block, otherwise it will replace `@options` with nil. I opted for this way because it looked slightly cleaner to my eyes and needed less changes.

Would you like me to rewrite it to skip the `.dup`?

WJWH:
@dvandersluis I looked at the failure but it does not look like it should be caused by the changes in this MR. I tried reproducing the error on my machine but cannot get it to fail the test suite even with many test queue workers. :neutral_face:

lib/rubocop/runner.rb
bbatsov: I see. I didn't notice the `ensure`. Probably it would have been nicer to have here `if/else` instead of `if` + `return`, although I assume RuboCop forced you to write the code like this.

WJWH: The following passes tests and rubocop inspection:

```
    # Warms up the RuboCop cache by forking a suitable number of RuboCop
    # instances that each inspects its allotted group of files.
    def warm_cache(target_files)
      if target_files.length <= 1
        puts 'Skipping parallel inspection: only a single file needs inspection' if @options[:debug]
      else
        begin
          saved_options = @options.dup
          puts 'Running parallel inspection' if @options[:debug]
          %i[autocorrect safe_autocorrect].each { |opt| @options[opt] = false }
          Parallel.each(target_files) { |target_file| file_offenses(target_file) }
        ensure
          @options = saved_options
        end
      end
    end
```

bbatsov:
Can you rebase on top of master?

bbatsov:
@koic Any idea why the Windows jobs are failing? This seems unrelated to the changes.

koic:
This looks like an expected failure on Windows CI.

```console
Failures:

  1) RuboCop::CLI options --parallel on Windows prints a warning
     Failure/Error: expect($stderr.string).to include('Process.fork is not supported by this Ruby')
       expected "" to include "Process.fork is not supported by this Ruby"
     # ./spec/rubocop/cli/options_spec.rb:19:in `block (4 levels) in <top (required)>'
     # ./spec/support/cli_spec_behavior.rb:26:in `block (2 levels) in <top (required)>'
     # ./lib/rubocop/rspec/shared_contexts.rb:30:in `block (4 levels) in <top (required)>'
     # ./lib/rubocop/rspec/shared_contexts.rb:30:in `chdir'
     # ./lib/rubocop/rspec/shared_contexts.rb:30:in `block (3 levels) in <top (required)>'
     # ./lib/rubocop/rspec/shared_contexts.rb:7:in `block (2 levels) in <top (required)>'
     # tasks/spec_runner.rake:40:in `block in run_specs'
     # tasks/spec_runner.rake:52:in `with_encoding'
     # tasks/spec_runner.rake:36:in `run_specs'
     # tasks/spec_runner.rake:165:in `block in <top (required)>'

Finished in 2 minutes 55.7 seconds (files took 4.16 seconds to load)
18732 examples, 1 failure, 15 pending

Failed examples:

rspec ./spec/rubocop/cli/options_spec.rb:17 # RuboCop::CLI options --parallel on Windows prints a warning
```

https://github.com/rubocop/rubocop/runs/7945357268

The message `Process.fork is not supported by this Ruby` is printed by Parallel gem.
https://github.com/grosser/parallel/blob/v1.22.1/lib/parallel.rb#L275

Probably this message is no longer printed because the test only has one inspection target file.

WJWH:
I'll have a look at it this weekend.

WJWH:
Making some files beforehand in the windows tests makes it pass the windows test again. The one remaining test is something about `--server` not restarting because of a version mismatch, but I don't understand how that would be connected to this change.

changelog/fix_speed_up_single_file_inspection.md
koic, line 1:
```suggestion
* [#1003](https://github.com/rubocop/rubocop/pull/10903): Skip forking off extra processes for parallel inspection when only a single file needs to be inspected. ([@wjwh][])
```

WJWH: Whoops, leftover from the original PR where something went wrong with the rebasing. Thanks for catching that.

changelog/fix_speed_up_single_file_inspection.md
```diff
@@ -0,0 +1 @@
+* [#1003](https://github.com/rubocop/rubocop/pull/10903): Skip forking off extra processes for parallel inspection when only a single file needs to be inspected. ([@wjwh][])
```
koic, line 1: Can you update the PR nimber 😅 

```suggestion
* [#10903](https://github.com/rubocop/rubocop/pull/10903): Skip forking off extra processes for parallel inspection when only a single file needs to be inspected. ([@wjwh][])
```

koic:
I commented on the changelog entry, anyway overall it looks good. Can you squash your commits into one?

WJWH:
Thanks for the help and guidance!

Issue #10904, opened 2022-08-10
Style/GuardClause false negative


GeorgeMarcusGG:
## Code

```ruby
def foo
  if true
    1
  else
    2
  end
end
```

## Expected behavior

The cop should have been triggered to suggest this change
```ruby
def foo
  return 1 if true

  2
end
```

This is especially important if the `else` branch has lots of lines.

## Actual behavior

No trigger.

If I remove the `else` branch then the cop is triggered.
```ruby
def foo
  if true
    1
  end
end
```

If I add a `return` to each branch then the cop is triggered.
```ruby
def foo
  if true
    return 1
  else
    return 2
  end
end
```

## Steps to reproduce the problem

Code above

## RuboCop version

```
bundle exec rubocop -V
1.34.1 (using Parser 3.1.2.1, rubocop-ast 1.21.0, running on ruby 3.1.1 x86_64-darwin21)
```

ydah:
That behavior seems correct. `Style/GuardClause` cop is offense for conditional branches with if-else if the body of the condition contains `return` or `break` or `next` or `raise` or `fail`.

Therefore, the following code is not an offense:
```ruby
def func
  if something
    work
  else
    test
  end
end

def func
  unless something
    work
  else
    test
  end
end
```

However, the following code is an offense:
```ruby
def func
  if something
    work
  else
    test || raise('message')
  end
end

def func
  if something
    work and return
  else
    test
  end
end
```

GeorgeMarcus:
Ok. I was just thinking that in this case the `if` is at the end of the function so both branches have an implicit `return`.

Also I don't understand why having or not having an `else` branch should make any difference.

If this is not intended or if you don't think this would be useful then feel free to close the issue.

bbatsov:
That's by design indeed - sometimes an `if`/`else` just reads better, so we wanted to leave people the option to chose what go for. I recall originally this cop was always enforcing the style you prefer, perhaps we left this as an option (or can add an option). At any rate - I want this cop to be relatively relaxed out-of-the-box.

bbatsov:
Btw, I see that the cop examples are clearly incorrect - e.g. we have this very same code listed as something that would be marked as an offense and most of the listings are not wrapped in the `def`s. We'll need to fix this.

PR #10907, opened 2022-08-10
Enhanced documentation for `Style/GuardClause`

ydah:
This PR is following documentation improvements have been made
- Fixed an incorrect bad case that was not wrapped in `def`
- Added a supplement on acceptable cases of conditions with `elsif` or `else` branches

refs: https://github.com/rubocop/rubocop/issues/10904 , https://github.com/rubocop/rubocop/issues/10904#issuecomment-1211741427

-----------------

* [-] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/guard_clause.rb
+++ b/lib/rubocop/cop/style/guard_clause.rb
@@ -6,6 +6,10 @@ module RuboCop
       # Use a guard clause instead of wrapping the code inside a conditional
       # expression
       #
+      # A condition with an `elsif` or `else` branch is allowed unless
+      # one of `return`, `break`, `next`, `raise`, or `fail` is used
+      # in the body of the conditional expression.
+      #
       # @example
       #   # bad
       #   def test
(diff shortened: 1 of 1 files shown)


dvandersluis:
This seems like a duplicate of #10868

ydah:
It appears to me that you are following the documentation of different cases.
In https://github.com/rubocop/rubocop/pull/10868 , it looks like a PR that makes the good case more correct.

Am I missing something?

ydah:
The title and text of this PR have been updated. What do you think?

lib/rubocop/cop/style/guard_clause.rb
```diff
@@ -27,6 +27,31 @@ module Style
...
+      #   end
+      #
+      #   # also good
+      #   def test
+      #     if something
+      #       foo
+      #     else
+      #       bar
+      #     end
+      #   end
```
koic, line 52: I'm worried about the incompatibility to the bad case. What about adding this allowed condition to the cop description text instead of adding an example?

lib/rubocop/cop/style/guard_clause.rb
```diff
@@ -6,6 +6,10 @@ module Style
       # Use a guard clause instead of wrapping the code inside a conditional
       # expression
       #
+      # When condition with an `elsif` or `else` branch is allowed unless
```
bbatsov, line 9: Probably you meant to say "A condition", as "when condition" seems wrong in this context.

lib/rubocop/cop/style/guard_clause.rb
```diff
@@ -6,6 +6,10 @@ module Style
...
+      # one of `return`, `break`, `next`, `raise`, or `fail` is used
+      # in the body of the conditional statement.
```
bbatsov, line 11: statement -> expression

lib/rubocop/cop/style/guard_clause.rb
ydah: Indeed it is. Thank you very much.

ydah:

bbatsov (approved):

PR #10911, opened 2022-08-11
Fix `Style/ClassMethodsDefinitions` for non-self receivers

sambostock:
The `Style/ClassMethodsDefinitions` cop controls if class methods should be defined using

```ruby
class Klass
  class << self
    def method_name
```

or

```ruby
class Klass
  def self.method_name
```

Given the current error message, it should not apply to singleton methods defined on objects that aren't `self`:

```ruby
object = Object.new
def object.method_name
```

> **Note**
> We could potentially follow up with a config that allows controlling if this cop applies **only** to **`self`** or to **_any_ receiver**.

-----------------

* [x] Added tests.


koic:
Thanks!

PR #10928, opened 2022-08-17
Add more autocorrect support on `Style/EachForSimpleLoop`

r7kamura:
If there is a block argument, it should be autocorrectable when the starting point of the range is 0.

```ruby
# bad
(0...10).each do |i|
  puts i
end

# good
10.times do |i|
  puts i
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/each_for_simple_loop.rb
+++ b/lib/rubocop/cop/style/each_for_simple_loop.rb
@@ -28,14 +28,14 @@ module RuboCop
         MSG = 'Use `Integer#times` for a simple loop which iterates a fixed number of times.'
 
         def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
-          return unless offending_each_range(node)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/each_for_simple_loop.rb
```diff
@@ -28,14 +28,14 @@ class EachForSimpleLoop < Base
         MSG = 'Use `Integer#times` for a simple loop which iterates a fixed number of times.'
 
         def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
-          return unless offending_each_range(node)
+          return unless offended?(node)
```
bbatsov, line 31: I think `offending?` is a better term here. :)

r7kamura: You're right, thanks 😃

koic:
Thanks!

PR #10966, opened 2022-08-26
Add autocorrect support to `Style/AccessModifierDeclarations`

r7kamura:
I would like to change `Style/AccessModifierDeclarations` to be able to autocorrect.

I made a little effort about indentation and empty lines with `RuboCop::Cop::RangeHelp#range_by_whole_lines` and some other treatments, but it does not handle them perfectly. They should be autocorrected by other Layout cops.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/access_modifier_declarations.rb
+++ b/lib/rubocop/cop/style/access_modifier_declarations.rb
@@ -9,6 +9,11 @@ module RuboCop
       # Applications of visibility methods to symbols can be controlled
       # using AllowModifiersOnSymbols config.
       #
+      # @safety
+      #   Autocorrection is not safe, because the visibility of dynamically
+      #   defined methods can vary depending on the state determined by
+      #   the group access modifier.
+      #
       # @example EnforcedStyle: group (default)
       #   # bad
       #   class Foo
@@ -63,7 +68,10 @@ module RuboCop
       #
       #   end
       class AccessModifierDeclarations < Base
+        extend AutoCorrector
+
         include ConfigurableEnforcedStyle
+        include RangeHelp
 
         GROUP_STYLE_MESSAGE = [
           '`%<access_modifier>s` should not be',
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -130,6 +158,54 @@ def message(range)
...
+            end
+          else
+            node.arguments.first
+          end
+        end
+
+        def find_argument_less_modifier_node(node)
+          node.parent.children.find do |child|
+            child&.send_type? && child&.method?(node.method_name) && child&.arguments&.empty?
+          end
```
koic, line 176:
```suggestion
          node.parent.children.compact.find do |child|
            child.send_type? && child.method?(node.method_name) && child.arguments.empty?
          end
```

lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -130,6 +158,54 @@ def message(range)
             format(INLINE_STYLE_MESSAGE, access_modifier: access_modifier)
           end
         end
+
+        def find_corresponding_def_node(node)
+          if access_modifier_with_symbol?(node)
+            method_name = node.arguments.first.value
+            node.parent.children.find do |child|
+              child&.def_type? && child&.method?(method_name)
+            end
```
koic, line 167:
```suggestion
            node.parent.children.compact.find do |child|
              child.def_type? && child.method?(method_name)
            end
```

koic:
Could you add `expect_correction` to the existing `expect_offense` in this cop's spec?

lib/rubocop/cop/style/access_modifier_declarations.rb
r7kamura: I thought that it would be wasteful to iterate twice on `#children` for this purpose.
By the way, if we want to remove this safe navigation operators, I think it is better to use `#each_child_node` than to use `#compact`, what do you think?

```suggestion
          node.parent.each_child_node.find do |child|
            child.send_type? && child.method?(node.method_name) && child.arguments.empty?
          end
```

koic: I think it's a good choice.

r7kamura:
I added these 2 changes:

1. Replace `#children` with `#each_child_node`
2. Add `expect_correction` and `expect_no_corrections`

Whether or not to explicitly write `expect_no_corrections` is debatable, but I thought it should be written here because I also want to make it clear that autocorrection is not supported in that situation.

lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -130,6 +158,54 @@ def message(range)
             format(INLINE_STYLE_MESSAGE, access_modifier: access_modifier)
           end
         end
+
+        def find_corresponding_def_node(node)
+          if access_modifier_with_symbol?(node)
+            method_name = node.arguments.first.value
+            node.parent.each_child_node.find do |child|
+              child.def_type? && child.method?(method_name)
```
koic, line 166: `each_child_node` can filter node types.

```suggestion
            node.parent.each_child_node(:def).find do |child|
              child.method?(method_name)
```

lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -130,6 +158,54 @@ def message(range)
...
+            end
+          else
+            node.arguments.first
+          end
+        end
+
+        def find_argument_less_modifier_node(node)
+          node.parent.each_child_node.find do |child|
+            child.send_type? && child.method?(node.method_name) && child.arguments.empty?
```
koic, line 175:
```suggestion
          node.parent.each_child_node(:send).find do |child|
            child.method?(node.method_name) && child.arguments.empty?
```

koic:
Thanks!

PR #10974, opened 2022-08-29
Improve safety section on `Style/SymbolProc`

r7kamura:
Since `Symbol#to_proc` is a bit difficult concept to understand in Ruby and is often tripped up by beginners, I thought it would be better to document it more precisely.

The original text explains it as follows:

> ``proc``s and blocks work differently

but this is not a very appropriate explanation I think, since both equally create an instance of the `Proc` class, so it is not reasonable to explain them as opposites. And worse, Ruby also has a keyword `proc`, so it is better to avoid using this word in a way that could be misinterpreted as a keyword as much as possible.

-----------------

* [ ] Added tests.


bbatsov:
Thanks!

PR #10975, opened 2022-08-29
Fix possible wrong autocorrection in namespace on `Style/PerlBackrefs`

r7kamura:
Inserting constants in the code without `::` prefix may cause problems in situations where `Module.nesting` is not empty, as the test case I added describes.

In order to eliminate the possibility of such unsafe autocorrection, it would be better to prepend `::`, even if it looks a bit verbose.

-----------------

* [x] Added tests.


bbatsov:
Fine by me. Thanks!

PR #10982, opened 2022-08-30
Do not omit parentheses for calls in assignments in conditional branches

gsamokovarov:
Parentheses are required in assignments that call methods in conditional branches. Removing them is a syntax error and the `Style/MethodCallWithArgsParentheses` cop with the style of `omit_parentheses` has been doing just that during autocorrection.

Here is an example that needs the parentheses but is improperly autocorrected:

```ruby
if success = destroy_remote(server)
  delete server
end
```

Removing them, as the autocorrection does, is a syntax error:

```ruby
if success = destroy_remote server # 💣
  delete server
end
```

Such is the case for any assignment or shorthand assignment in any Ruby conditional.

### DIFF
--- a/lib/rubocop/cop/style/method_call_with_args_parentheses/omit_parentheses.rb
+++ b/lib/rubocop/cop/style/method_call_with_args_parentheses/omit_parentheses.rb
@@ -20,8 +20,8 @@ module RuboCop
             return if require_parentheses_for_hash_value_omission?(node)
             return if syntax_like_method_call?(node)
             return if super_call_without_arguments?(node)
-            return if allowed_camel_case_method_call?(node)
             return if legitimate_call_with_parentheses?(node)
+            return if allowed_camel_case_method_call?(node)
             return if allowed_string_interpolation_method_call?(node)
 
             add_offense(offense_range(node), message: OMIT_MSG) do |corrector|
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/method_call_with_args_parentheses/omit_parentheses.rb
```diff
@@ -20,8 +20,9 @@ def omit_parentheses(node)
             return if require_parentheses_for_hash_value_omission?(node)
             return if syntax_like_method_call?(node)
             return if super_call_without_arguments?(node)
-            return if allowed_camel_case_method_call?(node)
             return if legitimate_call_with_parentheses?(node)
+            return if assignment_in_condition?(node)
```
bbatsov, line 24: I think this should go in the above method.

lib/rubocop/cop/style/method_call_with_args_parentheses/omit_parentheses.rb
```diff
@@ -202,6 +203,17 @@ def assigned_before?(node, target)
           def inside_string_interpolation?(node)
             node.ancestors.drop_while { |a| !a.begin_type? }.any?(&:dstr_type?)
           end
+
+          def assignment_in_condition?(node)
```
bbatsov, line 207: I think it'd be better to implement this in terms of pattern matching, as logic that goes through 2 levels of parents is both brittle and hard to understand.

gsamokovarov: I cannot use `def_node_pattern` here because of the way this cop is organized. This is a module that is mixed in into `Style/MethodCallWithArgsParentheses`. I can invoke `NodePattern` by itself, but I am yet to find a good pattern that looks for a grandparent.

I did had a redundant grandparent check, though. The new implementation is simpler.

PR #10988, opened 2022-09-04
Display warning message when both safe and unsafe autocorrect options are specified

koic:
If both `-a` option and `-A` option are specified then `-a` is applied. This is obviously misused and will raise an error.

```console
% echo 'puts "Hi"' | bundle exec rubocop --stdin example.rb -A -a
Error: Both safe and unsafe autocorrect options are specified, use only one.
```

-----------------

* [x] Added tests.


bbatsov:
>  A breaking change occurs if both options cause an error. It would be reasonable to display a warning message I think.

But would it affect anyone in practice? It seems to me it'd be much better to raise an error as the warning can easily be overlooked.

koic:
You are right! This is obviously misused and I've updated this PR to raise an error.

bbatsov:
Thanks!

PR #11001, opened 2022-09-13
[Fix #10977] Add option to raise cop errors, `--raise-cop-error`

### LINKED ISSUE #10977: Add option to surface more informative cop errors

Is there a way we can get `RuboCop::Options` to parse the `:raise_error` option (used in `Commissioner` for Cop errors)? If not, this would be a nice new option to have. 

From what I can see, there is no way to pass in the `@options[:raise_error]` that is used in `#with_cop_error_handling` [here](https://github.dev/rubocop/rubocop/blob/3360eac8244c1f2dce49cfffa00b6f3611c2de6b/lib/rubocop/cop/commissioner.rb#L162). However, it seems like the [spec does test for the option](https://github.dev/rubocop/rubocop/blob/3360eac8244c1f2dce49cfffa00b6f3611c2de6b/spec/rubocop/cop/commissioner_spec.rb#L107) though. 

Is that intentional or something that was overlooked? 

To have `RuboCop::Options` parse `raise_error` option for cop errors. Ideally, the option surfaces more useful information. For example the full `ErrorWithAnalyzedFileLocation.new(cause: e, node: node, cop: cop)` -- which contains the cause, node and cop, rather than just a `StandardError`.

In order to have this option, I'd have to bypass the `RuboCop::Options#parse` which is kind of a hack since it skips the options validator. Plus, it'd only surface a `StandardError` without information about the cop or node. 

## Additional context

The goal of this is for users to have the option to rescue cop-related errors, instead of just printing them to the standard error.

Thankss!

wildmaples:
Fix https://github.com/rubocop/rubocop/issues/10977

We want to be able to rescue cop-related errors. 

There exists `:raise_error`, but it's used for internal tests such as through [cop helper](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/rspec/cop_helper.rb#L49). 

I added a new option that raises with `RuboCop::ErrorWithAnalyzedFileLocation`, which surfaces more information. 

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_add_option_to_raise_cop_errors.md
+++ b/changelog/new_add_option_to_raise_cop_errors.md
@@ -1,0 +1,1 @@
+* [#11001](https://github.com/rubocop/rubocop/pull/11001): Add option to raise cop errors `--raise-cop-error`. ([@wildmaples][])
--- a/docs/modules/ROOT/pages/cops.adoc
+++ b/docs/modules/ROOT/pages/cops.adoc
@@ -10,6 +10,9 @@ enforce different coding conventions.
 
 You can also load xref:extensions.adoc#custom-cops[custom cops].
 
+Cop-related failures are silenced by default but can be turned on using the
+`--raise-cop-errors` option.
+
 == Style
 
 Style cops check for stylistic consistency of your code. Many of the them are
--- a/docs/modules/ROOT/pages/usage/basic_usage.adoc
+++ b/docs/modules/ROOT/pages/usage/basic_usage.adoc
@@ -240,6 +240,9 @@ $ rubocop --only Rails/Blank,Layout/HeredocIndentation,Naming/FileName
 | `--[no-]parallel`
 | Use available CPUs to execute inspection in parallel. Default is parallel.
 
+| `--raise-cop-error`
+| Raise cop-related errors with cause and location. This is used to prevent cops from failing silently. Default is false.
+
 | `-r/--require`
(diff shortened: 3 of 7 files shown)


dvandersluis:
I'm not sure I understand the use case for this? Why do you want rubocop to stop when an error is raised? (Which shouldn't happen in normal operation, of course)

Also there's already `--debug` which will output errors/backtraces when an exception is encountered.

wildmaples:
Hi @dvandersluis, thanks for taking a look at this!! 

We need this option to rescue any cop exceptions when running RuboCop on the Ruby LSP for a better use experience. The TL;DR is that we want to make sure users know there is an error (instead of failing silently using `--debug` which is what we are doing now) because it prevents the file from being processed while open on VS Code. 

We have a more detailed issue you can find here: https://github.com/Shopify/ruby-lsp/issues/250

bbatsov:
I see. That makes sense to me, but it has to be documented very clearly.

bbatsov:
@wildmaples Ping :-)

wildmaples:
Thanks for the reminder, @bbatsov! 

I've added some copy mentioning the new option in the `Cop` and `Basic Usage` section of the Rubocop documentation. I also provided additional context for to describe the option. (https://github.com/rubocop/rubocop/pull/11001/commits/b165c59ce7ff9c3c7f9a80e9b7528cc79ff0b5e9)

bbatsov:
Looks good. I'll include merge this after cutting the next bugfix release.

PR #11007, opened 2022-09-16
[Fix #11006] Allow multiple `elsif` for `Style/IfWithBooleanLiteralBranches`

### LINKED ISSUE #11006: Style/IfWithBooleanLiteralBranches: multiple branches readability

The below method would in my opinion loose in readability if it were corrected by rubocop:

```ruby
      def contains?(rhs, opts = {})
        if Feature::Geometry === rhs
          contains?(BoundingBox.new(@factory).add(rhs))
        elsif rhs.empty?
          true
        elsif empty?
          false
        elsif @min_x > rhs.min_x || @max_x < rhs.max_x || @min_y > rhs.min_y || @max_y < rhs.max_y
          false
        elsif @has_m && rhs.has_m && !opts[:ignore_m] && (@min_m > rhs.min_m || @max_m < rhs.max_m)
          false
        elsif @has_z && rhs.has_z && !opts[:ignore_z] && (@min_z > rhs.min_z || @max_z < rhs.max_z)
          false
        else
          true
        end
      end
```
<sup>

[source](https://github.com/rgeo/rgeo/blob/b2fe667df6c71db0637fdfbf647881d5c780a4e4/lib/rgeo/cartesian/bounding_box.rb#L282-L298)

</sup>

The branching helps understanding clearly which result is expected for which conditions. And in [the doc examples](https://docs.rubocop.org/rubocop/cops_style.html#styleifwithbooleanliteralbranches) we can see that it is well suited for simple if-else cases.

Having an option that allows for multiple branches to stay as it is:

```yaml
Style/IfWithBooleanLiteralBranches:
  Enabled: true
  MultipleBranches: false
```

disabling the rule, but this would make me miss the simple if-else which are to be removed.

koic:
Fixes #11006.

This PR allows multiple `elsif` for `Style/IfWithBooleanLiteralBranches`.

I think it's good to allow when multiple `elsif`s are used instead of a new config option. OTOH, if there is only single `elsif`, `elsif` will not exist due to autocorrection. So I think the current behavior can be kept.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/if_with_boolean_literal_branches.rb
+++ b/lib/rubocop/cop/style/if_with_boolean_literal_branches.rb
@@ -8,6 +8,20 @@ module RuboCop
       # The conditions to be checked are comparison methods, predicate methods, and double negative.
       # `nonzero?` method is allowed by default.
       # These are customizable with `AllowedMethods` option.
+      # This cop targets only `if`s with a single `elsif` or `else` branch.
+      #
+      # [source,ruby]
+      # ----
+      # if foo
+      #   true
+      # elsif bar > baz
+      #   true
+      # elsif qux > quux # Single `elsif` is warned, but two or more `elsif`s are not.
+      #   true
+      # else
+      #   false
+      # end
+      # ----
       #
       # @safety
       #   Autocorrection is unsafe because there is no guarantee that all predicate methods
@@ -57,7 +71,7 @@ module RuboCop
         def_node_matcher :double_negative?, '(send (send _ :!) :!)'
 
         def on_if(node)
-          return unless if_with_boolean_literal_branches?(node)
+          return if !if_with_boolean_literal_branches?(node) || multiple_elsif?(node)
 
           condition = node.condition
           range, keyword = offense_range_with_keyword(node, condition)
@@ -75,6 +89,12 @@ module RuboCop
         end
 
         private
+
+        def multiple_elsif?(node)
+          return false unless (parent = node.parent)
+
+          parent.if_type? && parent.elsif?
+        end
 
         def offense_range_with_keyword(node, condition)
           if node.ternary?
--- a/changelog/change_allow_multiple_elsif_for_style_if_with_boolean_literal_branches.md
+++ b/changelog/change_allow_multiple_elsif_for_style_if_with_boolean_literal_branches.md
@@ -1,0 +1,1 @@
+* [#11006](https://github.com/rubocop/rubocop/issues/11006): Allow multiple `elsif` for `Style/IfWithBooleanLiteralBranches`. ([@koic][])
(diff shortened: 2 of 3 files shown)


BuonOmo (commented):
That was fast! LGTM :)

lib/rubocop/cop/style/if_with_boolean_literal_branches.rb
```diff
@@ -8,6 +8,20 @@ module Style
       # The conditions to be checked are comparison methods, predicate methods, and double negative.
       # `nonzero?` method is allowed by default.
       # These are customizable with `AllowedMethods` option.
+      # Multiple `elsif`s with booleans in branches are allowed for readability.
```
bbatsov, line 11: Perhaps it'd be better to mention that this cop targets only `if`s with a single `else` branch. This seems clearer to me.

koic: Thank you! I've tweaked the text to "This cop targets only `if`s with a single `elsif` or `else` branch".

lib/rubocop/cop/style/if_with_boolean_literal_branches.rb
```diff
@@ -8,6 +8,20 @@ module Style
       # The conditions to be checked are comparison methods, predicate methods, and double negative.
       # `nonzero?` method is allowed by default.
       # These are customizable with `AllowedMethods` option.
+      # This cop targets only `if`s with a single `elsif` or `else` branch.
+      #
+      # [source,console]
```
bbatsov, line 13: Shouldn't it say `ruby` here instead of `console`?

lib/rubocop/cop/style/if_with_boolean_literal_branches.rb
```diff
@@ -76,6 +90,12 @@ def on_if(node)
 
         private
 
+        def multiple_elsif?(node)
```
bbatsov, line 93: We might add this to `rubocop-ast` as well.

lib/rubocop/cop/style/if_with_boolean_literal_branches.rb
koic: Oops! I've fixed.

lib/rubocop/cop/style/if_with_boolean_literal_branches.rb
koic: That's a good idea! I will open a PR to rubocop-ast.

koic: I've opened https://github.com/rubocop/rubocop-ast/pull/239 to rubocop-ast.

koic: The above PR has been closed because there are currently no plans to use it outside of this PR. I would like to extract to rubocop-ast when a concrete reuse use case comes up.

bbatsov:
I had forgotten about this PR. 🤦

PR #11014, opened 2022-09-24
Fix a false positive for `Style/Alias`cop when alias in a method def

ydah:
The following cases will result in a NoMethodError if replaced with alias_method.
Therefore, it is correct to regard it as a violation and not to replace it.

```ruby
def foo
  alias :ala :bala
end
```
-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/alias_spec.rb
+++ b/spec/rubocop/cop/style/alias_spec.rb
@@ -106,18 +106,48 @@ RSpec.describe RuboCop::Cop::Style::Alias, :config do
       RUBY
     end
 
-    it 'does not register an offense for alias_method with explicit receiver' do
+    it 'does not register registers an offense for alias in a def' do
       expect_no_offenses(<<~RUBY)
-        class C
-          receiver.alias_method :ala, :bala
+        def foo
+          alias :ala :bala
         end
       RUBY
     end
 
-    it 'does not register an offense for alias_method in a method def' do
-      expect_no_offenses(<<~RUBY)
-        def method
+    it 'registers an offense for alias in a defs' do
+      expect_offense(<<~RUBY)
+        def some_obj.foo
+          alias :ala :bala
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/alias_spec.rb
```diff
@@ -106,6 +106,51 @@ module M
...
+          alias :ala :bala
+          ^^^^^ Use `alias_method` instead of `alias`.
+        end
+      RUBY
+
+      expect_correction(<<~RUBY)
+        def foo
+          alias_method :ala, :bala
+        end
+      RUBY
```
koic, line 121: This looks strange and like a bug. I think `NoMethodError` is not expected. WDYT?

```ruby
def foo
  alias :to_string :to_s
end

foo
p 42.to_string # => "#<Integer:0x0000000000000055>"

def bar
  alias_method :to_str, :to_s # => undefined method `alias_method' for main:Object (NoMethodError)
end

bar
p 42.to_str
```

koic: @ydah ping :-)

ydah: Oops, I hadn't noticed. Thank you very much.
Indeed it is. It would seem that these should be corrected. Is there any problem to correct them in this PR? (In that case I will change the title)

ydah: Indeed it is. I think it is better to fix it. Can we include the bug fix in this PR?

koic: Yes. This wrong behavior spec cannot be merged. Please update this PR.

bbatsov:
@ydah Ping :-)

ydah:

bbatsov:
Thanks!

PR #11016, opened 2022-09-24
Fix a wrong AllowedPatterns example

ydah:
This PR is fix a wrong AllowedPatterns example
When specifying in the .rubocop.yml file, it is necessary
to specify `AllowedPatterns: ['foo']` instead of `AllowedPatterns: [/foo/]`.

Follow up: https://github.com/rubocop/rubocop-rspec/issues/1395

The test code was also modified to match the actual behavior as mentioned in https://github.com/rubocop/rubocop-rspec/pull/1396#issuecomment-1257026847 .

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/unreachable_loop.rb
+++ b/lib/rubocop/cop/lint/unreachable_loop.rb
@@ -79,7 +79,7 @@ module RuboCop
       #   # bad
       #   2.times { raise ArgumentError }
       #
-      # @example AllowedPatterns: [/(exactly|at_least|at_most)\(\d+\)\.times/] (default)
+      # @example AllowedPatterns: ['(exactly|at_least|at_most)\(\d+\)\.times'] (default)
       #
       #   # good
       #   exactly(2).times { raise StandardError }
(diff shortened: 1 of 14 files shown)


koic:
How about making `AllowedPatterns` and `IgnoredPattrerns` accept both `/regexp/` and `'regexp'`?
It will probably be more convenient for users. And there will be no need to modify documentation or tests.

lib/rubocop/cop/lint/unreachable_loop.rb
```diff
@@ -79,7 +79,7 @@ module Lint
       #   # bad
       #   2.times { raise ArgumentError }
       #
-      # @example AllowedPatterns: [/(exactly|at_least|at_most)\(\d+\)\.times/] (default)
+      # @example AllowedPatterns: ['(exactly|at_least|at_most)\(\d+\)\.times'] (default)
```
koic, line 82: No wonder users see the name Allowed**Patterns** and specify `/.../` instead of `'...'` for cases like this.

koic:
Ah this is my mistake. `!ruby/regexp` is required to specify regexp in YAML. I think your https://github.com/rubocop/rubocop/commit/d1b764c008cc251f2aece2274b4431c104ece0c3 is correct. I think that YAML representation should not have its own specifications for RuboCop's configuration. Can you revert to the previous commit?

ydah:
I should have checked in advance. I updated this PR. Thanks for the review.

koic:
Thanks!

PR #11036, opened 2022-10-01
[Fix #10409] Add `--no-exclude-limit` CLI option

### LINKED ISSUE #10409: Generate `.rubocop_todo.yml` without any exclude limit
(Apologies if this has been discussed previously, it's hard to search for with such a large issue volume.)

When setting up RuboCop on a project with a lot of existing offences, running `--auto-gen-config` results in many cops being entirely disabled (because there are more than the default limit of 15 occurrences).

To avoid that, I typically pass `--exclude-limit` with an arbitrary high limit such as `999999`.

I'd like to have a cleaner way to specify that no limit should supply, e.g:
- `--exclude-limit=0` (treat `0` as a special indicator)
- `--exclude-limit=-1` (treat any negative number as a special indicator)
- `--no-exclude-limit`

n/a

## Additional context

n/a

r7kamura:
Resolve https://github.com/rubocop/rubocop/issues/10409.

I tried this on our project, and as mentioned in #10409, it's better now that we don't have to pass an arbitrary high value (e.g. `99999`) that is a bit hard to understand, like this:

```diff
diff --git a/.rubocop_todo.yml b/.rubocop_todo.yml
index 3d93884d64..97229741b7 100644
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -1,6 +1,6 @@
 # This configuration was generated by
-# `rubocop --auto-gen-config --exclude-limit 99999 --no-auto-gen-timestamp`
+# `rubocop --auto-gen-config --no-exclude-limit --no-auto-gen-timestamp`
 # using RuboCop version 1.36.0.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
```

The reason why this option (or the 99999 workaround) is necessary is that without this, cops detected in 15 or more files will be disabled for the entire project, and new code violating that cop will continue to be added in the future.

-----------------

* [x] Added tests.


changelog/new_no_exclude_limit_cli_option.md
```diff
@@ -0,0 +1 @@
+* [#11036](https://github.com/rubocop/rubocop/pull/11036): Add `--no-exclude-limit` CLI option. ([@r7kamura][])
```
koic, line 1:
```suggestion
* [#10409](https://github.com/rubocop/rubocop/issues/10409): Add `--no-exclude-limit` CLI option. ([@r7kamura][])
```

spec/rubocop/cli/auto_gen_config_spec.rb
```diff
@@ -1236,6 +1236,45 @@ def function(arg1, arg2, arg3, arg4, arg5, arg6, arg7)
       expect(File.readlines('.rubocop_todo.yml')[2]).to match(/# using RuboCop version .*/)
     end
 
+    describe 'when --no-exclude-limit is given' do
+      before do
+        Array.new(16) do |i|
```
koic, line 1241: `Array.new(16)` creates an array with 16 `nil` elements. So the block value of `i` is always `nil`. Is this the expected behavior of `16.times`?

```suggestion
        16.times do |i|
```

koic: And, does `RuboCop::Options::DEFAULT_MAXIMUM_EXCLUSION_ITEMS + 1` resolve magic number `16`?

r7kamura: Indeed. At first I used `Array.new` based on `Performance/TimesMap` because I was using the return value, but now I don't use it, so `Integer#times` is more appropriate 👌

r7kamura: > And, does `RuboCop::Options::DEFAULT_MAXIMUM_EXCLUSION_ITEMS + 1` resolve the magic number `16`?

That's the hard part.
If it was the only place we use this value, it would be fine, but we need to check that `exampleN.rb` appears 16 times in .rubocop_todo.yml after this, but I think it would be pretty complicated code to try to reference that value there. The current simple implementation seems to have its advantages in terms of clarity.
What do you think of this?

changelog/new_no_exclude_limit_cli_option.md
r7kamura: > If the change has a related GitHub issue (e.g. a bug fix for a reported issue), put a link to the issue

Oh I didn't know that. Thank you!

spec/rubocop/cli/auto_gen_config_spec.rb
```diff
@@ -1236,6 +1236,45 @@ def function(arg1, arg2, arg3, arg4, arg5, arg6, arg7)
       expect(File.readlines('.rubocop_todo.yml')[2]).to match(/# using RuboCop version .*/)
     end
 
+    describe 'when --no-exclude-limit is given' do
+      before do
+        16.times do |i|
+          create_file("example#{i}.rb", [' '])
+        end
+      end
```
koic, line 1244:
```suggestion
      let(:num_over_default_max_exclusion) { RuboCop::Options::DEFAULT_MAXIMUM_EXCLUSION_ITEMS + 1 }

      before do
        num_over_default_max_exclusion.times do |i|
          create_file("example#{i}.rb", [' '])
        end
      end
```

spec/rubocop/cli/auto_gen_config_spec.rb
koic: It would be better whether `16` could be clarified. Please see https://github.com/rubocop/rubocop/pull/11036/files#r985230949.

bbatsov:
Looks good to me. Might be good to give some practical examples of the usefulness of this flag in the docs, as I doubt many people will figure it out otherwise.

PR #11041, opened 2022-10-02
[Fix #11040] Fix a false positive for `Style/IfUnlessModifier`

### LINKED ISSUE #11040: Unsafe Style/IfUnlessModifier Autocorrection for Code Using `defined?`
Say that I have code like:

```ruby
# Original Code
unless defined?(value)
  value = 'default'
end
```

This is reported as a `Style/IfUnlessModifier` failure.  If I autocorrect it, it changes to:

```ruby
# Autocorrected Code
value = 'default' unless defined?(value)
```

If `value` isn't defined, the original code will set it to `'default'` while the autocorrected code will set it to `nil`.

I usually try to avoid using `defined?` altogether.  But the example above is similar to what I was seeing in some legacy code and I wanted to document it.

Often code like this can be rewritten to use `||=`, as long a `false` value doesn't ever need to override the default.  Or it could be rewritten to use an additional `value_defined` variable that is set before the assignment of `value`.

I'm not sure what an autocorrection could be for this situation.  Maybe RuboCop shouldn't do anything here?

## RuboCop version

```
$ [bundle exec] rubocop -V
1.36.0 (using Parser 3.1.2.1, rubocop-ast 1.21.0, running on ruby 3.0.3) [x86_64-darwin20]
  - rubocop-rails 2.16.1
```

koic:
Fixes #11040.

This PR fixes a false positive for `Style/IfUnlessModifier` when `defined?`'s argument value is undefined.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_style_if_unless_modifier.md
+++ b/changelog/fix_a_false_positive_for_style_if_unless_modifier.md
@@ -1,0 +1,1 @@
+* [#11040](https://github.com/rubocop/rubocop/issues/11040): Fix a false positive for `Style/IfUnlessModifier` when `defined?`'s argument value is undefined. ([@koic][])
--- a/lib/rubocop/cop/style/if_unless_modifier.rb
+++ b/lib/rubocop/cop/style/if_unless_modifier.rb
@@ -11,6 +11,20 @@ module RuboCop
       # cop. The tab size is configured in the `IndentationWidth` of the
       # `Layout/IndentationStyle` cop.
       #
+      # NOTE: It is allowed when `defined?` argument has an undefined value,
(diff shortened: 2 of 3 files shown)


bbatsov:
Seems to me this will still be broken if you have more complex logic like `defined?(foo) &&`, so I'm wondering if it's worth trying to account for `defined?` at all. 

We should probably document this shortcoming with or without the proposed fix.

bbatsov:
@koic Seems we forgot about this PR.

koic:
Sorry for delay. I will retake this in a little while.

koic:
I've updated this PR to resolve the oversight.

bbatsov:
Thanks!

PR #11042, opened 2022-10-03
Mark `Lint/OrderedMagicComments` as unsafe autocorrection

koic:
This cop's autocorrection is unsafe because file encoding may change.

```console
# bad

# frozen_string_literal: true
# encoding: ascii
p [''.frozen?, ''.encoding] #=> [true, #<Encoding:UTF-8>]

# good

# encoding: ascii
# frozen_string_literal: true
p [''.frozen?, ''.encoding] #=> [true, #<Encoding:US-ASCII>]
```

It may be the intended file encoding, but the behavior is changed.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_mark_lint_ordered_magic_comments_as_unsafe_autocorrection.md
+++ b/changelog/change_mark_lint_ordered_magic_comments_as_unsafe_autocorrection.md
@@ -1,0 +1,1 @@
+* [#11042](https://github.com/rubocop/rubocop/pull/11042): Mark `Lint/OrderedMagicComments` as unsafe autocorrection. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Hmm, isn't it better to just avoid doing this autocorrection?

koic:
It may make sense. However, since RuboCop 0.53, this cop has already been released as safe autocorrection and I haven't seen any negative feedback. I found this problem while implementing #11044.
How about seeing what happens with this change?

bbatsov:
I meant not to disable the autocorrect, but to make the current autocorrect algorithm smarter, so it would avoid the problematic case. 

But I'd be okay with marking this as unsafe.

bbatsov:
I see what you mean, but ideally we'll need to explain this better - e.g. the unsafe behavior happens only if there was a bug in the source code to begin with.

PR #11045, opened 2022-10-04
Suggest to "Open the singleton class with `class << self`" when `Style/ModuleFunction` is set to `forbidden`

RDeckard:
**Context:**
``` yml
Style/ModuleFunction:
  EnforcedStyle: forbidden
```

When `extend self` or `module_function` is used in this context ⬆️, the cop message might suggest to "_Open the singleton class with `class << self` instead._".
WDYT?

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_module_function-forbidden_msg.md
+++ b/changelog/change_module_function-forbidden_msg.md
@@ -1,1 +1,0 @@
-* [#11045](https://github.com/rubocop/rubocop/pull/11045): Change the message by adding "*Open the singleton class with `class << self` instead*" when `Style/ModuleFunction` is set to `forbidden`. ([@rdeckard][])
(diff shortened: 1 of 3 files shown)


RDeckard:
[The broken `ruby-2.6-spec` CI step](https://circleci.com/gh/rubocop/rubocop/242160) might be flacky?

bbatsov:
We rarely use messages that are multi-sentence, so I think it would probably be better to improve the cop documentation instead. I see we have no suggestions for the forbidden style at all there and in general the documentation is pretty basic to begin with.

RDeckard:
> I think it would probably be better to improve the cop documentation instead.

Good idea, I've reverted the changes and just updated the documentation. 👍

bbatsov:
Thanks!

PR #11046, opened 2022-10-05
[FIx #9227] Change to better suggestion for Style/GuardClause

### LINKED ISSUE #9227: Misleading example for Style/GuardClause

KessaPassa:

follow up to #10868

-----------------

* [ ] Added tests.


lib/rubocop/cop/style/guard_clause.rb
```diff
@@ -157,11 +157,13 @@ def register_offense(node, scope_exiting_keyword, conditional_keyword)
...
 
-        def guard_clause_source(guard_clause)
+        def guard_clause_exiting_keyword(guard_clause)
           parent = guard_clause.parent
 
-          if parent.and_type? || parent.or_type?
+          if parent.and_type?
             guard_clause.parent.source
+          elsif parent.or_type?
+            "return #{guard_clause.parent.source}"
```
KessaPassa, line 166: Do you know any better condition to switch on/off return?

daipom, line 166: Looks like we need `return` on the `and` case too.

I think this is not related to `and` or `or`.
It seems that the same problem exists in `and`.

The following two codes can have different results.

```ruby
if something
  work and return
else
  test
end
```

```ruby
work and return if something
test
```

It is the following case:

* `something` : `true`
* `work`: `false`

```ruby
if true
  false and return
else
  p "hoge"
end
=> (no output)
```

```ruby
false and return if true
p "hoge"
=> "hoge"
```

KessaPassa: @daipom Thank you for your comment !
I think your code is correct.
I'll check it.

KessaPassa: I fixed this commit.
d1f0a224fce9e6f351fbae5e709b95fb00e92489

daipom: Hmm, indeed it is...

Given this example, perhaps we should recommend this trailing-if-style (such as `work if cond`) only when we are sure that a `raise` or `return` will occur in the block.

It would be correct not to recommend this if it is unclear whether a `raise` or `return` will definitely occur in the block.

bbatsov:
Can you summarize here what's the gist of the PR? I'm finding this a bit hard to understand from the code changes.

KessaPassa:
summary: This PR fix is rubocop's suggested code to be equivalent to the original code.
Currently, rubocop warns as follows.

when nested guard clause with `or` operator.
before
```ruby
if something
  foo || raise('exception')
else
  ok
end
```

after
```ruby
foo || raise('exception') if something
ok
```
when 
`something: true`
`foo: true`
`ok` is executed.

when nested guard clause with `and` operator.
before
```ruby
if something
  foo and raise('exception')
else
  ok
end
```

after
```ruby
foo and raise('exception') if something
ok
```
when 
`something: true`
`foo: false`
`ok` is executed.

This PR changes rubocop's suggested code as follows.

when nested guard clause with `or` operator.
before
```ruby
if something
  foo || raise('exception')
else
  ok
end
```

after
```ruby
return foo || raise('exception') if something
ok
```
when 
`something: true`
`foo: true`
`ok` is not executed.

when nested guard clause with `and` operator.
before
```ruby
if something
  foo and raise('exception')
else
  ok
end
```

after
```ruby
return (foo and raise('exception')) if something
ok
```
when 
`something: true`
`foo: false`
`ok` is not executed.

@bbatsov Does this comment solve your question?

spec/rubocop/cop/style/guard_clause_spec.rb
```diff
@@ -132,14 +132,20 @@ def func
           test
         end
       end
+
+      def func
```
bbatsov, line 136: I don't think that's the right place for this bit of code, given the description of the test. You can add a separate test for it.

lib/rubocop/cop/style/guard_clause.rb
```diff
@@ -226,9 +226,14 @@ def range_of_branch_to_remove(node, guard)
           branch.source_range
         end
 
-        def guard_clause_source(guard_clause)
+        def guard_clause_exiting_keyword(guard_clause)
```
bbatsov, line 229: I don't see the point in renaming this. Frankly, I find the new name confusing, so if you still want to change it should be something that reflects better the purpose of the method.

bbatsov:
Generally the `return` is not needed for one-line bodies method bodies, so in an ideal world this should be factored in when making such suggestions. Some of the tests were smaller they should have been.

KessaPassa:
> Generally the return is not needed for one-line bodies method bodies

Did you mean this code:
```ruby
def func
  if something
    foo || raise('exception')
  end
end
```

should be autocorrected as:
```ruby
def func
  foo || raise('exception') if something
end
```

instead of:
```ruby
def func
  return foo || raise('exception') if something
end
```

?

if so this PR does not process this case.

PR #11059, opened 2022-10-09
Re-enable xitted tests

tdeo:
There are 2 tests currently xitted in the codebase:
1. Behavior of `qualified_cop_name` when 2 cops have the same name in different namespaces

Re-enabling this one by adding test cops matching this condition

2. One about `ConfigObsoletion` behavior for extracted cops which had been disabled because of random failures.

I was able to reproduce by running `rspec ./spec/rubocop/config_obsoletion_spec.rb[1:1:2:3:1,1:1:3:1] --seed 47550 -fd` and tracked it down to pollution of the `RuboCop::ConfigObsoletion.files` class variables, fixed by setting a cleanup `after` block for all tests in that file.
I also re-ordered values of some `expect` to have the expected order `expect(value).to eq expected_value` for the failure messages to be accurate.

-----------------

* [x] Added tests.
* [ ] ~~Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.~~

### DIFF
--- a/spec/project_spec.rb
+++ b/spec/project_spec.rb
@@ -5,6 +5,7 @@ RSpec.describe 'RuboCop Project', type: :feature do
     RuboCop::Cop::Cop
       .registry
       .without_department(:Test)
+      .without_department(:Test2)
       .without_department(:InternalAffairs)
       .cops
       .map(&:cop_name)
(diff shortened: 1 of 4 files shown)


bbatsov:
Nicely done!

PR #11064, opened 2022-10-11
[Fix #11024] Add `require_always` option to `Style/EndlessMethod`

### LINKED ISSUE #11024: How to enforce endless methods?
Hi, I'm trying to get Rubocop to replace one-liners with endless methods in ruby 3.0.2.

I've tried different combinations of the `Style/EndlessMethod` and `Style/SingleLineMethods` cops, without any success.

--------

## Expected behavior

Suppose I have a file called `example.rb` with the following code:

```ruby
class Example
  def a_method
    "42"
  end
end
```

And the following `.rubocop.yml` file:

```yml
Style/EndlessMethod:
  Enabled: true
  EnforcedStyle: allow_single_line
```

I'd expect `rubocop -a example.rb` to report the method as endless-able, and correct it to:

```ruby
class Example
  def a_method = "42"
end
```

## Actual behavior

I only see other cops being ran, but there's no trace of the `Style/EndlessMethod` cop:

```
❯ rubocop -c .rubocop.yml -A example.rb --debug
configuration from /tmp/.rubocop.yml
Default configuration from /Users/andreaschiavini/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/rubocop-1.36.0/config/default.yml
Inspecting 1 file
Scanning /tmp/example.rb
C

Offenses:

example.rb:3:1: C: Style/Documentation: Missing top-level documentation comment for class Example.
class Example
^^^^^^^^^^^^^
example.rb:5:5: C: [Corrected] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
    "42"
    ^^^^

Finished in 0.13531199999852106 seconds
```

- create the ruby file above
- create the provided `.rubocop.yml` enabling the endless method cop
- run it

## RuboCop version

```
1.36.0 (using Parser 3.1.2.1, rubocop-ast 1.21.0, running on ruby 3.0.2) [arm64-darwin21]
```

koic:
Fixes #11024.

This PR adds `require_always` option to `Style/EndlessMethod` that all endless method definitions are required.

```ruby
# EnforcedStyle: require_always

# bad
def my_method
  x
end

# bad
def my_method
  x.foo
   .bar
   .baz
end

# good
def my_method = x

# good
def my_method = x.foo
                 .bar
                 .baz
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/endless_method.rb
+++ b/lib/rubocop/cop/style/endless_method.rb
@@ -5,8 +5,10 @@ module RuboCop
     module Style
       # Checks for endless methods.
       #
-      # It can enforce either the use of endless methods definitions
-      # for single-lined method bodies, or disallow endless methods.
+      # It can enforce endless method definitions whenever possible. It can also disallow multiline
+      # endless method definitions or all endless definitions.
+      #
+      # `require_always` style enforces endless method definitions for single statement methods.
       #
       # Other method definition types are not considered by this cop.
       #
@@ -15,36 +17,94 @@ module RuboCop
       # * allow_single_line (default) - only single line endless method definitions are allowed.
       # * allow_always - all endless method definitions are allowed.
       # * disallow - all endless method definitions are disallowed.
+      # * require_always - all endless method definitions are required.
       #
       # NOTE: Incorrect endless method definitions will always be
       # corrected to a multi-line definition.
       #
       # @example EnforcedStyle: allow_single_line (default)
-      #   # good
-      #   def my_method() = x
-      #
       #   # bad, multi-line endless method
-      #   def my_method() = x.foo
-      #                      .bar
-      #                      .baz
+      #   def my_method = x.foo
+      #                    .bar
+      #                    .baz
+      #
+      #   # good
+      #   def my_method
+      #     x
+      #   end
+      #
+      #   # good
+      #   def my_method = x
+      #
+      #   # good
+      #   def my_method
+      #     x.foo
+      #      .bar
+      #      .baz
+      #   end
       #
       # @example EnforcedStyle: allow_always
       #   # good
-      #   def my_method() = x
+      #   def my_method
+      #     x
+      #   end
       #
       #   # good
-      #   def my_method() = x.foo
-      #                      .bar
-      #                      .baz
+      #   def my_method = x
+      #
+      #   # good
+      #   def my_method = x.foo
+      #                    .bar
+      #                    .baz
+      #
+      #   # good
+      #   def my_method
+      #     x.foo
+      #      .bar
+      #      .baz
+      #   end
       #
       # @example EnforcedStyle: disallow
       #   # bad
-      #   def my_method() = x
+      #   def my_method = x
       #
       #   # bad
-      #   def my_method() = x.foo
-      #                      .bar
-      #                      .baz
+      #   def my_method = x.foo
+      #                    .bar
+      #                    .baz
+      #
+      #   # good
+      #   def my_method
+      #     x
+      #   end
+      #
+      #   # good
+      #   def my_method
+      #     x.foo
+      #      .bar
+      #      .baz
+      #   end
+      #
+      # @example EnforcedStyle: require_always
+      #   # bad
+      #   def my_method
+      #     x
+      #   end
+      #
+      #   # bad
+      #   def my_method
+      #     x.foo
+      #      .bar
+      #      .baz
+      #   end
+      #
+      #   # good
+      #   def my_method = x
+      #
+      #   # good
+      #   def my_method = x.foo
+      #                    .bar
+      #                    .baz
       #
       class EndlessMethod < Base
         include ConfigurableEnforcedStyle
@@ -56,12 +116,16 @@ module RuboCop
         CORRECTION_STYLES = %w[multiline single_line].freeze
         MSG = 'Avoid endless method definitions.'
         MSG_MULTI_LINE = 'Avoid endless method definitions with multiple lines.'
+        MSG_REQUIRE_ALWAYS = 'Use endless method definitions.'
 
         def on_def(node)
-          if style == :disallow
-            handle_disallow_style(node)
-          else
+          case style
+          when :allow_single_line, :allow_always
             handle_allow_style(node)
+          when :disallow
+            handle_disallow_style(node)
+          when :require_always
+            handle_require_always_style(node)
           end
         end
 
@@ -76,6 +140,14 @@ module RuboCop
           end
         end
 
+        def handle_require_always_style(node)
+          return if node.endless? || !node.body || node.body.begin_type? || node.body.kwbegin_type?
+
+          add_offense(node, message: MSG_REQUIRE_ALWAYS) do |corrector|
+            correct_to_endless_method(corrector, node)
+          end
+        end
+
         def handle_disallow_style(node)
           return unless node.endless?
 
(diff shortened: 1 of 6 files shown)


bbatsov:
We'll have to think about the naming (or at least improve the high-level description for the new style), though, as now I find it to be somewhat confusing.

koic:
I've updated to `EnforcedStyle: require_endless_method`.

The `require_endless_method` style enforces endless method definitions for a single statement (not a single line) containing a method chain with a line break.

```ruby
# EnforcedStyle: require_endless_method

# bad
def my_method
  x
end

# bad
def my_method
  x.foo
   .bar
   .baz
end

# good
def my_method = x

# bad
def my_method = x.foo
                 .bar
                 .baz
```

Possibly a bit different than what #11024 wants, but it's simpler in name and behavior and always enforces `EnforcedStyle: require_endless_method`. cc @jonas054 

I also thought of `EnforcedStyle: single_line_endless_method` which allows the following case.

```ruby
# EnforcedStyle: single_line_endless_method

# bad
def my_method
  x
end

# bad
def my_method = x.foo
                 .bar
                 .baz

# good
def my_method = x

# good
def my_method
  x.foo
   .bar
   .baz
end
```

But its name and behavior are a little more complicated than `EnforcedStyle: require_endless_method`.

So, I've updated this PR using `EnforcedStyle: require_endless_method`. How about?

bbatsov:
> require_endless_method

Or perhaps `always_require`?  I'll have to think about the naming here a bit more and how this will tie with `always_for_single_line` or something like this.

koic:
> Or perhaps `always_require`?

This name represents the current implementation that this PR adds. OHTH, there is already an option called `allow_always`, I updated it to `require_always` instead.

```diff
       # * allow_single_line (default) - only single line endless method definitions are allowed.
       # * allow_always - all endless method definitions are allowed.
       # * disallow - all endless method definitions are disallowed.
+      # * require_always - all endless method definitions are required.
```

lib/rubocop/cop/style/endless_method.rb
```diff
@@ -5,8 +5,11 @@ module Cop
     module Style
       # Checks for endless methods.
       #
-      # It can enforce either the use of endless methods definitions
-      # for single-lined method bodies, or disallow endless methods.
+      # It can enforce either the use of endless methods definitions for single-lined method bodies,
+      # or disallow endless methods.
+      #
+      # `require_always` style enforces endless method definitions for a single statement
+      # (not a single line) containing a method chain with a line break.
```
jonas054, line 12: I think the second line of this paragraph is a bit confusing, because the implementation of this style doesn't check whether there are any line breaks, and doesn't care if it's a method chain och just a numeric literal for example. So:

`require_always` style enforces endless method definitions for single statement methods.

lib/rubocop/cop/style/endless_method.rb
```diff
@@ -5,8 +5,11 @@ module Cop
     module Style
       # Checks for endless methods.
       #
-      # It can enforce either the use of endless methods definitions
-      # for single-lined method bodies, or disallow endless methods.
+      # It can enforce either the use of endless methods definitions for single-lined method bodies,
+      # or disallow endless methods.
```
jonas054, line 9: It's difficult to describe all the different styles now in a single compact paragraph, since it's a mixture of enforcing and allowing, and sometimes for single lines and sometimes not. Here's my suggestion:

It can enforce endless method definitions whenever possible. It can also disallow multiline endless method definitions or all endless definitions.

lib/rubocop/cop/style/endless_method.rb
```diff
@@ -15,36 +18,94 @@ module Style
       # * allow_single_line (default) - only single line endless method definitions are allowed.
       # * allow_always - all endless method definitions are allowed.
```
jonas054, line 19: By the way, is there any difference between `EnforcedStyle: allow_always` and `Enabled: false`?

lib/rubocop/cop/style/endless_method.rb
```diff
@@ -76,6 +141,14 @@ def handle_allow_style(node)
           end
         end
 
+        def handle_require_always_style(node)
+          return if node.endless? || !node.body || node.body.begin_type? || node.body.kwbegin_type?
+
+          add_offense(node, message: MSG_REQUIRE_ALWAYS) do |corrector|
```
jonas054, line 147: I see a possible problem if we report and autocorrect methods into endless methods that are too long for a single line. On the other hand, there's autocorrection in `Layout/LineLength` too, so some of these cases will be solved by the autocorrection loop.

jonas054 (changes requested):

lib/rubocop/cop/style/endless_method.rb
koic: I like the proposed text, it's simple! 🙏  I've update it!

lib/rubocop/cop/style/endless_method.rb
koic: Ah, indeed it is!
It's possible that the option will be deprecated, but the incompatibility is annoying.

lib/rubocop/cop/style/endless_method.rb
koic: I wasn't aware of this problem. I will take a look too.

bbatsov:
After thinking about this a bit more - without something like `require_single_line` I think that the cop will remain to be pretty weird. I think the use cases for `require_always` will be relatively limited in comparison.

metalelf0:
Is there any update on this PR? Can I help somehow to get this merged?

koic:
I was late retaking this. I will retake in a few days.

bbatsov:
@koic One more forgotten PR. :-)

smcabrera:
> I was late retaking this. I will retake in a few days.

Hey @koic any updates on this? Anything I can do to help?

ttanimichi:
I think one line method definition should be replaced to endless method definition.

```ruby
# bad
def edit; end

# good
def edit = nil
```

lib/rubocop/cop/style/endless_method.rb
kalivara:
```ruby
def handle_require_always_style(node)
  return if node.endless? || !node.body || node.body.begin_type? || node.body.kwbegin_type?

  line_length = node.body.source.length + definition.length
  return if line_length > config_store.line_length # or however this works

  add_offense(node, message: MSG_REQUIRE_ALWAYS) do |corrector|
    correct_to_endless_method(corrector, node)
  end
end

def definition(node) = "def #{node.method_name}#{arguments(node)} = "

def correct_to_endless_method(corrector, node)
  replacement = "#{definition}#{node.body.source}"
  corrector.replace(node, replacement)
end
```

Abdullah-l:
This PR is stale for a year, what can we do to get it merged?

lvonk:
It would be great if this came available... 🙏🏼 any help needed?

jtannas:
Probably a task for a follow-up PR, but `require_single_line` would be another nice option to have.
Please and thank you :bow: 
```
      # @example EnforcedStyle: require_single_line
      #   # bad
      #   def my_method
      #     x
      #   end
      #
      #   # bad
      #   def my_method = x.foo
      #                    .bar
      #                    .baz
      #
      #   # good
      #   def my_method = x
      #
      #   # good
      #   def my_method
      #     x.foo
      #      .bar
      #      .baz
      #   end
```

jtannas:
@koic I took a crack at building on top of your work to implement `require_single_line` if you're interested
https://github.com/koic/rubocop/pull/4

jtannas:
@koic It seems you're busy with other work.
My apologies if this is premature, but do you mind if I take over this task to get it over the finish line?
Please and thank you :bow:

koic:
Closing this as it has been addressed in #13874.

Issue #10930, opened 2022-08-17
CLI options for specifying source code range


r7kamura:
## Is your feature request related to a problem? Please describe.

I would like to have CLI options to autocorrect offenses by specifying a source code range like this:

```
rubocop \
  --autocorrect \
  --start-line 1 \
  --start-column 2 \
  --last-line 3 \
  --last-column 4 \
  foo.rb
```

This is not often used directly by humans, but mainly for program use, such as editor integrations. I'm creating a VSCode extension for Ruby and thought it would be nice to have this feature in its development.

- https://github.com/r7kamura/vscode-ruby-light

For example, the [Language Server Protocol](https://microsoft.github.io/language-server-protocol/) has the following specifications:

- DocumentRangeFormatting
- Diagnostics & Quick Fix

This feature can be used to provide VSCode with the ability to fix only certain offenses like this:

![diagnostics](https://user-images.githubusercontent.com/111689/185023861-e4bf6237-4cd6-4439-8497-b843dc3ea2c3.gif)

![184560748-a2057574-78e3-41df-956b-901d20627be4](https://user-images.githubusercontent.com/111689/185023921-bff195b3-2b74-42f2-9647-4896f36bbe4d.gif)

## Describe the solution you'd like

The implementation should not be difficult. All we need to do is to filter the collected offenses by the specified range. (Maybe at `RuboCop::Runner` or somewhere)

## Describe alternatives you've considered

The demo above appears to work well, but in fact it is realized in a very dirty way and does not work well in many cases. If RuboCop provided CLI options to specify a range, we would be able to achieve the functionality described above.

For now, I am rewriting the entire file once with `rubocop --autocorrect`, then converting the diff to an array of patches and applying only the patches in the range. This is the same method used in other editor integration with RuboCop, such as [vscode-ruby](https://github.com/rubyide/vscode-ruby).

This method has the flaw that it does not work well when the offense and patch locations are far apart, because it works depending on the location of the corrector, not on the location of the offense. For example, a reordering type of Cop, such as `Bundler/OrderedGems`, will fail with this method.

r7kamura:
> but in fact it is realized in a very dirty way and does not work well in many cases.

For now, I am rewriting the entire file once with `rubocop --autocorrect`, then converting the diff to an array of patches and applying only the patches in the range. This is the same method used in other editor integration with RuboCop, such as [vscode-ruby](https://github.com/rubyide/vscode-ruby).

[...]

Issue #10453, opened 2022-03-14
Run the auto-correct flags only for specified files and lines


axcdnt:
## Is your feature request related to a problem? Please describe.

I'm working on a code base that until very recently, wasn't using RuboCop. We've decided to apply it to make sure the team is using the same standards. To make things a bit easier, we're combining with [Pronto](https://github.com/prontolabs/pronto) gem, that basically helps by running RuboCop only for the modified files/lines.This way, we avoid huge non-related changes for the PR.

## Describe the solution you'd like

I've always used RuboCop's `-a` and `-A` flag for my own projects, but it always applies for the whole file. I'd like to have a way to specify a bunch of files and their respective lines as the input for auto-correct:

```
bundle exec rubocop -a /
    file_a.rb:33 /
    file_b.rb:120 /
```

## Describe alternatives you've considered

N/A

## Additional context

I guess this would be very handy for teams gradually starting to use RuboCop. 

Thanks a ton for the great work with RuboCop! 🤗 🚀

crokobit:
Need this feature too!

ballerolu:
In the same boat for my team. We're gradually enforcing Rubocop. It would be nice to run auto-correct for minor rules on only changed lines (same way Pronto runs check on only changed lines).

jonas054:
My suggestion is that you request this feature in [pronto](https://github.com/prontolabs/pronto). It's almost there already. By adding

```yaml
rubocop:
  suggestions: true
```

in `.pronto.yml` you get output like:

```
activerecord/lib/active_record/transactions.rb:295 W: Layout/LeadingCommentSpace: Missing space after `#`.

``suggestion
    def save(**) # :nodoc:
``
```

so all that's missing is for `pronto` to write those suggestions to the files.

PR #11072, opened 2022-10-12
Allow specifying code ranges for autocorrect

HeroProtagonist:
Adds command line flags that when present will filter autocorrections performed

Fixes https://github.com/rubocop/rubocop/issues/10930

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/autocorrect_logic.rb
+++ b/lib/rubocop/cop/autocorrect_logic.rb
@@ -4,8 +4,10 @@ module RuboCop
   module Cop
     # This module encapsulates the logic for autocorrect behavior for a cop.
     module AutocorrectLogic
-      def autocorrect?
-        autocorrect_requested? && correctable? && autocorrect_enabled?
+      # @param range [Parser::Source::Range] Used for bounding autocorrections
+      # @return [Boolean]
+      def autocorrect?(range = nil)
+        autocorrect_requested? && correctable? && autocorrect_enabled? && in_range?(range)
       end
 
       def autocorrect_with_disable_uncorrectable?
@@ -41,7 +43,36 @@ module RuboCop
         true
       end
 
+      def in_range?(range)
+        return true unless range
+
+        bounds = []
+
+        bounds << start_line_in_range?(range) if @options[:start_line]
+        bounds << start_column_in_range?(range) if @options[:start_column]
+        bounds << end_line_in_range?(range) if @options[:end_line]
+        bounds << end_column_in_range?(range) if @options[:end_column]
+
+        bounds.all?
+      end
+
       private
+
+      def start_line_in_range?(range)
+        range.line >= @options[:start_line]
+      end
+
+      def start_column_in_range?(range)
+        range.column >= @options[:start_column]
+      end
+
+      def end_line_in_range?(range)
+        range.last_line <= @options[:end_line]
+      end
+
+      def end_column_in_range?(range)
+        range.last_column <= @options[:end_column]
+      end
 
       def disable_offense(range)
         heredoc_range = surrounding_heredoc(range)
--- a/lib/rubocop/options.rb
+++ b/lib/rubocop/options.rb
@@ -129,7 +129,7 @@ module RuboCop
     # -a, --auto-correct         -            true          true               -
     #     --safe-auto-correct    -            true          true               -
     # -A, --auto-correct-all     -            true          -                  true
-    def add_autocorrection_options(opts) # rubocop:disable Metrics/MethodLength
+    def add_autocorrection_options(opts) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       section(opts, 'Autocorrection') do
         option(opts, '-a', '--autocorrect') { @options[:safe_autocorrect] = true }
         option(opts, '--auto-correct') do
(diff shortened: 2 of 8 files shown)


HeroProtagonist:
Might also resolve https://github.com/rubocop/rubocop/issues/10453?

lib/rubocop/cop/autocorrect_logic.rb
```diff
@@ -41,8 +41,37 @@ def autocorrect_enabled?
         true
       end
 
+      def inbounds?(range)
```
bbatsov, line 44: Won't `within_bounds?` or `(with)in_range?` read better?

lib/rubocop/options.rb
```diff
@@ -495,6 +499,14 @@ module OptionsHelp
       disable_uncorrectable:            ['Used with --autocorrect to annotate any',
                                          'offenses that do not support autocorrect',
                                          'with `rubocop:todo` comments.'],
+      first_line:                       ['Used with --autocorrect to specify',
```
bbatsov, line 502: I'm wondering if `start_line` and `end_line` won't be slightly better terminology here.

lib/rubocop/cop/autocorrect_logic.rb
```diff
@@ -4,8 +4,8 @@ module RuboCop
   module Cop
     # This module encapsulates the logic for autocorrect behavior for a cop.
     module AutocorrectLogic
-      def autocorrect?
-        autocorrect_requested? && correctable? && autocorrect_enabled?
+      def autocorrect?(range = nil)
```
bbatsov, line 7: As this is a public API it'd be good to document the purpose of `range`.

bbatsov:
@rafaelfranca tagging you here in case someone from Shopify is interested in providing feedback (as it's related to LSP).

HeroProtagonist:
@bbatsov Thanks for the feedback! Made the updates and also added a section to the docs

bbatsov:
@r7kamura Any feedback from you?

r7kamura:
Thank you @HeroProtagonist!

I am not sure if I can explain it well, but I will try to write my thoughts anyway.

Overall, I honestly think that my proposal at https://github.com/rubocop/rubocop/issues/10930 may have been too subtle. 

Of course it would be nice to have the ability to specify a range at the CLI or `RuboCop::Runner` API level. This is because this is probably the only way to realize a repetitive correction loop process while specifying a range in one instruction. But, on the other hand, I don't think there are that many use cases to require this feature. If it does not interfere with the growth of RuboCop, I would be happy to add it, but if this feature would make RuboCop more complicated, I think it would be better to give up adding this feature.


I am sorry for suggesting this, but after developing the editor extension since then, I realized that I could not provide a good editing experience if RuboCop was controlled from an external process via CLI. So I stopped implementing the Language Server in JavaScript and rewrote the Language Server from scratch in Ruby with using RuboCop's internal API.

- https://github.com/r7kamura/vscode-ruby-light
  - VSCode extension + Language Server implementation in JavaScript
  - Made in the early days
- https://github.com/r7kamura/rucoa
  - Language Server implementation in Ruby
  - Made later

This is the main reason why I have not implemented it until today, even though it was agreed upon in the Issue.

When implemented in Ruby, RuboCop's internal API can be used to retrieve corrected text from `RuboCop::Cop::Offense` with correction range information using `Parser::Source::TreeRewriter#as_replacements`. 

```ruby
offense.corrector.as_replacements
```

With this API, we can achieve most of what we want to do about range specification, by checking whether the range obtained from `offense` is within the range specified by the editor, limitation to only the range can be achieved, although not perfectly. Here is the code I actually use in my Language Server implementation:

- https://github.com/r7kamura/rucoa/blob/4ec510a66e0cd928c764712788d504800d2edebd/lib/rucoa/handlers/text_document_range_formatting_handler.rb#L57-L65
- https://github.com/whitequark/parser/blob/d0c86fe3d149a9cb6484ed094ffdf8595f2205cc/lib/parser/source/tree_rewriter.rb#L265-L283

In fact, other Ruby-made Lanauge Servers also use this method. For example, this is the case with ruby-lsp, which is developed by Shopify:

- https://github.com/Shopify/ruby-lsp/blob/d99eaf4107ede941cebdbf324f10525e8eacc67c/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb#L93

> although not perfectly

This may be an important topic for this Pull Request. The reason why I wrote it this way is because the logic just described above does not work perfectly in cases where one correction would create a new offense.

The advantage of the CLI option added in this Pull Request is that the iterative correction process using `RuboCop::Runner#do_inspection_loop` will be executed so that the correction is completed at once.

However, it is an occasional problem, and the cases where users want to specify a range are very rare (in the case of the Ruby-made Language Server, it is probably only used for `textDocument/rangeFormatting`), and even if it does occur, it can be resolved by re-running the formatting command several times.

That's why I said:

> I don't think there are that many use cases to require this feature.

at the beginning of this comment.

When I first tried to suggest this, I thought that [Quick Fix](https://user-images.githubusercontent.com/111689/185023861-e4bf6237-4cd6-4439-8497-b843dc3ea2c3.gif) would need this feature, but then I realized that the editor actually needs to fix things pretty fast and this feature would not work.

At the initial stage of diagnosing the offenses, the server needs to return the corrected text to the editor along with the offense, so that when the user presses the button, the correction is completed in an instant. So it is impractical to run autocorrect again when the button is pressed, and this CLI option cannot be used in this case.

---

I also have a question about the suggested implementation. If a new offense occurs as a result of autocorrection while specifying a range, how should it be fixed, and would the suggested implementation unintentionally modify code that was originally out of the range?

bbatsov:
Seems we won't be moving forward with this, so I'll close it.

PR #11080, opened 2022-10-16
[Fix #11079] Add new `Style/OperatorMethodCall` cop

### LINKED ISSUE #11079: Add a `Lint` cop to ask programmer to clarify `foo.&bar` into either safe navigation or bitwise and

Safe navigation uses the syntax `foo&.bar`.

However, when mistakenly writing `foo.&bar` instead, the Ruby parser happily accepts this.
After all:
```ruby
   foo.&bar
== foo.& bar
== foo.&(bar)
== foo & bar
```
That is, this is interpreted as usage of the 'bitwise and' operator. (Or an overloaded version of the binary `&` operator).

While the Ruby parser has to accept this syntax, it is highly unlikely that a user uses the particular way of writing `foo.&bar`
to mean `foo & bar`.
It is much more likely that the user meant to use the safe navigation `foo&.bar` syntax instead.

I would like a simple lint cop that warns when someone writes `foo.&bar`, and suggests the programmer to clarify:
- Whether they meant `foo&.bar`
- Or whether they meant `foo & bar`

I do not think that there are easy alternatives. 

We cannot alter what the Ruby parser itself does for this kind of situation, since it has to remain consistent with how method dispatching works.

## Extra context

One thing I am not sure about, is how to name this cop.

`Lint/AmbiguousBitwiseAnd` maybe? 

---

What do you think?
I would gladly contribute a PR that implements this cop :blush: .

koic:
Closes #11079.

Checks for redundant dot before operator method.
The target operator methods are `|`, `^`, `&`, `<=>`, `==`, `===`, `=~`, `>`, `>=`, `<`, `<=`, `<<`, `>>`, `+`, `-`, `*`, `/`, `%`, `**`, `~`, `!`, `!=`, and `!~`.

```ruby
# bad
foo.+ bar
foo.& bar

# good
foo + bar
foo & bar
```

It accepts cases where the dot cannot be omitted. e.g.:

```ruby
foo.+@ bar
foo.+ bar.to_s
```

-----------------

* [x] Added tests.


bbatsov:
I like the idea, but I'd prefer a name like `OperatorMethodCall` or something along those lines. 

One more thing - I assume we won't to treat `foo.+ bar` and `foo.+(baz)` the same, right?

koic:
`OperatorMethodCall` is a very good name! I've updated it.

> I assume we won't to treat `foo.+ bar` and `foo.+(baz)` the same, right? 

`foo.+ bar` and `foo.+(baz)` the same.
So, this cop will autocorrect `foo.+(baz)` to `foo + (baz)`. And redundant parentheses in `foo + (baz)` are left to `Style/RedundantParentheses` to fix.

bbatsov:
Thanks!

PR #11082, opened 2022-10-17
Fix an incorrect autocorrect for `Lint/SafeNavigationChain` when safenavigation on the right-hand side of the arithmetic operator

ydah:
This PR fixes an incorrect autocorrect when there is safe navigation on the right side of the arithmetic operator as follows:

```ruby
x + foo&.bar.baz
```

Expect autocorrect
```ruby
x + foo&.bar&.baz
```

Actual autocorrect
```ruby
x + foo&.bar&..baz
```

-----------------

* [ ] Added tests.


spec/rubocop/cop/lint/safe_navigation_chain_spec.rb
```diff
@@ -189,6 +189,19 @@
       RUBY
     end
 
+    %w[+ - * /].each do |operator|
```
koic, line 192: Can you separate each test case instead of the loop? Because the number is limited.

koic:
Thanks!

PR #11094, opened 2022-10-21
Add `InternalAffairs/ExampleHeredocDelimiter` cop

r7kamura:
In response to the following review comment, I have created a cop to consistently use `RUBY` delimiter for heredocs in cop tests.

- https://github.com/rubocop/rubocop-performance/pull/297#discussion_r930636473

cc: @koic 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/internal_affairs/example_heredoc_delimiter.rb
+++ b/lib/rubocop/cop/internal_affairs/example_heredoc_delimiter.rb
@@ -1,0 +1,111 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module InternalAffairs
+      # Use `RUBY` for heredoc delimiter of example Ruby code.
+      #
+      # Some editors may apply better syntax highlighting by using appropriate language names for
+      # the delimiter.
+      #
+      # @example
+      #  # bad
+      #  expect_offense(<<~CODE)
+      #    example_ruby_code
+      #  CODE
+      #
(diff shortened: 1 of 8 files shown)


lib/rubocop/cop/internal_affairs/example_heredoc_delimiter.rb
```diff
@@ -0,0 +1,106 @@
...
+  module Cop
+    module InternalAffairs
+      # Use `RUBY` for heredoc delimiter of example Ruby code.
+      #
+      # Some editors may apply better syntax highlighting by using appropriate language names for
+      # the delimiter.
+      #
+      # @safety
+      #   This cop is unsafe because the heredoc body may also contain a `RUBY` line,
+      #   in which case this word cannot be used for the delimiter.
```
koic, line 13: Since this is detectable, it should be able to be made safe. Can you add tests and adjust the implementation? An approach could be that this case is only detected and not auto-corrected.

r7kamura: That's right. Looking at some existing codes, there are cases where we really want to use a `RUBY` line in heredoc body, so I decided not to make such a case an offense.

koic:
Note: I'm going to merge this new cop because the cop is not for general users. Even if the next release is a bugfix release, it can be merged I think.

koic:
Thanks!

PR #11090, opened 2022-10-20
[Fix #11089] Fix an error for `Style/RedundantStringEscape` when using character literals (e.g. `?a`)

ydah:
Fix: #11089

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

Issue #11089, opened 2022-10-20
Style/RedundantStringEscape cop is crashing on some rare inputs


denis-kos:
The code as follows cause cop to crash:
```
["one", "two"] * ?.
# => "one.two"
```
--------
## Expected behavior
The cop should not crash. Rubocop probably can suggest to rewrite the code like `["one", "two"].join(".")`, but not crash.

## Actual behavior
`An error occurred while Style/RedundantStringEscape cop was inspecting ...`

## RuboCop version
```
$ [bundle exec] rubocop -V
1.37.0 (using Parser 3.1.2.1, rubocop-ast 1.22.0, running on ruby 3.1.1) [x86_64-linux]
  - rubocop-rails 2.15.2
  - rubocop-rspec 2.12.1
```

bbatsov:
@owst Please, take a look.

PR #11095, opened 2022-10-21
Fix an error for `Style/RedundantStringEscape` cop

koic:
This PR fixes the following error for `Style/RedundantStringEscape` cop when using `?\n` string character literal.

```console
% cat example.rb
?\n

% bundle exec rubocop --only Style/RedundantStringEscape -d
(snip)

An error occurred while Style/RedundantStringEscape cop was inspecting /Users/koic/src/github.com/koic/rubocop-issues/11090/example.rb:1:0.
undefined method `source' for nil:NilClass

          delimiters = [node.loc.begin.source[-1], node.loc.end.source[0]]
                                                               ^^^^^^^
/Users/koic/src/github.com/rubocop/rubocop/lib/rubocop/cop/style/redundant_string_escape.rb:162:in `delimiter?'
/Users/koic/src/github.com/rubocop/rubocop/lib/rubocop/cop/style/redundant_string_escape.rb:110:in `single_quoted?'
/Users/koic/src/github.com/rubocop/rubocop/lib/rubocop/cop/style/redundant_string_escape.rb:103:in `interpolation_not_enabled?'
/Users/koic/src/github.com/rubocop/rubocop/lib/rubocop/cop/style/redundant_string_escape.rb:85:in `allowed_escape?'
/Users/koic/src/github.com/rubocop/rubocop/lib/rubocop/cop/style/redundant_string_escape.rb:51:in `block in on_str'
```

Similar case to #11089, but not resolved by #11090. It also removes a bit of unnecessary implementation logic I found.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_string_escape.rb
+++ b/lib/rubocop/cop/style/redundant_string_escape.rb
@@ -42,10 +42,9 @@ module RuboCop
         MSG = 'Redundant escape of %<char>s inside string literal.'
 
         def on_str(node)
-          return if node.parent&.regexp_type? || node.parent&.xstr_type?
+          return if node.parent&.regexp_type? || node.parent&.xstr_type? || character_literal?(node)
 
           str_contents_range = str_contents_range(node)
-          return unless str_contents_range.source.include?('\\')
 
           each_match_range(str_contents_range, /(\\.)/) do |range|
             next if allowed_escape?(node, range.resize(3))
@@ -58,6 +57,10 @@ module RuboCop
 
         private
 
+        def character_literal?(node)
+          node.loc.respond_to?(:begin) && node.loc.begin && node.loc.begin.is?('?')
+        end
+
         def message(range)
           format(MSG, char: range.source.chars.last)
         end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/redundant_string_escape.rb
```diff
@@ -42,10 +42,9 @@ class RedundantStringEscape < Base
         MSG = 'Redundant escape of %<char>s inside string literal.'
 
         def on_str(node)
-          return if node.parent&.regexp_type? || node.parent&.xstr_type?
+          return if node.parent&.regexp_type? || node.parent&.xstr_type? || character_literal?(node)
```
koic, line 45: If https://github.com/rubocop/rubocop-ast/pull/242 becomes available, it can be rewritten as:

```suggestion
          return if node.parent&.regexp_type? || node.parent&.xstr_type? || node.character_literal?
```

bbatsov:
Thanks!

lib/rubocop/cop/style/redundant_string_escape.rb
bbatsov: That'd be nice - we have several checks for char literals in the codebase that can be simplified. And one cop that discourages them. :-)

PR #11097, opened 2022-10-21
Add new MatchCaseSensitive config in FileName cop

marctorrelles:
This adds a new config called `MatchCaseSensitive` in the `FileName` cop. It allows to have matching names without taking into account the sensitivity of the class name, as many eager loading methods allow it.

With `MatchCaseSensitive: true`, this is not allowed:
- filename: `us_collection.rb`
- classname: `USCollection`

With `MatchCaseSensitive: false`, this is allowed:
- filename: `us_collection.rb`
- classname: `USCollection`

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_add_new_match_case_sensitive_config.md
+++ b/changelog/new_add_new_match_case_sensitive_config.md
@@ -1,0 +1,1 @@
+* [#11097](https://github.com/rubocop/rubocop/pull/11097): Adds a new config called `MatchCaseSensitive` in the `FileName` cop, allowing to have non-sensitive file name matching. ([@marctorrelles][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2663,6 +2663,9 @@ Naming/FileName:
     - spec
     - test
     - src
+  # This allows to compare class names and module names insensitively. Many eager
+  # loading methods are able to match names with different case patterns.
+  MatchCaseSensitive: true
   # If non-`nil`, expect all source file names to match the following regex.
   # Only the file name itself is matched, not the entire file path.
   # Use anchors as necessary if you want to match the entire name rather than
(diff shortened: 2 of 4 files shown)


bbatsov:
Hmm, is using case-sensitive names even a thing in Ruby? I don't remember ever seeing filenames that weren't all lowercase.

bbatsov:
@rubocop/rubocop-core Any thoughts on this proposal? Are case-sensitive names a thing in Ruby-land?

bquorning:
> I don't remember ever seeing filenames that weren't all lowercase.

There is `Gemfile` and `Rakefile` to name a few.

bquorning:
But yes, what is the use case for having case sensitive matching at all?

bbatsov:
> There is Gemfile and Rakefile to name a few.

Yeah, but those are kind of special and generally the community has been trying to move away from them - e.g. newer versions of Bundler support also `gems.rb`.

Drenmi:
Sorry for the close-and-reopen. Fat fingers. 😅

marctorrelles:
@bbatsov @bquorning the issue is the other way around: filenames are usually in snake_case, but ClassNames may not be following the PascalCase convention. For instance:
```rb
# us_countries.rb
class USCountries
...
end
```
Following the rule, the class should be named as `UsCountries`, while many eager loading methods allow the first case.

bquorning:
> For instance:
> 
> ```ruby
> # us_countries.rb
> class USCountries
> ...
> end
> ```
> 
> Following the rule, the class should be named as `UsCountries`, while many eager loading methods allow the first case.

I still don’t understand. The added spec example passes against the current implementation (on master branch) when `cop_config` is not changed. I can only make it fail when configuring `ExpectMatchingDefinition` to true. And if doing that, I can add `US` to the `AllowedAcronyms` config. It seems to be the exact issue covered by this spec example: https://github.com/rubocop/rubocop/blob/6a5f8195cb61a6ae0b27d1afad94052f0d22ec62/spec/rubocop/cop/naming/file_name_spec.rb#L447-L460

bbatsov:

PR #11101, opened 2022-10-21
[Fix #10918] Add `<<` to `Style/StringConcatenation`

### LINKED ISSUE #10918: Add Style/StringConcatenation support for << with string literals?
`Style/StringConcatenation` cop detects string concatenation using `+` like so:

```ruby
'hello ' + name
# corrects to interpolation:
"hello #{name}"
```

It seems this cop could also detect concatenation using `<<` in a similar way if the left-hand operand is a literal:
```ruby
'hello ' << name
# would also correct to:
"hello #{name}"
```

HeroProtagonist:
Closes #10918

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_add_node_to_stylestringconcatenation.md
+++ b/changelog/new_add_node_to_stylestringconcatenation.md
@@ -1,0 +1,1 @@
+* [#10918](https://github.com/rubocop/rubocop/issues/10918): Add support for `<<` node to `Style/StringConcatenation`. ([@HeroProtagonist][])
--- a/lib/rubocop/cop/style/string_concatenation.rb
+++ b/lib/rubocop/cop/style/string_concatenation.rb
@@ -17,9 +17,9 @@ module RuboCop
       #
       # Two modes are supported:
       # 1. `aggressive` style checks and corrects all occurrences of `+` where
-      # either the left or right side of `+` is a string literal.
+      # either the left or right side of `+` or `<<` is a string literal.
       # 2. `conservative` style on the other hand, checks and corrects only if
-      # left side (receiver of `+` method call) is a string literal.
+      # left side (receiver of `+` or `<<` method call) is a string literal.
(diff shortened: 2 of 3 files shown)


HeroProtagonist:
Causes internal lints. Perhaps the cop should have this added as a config option with default false?

bbatsov:
Hmm, I thought that `+` and `<<`/`concat` had different semantics (new string vs modifying the receiver), but I guess this is OK for the current cop. 

> Causes internal lints. Perhaps the cop should have this added as a config option with default false?

It's fine to also update the existing code.

bbatsov:
I wonder if we should also handle `concat` is some way, but probably that'd be an overkill.

HeroProtagonist:
I realized that the rule is only valid if the receiver is a string. Right now I made it check both left and right hand sides which is wrong. When I correct this, the internal lints will no longer be flagged.


```ruby
[] << "foo" # not string concatenation

"#{[]}foo" # currently would yield an invalid correction
```

The issue mentioned this, but I missed it 🤦  
> It seems this cop could also detect concatenation using << in a similar way if the left-hand operand is a literal

vlad-pisanov:
@HeroProtagonist thanks for taking a look at this, but now I realize this substitution would be unsafe. 😞 Even if the receiver is a string literal, `<<` cannot be replaced with interpolation if the operand is a number:

```ruby
op = 65

"hello " << op    # "hello A"
"hello #{op}"     # "hello 65"
```

`<<` turns numbers into the corresponding ascii character, while interpolation turns them into decimal representation

bbatsov:
So, what are we doing with this one?

HeroProtagonist:
I don't think this should be added as it only applies in a few cases. Thoughts @vlad-pisanov?

bbatsov:
I'll take the lack of activity here as an indication we've decided to abandon this.

PR #11105, opened 2022-10-21
[Fix #11098] Fix false positive for Style/RedundantStringEscape

### LINKED ISSUE #11098: Bad autocorrect for Style/RedundantStringEscape
The following string is auto-corrected incorrectly by Style/RedundantStringEscape:

```
"#\{foo\}"
```

## Expected behavior

The second escape is not necessary, but the first one is. So one good correction is:

```
"#\{foo}"
```

It seems to be more common to escape the `#` instead, so an alternative good correction is:

```
"\#{foo}"
```

## Actual behavior

The string is corrected to the following, which is definitely not correct:

```
"#{foo}"
```

- Create a file `foo.rb` containing

```
"#\{foo\}"
```

- Run `rubocop -a foo.rb --only Style/RedundantStringEscape`.

## RuboCop version

```
$ [bundle exec] rubocop -V
1.37.0 (using Parser 3.1.2.1, rubocop-ast 1.22.0, running on ruby 3.1.2) [x86_64-linux]
```

tdeo:
Fix incorrect false-positives for escaped interpolation when escaping the character following the `#` instead of the `#` itself.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/redundant_string_escape_spec.rb
+++ b/spec/rubocop/cop/style/redundant_string_escape_spec.rb
@@ -21,20 +21,40 @@ RSpec.describe RuboCop::Cop::Style::RedundantStringEscape, :config do
       expect_no_offenses(wrap('\#$foo'))
     end
 
+    it 'does not register an offense for a $-escaped gvar interpolation' do
+      expect_no_offenses(wrap('#\$foo'))
+    end
+
     it 'does not register an offense for an escaped ivar interpolation' do
       expect_no_offenses(wrap('\#@foo'))
+    end
+
+    it 'does not register an offense for a @-escaped ivar interpolation' do
+      expect_no_offenses(wrap('#\@foo'))
     end
 
     it 'does not register an offense for an escaped cvar interpolation' do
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/redundant_string_escape_spec.rb
```diff
@@ -21,22 +21,42 @@ def wrap(contents)
...
+    it 'does not register an offense for a @-escaped ivar interpolation' do
+      expect_no_offenses(wrap('#\@foo'))
+    end
+
     it 'does not register an offense for an escaped cvar interpolation' do
       expect_no_offenses(wrap('\#@@foo'))
     end
 
+    it 'does not register an offense for a @-escaped cvar interpolation' do
+      expect_no_offenses(wrap('\#@@foo'))
```
koic, line 41: It looks the same as the L37.

koic:
It would be better if `expect_offense` and `expect_correction` is written about the following autocorrection.

```diff
-"#\{foo\}"
+"#\{foo}
```

bbatsov:
Looks good. Thanks!

PR #11108, opened 2022-10-22
[Fix #11107] Fix a false positive for `Style/OperatorMethodCall`

### LINKED ISSUE #11107: `Style/OperatorMethodCall` incorrectly flags module methods
`Style/OperatorMethodCall` incorrectly flags cases where the call is from a module or a class method, rather than the instance method that is normally associated with the operator.

E.g. `Sequel.-` produces an object that is used for generating an SQL query, and does not perform the operator, while `foo - bar` calls the instance method `foo#-`.

--------

## Expected behavior

This cop should be considered as unsafe for auto-correction, or should ignore non-instance method cases.

## Actual behavior

`Sequel.-` was autocorrected to `Sequel -`, which then produced errors on load as the Ruby syntax is interpreted as invalid.

Use a module or class method for an operator, and have the corrector modify this to an invalid syntax.

In our case, `Sequel.-(foo, bar)` was converted to `Sequel -(foo, bar)` which broke the Ruby interpreter when trying to parse that code.

## RuboCop version

```
$ [bundle exec] rubocop -V
1.37.0 (using Parser 3.1.2.1, rubocop-ast 1.23.0, running on ruby 3.1.2) [arm64-darwin21]
  - rubocop-performance 1.15.0
  - rubocop-rails 2.17.0
  - rubocop-sorbet 0.6.11
```

koic:
Fixes #11107.

This PR fixes a false positive for `Style/OperatorMethodCall` when a constant receiver uses an operator method.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_style_operator_method_call.md
+++ b/changelog/fix_a_false_positive_for_style_operator_method_call.md
@@ -1,0 +1,1 @@
+* [#11107](https://github.com/rubocop/rubocop/issues/11107): Fix a false positive for `Style/OperatorMethodCall` when a constant receiver uses an operator method. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #11110, opened 2022-10-22
Add new `Style/RedundantEach` cop

koic:
Follow up https://github.com/rubocop/rubocop-rails/pull/812/files#r994033301.

This PR adds new `Style/RedundantEach` cop that checks for redundant `each`.

This cop is unsafe, as it can produce false positives if the receiver is not an `Enumerable`.

```ruby
# bad
array.each.each { |v| do_something(v) }

# good
array.each { |v| do_something(v) }

# bad
array.each.each_with_index { |v, i| do_something(v, i) }

# good
array.each.with_index { |v, i| do_something(v, i) }

# bad
array.each.each_with_object { |v, o| do_something(v, o) }

# good
array.each.with_object { |v, o| do_something(v, o) }
```

-----------------

* [x] Added tests.


bbatsov:
I like the suggestion, but I'm wondering if this is not more of a Lint cop.

lib/rubocop/cop/style/redundant_each.rb
```diff
@@ -0,0 +1,109 @@
...
+      #   array.each.each { |v| do_something(v) }
+      #
+      #   # good
+      #   array.each { |v| do_something(v) }
+      #
+      #   # bad
+      #   array.each.each_with_index { |v, i| do_something(v, i) }
+      #
+      #   # good
+      #   array.each.with_index { |v, i| do_something(v, i) }
```
bbatsov, line 24: I'm guessing that using `each_with_index` directly is also good. Same with `each_with_object`. Probably we'll need some separate cop about such iterators (e.g. something suggesting `each.with_object` over `each_ with_object`).

koic: Sure,`each_with_index` is also accepted. I've updated it!
OTOH, I did a quick check and saw no difference in performance. This autocorrection chooses `each.with_object` because the receiver is an `Enumerator` and using method chain. But, I'm not sure which of `each_with_object` and `each.with_object` is definitely better as the default when receiver is a collection object (e.g. instance of Array, Hash, Set).

koic:
> I like the suggestion, but I'm wondering if this is not more of a Lint cop.

I was a little worried, but it's unsafe, so I chose Style over Lint. It's also because the original code is not a dangerous problem. For example, `reverse_each.each_with_index` would be a favorite style to eliminate the redundant expression `each` rather than a definite mistake.

bbatsov:
Thanks!

Issue #10080, opened 2021-09-14
`Lint/AmbiguousOperatorPrecedence` configuration


pyromaniac:
## Is your feature request related to a problem? Please describe.

The new `Lint/AmbiguousOperatorPrecedence` cop looks way like an overkill. Parenthesis in arithmetic expressions for all the operations create a lot of visual noise and might be not a god option for a lot of developers. I believe, it is going to be disabled by this reason a lot. 

But for boolean expressions additional parenthesis could bring a lot of value.

## Describe the solution you'd like

I believe we need to split operators into groups. I see two right now: boolean and arithmetic. Would be great to be able enabling them separately.

## Describe alternatives you've considered

The only alternative for now is to disable the cop completely.

dvandersluis:
Thank you for your feedback. The idea is while the cop is pending to gauge how well it works for RuboCop users and tweak as necessary.

The cop is intended to make precedence very explicit, even if you're not very familiar with Ruby's order of operations, and as such it may add more parentheses that you might expect, or that seem excessive even, depending on what your original code was. That might also be an indication that there's a useful refactoring there.

Being able to disable groups as you suggested is interesting, but has some pitfalls. First off, operations can be overridden, so `+` isn't necessarily arithmetic, for example. The shift operators (`<<` and `>>`) are a good example of how operators that are commonly overridden to mean different things. Secondly, I'm not sure that the other bitwise operators (`&`, `|`, `^`) cleanly fit into "arithmetic" either.

@bbatsov what do you think?

bbatsov:
I've also been thinking about some form of configuration, but I don't think that the idea about groups will work in practice. We can probably have a list of operators on which cop would apply that could narrow its scope, but I'm not eager to introduce this until we've gathered more feedback from users.

januszm:
it reports false positives, e.g.:

```
W: [Correctable] Lint/AmbiguousOperatorPrecedence: Wrap expressions with varying precedence with parentheses to avoid ambiguity.

...
something =  31 * 24 + 10 * 10
```

there is no ambiguity in this example.

bbatsov:
@januszm There's never any ambiguity with respect to operator precedence for Ruby itself. We used the word in the sense that there's potential ambiguity for the readers of the code (at the very least they have to read the code more carefully). Parentheses eliminate this.

bendilley:
[...]

PR #11111, opened 2022-10-22
Rename `Lint/AmbiguousOperatorPrecedence` to `Style/ExplicitOperatorPrecedence`

koic:
Follow up https://github.com/rubocop/rubocop/issues/10080#issuecomment-1006148714.

This PR renames `Lint/AmbiguousOperatorPrecedence` cop from `Style/ExplicitOperatorPrecedence`.

The above link comment would make a lot of sense. And other feedback on the issue has also been commented the cop shouldn't be at a particularly high "severity" level like `Lint`. I think it could be considered disabled by default, but this PR moves to `Style` instead of `Lint` anyway.

I understand that it shouldn't breaking change until 2.0, but I believe this is worth the breaking change for early adopters using pending cop.

As a side note, I think this should not be enforced by source code, as it can be completed in the editor if desired.
https://twitter.com/kddnewton/status/1485712576076042241

-----------------

* [x] Added tests.


bbatsov:
> I understand that it shouldn't breaking change until 2.0, but I believe this is worth the breaking change for early adopters using pending cop.

I'll have to think about this, but I'm open to the suggestion in general. Plus I think we should start focusing our attention on RuboCop 2.0 soon anyways, as a lot of breaking changes have piled up in the backlog. 

I also think that it's important for to start working on the "cop squads" or however we decide to name them in the end, as I think that allowing people to run some smaller more focused squads will solve much of such issues anyways.

pirj:
What I'm most interested regarding the inheritance approach is if would offences be displayed for both the old and the new cop?

It was one of the decision-making factors for [`rubocop-capybara` extraction](https://github.com/rubocop/rubocop-rspec/pull/1519) to avoid duplicates, see [this](https://github.com/rubocop/rubocop-rspec/pull/1519#issuecomment-1374883512).

pirj:
Just to avoid any confusion, redundant offences, not warnings.
This is caused by both cops being enlisted.

```
$ rubocop spec/cap_spec.rb
Inspecting 1 file
C

Offenses:

spec/cap_spec.rb:3:5: C: Capybara/SpecificActions: Prefer click_link over find('a').click.
    find('a[href]').click
    ^^^^^^^^^^^^^^^^^^^^^
spec/cap_spec.rb:3:5: C: RSpec/Capybara/SpecificActions: Prefer click_link over find('a').click.
    find('a[href]').click
    ^^^^^^^^^^^^^^^^^^^^^
```

Contrary to the inheritance approach, `rename` in obsoletion config tells RuboCop to fall back from the old cop to the new one, and there is no redundancy in the output. At a cost of some extra code for cops doc generation.

koic:
Yeah. It's offense, not warning. I was using the wrong terminology 😅

PR #11112, opened 2022-10-23
[Fix #11066] -- Update Docs for `Style/CollectionCompact`

### LINKED ISSUE #11066: Collection/Compact on ActionController::Parameters 6.0
I'm seeing this false positive. 
```
lib/blacklight/parameters.rb:14:14: C: [Correctable] Style/CollectionCompact: Use compact instead of reject { |_k, v| v.nil? }.
      params.reject { |_k, v| v.nil? }
             ^^^^^^^^^^^^^^^^^^^^^^^^^
```
When I make this change it fails because ActionController::Parameters 6.0 does not have a compact method.  It doesn't seem like one can infer that a `compact` method exists on an object just because a `.reject` method exists.

PolyCole:
- This PR fixes #11066. 
- Essentially, there's a situation that can happen where Rubocop says that `params.reject { |n| ... }` should be replaced with `params.compact`. However, the params object doesn't have an implementation of `compact`. 
- Since `Style/CollectionCompact` is already marked unsafe, the solution here was to update the associated documentation.

-----------------

* [x] Added tests.
    * Didn't actually add tests seeing as this is a documentation change, but if y'all have some way to run unit tests against docs lemme know and I'm happy to get that added 😁 
   * `"If the new code introduces user-observable changes"` -- I don't believe these changes are user-observable, but am more than happy to add this if it's necessary.

### DIFF
--- a/lib/rubocop/cop/style/collection_compact.rb
+++ b/lib/rubocop/cop/style/collection_compact.rb
@@ -8,7 +8,8 @@ module RuboCop
       #
       # @safety
       #   It is unsafe by default because false positives may occur in the
-      #   `nil` check of block arguments to the receiver object.
+      #   `nil` check of block arguments to the receiver object. Additionally,
+      #    .compact may not exist for the highlighted object.
       #
       #   For example, `[[1, 2], [3, nil]].reject { |first, second| second.nil? }`
       #   and `[[1, 2], [3, nil]].compact` are not compatible. This will work fine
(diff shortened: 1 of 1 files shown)


lib/rubocop/cop/style/collection_compact.rb
```diff
@@ -8,7 +8,8 @@ module Style
       #
       # @safety
       #   It is unsafe by default because false positives may occur in the
-      #   `nil` check of block arguments to the receiver object.
+      #   `nil` check of block arguments to the receiver object. Additionally,
+      #    .compact may not exist for the highlighted object.
```
bbatsov, line 12: I'd say something like "we can't know for certain the type of the receiver object, which may result in false positives as well".

PolyCole: Will do! Thanks for the quick response @bbatsov, I appreciate it :)

PR #11113, opened 2022-10-23
Report the count of files in the Worst and Offense Count formatters

hosamaly:
Report the count of files in the Worst and the Offence Counts formatters. Their output now looks like this:

```
# Worst:
4  exe/rubocop
3  spec/spec_helper.rb
2  lib/rubocop.rb
--
9  Total in 3 files

# Offenses:
2  CopC
1  CopA
1  CopB
--
4  Total in 1 files
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/formatter/worst_offenders_formatter.rb
+++ b/lib/rubocop/formatter/worst_offenders_formatter.rb
@@ -12,7 +12,7 @@ module RuboCop
     # 26  this/file/is/really/bad.rb
     # 3   just/ok.rb
     # --
-    # 29  Total
+    # 29  Total in 2 files
     class WorstOffendersFormatter < BaseFormatter
       attr_reader :offense_counts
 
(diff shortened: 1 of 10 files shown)


lib/rubocop/formatter/worst_offenders_formatter.rb
```diff
@@ -36,14 +36,17 @@ def finished(_inspected_files)
...
 
+        count_width = total_count.to_s.length + 2
         per_file_counts.each do |file_name, count|
-          output.puts "#{count.to_s.ljust(total_count.to_s.length + 2)}#{file_name}\n"
+          output.puts "#{count.to_s.ljust(count_width)}#{file_name}\n"
         end
+
         output.puts '--'
-        output.puts "#{total_count}  Total"
+        output.puts "#{total_count}  Total offenses in #{file_count} offending files"
```
bbatsov, line 49: I think the word "offending" is redundant here, given the context.

hosamaly: I had written it initially without "offending", but it felt a bit ambiguous: was this the total number of processed files, or the number of files with offenses?

I'm not worried too much, though, so I'm ok either way. What do you think?

bbatsov: I think here it should be quiet clear from the context, so probably less is more. Even the word "offenses" is kind of redundant IMO.

hosamaly: I updated it to show `3  Total in 2 files`.

bbatsov:
Thanks!

PR #11118, opened 2022-10-24
[Fix #11117] Fix a false positive for `Style/BlockDelimiters`

### LINKED ISSUE #11117: Style/BlockDelimiters forces do-end style even if it were one liner
I found a bug related to Style/BlockDelimters

--------

## Expected behavior
I hoped Rubocop didn't offend the code below.
```ruby
def method
  ary&.map { |e| some_func(e) }&.find { |e| some_func2(e) }
end
```

## Actual behavior

But actually, it's overwritten by RuboCop like below.
```ruby
def method
  ary&.map do |e| some_func(e) end&.find { |e| some_func2(e) }
end
```

And `rubocop --debug` said 
```
For /current/directory: configuration from /path/to/.rubocop.yml
Use parallel by default.
```

1. Set `Style/BlockDelimiters` EnforcedStyle: semantic
2. Chain some methods with at least one safe navigation operator
  Probably, it should be in a method. On the top level, it seems another kind of bug occurs against method chain. (It's under investigation now)
3. Run RuboCop
4. All pairs of `{}` in the chain except last one will be offended

## RuboCop version

```
$ [bundle exec] rubocop -V
1.37.1 (using Parser 3.1.2.1, rubocop-ast 1.23.0, running on ruby 3.1.2) [arm64-darwin21]
```

koic:
Fixes #11117.

This PR fixes a false positive for `Style/BlockDelimiters` when specifying `EnforcedStyle: semantic` and using a single line block with {} followed by a safe navigation method call.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_block_delimiters.md
+++ b/changelog/fix_a_false_positive_for_block_delimiters.md
@@ -1,0 +1,1 @@
+* [#11117](https://github.com/rubocop/rubocop/issues/11117): Fix a false positive for `Style/BlockDelimiters` when specifying `EnforcedStyle: semantic` and using a single line block with {} followed by a safe navigation method call. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #11121, opened 2022-10-25
[Fix #11120] Fix an incorrect autocorrect for `Lint/RedundantRequireStatement`

### LINKED ISSUE #11120: `Lint/RedundantRequireStatement` breaks code with modifiers
```ruby
require 'pp' unless respond_to?(:pretty_inspect)
```

Autocorrect breaks code when `require` is followed by postfix `unless`.

--------

## Expected behavior

Autocorrects into valid Ruby code, possibly like below (considering the modifier condition may have side effects.)

```ruby
respond_to?(:pretty_inspect)
```

## Actual behavior

```
Offenses:

main.rb:1:1: W: [Correctable] Lint/RedundantRequireStatement: Remove unnecessary require statement.
require 'pp' unless respond_to?(:pretty_inspect)
^^^^^^^^^^^^

```

effectively autocorrects code into

```ruby
unless respond_to?(:pretty_inspect)
```

, which is syntactically incorrect.

```
$ cat main.rb
require 'pp' unless respond_to?(:pretty_inspect)
$ rubocop --only Lint/RedundantRequireStatement
```

## RuboCop version

```
$ rubocop -V
1.37.1 (using Parser 3.1.2.1, rubocop-ast 1.23.0, running on ruby 3.1.1) [x86_64-darwin20]
$ 
```

koic:
Fixes #11120.

This PR fixes an incorrect autocorrect for `Lint/RedundantRequireStatement` when using redundant `require` with modifier form.

For example, in essential it is unknown how many times `do_something` will be executed when `require 'enumerator' while do_something`. So it does not autocorrects when modifier form is used to avoid side effects.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/redundant_require_statement.rb
+++ b/lib/rubocop/cop/lint/redundant_require_statement.rb
@@ -49,7 +49,13 @@ module RuboCop
           return unless redundant_require_statement?(node)
 
           add_offense(node) do |corrector|
-            range = range_with_surrounding_space(node.loc.expression, side: :right)
+            if node.parent.respond_to?(:modifier_form?) && node.parent.modifier_form?
+              corrector.insert_after(node.parent, "\nend")
+
+              range = range_with_surrounding_space(node.loc.expression, side: :right)
+            else
+              range = range_by_whole_lines(node.source_range, include_final_newline: true)
+            end
 
             corrector.remove(range)
           end
--- a/changelog/fix_an_incorrect_autocorrect_for_lint_redundant_require_statement.md
+++ b/changelog/fix_an_incorrect_autocorrect_for_lint_redundant_require_statement.md
@@ -1,0 +1,1 @@
+* [#11120](https://github.com/rubocop/rubocop/issues/11120): Fix an incorrect autocorrect for `Lint/RedundantRequireStatement` when using redundant `require` with modifier form. ([@koic][])
--- a/spec/rubocop/cop/lint/redundant_require_statement_spec.rb
+++ b/spec/rubocop/cop/lint/redundant_require_statement_spec.rb
@@ -13,6 +13,36 @@ RSpec.describe RuboCop::Cop::Lint::RedundantRequireStatement, :config do
     RUBY
   end
 
+  it 'registers an offense when using requiring `enumerator` with modifier form' do
+    expect_offense(<<~RUBY)
+      require 'enumerator' if condition
+      ^^^^^^^^^^^^^^^^^^^^ Remove unnecessary `require` statement.
+      require 'uri'
+    RUBY
+
+    expect_correction(<<~RUBY)
+      if condition
+      end
+      require 'uri'
+    RUBY
+  end
+
+  it 'registers an offense when using requiring `enumerator` in condition' do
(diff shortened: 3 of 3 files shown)


bbatsov:
> For example, in essential it is unknown how many times do_something will be executed when require 'enumerator' while do_something. So it does not autocorrects when modifier form is used to avoid side effects.

I think that's not the right approach and we should just kill the conditionals directly if they are `if` or `unless`.  I can't imagine who'd be writing a conditional require with `while` and `until` but, for them I agree we should not attempt to auto-correct them.

lib/rubocop/cop/lint/redundant_require_statement.rb
```diff
@@ -48,6 +48,8 @@ def on_send(node)
           return unless redundant_require_statement?(node)
 
           add_offense(node) do |corrector|
+            next if node.parent.respond_to?(:modifier_form?) && node.parent.modifier_form?
```
Drenmi, line 51: If we're confident no-one will do `require "foo" until x` maybe we can be more specific and check if `parent` is an `if_type?`. WDYT? 🙂

Drenmi (approved):

lib/rubocop/cop/lint/redundant_require_statement.rb
bbatsov: We probably need a lint about `require... until` :D That's some wild code! :D

Drenmi: Needless to say I have never seen it in the wild. I was trying to dream up some use case for it, but can honestly say I failed to even imagine one. 😅

bbatsov: Same here! 😆

tsugimoto:
Should this cop eagerly propose removing the conditionals along (works ~~almost~~ every real world case but theoretically:tm: turns this cop unsafe), or stay away from modifiers and be able to safely autocorrect `require`s without modifiers?

I see this PR is taking the safe side, which makes enough sense.

koic:
Yeah, this PR is on the safe side. As has been discussed, `require` with modifier condition is a corner case. So, I think leaving it to user's decision without autocorrection does not reduce usability.

bbatsov:
Btw, won't we have exactly the same problem if the parent note is some conditional or a loop? E.g.

```ruby
if foo
  require 'bar'
end
```

This won't result in broken code from a syntax perspective, but it's essentially the same as the modifier form, so I wonder if we should treat those differently.

koic:
Sure. I've tried to focus only removing redundant `require` statement for autocorrection. So, I've updated this PR to add `end` keyword for prevent the syntax error:

```ruby
it 'registers an offense when using requiring `enumerator` with modifier form' do
  expect_offense(<<~RUBY)
    require 'enumerator' if condition
    ^^^^^^^^^^^^^^^^^^^^ Remove unnecessary `require` statement.
    require 'uri'
  RUBY

  expect_correction(<<~RUBY)
    if condition
    end
    require 'uri'
  RUBY
end
```

This would make modifier form and normal form compatible. How about it?

bbatsov:
I like it more. Thanks!

PR #11032, opened 2022-09-30
Detect empty Hash literal braces containing only newlines and spaces on `Layout/SpaceInsideHashLiteralBraces`

r7kamura:
I changed `Layout/SpaceInsideHashLiteralBraces` so that it can detect the following pattern as empty Hash literal braces:

```ruby
# bad
foo = {
}

# good
foo = {}
```

One reason I think this should be is that `Layout/SpaceInsideArrayLiteralBrackets` already bahaves like this.

```ruby
# bad
foo = [
]

# good
foo = []
```

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #11130, opened 2022-10-27
Check blank percent literal by `Layout/SpaceInsidePercentLiteralDelimiters`

r7kamura:
Previously, no particular investigation was made for whitespace in blank percent array literals on `Layout/SpaceInsideArrayPercentLiteral`.

I will change it in this Pull Request so that they are inspected as well.

```ruby
# bad
%w[ ]
%w[    ]
%w[
]

# good
%w[]
```

This is a similar change to:

- https://github.com/rubocop/rubocop/pull/11032

---

edit: At first I tried to change `Layout/SpaceInsideArrayPercentLiteral` Cop, but after discussion, it was decided to use `Layout/SpaceInsidePercentLiteralDelimiters` Cop instead.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_check_blank_percent_literal.md
+++ b/changelog/change_check_blank_percent_literal.md
@@ -1,0 +1,1 @@
+* [#11130](https://github.com/rubocop/rubocop/pull/11130): Check blank percent literal by `Layout/SpaceInsidePercentLiteralDelimiters`. ([@r7kamura][])
--- a/lib/rubocop/cop/layout/space_inside_array_percent_literal.rb
+++ b/lib/rubocop/cop/layout/space_inside_array_percent_literal.rb
@@ -6,6 +6,9 @@ module RuboCop
       # Checks for unnecessary additional spaces inside array percent literals
       # (i.e. %i/%w).
       #
+      # Note that blank percent literals (e.g. `%i( )`) are checked by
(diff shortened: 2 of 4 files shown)


r7kamura:
I am wondering if we should offer several styles or if they are unnecessary at this time.

r7kamura:
Perhaps it would be better to implement this in `Layout/SpaceInsidePercentLiteralDelimiters` cop?

bbatsov:
> I am wondering if we should offer several styles or if they are unnecessary at this time.

I don't think so. I haven't seen much usages of spaces in empty literals in general. 

> Perhaps it would be better to implement this in Layout/SpaceInsidePercentLiteralDelimiters cop?

Well, it can't be an alternative really - it'd be complimentary to the change here.

bbatsov:
>  Perhaps it would be better to implement this in Layout/SpaceInsidePercentLiteralDelimiters cop?

My bad - I misread the cop names. 🤦 I agree that would actually be better, as it's more universal. I'd probably mention in the array literals cop that this additional aspect is control by the other cop, so it's clear we're consistent across the board with respect to the empty literals.

bbatsov:
There it would also make more sense to have some configuration option like `EnforcedStyleForEmptyLiteral`.

bbatsov:
Thanks for tackling this!

PR #11134, opened 2022-10-30
Escape cell delimiters in regexp on cops naming page

mojavelinux:
To use a cell delimiter in a table cell as a literal character, it must always be escaped.

### DIFF
--- a/docs/modules/ROOT/pages/cops_naming.adoc
+++ b/docs/modules/ROOT/pages/cops_naming.adoc
@@ -591,7 +591,7 @@ EOS
 | Name | Default value | Configurable values
 
 | ForbiddenDelimiters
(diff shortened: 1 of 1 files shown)


bbatsov:
Hmm, that's going to be tricky - the docs about config are auto-generated from the yaml configuration file, but if we change this there it would break the regexp.

mojavelinux:
Another option you have is to change the table cell delimiter/separator to something that won't conflict or use an attribute reference. See https://docs.asciidoctor.org/asciidoc/latest/tables/data-format/#escape-the-cell-separator

koic:
Thanks for the information. I've opened https://github.com/rubocop/rubocop/pull/11135 to update documentation generator with `[separator=¦]`.

bbatsov:
Superseded by #11135.

PR #11142, opened 2022-11-01
Fix `Style/RedundantEach` for non-chained `each_` calls

fatkodima:
`Style/RedundantEach` fails for `[1, 2, 3].each` with no method `block_type?` for nil.

And incorrectly detects as offense the following:
```ruby
build_enumerator(outer: [[1, 2, 3].each]).each
```


RDeckard:
Also, Shouldn't the documentation be corrected in this way:
> This cop is unsafe, as it can produce false positives if the receiver is not an ~`Enumerable`~ `Enumerator`.

koic:
Thanks!

koic:
@RDeckard Good catch. I've fixed that typo in the master.

PR #11152, opened 2022-11-04
[Fix #11150] Catch redundant hyphen escape within character class

### LINKED ISSUE #11150: Style/RedundantRegexpEscape should register and offence for escaped hypens in character groups when a hypen is the first/last character

Consider the following regexes:

```ruby
unescaped1 = /[0-9-]/
escaped1   = /[0-9\-]/

unescaped2 = /[-0-9]/
escaped2   = /[\-0-9]/
```

All 4 match numbers 0 through 9 and `-`. It is not neccesary to escape `-` when it is the first or last character in the group since there is no ambiguity about it. Either the user mistyped or it is intended like this. Regardless of that, the backslash is redundant.

Rubocop should register an offence for escaped hypens in character groups when the hypen is either the first or last character.

si-lens:
Fix: #11150 
It should catch redundant regexp escape when the hyphen is being escaped and at the same time is the first or the last character within the character class.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_regexp_escape.rb
+++ b/lib/rubocop/cop/style/redundant_regexp_escape.rb
@@ -44,7 +44,7 @@ module RuboCop
 
         def on_regexp(node)
           each_escape(node) do |char, index, within_character_class|
-            next if allowed_escape?(node, char, within_character_class)
+            next if allowed_escape?(node, char, index, within_character_class)
 
             location = escape_range_at_index(node, index)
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/redundant_regexp_escape.rb
```diff
@@ -65,12 +65,18 @@ def allowed_escape?(node, char, within_character_class)
...
-            ALLOWED_WITHIN_CHAR_CLASS_METACHAR_ESCAPES.include?(char)
+
+            ALLOWED_WITHIN_CHAR_CLASS_METACHAR_ESCAPES.include?(char) &&
+              !char_class_begins_or_ends_with_hypen?(node, index)
           else
             ALLOWED_OUTSIDE_CHAR_CLASS_METACHAR_ESCAPES.include?(char)
           end
         end
 
+        def char_class_begins_or_ends_with_hypen?(node, index)
```
bbatsov, line 76: It's not very clear what this function does, so perhaps it'd better to document it a bit.

bbatsov (approved):
Thanks!

PR #11163, opened 2022-11-08
Mark `Style/HashExcept` as unsafe

r7kamura:
I think it would be better to assume that `Style/HashExcept` is unsafe because it detects and replaces even things that should not be replaced.

For example:

```ruby
invalid_values = [3, 5]

[1, 2, 3, 4, 5, 6, 7, 8, 9, 10].each_slice(2).reject do |value1, _value2|
  invalid_values.include?(value1)
end
#=> [[1, 2], [7, 8], [9, 10]]
```

The above code is currently auto-corrected as follows:

```ruby
invalid_values = [3, 5]

[1, 2, 3, 4, 5, 6, 7, 8, 9, 10].each_slice(2).except(*invalid_values)
#=> undefined method `except' for #<Enumerator: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]:each_slice(2)> (NoMethodError)
```

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/hash_except.rb
+++ b/lib/rubocop/cop/style/hash_except.rb
@@ -13,6 +13,10 @@ module RuboCop
       # when used `==`.
       # And do not check `Hash#delete_if` and `Hash#keep_if` to change receiver object.
       #
+      # @safety
+      #   This cop is unsafe because it cannot be guaranteed that the receiver
+      #   is a `Hash` or responds to the replacement method.
+      #
       # @example
       #
       #   # bad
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/hash_except.rb
```diff
@@ -13,6 +13,9 @@ module Style
       # when used `==`.
       # And do not check `Hash#delete_if` and `Hash#keep_if` to change receiver object.
       #
+      # @safety
+      #   This cop is unsafe because the receiver may not be a Hash.
```
koic, line 17: Can you tweak it?

```suggestion
      #   This cop is unsafe because it cannot be guaranteed that the receiver
      #   is a `Hash` or responds to the replacement method.
```

r7kamura: Thanks, that's much better 👍

koic:
Thanks!

PR #11170, opened 2022-11-09
[Fix #11168] Fix an incorrect autocorrect for `Style/ClassEqualityComparison`

### LINKED ISSUE #11168: Style/ClassEqualityComparison is overly aggressive
## Expected behavior

1. Rubocop doesn't suggest using `#instance_of?` with an arguments whose type it doesn't know.
2. Rubocop doesn't automatically "correct" code to `#instance_of?` when it isn't sure the change is safe.

## Actual behavior

1. Rubocop reported the following error:

```
test.rb:2:10: C: [Correctable] Style/ClassEqualityComparison: Use instance_of?(@klass_name) instead of comparing classes.
  object.class.name == @klass_name
```

The error message contains a suggestion to use `#instance_of?(@klass_name)`, but Rubocop doesn't know the type of `@klass_name`. That suggestion can be incorrect in many cases.

2. Rubocop changed the code above to `object.instance_of?(@klass_name)`, resulting in `TypeError` at runtime.

Additionally, when the method is in a module/class like in the example below:

```ruby
class MyClass
  def check_class?(object)
    object.class.name == @klass_name
  end
end
```

then Rubocop converts it to:

```ruby
class MyClass
  def check_class?(object)
    object.instance_of?(::@klass_name)
  end
end
```

This produces a syntax error in Ruby 3.1.

1. Create a file called `test.rb` with the following code:

```ruby
class MyClass
  def check_class?(object)
    object.class.name == @klass_name
  end
end
```

2. Call `rubocop test.rb`.

## RuboCop version

```
% bundle3.1 exec rubocop -V
1.38.0 (using Parser 3.1.2.1, rubocop-ast 1.23.0, running on ruby 3.1.2) [arm64-darwin22]
  - rubocop-rails 2.17.2
```

koic:
Fixes #11168.

This PR fixes an incorrect autocorrect for `Style/ClassEqualityComparison` when using instance variable comparison in module.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_an_incorrect_autocorrect_for_style_class_equality_comparison.md
+++ b/changelog/fix_an_incorrect_autocorrect_for_style_class_equality_comparison.md
@@ -1,0 +1,1 @@
+* [#11168](https://github.com/rubocop/rubocop/issues/11168): Fix an incorrect autocorrect for `Style/ClassEqualityComparison` when using instance variable comparison in module. ([@koic][])
--- a/lib/rubocop/cop/style/class_equality_comparison.rb
+++ b/lib/rubocop/cop/style/class_equality_comparison.rb
@@ -97,12 +97,14 @@ module RuboCop
           if node.children.first.method?(:name)
             return class_node.receiver.source if class_node.receiver
 
-            value = class_node.source.delete('"').delete("'")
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #11171, opened 2022-11-10
Support inline visibility definition on checking visibility

r7kamura:
The following patterns have not been supported by `RuboCop::Cop::VisibilityHelp#node_visibility`:

```ruby
class A
  private def x
  end
end
```

```ruby
class A
  def x
  end
  private :x
end
```

```ruby
node_visibility(def_x_node) #=> :public
```

I think it would be good to support them as well.

The background leading me to this change is as follows:

I am trying to create a Cop that inspects unrecommended visibility of some method definitions. I tried to use the `#node_visibility` for this implementation, but realized that this would not work as intended in this case if inline style visibility was used. There seemed to be no obvious reason not to support it, so I decided to create this Pull Request.
(FYI: In fact, I'm trying to create a cop that detects some public methods in Rails controller that actually should be private.)

After making the changes, several tests began to fail, so I decided to improve those as well.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_support_visibility_definition.md
+++ b/changelog/change_support_visibility_definition.md
@@ -1,0 +1,1 @@
+* [#11171](https://github.com/rubocop/rubocop/pull/11171): Support inline visibility definition on checking visibility. ([@r7kamura][])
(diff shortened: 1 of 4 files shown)


bbatsov:
Good change! This was indeed just an oversight on our part.

PR #11179, opened 2022-11-15
Add `Style/RedundantConstantBase` cop

r7kamura:
Since how Ruby searches constant is a bit complicated, and it can often be difficult to understand from the code whether the `::` is intended or not.

Where `Module.nesting` is empty, there should be no need to append `::`, so I think it would be nice to consistently avoid such meaningless `::` prefix to avoid confusion.

What do you think?

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/constant_base.rb
```diff
@@ -0,0 +1,76 @@
...
+      #   # good
+      #   module A
+      #     ::Const
+      #   end
+      class ConstantBase < Base
+        extend AutoCorrector
+
+        MSG = 'Avoid meaningless `::` prefix on constant.'
+
+        # @param node [RuboCop::AST::CbaseNode]
```
koic, line 39: No param or return value need to be annotated like many other cops do. For example this `RuboCop::AST::CbaseNode` would not exist. Can you remove these annotations from this PR?

config/default.yml
```diff
@@ -1568,6 +1568,11 @@ Lint/CircularArgumentReference:
   Enabled: true
   VersionAdded: '0.33'
 
+Lint/ConstantBase:
```
koic, line 1571: It's a style cop and I think adding "Redundant" to the name makes it consistent.

```suggestion
Style/RedundantConstantBase:
```

r7kamura:
I understood about the code changes, but what about the opinion of adding this Cop?
I would like to know first about it because if it is dismissed, then there is no point in improving the details.

koic:
I personally think it's acceptable. If you're worried, just wait for other comments.

config/default.yml
r7kamura: This Cop is very similar to `Lint/ConstantResolution` in the sense that it decides whether or not to add `::` to a constant.

Can you elaborate a bit more on what the difference is between Lint and Style here? I'm not quite sure I understand it yet. I may have missed some documents about each department, so if you know of any, I'd be glad to know 🙏

koic: `Lint/ConstantResolution` cop is intended to prevent bugs as wrote in its description. OTOH, this cop looks for consistency of style. The purpose of the cop is different.

I wrote it on the slide I talked about at past RubyKaigi for the difference is between Lint and Style. Here's what I can quickly recall.
https://speakerdeck.com/koic/improve-ruby-coding-styles-and-lint?slide=55

koic: As a side note, it's worth remembering that Lint is the special department for Severity.
https://github.com/rubocop/rubocop/issues/10080#issuecomment-1006148714

r7kamura: Thank you so much, and that's a good resource :+1:

I reconsidered that it should be in the `Style` department. It seems unlikely that the presence or absence of this `::` will cause any grammatical problems or bugs, so we can say that it is a matter of preference whether this Cop should be followed or not. In that sense, as you say, this should not be in the `Lint` department.

spec/rubocop/cop/lint/constant_base_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        end
+      RUBY
+    end
+  end
+
+  context 'with prefixed constant in neither class nor module' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        ::Bar
+        ^^ Avoid meaningless `::` prefix on constant.
```
koic, line 28: It can be told more simply.

```suggestion
        ^^ Remove redundant `::`.
```

bbatsov:
> I understood about the code changes, but what about the opinion of adding this Cop?

I'm fine with the proposal in general, although I think that the rationale needs to be expanded in the cop's description and we should probably add some matching guidelines to the style guide.

bbatsov:
Btw, what's the actual term for `::` in Ruby? I myself was a bit confused by `constant base` terminology, so it'd nice to align on the wording here. (I'm not sure "base" is the right word)

koic:
https://github.com/JuanitoFatas/what-do-you-call-this-in-ruby mentions `double colon` and `namespace resolution operator`. But I'm not sure about the official Ruby terminology.

r7kamura:
I know that the wording "cbase" and "const base" is used within the source code of Ruby itself, but I am actually not familiar with whether this is an official wording. It may be equally confusing to users, so if there is some other good alternative, I would like to make it so. It might be better to focus on its appearance rather than its meaning, as in `RedundantPrefixedDoubleColon` or so.

bbatsov:
If it's `const base` in Ruby's code I guess that's good enough for me. The other alternatives we've discussed aren't that great either, anyways. Thanks for tackling this!

mobilutz:
This cop clashes with our use of `Lint/ConstantResolution` where we have many constants defined to be used with `::`. Yes, rubocop handles this, but still it is not ideal having one cop saying you must use `::` and another saying you do not.

Is there some more background information on this cop? The description is understandable, but I would love to get some more information on this topic.

PR #11181, opened 2022-11-15
Fix pattern to match .tool-version files that specify multiple runtimes

noelblaschke:
A .tool-versions file may contain different runtime version entries.

E.g.:
```
nodejs 16.15.1 
ruby 2.7.6
```

The current implementation only checks the beginning of the .tool-versions and fails to find any entry if it's not the first.

This PR implements matching against beginning of each line and uses the first match. Assumption: There is only one entry per runtime per .tool-versions file.
-----------------

* [x] Added tests.


changelog/fix_fix_pattern_to_match_toolversion_files.md
```diff
@@ -0,0 +1 @@
+* [#x](https://github.com/rubocop/rubocop/pull/x): Fix pattern to match .tool-version files that specify multiple runtimes. ([@noelblaschke][])
```
koic, line 1: Can you update with the PR number and .tool-versions filename?

```suggestion
* [#11181](https://github.com/rubocop/rubocop/pull/11181): Fix pattern to match .tool-versions files that specify multiple runtimes. ([@noelblaschke][])
```

lib/rubocop/target_ruby.rb
```diff
@@ -84,7 +84,7 @@ def version_file
     # @api private
     class ToolVersionsFile < RubyVersionFile
       TOOL_VERSIONS_FILENAME = '.tool-versions'
-      TOOL_VERSIONS_PATTERN = /\Aruby (?:ruby-)?(?<version>\d+\.\d+)/.freeze
+      TOOL_VERSIONS_PATTERN = /^(?:ruby )(?<version>\d+\.\d+)/m.freeze
```
koic, line 87: The test passes without the `m` option. Is this required?

```suggestion
      TOOL_VERSIONS_PATTERN = /^(?:ruby )(?<version>\d+\.\d+)/.freeze
```

noelblaschke:
@koic Thank you for your immediate feedback. Changes are pushed 👍

koic:
Thanks!

PR #11189, opened 2022-11-18
[RuboCop] RBS Signature Files

coding-bunny:
This pull request introduces RBS Signature files to the core of RuboCop. This gives tools such as Steep to perform type-checking, and allows upstream consumes to use our typing to write compatible code. The idea is to introduce some type-checking and reduce the potential of bugs. Only works with the new Ruby 3.1

### DIFF
--- a/sig/rubocop/target_ruby.rbs
+++ b/sig/rubocop/target_ruby.rbs
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module RuboCop
+  class TargetRuby
+    DEFAULT_VERSION: Float
+    KNOWN_RUBIES: Array[Float]
+    OBSOLETE_RUBIES: Hash[Float, String]
+    SOURCES: Array[Class]
+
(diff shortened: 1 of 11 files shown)


koic:
Thank you for opening this proposal. However, I'm going to close this PR. Because many RBS being added are internal APIs (e.g. `ConfigValidator`, `Server`, `TargetRuby`, and others). I think users should not see anything other than published APIs, because internal APIs are not stable. And, from the perspective of RuboCop development, the cost of maintaining type definitions is higher. To be honest, I can't imagine the benefits beyond that yet. As far as I know, the ideal goal of RBS is "get benefits of types without writing them", but with this PR the maintenance costs of writing types will outweigh the benefits of benefits. Of cause, Having say that, it would need to be written by people on low layers as a API provider :-) I think we can revisit it in the future when RBS is a bit more mature. So, let's wait a little longer for the RBS ecosystem to stabilize.

PR #11203, opened 2022-11-26
Support multiple arguments on `Lint/SendWithMixinArgument`

r7kamura:
These mixin methods can take multiple arguments, so these patterns should also be supported in this cop.

- `Module#include`
- `Module#prepend`
- `Object#extend`

-----------------

* [x] Added tests.


r7kamura:
This may be a trivial point, but I chose `change` as the category for the CHANGELOG entry because I figured that some existing users would expect it to work only with one argument. If another category is more appropriate, please let me know. (This may be `fix` if you strongly believe it should have been supported, or `new` if you believe it is a feature addition)

koic:
Looks good to me!

PR #11206, opened 2022-11-27
Change `Lint/InterpolationCheck` from `Safe: false` to `SafeAutoCorrect: false`

r7kamura:
In this Cop, I think it would be easier to understand to have a simple policy that all `#{...}` in single-quotes are considered offenses, regardless of how they are intended to be written. Given this, it would be nice to set `SafeAutoCorrect: false`, since I don't think there are any false-positives cases under this policy.

What do you think?

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/interpolation_check.rb
+++ b/lib/rubocop/cop/lint/interpolation_check.rb
@@ -6,9 +6,10 @@ module RuboCop
       # Checks for interpolation in a single quoted string.
       #
       # @safety
-      #   This cop is generally safe, but is marked as unsafe because
-      #   it is possible to actually intentionally have text inside
-      #   `#{...}` in a single quoted string.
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/interpolation_check.rb
```diff
@@ -6,9 +6,8 @@ module Lint
       # Checks for interpolation in a single quoted string.
       #
       # @safety
-      #   This cop is generally safe, but is marked as unsafe because
-      #   it is possible to actually intentionally have text inside
-      #   `#{...}` in a single quoted string.
+      #   This cop's autocorrection is unsafe because although it always replaces single quotes as
```
bbatsov, line 9: I think we need some example here that illustrates what's unsafe, as they way it's currently worded is not very clear IMO.

bbatsov:
Thanks!

PR #11211, opened 2022-11-29
Add autocorrect for `Lint/AssignmentInCondition`

r7kamura:
I think adding support for unsafe autocorrection to `Lint/AssignmentInCondition` would be more useful in some cases. Now that `SafeAutoCorrect: false` allows us to indicate to users that it is unsafe, I don't think we will have much trouble because of this support.

I decided to replace the code as assignment by default in this autocorrection because it does not change the behavior of the current code. If there seem to be many use cases where the behavior of autocorrection needs to be changed, we can add such configuration in this Pull Request or the future change. (In fact, is there any Cop that allows the autocorrection behavior to be switched by configuration in that way?)

Depending on the policy, this type of autocorrection support may not be acceptable in RuboCop. In which case, please do not hesitate to close this Pull Request.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/assignment_in_condition.rb
+++ b/lib/rubocop/cop/lint/assignment_in_condition.rb
@@ -11,6 +11,10 @@ module RuboCop
       # an assignment to indicate "I know I'm using an assignment
       # as a condition. It's not a mistake."
       #
+      # @safety
+      #   This cop's autocorrection is unsafe because it assumes that
+      #   the author meant to use an assignment result as a condition.
+      #
       # @example
       #   # bad
       #   if some_var = true
@@ -35,6 +39,8 @@ module RuboCop
       #   end
       #
       class AssignmentInCondition < Base
+        extend AutoCorrector
+
         include SafeAssignment
 
         MSG_WITH_SAFE_ASSIGNMENT_ALLOWED =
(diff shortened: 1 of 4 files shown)


r7kamura:
As background, I had to migrate a bit large old codebase from rubocop 0.44 to 1.39. I thought it would be easy to complete with autocorrection because the tests were well written, but this cop did not support autocorrection, so I had to manually rewrite a lot of such ambiguous assignment code, which was very complicated task.
In other words, I am saying that there are cases where it is useful enough to have it provided, even if it is not a perfect autocorrection.

koic:
I think any configuration is not required for this autocorrection. So this unsafe autocorrection approach would be better.

lib/rubocop/cop/lint/assignment_in_condition.rb
```diff
@@ -53,7 +59,11 @@ def on_if(node)
             next :skip_children if skip_children?(asgn_node)
             next if allowed_construct?(asgn_node)
 
-            add_offense(asgn_node.loc.operator)
+            add_offense(asgn_node.loc.operator) do |corrector|
+              next unless safe_assignment_allowed?
+
+              corrector.replace(asgn_node, "(#{asgn_node.source})")
```
koic, line 65: That's my two cents:

```suggestion
              corrector.wrap(asgn_node, '(', ')')
```

bbatsov:
I like the proposal. Thanks!

PR #11227, opened 2022-12-02
Handle ternaries in `Style/SafeNavigation`

### LINKED ISSUE #11116: Possible enhancement to Style/SafeNavigation
Currently, `Style/SafeNavigation` detects [many patterns](https://docs.rubocop.org/rubocop/cops_style.html#examples-183) where a safe navigation operator would result in a more compact expression.

I've encountered the following ternary expression in many codebases, and it might be a good candidate to be added to the `Style/SafeNavigation` list:

```ruby
# bad
x ? x.foo : nil

# good
x&.foo
```

fatkodima:
Closes #11116.

It handles new ternary cases like:
```ruby
x ? x.foo : nil

x.nil? ? nil : x.foo

!x.nil? ? x.foo : nil

!x ? nil : x.foo
```

### DIFF
--- a/changelog/change_safe_navigation_to_handle_ternaries.md
+++ b/changelog/change_safe_navigation_to_handle_ternaries.md
@@ -1,0 +1,1 @@
+* [#11116](https://github.com/rubocop/rubocop/issues/11116): Handle ternaries in `Style/SafeNavigation`. ([@fatkodima][])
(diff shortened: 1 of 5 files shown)


bbatsov:
Nice! I think it'd be a good idea to also mention this in the examples.

fatkodima:
Added examples.

bbatsov:
Thanks!

PR #11228, opened 2022-12-02
Add ability to count method arguments as one line to code length related `Metric` cops

### LINKED ISSUE #11070: Request: Allow to add method arguments to Metrics/MethodLength CountAsOne

I tend to write
```ruby
method(
  param1, param2, param3
)
```
Instead of
```ruby
method(
  param1,
  param2,
  param3
)
```
Which I think is clearer, just because I need to respect the method length limit.

Sometimes calling the method in one line isn't possible if I respect a 80 characters line width. 

I'd like to be able to specify that method calls can be counted as a single line. Something like:

```yaml
Metrics/MethodLength:
  CountAsOne: ['arguments']
```

fatkodima:
Closes #11070.

With the
```yaml
# .rubocop.yml
Metric/MethodLength:
  CountAsOne: ['arguments']
```

the following will be counted as 1 line:
```ruby
foo(
  1,
  2
)
```

### DIFF
--- a/changelog/change_count_as_one_accepts_arguments.md
+++ b/changelog/change_count_as_one_accepts_arguments.md
@@ -1,0 +1,1 @@
+* [#11070](https://github.com/rubocop/rubocop/issues/11070): Add ability to count method calls as one line to code length related `Metric` cops. ([@fatkodima][])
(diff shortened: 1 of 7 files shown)


bbatsov:
Looks good, but I am wondering a bit about the terminology here - is it really `arguments` we are after or we want method calls to always be counted as one line?

fatkodima:
@bbatsov Changed the name.

bbatsov:
Thanks!

PR #11234, opened 2022-12-03
CI: Let ruby/setup-run install gems

bquorning:
Simplify GHA CI setup a little bit.

No user-visible changes.


koic:
Good catch!

PR #11235, opened 2022-12-03
Add `Style/RequireOrder` cop

r7kamura:
Many projects have `require` and `require_relative` in alphabetical order by convention, the rubocop gem itself being one example. I thought it would be good to have a cop to provide such consistency.

As a good example, rubocop-rspec inspects this with tests. Such checks are replaced by this cop.

- https://github.com/rubocop/rubocop-rspec/blob/v2.15.0/spec/project/project_requires_spec.rb

Like `Bundler/OrderedGems`, code separated by a blank line is considered a different group, so if there are obvious dependencies but you really don't want to write `require_relative` in each file, it is better to separate the groups. lib/rubocop/formatter.rb is a good example of that.

- https://github.com/rubocop/rubocop/blob/v1.39.0/lib/rubocop/formatter.rb#L9-L27

Unfortunately, the dependencies written in lib/rubocop.rb are too complex, so I gave up on organizing this at the time of introducing this cop and decided to ignore it from this cop. Personally, if one file A depends on another file B, I prefer to indicate that fact by writing `require_relative` in file A, but I don't think that is the practice in the rubocop gem, so I don't want to go into this issue here.

Since this cop enforces the rule "always sort them in alphabetical order within a group", I don't think false-positive will occur, but applying autocorrection obviously changes the behavior from the original code, so I have set `SafeAutoCorrect: false`.

Ideally, I believe it is a good manner that the code to always be written in such a way that it does not depend on the order of `require`, but I set `Enabled: false` because this cop has a large impact and may or may not be acceptable depending on the preference of the project. I think this cop is rather opinionated, so I would be glad to get your opinion on the policy.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/require_order.rb
+++ b/lib/rubocop/cop/style/require_order.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Sort `require` and `require_relative` in alphabetical order.
+      #
+      # @safety
+      #   This cop's autocorrection is unsafe because it will obviously change the execution order.
+      #
+      # @example
+      #   # bad
+      #   require 'b'
+      #   require 'a'
+      #
+      #   # good
+      #   require 'a'
+      #   require 'b'
+      #
+      #   # bad
+      #   require_relative 'b'
+      #   require_relative 'a'
+      #
+      #   # good
+      #   require_relative 'a'
+      #   require_relative 'b'
+      #
+      #   # good (sorted within each section separated by a blank line)
+      #   require 'a'
+      #   require 'd'
+      #
+      #   require 'b'
+      #   require 'c'
+      #
+      #   # good
+      #   require 'b'
+      #   require_relative 'c'
+      #   require 'a'
+      class RequireOrder < Base
+        extend AutoCorrector
+
+        include RangeHelp
+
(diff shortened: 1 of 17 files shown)


bbatsov:
I love this suggestion and the thoughtful analysis you've submitted alongside it! I'm only wondering if this cop should also suggest to keep `require` and `require_relative` sections ordered as well (e.g. `require` first, `require_relative` afterwards or vice-versa).

lib/rubocop/cop/style/require_order.rb
```diff
@@ -0,0 +1,88 @@
...
+      #
+      #   # bad
+      #   require_relative 'b'
+      #   require_relative 'a'
+      #
+      #   # good
+      #   require_relative 'a'
+      #   require_relative 'b'
+      #
+      #   # good
```
bbatsov, line 28: You might mention here that a blank line introduces a new section or something like this.

r7kamura:
> 'm only wondering if this cop should also suggest to keep `require` and `require_relative` sections ordered as well (e.g. `require` first, `require_relative` afterwards or vice-versa).

I guess it would be difficult to enforce that, since there would be case where you want to write `relative_relative` before `require`, or vice versa, depending on the situation. If we want to do that, it might be better to prepare something like `Style/RequireSectionOrder` separately.

koic:
@zhuravel Thank you for the feedback. That issue will be resolved by #11254.

PR #11247, opened 2022-12-08
Optimize checking if the file is relevant for the cop

fatkodima:
Follow up to #11240.

Tested on Gitlab's codebase (`app/models` directory).

Because the project has a large set of rubocop todo files (https://gitlab.com/gitlab-org/gitlab/-/tree/master/.rubocop_todo), most items in `Exclude`/`Include` of which are exact file names (which I suppose correlates with most of the projects with todo files), it is wasteful to linearly check the list of patterns over and over to check if the file is relevant to the cop. 
Example of such configuration - https://gitlab.com/gitlab-org/gitlab/-/blob/master/.rubocop_todo/gitlab/namespaced_class.yml.

I optimized to firstly do cheaper checks first via partitioning an array of patterns into a set of exact match strings (and be able to do O(1) search) and the rest of the patterns.

## Before
Time: `86.4s`

```
==================================
  Mode: wall(1000)
  Samples: 86352 (0.09% miss rate)
  GC: 3948 (4.57%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
     18789  (21.8%)       11671  (13.5%)     RuboCop::PathUtil#match_path?
      6742   (7.8%)        6742   (7.8%)     RuboCop::PathUtil#maybe_hidden_file?
      4925   (5.7%)        3005   (3.5%)     RuboCop::AST::Descendence#each_child_node
      2814   (3.3%)        2814   (3.3%)     (sweeping)
      2475   (2.9%)        2475   (2.9%)     block (2 levels) in <class:Node>
      3383   (3.9%)        2304   (2.7%)     Parser::Lexer#advance
      2089   (2.4%)        2076   (2.4%)     RuboCop::Config#for_cop
      1935   (2.2%)        1935   (2.2%)     Gitlab::Styles::Rubocop::ModelHelpers#in_model?
      2920   (3.4%)        1639   (1.9%)     RuboCop::Cop::Base#cop_config
      1388   (1.6%)        1388   (1.6%)     Gitlab::Styles::Rubocop::ModelHelpers#in_model?
      1210   (1.4%)        1210   (1.4%)     RuboCop::QAHelpers#in_qa_file?
      1203   (1.4%)        1203   (1.4%)     Gitlab::Styles::Rubocop::ModelHelpers#in_model?
      1128   (1.3%)        1128   (1.3%)     (marking)
      1081   (1.3%)        1081   (1.3%)     RuboCop::AST::Node#node_parts
[...]

PR #11252, opened 2022-12-08
Tweak optimized checking if the file is relevant for the cop

fatkodima:
Follow up to https://github.com/rubocop/rubocop/pull/11247 (addresses style suggestions from the comments).


bbatsov:
Perfect!

PR #11260, opened 2022-12-09
Permit "extends" directive to support adding to a parent .rubocop.yml without changing path relativity

alexevanczuk:
This PR supports the idea of `extends`, which permits us to allow a child `.rubocop.yml` to make changes to a parent's configuration *without* affecting path relativity.

I chose this different directive since configuration inheritance changes are complex, so I wanted to be really clear that this is a new feature with different behavior and for different use cases to leave the existing feature unaffected.
-----------------

* [x] Added tests.

### DIFF
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -275,6 +275,42 @@ Style/For:
 
 In this example the `Exclude` would only include `bar.rb`.
 
+=== Extending a parent .rubocop.yml using extends
+
+The optional directive `extends` allows a child `.rubocop.yml` to specify
+that it changes a parent `.rubocop.yml` without affecting path-relativity rules.
+
+Given the following configs:
+
+[source,yaml]
+----
+# .rubocop.yml
+AllCops:
+  Exclude:
+    - 'generated/**/*.rb'
+
+Layout/LineLength:
+  Enabled: true
+  Include:
+    - 'child_directory/**/*.rb'
+
+Rails/HasAndBelongsToMany:
+  Enabled: false
+----
+
+[source,yaml]
+----
+# child_directory/.rubocop.yml
+extends: '../../.rubocop.yml'
+
+Rails/HasAndBelongsToMany:
+  Enabled: true
+----
+
+Using `extends: '../../.rubocop.yml` within the child `.rubocop.yml` file allows the path relativity of the parent `.rubocop.yml` to remain the same. That is, the `Include` for `Layout/LineLength` is still relative to `.rubocop.yml` rather than relative to `child_directory/.rubocop.yml`.
+
+Contrast this with `inherit_from: '../../.rubocop.yml`, which would make the `Include` for `Layout/LineLength` be `child_directory/child_directory/**/*.rb`, since paths would be made relative to `child_directory/.rubocop.yml`.
+
 == Pre-processing
 
 Configuration files are pre-processed using the ERB templating mechanism. This
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -285,31 +285,21 @@ Given the following configs:
 [source,yaml]
 ----
 # .rubocop.yml
-AllCops:
-  Exclude:
-    - 'generated/**/*.rb'
-
 Layout/LineLength:
   Enabled: true
   Include:
     - 'child_directory/**/*.rb'
-
-Rails/HasAndBelongsToMany:
-  Enabled: false
 ----
 
 [source,yaml]
 ----
 # child_directory/.rubocop.yml
-extends: '../../.rubocop.yml'
-
-Rails/HasAndBelongsToMany:
-  Enabled: true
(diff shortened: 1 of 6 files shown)


bbatsov:
Overall the change seems reasonable, but I'm worried if this won't be confusing for the end users as the difference between `inherit_from` and `extends` is not obvious. (even with some documentation already in place) I'll leave it to @jonas054 (our Master of Configuration) to weigh in on this. 

I'm also wondering if it won't be easier to add some extra configuration to `inherit_from` instead that alters its behavior.

docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -275,6 +275,42 @@ Style/For:
...
+    - 'child_directory/**/*.rb'
+
+Rails/HasAndBelongsToMany:
+  Enabled: false
+----
+
+[source,yaml]
+----
+# child_directory/.rubocop.yml
+extends: '../../.rubocop.yml'
```
jonas054, line 304: This should only go up one level, so `extends: '../.rubocop.yml'`

docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -275,6 +275,42 @@ Style/For:
...
+AllCops:
+  Exclude:
+    - 'generated/**/*.rb'
+
+Layout/LineLength:
+  Enabled: true
+  Include:
+    - 'child_directory/**/*.rb'
+
+Rails/HasAndBelongsToMany:
```
jonas054, line 297: The configuration of this cop is not relevant for the subject of inheritance and relativity, right? So better to remove it and keep the example minimal.

docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -275,6 +275,42 @@ Style/For:
...
+# child_directory/.rubocop.yml
+extends: '../../.rubocop.yml'
+
+Rails/HasAndBelongsToMany:
+  Enabled: true
+----
+
+Using `extends: '../../.rubocop.yml` within the child `.rubocop.yml` file allows the path relativity of the parent `.rubocop.yml` to remain the same. That is, the `Include` for `Layout/LineLength` is still relative to `.rubocop.yml` rather than relative to `child_directory/.rubocop.yml`.
+
+Contrast this with `inherit_from: '../../.rubocop.yml`, which would make the `Include` for `Layout/LineLength` be `child_directory/child_directory/**/*.rb`, since paths would be made relative to `child_directory/.rubocop.yml`.
```
jonas054, line 312: I can't reproduce the problem you describe. The current logic is that paths given in a base configuration file beginning with `.rubocop` are relative to that base configuration file. So I don't understand how you get `child_directory/child_directory/**/*.rb`.

jonas054 (changes requested):
Sorry for being so late with my review. All the code looks reasonable to me, but like I said in a comment, I don't see what the problem is as I'm unable to reproduce it.

docs/modules/ROOT/pages/configuration.adoc
alexevanczuk: Good call, removed!

docs/modules/ROOT/pages/configuration.adoc
alexevanczuk: Changed!

docs/modules/ROOT/pages/configuration.adoc
alexevanczuk: I have a reproduction of the issue here: https://github.com/alexevanczuk/rubocop-child-rubocop-yml-relative-path-issue

In this, you can see that the top-level `.rubocop.yml` file clearly covers files in a subdirectory, but when using `inherit_from` in a child `.rubocop.yml` to override specific cops, it no longer matches that same file.

I made a typo in the docs there that I corrected. The `Include` actually changes to `../**/*.rb` which ends up not matching the file path.

Let me know if this clarifies things!

jonas054: Thanks! That helped a lot.

After having reproduced the problem, I've come to the conclusion that there is a bug in how the paths are resolved when `inherit_from` is used. So we need to fix the bug, but there's no need for a new directive (`extends`).

The bug is that in the method that used to be called `fix_include_paths`, we made the include paths relative to the inheriting (child) configuration file, but when we later try to match file names against configured paths (in `PathUtil.match_path?`), relative paths are relative to the current directory.

The solution should be to make paths relative to `Dir.pwd` in `fix_include_paths` -- or maybe make them absolute.

alexevanczuk: Gotcha, that makes a lot of sense @jonas054 . If we can fix this bug and in doing so I'm able to get the behavior we need without a new feature, that would be great.

Is this something I can help with? If so, I can dig in this coming week! Otherwise, happy to look at any PRs 😄 

I'll close this branch for now since it's not the implementation we want to pursue.

alexevanczuk: Here's the PR: https://github.com/rubocop/rubocop/pull/11416

PR #11265, opened 2022-12-11
Allow extensions inject their defaults

pirj:
Presently, extensions (e.g. [rubocop-rspec](https://github.com/rubocop/rubocop-rspec/blob/63908ab48fad95e08c4e0dc21073a716cb91d06c/lib/rubocop/rspec/inject.rb#L8), [rubocop-rails](https://github.com/rubocop/rubocop-rails/blob/f94b23596821f80c861e6632807a959dbc306aab/lib/rubocop/rails/inject.rb#L8)) all include a similar piece of code, that is [called from their entry point](https://github.com/rubocop/rubocop-rspec/blob/63908ab48fad95e08c4e0dc21073a716cb91d06c/lib/rubocop-rspec.rb#L43).

This change would allow getting rid of this common code and streamline `config/default.yml` injection:
```ruby
# e.g. lib/rubocop-capybara.rb
require_relative 'rubocop/cop/capybara_cops'

project_root = File.join(__dir__, '..')
RuboCop::ConfigLoader.inject_defaults!(project_root)
```

:warning: no docs, no tests, no changelog entry - I consider this an experimental feature.
I'm all for this hack to be replaced with a better approach in RuboCop 2.0.

### Some food for the thought

```ruby
# e.g. lib/rubocop-capybara.rb
require_relative 'rubocop/cop/capybara_cops'

project_root = File.join(__dir__, '..')
obsoletion = File.join(project_root, 'config', 'obsoletion.yml')
RuboCop::ConfigObsoletion.files << obsoletion if File.exist?(obsoletion)
```

Looking at the above code for obsoletion config injection, it feels like adding something like
```ruby
# e.g. lib/rubocop-capybara.rb
require_relative 'rubocop/cop/capybara_cops'

project_root = File.join(__dir__, '..')
extension_config_defaults = File.join(project_root, 'config', 'defaults.yml')
RuboCop::ConfigLoader.defaults_files << extension_config_defaults
```
would be more natural. But I'm slightly concerned that `DEFAULT_FILE` is used in several places, and also the load order (extensions first, or default configs first?) and didn't dare to make such a change without knowing all potential consequences.

-----------------

* [-] Added tests.

### DIFF
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -137,6 +137,7 @@ module RuboCop
         end
       end
 
+      # @api private
       def inject_defaults!(project_root)
         path = File.join(project_root, 'config', 'default.yml')
         config = load_file(path)
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -137,6 +137,14 @@ module RuboCop
         end
       end
 
+      def inject_defaults!(project_root)
+        path = File.join(project_root, 'config', 'default.yml')
+        config = load_file(path)
+        new_config = ConfigLoader.merge_with_default(config, path)
+        puts "configuration from #{path}" if debug?
+        @default_configuration = new_config
+      end
+
       # Returns the path RuboCop inferred as the root of the project. No file
       # searches will go past this directory.
       # @deprecated Use `RuboCop::ConfigFinder.project_root` instead.


bbatsov:
@pirj So, the idea is to merge this as-is, but treat it as private for the time being, right? 

> would be more natural. But I'm slightly concerned that DEFAULT_FILE is used in several places, and also the load order (extensions first, or default configs first?) and didn't dare to make such a change without knowing all potential consequences.

I agree there's room for improvement here, but you'll have to try this out with some extensions to make sure nothing breaks.

lib/rubocop/config_loader.rb
```diff
@@ -137,6 +137,14 @@ def default_configuration
         end
       end
 
+      def inject_defaults!(project_root)
```
pirj, line 140:
```suggestion
      # @api private
      def inject_defaults!(project_root)
```

pirj:
> merge this as-is, but treat it as private for the time being, right?

Yes. I've marked it as private in docs. We can use it in our repos, but won't advertise it until it's more aligned with `RuboCop::ConfigObsoletion.files` interface.

> try this out with some extensions to make sure nothing breaks

Already tested with `rubocop-rspec` and the extracted `rubocop-capybard` by pointing them to this `rubocop`'s branch, checked out locally. Works as expected. This change is not pushed anywhere, but I'll gladly send a draft `rubocop-rspec` PR to demonstrate that.

bbatsov:
Okay, let's include this in the upcoming release.

jeromedalbert:
Thanks for doing this, I created my first rubocop extension and hit a few snags:

1) My gem's `config/default.yml` was never loaded when I required the gem in other projects, and https://docs.rubocop.org does not document how to create a rubocop extension. After asking GPTs and looking at the source code of popular rubocop extensions, I realized that I had to load the configuration in my gem entrypoint.
2) These popular extensions seem to inject in different ways but the main one was to use some sort of `Inject.defaults!` [duplicated](https://github.com/rubocop/rubocop-rspec/issues/246) across many of them. That duplication and hacking didn't feel great to me.

So I much prefer the simple `RuboCop::ConfigLoader.inject_defaults!(File.join(__dir__, '..'))` one-liner that this PR introduced. Thanks again!

---

The only worrying thing is that the gem docs for [`inject_defaults!`](https://www.rubydoc.info/gems/rubocop/RuboCop/ConfigLoader#inject_defaults!-class_method) has a scary message about potentially being private and being removed in the future:

Screenshot 2024-10-20 at 2 30 00 PM

So I will be watching this thread just in case!

PR #11270, opened 2022-12-12
[Fix #11266] Fix a false positive for `Style/RedundantConstantBase`

### LINKED ISSUE #11266: Lint/ConstantResolution and Style/RedundantConstantBase do not work well together
## Expected behavior

I expect that these 2 Cops would be able to work together and have the ability to distinguish between external gems and project code. My suggestion is that:

* `Lint/ConstantResolution` focusses on project code, and ignores gems loaded through the `Gemfile`
* `Style/RedundantConstantBase` focusses on external gems and doesn't check project references.

## Actual behavior

When both are enabled, you end up in an infinite chicken/egg problem because both cops do not agree on the correct constant resolution in the source code.

Use the following code snippet:

```
::RSpec.configure do |config|
  # Enable flags like --only-failures and --next-failure
  config.example_status_persistence_file_path = '.rspec_status'

  # Disable RSpec exposing methods globally on `Module` and `main`
  config.disable_monkey_patching!

  config.expect_with(:rspec) do |c|
    c.syntax = :expect
  end
end
```

This triggers the following output from RuboCop
```
spec/spec_helper.rb:12:1: E: [Correctable] Style/RedundantConstantBase: Remove redundant ::.
::RSpec.configure do |config|
^^
```

But if you apply the change and remove the leading `::`, the following is outputted instead:

```
spec/spec_helper.rb:12:1: E: Lint/ConstantResolution: Fully qualify this constant to avoid possibly ambiguous resolution.
RSpec.configure do |config|
^^^^^
```

This means it's impossible to find a correct code approach to resolve this unless one of the Cops is disabled.

## RuboCop version

```
$ [bundle exec] rubocop -V
1.40.0 (using Parser 3.1.3.0, rubocop-ast 1.24.0, running on ruby 2.7.6) [x86_64-darwin21]
  - rubocop-performance 1.15.1
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.15.0
  - rubocop-thread_safety 0.4.4
```

koic:
Fixes #11266.

This PR fixes a false positive for `Style/RedundantConstantBase` when enabling `Lint/ConstantResolution`.

It respects `Style/RedundantConstantBase` when enabling `Lint/ConstantResolution`. OTOH, it respects `Style/RedundantConstantBase` when disabling `Lint/ConstantResolution`.

This is to honor the desire to enable the Lint cop which is disabled by default.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/redundant_constant_base.rb
```diff
@@ -42,6 +42,7 @@ class RedundantConstantBase < Base
         MSG = 'Remove redundant `::`.'
 
         def on_cbase(node)
+          return if lint_constant_resolution_cop_enabled?
```
bbatsov, line 45: This behavior should be documented in the cop's description.

koic: Sure! I've update the docs of `Style/RedundantConstantBase` and `Lint/ConstantResolution` cops.

bbatsov:
Thanks!

PR #11274, opened 2022-12-13
[Fix #11273] Fix a false positive for `Lint/DuplicateMethods`

### LINKED ISSUE #11273: Lint/DuplicateMethods false positive when using `alias_method` to change/change back a method
This is something we sometimes do as a test helper method to ensure some sort of behavior out of a method in the class to be tested.  Imagine, for example, we wanted `save` in the tested class to always raise an error when called, but we also don't want it to _stay_ modified:

```ruby
module FooTest
  def make_save_always_fail
    Foo.class_eval do
      def failed_save
        raise
      end
      alias_method :original_save, :save
      alias_method :save, :failed_save
    end

    yield
  ensure
    Foo.class_eval do
      alias_method :save, :original_save
    end
  end
end
```

If you call `rubocop --only Lint/DuplicateMethods` on that snippet, you get a false positive for the "second" definition of `save` inside the `ensure` block.

https://wasabigeek.com/blog/practical-metaprogramming-in-ruby-minitest-mock/ outlines a similar form of this.

--------

## Expected behavior

Should allow multiple `alias_method` definitions of the same name if they happen in different scopes.

## Actual behavior

See above.

See example snippet and command above.

## RuboCop version

```
1.40.0 (using Parser 3.1.3.0, rubocop-ast 1.24.0, running on ruby 3.0.4) [x86_64-darwin21]
```

koic:
Fixes #11273.

This PR fixes a false positive for `Lint/DuplicateMethods` when there are same `alias_method` name outside `rescue` or `ensure` scopes.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_lint_duplicate_methods.md
+++ b/changelog/fix_a_false_positive_for_lint_duplicate_methods.md
@@ -1,0 +1,1 @@
+* [#11273](https://github.com/rubocop/rubocop/issues/11273): Fix a false positive for `Lint/DuplicateMethods` when there are same `alias_method` name outside `rescue` or `ensure` scopes. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #11278, opened 2022-12-13
[Fix #11277] Add new `Style/ConcatArrayLiterals` cop

### LINKED ISSUE #11277: Cop idea: recommend `push` over `concat` for single element lists
I often see new Ruby devs use `concat` to add an element to a list. This cop would recommend `push` (or `<<` depending on configuration).

```ruby
# bad
list.concat [item]
# good
list.push item
list << item
```

It also happens to be about 3x faster (10M iterations here):
```
        user     system      total        real
concat  1.055298   0.037218   1.092516 (  1.092645)
push    0.368790   0.017711   0.386501 (  0.386501)
```

koic:
Fixes #11277.

## Summary

Enforces the use of `Array#push(item)` instead of `Array#concat([item])` to avoid redundant array literal.

## Safety

This cop is unsafe, as it can produce false positives if the receiver is not an `Array` object.

## Examples

```ruby
# bad
list.concat([item])

# good
list.push(item)
```

If the array literal has multiple elements like `item.concat([foo, bar])`, replace it with `item.push(foo, bar)`. `<<` is not allowed multiple arguments. So for autocorrection choose `push`.

## Benchmark

```ruby
# concat_array_literals.rb
require 'benchmark/ips'

ary = [:a, :b, :c]

Benchmark.ips do |x|
  x.report('concat') { ary.concat([:d, :e, :f]) }
  x.report('push')   { ary.push(:d, :e, :f) }

  x.compare!
end
```

```console
% ruby concat_array_literals.rb
              concat   616.057k i/100ms
                push     1.099M i/100ms
              concat      7.382M (±14.5%) i/s -     35.115M in   5.388710s
                push     11.604M (± 2.9%) i/s -     58.241M in   5.023063s

Comparison:
                push: 11604292.6 i/s
              concat:  7382414.4 i/s - 1.57x  (± 0.00) slower
```

## Additional Infromation

Percent array literals are not supported because the complexity of autocorrection logic.

And I think this cop can be placed in Style department, not Performance department. Because there doesn't seem to be much risk of lowering readability.

-----------------

* [x] Added tests.


dvandersluis:
Should this be in `Performance` instead of `Style`?

koic:
Quote from this PR description:

> And I think this cop can be placed in Style department, not Performance department. Because there doesn't seem to be much risk of lowering readability.

I think performance is secondary and can be put in `Style`. For example, this style has the advantage of eliminating the use of redundant array literals. That said, of course any feedback on this is welcome.

dvandersluis:
My apologies, I missed that line. I don't have a strong feeling either way, just by looking at the benchmark that was my immediate thought.

lib/rubocop/cop/style/concat_array_literals.rb
```diff
@@ -0,0 +1,62 @@
...
+      # to avoid redundant array literals.
+      #
+      # @safety
+      #   This cop is unsafe, as it can produce false positives if the receiver
+      #   is not an `Array` object.
+      #
+      # @example
+      #
+      #   # bad
+      #   list.concat([item])
```
bbatsov, line 16: I think it's better to add examples with multiple elements as well, so people won't think this works only one-element concats.

koic: Exactly! I've added examples.

bbatsov:
Thanks!

PR #11049, opened 2022-10-05
Fixes documentation message for EnforcedShorthandSyntax: consistent

lucianghinda:
`EnforcedShorthandSyntax: consistent` behave more like `either` than `always`

Why is it not like `always`:

- `consistent` allows regular and shorthand hash syntax, but not mixed
- `always`     allows only the shorthand syntax

=> `consistent` does not behave like `always

Why is like `either`:
- `consistent` allows regular and shorthand hash syntax, but not mixed
- `either`     allows regular and shorthand hash syntax.

=> `consistent` is similar to `either` but does not allow mixing syntax.

**Replace this text with a summary of the changes in your PR.
The more detailed you are, the better.**

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [ ] Added tests.
* [ ] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [ ] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

Issue #10933, opened 2022-08-17
"EnforcedShorthandSyntax: consistent" doesn't work?


jdelStrother:
I'm using rubocop 1.35, which includes the `EnforcedShorthandSyntax: consistent` option in Style/HashSyntax (https://github.com/rubocop/rubocop/pull/10776)

However in my experience, `consistent` just behaves like `always`.

## Expected behavior

`EnforcedShorthandSyntax: consistent` should allow `{foo:, bar:}` and `{foo: foo, bar: bar}`

## Actual behavior

`EnforcedShorthandSyntax: consistent` only allows `{foo:, bar:}`

.rubocop.yml :

```yaml
AllCops:
  TargetRubyVersion: 3.1
  DisabledByDefault: true

Style/HashSyntax:
  Enabled: true
  EnforcedShorthandSyntax: consistent
```

test.rb :

```rb
foo, bar, baz = nil
short = { foo: }
long = { foo: foo }
short_consistent = { foo:, bar: }
long_consistent = { foo: foo, bar: bar}
inconsistent = { foo: foo, bar: }
```

```
$ bundle exec rubocop --debug
For /Users/jon/Developer/consistent-hash: configuration from /Users/jon/Developer/consistent-hash/.rubocop.yml
Default configuration from /Users/jon/.gem/ruby/3.1.0/gems/rubocop-1.35.0/config/default.yml
Use parallel by default.
Running parallel inspection
Loading cache from /Users/jon/.cache/rubocop_cache/71589f41d68f90ef315fa6058661f86c7fa52057/6d7a3b621ca1730e04accd938619e4bdab66cfb1/009fb4a3f5837ba88be0887547b8a871a0c297c1
Inspecting 2 files
Scanning /Users/jon/Developer/consistent-hash/Gemfile
Loading cache from /Users/jon/.cache/rubocop_cache/71589f41d68f90ef315fa6058661f86c7fa52057/6d7a3b621ca1730e04accd938619e4bdab66cfb1/009fb4a3f5837ba88be0887547b8a871a0c297c1
.Scanning /Users/jon/Developer/consistent-hash/test.rb
Loading cache from /Users/jon/.cache/rubocop_cache/71589f41d68f90ef315fa6058661f86c7fa52057/6d7a3b621ca1730e04accd938619e4bdab66cfb1/495fb8579da040f56bf4f00e9f5114634a98f29e
C

Offenses:

[...]

PR #11286, opened 2022-12-14
Improve doc of EnforcedShorthandSyntax: consistent

tight:
This PR improves the documentation of `Style/HashSyntax`'s `EnforcedShorthandSyntax` for the `consistent` style.

The documentation (https://docs.rubocop.org/rubocop/1.40/cops_style.html#stylehashsyntax) says

> always - forces use of the 3.1 syntax (e.g. {foo:})
> never - forces use of explicit hash literal value
> either - accepts both shorthand and explicit use of hash literal value
> consistent - like "either", but will avoid mixing styles in a single hash

There is a recent change (https://github.com/rubocop/rubocop/pull/11049, the documentation was `like "always", but will avoid mixing styles in a single hash` before) about this, but it's still unclear IMHO (see also https://github.com/rubocop/rubocop/issues/10933).

This PR change the documentation to :

> consistent - forces use of the 3.1 syntax only if all values can be omitted in the hash

and updates the examples to be more explicit about why they are good or bad.

-----------------

* [x] Added tests.
➡️ No code change
➡️ ❓ Is it required for a documentation change ?

### DIFF
--- a/lib/rubocop/cop/style/hash_syntax.rb
+++ b/lib/rubocop/cop/style/hash_syntax.rb
@@ -28,7 +28,7 @@ module RuboCop
       # * always - forces use of the 3.1 syntax (e.g. {foo:})
       # * never - forces use of explicit hash literal value
       # * either - accepts both shorthand and explicit use of hash literal value
-      # * consistent - like "either", but will avoid mixing styles in a single hash
+      # * consistent - forces use of the 3.1 syntax only if all values can be omitted in the hash
       #
       # @example EnforcedStyle: ruby19 (default)
       #   # bad
@@ -92,17 +92,20 @@ module RuboCop
       #
       # @example EnforcedShorthandSyntax: consistent
       #
-      #   # bad
-      #   {foo: , bar: bar}
-      #
       #   # good
       #   {foo:, bar:}
       #
-      #   # bad
-      #   {foo: , bar: baz}
+      #   # good - can't omit `qux`
+      #   {foo: foo, bar: qux}
       #
-      #   # good
-      #   {foo: foo, bar: baz}
+      #   # bad - `foo' and `bar` can be omitted
+      #   {foo: foo, bar: bar}
+      #
+      #   # bad - `bar` can be omitted
+      #   {foo:, bar: bar}
+      #
+      #   # bad - mixed syntaxes
+      #   {foo:, bar: qux}
       #
       class HashSyntax < Base
         include ConfigurableEnforcedStyle


lib/rubocop/cop/style/hash_syntax.rb
```diff
@@ -92,17 +92,20 @@ module Style
...
-      #   # bad
-      #   {foo: , bar: bar}
-      #
       #   # good
       #   {foo:, bar:}
       #
-      #   # bad
-      #   {foo: , bar: baz}
+      #   # good - can't omit `qux`
+      #   {foo: foo, bar: qux}
```
koic, line 99: After `foo` and `bar`, `baz` is common. Not `qux`.
https://www.rfc-editor.org/rfc/rfc3092.txt

```suggestion
      #   # good - can't omit `baz`
      #   {foo: foo, bar: baz}
```

lib/rubocop/cop/style/hash_syntax.rb
```diff
@@ -92,17 +92,20 @@ module Style
...
       #
-      #   # good
-      #   {foo: foo, bar: baz}
+      #   # bad - `foo' and `bar` can be omitted
```
koic, line 101:
```suggestion
      #   # bad - `foo` and `bar` values can be omitted
```

lib/rubocop/cop/style/hash_syntax.rb
```diff
@@ -92,17 +92,20 @@ module Style
...
+      #   {foo: foo, bar: bar}
+      #
+      #   # bad - `bar` can be omitted
+      #   {foo:, bar: bar}
+      #
+      #   # bad - mixed syntaxes
+      #   {foo:, bar: qux}
```
koic, line 108: Ditto:

```suggestion
      #   # bad - mixed syntaxes
      #   {foo:, bar: baz}
```

koic:
It seems to be written in the order of `# bad`, `# good`. It would be ordered like this:

```ruby
      #
      # @example EnforcedShorthandSyntax: consistent
      #
      #   # bad - `foo` and `bar` can be omitted
      #   {foo: foo, bar: bar}
      #
      #   # bad - `bar` can be omitted
      #   {foo:, bar: bar}
      #
      #   # bad - mixed syntaxes
      #   {foo:, bar: baz}
      #
      #   # good
      #   {foo:, bar:}
      #
      #   # good - can't omit `baz`
      #   {foo: foo, bar: baz}
      #
```

koic:
> Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named {change_type}_{change_description}.md if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.
> ➡️ ❓ Is it required for a documentation change ?

It is unnecessary because it is a document update.

koic:
I've updated and cherry-picked this patch into the master branch. Thank you.
https://github.com/rubocop/rubocop/commit/fa401dc8942491d30c791cbe1ab583f75d196af3

tight:
Sorry, I didn't have time for this PR. Thank you for caring about it 👍

PR #11294, opened 2022-12-16
[Fix #11293] Fix a false negative for `Style/Documentation`

### LINKED ISSUE #8788: Style/Documentation with only macros
I'd like Style/Documentation to accept the following:

```ruby
module RuboCop
  module AST
    # Responsible for compatibility with main gem
    # @api private
    module RuboCopCompatibility
      # ...
    end

    extend RuboCopCompatibility  # => This triggers Style/Documentation: Missing top-level module documentation comment
  end
end
```

I think that (at least) macro calls like `extend` should not trigger it.

### LINKED ISSUE #11293: `Style/Documentation` macro-only check is incorrect
This does not get a lint error:

```ruby
class Thing < ApplicationRecord
  belongs_to :user
end
```

I think it should, because it is a class with legitimate contents, but no documentation.

I think the changes made in https://github.com/rubocop/rubocop/pull/8812 have introduced a false positive.

--------

## Expected behavior

`Style/Documentation` lint error, as the class has no documentation.

## Actual behavior

No lint error.

## RuboCop version

```
1.40.0 (using Parser 3.1.3.0, rubocop-ast 1.24.0, running on ruby 3.1.3) [x86_64-darwin20]
  - rubocop-minitest 0.25.0
  - rubocop-performance 1.15.1
  - rubocop-rails 2.17.3
  - rubocop-thread_safety 0.4.4
```

koic:
Fixes #11293.

This PR fixes a false negative for `Style/Documentation` when using custom macro.

The built-in macros (e.g. `include`, `extend`, `prepend`) and DSL like `belongs_to` should be treated differently. Only `include`, `extend`, and `prepend` will be needed to resolve #8788.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_negative_for_style_documentation.md
+++ b/changelog/fix_false_negative_for_style_documentation.md
@@ -1,0 +1,1 @@
+* [#11293](https://github.com/rubocop/rubocop/issues/11293): Fix a false negative for `Style/Documentation` when using macro. ([@koic][])
--- a/lib/rubocop/cop/style/documentation.rb
+++ b/lib/rubocop/cop/style/documentation.rb
@@ -86,6 +86,11 @@ module RuboCop
           (send nil? {:public_constant :private_constant} ({sym str} _))
         PATTERN
 
+        # @!method include_statement?(node)
+        def_node_matcher :include_statement?, <<~PATTERN
+          (send nil? {:include :extend :prepend} const)
+        PATTERN
+
         def on_class(node)
           return unless node.body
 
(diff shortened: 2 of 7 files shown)


bbatsov:
I think it will also be a good idea to update the docs to reflect more precisely what's allowed when it comes to macros, as right now the documentation is extremely basic in this regard.

bbatsov:
@koic Ping :-)

koic:
I'm not sure if this change should be documented. It essentially looks like #8812's bug against #8788.
This is because it seems redundant to document that `extend Bar` does not require comments.

```ruby
module Foo
  # documentation.
  module Bar
  end

  extend Bar
end
```

So, I think it's clear that this cop doesn't require documentation for `extend Bar`.

bbatsov:
Okay, fair enough.

ghiculescu:
Thank you!!

PR #11301, opened 2022-12-17
Print `memory_profiler` report to the file

fatkodima:
Previously, it was just dumping a lot of the lines into the stdout making it hard to work with.


koic:
This looks good! Thanks!

PR #11310, opened 2022-12-21
[Fix #11309] Fix a false positive for `Style/RedundantStringEscape`

### LINKED ISSUE #11309: autocorrecting Style/RedundantStringEscape changes program behavior if there are two redundant escapes
When Style/RedundantStringEscape sees a string like this

```ruby
"\#\{foo}"
```

it identifies both `\#` and `\{` as redundant string escapes, and if asked to autocorrect, will unescape _both_, changing the meaning of the program.

--------

## Expected behavior

Because Style/RedundantStringEscape promises safe autocorrection, I would expect that it would change the offending string to either`"#\{foo}"` (removing first slash) or `"\#{foo}"` (removing second slash).

## Actual behavior

It actually removes both, resulting in `#{foo}`. This causes the program to interpolate the value of `foo`.

Assuming RuboCop is installed, this script will demonstrate the problem:

```sh
#!/bin/bash -x

echo 'puts "\#\{foo}"' > demo.rb
ruby demo.rb
rubocop --debug --autocorrect --only Style/RedundantStringEscape demo.rb
ruby demo.rb
```

Here is the output:

```
+ echo 'puts "\#\{foo}"'
+ ruby demo.rb
#{foo}
+ rubocop --debug --autocorrect --only Style/RedundantStringEscape demo.rb
For /Users/benragheb/Projects/rubocop-bug: Default configuration from /Users/benragheb/.rvm/gems/ruby-2.6.10/gems/rubocop-1.41.0/config/default.yml
Use parallel by default.
Inspecting 1 file
Scanning /Users/benragheb/Projects/rubocop-bug/demo.rb
Loading cache from /Users/benragheb/.cache/rubocop_cache/5dc345701efc12c82aae8fbc7c2d51bbefd152fa/05097446ed3d3d6bb5a46aa88fe4d53f18d298af/5c20c30dd0b384e0122a1414df3f4ffec395ac2c
C

Offenses:

demo.rb:1:7: C: [Corrected] Style/RedundantStringEscape: Redundant escape of # inside string literal.
puts "\#\{foo}"
      ^^
demo.rb:1:9: C: [Corrected] Style/RedundantStringEscape: Redundant escape of { inside string literal.
puts "\#\{foo}"
        ^^

Finished in 0.1450700000859797 seconds
+ ruby demo.rb
demo.rb:1:in `<main>': undefined local variable or method `foo' for main:Object (NameError)
```

## RuboCop version

```
1.41.0 (using Parser 3.1.3.0, rubocop-ast 1.24.0, running on ruby 2.6.10) [x86_64-darwin21]
```

koic:
Fixes #11309.

This PR fixes a false positive for `Style/RedundantStringEscape` when using a redundant escaped string interpolation `\#\{foo}`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_style_redundant_string_escape.md
+++ b/changelog/fix_a_false_positive_for_style_redundant_string_escape.md
@@ -1,0 +1,1 @@
+* [#11309](https://github.com/rubocop/rubocop/issues/11309): Fix a false positive for `Style/RedundantStringEscape` when using a redundant escaped string interpolation `\#\{foo}`. ([@koic][])
--- a/lib/rubocop/cop/style/redundant_string_escape.rb
+++ b/lib/rubocop/cop/style/redundant_string_escape.rb
@@ -169,9 +169,10 @@ module RuboCop
           # Allow \#{foo}, \#$foo, \#@foo, and \#@@foo
           # for escaping local, global, instance and class variable interpolations
           return true if range.source.match?(/\A\\#[{$@]/)
-
           # Also allow #\{foo}, #\$foo, #\@foo and #\@@foo
           return true if range.adjust(begin_pos: -2).source.match?(/\A[^\\]#\\[{$@]/)
+          # For `\#\{foo} allow `\#` and warn `\{`
+          return true if range.adjust(end_pos: 1).source == '\\#\\{'
 
           false
         end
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #11316, opened 2022-12-21
Fix a false positive for `Naming/BlockForwarding` when the block argument is reassigned

### LINKED ISSUE #11313: Failing example for Naming/BlockForwarding and Lint/UselessAssignment
The following code gets refactored in a wrong way by `rubocop -A` (with `rubocop-rails` installed:

## Source

```ruby
def foo(&block)
  block = proc {} unless block_given?
  pass_to_bar(&block)
end
```

## Misbehavior

- `Naming/BlockForwarding: Use anonymous block forwarding`
- `Lint/UselessAssignment: Useless assignment to variable - block.` but the block is actually used on line 3 (after correction, possibly fixed implicitely when BlockForwarding is fixed)

The code becomes:

```ruby
def foo(&)
  block = proc {} unless block_given?
  pass_to_bar(&)
end
```

--> Line 2 is now ineffective and the `pass_to_bar` crashes then given a nil block.

## RuboCop version

```
1.41.0 (using Parser 3.1.3.0, rubocop-ast 1.24.0, running on ruby 3.1.3) [x86_64-linux]
  - rubocop-rails 2.17.3
```

fatkodima:
Fixes #11313.

`node_parts[0]` is a name of a local variable.

### DIFF
--- a/changelog/fix_block_forwarding_reassignments.md
+++ b/changelog/fix_block_forwarding_reassignments.md
@@ -1,0 +1,1 @@
+* [#11313](https://github.com/rubocop/rubocop/issues/11313): Fix a false positive for `Naming/BlockForwarding` when the block argument is reassigned. ([@fatkodima][])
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

PR #11317, opened 2022-12-22
Add support `@see` tags in our document

ydah:
Resolve: https://github.com/rubocop/rubocop-rspec/issues/772

-----------------

* [-] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/class_structure.rb
+++ b/lib/rubocop/cop/layout/class_structure.rb
@@ -132,7 +132,6 @@ module RuboCop
       #     end
       #   end
       #
-      # @see https://rubystyle.guide#consistent-classes
       class ClassStructure < Base
         include VisibilityHelp
         extend AutoCorrector
--- a/lib/rubocop/cops_documentation_generator.rb
+++ b/lib/rubocop/cops_documentation_generator.rb
@@ -33,7 +33,7 @@ class CopsDocumentationGenerator # rubocop:disable Metrics/ClassLength
     cops.with_department(department).sort!
   end
 
-  def cops_body(cop, description, examples_objects, safety_objects, pars) # rubocop:disable Metrics/AbcSize
+  def cops_body(cop, description, examples_objects, safety_objects, see_objects, pars) # rubocop:disable Metrics/AbcSize, Metrics/ParameterLists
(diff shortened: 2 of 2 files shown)


lib/rubocop/cop/layout/class_structure.rb
```diff
@@ -132,7 +132,6 @@ module Layout
       #     end
       #   end
       #
-      # @see https://rubystyle.guide#consistent-classes
```
ydah, line 135: [memo] It was the same as the output in the `Reference`, so it has been removed here.

lib/rubocop/cops_documentation_generator.rb
```diff
@@ -33,14 +33,15 @@ def cops_of_department(department)
...
-  def cops_body(cop, description, examples_objects, safety_objects, pars) # rubocop:disable Metrics/AbcSize
+  def cops_body(cop, description, examples_objects, safety_objects, see_objects, pars) # rubocop:disable Metrics/AbcSize, Metrics/ParameterLists
     check_examples_to_have_the_default_enforced_style!(examples_objects, cop)
 
     content = h2(cop.cop_name)
     content << required_ruby_version(cop)
     content << properties(cop)
     content << "#{description}\n"
     content << safety_object(safety_objects) if safety_objects.any? { |s| !s.text.blank? }
+    content << see_object(see_objects) if see_objects.any?
```
ydah, line 44: [memo] I'm putting `@see` here because it often comes before `@example`[^1], but open to suggestions for other locations.

[^1]: Currently, RuboCop RSpec does so.

koic:
I doubt about extension to this document generation. The cops (i.e. published cops in lib/rubocop/cop) document could be represented as `StyleGuide` or `Reference` instead of `@see`. 

So I think the only thing that matters in this PR is the removal of `@see https://rubystyle.guide#consistent-classes`.

ydah:
Ah, I see. Indeed it is, we can write it in the `Reference`.
One more point I'd like to confirm, if the `@see` tag is used, is it excessive to add it to the `Reference`?

bbatsov:
I'm on the fence about this - but I'll be okay if `@see` gets added to `References` as that's what it is essentially.

bbatsov:
Thanks!

PR #11321, opened 2022-12-23
[Fix #11320] Fix a false positive for `Lint/RequireParentheses`

### LINKED ISSUE #11320: Lint/RequireParentheses and Style/TernaryParentheses contradicting each other
Sample code:

```ruby
self.foo = bar && baz ? 'qux' : 'quux'
```

Running Rubocop against that, you get:

```
example.rb:1:1: W: Lint/RequireParentheses: Use parentheses in the method call to avoid confusion about precedence.
self.foo = bar && baz ? 'qux' : 'quux'
^^^^^^^^^^^^^^^^^^^^^
```

However, correcting it to:

```ruby
self.foo = (bar && baz) ? 'qux' : 'quux'
```

Rubocop now reports:

```
example.rb:1:12: C: [Correctable] Style/TernaryParentheses: Omit parentheses for ternary conditions.
self.foo = (bar && baz) ? 'qux' : 'quux'
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Correcting back-and-forth just keeps you in this loop forever.

Interesting wildcard: the first offense happens when the sample code is `self.foo =`, but **not** when it's `foo =`, which makes me think the first offense is a bug.

--------

## Expected behavior

Based on the docs for `Lint/RequireParentheses`, it seems like it's only supposed to apply when predicate methods are used, which isn't happening in the sample code, so it probably shouldn't apply here.

## Actual behavior

See above.

See above.

## RuboCop version

```
1.41.1 (using Parser 3.1.3.0, rubocop-ast 1.24.0, running on ruby 3.0.4) [x86_64-darwin21]
```

koic:
Fixes #11320.

This PR fixes a false positive for `Lint/RequireParentheses` when assigning ternary operator.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positive_for_lint_require_parentheses.md
+++ b/changelog/fix_false_positive_for_lint_require_parentheses.md
@@ -1,0 +1,1 @@
+* [#11320](https://github.com/rubocop/rubocop/issues/11320): Fix a false positive for `Lint/RequireParentheses` when assigning ternary operator. ([@koic][])
--- a/lib/rubocop/cop/lint/require_parentheses.rb
+++ b/lib/rubocop/cop/lint/require_parentheses.rb
@@ -46,7 +46,9 @@ module RuboCop
         private
 
         def check_ternary(ternary, node)
-          return if node.method?(:[]) || !ternary.condition.operator_keyword?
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #11323, opened 2022-12-23
[Fix #11322] Fix message for Lint/UnusedMethodArgument

### LINKED ISSUE #11322: Lint/UnusedMethodArgument should say "If it's unnecessary, remove it."

I have a code like this:

```ruby
def double(number, other)
  number * 2
end
```

Then, RuboCop warns:

```
foo.rb:17:20: W: [Correctable] Lint/UnusedMethodArgument: Unused method argument - other. If it's necessary, use _ or _other as an argument name to indicate that it won't be used.
def double(number, other)
                   ^^^^^
```

If I run `rubocop -A`, it will be fixed like this:

```ruby
def double(number, _other)
  number * 2
end
```

However, that is not a good fix. The second argument should be removed in this case:

```ruby
def double(number)
  number * 2
end
```

RuboCop should determine the argument is necessary or not. If it's necessary, it would be okay to rename to `_other`. But if not, it should be removed instead of renaming it.

However, I guess it would be tough to say necessary or not for RuboCop. At least the message should be changed like this:

> If it's necessary, use _ or _other as an argument name to indicate that it won't be used. If it's unnecessary, remove it.

## Additional context

As a programming mentor, I've seen unnecessary arguments like `_others` so many times since newbies tend to apply `rubocop -A` blindly. It might be good idea to disable auto-correct by default because they should consider by themselves to rename or remove warned arguments.

JunichiIto:
Closes https://github.com/rubocop/rubocop/issues/11322

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/unused_method_argument.rb
+++ b/lib/rubocop/cop/lint/unused_method_argument.rb
@@ -100,7 +100,8 @@ module RuboCop
 
           unless variable.keyword_argument?
             message << " If it's necessary, use `_` or `_#{variable.name}` " \
-                       "as an argument name to indicate that it won't be used."
+                       "as an argument name to indicate that it won't be used. " \
+                       "If it's unnecessary, remove it."
           end
 
           scope = variable.scope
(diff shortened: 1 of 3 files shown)


changelog/fix_message_for_lint_unused_method_argument.md
```diff
@@ -0,0 +1 @@
+* [#11322](https://github.com/rubocop/rubocop/issues/11322): Fix message for `Lint/UnusedMethodArgument`. ([@JunichiIto][])
```
koic, line 1: Changelog is not required. Can you remove the changelog entry?

JunichiIto: Okay, I removed it.

bbatsov:
This message reads like a novel at this point! :D

PR #11328, opened 2022-12-24
Support autocorrection for percent literals in `Style/ConcatArrayLiterals`

### LINKED ISSUE #11325: Style/ConcatArrayLiterals autocorrection isn't working
## Expected behavior

Docs says this cop supports autocorrection:

> image

## Actual behavior

Autocorrection is not working (not even with `-A` flag):

```bash
$ rubocop -A

# (...)

Offenses:

file.rb:33:17: C: Style/ConcatArrayLiterals: Use push with elements as arguments without array brackets instead of concat(%i[subject preview_text audience body sender_id]).
          attrs.concat(%i[subject body])
                ^^^^^^^^^^^^^^^^^^^^^^^^

162 files inspected, 1 offense detected
```

Here's the tested Ruby code:

```ruby
attrs.concat(%i[subject body])
```

## RuboCop version

```
$ rubocop -V
1.41.1 (using Parser 3.1.3.0, rubocop-ast 1.24.0, running on ruby 3.1.3) [arm64-darwin22]
  - rubocop-performance 1.15.1
  - rubocop-rails 2.17.3
  - rubocop-rspec 2.16.0
```

fatkodima:
Fixes #11325.

`%W`/`%I` literals with interpolations inside are skipped for autocorrection, because they are rare and could further complicate this cop's logic. Other cases are pretty common.

### DIFF
--- a/spec/rubocop/cop/style/concat_array_literals_spec.rb
+++ b/spec/rubocop/cop/style/concat_array_literals_spec.rb
@@ -3,41 +3,61 @@
 RSpec.describe RuboCop::Cop::Style::ConcatArrayLiterals, :config do
   it 'registers an offense when using `concat` with single element array literal argument' do
     expect_offense(<<~RUBY)
-      item.concat([item])
-           ^^^^^^^^^^^^^^ Use `push(item)` instead of `concat([item])`.
+      arr.concat([item])
+          ^^^^^^^^^^^^^^ Use `push(item)` instead of `concat([item])`.
     RUBY
 
     expect_correction(<<~RUBY)
-      item.push(item)
+      arr.push(item)
     RUBY
   end
 
   it 'registers an offense when using `concat` with multiple elements array literal argument' do
     expect_offense(<<~RUBY)
-      item.concat([foo, bar])
-           ^^^^^^^^^^^^^^^^^^ Use `push(foo, bar)` instead of `concat([foo, bar])`.
+      arr.concat([foo, bar])
+          ^^^^^^^^^^^^^^^^^^ Use `push(foo, bar)` instead of `concat([foo, bar])`.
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/concat_array_literals_spec.rb
```diff
@@ -37,7 +37,27 @@
   it 'registers an offense when using `concat` with single element `%i` array literal argument' do
     expect_offense(<<~RUBY)
       item.concat(%i[item])
-           ^^^^^^^^^^^^^^^^ Use `push` with elements as arguments without array brackets instead of `concat(%i[item])`.
+           ^^^^^^^^^^^^^^^^ Use `push(:item)` instead of `concat(%i[item])`.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      item.push(:item)
```
vlad-pisanov, line 44: minor -- I find it confusing that this test file uses `item` for both the list and the item being inserted. `list.push(:item)` would make more sense

fatkodima: Agreed, but this is how this is done for this entire spec file. Do not want to change the lines history and make diff bigger than is needed.
But if requested, will do with no problems..

bbatsov: Feel free to add a second commit that changes this. I'd suggest using something like `arr` instead of `item`.

fatkodima: Updated.

koic:
This looks good to me. Can you squash your commits into one?

fatkodima:
Squashed.

koic:
Thanks!

PR #11329, opened 2022-12-24
Accept simple freezed constants in `Layout/ClassStructure` and correctly handle class methods

### LINKED ISSUE #8034: Auto-correction of Layout/ClassStructure cop breaks codes

I ran `rubocop -a --only Layout/ClassStrucutre` on the following codes.

``` ruby
# example.rb
class Foo
  C1 = "abc".freeze
  C2 = "efg".freeze

  def self.class_m1
    puts __callee__
  end

  def self.class_m2
    puts __callee__
  end

  def instance_m1
    puts __callee__
  end

  private

  def private_m1
    puts __callee__
  end

  def initialize(args)
    @args = args
  end

  def private_m2
    puts __callee__
  end
end
```

## Expected behavior

I expected no error and expected get following codes.

``` ruby
class Foo
  C1 = "abc".freeze
  C2 = "efg".freeze

  def initialize(args)
    @args = args
  end

  def instance_m1
    puts __callee__
  end

  private

  def private_m1
    puts __callee__
  end

  def private_m2
    puts __callee__
  end

  def self.class_m1
    puts __callee__
  end

  def self.class_m2
    puts __callee__
  end
end
```

## Actual behavior

I encountered the following results and broken codes.

```
$ rubocop -a --only Layout/ClassStructure example.rb
Inspecting 1 file
E

Offenses:

example.rb:8:3: C: [Corrected] Layout/ClassStructure: initializer is supposed to appear before private_methods.
  def initialize(args) ...
  ^^^^^^^^^^^^^^^^^^^^
example.rb:9:3: C: [Corrected] Layout/ClassStructure: initializer is supposed to appear before private_methods.
  def initialize(args) ...
  ^^^^^^^^^^^^^^^^^^^^
example.rb:16:3: C: [Corrected] Layout/ClassStructure: constants is supposed to appear before private_methods.
  C1 = "abc".freeze
  ^^^^^^^^^^^^^^^^^
example.rb:17:3: C: [Corrected] Layout/ClassStructure: constants is supposed to appear before private_methods.
  C1 = "abc".freeze
  ^^^^^^^^^^^^^^^^^
example.rb:23:3: C: [Corrected] Layout/ClassStructure: initializer is supposed to appear before private_methods.
  def initialize(args) ...
  ^^^^^^^^^^^^^^^^^^^^
example.rb:33:1: E: Lint/Syntax: unexpected token $end

1 file inspected, 6 offenses detected, 5 offenses corrected
```

``` ruby
class Foo
  private

  def private_m1
    puts __callee__
  end

  def initialize(args)
    @args = args
  end

  def private_m2
    puts __callee__
  end

  C1 = "abc".freeze
  C2 = "efg".freeze

  def self.class_m1
    puts __callee__
  end

  def self.class_m2
    puts __callee__
  end

  def instance_m1
    puts __callee__
  end

  C1 = "abc".freeze
  C1 = "abc".freezeend
```

## RuboCop version
```
$ [bundle exec] rubocop -V
0.84.0 (using Parser 2.7.1.2, rubocop-ast 0.0.3, running on ruby 2.5.7 x86_64-linux)
```

fatkodima:
This PR fixes 2 problems:
1. Previously, rubocop hasn't correctly detected and autocorrected class methods (as you can see from the 2nd commit), because of the misspelling.
2. Simple `.freeze` constants are considered as "dynamic constants" and are ignored from the consideration. Changed that.

This PR also fixes #8034.

As a follow up to this, I will extend this cop to be able to detect private constants and class methods, to avoid having disables like https://github.com/rubocop/rubocop/blob/20b32e434160716d07182d512be6da3d8ba96d71/lib/rubocop/cop/base.rb#L330-L333 - that constant is used in the method that follows and makes sense to keep it near that method (not at the top as `Layout/ClassStructure` currently suggests, thats why `# rubocop:disable`).

### DIFF
--- a/changelog/fix_class_structure_freezed_consts_and_public_class_methods.md
+++ b/changelog/fix_class_structure_freezed_consts_and_public_class_methods.md
@@ -1,0 +1,1 @@
+* [#11329](https://github.com/rubocop/rubocop/pull/11329): Accept simple freezed constants in `Layout/ClassStructure` and correctly handle class methods. ([@fatkodima][])
--- a/lib/rubocop/config.rb
+++ b/lib/rubocop/config.rb
@@ -21,6 +21,13 @@ module RuboCop
     DEFAULT_RAILS_VERSION = 5.0
     attr_reader :loaded_path
 
+    def self.create(hash, path, check: true)
+      config = new(hash, path)
+      config.check if check
+
+      config
+    end
+
     # rubocop:disable Metrics/AbcSize
     def initialize(hash = {}, loaded_path = nil)
       @loaded_path = loaded_path
@@ -38,13 +45,6 @@ module RuboCop
       @clusivity_config_exists_cache = {}
     end
     # rubocop:enable Metrics/AbcSize
-
-    def self.create(hash, path, check: true)
-      config = new(hash, path)
-      config.check if check
-
-      config
-    end
 
     def loaded_features
       @loaded_features ||= ConfigLoader.loaded_features
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -68,6 +68,64 @@ module RuboCop
         Documentation.url_for(self) if builtin?
       end
 
+      def self.inherited(subclass)
+        super
+        Registry.global.enlist(subclass)
+      end
+
+      # Call for abstract Cop classes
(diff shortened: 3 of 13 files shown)


bbatsov:
Looks good. Thanks!

PR #11330, opened 2022-12-25
Fix an error for `Style/RequireOrder` when using `require` inside `rescue` body

fatkodima:
For the following code
```ruby
begin
rescue
  require "foo"
end
```

I got
```
undefined method `if_type?' for nil:NilClass

          node.if_type? && !node.modifier_form?
              ^^^^^^^^^
rubocop/lib/rubocop/cop/style/require_order.rb:102:in `not_modifier_form?'
rubocop/lib/rubocop/cop/style/require_order.rb:121:in `sibling_node'
rubocop/lib/rubocop/cop/style/require_order.rb:107:in `block in find_previous_older_sibling'
rubocop/lib/rubocop/cop/style/require_order.rb:106:in `each'
rubocop/lib/rubocop/cop/style/require_order.rb:106:in `find'
rubocop/lib/rubocop/cop/style/require_order.rb:106:in `find_previous_older_sibling'
rubocop/lib/rubocop/cop/style/require_order.rb:87:in `on_send'
rubocop/lib/rubocop/cop/commissioner.rb:137:in `public_send'
```

`Node#left_siblings` returns all (sibling) children of the parent node (not only of `AST::Node` type), but we should work only with nodes.

### DIFF
--- a/changelog/fix_require_order_for_require_inside_rescue.md
+++ b/changelog/fix_require_order_for_require_inside_rescue.md
@@ -1,0 +1,1 @@
+* [#11330](https://github.com/rubocop/rubocop/pull/11330): Fix an error for `Style/RequireOrder` when using `require` inside `rescue` body. ([@fatkodima][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

Issue #8366, opened 2020-07-19
Layout/ClassStructure and private constants


marcandre:
I believe the following code should not raise an offense:

```ruby
class Foo
  def foo
  end

  BAR = 42
  private_constant :BAR

  def bar
    # use BAR
  end
end
```

This is made worse by autocorrection that only moves the constant and not the `private_constant` just following it.

nguyenquangminh0711:
@marcandre Note on this issue. Ruby also has `public_constant`. Those lines of codes are totally valid.

```ruby
class Hello
  BAR = 10
  private_constant :BAR

  private

  FOO = 10
  public_constant :FOO
end
```

In this case, I think the behavior would be:
- Ignore constants which have private_constant calls somewhere.
- Add offeneses when meet normal constants or constants with public_constant.

Similarly, methods have the same methods, including: `private`, `public`, `private_class_method`, `public_class_method`. However, I think that methods would be their own places, the same as normal private/public method. So, those cases would be caught. 

```ruby
class A
  class << self
    def a_public_class_method; end

    private

    def a_private_class_method; end
  end

  public_class_method :a_private_class_method
  private_class_method :a_public_class_method

  def a_public_method; end

  private :a_public_method

  def another_public_method; end

  private

  def a_private_method; end

  public :a_private_method
end

A.a_private_class_method
A.new.a_private_method

A.new.another_public_method

A.a_public_class_method # Raise exception
A.new.a_public_method # Raise exception
```

Should we tackle them too? I would love to work on this.

marcandre:
Sorry I didn't reply earlier.

1) I would not deal with methods and constants in the same cop.
2) `private` does not influence the visibility of constants. The only use of `public_constant` is to change the visibility of a constant that was made private with `private_constant`.
3) I don't know the multiple cops we have, there's probably one to add offenses for constants defined after a `private` that lack a `private_constant`, but I'm not sure; that could be useful.

If you want to work on this, I would recommend to first address this current issue, where `Layout/ClassStructure` should ignore constants that are made private with `private_constant`. Interested?

thearjunmdas:
@marcandre - Is this still an issue? I tried to recreate this, but there were no offense generated for the stated code.

marcandre:
Definitely an issue. I have a big PR that I need to complete, which addresses this: https://github.com/rubocop/rubocop/pull/9575

thearjunmdas:
[...]

PR #11332, opened 2022-12-25
Ignore private constants in `Layout/ClassStructure` cop

fatkodima:
Follow up to https://github.com/rubocop/rubocop/pull/11329.
Closes https://github.com/rubocop/rubocop/issues/8366.

Example: often times we need to have a constant to act like a cache (see autocorrected changes in this PR) and it makes sense to make it private and keep it near the method that uses it. Rubocop suggests to move it to the top, which is not much useful. 

The same probably needs to be done for private class methods, as they are currently incorrectly considered as "public class methods". But we need to decide where to place them in the class hierarchy (after the public class methods, or before private instance methods, or ...)?

### DIFF
--- a/lib/rubocop/cop/layout/class_structure.rb
+++ b/lib/rubocop/cop/layout/class_structure.rb
@@ -216,7 +216,7 @@ module RuboCop
         def walk_over_nested_class_definition(class_node)
           class_elements(class_node).each do |node|
             classification = classify(node)
-            next if ignore?(classification)
+            next if ignore?(node, classification)
 
             yield node, classification
           end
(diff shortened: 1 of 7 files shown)


lib/rubocop/cop/layout/class_structure.rb
```diff
@@ -264,6 +267,22 @@ def dynamic_constant?(node)
...
+          return false unless node.casgn_type? && node.namespace.nil?
+          return false unless (parent = node.parent)
+
+          parent.each_child_node(:send) do |child_node|
+            return true if marks_constant_private?(child_node, node.name)
+          end
+          false
+        end
+
+        def marks_constant_private?(node, name)
```
koic, line 280: What about a method name like this?

```suggestion
        def marked_as_private_constant?(node, name)
```

lib/rubocop/cop/layout/class_structure.rb
```diff
@@ -264,6 +267,22 @@ def dynamic_constant?(node)
...
+          name = name.to_s
+          node.method?(:private_constant) &&
+            node.arguments.any? { |arg| (arg.sym_type? || arg.str_type?) && arg.value.to_s == name }
```
koic, line 283: Can you tweak it?

```suggestion
          return false unless node.method?(:private_constant)
          
          node.arguments.any? { |arg| (arg.sym_type? || arg.str_type?) && arg.value == name }
```

fatkodima:
Updated.

koic:
We could consider private class methods separately. Anyway I think this PR is useful so I'm going to merging it.

fatkodima:
> We could consider private class methods separately. 

@koic How do you think, should we just ignore them or be able to configure their placement via `ExpectedOrder` config option?

PR #11333, opened 2022-12-25
Add new `Style/YodaExpression` cop

### LINKED ISSUE #9222: New cop for yoda expressions
Inspired by https://palantir.github.io/tslint/rules/binary-expression-operand-order/

This is in addition to existing `Style/YodaCondition` cop.
```ruby
# bad
y = 1 + x

# good
y = x + 1
```

fatkodima:
Closes #9222.

Considerations:
1. I excluded `*` from the checked binary operators, as it looks fine for me both ways: `1000 * milliseconds` vs `milliseconds * 1000`.
2. I check only for integer values, because of the same reason as above: `0.8 + 0.2 * count` vs `0.2 * count + 0.8`.
3. This cop only forbids yoda expressions. Should it be able to enforce it?
4. When forbiding, which edge cases we should ignore? 
5. Etc.

### DIFF
--- a/lib/rubocop/cop/style/yoda_expression.rb
+++ b/lib/rubocop/cop/style/yoda_expression.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Forbids Yoda expressions, i.e. binary operations (using `*`, `+`, `&`, `|`,
+      # and `^` operators) where the order of expression is reversed, eg. `1 + x`.
+      # This cop complements `Style/YodaCondition` cop, which has a similar purpose.
+      #
+      # @safety
+      #   This cop is unsafe because binary operators can be defined
+      #   differently on different classes, and are not guaranteed to
+      #   have the same result if reversed.
+      #
+      # @example SupportedOperators: ['*', '+', '&'']
+      #   # bad
+      #   1 + x
+      #   10 * y
+      #   1 & z
+      #
+      #   # good
+      #   60 * 24
+      #   x + 1
+      #   y * 10
+      #   z & 1
+      #
+      #   # good
+      #   1 | x
+      #
+      class YodaExpression < Base
+        extend AutoCorrector
+
+        MSG = 'Non literal operand should be first.'
+
+        RESTRICT_ON_SEND = %i[* + & | ^].freeze
+
+        def on_new_investigation
+          @offended_nodes = nil
+        end
+
+        def on_send(node)
+          return unless supported_operators.include?(node.method_name.to_s)
+
+          lhs = node.receiver
+          rhs = node.first_argument
+          return if !lhs.numeric_type? || rhs.numeric_type?
+
+          return if offended_ancestor?(node)
+
+          add_offense(node) do |corrector|
+            corrector.swap(lhs, rhs)
+          end
+
+          @offended_nodes ||= Set.new.compare_by_identity
+          @offended_nodes.add(node)
+        end
+
+        private
+
+        def supported_operators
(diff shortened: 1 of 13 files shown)


lib/rubocop/cop/style/yoda_expression.rb
```diff
@@ -0,0 +1,45 @@
...
+      #   60 * 24
+      #   x + 1
+      #   10 * y
+      #   y * 10
+      #   z & 1
+      #
+      class YodaExpression < Base
+        extend AutoCorrector
+
+        MSG = 'Reverse the order of the operands.'
```
bbatsov, line 29: Perhaps the message should make it clear the intention is for the non-literal operand to go first?

lib/rubocop/cop/style/yoda_expression.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Forbids Yoda expressions,
```
bbatsov, line 6: I'd suggest being more explicit about the exact scope of the cop, as it covers a relatively small number of expressions. Perhaps we should also mention it complements `YogaConditions`.

bbatsov:
> I excluded * from the checked binary operators, as it looks fine for me both ways: 1000 * milliseconds vs milliseconds * 1000.

I'd suggest making the list of operators configurable and apply it to `*` out-of-the-box for consistency's sake.  

> I check only for integer values, because of the same reason as above: 0.8 + 0.2 * count vs 0.2 * count + 0.8.

Seems to me it's fine to work with floats in the same way. 

> This cop only forbids yoda expressions. Should it be able to enforce it?

No, I think that'd be an overkill.

fatkodima:
Made suggested changes - please take a look.

lib/rubocop/cop/style/yoda_expression.rb
```diff
@@ -0,0 +1,70 @@
...
+      #   y * 10
+      #   z & 1
+      #
+      #   # good
+      #   1 | x
+      #
+      class YodaExpression < Base
+        extend AutoCorrector
+
+        MSG = 'Non literal operand should be first.'
```
bbatsov, line 33: Non-literal

I'd also mention the name of the operand explicitly in the message.

bbatsov:
I'll merge it, as I'm planning a release for today. You can do the small improvements I suggested later.

jeppester:
We have the following code for setting a port for capybara:
```rb
config.server_port = 9000 + ENV["TEST_ENV_NUMBER"].to_i
```

Which rubocop now flags.

I personally prefer the code we have now as the test env number will be a very small addition on top of a much larger number. A bit like how sixty-five reads better than five-and-sixty (which is source of much confusion in my local Danish language).

I think it's generally the case when you talk about things that are added together that the largest part is mentioned first. If it's last it's usually a deliberate dramatic or comedic device.

So while I totally understand the idea behind this cop, I'm a little bit unsure if it's practically a good idea, or even if there might be a good way to solve this edge case.

jdelStrother:
Yeah, I find there's a lot of cases this doesn't work so well - don't think any of the offenses that this cop produces in our codebase read better the other way round, and most are actively worse. I'm disabling it for now, but I'm curious to hear from others.

bbatsov:
Ideas for improvements are welcome! I'm also open to just agreeing to disable this cop out-of-the-box.

koic:
I get the idea of this cop. However, it seems hard to suppress cases that confuse users. I think this should be disabled by default so that an intent of code can be respected. I've opened #11380.

PR #11336, opened 2022-12-26
Fix `Style/IdenticalConditionalBranches` to ignore identical leading lines when branch has single child and is used in return context

### LINKED ISSUE #10858: Auto-correction for Style/IdenticalConditionalBranches cop breaks behavior

fatkodima:
Fixes #10858.

We should ignore branches with identical leading lines, if any of them has only one child and the whole `if`/`case`/etc is used in return contexts, like
```ruby
def method
  if something
    foo
  else
    foo
    bar
  end
end
```

because otherwise it will be changed to
```ruby
def method
  foo
  if something
  else
    bar
  end
end
```

and incorrectly return `nil` when `something` is truthy, instead of the value of `foo` (as before).

### DIFF
--- a/changelog/fix_identical_conditional_branches_ignore_empty_left_branches.md
+++ b/changelog/fix_identical_conditional_branches_ignore_empty_left_branches.md
@@ -1,0 +1,1 @@
+* [#10858](https://github.com/rubocop/rubocop/issues/10858): Fix `Style/IdenticalConditionalBranches` to ignore identical leading lines when branch has single child and is used in return context. ([@fatkodima][])
--- a/lib/rubocop/cop/style/identical_conditional_branches.rb
+++ b/lib/rubocop/cop/style/identical_conditional_branches.rb
@@ -136,6 +136,7 @@ module RuboCop
 
         private
 
+        # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
         def check_branches(node, branches)
           # return if any branch is empty. An empty branch can be an `if`
           # without an `else` or a branch that contains only comments.
@@ -144,9 +145,13 @@ module RuboCop
           tails = branches.map { |branch| tail(branch) }
           check_expressions(node, tails, :after_condition) if duplicated_expressions?(node, tails)
 
+          return if last_child_of_parent?(node) &&
+                    branches.any? { |branch| single_child_branch?(branch) }
+
           heads = branches.map { |branch| head(branch) }
           check_expressions(node, heads, :before_condition) if duplicated_expressions?(node, heads)
         end
+        # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
 
         def duplicated_expressions?(node, expressions)
           unique_expressions = expressions.uniq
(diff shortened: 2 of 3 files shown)


koic:
Looks good. Thanks!

PR #11337, opened 2022-12-26
Support pattern matching for `Lint/OutOfRangeRegexpRef` cop

### LINKED ISSUE #10976: Lint/OutOfRangeRegexpRef does not detect captures from pattern matching syntax.
The Lint/OutOfRangeRegexpRef does not seem to recognize regular expressions (which have capture groups) that are used in Ruby's new pattern matching syntax. The cop fails to recognize captures when the regex is used in a `case-in` expression as well as the `=>` array syntax.

Both forms invoke `Regexp#===` under the hood, which still trigger's ruby's `Regexp.last_match` capabilities (as well as setting the 'last-match' special global variables like `$1`). Since the variables _are_ set from the captured groups, I would expect rubocop's linter to recognize them as valid capture references.

--------

## Expected behavior

Ruby's new pattern matching syntax accepts regular expressions and captures groups. It still stores captures in the special globals as documented here: https://ruby-doc.org/core-3.1.2/Regexp.html#class-Regexp-label-Special+global+variables

```rb
case "rubocop"
in /rubo(.*)/ then p $1 # prints "cop"
end
```

```rb
"rubocop" => /rubo(.*)/
p $1 # prints "cop"
```

## Actual behavior

Rubocop doesn't recognize the regex pattern in the "case-in", nor "arrow" pattern matching syntax.

in both cases, it prints:

```
  app/lib/vtiger/webservice/id.rb:6:39: Lint/OutOfRangeRegexpRef: $1 is out of range (no regexp capture groups detected).
```

The two ruby snippets above will reproduce the cop error.

## RuboCop version

```
    rubocop (1.35.0)
    rubocop-ast (1.21.0)
    rubocop-performance (1.14.3)
    standard (1.16.0)
```

fatkodima:
Closes #10976.

Note: This is the first time I familiarized myself with the pattern matching in ruby, so be a little more careful when reviewing this PR.

### DIFF
--- a/lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
+++ b/lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
@@ -68,6 +68,12 @@ module RuboCop
           @valid_ref = regexp_conditions.map { |condition| check_regexp(condition) }.compact.max
         end
 
+        def on_in_pattern(node)
+          regexp_patterns = patterns(node).select(&:regexp_type?)
+
+          @valid_ref = regexp_patterns.map { |pattern| check_regexp(pattern) }.compact.max
+        end
+
         def on_nth_ref(node)
           backref, = *node
           return if @valid_ref.nil? || backref <= @valid_ref
@@ -83,6 +89,19 @@ module RuboCop
         end
 
         private
+
+        def patterns(pattern_node)
+          pattern = pattern_node.node_parts[0]
+
+          case pattern.type
+          when :array_pattern, :match_alt
+            pattern.children
+          when :match_as
+            patterns(pattern)
+          else
+            [pattern]
+          end
+        end
 
         def check_regexp(node)
           return if node.interpolation?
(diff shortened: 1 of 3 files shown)


bbatsov:
Looks good to me.

lib/rubocop/cop/lint/out_of_range_regexp_ref.rb
```diff
@@ -84,6 +90,19 @@ def on_nth_ref(node)
 
         private
 
+        def patterns(pattern_node)
+          pattern = pattern_node.node_parts[0]
+
+          case pattern.type
+          when :array_pattern, :match_alt
```
k-tsj, line 97: @fatkodima @bbatsov 
I may be wrong, as I only glanced at it, but I don't think it supports hash pattern/find pattern/value pattern with the pin operator(`^(expression)`).

```ruby
{a: 'foo'} => {a: /(.)/}
p $1

['foo'] => [*, /(.)/, *]
p $1

'foo' => ^(/(.)/)
p $1
```

The last one may not need to be fixed since it is difficult to fully support and is not a practical code.

Issue #11338, opened 2022-12-27
New cop idea: Use Comparable#clamp, Array#{min,max} instead of comparison operators


pocke:
## Is your feature request related to a problem? Please describe.

Today I suggested using `Comparable#clamp` instead of comparison operators in a private project, like the following.

```ruby
# bad - I feel it's not readable...
user_input > threshold ? threshold : user_input

# good - I can imagine the intention of this code easily!
user_input.clamp(..threshold)
```

I think RuboCop can aid with this problem, but there is no cop to detect this problem.

## Describe the solution you'd like

Add a cop to detect the comparison operators' usage. The cop suggests `user_input.clamp(..threshold)` instead of comparison operators.

### `Array#min`

The following code has the same meaning as the above example also.

```ruby
[user_input, threshold].min
```

Which of these is appropriate depends on the situation. If one of the values is a threshold, `clamp` clarifies the intention. But if both values are user inputs, `min` is better than `clamp`.

```ruby
# It's good
[user_input_a, user_input_b].min

# It's strange...
user_input_a.clamp(..user_input_b)
```

Therefore I think the cop should suggest both `min` and `clamp`.

### `Array#max` and `a.clamp(x..y)`

By the way, the cop can detect cases for `Array#max` and `clamp(lower_threshold..upper_threhold)` (or `clamp(lower_threshold, upper_threhold)`).

## Describe alternatives you've considered

nothing

## Additional context

There are performance differences between comparison operators, `clamp`, and `min`.

```ruby
require 'benchmark/ips'

require 'benchmark'
Benchmark.ips do |x|
  a = 42
  b = 43

  x.report('ops') { a > b ? b : a }
  x.report('min') { [a, b].min }
  x.report('clamp') { a.clamp(..b) }

  x.compare!
end
```

```console
$ ruby -v rubocop-clamp-min-performance.rb
ruby 3.2.0 (2022-12-25 revision a528908271) [arm64-darwin21]
Warming up --------------------------------------
                 ops     2.035M i/100ms
                 min     2.203M i/100ms
               clamp   879.718k i/100ms
Calculating -------------------------------------
                 ops     20.230M (± 1.4%) i/s -    101.765M in   5.031438s
                 min     21.694M (± 1.5%) i/s -    110.157M in   5.078819s
               clamp      8.694M (± 1.9%) i/s -     43.986M in   5.060963s

Comparison:
                 min: 21694180.2 i/s
[...]

PR #11341, opened 2022-12-27
Add new `Style/MinMaxComparison` cop

koic:
Implement part of #11338.

This PR adds new `Style/MinMaxComparison` cop.

This cop Enforces the use of `max` or `min` instead of comparison for greater or less. It is unsafe because even if a value has `<` or `>` method, it is not necessarily `Comparable`.

```ruby
# bad
a > b ? a : b
a >= b ? a : b

# good
[a, b].max

# bad
a < b ? a : b
a <= b ? a : b

# good
[a, b].min
```

-----------------

* [x] Added tests.


lib/rubocop/cop/style/min_max_comparison.rb
```diff
@@ -0,0 +1,59 @@
...
+      #   [a, b].min
+      #
+      class MinMaxComparison < Base
+        extend AutoCorrector
+
+        MSG = 'Use `%<prefer>s` instead.'
+
+        def on_if(node)
+          lhs, operator, rhs = *node.condition
+          return unless %i[> <].include?(operator)
```
vlad-pisanov, line 33: I think this cop would also work with `>=` and `<=`

```ruby
# these can be both replaced with `[a, b].max`
a >  b ? a : b
a >= b ? a : b
```

lib/rubocop/cop/style/min_max_comparison.rb
```diff
@@ -0,0 +1,59 @@
...
+      #   This cop is unsafe because even if a value has `<` or `>` method,
+      #   it is not necessarily `Comparable`.
+      #
+      # @example
+      #
+      #   # bad
+      #   a > b ? a : b
+      #
+      #   # bad
+      #   [a, b].max
```
vlad-pisanov, line 18: this should be `# good` 😄

bbatsov:
Looks good to me. Thanks!

PR #11345, opened 2022-12-28
Distribute Rake Tasks for RuboCop extensions

ydah:
The motivation for this PR is currently that RuboCop RSpec uses the exact same cut_release task as RuboCop does. We thought that by distributing these on the RuboCop side, we could abolish double maintenance.

With this change, on the RuboCop extension side, the following code can be added to the Rakefile to use the Rake Task defined on any RuboCop side.

```ruby
load "rubocop/tasks/cut_release.rake"
```

-----------------

* [-] Added tests.


koic:
I'm interested, but I'm closing it for the following reasons.

- It seems that the core release task and the extension release task are actually different. This is because each repository configuration is different. RuboCop RSpec is not the only extension gem.
- I doubt that code for internal work should be distributed API.
- This may be a small point, but many users wonder about distributing unnecessary code as a gem. Only a few extensions are relevant. Not required for all extensions.

There are probably different approaches that each extension can take. But at least not by distributing these.

PR #11364, opened 2022-12-31
[Fix #11338] Add new `Style/ComparableClamp` cop

### LINKED ISSUE #11338: New cop idea: Use Comparable#clamp, Array#{min,max} instead of comparison operators

koic:
## Summary

Fixes #11338.

This PR adds new `Style/ComparableClamp` cop.
It enforces the use of `Comparable#clamp` instead of comparison by minimum and maximum.

```ruby
# bad
[[x, low].max, high].min

# bad
if x < low
  low
elsif high < x
  high
else
  x
end

# good
x.clamp(low, high)
```

This cop supports autocorrection for `if/elsif/else` bad style only.
Because `ArgumentError` occurs if the minimum and maximum of `clamp` arguments are reversed.
When these are variables, it is not possible to determine which is the minimum and maximum:

```ruby
[1, [2, 3].max].min # => 1
1.clamp(3, 1)       # => min argument must be smaller than max argument (ArgumentError)
```

## Additional Information

- `Comparable#clamp` was introduced in Ruby 2.4
- Part of #11338 was implemented by #11341

-----------------

* [x] Added tests.


pocke:
```ruby
if a < low
  low
elsif high < a
  high
else
  a
end
```

This cop can detect the above code too. Do you have the plan to implement it?

koic:
I had overlooked that cases. I will take it in this PR.

koic:
I've updated the entire PR. In particular, this cop's autocorrection strategy has changed and is documented in this PR and cop description.

bbatsov:
Looks good to me. I had completely forgotten about `clamp`, and I guess I'm not the only one. We should look in the newer Ruby APIs and potentially come up with other similar cops that can promote them.

PR #11384, opened 2023-01-03
Add new `Lint/UselessRescue` cop

### LINKED ISSUE #11359: New cop: `Lint/UselessRescue`
It should detect `rescue` clauses which are only reraising the catched error. This is possibly a bug, or just a useless code.

```ruby
# bad
def foo
  do_something
rescue SomeError
  raise
end

# bad
def foo
  do_something
rescue e
  raise e
end

# good
def foo
  do_something
rescue e
  do_something_more
  raise e
end
```

fatkodima:
Closes #11359.

Tested on rails - 0 offenses, discourse - 1 offense, mastodon- 0 offenses, gitlab - 1 offense.


lib/rubocop/cop/lint/useless_rescue.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        private
+
+        def only_reraising?(resbody_node)
+          body = resbody_node.body
+          return false if body.nil? || !body.send_type? || !body.method?(:raise)
+
+          !body.arguments? ||
+            (body.arguments.size == 1 &&
+              body.first_argument.source == resbody_node.exception_variable&.source)
```
vlad-pisanov, line 58: Should this also handle things like:
```ruby
rescue
  raise $!
```

```ruby
rescue
  raise $ERROR_INFO
```

It's kinda dumb, but I've seen devs do this.

fatkodima: Thanks, updated with these suggestions.

bbatsov: You should also mention them in the doc examples.

fatkodima: Added a mention.

PR #11378, opened 2023-01-03
Fix `OperatorMethodCall` when forwarding arguments

sambostock:
`...` cannot be used as an argument to an operator, so it should not be an offense. For example,

```ruby
def m(...) = obj.==(...)
```

would be autocorrected as

```ruby
def m(...) = obj ==(...)
```

which is a syntax error, as would be

```ruby
def m(...) = obj == (...)
```

and

```ruby
def m(...) = obj == ...
```

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

---

Closes #11377

PR #11386, opened 2023-01-03
Fix a false positive for `Style/OperatorMethodCall`

koic:
This PR fixes a false positive for `Style/OperatorMethodCall` when using anonymous forwarding.

A similar case to #11378. It adds tests for anonymous block forwarding, anonymous rest arguments forwarding, and anonymous keyword rest argument forwarding to pass.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/operator_method_call.rb
+++ b/lib/rubocop/cop/style/operator_method_call.rb
@@ -28,7 +28,7 @@ module RuboCop
           return if node.receiver.const_type?
 
           _lhs, _op, rhs = *node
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/operator_method_call.rb
```diff
@@ -38,10 +38,20 @@ def on_send(node)
 
         private
 
-        def parenthesized_argument?(argument)
+        # Checks for an acceptable case of `foo.+(bar).baz`.
+        def method_call_to_parenthesized_arg?(argument)
```
bbatsov, line 42: I think it's better to use `with` instead of `to` here, as you don't really call "to" arguments, but rather "with" them.

PR #11387, opened 2023-01-04
[Fix #11303] Make `Metrics/ParameterLists` aware of `Struct.new` and `Data.define` blocks

### LINKED ISSUE #11303: Add Metrics/ParameterLists support for Data/Struct initialize

Hello. :wave: I think this is a potential issue when Ruby 3.2.0 drops in a few days.

I'd like to be able to define defaults for a data object with a large set of attributes. Example:

``` ruby
DataDemo = Data.define :one, :two, :three, :four do
  def initialize one: 1, two: 2, three: 3, four: 4
    super
  end
end
```

If I save the above as `demo.rb` and run `rubocop demo.rb`, I'll get the following error:

``` 
Offenses:

test.rb:4:18: C: Metrics/ParameterLists: Avoid parameter lists longer than 3 parameters. [4/3]
  def initialize one: 1, two: 2, three: 3, four: 4
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

With the introduction of the new [Data](https://www.alchemists.io/articles/ruby_data) primitive, you have the ability to define defaults for your attributes if desired. The `Data` class expects you to define _all defaults_ for a `Data` object when using `#initialize` which means the parameter length can get quite long.

In a similar vein, you can do this with a [Struct](https://www.alchemists.io/articles/ruby_structs) as well. The biggest difference being that you _don't have to define all keyword arguments_ so you can simply supply the defaults you need. Example:

``` ruby
StructDemo = Struct.new:one, :two, :three, :four do
  def initialize(one: 1, three: 3, **)
    super
  end
end
```

I think it might be good to allow `#initialize` to not be flagged by `Metrics/ParameterLists` due to this use case for `Data` and `Struct` objects. Something along the lines of the following might be nice:

``` yaml
Metrics/ParameterLists:
  AllowedPatterns:
    - Data#initialize
    - Struct#initialize
  Max: 3
```

I default to `Max: 3` not as a feature request suggestion but only to contrast how I currently use this feature against allowing `Data` and `Struct` to be excluded due to the nature of how these objects are meant to be used. ...but emphasis on only the `#initialize` is important since any other method should be under normal scrutiny.

At the moment, the only alternative is to disable the `Metrics/ParameterLists` or configure `CountKeywordArgs` as `false` but this is too broad of a stroke. I think the value in this work is more specifically related to `Data` and `Struct` objects.

## Additional context

Here's what I'm running locally using [Ruby 3.2.0 RC1](https://www.ruby-lang.org/en/news/2022/12/06/ruby-3-2-0-rc1-released):

``` 
1.40.0 (using Parser 3.1.3.0, rubocop-ast 1.24.0, running on ruby 3.2.0) [arm64-darwin22.2.0]
  - rubocop-performance 1.15.1
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.16.0
  - rubocop-thread_safety 0.4.4
```

koic:
Fixes #11303.

This PR makes `Metrics/ParameterLists` aware of `Struct.new` and `Data.define` blocks.

The number of these `initialize` arguments depends on the number of arguments of `Struct.new` and `Data.define`, so it wouldn't make sense as a metrics.

One abandoned design, the number of `initialize` argument will check no more than the number of `Struct.new` and `Data.define` arguments. However, considering a special case that the `keyword_init` parameter can be included in the number, I think there is no much benefit for the complexity of the design.

So this PR chooses an approach that doesn't always check for the number of `initialize` arguments.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_make_metrics_parameter_lists_aware_of_struct_new_and_data_define.md
+++ b/changelog/change_make_metrics_parameter_lists_aware_of_struct_new_and_data_define.md
@@ -1,0 +1,1 @@
+* [#11303](https://github.com/rubocop/rubocop/issues/11303): Make `Metrics/ParameterLists` aware of `Struct.new` and `Data.define` blocks. ([@koic][])
--- a/lib/rubocop/cop/metrics/parameter_lists.rb
+++ b/lib/rubocop/cop/metrics/parameter_lists.rb
@@ -9,6 +9,20 @@ module RuboCop
       # Keyword arguments can optionally be excluded from the total count,
       # as they add less complexity than positional or optional parameters.
       #
+      # Any number of arguments for `initialize` method inside a block of
+      # `Struct.new` and `Data.define` like this is always allowed:
+      #
+      # [source,ruby]
+      # ----
+      # Struct.new(:one, :two, :three, :four, :five, keyword_init: true) do
+      #   def initialize(one:, two:, three:, four:, five:)
+      #   end
+      # end
+      # ----
+      #
+      # This is because checking the number of arguments of the `initialize` method
+      # does not make sense.
+      #
       # NOTE: Explicit block argument `&block` is not counted to prevent
       # erroneous change that is avoided by making block argument implicit.
       #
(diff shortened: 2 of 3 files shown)


bbatsov:
The proposed solution looks reasonable. Thanks!

PR #11389, opened 2023-01-04
Add autocorrect for `Style/MissingElse`

FnControlOption:
If the enforced style for [`EmptyElse`](https://docs.rubocop.org/rubocop/cops_style.html#styleemptyelse) is `empty`, `MissingElse` is autocorrected like so:

```rb
if cond; foo end
# becomes
if cond; foo else; nil; end
```

If the enforced style is `nil`:

```rb
if cond; foo end
# becomes
if cond; foo else; end
```

No corrections are made if the enforced style is `both`

-----------------

* [x] Added tests.


sambostock (commented):
It looks like all the test cases are on single line (using `;`). How does this behave on multiline conditional blocks?

bbatsov:
Looks good. Thanks!

PR #11403, opened 2023-01-08
Fix bad offense for parenthesised calls in literals

gsamokovarov:
In `omit_parentheses` style of `Style/MethodCallWithArgsParentheses` we issued an offense for code like this:

```ruby
[
  Stat.new("Users Today") do
          ^^^^^^^^^^^^^^^ Omit parentheses for method calls with arguments.
    User.today.count
  end,
  Stat.new("Servers Today") do
          ^^^^^^^^^^^^^^^^^ Omit parentheses for method calls with arguments.
    Server.today.count
  end
]
```
The problem here is that removing the parentheses leads invalid Ruby, and we should allow them:

```ruby
[
  Stat.new "Users Today" do
    User.today.count
  end,
  Stat.new "Servers Today" do
    Server.today.count
  end
]

# => test.rb:2: syntax error, unexpected string literal, expecting ']'
#     Stat.new "Users Today" do
#    test.rb:2: syntax error, unexpected `do', expecting end-of-input
#     Stat.new "Users Today" do
```

We have a check that allows parentheses in calls in literals but we missed the case for the calls with attached blocks.

### DIFF
--- a/changelog/fix_bad_offense_for_parenthesised_calls.md
+++ b/changelog/fix_bad_offense_for_parenthesised_calls.md
@@ -1,0 +1,1 @@
+* [#11403](https://github.com/rubocop/rubocop/pull/11403): Fix bad offense for parenthesised calls in literals for `omit_parentheses` style in `Style/MethodCallWithArgsParentheses`. ([@gsamokovarov][])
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

PR #11404, opened 2023-01-08
[Fix #8990] Make `Style/HashEachMethods` aware of `Thread.current`

### LINKED ISSUE #8990: Style/HashEachMethods is marked as a violation for non-hash objects
## Expected behavior

Rubocop should mark violations only for Hash objects.

## Actual behavior

Rubocop marks any object with a `keys` method and a call to `each` on the keys a violation of this rule.
This cop breaks for the standard library `Thread.current.keys.each`, which does not provide a `each_key`.

Create a file with:

```ruby
Thread.current.keys.each { |k| puts k.to_s }
```

Running rubocop or `rubocop -A` will report violations for this line.
The same is valid for any object with `keys` and `each` called on it:

```ruby
class Vault
  def self.keys
    []
  end
end
```

This type of code will also trigger the violation without having an actual `each_key` method.

## RuboCop version

```
$ rubocop -V
1.0.0 (using Parser 2.7.2.0, rubocop-ast 1.0.1, running on ruby 2.6.6 x86_64-darwin19)
```

koic:
Fixes #8990.

This PR makes `Style/HashEachMethods` aware of built-in `Thread.current`. And `Thread.current` is set to `AllowedReceivers` by default that can be specified for method calls.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_make_style_hash_each_methods_aware_of_thread_current.md
+++ b/changelog/fix_make_style_hash_each_methods_aware_of_thread_current.md
@@ -1,0 +1,1 @@
+* [#8990](https://github.com/rubocop/rubocop/issues/8990): Make `Style/HashEachMethods` aware of built-in `Thread.current`. ([@koic][])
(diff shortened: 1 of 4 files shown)


bbatsov:
Thanks!

PR #11414, opened 2023-01-10
Apply concurrency for GHA workflow `pull_requests` envents only

koic:
~~This reverts commit bf6ea9bd153d33d9e5e263c8c7f584b9314b646a.~~

~~This reverts the setting introduced in #11402.~~

When multiple commits continue, the workflow that wants to know the outcome is also canceled, like this: https://github.com/rubocop/rubocop/actions

github-actions-cuncurrency

~~That "GitHub Actions concurrency" way, if CI has a failing commit, we don't know where it failed on CI. I think this risk reduction should be taken over GitHub Actions concurrency in this repository. There may be repositories where this is useful, but at least not in this repository.~~

To reduce this risk, apply GitHub Actions concurrency only to `pull_requests` events:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value

-----------------

* [ ] Added tests.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -5,7 +5,7 @@ on:
 permissions:  # added using https://github.com/step-security/secure-workflows
   contents: read
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 jobs:
   yamllint:
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -13,7 +13,7 @@ permissions:  # added using https://github.com/step-security/secure-workflows
   contents: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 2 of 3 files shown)


.github/workflows/linting.yml
```diff
@@ -4,9 +4,7 @@ on:
   - workflow_dispatch
 permissions:  # added using https://github.com/step-security/secure-workflows
   contents: read
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+
```
ydah, line 7: The following will ensure that only Pull Request events are cancelled by concurrency, but not all Push events. WDYT?

```suggestion
concurrency:
  group: format('{0}-{1}', ${{ github.workflow }}-${{ github.ref }}, ${{ github.head_ref || github.run_id }})
  cancel-in-progress: true
```

ydah: This is the approach described in the GitHub documentation.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value

ydah: And we are testing in the following repositories and have confirmed that it is canceled on Pull Request but not on Push.
https://github.com/ydah/concurrency-test/actions
concurrency

koic: > The following will ensure that only Pull Request events are cancelled by concurrency, but not all Push events.

I like the idea. But `github.ref` looks unnecessary. `github.head_ref` if PR, otherwise `github.run_id` suffices. I've updated to the below:

```ruby
group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

koic:
I've confirmed the operation of that.

PR #11416, opened 2023-01-10
Have `fix_include_path` make paths relative to Dir.pwd

alexevanczuk:
This PR is intended to fix a bug for how relative paths are calculated for configuration created from using `inherit_from` with a child `.rubocop.yml`.

See https://github.com/rubocop/rubocop/pull/11260#discussion_r1064169254 for more information.

cc @jonas054 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/config_loader_resolver.rb
+++ b/lib/rubocop/config_loader_resolver.rb
@@ -33,7 +33,7 @@ module RuboCop
                       inherit_mode: determine_inherit_mode(hash, k))
           end
           hash[k] = v
-          fix_include_paths(base_config.loaded_path, hash, path, k, v) if v.key?('Include')
+          fix_include_paths(base_config.loaded_path, hash, k, v) if v.key?('Include')
         end
       end
     end
@@ -42,13 +42,12 @@ module RuboCop
     # base configuration are relative to the directory where the base configuration file is. For the
     # derived configuration, we need to make those paths relative to where the derived configuration
     # file is.
-    def fix_include_paths(base_config_path, hash, path, key, value)
+    def fix_include_paths(base_config_path, hash, key, value)
       return unless File.basename(base_config_path).start_with?('.rubocop')
 
       base_dir = File.dirname(base_config_path)
-      derived_dir = File.dirname(path)
       hash[key]['Include'] = value['Include'].map do |include_path|
-        PathUtil.relative_path(File.join(base_dir, include_path), derived_dir)
+        PathUtil.relative_path(File.join(base_dir, include_path), Dir.pwd)
       end
     end
 
--- a/spec/rubocop/config_loader_spec.rb
+++ b/spec/rubocop/config_loader_spec.rb
@@ -283,7 +283,7 @@ RSpec.describe RuboCop::ConfigLoader do
       end
 
       it 'gets an Include that is relative to the subdirectory' do
-        expect(configuration_from_file['Style/StringLiterals']['Include']).to eq(['**/*.rb'])
+        expect(configuration_from_file['Style/StringLiterals']['Include']).to eq(['dir/**/*.rb'])
       end
 
       it 'ignores parent AllCops/Exclude if ignore_parent_exclusion is true' do
@@ -341,7 +341,7 @@ RSpec.describe RuboCop::ConfigLoader do
       end
 
       it 'gets an Include that is relative to the subdirectory' do
-        expect(configuration_from_file['Style/StringLiterals']['Include']).to eq(['../src/**/*.rb'])
+        expect(configuration_from_file['Style/StringLiterals']['Include']).to eq(['src/**/*.rb'])
       end
     end
 
--- a/spec/rubocop/cop/bundler/duplicated_gem_spec.rb
+++ b/spec/rubocop/cop/bundler/duplicated_gem_spec.rb
@@ -5,7 +5,7 @@ RSpec.describe RuboCop::Cop::Bundler::DuplicatedGem, :config do
 
   context 'when investigating Ruby files' do
     it 'does not register any offenses' do
-      expect_no_offenses(<<~RUBY, 'foo.rb')
+      expect_no_offenses(<<~RUBY, '/foo.rb')
         # cop will not read these contents
         gem('rubocop')
         gem('rubocop')
--- a/spec/rubocop/cop/bundler/gem_comment_spec.rb
+++ b/spec/rubocop/cop/bundler/gem_comment_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe RuboCop::Cop::Bundler::GemComment, :config do
 
   context 'when investigating Ruby files' do
     it 'does not register any offenses' do
-      expect_no_offenses(<<~RUBY, 'foo.rb')
+      expect_no_offenses(<<~RUBY, '/foo.rb')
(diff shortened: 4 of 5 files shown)


lib/rubocop/cop/base.rb
```diff
@@ -430,7 +430,7 @@ def file_name_matches_any?(file, parameter, default_result)
         return default_result unless patterns
 
         patterns = FilePatterns.from(patterns)
-        patterns.match?(config.path_relative_to_config(file)) || patterns.match?(file)
+        patterns.match?(config.path_relative_to_config(file)) || patterns.match?(file) || patterns.match?(Pathname.new(file).relative_path_from(Pathname.pwd).to_s)
```
alexevanczuk, line 433: I added this because if we have an include like this:
```yml
Style/SomeCop:
  Include:
    - packs/**/*
```

And a file:
```
Pathname.pwd.join('packs/my_pack/blah.rb')
# /Users/me/path/to/directory/packs/my_pack/blah.rb
```

Then that `Include` wouldn't match that path otherwise.

To put it another way – the first match check checks if the include matches the file when the file is relative to the directory where the config file lives. The second does a straight match, which would work here if the path were absolute, and lastly we need to check if the pattern matches the file relative to the working directory, which mirrors the change above to change `Include` paths to be relative to the `pwd`.

Note that if the `file` doesn't have the same base as `Pathname.pwd`, we get this error:
```
     ArgumentError:
       different prefix: "" and "/Users/alex.evanczuk/workspace/rubocop"
```

However, I don't think it's supported in rubocop to run rubocop against a file that is *not* in the `Pathname.pwd`, so an error seems to make sense here.

When I try to `bundle exec rubocop /path/to/file/in/other/repo`, it leads to strange and unpredictable errors. Therefore, I changed the tests to match to reflect the fact that they should be considered in the same pwd as the one rubocop runs in.

alexevanczuk: I went ahead and rescued this error, for two reasons:
1) In the specs for `windows`, `create_file` was creating a file in a directory that was different than `Dir.pwd`. We should probably fix this, but it made me realize that...
2) In the case where folks have valid use cases for running rubocop on a file in a directory different than the `Dir.pwd`, rather than breaking their workflows, we can simply `rescue` which would fall back to the existing behavior.

jonas054: I don't really understand why the include path in the `.rubocop.yml` you give at the beginning of your comment would have to match paths relative to the current working directory. These path patterns should only be interpreted as relative to the configuration file's directory IMO.

Maybe you have a point that I was unable to grasp, but I tried reverting `base.rb` back to its previous state, i.e. I removed the last part of the condition, and all specs still passed. This means that we don't have a spec that shows why this logic needs to be changed.

alexevanczuk: Thanks @jonas054 , I see what you're saying, and you're exactly right here. I reverted it as well and ran it fully integrated on our application's codebase and found that I didn't need it. I've been on vacation so I forgot why I thought I needed it, but it looks like as you said this is not necessary.

Let me know if you have any other feedback about this PR!

jonas054 (approved):
Looks good to me now! There's a coverage check that failed, but not because the test coverage is any lower than before. Maybe rerunning the job would fix the problem.

bbatsov:
@alexevanczuk Just squash those commits into one and we're good to go.

alexevanczuk:
Squashed and ready to go @bbatsov ! 

@jonas054 Looks like `cc-upload-coverage` is [still failing](https://app.circleci.com/pipelines/github/alexevanczuk/rubocop/18/workflows/9e309624-4098-4232-a840-1d8ca03024bc/jobs/359) with the following errors:

```
#!/bin/bash -eo pipefail
./tmp/cc-test-reporter sum-coverage tmp/codeclimate.*.json --parts 5 --output tmp/codeclimate.total.json
./tmp/cc-test-reporter upload-coverage --input tmp/codeclimate.total.json

Error: response from https://api.codeclimate.com/v1/test_reports.
HTTP 401: You are not authorized for this action
Usage:
  cc-test-reporter upload-coverage [flags]

...

Exited with code exit status 255
CircleCI received exit code 255
```

My guess is that that job is coupled to some authorization token/environment variable/etc. that is present in certain CI accounts but not mine. I also found this very similar issue which may help resolve this so other contributors don't face the same issue! https://github.com/codeclimate/test-reporter/issues/342

I also noticed I didn't see `CC_TEST_REPORTER_ID` set in the environment variables output in CI: https://app.circleci.com/pipelines/github/alexevanczuk/rubocop/18/workflows/9e309624-4098-4232-a840-1d8ca03024bc/jobs/359/parallel-runs/0/steps/0-99

bbatsov:
Probably you're right about the root cause of the CC issue and I guess it won't be present in the main CI job, so I'll just go ahead and merge this PR.

stoivo:
To me it seems like this patch did not fix issue from the reproduction case. https://github.com/alexevanczuk/rubocop-child-rubocop-yml-relative-path-issue. I tested with different version of rubocop. I altered the gemfile file with this 
```rb
# frozen_string_literal: true

source 'https://rubygems.org'

gem 'pry-byebug'
# gem 'rubocop', '1.26.1'
# gem 'rubocop', git: 'https://github.com/rubocop/rubocop.git', ref: '8ef0ed0ef'
# gem 'rubocop', git: 'https://github.com/rubocop/rubocop.git', ref: 'c9ccb2f35'

gem 'rubocop', '1.45.1'
```

It doesn't give offences with any of this versions.

(I hope I am doing something wrong and this is allready fixed, but I don't think so).

I acctualy had this working for a custom cop and after this commit it broke. Here are my reproduction case https://github.com/stoivo/rubocop-bug-reproduction-relative-files. If you checkout the history the first commit gives the warning but not the secong. I left some comment in there which is my workaround.

stoivo:
I can open a new issue but just want to confirm if it is a bug or me doing something wrong

alexevanczuk:
Hey @stoivo that reproduction case ended up changing quite a bit since the original issue as I explored it a bit. I also ended up changing the implementation to address the problem I was trying to solve. This PR actually does solve the issue in our application -- we are able to have a child `.rubocop.yml` that inherits from a parent one and manipulates paths as expected. There may still be some specific use cases that are not solved though.

stoivo:
Can you update https://github.com/alexevanczuk/rubocop-child-rubocop-yml-relative-path-issue so it works, so I can copy from that?

alexevanczuk:
Sure thing @stoivo ! I updated it to match our new implementation where the child rubocop configuration depends on the parent one referencing paths that are relative to the child configuration. This doesn't work properly on the old version of rubocop but works on the new one.

naveg:
@alexevanczuk @jonas054 I believe that this change broke our configuration. We use `inherit_from` to organize our configuration into multiple files in the top-level directory. With this change, we can no longer run rubocop from a child directory, because the Include paths get modified and no longer match.

For our use case, it was correct before. Intuitively, when a derived config inherits from a base config, it's the relative location of the two files that matters, not the pwd. Rubocop could be run from deep inside a directory structure where there are no additional configuration files.

I think the solution from the other thread to use absolute paths might work.

naveg:
This should be suitable for our purposes: https://github.com/rubocop/rubocop/pull/11968

PR #11425, opened 2023-01-11
Fix a false negative on `Lint/Void`

krishanbhasin-shopify:
Co-authored-by: Ryan Quanz <ryan.quanz@shopify.com>
Co-authored-by: Anderson Sequera <anderson.sequera@shopify.com>

`Lint/Void` cop detects nonmutating methods that act in a void context and marks them as an offence. Prior to this PR, if those methods took a block the cop would fail to detect the offence; this PR fixes those false negatives.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_negative_for_methods_that_take_a_block_in_void_context.md
+++ b/changelog/fix_a_false_negative_for_methods_that_take_a_block_in_void_context.md
@@ -1,0 +1,1 @@
+* [#11425](https://github.com/rubocop/rubocop/pull/11425): Fix a false negative for `Lint/Void` when using methods that takes blocks. ([@krishanbhasin-shopify][])
--- a/lib/rubocop/cop/lint/void.rb
+++ b/lib/rubocop/cop/lint/void.rb
@@ -124,7 +124,7 @@ module RuboCop
         end
 
         def check_nonmutating(node)
-          return unless node.send_type? && NONMUTATING_METHODS.include?(node.method_name)
+          return unless NONMUTATING_METHODS.include?(node.method_name)
 
           add_offense(node, message: format(NONMUTATING_MSG, method: node.method_name))
         end
--- a/spec/rubocop/cop/lint/void_spec.rb
+++ b/spec/rubocop/cop/lint/void_spec.rb
@@ -218,6 +218,28 @@ RSpec.describe RuboCop::Cop::Lint::Void, :config do
       RuboCop::Config.new('Lint/Void' => { 'CheckForMethodsWithNoSideEffects' => true })
     end
 
+    it 'registers offense for nonmutating method that takes a block' do
+      expect_offense(<<~RUBY)
+        [1,2,3].map do |n|
+        ^^^^^^^^^^^^^^^^^^ Method `#map` used in void context. Did you mean `#map!`?
+          n.to_s
+        end
+        "done"
+      RUBY
+    end
+
+    context 'Ruby 2.7', :ruby27 do
+      it 'registers offense for nonmutating method that takes a numbered parameter block' do
+        expect_offense(<<~RUBY)
+          [1,2,3].map do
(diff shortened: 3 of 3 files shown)


changelog/fix_a_false_negative_for_methods_that_take_a_block_in_void_context.md
```diff
@@ -0,0 +1 @@
+* [#11425](https://github.com/rubocop/rubocop/pull/11425): Fix an error for `Lint/Void` when using methods that takes blocks. ([@krishanbhasin-shopify][])
```
koic, line 1: Can you tweak it?

```suggestion
* [#11425](https://github.com/rubocop/rubocop/pull/11425): Fix a false negative for `Lint/Void` when using methods that takes blocks. ([@krishanbhasin-shopify][])
```

lib/rubocop/cop/lint/void.rb
```diff
@@ -124,7 +124,8 @@ def check_void_expression(node)
         end
 
         def check_nonmutating(node)
-          return unless node.send_type? && NONMUTATING_METHODS.include?(node.method_name)
+          return unless (node.send_type? || node.block_type?) &&
```
koic, line 127: Can you add the numbered block parameter type?

```suggestion
          return unless (node.send_type? || node.block_type? || node.numblock_type?) &&
```

spec/rubocop/cop/lint/void_spec.rb
```diff
@@ -218,6 +218,16 @@ def foo
...
 
+    it 'registers offense for nonmutating method that takes a block' do
+      expect_offense(<<~RUBY)
+        [1,2,3].map do |n|
+        ^^^^^^^^^^^^^^^^^^ Method `#map` used in void context. Did you mean `#map!`?
+          n.to_s
+        end
+        "done"
+      RUBY
+    end
```
koic, line 229: Can you add the following test?

```suggestion
    end
    
    context 'Ruby 2.7', :ruby27 do
      it 'registers offense for nonmutating method that takes a numbered parameter block' do
        expect_offense(<<~RUBY)
          [1,2,3].map do
          ^^^^^^^^^^^^^^ Method `#map` used in void context. Did you mean `#map!`?
            _1.to_s
          end
          "done"
        RUBY
      end
    end
```

lib/rubocop/cop/lint/void.rb
```diff
@@ -124,7 +124,8 @@ def check_void_expression(node)
         end
 
         def check_nonmutating(node)
-          return unless node.send_type? && NONMUTATING_METHODS.include?(node.method_name)
+          return unless (node.send_type? || node.block_type? || node.numblock_type?) &&
```
serioushaircut, line 127: Could we de-morgan this logic a bit maybe like so
```
wrong_type = !(node.send_type? || node.block_type? || node.numblock_type?)
wrong_method = !NONMUTATING_METHODS.include?(node.method_name)

return if wrong_type
return if wrong_method
```
or with the `unless` and the negation before the returns, but in two steps, I mean

lib/rubocop/cop/lint/void.rb
```diff
@@ -124,7 +124,10 @@ def check_void_expression(node)
         end
 
         def check_nonmutating(node)
-          return unless node.send_type? && NONMUTATING_METHODS.include?(node.method_name)
+          right_node_type = (node.send_type? || node.block_type? || node.numblock_type?)
+          right_method = NONMUTATING_METHODS.include?(node.method_name)
+
+          return unless right_node_type && right_method
```
koic, line 130: It seems that there were redundant conditions. So this is probably better:

```suggestion
          return unless NONMUTATING_METHODS.include?(node.method_name)
```

koic:
Thanks!

Issue #8270, opened 2020-07-08
Cop idea: prefer each over map when return value is not used


okuramasafumi:
## Is your feature request related to a problem? Please describe.

I found a code like this:

```ruby
Model.all.map { do_something_as_side_effect }
```

I think everyone will agree that we should use `each` instead of `map` in this case, since `map` is mainly used for its return value but here it's not used.

This sort of thing can be automated by RuboCop!

## Describe the solution you'd like

When a method named `map` is used and the return value is not used, the cop warns.
It can support auto-correct to replace `map` with `each`.

## Describe alternatives you've considered

Actually, there are some more methods from `Enumerable` module that we use for their return values, such as `select`.
We might let users specify method names to replace with `each`

## Additional context
 
When a user defines their own method named `map`, this cop might produce false positive. However, when it's named as `map`, it clearly shows that it's expected to return value and we care it.

fatkodima:
Duplicate https://github.com/rubocop-hq/rubocop/issues/1052 ?

okuramasafumi:
Thanks I didn't notice this old issue.

koic:
@fatkodima Good catch!
This issue has been closed in the past. Opening as a PR would be welcome. See #1052 for more information. Thank you.

Issue #4305, opened 2017-04-23
Detect useless `map`


backus:
The following code could probably be flagged as using `map` when it should just use `each`.

```ruby
def prune_inactive_images
  images = Image.where("last_viewed_at > ?", 2.years.ago)

  images.map do |image|
    puts "Deleting image #{image.name}"
  end

  images.count
end
```

* The `map` result isn't assigned
* It isn't the last statement in the method

bbatsov:
Seems like a good idea to me.

Drenmi:
Seems very reasonable to me. 🙂 Indicating that you will use the return value and then not using it is confusing.

One thing to watch out for is that the statement following isn't a conditional.

rrosenblum:
> One thing to watch out for is that the statement following isn't a conditional.

I think the bigger thing here is that the return of `map` is not being assigned to anything and the `map` statement isn't the last line of the method.

I think the same thing can be said for `find` and `select` instead of `any?` and `reject` instead of `none?`. These tend to be scoped to being used in if statements.

```ruby
if foo.find { |f| f == 1 }
  something
end
```

koic:
When implementing this cop, I think that it should be considered that method chain is not used. It is a similar reason that the return value is not used.

Issue #1052, opened 2014-04-29
Warning for unused map output


michaelmior:
If the output of a call to `map` is thrown away, then it feels like it would be much better to use `each`.

```ruby
[1, 2, 3].map { |n| puts n }
```

vs

```ruby
[1, 2, 3].each { |n| puts n }
```

bbatsov:
Seems we won't be doing this any time soon. If someone wants to pick it up - PRs are welcome! :-)

PR #11426, opened 2023-01-11
Suggest `each` when `map` or `collect` is used in void context

krishanbhasin-shopify:
Using `map` in a void context can either be replaced with `map!` if mutation of the underlying object is desired, or with `each` when it is only being invoked for the side effects produced by a block passed in.

This PR updates the cop to suggest `each` in instead of `map!`.

Fixes https://github.com/rubocop/rubocop/issues/8270
Fixes https://github.com/rubocop/rubocop/issues/4305
Fixes https://github.com/rubocop/rubocop/issues/1052

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/void.rb
+++ b/lib/rubocop/cop/lint/void.rb
@@ -46,19 +46,23 @@ module RuboCop
         LIT_MSG = 'Literal `%<lit>s` used in void context.'
         SELF_MSG = '`self` used in void context.'
         EXPRESSION_MSG = '`%<expression>s` used in void context.'
-        NONMUTATING_MSG = 'Method `#%<method>s` used in void context. Did you mean `#%<method>s!`?'
+        NONMUTATING_MSG = 'Method `#%<method>s` used in void context. Did you mean `#%<suggest>s`?'
 
         BINARY_OPERATORS = %i[* / % + - == === != < > <= >= <=>].freeze
         UNARY_OPERATORS = %i[+@ -@ ~ !].freeze
         OPERATORS = (BINARY_OPERATORS + UNARY_OPERATORS).freeze
         VOID_CONTEXT_TYPES = %i[def for block].freeze
-        NONMUTATING_METHODS = %i[capitalize chomp chop collect compact
-                                 delete_prefix delete_suffix downcase
-                                 encode flatten gsub lstrip map merge next
-                                 reject reverse rotate rstrip scrub select
-                                 shuffle slice sort sort_by squeeze strip sub
-                                 succ swapcase tr tr_s transform_values
-                                 unicode_normalize uniq upcase].freeze
+        NONMUTATING_METHODS_WITH_BANG_VERSION = %i[capitalize chomp chop compact
+                                                   delete_prefix delete_suffix downcase
+                                                   encode flatten gsub lstrip merge next
+                                                   reject reverse rotate rstrip scrub select
+                                                   shuffle slice sort sort_by squeeze strip sub
+                                                   succ swapcase tr tr_s transform_values
+                                                   unicode_normalize uniq upcase].freeze
+        METHODS_REPLACABLE_BY_EACH = %i[collect map].freeze
+
+        NONMUTATING_METHODS = (NONMUTATING_METHODS_WITH_BANG_VERSION +
+                               METHODS_REPLACABLE_BY_EACH).freeze
 
         def on_block(node)
           return unless node.body && !node.body.begin_type?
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/lint/void.rb
```diff
@@ -47,18 +47,23 @@ class Void < Base
...
-                                 succ swapcase tr tr_s transform_values
-                                 unicode_normalize uniq upcase].freeze
+        NONMUTATING_METHODS_WITH_BANG_VERSION = %i[capitalize chomp chop compact
+                                                   delete_prefix delete_suffix downcase
+                                                   encode flatten gsub lstrip merge next
+                                                   reject reverse rotate rstrip scrub select
+                                                   shuffle slice sort sort_by squeeze strip sub
+                                                   succ swapcase tr tr_s transform_values
+                                                   unicode_normalize uniq upcase].freeze
+        METHODS_REPLACABLE_BY_EACH = %i[collect map].freeze
```
krishanbhasin-shopify, line 63: `collect` feels more tenuous here; I'm happy to remove it and keep the cop suggesting `collect!` instead

koic: `collect` is an alias for `map`, it should be consistent with the same behavior.

changelog/fix_a_false_negative_for_methods_that_take_a_block_in_void_context copy.md
```diff
@@ -0,0 +1 @@
+* [#11426](https://github.com/rubocop/rubocop/pull/11426): Update `Lint/Void` to suggest `each` in place of `map` or `collect` in void context. ([@krishanbhasin-shopify][])
```
koic, line 1: Please remove this changelog entry. Because message-only adjustments have no direct behavioral impact.

lib/rubocop/cop/lint/void.rb
koic: IMHO, I remember seeing `map` used in code where `each` should have been used. OTOH, In my experience `reverse` doesn't look that familiar. So I think it can start with just `map` and `collect`.

lib/rubocop/cop/lint/void.rb
```diff
@@ -47,18 +47,23 @@ class Void < Base
         SELF_MSG = '`self` used in void context.'
         EXPRESSION_MSG = '`%<expression>s` used in void context.'
         NONMUTATING_MSG = 'Method `#%<method>s` used in void context. Did you mean `#%<method>s!`?'
+        REPLACABLE_BY_EACH_MSG = 'Method `#%<method>s` used in void context. Did you mean `#each`?'
```
bbatsov, line 50: replacable -> replaceable :-)

koic: @krishanbhasin-shopify ping :-)

krishanbhasin-shopify: Good spot, thank you!

lib/rubocop/cop/lint/void.rb
```diff
@@ -126,7 +131,11 @@ def check_void_expression(node)
         def check_nonmutating(node)
           return unless NONMUTATING_METHODS.include?(node.method_name)
 
-          add_offense(node, message: format(NONMUTATING_MSG, method: node.method_name))
+          if METHODS_REPLACABLE_BY_EACH.include?(node.method_name)
+            add_offense(node, message: format(REPLACEABLE_BY_EACH_MSG, method: node.method_name))
+          else
+            add_offense(node, message: format(NONMUTATING_MSG, method: node.method_name))
+          end
```
koic, line 138: Can you update it as below? Because the new `REPLACEABLE_BY_EACH_MSG` constant can be removed.

```diff
-        NONMUTATING_MSG = 'Method `#%<method>s` used in void context. Did you mean `#%<method>s!`?'
-        REPLACEABLE_BY_EACH_MSG = 'Method `#%<method>s` used in void context. Did you mean `#each`?'
+        NONMUTATING_MSG = 'Method `#%<method>s` used in void context. Did you mean `#%<suggest>s`?'```
```

```ruby
        def check_nonmutating(node)
          method_name = node.method_name
          return unless NONMUTATING_METHODS.include?(method_name)

          suggest = METHODS_REPLACABLE_BY_EACH.include?(method_name) ? 'each' : "#{method_name}!"
          add_offense(node, message: format(NONMUTATING_MSG, method: method_name, suggest: suggest))
        end
```

bbatsov:
Thanks!

PR #11428, opened 2023-01-11
Apply value omission exceptions in super invocations

gsamokovarov:
Invocation to `super` were not covered by the special cases of value omissions that method calls have.

We ended up with a bad autocorrection in a super invocation. We got:

```ruby
return super current_user, primary: primary unless condition
```

autocorrected to:

```ruby
return super current_user, primary: unless condition
```

The code above is invalid Ruby. If we apply them method call edge cases to super invocations we end up with valid Ruby.


koic:
Thanks!

PR #11431, opened 2023-01-11
Fix a crash in Style/HashSyntax

gsamokovarov:
The crash used to happen when an offense is registered in method receiving hash literals. Here is real world example that results in a crash inside RuboCop:

```ruby
attributes = {account: account, integ_name: shop_name, primary_integ_id: primary_integ_id}.merge attributes
```

The crash was:

```
undefined method `pairs' for s(:lvar, :attributes):RuboCop::AST::Node

            corrector.insert_after(last_argument, ')') if node == last_argument.pairs.last
                                                                  ^^^^^^^^^^^^^
```


bbatsov:
Looks good. Thanks!

Issue #11399, opened 2023-01-06
Cop idea: minimize number of negations in a condition


vlad-pisanov:
Currently, `Style/NegatedIf` (and `Style/NegatedUnless`) recommend `if x` over `unless !x`, which is good because negations are less natural for the reader.

What might be useful is a cop that goes a step further and recommends `if x || y` over `unless !x && !y`. That is, a cop that reduces the number of negations in a boolean expression as much as possible by leveraging DeMorgan's Rule.

In its basic form, it could look for conjunctions (or disjunctions) where _all_ terms are negated:

```ruby
# bad (all terms negated)
unless !x && !y && !z

# good (all terms positive)
if x || y || z
```

A more complex version could look for conjunctions (or disjunctions) where there are _more negations_ than positives:
```ruby
# bad (3 negations, 1 positive)
unless !x || !y || !z || w

# good (3 positives, 1 negation)
if x && y && z && !w
```

Does this sound reasonable? 🤔

fatkodima:
I had a similar suggestion a while ago - https://github.com/rubocop/rubocop/issues/8879. Just need to implement 😄

fatkodima:
Valid.

PR #11432, opened 2023-01-12
Add new `Style/InvertibleUnlessCondition` cop

### LINKED ISSUE #11410: New cop: `Style/RedundantUnless`
It is a known fact, that `unless` adds cognitive overhead. Personally, I always try to avoid it if `if` looks better/simpler. 
This cop will detect places where `unless` is completely unneeded and can be replaced by `if`.

I was thinking, that there is a similar cop already, but was not able to find it.

For example,
```ruby
# bad
foo unless !bar
foo unless x != y
foo unless x >= 10
foo unless x.positive?

# good
foo if bar
foo if x == y
foo if x < 10
foo if x.negative?
```

I saw at least many offenses in the rubocop itself. Here are only a few examples: https://github.com/rubocop/rubocop/blob/1e5f19a9afddb0e5123c86efb0f5957431c7a6d4/lib/rubocop/cop/layout/trailing_empty_lines.rb#L59
https://github.com/rubocop/rubocop/blob/1e5f19a9afddb0e5123c86efb0f5957431c7a6d4/lib/rubocop/cop/layout/multiline_assignment_layout.rb#L99
https://github.com/rubocop/rubocop/blob/1e5f19a9afddb0e5123c86efb0f5957431c7a6d4/lib/rubocop/cop/metrics/parameter_lists.rb#L82

Related: https://github.com/rubocop/rubocop/issues/11399

There is a kinda related cop already (`Style/NegatedUnless`), but it only checks for simple negations with `!`.

fatkodima:
Closes #11410.

Tested on `rails/rails` - all the offenses looks valid to me.

As not everyone would agree that the new version becomes better (I am *not* one of them, I would rather prefer `if` over `unless` in any case), I decided to accept by default cases like the following:
```ruby
foo unless x == y # otherwise would be converted to `foo if x != y`
foo unless x =~ y # otherwise would be converted to `foo if x !~ y`
foo unless x < Foo # checking for class inheritance, `x >= Foo` looks confusing imo?
```


bbatsov:
I don't really like many of the changes this cop does, so I'd suggest to have it be disabled by default. E.g. I'd never make this change as I don't see any readability gains from it:

```
# before
return [] unless lockfile.dependencies.any?

# after
return [] if lockfile.dependencies.none?
```

I get that some people might dislike `unless`, but I think that only when the condition was negated there are some real gains.

fatkodima:
Instead of completely disabling, we can tweak `InverseMethods` config, for example: remove `:any?: :none?` and `:zero?: :nonzero?` conversions, so we do not get `if` with negations, like in the example you provided.

If you insist on disabling by default, will do that with no problems.

bbatsov:
I think the config makes sense for whoever wants to use the cop (that's why I think the config is fine as it is), but I don't really have issues with `unless` myself, so I'm interested in applying this to RuboCop's codebase. 

Keep in mind that we've been promising people for a while to enable fewer controversial cops (if I could start all over I'd probably disable half of all cops by default), that's why I think it's best to leave it disabled by default as well and leave it to whoever needs it to enable it.

fatkodima:
> so I'm interested in applying this to RuboCop's codebase.

Sorry, did you mean just to introduce the cop or to also enable it in the rubocop itself (in `.rubocop.yml` file)?

koic:
IMO, RuboCop’s codebase doesn't need to be different from the default, so the .rubocop.yml shouldn't be applied differently than the default for the new rule.

vlad-pisanov:
My 2c -- I think this cop should only discourage `unless` with an invertible _negated_ condition, not all invertible `unless` statements (similar to my proposal #11399).

```ruby
unless !x         # double negation ❌
unless x.none?    # double negation ❌
unless x.nonzero? # double negation ❌
unless x          # ok ✔️
unless x.any?     # ok ✔️
unless x.zero?    # ok ✔️
```

vlad-pisanov:
@koic yes, it's a somewhat different rule, but I feel it's driven by the same underlying sentiment -- readable code should negate things as few times as possible. 🙂

bbatsov:
> Sorry, did you mean just to introduce the cop or to also enable it in the rubocop itself (in .rubocop.yml file)?

I made a typo and forgot a "not". :-)

```ruby
unless !x         # double negation ❌
unless x.none?    # double negation ❌
unless x.nonzero? # double negation ❌
```

Frankly, only the first one reads like a true negation to me. Yeah, you can `x unless none` and `x if some` are pretty much the same in terms of readability for me. I guess I'm more inclined to agree for `nonzero` as that is one pretty weird predicate, but my point stands - this is way too subjective to be encouraged broadly. 

Funny enough the cop this make the following suggestion to flip an `unless`:

```
# original
return false unless (exponent.to_i % 3).zero?
# changed
return false if (exponent.to_i % 3).nonzero?
```

So here, we actually added a negation and made the predicate heavier to process mentally.

fatkodima:
Made the cop disabled by default and reverted cop corrections.

@bbatsov 
> Funny enough the cop this make the following suggestion to flip an unless:
> 
> ```ruby
> # original
> return false unless (exponent.to_i % 3).zero?
> # changed
> return false if (exponent.to_i % 3).nonzero?
> ```

It would be easier to process (for me) if the changed version was written like `return false if exponent.to_i % 3 != 0`

But this is a style cop, so not everyone would like it.

@vlad-pisanov 
> My 2c -- I think this cop should only discourage unless with an invertible negated condition, not all invertible unless statements (similar to my proposal https://github.com/rubocop/rubocop/issues/11399).

It has a flexible `InverseMethods`, which can be configured to to handle this (for example, to not allow `zero?` => `nonzero?` conversion).

I will be happy to apply more flexibility, if there would be a need.

lib/rubocop/cop/style/invertible_unless_condition.rb
```diff
@@ -0,0 +1,112 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for usages of `unless` which can be replaced by `if` with inverted condition.
```
bbatsov, line 6: Probably it's a good idea to mention here that the idea is that code without `unless` is easier to read, but that's subjective that's why the cop is disabled by default.

bbatsov:
I'm also wondering if a more generic version of this cop that suggests replacing negated methods with their complementary counterparts won't be useful.

lib/rubocop/cop/style/invertible_unless_condition.rb
fatkodima: Added.

fatkodima:
There are already `NegatedIf`, `NegatedUnless`, `NegatedWhile`, `InverseMethods` (this looks the most like what you suggested) cops with a related purpose. Do we need to check something more that I missed? 🤔

bbatsov:
Yeah, seems I was thinking of `InverseMethods`, but I had forgotten how we'd named it. Too many cops to remember them all! 😆

PR #11442, opened 2023-01-13
Fix a crash during anonymous rest argument forwarding

gsamokovarov:
For `Style/MethodCallWithArgsParentheses` with style of `omit_parentheses`, we can crash RuboCOp with the following code:

```ruby
hash.should eq({})
```

Further more, I moved this check to the OmitParentheses module where it belongs.


koic:
Thanks!

PR #11453, opened 2023-01-15
Use `Node#argument_type?` from `rubocop-ast`

fatkodima:


lib/rubocop/cop/metrics/utils/abc_size_calculator.rb
```diff
@@ -25,10 +25,7 @@ class AbcSizeCalculator
           # > http://c2.com/cgi/wiki?AbcMetric
           CONDITION_NODES = CyclomaticComplexity::COUNTED_NODES.freeze
 
-          # TODO: move to rubocop-ast
-          ARGUMENT_TYPES = %i[arg optarg restarg kwarg kwoptarg kwrestarg blockarg].freeze
```
koic, line 29: This is a nice DRY as it was drifting away from the RuboCop AST's definition:
https://github.com/rubocop/rubocop-ast/blob/v1.24.1/lib/rubocop/ast/node.rb#L80-L81

PR #11454, opened 2023-01-15
Remove stale TODO

fatkodima:
The author forgot to delete it in https://github.com/rubocop/rubocop/pull/11054.


koic:
Good catch!

PR #11456, opened 2023-01-15
Fix value omissions in `yield` invocations

gsamokovarov:
I #11428, I have tried to fix the value omission syntax in `super` invocations. I have missed `yield` invocations too. They can have value omissions too.

### DIFF
--- a/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
+++ b/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
@@ -54,7 +54,7 @@ module RuboCop
               next corrector.replace(node, replacement)
             end
 
-            white_spaces = range_between(def_node.loc.selector.end_pos,
+            white_spaces = range_between(def_node.selector.end_pos,
                                          def_node.first_argument.source_range.begin_pos)
             corrector.replace(white_spaces, '(')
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
```diff
@@ -101,15 +101,18 @@ def def_node_that_require_parentheses(node)
...
 
-        def_node unless def_node && def_node.arguments.empty?
+        DefNode.new(def_node) unless def_node && def_node.arguments.empty?
       end
 
       def find_ancestor_method_dispatch_node(node)
         ancestor = node.parent.parent
 
-        ancestor if (ancestor&.call_type? || ancestor&.super_type?) && !brackets?(ancestor)
+        return unless ancestor&.call_type? || ancestor&.super_type? || ancestor&.yield_type?
```
koic, line 112: Can you update it to the below?

```ruby
return unless (ancestor = node.parent.parent)
return unless ancestor.call_type? || ancestor.super_type? || ancestor.yield_type?
```

gsamokovarov: Done ✅

koic:
Thanks!

PR #11459, opened 2023-01-16
Make `Lint/UselessRuby2Keywords` aware of conditions

splattael:
The following code is not flagged yet:

```ruby
  def foo(**args)
  end
  ruby2_keywords :foo if respond_to?(:ruby2_keywords, true)
```

This PR fixes this issue.

Real world example: https://gitlab.com/gitlab-org/gitlab/-/blob/4598296ccc6d31414dd22d288882603d23a12296/config/initializers/6_labkit_middleware.rb#L13. Discovered during enabling this rule in https://gitlab.com/gitlab-org/ruby/gems/gitlab-styles/-/merge_requests/160.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/useless_ruby2_keywords.rb
+++ b/lib/rubocop/cop/lint/useless_ruby2_keywords.rb
@@ -98,12 +98,20 @@ module RuboCop
           return unless node.parent
 
           method_name = sym_node.value
-          definition = node.parent.each_child_node.detect { |n| method_definition(n, method_name) }
+          definition = find_method_definition(node, method_name)
 
           return unless definition
           return if allowed_arguments(definition.arguments)
 
           add_offense(node, message: format(MSG, method_name: method_name))
+        end
+
+        def find_method_definition(node, method_name)
+          node.each_ancestor.lazy.map do |ancestor|
+            ancestor.each_child_node(:def, :block, :numblock).find do |child|
+              method_definition(child, method_name)
+            end
+          end.find(&:itself)
         end
 
         # `ruby2_keywords` is only allowed if there's a `restarg` and no keyword arguments
--- a/changelog/fix_make_lint_useless_ruby2_keywords_aware_of.md
+++ b/changelog/fix_make_lint_useless_ruby2_keywords_aware_of.md
@@ -1,0 +1,1 @@
+* [#11459](https://github.com/rubocop/rubocop/pull/11459): Make `Lint/UselessRuby2Keywords` aware of conditions. ([@splattael][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/lint/useless_ruby2_keywords.rb
```diff
@@ -98,14 +98,22 @@ def inspect_sym(node, sym_node)
           return unless node.parent
 
           method_name = sym_node.value
-          definition = node.parent.each_child_node.detect { |n| method_definition(n, method_name) }
+          definition = find_method_definition(node, method_name)
```
splattael, line 101: Previously we only checked node's parent which is `begin` and not `def`, `block` or `numblock` in case of:

```ruby
def foo
end
ruby2_keywords :foo if respond_to?(:ruby2_keywords)
```

```ruby
s(:begin,
  s(:def, :foo,
    s(:args), nil),
  s(:if,
    s(:send, nil, :respond_to?,
      s(:sym, :ruby2_keywords)),
    s(:send, nil, :ruby2_keywords
      s(:sym, :foo)), nil))
```

lib/rubocop/cop/lint/useless_ruby2_keywords.rb
```diff
@@ -98,14 +98,22 @@ def inspect_sym(node, sym_node)
...
 
           return unless definition
           return if allowed_arguments(definition.arguments)
 
           add_offense(node, message: format(MSG, method_name: method_name))
         end
 
+        def find_method_definition(node, method_name)
+          node.each_ancestor.lazy.map do |ancestor|
```
splattael, line 110: Because we still support Ruby 2.6 for windows `filter_map` is not available. See https://github.com/rubocop/rubocop/actions/runs/3929998645/jobs/6719462639

With this we could refactor to:

```ruby
def find_method_definition(node, method_name)
  node.each_ancestor.lazy.filter_map do |ancestor|
    ancestor.each_child_node(:def, :block, :numblock).find do |child|
      method_definition(child, method_name)
    end
  end.first
end
```

koic: FYI, `filter_map` method has been introduced in Ruby 2.7. So, it's not a Windows platform-specific issue.

koic:
Thanks!

lib/rubocop/cop/lint/useless_ruby2_keywords.rb
splattael: @koic Yes, you are right and sorry for implying it was the case. It was unintentional. My main question was why we support Ruby 2.6 only for windows? :shrug:

koic: Ruby 2.6 runtime also supports non-Windows platforms. CI on other OS would surely have failed on CircleCI.

Issue #11462, opened 2023-01-16
Over indentation when autocorrecting `Layout/FirstHashElementLineBreak`


sambostock:
In certain circumstances, when autocorrecting `Layout/FirstHashElementLineBreak`, the resulting `Hash` is way over indented.

I have found this tricky to minimally reproduce, as it seems to involve multiple cops. I was able to distill it down to this:

```yaml
# .rubocop.yml
AllCops:
  SuggestExtensions: false
  DisabledByDefault: true
Style/FrozenStringLiteralComment:
  Enabled: false

Layout/FirstHashElementIndentation:
  Enabled: true
Layout/FirstHashElementLineBreak:
  Enabled: true
Layout/TrailingWhitespace:
  Enabled: true
```

```ruby
# example.rb
def example
  { key1:
    { key2:
      { key3:
        { key4:
          { key5:
            :value
          }
        }
      }
    }
  }
end
```

--------

## Expected behavior

Running `rubocop -a` should produce an output with reasonable indentation. Something like

```ruby
# example.rb
def example
  {
    key1:
      {
        key2:
          {
            key3:
              {
                key4:
                  {
                    key5:
                      :value
                  }
              }
          }
      }
  }
end
```

A human would probably manually correct it as this , but I'm not sure if they're a setting to move the `{`  onto the previous line.

```ruby
# example.rb
def example
  {
    key1: {
      key2: {
        key3: {
          key4: {
            key5: :value
          }
        }
      }
    }
  }
end
```

In either scenario, the indentation is reasonable (2 spaces indented compared to parent line).

## Actual behavior

Running `rubocop -a` over indents the contents of the hash:

```ruby
# example.rb
def example
  {
    key1:
        {
          key2:
                {
                  key3:
                          {
                            key4:
                                      {
                                        key5:
                                                    :value
                                      }
                          }
                }
        }
  }
end
```

Notice the indentation seems compound and get worse the more it is nested. It appears
- the value for `:key1` is over nested by 2 spaces relative to the `key1:` line
- the value for `:key2` is over nested by 4 spaces relative to the `key2:` line
- the value for `:key3` is over nested by 6 spaces relative to the `key3:` line
- the value for `:key4` is over nested by 8 spaces relative to the `key4:` line
- the value for `:key5` is over nested by 10 spaces relative to the `key5:` line

<details><summary><code>rubocop --debug -a</code> output</summary>

```
bundle exec rubocop --debug -a
[...]

PR #11463, opened 2023-01-17
[WIP] [Fix: #11462] Fix over-indentation in `Layout/FirstElementHashIndentation`

sambostock:
This investigates fixing #11462. So far, it adds an autocorrection test. I'm still figuring out the details of the autocorrector logic, which I am very much unfamiliar with (cc. @dvandersluis ).

Starting with the example:

```ruby
def example
  { key1:
    { key2:
      { key3:
	{ key4:
	  { key5:
	    :value
	  }
	}
      }
    }
  }
end
```

What seems to be happening is that a line break is inserted between `{` and `keyN:`. `Layout/TrailingWhitespace` handles getting rid of the space remaining after `{`, but everything goes wrong in `Layout/FirstHashElementIndentation`.

```ruby
def example
  {
key1:
    {
key2:
      {
key3:
	{
key4:
	  {
key5:
	    :value
	  }
	}
      }
    }
  }
end
```

It finds `key1:` in column 0, computes a column delta of 4, which it applies to the key and value lines, even though the value lines don't have the same indentation.

```ruby
def example
  {
    key1:
        {
    key2:
          {
    key3:
    	{
    key4:
    	  {
    key5:
    	    :value
    	  }
    	}
          }
        }
  }
end
```

This repeats for `key2:`, which is not indented as much as it should be relative to its opening `{`, so the entire thing is indented by the column delta.

```ruby
def example
  {
    key1:
        {
          key2:
                {
          key3:
          	{
          key4:
          	  {
          key5:
          	    :value
          	  }
          	}
                }
        }
  }
end
```

This continues to repeat for `key3:`, `key4:`, and `key5:`, respectively:

```ruby
def example
  {
    key1:
        {
          key2:
                {
                  key3:
                  	{
                  key4:
                  	  {
                  key5:
                  	    :value
                  	  }
                  	}
                }
        }
  }
end
```

```ruby
def example
  {
    key1:
        {
          key2:
                {
                  key3:
                  	{
                          key4:
                          	  {
                          key5:
                          	    :value
                          	  }
                  	}
                }
        }
  }
end
```

```ruby
def example
  {
    key1:
        {
          key2:
                {
                  key3:
                  	{
                          key4:
                          	  {
                                    key5:
                                    	    :value
                          	  }
                  	}
                }
        }
  }
end
```

Every level is incorrectly corrected once for each of it's parents, and finally for itself, compounding the error.

-----------------

* [ ] Added tests.

### DIFF
--- a/spec/rubocop/cli/autocorrect_spec.rb
+++ b/spec/rubocop/cli/autocorrect_spec.rb
@@ -2763,4 +2763,50 @@ RSpec.describe 'RuboCop::CLI --autocorrect', :isolated_environment do # rubocop:
       end
     RUBY
   end
+
+  it 'indents hashes correctly when adding a line break' do
+    source_file = Pathname('example.rb')
+    create_file(source_file, <<~RUBY)
+      def example
+        { key1:
+          { key2:
+            { key3:
+              { key4:
+                { key5:
+                  :value
+                }
+              }
+            }
+          }
+        }
+      end
+    RUBY
+
+    status = cli.run(%W[--autocorrect-all --only #{%w[
+      Layout/FirstHashElementIndentation
+      Layout/FirstHashElementLineBreak
+      Layout/TrailingWhitespace
+    ].join(',')}])
+    expect(status).to eq(0)
+    expect(source_file.read).to eq(<<~RUBY)
+      def example
+        {
+          key1:
+          {
+            key2:
+            {
+              key3:
+              {
+                key4:
+                {
+                  key5:
+                  :value
+                }
+              }
+            }
+          }
+        }
+      end
+    RUBY
+  end
 end


spec/rubocop/cli/autocorrect_spec.rb
```diff
@@ -2763,4 +2763,60 @@ module Foo#{trailing_whitespace}
...
+
+      Layout/FirstHashElementIndentation:
+        Enabled: true
+      Layout/FirstHashElementLineBreak:
+        Enabled: true
+      Layout/TrailingWhitespace:
+        Enabled: true
+    YAML
+
+    status = cli.run(%w[--autocorrect-all])
```
dvandersluis, line 2799: You should probably just specify (with `--only`, see line 2758 for an example) which cops you want to run for this test, and then you wouldn't need to explicitly enable/disable cops in the yaml file.

PR #11469, opened 2023-01-19
Add `Gemspec/DevelopmentDependencies` cop

sambostock:
This cop checks that development dependencies are specified in `Gemfile` rather than `gemspec`. It also supports `EnforcedStyle: gems.rb`, which behaves the same way, but mentions `gems.rb` in the error message, or `EnforcedStyle: gemspec`, which inverts the check and enforces that all dependencies be specified in the gemspec file.

An `AllowedGems` config is also included, for projects that want to include some gems in their gemspec. For example, Rubocop itself specifies

```ruby
s.add_development_dependency 'bundler'
```

but specifies all other development dependencies in `Gemfile`.

-----------------

* ~Commit message starts with `[Fix #issue-number]` (if the related issue exists)~.
* [x] Added tests.

### DIFF
--- a/changelog/new_add_gemspec_development_dependencies_cop.md
+++ b/changelog/new_add_gemspec_development_dependencies_cop.md
@@ -1,0 +1,1 @@
+* [#11469](https://github.com/rubocop/rubocop/pull/11469): Add `Gemspec/DevelopmentDependencies` cop. ([@sambostock][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -262,6 +262,22 @@ Gemspec/DeprecatedAttributeAssignment:
   Include:
     - '**/*.gemspec'
 
+Gemspec/DevelopmentDependencies:
+  Description: Checks that development dependencies are specified in Gemfile rather than gemspec.
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  EnforcedStyle: Gemfile
+  SupportedStyles:
+    - Gemfile
+    - gems.rb
+    - gemspec
+  AllowedGems:
+    - bundler
+  Include:
+    - '**/*.gemspec'
+    - '**/Gemfile'
+    - '**/gems.rb'
+
 Gemspec/DuplicatedAssignment:
   Description: 'An attribute assignment method calls should be listed only once in a gemspec.'
   Enabled: true
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -166,6 +166,7 @@ require_relative 'rubocop/cop/bundler/ordered_gems'
 
 require_relative 'rubocop/cop/gemspec/dependency_version'
 require_relative 'rubocop/cop/gemspec/deprecated_attribute_assignment'
+require_relative 'rubocop/cop/gemspec/development_dependencies'
 require_relative 'rubocop/cop/gemspec/duplicated_assignment'
 require_relative 'rubocop/cop/gemspec/ordered_dependencies'
 require_relative 'rubocop/cop/gemspec/require_mfa'
--- a/lib/rubocop/cop/gemspec/development_dependencies.rb
+++ b/lib/rubocop/cop/gemspec/development_dependencies.rb
@@ -1,0 +1,107 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Gemspec
+      # Enforce that development dependencies for a gem are specified in
+      # `Gemfile`, rather than in the `gemspec` using
+      # `add_development_dependency`. Alternatively, using `EnforcedStyle:
+      # gemspec`, enforce that all dependencies are specified in `gemspec`,
+      # rather than in `Gemfile`.
+      #
+      # @example EnforcedStyle: Gemfile (default)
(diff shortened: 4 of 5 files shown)


bbatsov:
Looks like a good addition to me! 

> s.add_development_dependency 'bundler'

Like that was an oversight on our part. I'm also wondering if `Gemfile` or `gems.rb` should be separate styles - probably this should be some global configuration for `Gemfile` or `gems.rb` (or auto-detection), but that's something we can address down the road.

DannyBen:
Excellent cop, thank you.

mvz:
What's the reasoning behind this new cop?

mvz:
Thanks for the quick response, @sambostock.

I'm totally on board with the consistency, but would much prefer the default to be for the development dependencies to be specified in the gemspec, since that way rubygems.org will know about these dependencies. This helps developers of supporting gems (like myself) to know which gems make use of them. I wouldn't want this resource to be lost.

sambostock:
> I believe I saw somewhere one good reason for defining dev dependencies in gemspec, and it was related to whether the users of the gem might have two environment, like in Rails for example, in which the development dependency is needed foe the user's development pipeline.

Rails' development dependencies shouldn't be needed by consumers. Those dependencies should only be needed if actually working on Rails itself.

@mvz brings up an interesting use case above (providing information for gem maintainers about which gems depend on theirs as development dependencies), however, I am uncertain about the reliability of this, given gems could be listed in `Gemfile` instead.

mvz:
@sambostock it indeed does not provide a full picture, but I think something is better than nothing, and if RuboCop nudges developers toward supplying this information in the gemspec, this would only improve.

DannyBen:
I found one of the discussions I looked at before, when I was trying to understand when and if I should use `add_development_dependencies`.

It is an old, long but still very interesting discussion: https://github.com/rubygems/rubygems/discussions/5065

I particularly relate to this amusing statement:

> It's like a conspiracy - "everyone" acts like it's a well-thought-out super-useful feature, but no one can explain why.

*(originally issue https://github.com/rubygems/rubygems/issues/1104)*

hlascelles:
I'm happy with development dependencies being in the Gemfile (more control over versions?) though I do see @mvz's point about rubygems dependency knowledge.

Also, for what it's worth, here is a reason to have test gems in the `gemspec`: If you are using the [appraisal](https://github.com/thoughtbot/appraisal) gem you will run into problems.

The gem works by creating a folder of Gemfiles which the tests are run with. They are generated from the `gemspec` and an `Appraisal` file.

If you have (say) `rspec` in your `gemspec` as a `development_dependency` then this works just fine. But if you move it to your `Gemfile` (as per this cop implies) then it is *not* included when the tests are run (as that main `Gemfile` is not used), so the tests crash with "can't find executable rspec for gem rspec-core".

Here is an example of the error: https://github.com/hlascelles/que-scheduler/actions/runs/4000393109/jobs/6865465909

The fix is to put the test gems in the `Appraisal` file, which feels a bit weird as that is where you are supposed to put "gem version differences to test". Or move them back to the `gemspec` with a cop disable (which I have done).

This is a rare(r) corner case, and for simple gems maybe the cop is good. But when using [appraisal](https://github.com/thoughtbot/appraisal) it's one to remember.

mvz:
I just found out that no offense is reported if a version is specified for the development dependency :thinking:.

ryanwjackson:
@sambostock can you provide some guidance on how this cop expects development dependencies to be specified in the Gemfile? Is it in the `:development` group?

DannyBen:
@ryanwjackson - in a folder with `gemspec` and `Gemfile`, it means that this folder is used for building a gem.

So - since the `Gemfile` will only be used by developers of the gem (i.e. those who git cloned your repository), it means everything in the `Gemfile` is for development. You can just place them there as usual. No need for any grouping, unless you really want to (e.g. `test` group).

AlexWayfer:
I still don't understand why I should change all my 20+ gems. Because someone wanted it? Without any practical reason? I see no cons with `development_dependency`. There is can be a such cop, but I doubt it should be enabled by default.

luke-hill:
Question (Sorry to revive a dead horse).

In many gems (5+), that I look after. We have the concept of "low spec" or "medium spec" gemfiles. The way these work currently is...

They install the "overridden" gems first, so we maintain a small gemfile of 2-4 gems per "type of gemfile". Then it installs the rest which can just be default. For a legit example see here: https://github.com/cucumber/cucumber-rails/blob/main/gemfiles/rails_5_1.gemfile

This "lower" spec gemfile we manually specify 7 gems. But there a bunch of others we also must install from the dev dependencies - I think this is 14-7=7 but this was a quick 5s glance.

My understanding is that if we no longer use dev dependencies, we should place them in the gemfile. Which is fine. However then using a pipeline like GHA we specify 1 specific gemfile. In this case the rails 5.1 gemfile above. This will (To my knowledge), only install the 7 dev dependency gems.

TL;DR - I believe this cop increases the workload for people maintaining gems that have lots of "lower" spec gemfiles to maintain.

PR #11470, opened 2023-01-19
[Fix #11467] Fix a false negative for `Style/MethodCallWithoutArgsParentheses`

### LINKED ISSUE #11467: False negative for `Style/MethodCallWithoutArgsParentheses`
`Style/MethodCallWithoutArgsParentheses` doesn't report an issue when assigning to a variable with the same name as the method, e.g.:
```ruby
foo = x.foo()
```
It correctly reports an issue when the method and variable names are different:
```ruby
bar = x.foo()
```

This seems like an unintended side-effect of #2056.

## Expected behavior

Both of the above should report an issue.

## Actual behavior

Only the latter case reports an issue.

```console
~/src % cat >test.rb
foo = x.foo()
bar = x.foo()
~/src % rubocop --only Style/MethodCallWithoutArgsParentheses test.rb 
Inspecting 1 file
C

Offenses:

test.rb:2:12: C: Style/MethodCallWithoutArgsParentheses: Do not use parentheses for method calls with no arguments.
bar = x.foo()
           ^^

```

## RuboCop version

```console
~/src % rubocop -V       
0.88.0 (using Parser 3.1.0.0, rubocop-ast 0.2.0, running on ruby 2.7.7 arm64-darwin21)
```

koic:
Fixes #11467.

This PR fixes a false negative for `Style/MethodCallWithoutArgsParentheses` when calling method on a receiver and assigning to a variable with the same name.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_negative_for_style_method_call_without_args_parentheses.md
+++ b/changelog/fix_a_false_negative_for_style_method_call_without_args_parentheses.md
@@ -1,0 +1,1 @@
+* [#11467](https://github.com/rubocop/rubocop/issues/11467): Fix a false negative for `Style/MethodCallWithoutArgsParentheses` when calling method on a receiver and assigning to a variable with the same name. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Looks good. Thanks!

harto:
Thanks for the quick fix! 👏

PR #11472, opened 2023-01-19
Use `Corrector#swap` where possible

fatkodima:
`Corrector#swap` was introduced in the PR for `YodaExpression` cop - https://github.com/rubocop/rubocop/pull/11333/files#diff-1341c76fbde042e0ad98b3df91749671b56de56e87ff63496b5c422db1d9e2b9R81-R91.

I tweak it to not raise clobbering error for two adjacent ranges and make it use where possible in the codebase.


bbatsov:
Thanks!

PR #11473, opened 2023-01-19
Add intentionally empty file to implicitly test cops on it

fatkodima:
Here and there are error reports that some cop fails on an empty file. The latest of them - https://github.com/rubocop/rubocop-minitest/issues/225. Some of the cops already have test cases for empty files, but it is tedious to write them and very often just forgotten.

So I added an empty file into the project so that cops (at least which are not disabled) are always tested (implicitly) on an empty file.
I think, we should add a similar file to the `test/` directory of the `rubocop-minitest` gem and somewhere to the `rubocop-rails` too.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -80,6 +80,11 @@ Lint/UselessAccessModifier:
     - 'def_matcher'
     - 'def_node_matcher'
 
+Lint/EmptyFile:
+  Exclude:
+    # This file is intentionally empty to catch rubocop cops failing on empty files.
+    - spec/rubocop/intentionally_empty_file.rb
+
 Metrics/BlockLength:
   Exclude:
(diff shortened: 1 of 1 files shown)


bbatsov:
I get the idea, but I don't like having an empty lib file simply for test purposes (as it will look weird if someone's browsing the code). I guess we can just have the `test` task create and delete one or something along those lines.

fatkodima:
That is actually the response I was expected 😄 

How about moving it into the `spec/` directory? Or tweaking `spec/spec_helper.rb` to add such a test case to all cop test suites? (not sure how to do it - not a rspec user).

bbatsov:
Something like this would be fine by me.

fatkodima:
Moved the file to `spec/`.

Issue #11076, opened 2022-10-14
Lint/LiteralInInterpolation does incorrect replacement for hashes


tristil:
Running the autocorrect for Lint/LiteralInInterpolation does the right thing for strings and numbers, but it does an incorrect replacement for hashes. This is because unlike strings and numbers the string representation of a hash is different from how it is written in code. 

--------

## Expected behavior

```ruby
# original code
"#{ { foo: 'bar'} }" #=> "{:foo=>\"bar\"}"

# autocorrected code
"{:foo=>\"bar\"}"
```

Alternatively, and maybe more preferably, this cop could just not complain about hash literals since depending on the string representation of a hash is an unusual use case.

## Actual behavior

```ruby
# original code
"#{ { foo: 'bar'} }" #=> "{:foo=>\"bar\"}"

# autocorrected code
"{ foo: 'bar' }"
```

## Steps to reproduce the problem

1. Create a Ruby file `foo.rb` with contents `"#{ { foo: 'bar'} }" `
2. run `rubocop --autocorrect --only Lint/LiteralInInterpolation foo.rb`

## RuboCop version

```
$ [bundle exec] rubocop -V
1.36.0 (using Parser 3.1.2.1, rubocop-ast 1.21.0, running on ruby 2.7.5) +server [x86_64-darwin21]
  - rubocop-rails 2.15.2
```

KessaPassa:
I think @tristil said right.
Actual behavior is not better.

`hash type` is not defined. so, `hash type` is using `else`.
I'll try to fix it to create a `when :hash`.
https://github.com/rubocop/rubocop/blob/4b4c2a409b14908c4344995720e04a0c65dc55e4/lib/rubocop/cop/lint/literal_in_interpolation.rb#L62-L79

piroor:
I agree to @KessaPassa, because linter's autocorrect should not break compatibility of the target code. On this case it is expected that rubocop corrects the code exactly same as the result Ruby evaluates.

PR #11475, opened 2023-01-20
[#11076] make hash type autocorrect with test

KessaPassa:
Fixes https://github.com/rubocop/rubocop/issues/11076.
This PR add autocorrect to `Lint/LiteralInInterpolation` for hash type.

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/literal_in_interpolation.rb
```diff
@@ -100,6 +102,40 @@ def autocorrected_value_for_array(node)
...
+          case node.type
+          when :int
+            node.children.last.to_i.to_s
+          when :float
+            node.children.last.to_f.to_s
+          when :str
+            '\"' + node.value.to_s.gsub('"') { '\\\\\"' } + '\"'
+          when :sym
+            # TODO: We need to detect symbol unacceptable names more correctly
+            if / |"|'/ =~ node.value.to_s
```
KessaPassa, line 125: A gem [unparser](https://github.com/mbj/unparser) provides more simple (and better) way to generate code strings for autocorrect result. 
Should we add it to the list of dependencies?
```suggestion
            Unparser.unparse(node).gsub('"') { '\"' }
```

lib/rubocop/cop/lint/literal_in_interpolation.rb
```diff
@@ -91,7 +93,16 @@ def autocorrected_value_for_symbol(node)
           end_pos =
             node.loc.end ? node.loc.end.begin_pos : node.loc.expression.end_pos
 
-          range_between(node.loc.begin.end_pos, end_pos).source
+          range_between(node.loc.begin.end_pos, end_pos).source.gsub('"', '\"')
+        end
+
+        def autocorrected_value_in_hash_for_symbol(node)
+          # TODO: We need to detect symbol unacceptable names more correctly
+          if / |"|'/.match?(node.value.to_s)
```
KessaPassa, line 101: A gem [unparser](https://github.com/mbj/unparser) provides more simple (and better) way to generate code strings for autocorrect result. 
Should we add it to the list of dependencies?
```suggestion
            Unparser.unparse(node).gsub('"') { '\"' }
```

bbatsov: Probably that'd be an overkill for a single usage. I'm always wary of adding additional runtime deps.

spec/rubocop/cop/lint/literal_in_interpolation_spec.rb
```diff
@@ -114,6 +121,72 @@
   it_behaves_like('literal interpolation', '%I[s1 s2]', '[\"s1\", \"s2\"]')
   it_behaves_like('literal interpolation', '%i[s1     s2]', '[\"s1\", \"s2\"]')
   it_behaves_like('literal interpolation', '%i[ s1   s2 ]', '[\"s1\", \"s2\"]')
+  it_behaves_like('literal interpolation', '{"a" => "b"}', '{\"a\"=>\"b\"}')
```
bbatsov, line 124: I think it might be nice to have the hash-related tests in their own section. The tests for this cop are already quite messy and we should try to improve this.

KessaPassa: @bbatsov thank you!
I created sctions for each type.
How about this change?
https://github.com/rubocop/rubocop/pull/11475/commits/c9fdcdc8c4a864b8476fe54560721b4fe997fe98

lib/rubocop/cop/lint/literal_in_interpolation.rb
KessaPassa: I aimed to separate codes to unparse hash from the rubocop to make its project scope more clear.
But after more research I realized that the `unparser gem` sometimes generate results not exact same to Ruby's one.
Thus I agree to cancel this suggestion.

mbj: @KessaPassa Unparser only guarantess: `parse(code) == parse(unparse(parse(code)))`, so it guarantees you generate concrete syntax that if parsed produces the original AST, not the same concrete syntax. If you knew the concrete syntax already there is no need to unparse in the first place ;)

PR #11484, opened 2023-01-23
Fix `Layout/ClosingParenthesisIndentation` for keyword splat arguments

### LINKED ISSUE #11483: Wrong behavior when Layout/ClosingParenthesisIndentation sees double splat
Consider the following code in test.rb:
```rb
foo(bar,
  **baz
)
```

## Expected behavior

`./exe/rubocop --only=Layout/ClosingParenthesisIndentation test.rb` should not see any offenses. [The cop documentation](https://github.com/rubocop/rubocop/blob/6139e4019cfa349fbddaf7244eeb21e780a1a6d4/lib/rubocop/cop/layout/closing_parenthesis_indentation.rb#L63-L68) says “when other params are not lined up on multiple lines, outdent right paren by `IndentationWidth`”. Also this code does not cause offense:

```rb
foo(bar,
  baz
)
```

## Actual behavior

```
Offenses:

test.rb:3:1: C: [Correctable] Layout/ClosingParenthesisIndentation: Align ) with (.
)
^
```

See above.

## RuboCop version

```
❯ ./exe/rubocop -V
1.44.0 (using Parser 3.2.0.0, rubocop-ast 1.24.1, running on ruby 3.1.2) [x86_64-darwin21]
  - rubocop-performance 1.15.2
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.18.1
```

```
❯ git rev-parse @                                                                      
51ded37abcd63590cca74a86e2b26812e5ac8ae3
```

fatkodima:
Fixes #11483.

We need to check all the hash children nodes (not only `pair`s).

```
$ ruby-parse -e 'foo(x: 1, **options, y: 2)'
```
```
(send nil :foo
  (kwargs
    (pair
      (sym :x)
      (int 1))
    (kwsplat
      (send nil :options))
    (pair
      (sym :y)
      (int 2))))
```

### DIFF
--- a/changelog/fix_closing_parenthesis_indentation_for_keyword_splat.md
+++ b/changelog/fix_closing_parenthesis_indentation_for_keyword_splat.md
@@ -1,0 +1,1 @@
+* [#11483](https://github.com/rubocop/rubocop/issues/11483): Fix `Layout/ClosingParenthesisIndentation` for keyword splat arguments. ([@fatkodima][])
--- a/lib/rubocop/cop/layout/closing_parenthesis_indentation.rb
+++ b/lib/rubocop/cop/layout/closing_parenthesis_indentation.rb
@@ -157,7 +157,7 @@ module RuboCop
         def all_elements_aligned?(elements)
           elements.flat_map do |e|
             if e.hash_type?
-              e.each_pair.map { |pair| pair.loc.column }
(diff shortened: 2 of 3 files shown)


bquorning (approved):
🎉

koic:
Thanks!

PR #11486, opened 2023-01-23
Fix an incorrect autocorrect for `Naming/BlockForwarding`

### LINKED ISSUE #11400: Autocorrect of ambiguous splat and anonymous block creates a syntax error
When formatting code that contains _both_ a named block forwarding and an ambiguous keyword splat, `rubocop -a` will generate syntactically invalid code. 

--------

## Expected behavior

Given the following code:

```rb
def foo(options, &block)
  bar **options, &block
end
```

I expect Rubocop to generate the following output:

```rb
def foo(options, &)
  bar(**options, &)
end
```

## Actual behavior

Rubocop produces this output:

```rb
def foo(options, &)
  bar((**options, &))
end
```

Put the above code in a file and run `rubocop -a testfile.rb`

## RuboCop version

```
$ bundle exec rubocop -V
1.42.0 (using Parser 3.2.0.0, rubocop-ast 1.24.1, running on ruby 3.1.2) [arm64-darwin21]
```

fatkodima:
Fixes #11400.


koic:
This looks good!

PR #11464, opened 2023-01-17
Fix a false negative for `Lint/FormatParameterMismatch` when include interpolated string

ydah:
This PR is fix a false negative for `Lint/FormatParameterMismatch` when include interpolated string.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [-] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #11487, opened 2023-01-24
Fix a false positive for `Lint/FormatParameterMismatch` when format string is only interpolated string

ydah:
This PR is Follow up: https://github.com/rubocop/rubocop/pull/11464#issuecomment-1400880819

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/format_parameter_mismatch.rb
+++ b/lib/rubocop/cop/lint/format_parameter_mismatch.rb
@@ -80,6 +80,7 @@ module RuboCop
           num_of_format_args, num_of_expected_fields = count_matches(node)
 
           return false if num_of_format_args == :unknown
+          return false if num_of_expected_fields.zero? && node.child_nodes.first.dstr_type?
 
           matched_arguments_count?(num_of_expected_fields, num_of_format_args)
         end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/format_parameter_mismatch.rb
```diff
@@ -80,6 +70,7 @@ def offending_node?(node)
           num_of_format_args, num_of_expected_fields = count_matches(node)
 
           return false if num_of_format_args == :unknown
+          return false if num_of_expected_fields&.zero? && node.child_nodes.first.dstr_type?
```
koic, line 73: The test seems to pass on `return false if num_of_expected_fields.zero?` . There seems to be too much or too little either implementation or testing.

ydah: There seemed to be a lack of testing. Thank you very much.

lib/rubocop/cop/lint/format_parameter_mismatch.rb
```diff
@@ -80,6 +80,7 @@ def offending_node?(node)
...
```
koic, line 83: Is the safe navigation operator required?

```suggestion
          return false if num_of_expected_fields.zero? && node.child_nodes.first.dstr_type?
```

ydah: Ah, initially it was necessary because the order was reversed with L82, so :unknown came in, but now it is not necessary. Thank you very much.

hmcguire-shopify:
Thanks for the quick fix!

PR #11491, opened 2023-01-24
Fix a crash on Lint/UselessAssignment

gsamokovarov:
The crash is:

```
TypeError:
  no implicit conversion from nil to integer
# ./lib/rubocop/cop/variable_force/scope.rb:53:in `body_node'
```

it happened for nodes like:

```
s(:numblock,
  s(:send,
    s(:begin,
      s(:irange,
        s(:int, 0),
        s(:int, 10))), :each), 1,
  s(:op_asgn,
    s(:lvasgn, :checksum_total), :+,
    s(:send,
      s(:send,
        s(:lvar, :checksum_multiplier_by_position), :[],
        s(:lvar, :_1)), :*,
      s(:send,
        s(:send,
          s(:lvar, :contact_tax_number), :[],
          s(:lvar, :_1)), :to_i))))
```

### DIFF
--- a/spec/rubocop/cop/lint/useless_assignment_spec.rb
+++ b/spec/rubocop/cop/lint/useless_assignment_spec.rb
@@ -266,6 +266,13 @@ RSpec.describe RuboCop::Cop::Lint::UselessAssignment, :config do
         end
       RUBY
     end
+
+    it 'registers offenses for self assignment in numblock', :ruby27 do
+      expect_offense(<<~RUBY)
+        do_something { foo += _1 }
+                       ^^^ Useless assignment to variable - `foo`. Use `+` instead of `+=`.
+      RUBY
+    end
   end
 
   context 'when a variable is assigned in loop body and unreferenced' do
--- a/changelog/fix_a_crash_on_lint_useless_assignment.md
+++ b/changelog/fix_a_crash_on_lint_useless_assignment.md
@@ -1,0 +1,1 @@
+* [#11491](https://github.com/rubocop/rubocop/pull/11491): Fix a crash on Lint/UselessAssignment. ([@gsamokovarov][])
--- a/lib/rubocop/cop/variable_force/scope.rb
+++ b/lib/rubocop/cop/variable_force/scope.rb
@@ -45,9 +45,9 @@ module RuboCop
             node
           else
             child_index = case node.type
-                          when :module, :sclass     then 1
-                          when :def, :class, :block then 2
-                          when :defs                then 3
+                          when :module, :sclass then 1
+                          when :def, :class, :block, :numblock then 2
+                          when :defs then 3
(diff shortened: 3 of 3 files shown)


spec/rubocop/cop/lint/useless_assignment_spec.rb
```diff
@@ -266,6 +266,16 @@ def some_method
...
     end
+
+    it 'registers offenses for the assignment in numblock' do
+      expect_offense(<<~'RUBY')
+        1.times do
+          puts "Iteration #{_1 + 1}"
+          foo = 2
+          ^^^ Useless assignment to variable - `foo`.
+        end
+      RUBY
```
koic, line 277: Is it possible to reproduce the error with this test case? I reverted `:numblock` from the `case` / `when` branch in the scope.rb and ran the test, but the error was not reproduced.

gsamokovarov: It's still hard for me to get the exact repro I get on my codebase. Let me try a bit more.

gsamokovarov:
```ruby
def validate_format(foo)
  return false unless foo.in_format?

  bar = [1]
  baz = 1
  bap = 2

  baq = 0
  (0..10).each { baq += bar[_1] * foo[_1].to_i }

  baq % baz == bap
end
```

This is the code that crashes because of the `numblock` locally, but I still cannot reproduce it in isolation. @koic I'm posting the code if you have an idea here.

koic, line 277: I was able to reproduce it below test case. Can you review and update it?

```suggestion
    it 'registers offenses for the self assignment in numblock', :ruby27 do
      expect_offense(<<~'RUBY')
        do_something { foo += _1 }
                       ^^^ Useless assignment to variable - `foo`. Use `+` instead of `+=`.
      RUBY
```

For reference, the reproduced error is below:

```console
  1) RuboCop::Cop::Lint::UselessAssignment when a block local variable is assigned and unreferenced registers offenses for the self assignment in numblock
     Failure/Error: node.children[child_index]

     TypeError:
       no implicit conversion from nil to integer
     # ./lib/rubocop/cop/variable_force/scope.rb:53:in `body_node'
     # ./lib/rubocop/cop/lint/useless_assignment.rb:99:in `return_value_node_of_scope'
```

gsamokovarov: Oh, I have forgotten to mark the test as :ruby27, that was it! 🤦‍♂️

gsamokovarov: Thank you, @koic! I have amended my commit with the fix.

PR #11502, opened 2023-01-25
Add a description of the `AllowedMethods` module documentation

ydah:
This PR is fixes: https://github.com/rubocop/rubocop/pull/10829#issuecomment-1403082579

-----------------

* [-] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/allowed_methods.rb
+++ b/lib/rubocop/cop/mixin/allowed_methods.rb
@@ -3,7 +3,9 @@
 module RuboCop
   module Cop
     # This module encapsulates the ability to allow certain methods when
-    # parsing.
+    # parsing. Even if the code is in offense, if it contains methods
+    # that are allowed. This module is equivalent to the IgnoredMethods module,
+    # which will be deprecated in RuboCop 2.0.
     module AllowedMethods
       private
 


lib/rubocop/cop/mixin/allowed_methods.rb
```diff
@@ -3,7 +3,9 @@
 module RuboCop
   module Cop
     # This module encapsulates the ability to allow certain methods when
-    # parsing.
+    # parsing. Function to allow even code that would be an offense in this module.
```
bbatsov, line 6: Function to allow even code that would be an offense in this module. - this sentence confuses me. 

Perhaps we'd be better off without it. You can also see the comment I left for the ticket that spawned this conversation.

ydah: Thank you for your comments. I have improved the expression. How is it now?

lib/rubocop/cop/mixin/allowed_methods.rb
```diff
@@ -3,7 +3,8 @@
 module RuboCop
   module Cop
     # This module encapsulates the ability to allow certain methods when
-    # parsing.
+    # parsing. Even if the code is in offense, if it contains methods
+    # that are allowed, it will deal with not be registered as an offense.
```
bbatsov, line 7: "deal with" seems redundant here. 

You should probably also mentioned that this is functionality equivalent to the `IgnoredMethods` class that is deprecated and will be removed in RuboCop 2.0.

ydah: Thank you so much. I updated this PR.

bbatsov (approved):

PR #11507, opened 2023-01-26
Add `Style/HashNewKeywordArguments` cop

r7kamura:
I have heard that future versions of Ruby are likely to deprecate the ability to pass default values with keyword arguments to `Hash.new`.

- https://bugs.ruby-lang.org/issues/19236

So I thought it would be a good idea to prepare a Cop that detects such usage and autocorrects it while it is still available. Since the signature is originally `Hash.new(default_value = nil)`, there should be no problem deprecating the keyword arguments.

- https://ruby-doc.org/3.1.2/Hash.html#method-c-new

-----------------

* [x] Added tests.


r7kamura:
I am a little concerned about whether `Style` is the correct department, and also whether this is the best way to detect whether the target node is keyword arguments or not.

lib/rubocop/cop/style/hash_new_keyword_arguments.rb
```diff
@@ -0,0 +1,49 @@
...
+          add_offense(keyword_arguments) do |corrector|
+            corrector.wrap(keyword_arguments, '{ ', ' }')
+          end
+        end
+
+        private
+
+        def keyword_arguments?(argument)
+          argument.hash_type? && !argument.source.lstrip.start_with?('{')
+        end
```
r7kamura, line 45: Is there any other good way to inspect keyword arguments?

ydah: It seems that `kwargs` is rounded to `hash` in rubocop-ast.
https://github.com/rubocop/rubocop-ast/blob/57bc6792ed83dfb2890ed1ce1eebc2580fd83aa2/lib/rubocop/ast/builder.rb#L64

I don't know if this is better, but would it be better to define a `KwargsNode` to replace it?

r7kamura: Since the `emit_kwargs` option is not enabled in RuboCop's parser, I don't think the `kwargs` type is actually passed there, but changing this option might have too much impact.

- https://github.com/whitequark/parser/blob/8de8b7fa7af471a2159860d6a0a5b615eac9c83c/lib/parser/builders/default.rb#L131-L177

koic:
Interesting. However, it is too early to judge the direction of this cop force yet. So, it should be decided after https://bugs.ruby-lang.org/issues/19236 is settled. Let's come back to this PR once https://bugs.ruby-lang.org/issues/19236 is closed.

bbatsov:
Let's mark this PR as WIP in the mean time.

Zopolis4:
https://bugs.ruby-lang.org/issues/19236 has been closed by ruby/ruby#7828, which has been shipped in Ruby 3.3.

bbatsov (approved):

r7kamura:
Nice catch. I was completely unaware of its existence as well 😆

The purpose is the same, but in terms of the thoroughness of the explanation and the consideration for edge cases like `capacity`, Lint/HashNewWithKeywordArgumentsAsDefault is definitely superior.

I'll go ahead and close this pull request — thanks again!

Issue #11485, opened 2023-01-23
`Lint/UselessAssignment` false positive when variables are used in blocks with numbered parameters


Fustrate:
## Expected behavior

Variables that are set and then used inside blocks with numbered parameters should not be marked as unused.

It looks like it was caused by b48ebef423f74c3083bf2cd7ab79e8546f28a95e - I tried the commit before and it passed. This also causes `Lint/UnusedMethodArgument` and `Lint/UnusedBlockArgument` warnings.

## Actual behavior

```
For /Users/Fustrate/Development/DummyProject: configuration from /Users/Fustrate/Development/DummyProject/.rubocop.yml
configuration from /Users/Fustrate/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/rubocop-performance-1.15.2/config/default.yml
configuration from /Users/Fustrate/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/rubocop-performance-1.15.2/config/default.yml
Default configuration from /Users/Fustrate/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/rubocop-1.44.0/config/default.yml
configuration from /Users/Fustrate/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/rubocop-rails-2.17.4/config/default.yml
configuration from /Users/Fustrate/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/rubocop-rails-2.17.4/config/default.yml
Use parallel by default.
Skipping parallel inspection: only a single file needs inspection
Inspecting 1 file
Scanning /Users/Fustrate/Development/DummyProject/test.rb
W

Offenses:

test.rb:3:1: W: Lint/UselessAssignment: Useless assignment to variable - target.
target = 6
^^^^^^

1 file inspected, 1 offense detected
Finished in 0.47589999999945576 seconds
```

## Steps to reproduce the problem

```
# rubocop:disable all
# rubocop:enable Lint/UselessAssignment
target = 6

[1, 2, 3].each { puts _1 if _1 * 2 == target }

```

Changing it to a normal block with named arguments passes:

```
target = 6

[1, 2, 3].each do |num|
  puts  num if num * 2 == target
end
```

## RuboCop version

```
$ rubocop -V
1.44.0 (using Parser 3.2.0.0, rubocop-ast 1.24.1, running on ruby 3.2.0) [x86_64-darwin22]
[...]

PR #11510, opened 2023-01-26
Fix `Lint/UselessAssignment` false positive

sambostock:
A variable assigned in a block is not unused, if it was captured from the surrounding scope, where it is later used.

```ruby
var = nil
do_something { |arg| var = arg }
use(var)
```

We already cover this case for normal blocks, but missed the edge case where the block uses numbered variables (i.e. `numblock` vs `block`).

```ruby
var = nil
do_something { var = _1 }
             # ^^^ False positive for useless assignment
use(var)
```

This fixes the false positive.

-----------------

* [x] Added tests.


lib/rubocop/cop/variable_force/variable_table.rb
```diff
@@ -113,14 +113,14 @@ def variable_exist?(name)
         def accessible_variables
           scope_stack.reverse_each.with_object([]) do |scope, variables|
             variables.concat(scope.variables.values)
-            break variables unless scope.node.block_type?
+            break variables unless scope.node.block_type? || scope.node.numblock_type?
```
sambostock, line 116: As mentioned in the commit message, I tentatively added this to match the other `block_type?` call sites in this file:

- #11485 added it in `find_variable` above, to resolve a false positive (#11485)
- I am adding it in `mark_variable_as_captured_by_block_if_so` below, to resolve a false positive

I don't know what the false positive is here, but I'm assuming there is one. If someone with more context knows what it might be, I'm can add a corresponding test and update the commit message and changelog.

spec/rubocop/cop/lint/useless_assignment_spec.rb
```diff
@@ -1326,6 +1326,26 @@ def some_method(name: value, **)
...
+
+    it 'registers an offense when the variable is assigned and but unused' do
+      expect_offense(<<~RUBY)
+        var = nil
+        ^^^ Useless assignment to variable - `var`.
+
+        do_something { var = _1 }
+                       ^^^ Useless assignment to variable - `var`.
+      RUBY
+    end
```
sambostock, line 1348: Hmm... I just added this test as the inverse of the one above, but it looks like it fails. I think it's because
- `var = nil` is considered fine, because a block later uses it
- `var = _1` is considered fine, because it's assigning to a shadowed variable

I don't think there's anything checking that the variable is used in neither the block, nor the surrounding scope.

I'm not sure how to go about implementing that, so I may just remove the test and accept the unlikely false negative. 🤷

bbatsov:
@sambostock The CI is not passing currently.

changelog/fix_tentatively_check_against_numblock_type_too.md
```diff
@@ -0,0 +1 @@
+* [#11510](https://github.com/rubocop/rubocop/pull/11510): Tentatively check against `numblock_type?` too. ([@sambostock][])
```
koic, line 1: Can you remove this changelog entry?

spec/rubocop/cop/lint/useless_assignment_spec.rb
koic: Yep. It's an existing issue that also happens with normal block (not numbered block). So, can you remove this test?

koic:
I left a comment about removing 2 places. Please squash your commits into one after the remove.

sambostock:
@koic Sorry for the delay; updated. 👍

changelog/fix_fix_lint_useless_assignment_false_positive.md
```diff
@@ -0,0 +1 @@
+* [#11510](https://github.com/rubocop/rubocop/pull/11510): Fix `Lint/UselessAssignment` false positive. ([@sambostock][])
```
koic, line 1: Can you supplement this changelog entry?

```suggestion
* [#11510](https://github.com/rubocop/rubocop/pull/11510): Fix `Lint/UselessAssignment` false positive when using numbered block parameters. ([@sambostock][])
```

koic:
@sambostock Finally noticed about the changelog. Can you take a look?

koic:
Thanks!

PR #11512, opened 2023-01-27
[Fix #11511] Fix a false negative for `Style/YodaCondition`

### LINKED ISSUE #11511: Should `Style/YodaCondition` treat constants like literals?
Currently, `Style/YodaCondition` flags something like this:

```ruby
if 'red' == color
  foo
elsif 'green' == color
  bar
end
```

but not something like this:
```ruby
if Colors::RED == color
  foo
elsif Colors::GREEN == color
  bar
end
```

Should this cop treat constants like literals? (Or is there some inherent danger in doing so?)

koic:
Fixes #11511.

This PR fixes a false negative for `Style/YodaCondition` when using constant.
Yoda conditions targets constant but only handles literal. This patch makes constant aware.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_negative_for_style_yoda_condition.md
+++ b/changelog/fix_a_false_negative_for_style_yoda_condition.md
@@ -1,0 +1,1 @@
+* [#11511](https://github.com/rubocop/rubocop/issues/11511): Fix a false negative for `Style/YodaCondition` when using constant. ([@koic][])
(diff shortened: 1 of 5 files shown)


bbatsov:
I'd also update the cop examples to reflect this. Does this affect the other Yoga cop as well?

koic:
Indeed! I've update example of this cop and opened https://github.com/rubocop/rubocop/pull/11520 to affect other Yoda cop as the same.

bbatsov:
Thanks!

PR #11516, opened 2023-01-28
Fix missing parentheses in shorthand hash syntax as argument calls

gsamokovarov:
Previously, `Style/HashSyntax` would autocorrect the code:

```ruby
raise Errors.new 'No!', foo: foo
```
to 👇:

```ruby
raise Errors.new 'No!', foo:
```

While it may lead to valid code _in this case_, parentheses around `Error.new` are required in edge cases like condition that we can inline to a modifier form after we apply the shorthand hash syntax.

In this example:

```ruby
if condition?
  raise Error.new 'A long error message...', foo: foo
end
```

we would autocorrect to invalid Ruby code:

```ruby
raise Error.new 'A long error message...', foo: if condition?
```

Putting parentheses in method calls that return value as an argument in another call is the safest autocorrection in this case. I'm starting to wonder whether we always put the parentheses in them method calls as it's generally safer to autocorrect.

### DIFF
--- a/changelog/fix_parentheses_in_shorthand_hash_syntax_as.md
+++ b/changelog/fix_parentheses_in_shorthand_hash_syntax_as.md
@@ -1,0 +1,1 @@
+* [#11516](https://github.com/rubocop/rubocop/pull/11516): Fix missing parentheses in shorthand hash syntax as argument calls. ([@gsamokovarov][])
--- a/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
+++ b/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
@@ -95,16 +95,19 @@ module RuboCop
           use_modifier_form_without_parenthesized_method_call?(method_dispatch_node)
       end
 
+      # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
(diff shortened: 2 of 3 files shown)


bbatsov:
> Putting parentheses in method calls that return value as an argument in another call is the safest autocorrection in this case. I'm starting to wonder whether we always put the parentheses in them method calls as it's generally safer to autocorrect.

I recall in the past one idea was to check if the version without paren is valid or different (at the AST level) from the one with parens to decide whether those are needed or not. That'd be pretty slow, though. No easy solutions...

gsamokovarov:
My idea is to always put the parens when autocorrecting shorthand values and letting the user decide afterwards whether to keep them or not. The idea being that there are other cops that can autocorrect the same code afterward. I have run into other autocorrections that left me with broken code – inlining conditions as modifier forms, being one.

Anyway, let's see how many other edge cases we will find. I am trying to enforce a big codebase into `EnforcedShorthandSyntax: always` style and will get back with the issues I find.

bbatsov:
> My idea is to always put the parens when autocorrecting shorthand values and letting the user decide afterwards whether to keep them or not. The idea being that there are other cops that can autocorrect the same code afterward. I have run into other autocorrections that left me with broken code – inlining conditions as modifier forms, being one.

Yeah, that might reasonable as well, and probably it's the simplest to implement.

PR #11527, opened 2023-01-31
Documentation: Clarify reference to TargetRubyVersion

pboling:
Due to the structural and grammatical similarity of terms it isn't clear that "here" refers to the link and not the section of the page being read. This makes the assumption explicit, which will help those unfamiliar with the reference.

When used in an anchor tag' text "here" commonly refers to what is in the anchor's href target.

However, less familiar RuboCop users may not know that there is a setting literally called `TargetRubyVersion`, while the preceeding sentence uses the exact same language "target ruby version", and is not directly referring to the setting.  This can result in the reference to "here" being ambiguous.

Better to be explicit.

-----------------

* [x] Added tests.


koic:
Thanks!

PR #11529, opened 2023-02-01
Fix an incorrect autocorrect for `Layout/ClassStructure` when definitions that need to be sorted are defined alternately

ydah:
This PR is fix an incorrect autocorrect for `Layout/ClassStructure` when definitions that need to be sorted are defined alternately.

## Code to reproduce

```ruby
class A
  private def foo; end

  def bar; end

  private def baz; end

  def qux; end
end
```

## Expected behavior

Automatically corrected as follows:

```ruby
class A
  def bar; end

  def qux; end

  private def foo; end

  private def baz; end
end
```

## Actual behavior

Automatically corrected as follows:

```ruby
class A
  def qux; end

  def bar; end

  def qux; end # <------ Duplicate

  private def foo; end

  private def baz; end
end
```

-----------------

* [x] Added tests.


koic:
There is some impact on existing autocorrection, but overall I think it's an acceptable trade-off. Thank you.

PR #11530, opened 2023-02-01
Fix an incorrect autocorrect for `Style/AccessModifierDeclarations` when multiple groupable access modifiers are defined

ydah:
This PR is fix an incorrect autocorrect for `Style/AccessModifierDeclarations` when multiple groupable access modifiers are defined.

## Code to reproduce

```ruby
class Test
  private def foo; end
  private def bar; end
  def baz; end
end
```

## Expected behavior

Automatically corrected as follows:

```ruby
class Test
  def baz; end

  private

  def foo; end

  def bar; end
end
```

## Actual behavior

Automatically corrected as follows:

```ruby
class Test
  def baz; end
private

def bar; end # <--- The order is changing
private # <--- Duplicate

def foo; end # <--- The order is changing
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/access_modifier_declarations.rb
+++ b/lib/rubocop/cop/style/access_modifier_declarations.rb
@@ -129,7 +129,8 @@ module RuboCop
         end
 
         def offense?(node)
-          (group_style? && access_modifier_is_inlined?(node)) ||
+          (group_style? && access_modifier_is_inlined?(node) &&
+            !right_siblings_same_inline_method?(node)) ||
             (inline_style? && access_modifier_is_not_inlined?(node))
         end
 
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -129,10 +129,17 @@ def allow_modifiers_on_symbols?(node)
         end
 
         def offense?(node)
-          (group_style? && access_modifier_is_inlined?(node)) ||
+          (group_style? && access_modifier_is_inlined?(node) &&
+            !right_siblings_same_inline_method?(node)) ||
             (inline_style? && access_modifier_is_not_inlined?(node))
         end
 
+        def right_siblings_same_inline_method?(node)
```
koic, line 137: That's just my two cents. Can you move this new method to under the `access_modifier_is_not_inlined` method? (i.e. around L158)

ydah: Thank you. I updated this PR.

koic:
Thanks!

PR #11536, opened 2023-02-03
[Fix #11535] Fix a false positive for `Style/NumberedParametersLimit`

### LINKED ISSUE #11535: Style/NumberedParametersLimit incorrectly(?) counting largest parameter number instead of number of parameters used
Added in https://github.com/rubocop/rubocop/pull/10111 , `Style/NumberedParametersLimit` reports that more than one numbered parameter is used, when there is only one parameter used but it is "higher" than `_1` e.g. `_2`.

For example, `.group_by { _2 }` causes an offense of 
> Avoid using more than 1 numbered parameter; 2 detected.

(but `group_by { _1 }` does not)

I suspect that the intention of the cop was to avoid using more than one numbered parameter in a single block (e.g. `{ _1 + _2 + _3 }`), rather than a single reference to a not-first numbered parameter (e.g. `{ _2 }`).

--------

## Expected behavior

No offense - a single parameter is used

## Actual behavior

An offense was raised.

`TargetRubyVersion` must be (at least) 2.7. Then the following `repro.rb`:

```ruby
grouped = { a: 1, b: 1, c: 2 }.group_by { _2 }
```
gives:
```
$ rubocop --config =(echo -e "AllCops:\n  TargetRubyVersion: 2.7") --debug --cache false --only Style/NumberedParametersLimit re
pro.rb
configuration from /tmp/zshJPIBdI
Default configuration from /Users/owenstephens/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/rubocop-1.41.0/config/default.yml
Inspecting 1 file
Scanning /private/tmp/repro.rb
C

Offenses:

repro.rb:1:11: C: Style/NumberedParametersLimit: Avoid using more than 1 numbered parameter; 2 detected.
grouped = { a: 1, b: 1, c: 2 }.group_by { _2 }
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```
## RuboCop version

```
$ [bundle exec] rubocop -V
1.44.1 (using Parser 3.2.0.0, rubocop-ast 1.24.1, running on ruby 3.1.3) [x86_64-darwin21]
```

koic:
Fixes #11535.

This PR fixes a false positive for `Style/NumberedParametersLimit` when only `_2` or higher numbered parameter is used.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_style_numbered_parameters_limit.md
+++ b/changelog/fix_a_false_positive_for_style_numbered_parameters_limit.md
@@ -1,0 +1,1 @@
+* [#11535](https://github.com/rubocop/rubocop/issues/11535): Fix a false positive for `Style/NumberedParametersLimit` when only `_2` or higher numbered parameter is used. ([@koic][])
--- a/lib/rubocop/cop/style/numbered_parameters_limit.rb
+++ b/lib/rubocop/cop/style/numbered_parameters_limit.rb
@@ -12,10 +12,11 @@ module RuboCop
       #
       # @example Max: 1 (default)
       #   # bad
-      #   foo { _1.call(_2, _3, _4) }
+      #   use_multiple_numbered_parameters { _1.call(_2, _3, _4) }
       #
       #   # good
-      #   foo { do_something(_1) }
+      #   array.each { use_array_element_as_numbered_parameter(_1) }
+      #   hash.each { use_only_hash_value_as_numbered_parameter(_2) }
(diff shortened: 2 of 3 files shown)


bbatsov:
Seems like something that we should mention in the examples as well. I definitely hadn't thought of this use-case until today. :-)

koic:
Yeah, I've updated the examples for the cop :-)

bbatsov:
Thanks!

PR #11518, opened 2023-01-29
[Fix #11517] Make `Lint/Debugger` aware of some debug methods

koic:
Fixes #11517.

This PR makes `Lint/Debugger` aware of `p`, `PP.pp`, and `pp` methods. These are used for debugging purposes basically. OTOH, `puts` is not included in the config because it is used for the intended standard output.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [ ] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #11552, opened 2023-02-08
Fix a false positive for `Lint/Debugger` when methods containing different method chains

ydah:
This PR is fix a false positive for `Lint/Debugger` when methods containing different method chains.

## Motivation
I don't think that methods such as `p.do_something` for example should be an offense of this cop. I think it would be better to make it a violation only if it is an exact match, including the method chain, but what do you think?

## This PR will result in the following

```ruby
# bad
p 'foo'
foo(p 'foo')
p(p 'foo')

# good
p.do_something # <--- Before the change, it is a bad code
Foo.p
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_lint_debugger.md
+++ b/changelog/fix_a_false_positive_for_lint_debugger.md
@@ -1,0 +1,1 @@
+* [#11552](https://github.com/rubocop/rubocop/pull/11552): Fix a false positive for `Lint/Debugger` when methods containing different method chains. ([@ydah][])
--- a/lib/rubocop/cop/lint/debugger.rb
+++ b/lib/rubocop/cop/lint/debugger.rb
@@ -82,44 +82,25 @@ module RuboCop
         def debugger_methods
           @debugger_methods ||= begin
             config = cop_config.fetch('DebuggerMethods', [])
-            values = config.is_a?(Array) ? config : config.values.flatten
-            values.map do |v|
-              next unless v
-
-              *receiver, method_name = v.split('.')
-              {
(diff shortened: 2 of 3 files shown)


bbatsov:
Yeah, I agree that's an improvement. Thanks!

hartator:
Thanks for all the work @ydah. Good job and it's deeply appreciated. ❤️

This has immediately created an issue on our end unfortunately. We use `Lint/Debugger` to catch potential forgotten `binding.pry`. However, we also do have legitimate utility scripts that uses `p` and `pp` to add some verbosity to their outputs. These are not forgotten and are on purpose.

Maybe it makes sense to move `p` and `pp` outside of `Lint/Debugger`? To something like `Lint/DebuggerPrints`? And keep `Lint/Debugger` only for REPLs?

koic:
`p` and `pp` were added for debugging methods that call `inspect` (not `to_s`) internally.
https://github.com/rubocop/rubocop/pull/11518

hartator:
> `p` and `pp` were added for debugging methods that call `inspect` (not `to_s`) internally.
> #11518

Why not catching `.inspect` as well then? Following this logic, `.inspect` should be also used only for debugging and shouldn't be in a production codebase. However, there are out there legitimate use cases of this as well. I would believe this opens up a bad rabbit hole.

Can you guys reconsider? The current implementation will make us remove `Lint/Debugger` as a linter unfortunately.

koic:
I reconsidered based on some feedback and opened https://github.com/rubocop/rubocop/pull/11564. Thank you.

hartator:
> I reconsidered based on some feedback and opened #11564. Thank you.

Thank _you_. 👍

PR #11557, opened 2023-02-08
[Fix #11556] Fix a false positive for `Lint/Debugger`

### LINKED ISSUE #11556: False positive in Lint/Debugger
There is a regression caused by #11555, where it thinks that any use of the local variable `p` is a method call to the debugger.

Example:
```
spec/views/spotlight/sir_trevor/blocks/_solr_documents_features_block.html.erb_spec.rb:49:23: W: Lint/Debugger: Remove debugger entry point p.
      render partial: p, locals: { item_carousel_block: block }
                      ^
```

koic:
Fixes #11556.

This PR fixes a false positive for `Lint/Debugger` when `p` is an argument of method call. Basically, debugger methods are not used as a method argument without arguments.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_lint_debugger.md
+++ b/changelog/fix_a_false_positive_for_lint_debugger.md
@@ -1,0 +1,1 @@
+* [#11556](https://github.com/rubocop/rubocop/issues/11556): Fix a false positive for `Lint/Debugger` when `p` is an argument of method call. ([@koic][])
--- a/lib/rubocop/cop/lint/debugger.rb
+++ b/lib/rubocop/cop/lint/debugger.rb
@@ -70,6 +70,9 @@ module RuboCop
         def on_send(node)
           return unless debugger_method?(node)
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Looks okay, sans the CI failures.

I'm starting to wonder, though, if it was a good idea to mix the debug methods that actually start some debugger (and don't take arguments) with print debugging methods (which take some arguments), as for them we could just check if they are invoked without any arguments. Anyways, the fix will do for now.

koic:
> the debug methods that actually start some debugger (and don't take arguments) with print debugging methods (which take some arguments)

Ah, it may have been better to distinguish between them. For now, I'd like to look feedback after https://github.com/rubocop/rubocop/issues/11556 issue is resolved.

Greg-Myers-SB:
I'm not sure this was intentional but this appears to have changed the behaviour of using pretty_print `pp MyCode` used to trigger, now it no longer does.
I'm not sure if PrettyPrint was supposed to trigger this rule in the first place though?

PR #11559, opened 2023-02-08
Fix octal handling in `Style/RedundantRegexpCharacterClass`

jaynetics:
regexp_parser v2.7.0 introduced a fix that makes it correctly detect octal escapes (e.g. "\141", which equals "a") inside character classes, e.g. `/[\141]/`.

Previously, these were incorrectly interpreted and chunked, c.f. https://github.com/ammar/regexp_parser/issues/86

The incorrect chunking made your build green. This is because `\032` emitted `\0`, which was checked for specifically in `#backslash_zero?`.

The fix in regexp_parser v2.7.0, however, made this spec red.

The correct output by rubocop would be to detect this case as a redundant use of a char class, because it is.

The same goes for the original example "\0":

```ruby
[/\0/.match?("\0"), /[\0]/.match?("\0")] # => [true, true]
```

For `\1` to `\7`, putting them in a char class is not redundant but actually required, because they would otherwise be treated as backreferences.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_regexp_character_class.rb
+++ b/lib/rubocop/cop/style/redundant_regexp_character_class.rb
@@ -74,7 +74,7 @@ module RuboCop
 
           non_redundant =
             whitespace_in_free_space_mode?(node, class_elem) ||
-            backslash_b?(class_elem) || backslash_zero?(class_elem) ||
+            backslash_b?(class_elem) || octal_requiring_char_class?(class_elem) ||
             requires_escape_outside_char_class?(class_elem)
 
           !non_redundant
@@ -104,11 +104,10 @@ module RuboCop
           elem == '\b'
         end
 
-        def backslash_zero?(elem)
-          # See https://github.com/rubocop/rubocop/issues/11067 for details - in short "\0" != "0" -
-          # the former means an Unicode code point `"\u0000"`, the latter a number character `"0"`.
-          # Similarly "\032" means "\u001A". Other numbers starting with "\0" can also be mentioned.
-          elem == '\0'
+        def octal_requiring_char_class?(elem)
+          # The octal escapes \1 to \7 only work inside a character class
+          # because they would be a backreference outside it.
+          elem.match?(/\A\\[1-7]\z/)
         end
 
         def requires_escape_outside_char_class?(elem)
--- a/changelog/fix_octal_handling_in_regexp_char_classes_.md
+++ b/changelog/fix_octal_handling_in_regexp_char_classes_.md
@@ -1,0 +1,1 @@
+* [#11559](https://github.com/rubocop/rubocop/pull/11559): Fixed false positives and negatives in `Style/RedundantRegexpCharacterClass` when using octal escapes (e.g. "\0"). ([@jaynetics][])
--- a/spec/rubocop/cop/style/redundant_regexp_character_class_spec.rb
+++ b/spec/rubocop/cop/style/redundant_regexp_character_class_spec.rb
@@ -307,12 +307,29 @@ RSpec.describe RuboCop::Cop::Style::RedundantRegexpCharacterClass, :config do
     end
   end
 
-  context 'with a character class containing an escaped-0' do
+  context 'with a character class containing an octal escape sequence that also works outside' do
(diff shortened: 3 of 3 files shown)


lib/rubocop/cop/style/redundant_regexp_character_class.rb
```diff
@@ -104,11 +104,10 @@ def backslash_b?(elem)
...
-        def backslash_zero?(elem)
-          # See https://github.com/rubocop/rubocop/issues/11067 for details - in short "\0" != "0" -
-          # the former means an Unicode code point `"\u0000"`, the latter a number character `"0"`.
-          # Similarly "\032" means "\u001A". Other numbers starting with "\0" can also be mentioned.
-          elem == '\0'
+        def octal_requiring_char_class?(elem)
+          # The octal escapes \1 to \7 only work inside a character class
+          # because they would be a backreference outside it.
+          elem.match?(/\A\\[1-7]\z/)
         end
```
koic, line 111: As far as I remember `add_runtime_dependency` cannot be updated to `>= 2.7`: https://github.com/rubocop/rubocop/pull/9154 
Isn't backwards compatibility like below needed?

```suggestion
        if Gem::Version.new(Regexp::Parser::VERSION) >= Gem::Version.new('2.7')
          def octal_requiring_char_class?(elem)
            # The octal escapes \1 to \7 only work inside a character class
            # because they would be a backreference outside it.
            elem.match?(/\A\\[1-7]\z/)
          end
        else
          def octal_requiring_char_class?(elem)
           # See https://github.com/rubocop/rubocop/issues/11067 for details - in short "\0" != "0" -
           # the former means an Unicode code point `"\u0000"`, the latter a number character `"0"`.
           # Similarly "\032" means "\u001A". Other numbers starting with "\0" can also be mentioned.
           elem == '\0'
          end
        end
```

koic:
Can you add a changelog entry?

lib/rubocop/cop/style/redundant_regexp_character_class.rb
jaynetics: no, because `\0` (or anything starting with `\0`) acts the same way inside and outside a char class.

basically, this commit fixes both the spec (which was only green due to a bug in regexp_parser), as well as the behavior.

before this commit:

```ruby
/[\0]/   # no offense (wrong - this char class *is* redundant)
/[\032]/ # no offense (wrong - this char class *is* redundant)
/[\1]/   # offense (wrong - this char class *is* needed)
/[\132]/ # no offense (wrong - this char class *is* redundant)
/[\9]/   # offense (correct - no octal, so both char class and escaping are redundant)
```

after:

```ruby
/[\0]/   # offense (correct)
/[\032]/ # no offense on regexp_parser 2.6 (still wrong), offense on 2.7 (correct)
/[\1]/   # no offense (correct)
/[\132]/ # no offense (correct)
/[\9]/   # offense (correct)
```

jaynetics:
> Can you add a changelog entry?

ok, i just did.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#11559](https://github.com/rubocop/rubocop/pull/11559): Fixed false positives and negatives in `Style/RedundantRegexpCharacterClass` when using octal escapes (e.g. "\0"). ([@jaynetics][])
```
koic, line 5: Can you write the changelog entry using the below instead of directly?

> Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named {change_type}_{change_description}.md if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

jaynetics: should i create this folder?

Screenshot 2023-02-09 at 09 59 54

lib/rubocop/cop/style/redundant_regexp_character_class.rb
koic: I get it. Thank you for the explanation!

CHANGELOG.md
jaynetics: done

koic:
Great! Thank you!

PR #11562, opened 2023-02-09
Fix handling of escaped octals in duplicate regexp cop

rwstauner:
This improves the handling of escaped octals (as long as we allow regexp_parser < 2.7)
for the duplicate regexp character class element cop.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/duplicate_regexp_character_class_element.rb
+++ b/lib/rubocop/cop/lint/duplicate_regexp_character_class_element.rb
@@ -32,35 +32,53 @@ module RuboCop
           end
         end
 
-        # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         def each_repeated_character_class_element_loc(node)
           node.parsed_tree&.each_expression do |expr|
             next if skip_expression?(expr)
 
             seen = Set.new
-            enum = expr.expressions.to_enum
-            expression_count = expr.expressions.count
+            group_expressions(node, expr.expressions) do |group|
+              group_source = group.map(&:to_s).join
 
-            expression_count.times do |current_number|
-              current_child = enum.next
-              next if within_interpolation?(node, current_child)
+              yield source_range(group) if seen.include?(group_source)
 
-              current_child_source = current_child.to_s
-              next_child = enum.peek if current_number + 1 < expression_count
-
-              if seen.include?(current_child_source)
-                next if start_with_escaped_zero_number?(current_child_source, next_child.to_s)
-
-                yield current_child.expression
-              end
-
-              seen << current_child_source
+              seen << group_source
             end
           end
         end
-        # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
 
         private
+
+        def group_expressions(node, expressions)
+          expressions = expressions.to_a
+
+          until expressions.empty?
+            group = [expressions.shift]
+            next if within_interpolation?(node, group.first)
+
+            pop_octal_digits(group, expressions) if escaped_octal?(group.first.to_s)
+
+            yield(group)
+          end
+        end
+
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/duplicate_regexp_character_class_element.rb
```diff
@@ -32,35 +32,53 @@ def on_regexp(node)
...
+            end
+          end
+        end
+
+        private
 
-              if seen.include?(current_child_source)
-                next if start_with_escaped_zero_number?(current_child_source, next_child.to_s)
+        def group_expressions(node, expressions)
+          expressions = expressions.to_a
```
koic, line 53: This code seems unnecessary, right?

```suggestion
```

lib/rubocop/cop/lint/duplicate_regexp_character_class_element.rb
```diff
@@ -32,35 +32,53 @@ def on_regexp(node)
...
 
-                yield current_child.expression
-              end
+          until expressions.empty?
+            group = [expressions.shift]
+            next if within_interpolation?(node, group.first)
 
-              seen << current_child_source
-            end
+            pop_octal_digits(group, expressions) if escaped_octal?(group.first.to_s)
```
koic, line 59: How about leaving a comment like this?

```suggestion
            # Compatible logic for regexp_parser < 2.7.
            pop_octal_digits(group, expressions) if escaped_octal?(group.first.to_s)
```

lib/rubocop/cop/lint/duplicate_regexp_character_class_element.rb
```diff
@@ -32,35 +32,53 @@ def on_regexp(node)
...
+
+            yield(group)
           end
         end
-        # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
 
-        private
+        def pop_octal_digits(current_child, expressions)
+          2.times do
```
koic, line 66: It looks better when named with `2` as a constant

lib/rubocop/cop/lint/duplicate_regexp_character_class_element.rb
```diff
@@ -75,9 +93,12 @@ def within_interpolation?(node, child)
           interpolation_locs(node).any? { |il| il.overlaps?(parse_tree_child_loc) }
         end
 
-        def start_with_escaped_zero_number?(current_child, next_child)
-          # Represents escaped code from `"\00"` (`"\u0000"`) to `"\07"` (`"\a"`).
-          current_child == '\\0' && next_child.match?(/[0-7]/)
+        def escaped_octal?(string)
+          string.size == 2 && string[0] == '\\' && octal?(string[1])
```
koic, line 97: That's just my two cents:

```suggestion
          string.length == 2 && string[0] == '\\' && octal?(string[1])
```

lib/rubocop/cop/lint/duplicate_regexp_character_class_element.rb
rwstauner: i was using the array as a "enum" so i can just remove items until it's empty.
it required less state tracking than the original enumerator.

rwstauner: part of why i ended up here was because the original method was triggering several linters about method length and complexity, but i couldn't update the counter from a separate method, so in breaking it into smaller methods using a mutable array seemed simplest.

bbatsov:
As I'm not familiar with the nuances of `regexp_parser` I'll defer to @koic here.

bbatsov:
@koic Are we good to merge this PR?

koic:
Yep! Let's take this change!

koic:
Thanks!

PR #11565, opened 2023-02-10
Add new `Lint/RedundantRegexpQuantifiers` cop

jaynetics:
This cop checks for redundant quantifiers (a.k.a. repeat operators) in regular expression literals.

Such redundant quantifiers also trigger a warning in the Ruby parser / Onigmo:

```sh
ruby -e '/(?:a+)+/'
-e:1: warning: regular expression has redundant nested repeat operator '+': /(?:a+)+/
```
-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -4880,6 +4880,11 @@ Style/RedundantRegexpEscape:
   Enabled: true
   VersionAdded: '0.85'
 
+Style/RedundantRegexpQuantifiers:
+  Description: 'Checks for redundant quantifiers in Regexps.'
+  Enabled: true
```
jaynetics, line 4885: or `Enabled: pending`? not sure what to choose

lib/rubocop/cop/style/redundant_regexp_quantifiers.rb
```diff
@@ -0,0 +1,120 @@
...
+        def on_regexp(node)
+          each_redundantly_quantified_pair(node) do |group, child|
+            replacement = merged_quantifier(group, child)
+            add_offense(
+              quantifier_range(group, child),
+              message: message(group, child, replacement)
+            ) do |corrector|
+              # drop outer quantifier
+              corrector.replace(group.loc.quantifier, '')
+              # replace inner quantifier
```
jaynetics, line 43: it could also be done the other way round - replacing the outer quantifier and dropping the inner quantifier.

dropping the outer quantifier: `/(?:a+)?/ => /(?:a*)/`
dropping the inner quantifier: `/(?:a+)?/ => /(?:a)*/`

both are equivalent. i went with the first option but i don't think there's a big difference between them.

lib/rubocop/ext/regexp_parser.rb
```diff
@@ -68,7 +68,9 @@ def build_location
             return { body: expression } unless (q = quantifier)
 
             body = expression.adjust(end_pos: -q.text.length)
-            q_loc = expression.with(begin_pos: body.end_pos)
+            q.origin = origin
```
jaynetics, line 71: these changes fix a bug:

as i noticed in [this test](https://github.com/rubocop/rubocop/pull/11565/files#diff-2955709183e371bdcab246c6a457bf346eb5ed3dbcd6626bf2db521f1163ae9dR124), the quantifier location was previously incorrect if the regexp was in x-mode and there was whitespace between the quantified expression and the quantifier.

e.g. `/a   +/x`.

in such cases, the starting position of the quantifier is *not* the end position of the body.

as far as i could tell, this quantifier location was not used by any existing cop so far, so i though this bugfix was not worth its own PR.

lib/rubocop/cop/style/redundant_regexp_quantifiers.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for redundant quantifiers inside Regexp literals.
+      #
+      # @example
+      #   # bad
+      #   /(?:x+)+/
```
jaynetics, line 10: unfortunately, some of these examples emit warnings during the `documentation_syntax_check` task.

is there an established way to prevent this?

config/default.yml
bbatsov: All new cops should be marked as pending.

lib/rubocop/cop/style/redundant_regexp_quantifiers.rb
bbatsov: Don't think so. @koic can correct me about this, but in general we have a ton of AsciiDocs warnings these days that we'll have to address at some point.

lib/rubocop/cop/style/redundant_regexp_quantifiers.rb
```diff
@@ -0,0 +1,120 @@
...
+      #   # good
+      #   /(?:x)*/
+      #
+      #   # good
+      #   /(?:x*)/
+      class RedundantRegexpQuantifiers < Base
+        include RangeHelp
+        extend AutoCorrector
+
+        MSG_REDUNDANT_QUANTIFIER = 'Redundant quantifiers in Regexp literal, ' \
```
bbatsov, line 30: Is this the message that Ruby uses? If not - I'd reword it to shorten it. E.g. "Remove redundant ...".

jaynetics:
@bbatsov 

thanks for the feedback.

i made it a Lint cop, set it to pending, and shortened the message.

> @koic can correct me about this, but in general we have a ton of AsciiDocs warnings these days that we'll have to address at some point.

the examples of this cop emit warnings were no warnings were emitted before, as seen in the lower lines of this shell output:

```sh
$ bundle exec rake documentation_syntax_check
Files:         651
Modules:        90 (   13 undocumented)
Classes:       614 (    4 undocumented)
Constants:    1088 ( 1072 undocumented)
Attributes:     35 (    0 undocumented)
Methods:      1391 ( 1227 undocumented)
 28.03% documented
~/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/parser-3.2.0.0/lib/parser/builders/default.rb:2249: warning: regular expression has redundant nested repeat operator '+': /(?:x+)+/
~/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/parser-3.2.0.0/lib/parser/builders/default.rb:2249: warning: nested repeat operator '+' and '?' was replaced with '*' in regular expression: /(?:x+)?/
$ echo $? # => 0
```

other existing examples that trigger warnings in Ruby, such as [this one](https://github.com/rubocop/rubocop/blob/f44b9e51efeaa38190b1d523d46cacaea596a397/lib/rubocop/cop/lint/duplicate_hash_key.rb#L14), do not pop up here. that's why i thought maybe there is some ignorelist or suppression, but i couldn't find anything.

jaynetics:
@koic @bbatsov would you like to merge this? should we suppress the warnings caused by "bad code examples" mentioned in the comment above or just live with them?

bbatsov:
> would you like to merge this?

Yeah, sorry about the slow turnaround here!

>  should we suppress the warnings caused by "bad code examples" mentioned in the comment above or just live with them?

Ideally yes, if we figure out how to do it. :-) It's not a stopper to merge this for now, though - I'm preparing for a new release soon.

bbatsov:
> Such redundant quantifiers also trigger a warning in the Ruby parser / Onigmo:

You might also mention this in the cop description. More rationale is always better. (a follow-up PR is fine)

PR #11574, opened 2023-02-13
Fix a broken shorthand syntax autocorrection

gsamokovarov:
Calls like `foo(bar: bar).baz` were autocorrected to `foo(bar:)).baz`. Notice the extra parentheses. I have broken this in #11516.


koic:
Thanks!

PR #11575, opened 2023-02-13
Fix parentheses in value omissions for multiple assignments

gsamokovarov:
I'm on a quest to run the `EnforcedShorthandSyntax: always` style on a big codebase. One of the latest thins I found was a value omission in multiple assignments. Not the best idea, but it is a real "production" code usage, I guess. Here is an estimation of the code:

```ruby
foo = foo = do_stuff arg, opt1: opt1,
                          opt2: opt2

something.each { pass }
```

The multiple assignment in the code above tricks the current version of the parentheses detection code and we don't put the parens when autocorrecting:

```ruby
foo = foo = do_stuff arg, opt1:,
                          opt2:

something.each { pass }
```

While not a syntax error, Ruby is treating it like:

```ruby
foo = foo = do_stuff arg, opt1:,
                          opt2: something.each { pass }
```

We do need to put the parentheses for end result like:

```ruby
foo = foo = do_stuff(arg, opt1:,
                          opt2:)

something.each { pass }
```

### DIFF
--- a/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
+++ b/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
@@ -101,7 +101,7 @@ module RuboCop
         return unless last_pair.key.source == last_pair.value.source
         return unless (dispatch_node = find_ancestor_method_dispatch_node(node))
         return if dispatch_node.parenthesized?
-        return unless last_expression?(dispatch_node) || method_dispatch_as_argument?(dispatch_node)
+        return if last_expression?(dispatch_node) && !method_dispatch_as_argument?(dispatch_node)
 
         def_node = node.each_ancestor(:send, :csend, :super, :yield).first
 
@@ -132,9 +132,12 @@ module RuboCop
         ancestor.ancestors.any? { |node| node.respond_to?(:modifier_form?) && node.modifier_form? }
       end
 
-      def last_expression?(ancestor)
-        ancestor.right_sibling ||
-          ancestor.each_ancestor.find { |node| node.assignment? || node.send_type? }&.right_sibling
+      def last_expression?(node)
+        return false if node.right_sibling
+        return true unless (assignment_node = node.each_ancestor.find(&:assignment?))
+        return last_expression?(assignment_node.parent) if assignment_node.parent&.assignment?
+
+        !assignment_node.right_sibling
       end
 
       def method_dispatch_as_argument?(method_dispatch_node)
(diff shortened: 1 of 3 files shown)


gsamokovarov:
The failing test seems _suspiciously_ related. Working on it! 😅

gsamokovarov:
I have fixed it, the failing test was legitimate.

spec/rubocop/cop/style/hash_syntax_spec.rb
```diff
@@ -1154,6 +1154,22 @@ def do_something
         RUBY
       end
 
+      it 'registers an offense when expression follows multiple assignments' do
+        expect_offense(<<~RUBY)
+          foo = foo = do_stuff arg, opt1: opt1,
```
bbatsov, line 1159: I think `foo = bar =` would read a bit better here.

lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
```diff
@@ -133,8 +133,16 @@ def use_modifier_form_without_parenthesized_method_call?(ancestor)
       end
 
       def last_expression?(ancestor)
-        ancestor.right_sibling ||
-          ancestor.each_ancestor.find { |node| node.assignment? || node.send_type? }&.right_sibling
+        return false if ancestor.right_sibling
```
bbatsov, line 136: As the logic in the method became a bit dense I'd sparkle a few comments describing what exactly are we checking for in the various branches.

spec/rubocop/cop/style/hash_syntax_spec.rb
gsamokovarov: `price = price = generate_something_with args, and_shorthands:` read the production code. :D

lib/rubocop/cop/mixin/hash_shorthand_syntax.rb
gsamokovarov: I have simplified the method, I hope it's self explanatory now. The idea is that if you have a right node, there is code afterwards, unless the current node is the right node of an assignment. Then you have to check the right sibling of the assignment node itself. This can be recursive, because the current node can be a right sibling of an assignment itself, like the multiple assignment case I hit.

@koic there used to be a condition about being the right node of a `send` (method dispatch). Removing this does not fail any tests, are you sure the check was needed and the current approach is exhaustive?

koic: I would have written test as much as possible, so it may be removed I think. If the deletion caused an error, a test may have been forgotten.

PR #11588, opened 2023-02-17
Optimize `Style/WordArray` complex matrix check

sambostock:
One of the checks we do in `Style/WordArray` is to check if an otherwise simple array is a member of a parent array consisting exclusively of arrays, including at least one with "complex" content.

This means that for each sub-array, we navigate up to the parent, and check all of its children, resulting in `n²` array complexity checks.

By maintaining a cache of checked parent array nodes per investigation (i.e. per file), we can reduce this to `n` array complexity checks.

This has a massive impact on performance when investigating files containing large arrays of arrays. A test is added with a timeout that does not pass when using the non-caching implementation **(it would take over a minute; now takes under a second)**.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/word_array.rb
+++ b/lib/rubocop/cop/style/word_array.rb
@@ -82,10 +82,19 @@ module RuboCop
           attr_accessor :largest_brackets
         end
 
+        def on_new_investigation
+          super
+
+          # Prevent O(n2) checks (checking the entire matrix once for each child array) by caching
+          @matrix_of_complex_content_cache = Hash.new do |cache, node|
+            cache[node] = matrix_of_complex_content?(node)
+          end
+        end
+
         def on_array(node)
           if bracketed_array_of?(:str, node)
             return if complex_content?(node.values)
-            return if within_2d_array_of_complex_content?(node)
+            return if within_matrix_of_complex_content?(node)
 
             check_bracketed_array(node, 'w')
           elsif node.percent_literal?(:string)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/word_array.rb
```diff
@@ -82,10 +82,16 @@ class << self
           attr_accessor :largest_brackets
         end
 
+        def on_new_investigation
+          super
+
+          reset_matrix_of_complex_content_cache!
+        end
```
fatkodima, line 89: Up to you, but `super` is not needed in this case and I would inline `reset_matrix_of_complex_content_cache!` here.

fatkodima (commented):
I am the one who introduced this O(n2) 😄 Lgtm 👍

lib/rubocop/cop/style/word_array.rb
sambostock: https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/base.rb#L135-L139 specifically says to always call `super`, which is why I did it, and extracted the method.

Fine with inlining though, if you think that's better. It just felt weird to have logic in `on_new_investigation`.

fatkodima: > specifically says to always call super, which is why I did it, and extracted the method.

Ah, good, looks like this is currently missing in most of the places.

> It just felt weird to have logic in on_new_investigation

Seems like thats how it is done throughout the other places.

bbatsov:
Nice catch!

Issue #3191, opened 2016-06-02
allow comments after rubocop:disable


grosser:
a few times I wanted to add a trailing comment to a disable, but rubocop does not allow that ...

```
Translation # rubocop:disable Lint/Void needs to be loaded before un-marshaling
```

but that results in:

```
xyz.rb:87:21: W: Lint/UnneededDisable: Unnecessary disabling of Lint/Void needs to be loaded before un (unknown cop).
```

rubocop: 0.40

PR #11590, opened 2023-02-17
Document that disable comments can include a descriptive comment

bensheldon:
Updates the documentation to reflect that disabling comments _can_ include descriptive text after the cop name. This feature was introduced in #3191 and as far as I can tell is still supported. 

_This is only a documentation update, I'll do my best to pick the relevant tasks below 🙇🏻_

-----------------

* [ ] Added tests.

### DIFF
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -739,11 +739,11 @@ Running `rubocop --autocorrect --disable-uncorrectable` will
 create comments to disable all offenses that can't be automatically
 corrected.
 
-Do not write anything other than cop name in the disabling comment. E.g.:
+You can also include a comment to explain why the cop has been disabled. E.g.:
 
 [source,ruby]
 ----
-# rubocop:disable Layout/LineLength --This is a bad comment that includes other than cop name.
+# rubocop:disable Layout/LineLength This is a comment that explains why Layout/LineLength is disabled.
 ----
 
 == Temporarily enabling cops in source code


docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -739,11 +739,11 @@ Running `rubocop --autocorrect --disable-uncorrectable` will
 create comments to disable all offenses that can't be automatically
 corrected.
 
-Do not write anything other than cop name in the disabling comment. E.g.:
+Add a comment to explain why the cop has been disabled E.g.:
```
bbatsov, line 742: I'd change the wording to something like "You can also...", so it sounds like a suggestion. I think you removed the `.` before e.g. by accident.

bensheldon: 👍🏻 I updated the language to be `You can also include...` and re-added the period.

bensheldon:
@bbatsov I did some more digging into the git history and learned that the documentation I'm updating is actually much more recent than expected. It was introduced in 2020: https://github.com/rubocop/rubocop/pull/7995, whereas the behavior to allow the descriptive text was introduced in 2016.

Am I being accurate that disabling comments _are_ intended to accept an additional "reason/comment"?

bbatsov:
Yeah, it seems that actually that's not really supported. My bad. I keep forgetting how certain features work. @koic can add more on the topic I guess.

koic:
https://github.com/rubocop/rubocop/issues/7993 was misdetecting comments as cop name, https://github.com/rubocop/rubocop/pull/7995 fixes that.

Perhaps I wanted to be told about not recommending any description other than cop name. So, it can be written free text comment on the same line, but it is not recommended. It's probably better to write it on a separate line.

bbatsov:
@koic Agreed.

bensheldon:
> So, it can be written free text comment on the same line, but it is not recommended. It's probably better to write it on a separate line.

That sounds like it _is_ a supported feature, but not recommended (very Rubocop-y 😄 ). I can try to capture that in the documentation.

Is the reason purely that it's prone-to-breakage?

I think the trailing reason comment is very useful to have on the same-line (it's visually closest to the code being commented; it could be removed via `Lint/UnneededDisable`; it _could_ be parsed/reported). So I like it as a feature, but I also trust you all if you don't think it's possible to support it. 

I see two options and I'd like your advice:

- Capture in the documentation that it is a feature, but it's not recommended because it's easy to mess up the parsing. I can also add some tests and attempt to make it _less_ breakable and better define what is acceptable syntax.
- Close this PR because ultimately it's not a supported feature.

Thanks for chatting through this with me!

bbatsov:
> Is the reason purely that it's prone-to-breakage?

For me the main problem is that this makes sense only for a single cop/department being listed on the line, but what happens if you have several? And obviously it's easier to parse the directive contents if you know there's no free-form text somewhere in there. 

I do get your point, however, and I wonder what's the best approach in general. Normally I put some explanatory comments before directives and that seems fine to me overall.

PR #11595, opened 2023-02-18
Add new `Style/RedundantRegexpArgument` cop

koic:
This PR adds new `Style/RedundantRegexpArgument` cop.

It identifies places where argument can be replaced from a deterministic regexp to a string.

```ruby
# bad
'foo'.byteindex(/f/)
'foo'.byterindex(/f/)
'foo'.gsub(/f/, 'x')
'foo'.gsub!(/f/, 'x')
'foo'.index(/f/)
'foo'.partition(/f/)
'foo'.rindex(/f/)
'foo'.rpartition(/f/)
'foo'.scan(/f/)
'foo'.split(/f/)
'foo'.sub(/f/, 'x')
'foo'.sub!(/f/, 'x')

# good
'foo'.byteindex('f')
'foo'.byterindex('f')
'foo'.gsub('f', 'x')
'foo'.gsub!('f', 'x')
'foo'.index('f')
'foo'.partition('f')
'foo'.rindex('f')
'foo'.rpartition('f')
'foo'.scan('f')
'foo'.split('f')
'foo'.sub('f', 'x')
'foo'.sub!('f', 'x')
```

This is a superset of `Performance/RedundantSplitRegexpArgument` which only handles `String#split`: https://docs.rubocop.org/rubocop-performance/1.16/cops_performance.html#performanceredundantsplitregexpargument

Changing the argument from regexp to string have a positive effect on readability as well as performance. So this cop is promoting Style department, not Performance department.

In the future `Performance/RedundantSplitRegexpArgument` can be deprecated and replaced by this cop.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/redundant_regexp_argument.rb
```diff
@@ -0,0 +1,88 @@
...
+      #   'foo'.sub('f', 'x')
+      #   'foo'.sub!('f', 'x')
+      class RedundantRegexpArgument < Base
+        extend AutoCorrector
+
+        MSG = 'Use string `%<prefer>s` as argument instead of regexp `%<current>s`.'
+        RESTRICT_ON_SEND = %i[
+          byteindex byterindex gsub gsub! index partition rindex rpartition scan split sub sub!
+        ].freeze
+        DETERMINISTIC_REGEX = /\A(?:#{LITERAL_REGEX})+\Z/.freeze
```
jaynetics, line 44: it is maybe not directly relevant to this PR as it already existed before, but the `LITERAL_REGEX` has both false positives and false negatives.

```ruby
# current regex
LITERAL_REGEX = %r{[\w\s\-,"'!#%&<>=;:`~/]|\\[^AbBdDgGhHkpPRwWXsSzZ0-9]}

# some false positives
LITERAL_REGEX.match?(/foo # comment/x.source) # => true
LITERAL_REGEX.match?(/a\Kb/.source) # => true (\K is a lookbehind)

# some false negatives
LITERAL_REGEX.match?(/😃/.source) # => false
LITERAL_REGEX.match?(/]/.source) # => false
LITERAL_REGEX.match?(/@/.source) # => false
LITERAL_REGEX.match?(/ü/.source) # => false
```

it might make sense to improve `LITERAL_REGEX` a bit, or maybe `regexp_parser` could be used, approximately like so (i haven't tested this):

```ruby
Regexp::Parser.parse(regexp).each_expression.all? do |(exp)|
  exp.one_of?(:literal, :escape) && !exp.quantified?
end
```

lib/rubocop/cop/style/redundant_regexp_argument.rb
```diff
@@ -0,0 +1,88 @@
...
+        STR_SPECIAL_CHARS = %w[\n \" \' \\\\ \t \b \f \r].freeze
+
+        def on_send(node)
+          return unless (regexp_node = node.first_argument)
+          return unless regexp_node.regexp_type?
+          return if regexp_node.ignore_case? || regexp_node.content == ' '
```
jaynetics, line 50: i'm not sure what `regexp_node.content == ' '` implies, but maybe `''` was intended instead of `' '`, to allow something like `split(//)`?

bbatsov:
@koic Please address the feedback provided by @jaynetics.

lib/rubocop/cop/style/redundant_regexp_argument.rb
koic: `/foo # comment/` does not seem to be a false positive. OTOH, regular expression `/a\Kb/` may indeed be a false positive, or an incorrect autocorrection. `LITERAL_REGEXP` is a constant used in some existing RuboCop Performances cops, so I would like to modify it separately.

lib/rubocop/cop/style/redundant_regexp_argument.rb
koic: It prevents false positives in regex matches for spaces.

```ruby
'foo         bar'.split(/ /) # => ["foo", "", "", "", "", "", "", "", "", "bar"]
'foo         bar'.split(' ') # => ["foo", "bar"]
```

The test is written below for that.
https://github.com/rubocop/rubocop/pull/11595/files#diff-86f0d8dfae0d2d9eb2236fbfd1ae5c61165805156ed3c04b0847e6f82823252eR121-R125

lib/rubocop/cop/style/redundant_regexp_argument.rb
jaynetics: > /foo # comment/ does not seem to be a false positive

It is not a false positive in the sense that it is impossible to replace it with a string, but IIUC the cop would replace it with `'foo # comment'`, which would not behave the same?

lib/rubocop/cop/style/redundant_regexp_argument.rb
jaynetics: Ah, i had no idea that split had a custom "awk-like" behavior for single spaces!

lib/rubocop/cop/style/redundant_regexp_argument.rb
koic: For example, the following are compatible. What code specifically causes the problem?

```ruby
'foo # comment extra'.match?(/foo # comment/) # => true
'foo # comment extra'.match?('foo # comment') # => true
```

jaynetics: `/foo # comment/x`

jaynetics: also `/ foo /x`

koic: Ah, sure! It changed to allow when options are present.

lib/rubocop/cop/style/redundant_regexp_argument.rb
```diff
@@ -0,0 +1,88 @@
...
+      #   'foo'.scan('f')
+      #   'foo'.split('f')
+      #   'foo'.sub('f', 'x')
+      #   'foo'.sub!('f', 'x')
+      class RedundantRegexpArgument < Base
+        extend AutoCorrector
+
+        MSG = 'Use string `%<prefer>s` as argument instead of regexp `%<current>s`.'
+        RESTRICT_ON_SEND = %i[
+          byteindex byterindex gsub gsub! index partition rindex rpartition scan split sub sub!
```
jaynetics, line 42:
```suggestion
          [] []= byteindex byterindex gsub gsub! index partition rindex rpartition scan slice slice! split start_with? sub sub!
```

koic: Good catch! However, `[]` and `[]=` are not applicable because they are false positives when the receiver is a hash. Others have applied :-)

jaynetics: could we check the receiver? `index`, `rindex`, `partition`, `slice`, `slice!` are also Array/Enumerable/Hash methods and could lead to false positives otherwise.

koic: Ah, it's true that `slice` and `slice!` are bad when the receiver is a `Hash` object, so l've removed them. Other methods can usually be kept since an Integer is used as an argument when the receiver is an `Array` and `Enumerable`, I think.

jaynetics: hmm, thats true for `partition`, `slice` and `slice!`, which only take a regexp when called on strings, but not for `index` and `rindex`:

```ruby
[/a/, /b/, /b/, /c/].index /b/  # => 1
[/a/, /b/, /b/, /c/].rindex /b/ # => 2
```

jaynetics:
please note this will still cause false positives for `Array#{index,rindex}` as mentioned [here](https://github.com/rubocop/rubocop/pull/11595#discussion_r1232129711) and still needs to preserve some more STR_SPECIAL_CHARS as mentioned [here](https://github.com/rubocop/rubocop/pull/11595#discussion_r1231268032)

koic:
Yeah, I've made corrections for `Array#{index,rindex}` in https://github.com/rubocop/rubocop/commit/d9b431b51f91acc36d4d66deca73882af65be101. As for `STR_SPECIAL_CHARS`, I'll look into it in conjunction with `Performance/RedundantSplitRegexpArgument`, which it is based on. Thank you.

PR #11602, opened 2023-02-21
[Fix #11596] Make `Style/AccessorGrouping` aware of method call before accessor

### LINKED ISSUE #11596: Style/AccessorGrouping false positives with sorbet-runtime type signatures
## Expected behavior

Sorbet allows type signatures to be [attached](https://sorbet.org/docs/faq#how-should-i-add-types-to-methods-defined-with-attr_reader) to `attr_*` methods:

```ruby
class C
  extend T::Sig

  sig { returns(String) }
  attr_reader :str

  sig { returns(Integer) }
  attr_reader :int
end
```

I expect this code to be accepted by the `Style/AccessorGrouping`, as the conflicting type signatures do not allow the `attr_reader`s to be grouped.

## Actual behavior

The above code triggers `Style/AccessorGrouping` violations:

```
rubocop foo.rb --only Style/AccessorGrouping
Inspecting 1 file
C

Offenses:

foo.rb:8:3: C: [Correctable] Style/AccessorGrouping: Group together all attr_reader attributes.
  attr_reader :str
  ^^^^^^^^^^^^^^^^
foo.rb:11:3: C: [Correctable] Style/AccessorGrouping: Group together all attr_reader attributes.
  attr_reader :int
  ^^^^^^^^^^^^^^^^
```

Create a file `foo.rb` with the example at the top, then invoke `rubocop foo.rb --only Style/AccessorGrouping`

## RuboCop version

```
$ rubocop -V
1.45.1 (using Parser 3.2.0.0, rubocop-ast 1.24.1, running on ruby 3.1.1) [x86_64-darwin21]
```

koic:
Fixes #11596.

This PR makes `Style/AccessorGrouping` aware of method call before accessor.

If there is a method call before the accessor method it is always allowed as it might be intended like Sorbet . Note, it respects user's intentions, not just Sorbet. Because (annotation) comments are already allowed, so (annotation) methods can be generalized as well.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_make_style_accessor_grouping_aware_of_method_call_before_accessor_method.md
+++ b/changelog/change_make_style_accessor_grouping_aware_of_method_call_before_accessor_method.md
@@ -1,0 +1,1 @@
+* [#11596](https://github.com/rubocop/rubocop/issues/11596): Make `Style/AccessorGrouping` aware of method call before accessor. ([@koic][])
--- a/lib/rubocop/cop/style/accessor_grouping.rb
+++ b/lib/rubocop/cop/style/accessor_grouping.rb
@@ -7,8 +7,8 @@ module RuboCop
       # By default it enforces accessors to be placed in grouped declarations,
       # but it can be configured to enforce separating them in multiple declarations.
       #
-      # NOTE: `Sorbet` is not compatible with "grouped" style. Consider "separated" style
-      # or disabling this cop.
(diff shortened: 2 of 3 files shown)


bbatsov:
The cop examples need to be improved, but I'll merge this regardless as I'm about to cut a new release now.

PR #10987, opened 2022-09-03
Opt-in cop compatibility in redundant directives

tdeo:
Add compatibility in `Lint/RedundantCopDisableDirective`, `Lint/RedundantCopEnableDirective` and `Lint/MissingCopEnableDirective` for opt-in cops.

I mean by "opt-in" having a cop which is globally disabled in the configuration, but that we want to enable only on specific parts of a file.

My use-case specifically is the following: we sometimes have large-ish arrays maintaining list of allowed values for an attribute, let's imagine:
```ruby
allowed_prefixes = %w[
  ab_
  ad_
  bc_
  ...
  zz_
]
```
I wrote a custom cop that checks that this array remains alphabetically sorted for human readibility. Obviously, I don't want to enable it globally in the codebase as array ordering is important in many use-cases, but would like to enable it specifically on that part with:
```ruby
# rubocop:enable Pennylane/SortedArray
allowed_prefixes = %w[
  ...
]
# rubocop:disable Pennylane/SortedArray
```

But rubocop currently reports various offenses with this pattern:
```
sample.rb:3:18: W: [Correctable] Lint/RedundantCopEnableDirective: Unnecessary enabling of Pennylane/SortedArray.
# rubocop:enable Pennylane/SortedArray
                 ^^^^^^^^^^^^^^^^^^^^^
sample.rb:7:1: W: Lint/MissingCopEnableDirective: Re-enable Pennylane/SortedArray cop with # rubocop:enable after disabling it.
# rubocop:disable Pennylane/SortedArray
^
sample.rb:7:1: W: [Correctable] Lint/RedundantCopDisableDirective: Unnecessary disabling of Pennylane/SortedArray.
# rubocop:disable Pennylane/SortedArray
```

To be able to implement that, I had to inject both `config` as well as the cop registry into `processed_source`, which turns out to be pretty isolated changes in `RuboCop::Runner` as well as in the `CopHelper` test module to mimic the behaviour.

I also made a small api change to `RuboCop::Registry#enabled` as 2 arguments were already available from instance variables, please let me know if that isn't an acceptable change.

I didn't squash commits yet to be able to discuss implementation strategy. I originally (until https://github.com/rubocop/rubocop/pull/10987/commits/bf4e4b7ca2bb18515f97323471c7884e09c3156a) tried to only inject `config` within `CommentConfig` instead of `ProcessedSource`, but things were getting a bit more out of hand.

-----------------

Before submitting the PR make sure the following are checked:

[...]

PR #11603, opened 2023-02-21
Actually run temporarily enabled cops

tdeo:
A follow-up on https://github.com/rubocop/rubocop/pull/10987, I was trying to use the temporarily cop enablement feature, and realized that it wasn't working as intended. Although rubocop does not complain about unneeded directives in such case, it turns out those temporarily enabled cops weren't reporting any offense.

I think I overlooked 2 things in the original implementation:
- the fact that `Team.mobilize_cops` was going to exclude those cops from the beginning and never invoke them
- the [test](https://github.com/rubocop/rubocop/pull/10987/files#diff-46b3c2cf6f8085fc01a6be8ddf84f83b741c9065ecb04cd0cb3409ee29bf9379R282) that I wrote was checking that the temporarily enabled cop wasn't reported as an unneeded directive, but not that it was actually reporting offenses, this is now fixed.

A few points that I'd like to hear your opinions about:

- I'm not super happy with the implementation in `RuboCop::Team` which instantiates a new `Registry`, but changing the class methods to pass the whole registry to `initialize` looked fishy as there seems to be backward-compatible reasons, happy to hear other suggestions.

- Do you think I should keep the removed test in `team_spec` for disabled cops, as this would now cover a private method (`roundup_relevant_cops`) or trust the other test cases to make sure the implementation is correct?

- I generated a bugfix changelog since the behavior is already in the docs, but since the original implementation actually had no effect, should this be considered a new feature?

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_actually_run_temporarily_enabled_cops.md
+++ b/changelog/fix_actually_run_temporarily_enabled_cops.md
@@ -1,0 +1,1 @@
+* [#11603](https://github.com/rubocop/rubocop/pull/11603): Actually run temporarily enabled cops. ([@tdeo][])
--- a/lib/rubocop/comment_config.rb
+++ b/lib/rubocop/comment_config.rb
@@ -42,6 +42,10 @@ module RuboCop
       disabled_line_ranges.none? { |range| range.include?(line_number) }
     end
 
+    def cop_opted_in?(cop)
+      opt_in_cops.include?(cop.cop_name)
+    end
+
     def cop_disabled_line_ranges
       @cop_disabled_line_ranges ||= analyze
     end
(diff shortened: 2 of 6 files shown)


bbatsov:
> I'm not super happy with the implementation in RuboCop::Team which instantiates a new Registry, but changing the class methods to pass the whole registry to initialize looked fishy as there seems to be backward-compatible reasons, happy to hear other suggestions.

It's not great, but it will do for now. We can always improve this down the road.

> Do you think I should keep the removed test in team_spec for disabled cops, as this would now cover a private method (roundup_relevant_cops) or trust the other test cases to make sure the implementation is correct?

That's fine. I'm not fond of tests for private APIs. 

> I generated a bugfix changelog since the behavior is already in the docs, but since the original implementation actually had no effect, should this be considered a new feature?

It should be a bugfix indeed.

tdeo:
Thanks for the quick review and merge

PR #11619, opened 2023-02-24
[Fix #11615] Fix false negative and positive for `Lint/MissingSuper`

### LINKED ISSUE #11615: `Lint/MissingSuper` does not handle dynamically-generated classes
Currently, `Lint/MissingSuper` doesn't seem to handle classes generated dynamically via `Class.new`:

```ruby
Class.new do
  def initialize; end # no offence ✔️
end

Class.new(String) do
  def initialize; end # no offence ❌(inherited from String -> should call super)
end
```

```ruby
class X
  Class.new do
    def initialize; end # no offence ✔️
  end
end

class X < Y
  # X is inherited, but this class isn't -> we shouldn't have to call super
  Class.new do
    def initialize; end # Lint/MissingSuper: Call super to initialize state of the parent class ❌
  end
end

```

koic:
Fixes https://github.com/rubocop/rubocop/issues/11615.

This PR fixes two bugs reported in #11615:

- Fix a false negative when no `super` call with `Class.new` block.
- Fix a false positive when using `Class.new` without parent class argument.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_negative_for_lint_missing_super.md
+++ b/changelog/fix_a_false_negative_for_lint_missing_super.md
@@ -1,0 +1,1 @@
+* [#11615](https://github.com/rubocop/rubocop/issues/11615): Fix a false negative for `Lint/MissingSuper` when no `super` call with `Class.new` block. ([@koic][])
--- a/changelog/fix_a_false_positive_for_lint_missing_super.md
+++ b/changelog/fix_a_false_positive_for_lint_missing_super.md
@@ -1,0 +1,1 @@
+* [#11615](https://github.com/rubocop/rubocop/issues/11615): Fix a false negative for `Lint/MissingSuper` when using `Class.new` without parent class argument. ([@koic][])
--- a/lib/rubocop/cop/lint/missing_super.rb
+++ b/lib/rubocop/cop/lint/missing_super.rb
@@ -110,8 +110,8 @@ module RuboCop
         end
 
         def inside_class_with_stateful_parent?(node)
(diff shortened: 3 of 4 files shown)


bbatsov:
Looks good to me. Thanks!

simi:
:thinking: this seems to add this warning (not present in 1.46.0). Everything is ok without `header` method. It seems `rubocop:enable Layout/ExtraSpacing` is also broken when `ー` is used (as you can see, it is wrongly reported as `Layout/ExtraSpacing` violation).

```ruby
# frozen_string_literal: true

# base exporter
class BaseExporter
  def initialize
    @c = 'D'
  end
end

module Exporter
  # my exporter
  class MyExporter < BaseExporter
    def initialize # rubocop:disable Lint/MissingSuper
      @a = 'B'
    end

    # rubocop:disable Layout/ExtraSpacing
    def header
      [
        '配ー',         #  A reference
        '配'            #  B name
      ]
    end
    # rubocop:enable Layout/ExtraSpacing
  end
end

```

```
rubocop exporter.rb
Inspecting 1 file
W

Offenses:

exporter.rb:13:20: W: [Correctable] Lint/RedundantCopDisableDirective: Unnecessary disabling of Lint/MissingSuper.
    def initialize # rubocop:disable Lint/MissingSuper
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

koic:
First, I've opened #11654 to fix a new false negative of `Lint/MissingSuper`.

PR #11624, opened 2023-02-25
Fix an incorrect autocorrect for `Layout/ClassStructure` using heredoc inside method

### LINKED ISSUE #11621: Bad `Layout/ClassStructure` autocorrect when heredoc is used
Consider:

```ruby
class C
  def bar
    'bar'
  end

  def self.foo
    <<~EOF
      foo
    EOF
  end
end
```

autocorrecting with `--only Layout/ClassStructure -a` results in `bar` being absorbed into `foo` 😨:

```ruby
class C
  def self.foo
    <<~EOF
      foo
    EOF

  def bar
    'bar'
  end
  end
end
```

Version
```sh
1.46.0 (using Parser 3.2.1.0, rubocop-ast 1.26.0, running on ruby 2.6.9) [x86_64-darwin21]
  - rubocop-minitest 0.28.0
  - rubocop-performance 1.16.0
  - rubocop-rails 2.17.4
```

fatkodima:
Fixes #11621.

### DIFF
--- a/changelog/fix_autocorrect_class_structure_method_with_heredoc.md
+++ b/changelog/fix_autocorrect_class_structure_method_with_heredoc.md
@@ -1,0 +1,1 @@
+* [#11621](https://github.com/rubocop/rubocop/issues/11621): Fix an incorrect autocorrect for `Layout/ClassStructure` using heredoc inside method. ([@fatkodima][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #11627, opened 2023-02-26
[Fix #11626] Fix a false positive for `Style/ZeroLengthPredicate`

### LINKED ISSUE #11626: `Style/ZeroLengthPredicate` breaks file.size.zero?
It seems that the Style/ZeroLengthPredicate cop autocorrect will turn this code:

```ruby
File.new(path).size.zero?
```

into:

```ruby
File.new(path).empty?
```

However, `File#empty?` method does not exist in Ruby (see [File docs](https://ruby-doc.org/core-2.4.0/File.html)).

What could be used instead is `File::empty?`:

```ruby
File.empty?(path)
```

koic:
Fixes #11626.

This PR fixes a false positive for `Style/ZeroLengthPredicate` when using `File.new(path).size.zero?`.
The same goes for the existing `Tempfile` and `StringIO`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_style_zero_length_predicate.md
+++ b/changelog/fix_a_false_positive_for_style_zero_length_predicate.md
@@ -1,0 +1,1 @@
+* [#11626](https://github.com/rubocop/rubocop/issues/11626): Fix a false positive for `Style/ZeroLengthPredicate` when using `File.new(path).size.zero?`. ([@koic][])
--- a/lib/rubocop/cop/style/zero_length_predicate.rb
+++ b/lib/rubocop/cop/style/zero_length_predicate.rb
@@ -4,10 +4,13 @@ module RuboCop
   module Cop
     module Style
       # Checks for numeric comparisons that can be replaced
-      # by a predicate method, such as receiver.length == 0,
-      # receiver.length > 0, receiver.length != 0,
-      # receiver.length < 1 and receiver.size == 0 that can be
-      # replaced by receiver.empty? and !receiver.empty?.
+      # by a predicate method, such as `receiver.length == 0`,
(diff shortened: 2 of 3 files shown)


bbatsov:
Maybe we should also documented all those file-related exceptions, so it's clearer what the cop is aware of. Funny enough I just noticed that files have a `zero?` predicate, but still don't have an `empty?` predicate.

bbatsov:
Actually Ruby 2.4 added `File.empty?` as an alias to `File.zero?`, but the problem remains that those are only class methods. Might also be worth to recommend  (in another cop) using `File.empty?` over `File.zero?` and `file.size.zero?` as it reads much better IMO.

koic:
First, I've updated the documentation.

> Actually Ruby 2.4 added `File.empty?` as an alias to `File.zero?`, but the problem remains that those are only class methods. Might also be worth to recommend  (in another cop) using `File.empty?` over `File.zero?` and `file.size.zero?` as it reads much better IMO. 

Yeah, I made it follow the behavior of the existing `Tempfile` and `StringIO` on this patch. OTOH, I was thinking it might be a worth change :-)
Anyway, this looks like a good idea! It's limited to Ruby 2.4 and higher, so consider it separately.

bbatsov:
>  Anyway, this looks like a good idea! It's limited to Ruby 2.4 and higher, so consider it separately.

Fine by me. I also noticed 2.4 introduced `Dir.empty?` to address code like:

```ruby
Dir.entries("/usr/lib").size == 2 # empty dirs stills have `.` and `..` inside
```

Another new cop idea. 😄

bbatsov:
I've created a couple of tickets (#11628 and #11629), so we won't forgot about the additional cops.

PR #11638, opened 2023-02-28
Fix a false positive for `Lint/UselessAccessModifier` when using same access modifier inside and outside the `included` block

ydah:
This PR is fix a false positive for `Lint/UselessAccessModifier` when using same access modifier inside and outside the `included` block.

If an included block is used within a class, as shown below, I think it should not be a violation even if the same access modifier is used inside and outside the `included` block.

```ruby
class SomeClass
  included do
    private
    def foo; end
  end
  private
  def bar; end
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/useless_access_modifier_spec.rb
+++ b/spec/rubocop/cop/lint/useless_access_modifier_spec.rb
@@ -1081,4 +1081,113 @@ RSpec.describe RuboCop::Cop::Lint::UselessAccessModifier, :config do
       it_behaves_like('nested modules', keyword, modifier)
     end
   end
+
+  context 'when `AllCops/ActiveSupportExtensionsEnabled: true`' do
+    let(:config) do
+      RuboCop::Config.new('AllCops' => { 'ActiveSupportExtensionsEnabled' => true })
+    end
+
+    context 'when using same access modifier inside and outside the included block' do
+      it 'does not register an offense' do
+        expect_no_offenses(<<~RUBY)
+          class SomeClass
+            included do
+              private
+              def foo; end
+            end
+            private
+            def bar; end
+          end
+        RUBY
+      end
+
+      it 'registers an offense when using repeated access modifier inside/outside the included block' do
+        expect_offense(<<~RUBY)
+          class SomeClass
+            included do
+              private
+              private
+              ^^^^^^^ Useless `private` access modifier.
+              def foo; end
+            end
+            private
+            private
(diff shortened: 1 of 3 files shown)


koic:
The `included` block is a `ActiveSupport::Concern`'s method, right? RuboCop core is for vanilla Ruby. If Rails content will be included in the core, it may be better to include `active_support_extensions_enabled?` (`ActiveSupportExtensionsEnabled` config) as a condition.
https://docs.rubocop.org/rubocop/configuration.html#enable-checking-active-support-extensions

`Style/HashExcept` may be a reference for implementation.

ydah:
> The `included` block is a `ActiveSupport::Concern`'s method, right? RuboCop core is for vanilla Ruby. If Rails content will be included in the core, it may be better to include `active_support_extensions_enabled?` (`ActiveSupportExtensionsEnabled` config) as a condition. https://docs.rubocop.org/rubocop/configuration.html#enable-checking-active-support-extensions
> 
> `Style/HashExcept` may be a reference for implementation.

I was not aware of `active_support_extensions_enabled?`. Thanks for letting me know about it 😀
I updated this PR.

spec/rubocop/cop/lint/useless_access_modifier_spec.rb
```diff
@@ -1081,4 +1081,113 @@ def foo
...
+      end
+    end
+  end
+
+  context 'when `AllCops/ActiveSupportExtensionsEnabled: false`' do
+    let(:config) do
+      RuboCop::Config.new('AllCops' => { 'ActiveSupportExtensionsEnabled' => false })
+    end
+
+    context 'when using same access modifier inside and outside the included block' do
```
koic, line 1139:
```suggestion
    context 'when using same access modifier inside and outside the `included` block' do
```

spec/rubocop/cop/lint/useless_access_modifier_spec.rb
```diff
@@ -1081,4 +1081,113 @@ def foo
...
+            included do
+              private
+              def foo; end
+            end
+            def bar; end
+          end
+        RUBY
+      end
+
+      it 'registers an offense when using repeated access modifier inside/outside the included block' do
```
koic, line 1164:
```suggestion
      it 'registers an offense when using repeated access modifier inside/outside the `included` block' do
```

koic:
There may be a better way, but since there are no new parameters added, so there will be no impact on future API compatibility. I'm going to merge this PR. Thank you.

PR #11662, opened 2023-03-04
[Fix #11399] Add new `Style/UnlessMinimizeNegations` cop

### LINKED ISSUE #11399: Cop idea: minimize number of negations in a condition

SparLaimor:
Fixes #11399.
This PR adds new `Style/UnlessMinimizeNegations` cop.

It minimizes the number of negations in an `unless` using De Morgan’s laws.

```ruby
# bad (all terms negated)
do_something unless !x && !y

# good (all terms positive)
do_something if x || y

# bad (2 negations, 1 positive)
do_something unless !x || !y || z

# good (2 positives, 1 negation)
do_something if x && y && !z
```
-----------------

* [x] Added tests.


lib/rubocop/cop/style/unless_minimize_negations.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Minimizes the number of negations in an `unless` using De Morgan’s laws.
```
bbatsov, line 6: Why in `unless` only? Seems to me this should apply to all conditionals.

SparLaimor: More often people avoid `unless` than use it in difficult conditions. 
```
# I don't think the corrected version is much better. Can be seen as a double negation.
do_something if x && !y && !z
do_something unless !x || y || z

# But such a case is not bad.
do_something if !x || !y || !z
do_something unless x && y && z
```

For `if` I would make a parameter (disabled by default). But, this case is related to other cops. `Style/UnlessElse` and `Style/UnlessLogicalOperators` with `forbid_logical_operators`.

bbatsov: I see. I'd be fine to keeping this limited to unless but the rationale needs to be explained clearly in the cop and it's relation to existing cops.

vlad-pisanov: my 2c: I like the idea of conditionally supporting both `if` and `unless`. Perhaps the cop name could be just `Style/MinimizeNegations` with an option to target if/unless/both?

SparLaimor: I'll finalize it as soon as I have time.

Issue #11657, opened 2023-03-03
Excessive memory usage in Rubocop JUnit Reporter starting in 1.46.0 or 1.47.0


broksonic21:
--------

## Expected behavior

Version 1.46.0 and Version 1.47.0 have same memory footprint as 1.45.1 and below and are relatively drop-in.

## Actual behavior

When running with any of 1.45.1, 1.46.0 or 1.47.0 vs our source code base, we are fine when we run this and it takes a minute or so. Max memory is 200MB or so.

```
bundle exec rubocop --fail-level W --display-only-fail-level-offenses
```

Similar on 1.45.1 when running in unit mode - Max memory is 200MB or so.
But memory goes into the gigabytes when we run this on 1.46.0 or 1.47.0 and it takes 30+ minutes if it even finishes:

```
bundle exec rubocop --format junit --fail-level W --display-only-failed --display-only-fail-level-offenses --out results/rubocop.xml
```

For example, I kicked these off at the same time, as they are mid-running:

The top row is Rubocop 1.47.0, and bottom one is Rubocop 1.45.1, both running the junit version of the command . Same code repository, just git-cloned into two different folders

You can see the massive difference 
<img width="781" alt="image" src="https://user-images.githubusercontent.com/24306571/222757798-080137fc-67ab-4568-9536-fab276fd9417.png">

## Steps to reproduce the problem

This is a private source code base, so I can't share out unfortunately.
Never had an issue on rubocop upgrades before,

Upgrade to a rubocop 1.46.0 or 1.47.0 and run this command:
```
bundle exec rubocop --format junit --fail-level W --display-only-failed --display-only-fail-level-offenses --out results/rubocop.xml
```

Note: I also tried upgrading rubocop-rails along with rubocop, same issue, so it's not that.

## RuboCop version

```
1.45.1 (using Parser 3.2.1.0, rubocop-ast 1.27.0, running on ruby 2.7.7) [arm64-darwin21]
  - rubocop-minitest 0.28.0
  - rubocop-performance 1.16.0
  - rubocop-rails 2.17.4
  - rubocop-rake 0.6.0
```

```
1.47.0 (using Parser 3.2.1.0, rubocop-ast 1.27.0, running on ruby 2.7.7) [arm64-darwin21]
  - rubocop-minitest 0.28.0
  - rubocop-performance 1.16.0
  - rubocop-rails 2.17.4
  - rubocop-rake 0.6.0
```

fatkodima:
[...]

PR #11665, opened 2023-03-05
Use cop name to check if cop inside registry is enabled

fatkodima:
Fixes https://github.com/rubocop/rubocop/issues/11657.

Previously, `@for_cop` cache https://github.com/rubocop/rubocop/blob/951507a2cd70518aeca2351b405a301534be195b/lib/rubocop/config.rb#L34-L40
always received only cop names (strings) or cop classes, but after https://github.com/rubocop/rubocop/pull/11603/files#diff-1c2f3a3b82c13964ec1f4f0ce72bb04f429cc9258514611cd6b6d735ec8aea6aR163 it started to receive cop instances, and so this cache became needlessly bigger and bigger.

### DIFF
--- a/lib/rubocop/cop/registry.rb
+++ b/lib/rubocop/cop/registry.rb
@@ -197,7 +197,9 @@ module RuboCop
       def enabled?(cop, config)
         return true if options[:only]&.include?(cop.cop_name)
 
-        cfg = config.for_cop(cop)
+        # We need to use `cop_name` in this case, because `for_cop` uses caching
+        # which expects cop names or cop classes as keys.
+        cfg = config.for_cop(cop.cop_name)
 
         cop_enabled = cfg.fetch('Enabled') == true || enabled_pending_cop?(cfg, config)
 
--- a/lib/rubocop/cop/team.rb
+++ b/lib/rubocop/cop/team.rb
@@ -160,8 +160,8 @@ module RuboCop
       def roundup_relevant_cops(processed_source)
         cops.select do |cop|
           next true if processed_source.comment_config.cop_opted_in?(cop)
-          next false unless @registry.enabled?(cop, @config)
           next false if cop.excluded_file?(processed_source.file_path)
+          next false unless @registry.enabled?(cop, @config)
 
           support_target_ruby_version?(cop) && support_target_rails_version?(cop)
         end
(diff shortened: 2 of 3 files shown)


fatkodima:
So did this actually fully solved an issue?

The excessive memory usage by junit formatter will be fixed by another PR.

broksonic21:
As far as I can tell, @fatkodima , completely solved the issue we've been discussing today.

lib/rubocop/cop/registry.rb
```diff
@@ -197,7 +197,7 @@ def disabled(config)
       def enabled?(cop, config)
         return true if options[:only]&.include?(cop.cop_name)
 
-        cfg = config.for_cop(cop)
+        cfg = config.for_cop(cop.cop_name)
```
koic, line 200: Perhaps it would be useful if a source code comment will be described about the problem avoiding the memory increase.

lib/rubocop/cop/team.rb
```diff
@@ -160,8 +160,8 @@ def investigate_partial(cops, processed_source, offset:, original:)
       def roundup_relevant_cops(processed_source)
         cops.select do |cop|
           next true if processed_source.comment_config.cop_opted_in?(cop)
-          next false unless @registry.enabled?(cop, @config)
           next false if cop.excluded_file?(processed_source.file_path)
+          next false unless @registry.enabled?(cop, @config)
```
koic, line 164: This is also the same if the order of the conditions is meaningful.

fatkodima: The order is not meaningful, I just swapped them because `cop.excluded_file?` will short circuit more often (in my opinion) and is cheaper to compute.

lib/rubocop/cop/registry.rb
fatkodima: Added a comment.

lib/rubocop/cop/registry.rb
```diff
@@ -197,7 +197,9 @@ def disabled(config)
       def enabled?(cop, config)
         return true if options[:only]&.include?(cop.cop_name)
 
-        cfg = config.for_cop(cop)
+        # We need to use cop_name in this case, because for_cop uses caching
```
koic, line 200: Can you tweak it?

```suggestion
        # We need to use `cop_name` in this case, because `for_cop` uses caching
```

fatkodima: Done.

koic:
Nicely done! Thanks!

koic:
@bbatsov This performance regression fix might be worth bugfix releasing early.

bbatsov:
I'll cut a new release later today.

PR #11673, opened 2023-03-07
Fix incorrect Style/HashSyntax autocorrection for assignment methods

gsamokovarov:
In Ruby 3.1 and above, code like:

```ruby
object.attr = {foo: foo}
                    ^^^ Omit the hash value.
pass
```

was incorrectly autocorrected to:

```ruby
object.attr({foo:})
pass
```

The autocorrected code is valid Ruby, but is semantically different. This may and will lead to subtle bugs in user codebases.


koic:
Thanks!

PR #11675, opened 2023-03-07
Style/AccessorGrouping: Fix sibling detection for methods with type sigs

issyl0:
- Over in `Homebrew/brew`, trying to enable this cop after 008506d59f094140636269aa588e32fb69001018 was released, we were seeing the following offenses which were incorrect since we can't combine attrs that follow on from typed ones.

```
  sig { returns(Pathname) }
  attr_reader :cached_location

  attr_reader :cache, :meta, :name, :version
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Group together all `attr_reader` attributes
```

- This commit ensures that sibling accessors are groupable before raising an offense that they should be grouped.
- Also add more tests to ensure the offense detection and subsequent autocorrection still works.
- Follow up to #11650.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/accessor_grouping_spec.rb
+++ b/spec/rubocop/cop/style/accessor_grouping_spec.rb
@@ -141,6 +141,47 @@ RSpec.describe RuboCop::Cop::Style::AccessorGrouping, :config do
       RUBY
     end
 
+    it 'does not register an offense for grouped accessors below a typechecked accessor method' do
+      expect_no_offenses(<<~RUBY)
+        class Foo
+          extend T::Sig
+
+          sig { returns(Integer) }
+          attr_reader :one
+
+          attr_reader :two, :three
+        end
+      RUBY
+    end
+
+    it 'registers an offense for grouped accessors distinct from a typechecked accessor method' do
+      expect_offense(<<~RUBY)
+        class Foo
+          extend T::Sig
+
+          sig { returns(Integer) }
+          attr_reader :one
+
+          attr_reader :two, :three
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/accessor_grouping_spec.rb
```diff
@@ -141,6 +141,47 @@ class Foo
...
+
+          attr_reader :two, :three
+          ^^^^^^^^^^^^^^^^^^^^^^^^ Group together all `attr_reader` attributes.
+
+          attr_reader :four
+          ^^^^^^^^^^^^^^^^^ Group together all `attr_reader` attributes.
+        end
+      RUBY
+
+      expect_correction(<<~RUBY)
```
issyl0, line 173: I ran this test on RuboCop `master` (v1.48.0) and the autocorrect is completely bugged because of the sibling grouping detection. Sorry for not thinking about this sooner! 😳

```
  1) RuboCop::Cop::Style::AccessorGrouping when EnforcedStyle is grouped registers an offense for grouped accessors distinct from a typechecked accessor method
     Failure/Error: expect(new_source).to eq(correction)
     
       expected: "class Foo\n  extend T::Sig\n\n  sig { returns(Integer) }\n  attr_reader :one\n\n  attr_reader :two, :three, :four\nend\n"
            got: "class Foo\n  extend T::Sig\n\n  sig { returns(Integer) }\n  attr_reader :one\nend\n"
     
       (compared using ==)
     
       Diff:
       @@ -3,7 +3,5 @@
        
          sig { returns(Integer) }
          attr_reader :one
       -
       -  attr_reader :two, :three, :four
        end
```

PR #11678, opened 2023-03-08
Improve `Naming/MethodName` example `AllowedPatterns`

sambostock:
None of the code paths lead to scenarios in which a method name could possibly include leading or trailing whitespace, so there is no reason for the documentation or test patterns to include optional whitespace.

Instead, we can show an example of anchoring on both ends, to exactly match the method name.

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/naming/method_name.rb
+++ b/lib/rubocop/cop/naming/method_name.rb
@@ -10,8 +10,8 @@ module RuboCop
       #
       #   Naming/MethodName:
       #     AllowedPatterns:
-      #       - '\A\s*onSelectionBulkChange\s*'
-      #       - '\A\s*onSelectionCleared\s*'
+      #       - '\AonSelectionBulkChange\z'
+      #       - '\AonSelectionCleared\z'
       #
       # Method names matching patterns are always allowed.
       #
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/naming/method_name.rb
```diff
@@ -10,8 +10,8 @@ module Naming
       #
       #   Naming/MethodName:
       #     AllowedPatterns:
-      #       - '\A\s*onSelectionBulkChange\s*'
-      #       - '\A\s*onSelectionCleared\s*'
```
sambostock, line 14: Am I wrong about this? I looked at the code, and I can't see how the method name could ever have a space in it, and be intended to still be allowed.

We pass in

```ruby
def my_method_name
    ^^^^^^^^^^^^^^
    
attr_reader :my_attr_name
             ^^^^^^^^^^^^
end
```

so I don't see how spaces would appear.

But perhaps I missed something, which is why I'm asking.

sambostock: I also checked #7318 where these were introduced, and it looks like these were the patterns right from the start, so it doesn't look like there were added later.

koic: Probably used values from user feedback for the examples. So I have no particular opinion.

changelog/change_improve_naming_method_name_example.md
```diff
@@ -0,0 +1 @@
+* [#11678](https://github.com/rubocop/rubocop/pull/11678): Improve `Naming/MethodName` example `AllowedPatterns`. ([@sambostock][])
```
koic, line 1: Change log entry is not required because no user impact. Can you remove it?

sambostock: Done!

Issue #3018, opened 2016-04-07
Feature Request: Style/EachForSimpleLoop


deivid-rodriguez:
I'd like RuboCop to detect and autocorrect things like:

```ruby
(1..8).each do |_n|
   puts "Do something that does not use the block parameter..."
   # ...
end
```

The preferred style would be

```ruby
8.times do
  puts "Do something that does not use the block parameter..."
  # ...
end 
```

I guess it could be generalized to

```ruby
collection.each do |_item|
  puts "Do something that does not use `_item` or mutate `collection`"
  # ...
end
```

Autocorrected the same way.

Does this sound like a good idea?

Drenmi:
Hi @deivid-rodriguez!

There's one called `Lint/UnusedBlockArgument`. Is this something like what you're looking for? It doesn't have auto-correct, though.

https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/lint/unused_block_argument.rb

deivid-rodriguez:
Hi @Drenmi! No, it's not that. The code I'm talking about has already corrected by the cop you mention.

I'm looking for a smarter cop that detects that you're using things like `(1..10).each { |n| ... }` when you could use `10.times { ... }`. It could reuse the `Lint/UnusedBlockArgument` though since a necessary condition for this replacement to make sense is that the block argument is unused.

alexdowad:
> a necessary condition for this replacement to make sense is that the block argument is unused

Actually, it's not a necessary condition, because `times` also yields numbers from 0 up to N. But many people might find that less readable.

deivid-rodriguez:
Right, I find the `times` version more readable only when the iteration index is unused.

PR #11681, opened 2023-03-08
Fix `Style/EachForSimpleLoop` handling for blocks with a parameter

j1wilmot:
**Problem:**
The `Style/EachForSimpleLoop` cop [documents](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/style/each_for_simple_loop.rb#L10) that "This check only applies if the block takes no parameters", however that is not how it behaves. Not running for blocks with parameters was also the [original intent](https://github.com/rubocop/rubocop/issues/3018#issuecomment-222643692) when this cop was introduced. The issue is that the `each_range_with_zero_origin` matcher matches against a block that takes arguments when it should not.

**Solution:**
- Add tests to ensure an offense is not generated for blocks with params
- Remove the `each_range_with_zero_origin matcher`

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_each_for_simple_loop_block_with_param.md
+++ b/changelog/fix_each_for_simple_loop_block_with_param.md
@@ -1,0 +1,1 @@
+* [#11681](https://github.com/rubocop/rubocop/pull/11681): `Style/EachForSimpleLoop` now does not register offense for blocks with a parameter. ([@j1wilmot][])
--- a/lib/rubocop/cop/style/each_for_simple_loop.rb
+++ b/lib/rubocop/cop/style/each_for_simple_loop.rb
@@ -46,7 +46,7 @@ module RuboCop
         private
 
         def offending?(node)
-          each_range_with_zero_origin?(node) || each_range_without_block_argument?(node)
+          each_range_without_block_argument?(node)
         end
 
         # @!method each_range(node)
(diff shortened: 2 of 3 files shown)


bbatsov:
I'm a bit puzzled by the proposed change, as we could have just as easily updated the docs. As `times` actually passes a parameter to the block it seems to me that use-case is just as valid as the one without passing a param to the block. Am I missing something here?

j1wilmot:
> As times actually passes a parameter to the block it seems to me that use-case is just as valid as the one without passing a param to the block.

There are a couple of reasons why `times` is less ideal than `Range#each` when a block is passed:
1. `times` is more explicit when there is a known number of iterations to run, but if a specific value needs to be iterated then using `Range#each` is more explicit.

```
# Example:

# Using a range makes it explicit and easy to visualize which values will be passed to the block.
(0..4).each { |n| p n }

# Using times makes it implicit which values will be passed to the block. 
# We can reason that this will pass 0, 1, 2, 3 to the block, but it is less visually clear.
4.times { |n| p n }
```


2. Using times as a replacement for calling `each` with a block on a range will only work for ranges that start at 0. So it is more consistent to use `Range#each` when the value being iterated is needed.

```
# Example:
# These are not equivalent.
# Only ranges starting at 0 will have the same block value as `times`.

(1..4).each { |n| p n } # => "1\n2\n3\n4\n" 
4.times { |n| p n } # => "0\n1\n2\n3\n" 
```

bbatsov:
@j1wilmot The changes should also be reflected in the cop's documentation - description & examples.

j1wilmot:
> The changes should also be reflected in the cop's documentation - description & examples.

@bbatsov I think the documentation already reflects the desired behavior, but the implementation previously diverged from the documentation:

"This check only applies if the block takes no parameters."
https://github.com/rubocop/rubocop/pull/11681/files#diff-4ef0af7d2699cabc4a94f5e625da92a87d8ec254c054d94ba8c0b695c9d58980L10

j1wilmot:
Looks like this was fixed in #12601. Closing.

PR #11688, opened 2023-03-10
[Fix #11677] Fix the severity for `Lint/Syntax`

koic:
Fixes #11677.

This PR fixes the severity for `Lint/Syntax`. The `Lint/Syntax` severity must always be `fatal`.

> There is one exception from the general rule above and that is `Lint/Syntax`, a special cop that checks for syntax errors before the other cops are invoked. It cannot be disabled and its severity (`fatal`) cannot be changed in configuration.

https://docs.rubocop.org/rubocop/1.46/configuration.html#severity

In addition to #11666, it also fixes the following issues:

```ruby
def foobar
  t = {
    a: "foo"  # missing "," syntax error
    b: 'bar'
  }
  return t
end
```

## Before

The severity that mult be `fatal` (`F`) is unexpectedly `error` (`E`):

```console
$ bundle exec rubocop
(snip)

Inspecting 1 file
E

Offenses:

example.rb:4:5: E: Lint/Syntax: unexpected token tIDENTIFIER
(Using Ruby 3.2 parser; configure using TargetRubyVersion parameter, under AllCops)
    b: 'bar'
    ^

1 file inspected, 1 offense detected
```

## After

It will be `fatal` (`F`) as expected:

```console
$ bundle exec rubocop
(snip)

Inspecting 1 file
F

Offenses:

example.rb:4:5: F: Lint/Syntax: unexpected token tIDENTIFIER
(Using Ruby 3.2 parser; configure using TargetRubyVersion parameter, under AllCops)
    b: 'bar'
    ^

1 file inspected, 1 offense detected
```

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #11689, opened 2023-03-10
Fix `Lint/Syntax` behavior when `Enabled: false` of `Lint` department

koic:
Here's another fix I found in work on #11688.

Ruby code has a syntax error and .rubocop.yml with `Lint` department is disabled:

```ruby
def foobar
  t = {
    a: "foo"  # missing "," syntax error
    b: 'bar'
  }
  return t
end
```

```yaml
# .rubocop.yml
Lint:
  Enabled: false
```

But `Lint/Syntax` must not be disabled:

> There is one exception from the general rule above and that is `Lint/Syntax`, a special cop that checks for syntax errors before the other cops are invoked. It cannot be disabled and its severity (`fatal`) cannot be changed in configuration.

https://docs.rubocop.org/rubocop/1.46/configuration.html#severity

## Before

The syntax is invalid but an offense is not registerd:

```console
$ bundle exec rubocop
(snip)

Inspecting 1 file
.

```

## After

`Lint/Syntax` detects invalid syntax as expected:

```console
$ bundle exec rubocop
(snip)

Offenses:

example.rb:4:5: F: Lint/Syntax: unexpected token tIDENTIFIER
(Using Ruby 3.2 parser; configure using TargetRubyVersion parameter, under AllCops)
    b: 'bar'
    ^

```

The `Lint/Syntax` behavior for `DisabledByDefault: true` configuration is also fixed.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_lint_syntax_when_enabled_false.md
+++ b/changelog/fix_lint_syntax_when_enabled_false.md
@@ -1,0 +1,1 @@
+* [#11686](https://github.com/rubocop/rubocop/pull/11686): Fix `Lint/Syntax` behavior when `Enabled: false` of `Lint` department. ([@koic][])
--- a/lib/rubocop/config.rb
+++ b/lib/rubocop/config.rb
@@ -304,8 +304,8 @@ module RuboCop
     end
 
     def enable_cop?(qualified_cop_name, cop_options)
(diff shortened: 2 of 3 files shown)


bbatsov:
Nice catch!

PR #11694, opened 2023-03-13
[Fix #11122] Add new `Style/RedundantLineContinuations` cop

### LINKED ISSUE #11122: Cop idea: Check for unnecessary line continuations
I sometimes see code like this
```rb
foo \
  .bar
```

or this
```rb
foo. \
  bar
```

where the backslash is completely unnecessary. Can we create a cop that can detect (and remove) those backslashes?

ydah:
Fix: #11122

-----------------

* [x] Added tests.


koic:
This cop should be able to detect various redundat `\` cases such as:

```ruby
def foo \
end

class Foo \
end

foo do \
end
```

and many cases.

lib/rubocop/cop/style/redundant_line_continuations.rb
```diff
@@ -0,0 +1,89 @@
...
+      #   foo&.
+      #     bar
+      #   [foo,
+      #     bar]
+      #   {foo:
+      #     bar}
+      #   foo(bar,
+      #     baz)
+      #   foo('bar' \
+      #     'baz')
```
koic, line 39: These examples looks complicated. Can you update to simplify and clarify?

lib/rubocop/cop/style/redundant_line_continuations.rb
```diff
@@ -0,0 +1,89 @@
...
+          register_offense(arguments_range(node)) if node.arguments?
+        end
+        alias on_csend on_send
+
+        def on_array(node)
+          return unless redundant_line_continuation?(node)
+
+          register_offense(node.source_range)
+        end
+        alias on_hash on_array
```
koic, line 61: Perhaps a more generic approach should be used rather than `on_send` and `on_array`.

lib/rubocop/cop/style/redundant_line_continuations.rb
```diff
@@ -0,0 +1,89 @@
...
+      #   [foo,
+      #     bar]
+      #   {foo:
+      #     bar}
+      #   foo(bar,
+      #     baz)
+      #   foo('bar' \
+      #     'baz')
+      #
+      class RedundantLineContinuations < Base
```
bbatsov, line 41: I'd change the name to be in singular form - "RedundantLineContinuation". I think that's how we typically name cops.

lib/rubocop/cop/style/redundant_line_continuations.rb
```diff
@@ -0,0 +1,89 @@
...
+        include MatchRange
+        include RangeHelp
+        extend AutoCorrector
+
+        MSG = 'Avoid redundant line continuations.'
```
bbatsov, line 46: Probably the message here can be just `Redundant line continuation.".

lib/rubocop/cop/style/redundant_line_continuations.rb
ydah: Indeed it is. I changed it to `RedundantLineContinuation`.

lib/rubocop/cop/style/redundant_line_continuations.rb
ydah: Examples are organized.

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,99 @@
...
+        def on_array(node)
+          return unless redundant_line_continuation?(node)
+
+          register_offense(node.source_range)
+        end
+        alias on_hash on_array
+        alias on_def on_array
+        alias on_defs on_array
+        alias on_class on_array
+        alias on_block on_array
```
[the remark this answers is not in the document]
ydah: Sorry, I could not understand the [more generic approach](https://github.com/rubocop/rubocop/pull/11694#discussion_r1134204555) you're referring to. So I was about to ask a question. However, I apologize for the delay in asking the question and getting confirmation.

bbatsov:
@ydah I see a lot of CI failures.

ydah:
Thank you so much for feedback. I updated this PR. `on_new_investigation` and added test cases and comments accordingly.

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,130 @@
...
+      #   does not result in a syntax error.
+      #
+      #   [source,ruby]
+      #   ----
+      #   1 \
+      #     + 2 \
+      #       - 3
+      #   some_method \
+      #     (argument)
+      #   ----
```
koic, line 26: This cop should be implemented so that false positives do not occur as a safe cop. So this case should be allowed.

ydah: I see, I have updated this PR to avoid false positives for those cases.

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,147 @@
...
+module RuboCop
+  module Cop
+    module Style
+      # Check for redundant line continuation.
+      #
+      # This cop is marked as redundant if removing the backslash
+      # does not result in a syntax error.
+      # However, a backslash at the end of a comment or
+      # for string concatenation is not redundant and is not considered an offense.
+      # And if removing the backslash changes the behavior, it will not be considered an offense.
```
koic, line 12: It's obvious, so I think it's unnecessary.

```suggestion
```

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,147 @@
...
+      #       - 3
+      #
+      #   # also good - backslash with newline between the method name and its arguments,
+      #   # it is not redundant.
+      #   some_method \
+      #     (argument)
+      #
+      class RedundantLineContinuation < Base
+        include MatchRange
+        include RangeHelp
```
koic, line 70: Can you remove the redundant module mix-in?

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,147 @@
...
+
+            add_offense(range) do |corrector|
+              corrector.remove_leading(range, 1)
+            end
+          end
+        end
+
+        private
+
+        def not_redundant?(range)
```
koic, line 91: `not_` may be confusing. Can you make the logic match a name like `require_line_continuation?`?

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,147 @@
...
+
+        def ends_with_backslash_without_comment?(source_line)
+          source_line.gsub(/#.+/, '').end_with?('\\')
+        end
+
+        def string_concatenation?(source_line)
+          /["']\s*\\\z/.match?(source_line)
+        end
+
+        def valid_syntax?(range)
```
koic, line 105:
```suggestion
        def redundant_line_continuation?(range)
```

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,147 @@
...
+          return true unless (node = find_node_for_line(range.line))
+          return false if argument_newline?(node)
+
+          ProcessedSource.new(node.source.gsub(/\\\n/, "\n"),
+                              processed_source.ruby_version).valid_syntax?
```
koic, line 110:
```suggestion
          parse(node.source.gsub(/\\\n/, "\n")).valid_syntax?
```

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,147 @@
...
+        end
+
+        def argument_newline?(node)
+          return unless node.send_type?
+
+          /#{node.method_name}\s*\\\n/.match?(node.source)
```
koic, line 116: AST should be preferred over regular expressions. How is it?

```suggestion
          return if !node.send_type? || node.arguments.empty?

          node.loc.selector.line != node.first_argument.loc.line
```

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,147 @@
...
+        end
+
+        def find_node_for_line(line)
+          processed_source.ast.each_node do |node|
+            return node if same_line?(node, line)
+          end
+
+          nil
```
koic, line 124:
```suggestion
```

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,147 @@
...
+        end
+
+        def same_line?(node, line)
+          return unless (source_range = node.source_range)
+
+          if STR_TYPES.include?(node.type)
```
koic, line 130:
```suggestion
          if node.is_a?(AST::StrNode)
```

spec/rubocop/cop/style/redundant_line_continuation_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+        ? bar
+          : baz
+      1 \
+        + 2 \
+          - 3
+      some_method \
+        (argument)
+      some_method \
+        argument
+    RUBY
```
koic, line 169: Too many test cases may be written. Can you please split it for line continuations are not redundant?

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,142 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Check for redundant line continuation.
+      #
+      # This cop is marked as redundant if removing the backslash
```
bbatsov, line 8: This cop marks a line continuation as redundant if...

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -0,0 +1,142 @@
...
+      #   # it is not redundant.
+      #   some_method \
+      #     (argument)
+      #
+      class RedundantLineContinuation < Base
+        include MatchRange
+        extend AutoCorrector
+
+        MSG = 'Redundant line continuation.'
+        STR_TYPES = %i[str dstr xstr].freeze
```
koic, line 72: This constant can be removed.

```suggestion
```

ydah:
Thank you for lots. I updated this PR.

bbatsov:
Seems we're good to go here. Thanks!

nobuyo:
In looking at the release notes for v1.49, I noticed that the changelog entry on this was missing 👀

gsamokovarov:
Here is an example of a false-positive case:


```ruby
merged_document.public_url.should eq \
  "#{Settings.server.base_uri}/file_download?filename=graphql_receipts_merge%2Ftest1234"
```

Gets autocorrect to:

```ruby
merged_document.public_url.should eq
"#{Settings.server.base_uri}/file_download?filename=graphql_receipts_merge%2Ftest1234"
```

This is interpreted as `eq()` by Ruby itself and fails my test.

```ruby
merged_document.public_url.should eq()
"#{Settings.server.base_uri}/file_download?filename=graphql_receipts_merge%2Ftest1234"
```

I have to put parentheses to preserve the correct behaviour now, but I actually preferred omitting them in that case.

koic:
@gsamokovarov Thanks! I've opened #11752 to fix the false positive.

frodsan:
I think there is a false positive:

```ruby
def ==(other)
        foo == other.foo \
          && bar == other.bar \
          &&  baz == other.baz
end
```

I get the following after autocorrect:

```
F: Lint/Syntax: unexpected token tAMPER
(Using Ruby 3.2 parser; configure using TargetRubyVersion parameter, under AllCops)
        && bar == other.bar
```

ydah:
@frodsan Thank you so much. I've opened https://github.com/rubocop/rubocop/pull/11754 to fix the false positive.

PR #11705, opened 2023-03-16
[Fix #11704] Fix a false positive for `Lint/UselessMethodDefinition`

### LINKED ISSUE #11704: Lint/UselessMethodDefinition false positive on methods transformed by class method
## Expected behavior

The following code should not get picked up / auto-corrected by the `UselessMethodDefinition` cop:
```ruby
memoize def my_method
  super
end
```
(illustrated here in conjunction with [Memery](https://github.com/tycooon/memery), but this applies more generally)

## Actual behavior

The code gets auto-corrected to the following invalid code:

```ruby
memoize
```

koic:
Fixes #11704.

This PR fixes a false positive for `Lint/UselessMethodDefinition` when method definition with non access modifier containing only `super` call.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/useless_method_definition.rb
+++ b/lib/rubocop/cop/lint/useless_method_definition.rb
@@ -41,14 +41,22 @@ module RuboCop
         MSG = 'Useless method definition detected.'
 
         def on_def(node)
-          return if use_rest_or_optional_args?(node)
+          return if method_definition_with_modifier?(node) || use_rest_or_optional_args?(node)
           return unless delegating?(node.body, node)
 
-          add_offense(node) { |corrector| corrector.remove(node) }
+          add_offense(node) do |corrector|
+            range = node.parent&.send_type? ? node.parent : node
+
+            corrector.remove(range)
+          end
         end
         alias on_defs on_def
 
         private
+
+        def method_definition_with_modifier?(node)
+          node.parent&.send_type? && !node.parent&.non_bare_access_modifier?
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/useless_method_definition.rb
```diff
@@ -41,15 +41,23 @@ class UselessMethodDefinition < Base
...
+            range = node.parent&.send_type? ? node.parent : node
+
+            corrector.remove(range)
+          end
         end
         alias on_defs on_def
 
         private
 
+        def modifiered_method_definition?(node)
```
bbatsov, line 57: I'd go with "method_definition_with_modifier?" here.

spec/rubocop/cop/lint/useless_method_definition_spec.rb
```diff
@@ -170,6 +214,14 @@ def method(x: 1)
     RUBY
   end
 
+  it 'does not register an offense when method definition with non access modifier containing only `super` call' do
```
bbatsov, line 217: Technically speaking if it's not an access modifier we should probably call it should else - e.g. "generic method macro" or whatever. I'm not sure what's the best terminology here.

koic: I'm not sure the best terminology, but I think "generic method macro" is definitely a better name. Thank you for the suggestion! I've updated this PR.

PR #11712, opened 2023-03-18
Fix a crash in `Lint/EmptyConditionalBody`

gsamokovarov:
```ruby
var =
  unless object.action value:, other:
    condition || other_condition # This is the value of `other:`, like so:
                                 # `other: condition || other_condition`
  end
```

The code above is tricky, because we may think `condition || other_condition` is the body of the condition, however the code is a valid case for the `Lint/EmptyConditionalBody` rule. It is equivalent to:

```ruby
var =
  unless object.action value:, other: condition || other_condition
  end
```

However, the variable assignment is tricking `CommentsHelp#find_end_line` method into calling an `#end` method of `Parser::Source::Map::Variable` object. `Parser::Source::Map::Variable` **does not respond** to it. Other `Parser::Source::Map` subclasses also do not respond to `#end` methods so we need to check for its existing. `Parser::Source::Map::RescueBody` can be a parent node too.

### DIFF
--- a/changelog/fix_fix_a_crash_in_lint_empty_conditional_body.md
+++ b/changelog/fix_fix_a_crash_in_lint_empty_conditional_body.md
@@ -1,0 +1,1 @@
+* [#11712](https://github.com/rubocop/rubocop/pull/11712): Fix a crash in `Lint/EmptyConditionalBody`. ([@gsamokovarov][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Looks good. I really hope we already has some cop that advises against splitting maps in such a manner over multiple lines, as such code looks horrible.

PR #11724, opened 2023-03-21
[Fix #11723] Fix a false positive for `Style/IfUnlessModifier`

### LINKED ISSUE #11723: False positive for `IfUnlessModifier` with pattern matching
If an `if` condition contains a pattern matching with a variable assignment, rubocop may erroneously suggest changing this code to use an `if` modifier, but the changed code does not work.

--------

## Expected behavior

No offences, no autocorrection

## Actual behavior

```
For /home/meh/Projects/sandbox: Default configuration from /home/meh/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/rubocop-1.48.1/config/default.yml
Use parallel by default.
Inspecting 1 file
Scanning /home/meh/Projects/sandbox/rubocop_pattern_match.rb
C

Offenses:

rubocop_pattern_match.rb:3:1: C: [Correctable] Style/IfUnlessModifier: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.
if [1] in [x]
^^

Finished in 0.1424515729995619 seconds
```

Code after autocorrection:
```ruby
# frozen_string_literal: true

puts x if [1] in [x]
```
This code does not work because the variables declared to the right of if are not visible to the code to the left of if. As a result, we get an error:
```
undefined local variable or method `x' for main:Object (NameError)

puts x if [1] in [x]
     ^
```

```ruby
# frozen_string_literal: true

if [1] in [x]
  puts x
end
```

## RuboCop version

```
$ rubocop -V
1.48.1 (using Parser 3.2.1.0, rubocop-ast 1.27.0, running on ruby 3.2.1) [x86_64-linux]
```

koic:
Fixes #11723.

This PR fixes a false positive for `Style/IfUnlessModifier` when using one-line pattern matching as a `if` condition.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_style_if_unless_modifier.md
+++ b/changelog/fix_a_false_positive_for_style_if_unless_modifier.md
@@ -1,0 +1,1 @@
+* [#11723](https://github.com/rubocop/rubocop/issues/11723): Fix a false positive for `Style/IfUnlessModifier` when using one-line pattern matching as a `if` condition. ([@koic][])
--- a/lib/rubocop/cop/style/if_unless_modifier.rb
+++ b/lib/rubocop/cop/style/if_unless_modifier.rb
@@ -11,6 +11,18 @@ module RuboCop
       # cop. The tab size is configured in the `IndentationWidth` of the
       # `Layout/IndentationStyle` cop.
       #
+      # One-line pattern matching is always allowed. To ensure that there are few cases
+      # where the match variable is not used, and to prevent oversights. The variable `x`
+      # becomes undefined and raises `NameError` when the following example is changed to
+      # the modifier form:
+      #
+      # [source,ruby]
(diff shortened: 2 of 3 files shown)


bbatsov:
I'm wondering about the cases when the match variable is not used in the body, as I guess those should be extremely uncommon (and often some oversight). Perhaps it's better to treat all instances of one-line pattern matching the same?

koic:
To be honest, I was also worried about the same thing. OTOH, I thought the cop's intention was better respected by being in modifier form when possible. However, you're right. It seems plausible to assume that match variable is rarely not used. And it would be more valuable to always allow one-liner matching to prevent it from being oversights. I've updated this PR.

bbatsov:
Thanks!

PR #11728, opened 2023-03-22
[Fix #11696] Add new `Style/DataInheritance` cop

### LINKED ISSUE #11696: DataInheritance cop similar to StructInheritance

https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/style/struct_inheritance.rb exists because using the bad form creates a useless anonymous class ([as documented](https://ruby-doc.org/core-2.7.0/Struct.html#:~:text=Subclassing%20an%20anonymous%20struct%20creates%20an%20extra%20anonymous%20class%20that%20will%20never%20be%20used.)):
```
irb(main):001:0> class MyStruct < Struct.new(:foo); end
=> nil
irb(main):002:0> MyStruct.ancestors
=> [MyStruct, #<Class:0x0000000105e95790>, Struct, Enumerable, Object, PP::ObjectMixin, Kernel, BasicObject]
```

[Ruby 3.2's Data class](https://ruby-doc.org/3.2.1/Data.html) has the same issue, so it makes sense to have that cop (or a new cop) detect the same pattern:

```
irb(main):003:0> class Point < Data.define(:x, :y); end
=> nil
irb(main):004:0> Point.ancestors
=> [Point, #<Class:0x0000000105e9ef70>, Data, Object, PP::ObjectMixin, Kernel, BasicObject]
```

```ruby
# bad
class Point < Data.define(:x, :y); end

# good
Point = Data.define(:x, :y)
```

N/A

## Additional context

N/A

ktopolski:
Fixes #11696. 
This PR adds new cop: `Style/DataInheritance`. It's based on the `Style/StructInheritance` cop.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/data_inheritance.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for inheritance from Data.define.
```
bbatsov, line 6: I'd also mention here what's the problem we're trying to solve (avoid the creation of the anonymous parent class).

bbatsov:
It would also be nice to add a matching guideline in the Ruby style guide and mention it in the cop's metadata.

bbatsov:
I'm referring to https://github.com/rubocop/ruby-style-guide#dont-extend-structnew

koic:
`Data.define` was added in Ruby 3.2. Using `minimum_target_ruby_version 3.2` is more robust.

lib/rubocop/cop/style/data_inheritance.rb
ktopolski: Added couple words, thanks for the comment! I was considering adding outputs of `.ancestors` and whatnot, but was afraid it will be too verbose then. WDYT about current form?

ktopolski:
Thanks for the comments @koic and @bbatsov ! I've added the `minimum_target_ruby_version 3.2` (nice catch) and opened [PR to style guide](https://github.com/rubocop/ruby-style-guide/pull/917). Looking forward to the re-review :)

spec/rubocop/cop/style/data_inheritance_spec.rb
```diff
@@ -0,0 +1,132 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Style::DataInheritance, :config do
+  context 'Ruby >= 3.2', :ruby32 do
+    it 'registers an offense when extending instance of Data' do
```
koic, line 5: Can you replace the entire test description with `Data.define`?

```suggestion
    it 'registers an offense when extending instance of `Data.define`' do
```

spec/rubocop/cop/style/data_inheritance_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    it 'accepts assignment to block form of Data.define' do
+      expect_no_offenses(<<~RUBY)
+        Person = Data.define(:first_name, :last_name) do
+          def age
+            42
+          end
+        end
+      RUBY
+    end
+  end
```
koic, line 131: Can you add a test for `context 'Ruby <= 3.1', :ruby31 do` that does not register an offense?

lib/rubocop/cop/style/data_inheritance.rb
```diff
@@ -0,0 +1,75 @@
...
+          add_offense(node.parent_class.source_range) do |corrector|
+            corrector.remove(range_with_surrounding_space(node.loc.keyword, newlines: false))
+            corrector.replace(node.loc.operator, '=')
+
+            correct_parent(node.parent_class, corrector)
+          end
+        end
+
+        # @!method data_constructor?(node)
+        def_node_matcher :data_constructor?, <<~PATTERN
```
koic, line 48:
```suggestion
        def_node_matcher :data_define?, <<~PATTERN
```

lib/rubocop/cop/style/data_inheritance.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for inheritance from Data.define to avoid creating the anonymous parent class.
```
koic, line 6:
```suggestion
      # Checks for inheritance from `Data.define` to avoid creating the anonymous parent class.
```

ktopolski:
Thanks @koic for bunch of useful suggestions, fixed them all 👍 

@bbatsov - sorry, didn't have much time on my hands during the week 😅 hopefully I'm still in time for the next release!

bbatsov (approved):

koic:
Thanks!

PR #11742, opened 2023-03-29
Fix error when trying to display error message

composerinteralia:
When RuboCop errors, it prints a message that includes the issue tracker URL. It pulls this URL from the gem spec metadata.

Relying on the gem spec can cause problems in bundler standalone mode, where everything is set up by the standalone file and `Gem.loaded_specs` is not populated with the bundled gems.

```
$ bundle init
$ bundle add rubocop --skip-install
$ bundle install --standalone
$ ruby -r ./bundle/bundler/setup.rb -e 'require "rubocop"; Gem.loaded_specs["bundler"].metadata['bug_tracker_uri']'
undefined method `metadata' for nil:NilClass (NoMethodError)
```

This means that if RuboCop errors, instead of a helpful error message we get a cryptic error that masks the underlying problem. We've seen this a number of times at GitHub (although mostly when developing our own internal cops and not from bugs in RuboCop itself).

This commit makes the error reporting more resilient by skipping the issue tracker URL when the rubocop spec is unavailable. Leaving out the URL in this case seems OK to me, since it's basically falling back to the error message before https://github.com/rubocop/rubocop/commit/ef74a1886aa9174aa8c7fcb5441e00f3843fb7e8. Hard-coding the URL is another option, although there's some risk of it falling out of sync some day.

I didn't find any existing test coverage for this code (perhaps because it only gets run if something else is broken), so I haven't added any additional coverage here.

-----------------

* [ ] Added tests. (See note above. Open to suggestions!)

### DIFF
--- a/changelog/fix_error_handling_in_bundler_standalone_mode.md
+++ b/changelog/fix_error_handling_in_bundler_standalone_mode.md
@@ -1,0 +1,1 @@
+* [#11742](https://github.com/rubocop/rubocop/pull/11742): Fix error handling in bundler standalone mode. ([@composerinteralia][])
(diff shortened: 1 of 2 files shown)


koic:
Thanks!

PR #11745, opened 2023-03-30
Add new `Style/YAMLFileRead` cop

koic:
This PR adds new `Style/YAMLFileRead` cop.

It checks for the use of `YAML.load`, `YAML.safe_load`, and `YAML.parse` with `File.read` arguemnt.

NOTE: `YAML.safe_load_file` was introduced in Ruby 3.0.

```ruby
# bad
YAML.load(File.read(path))
YAML.parse(File.read(path))

# good
YAML.load_file(path)
YAML.parse_file(path)

# bad
YAML.safe_load(File.read(path)) # Ruby 3.0 and newer

# good
YAML.safe_load_file(path)       # Ruby 3.0 and newer
```

These good cases use intentionally shorthand methods.

-----------------

* [x] Added tests.


bbatsov:
Are there any practical advantages of this other than it's more compact?

koic:
TBH, I'm not sure about anything other than compactness, but for example, replacement with these appropriate standard APIs can reduce `Metrics/AbcSize` (i.e. complexity). RuboCop itself will also benefit from the following in the future:

```diff
diff --git a/lib/rubocop/config_obsoletion.rb b/lib/rubocop/config_obsoletion.rb
index d3ae3950a..3c1e45a04 100644
--- a/lib/rubocop/config_obsoletion.rb
+++ b/lib/rubocop/config_obsoletion.rb
@@ -47,9 +47,9 @@ module RuboCop

     # Default rules for obsoletions are in config/obsoletion.yml
     # Additional rules files can be added with `RuboCop::ConfigObsoletion.files << filename`
-    def load_rules # rubocop:disable Metrics/AbcSize
+    def load_rules
       rules = self.class.files.each_with_object({}) do |filename, hash|
-        hash.merge!(YAML.safe_load(File.read(filename))) do |_key, first, second|
+        hash.merge!(YAML.safe_load_file(filename)) do |_key, first, second|
           case first
           when Hash
             first.merge(second)
```

lib/rubocop/cop/style/yaml_file_read.rb
```diff
@@ -0,0 +1,66 @@
...
+    module Style
+      # Checks for the use of `YAML.load`, `YAML.safe_load`, and `YAML.parse` with
+      # `File.read` argument.
+      #
+      # NOTE: `YAML.safe_load_file` was introduced in Ruby 3.0.
+      #
+      # @example
+      #
+      #   # bad
+      #   YAML.load(File.read(path))
```
bbatsov, line 14: Btw, I'm guessing we should be explicit and mention that it's fine to pass a file data explicitly if you already happen to have it. (e.g. a good example)

I'm also wondering if this not a target for a broader cop - I assume there are more methods that work with a file object, file data or a file path in ruby.

bbatsov:
I see. I assume the shorter version might also be more memory efficient if you don't need to load all the contents in memory. I was mostly wondering if we can improve the cop rationale.

I'm also wondering if that's not a broader problem and we can expand the cop to cover more similar API.

koic:
My apologies for the delayed response. I've been considering similar API cases where this cop might want to shorten names, such as with `YAML.load_file`, but I haven't come up with any. Therefore, I think it's ready for release as is. Thank you for your review and merge.

PR #11748, opened 2023-03-31
[Fix #10766] Use the path given by `--cache-root` to be the parent for  `rubocop_cache` dir like other ways to specify it

### LINKED ISSUE #10766: Odd inconsistency with `--cache-root`
## Expected behavior

Configuring `--cache-root` would behave exactly like setting `RUBOCOP_CACHE_ROOT` or `AllCops: CacheRootDirectory`, modulo precedence.

## Actual behavior

The `--cache-root` flag specifies the exact cache folder. The other two options end up creating a nested `rubocop_cache` directory. I think I see from the code why this is happening but it is perplexing as an end user (and messed up my CI cache at first).

Specify a directory to `--cache-root` and separately either of the other options and observe that the directory structure it creates is different.

## RuboCop version

```
1.31.0 (using Parser 3.1.2.0, rubocop-ast 1.18.0, running on ruby 2.7.6 x86_64-linux)
  - rubocop-performance 1.14.2
  - rubocop-rspec 2.11.1
```

I'm not sure if this should be changed to match or if the difference should just be documented but it is very surprising as-is. I actually prefer the `--cache-root` behavior (so i don't have an extra nesting of directories) but the main issue here is that it is confusing.

nobuyo:
Fixes #10766. 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/result_cache.rb
+++ b/lib/rubocop/result_cache.rb
@@ -86,7 +86,7 @@ module RuboCop
     attr :path
 
     def initialize(file, team, options, config_store, cache_root = nil)
-      cache_root ||= options[:cache_root]
+      cache_root ||= File.join(options[:cache_root], 'rubocop_cache') if options[:cache_root]
       cache_root ||= ResultCache.cache_root(config_store)
       @allow_symlinks_in_cache_location =
         ResultCache.allow_symlinks_in_cache_location?(config_store)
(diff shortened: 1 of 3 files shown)


lib/rubocop/result_cache.rb
```diff
@@ -91,6 +91,7 @@ def initialize(file, team, options, config_store, cache_root = nil)
       @allow_symlinks_in_cache_location =
         ResultCache.allow_symlinks_in_cache_location?(config_store)
       @path = File.join(cache_root,
+                        'rubocop_cache',
```
koic, line 94: When `ResultCache.cache_root(config_store)` (L90) is assigned instead of `options[:cache_root]` (L89) seems to create an unexpected nested `rubocop_cache/rubocop_cache` directory. Can you add a test to prevent the unexpected behavior and fix the implementation?

nobuyo: Sorry, you're right. I've fixed the implementation.

koic:
Thanks!

PR #11749, opened 2023-04-01
Add new `Lint/DuplicateMatchPattern` cop

koic:
This PR adds new `Lint/DuplicateMatchPattern` cop. It checks that there are no repeated patterns used in `in` keywords.

```ruby
# bad
case x
in 'first'
  do_something
in 'first'
  do_something_else
end

# good
case x
in 'first'
  do_something
in 'second'
  do_something_else
end
```

I was wondering whether to extend existing `Lint/DuplicateCaseCondition`, but it is a different inspection target from the concept named "`case` condition". So, I implemented it as a new `Lint/DuplicateMatchPattern` for "match pattern".

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/duplicate_in_pattern.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Checks that there are no repeated patterns used in `in` keywords.
+      #
+      # @example
```
bbatsov, line 8: It'd be nice to show examples with more complex patterns, so it's clear that this cop is pretty smart about them.

bbatsov:
I wonder if `DuplicateMatchPattern` might not be a better name for the cop (or at least a more descriptive name).

koic:
Thank you for the suggestions. `DuplicateMatchPattern` looks like a better name and I've also updated the examples.

bbatsov:
Looks great. Thanks!

PR #11752, opened 2023-04-03
Fix a false positive for `Style/RedundantLineContinuation`

koic:
Follow up https://github.com/rubocop/rubocop/pull/11694#issuecomment-1493936758

This PR fixes a false positive for `Style/RedundantLineContinuation` when using line concatenation and calling a method without parentheses.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/redundant_line_continuation_spec.rb
+++ b/spec/rubocop/cop/style/redundant_line_continuation_spec.rb
@@ -150,6 +150,20 @@ RSpec.describe RuboCop::Cop::Style::RedundantLineContinuation, :config do
     RUBY
   end
 
+  it 'does not register an offense when using line concatenation and calling a method without parentheses' do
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/redundant_line_continuation_spec.rb
```diff
@@ -150,6 +150,13 @@ def self.foo(bar,#{trailing_whitespace}
     RUBY
   end
 
+  it 'does not register an offense when using string concatenation and calling a method without parentheses' do
```
bbatsov, line 153: string concatenation -> line continuation (or you can just remove this, as it doesn't seem very relevant)

koic: Sure! Indeed this is "line continuation" 💦 I've updated it. (It can be removed, but it remains for the testing information)

PR #11753, opened 2023-04-03
Fix a method name for `Style/RedundantLineContinuation`

ydah:
This PR is a follow-up of #11694.
Wrong method name was used in the cop.

-----------------

* [-] Added tests.


koic:
I'm going to close this PR. Unfortunately this request is not an essential solution. I will try to solve that.
https://github.com/rubocop/rubocop/pull/11694#issuecomment-1494003899

ydah:
I think this will probably resolve the issue. We need to check whether the syntax is valid by including the parent node if there is one.
- https://github.com/rubocop/rubocop/pull/11754

PR #11759, opened 2023-04-04
[Fix #11758] Fix a false positive for `Style/RedundantLineContinuation`

### LINKED ISSUE #11758: Style/RedundantLineContinuation should not look inside strings
When the trailing backslash is inside a string literal, I think it needs to be disregarded by the Style/RedundantLineContinuation cop.

--------

## Expected behavior

I expect these two spec examples to pass in spec/rubocop/cop/style/redundant_line_continuation_spec.rb:

```rb
expect_no_offenses(<<~'RUBY')
  "foo \
    bar"
RUBY

expect_no_offenses(<<~'RUBY')
  'foo \
    bar'
RUBY
```

## Actual behavior

The spec examples fail, suggesting to remove the trailing backslash.

See above.

## RuboCop version

7c35167233e567cd0bad3879974de787b3c372bd

koic:
Fixes #11758.

This PR fixes a false positive for `Style/RedundantLineContinuation` when line continuations for string.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_line_continuation.rb
+++ b/lib/rubocop/cop/style/redundant_line_continuation.rb
@@ -69,6 +69,7 @@ module RuboCop
         extend AutoCorrector
 
         MSG = 'Redundant line continuation.'
+        ALLOWED_STRING_TOKENS = %i[tSTRING tSTRING_CONTENT].freeze
 
         def on_new_investigation
           return unless processed_source.ast
@@ -88,7 +89,8 @@ module RuboCop
         def require_line_continuation?(range)
           !ends_with_backslash_without_comment?(range.source_line) ||
             string_concatenation?(range.source_line) ||
-            start_with_arithmetic_operator?(processed_source[range.line])
+            start_with_arithmetic_operator?(processed_source[range.line]) ||
+            inside_string_literal?(range)
         end
 
         def ends_with_backslash_without_comment?(source_line)
@@ -97,6 +99,12 @@ module RuboCop
 
         def string_concatenation?(source_line)
           /["']\s*\\\z/.match?(source_line)
+        end
+
+        def inside_string_literal?(range)
+          processed_source.tokens.each.any? do |token|
+            ALLOWED_STRING_TOKENS.include?(token.type) && token.pos.overlaps?(range)
+          end
         end
 
         def redundant_line_continuation?(range)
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/redundant_line_continuation_spec.rb
```diff
@@ -134,6 +134,20 @@ def self.foo(bar,#{trailing_whitespace}
     RUBY
   end
 
+  it 'does not register an offense when line continuations for double quoted string' do
+    expect_no_offenses(<<~'RUBY')
+      "foo \
+        bar"
```
bquorning, line 140: The following code causes the spec to fail.
```suggestion
      a = "foo \
            bar"
```

koic: Oops, I overlooked this case. I've fixed it and updated this PR.

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -87,16 +88,20 @@ def on_new_investigation
...
         def ends_with_backslash_without_comment?(source_line)
           source_line.gsub(/#.+/, '').end_with?('\\')
         end
 
-        def string_concatenation?(source_line)
-          /["']\s*\\\z/.match?(source_line)
+        def string_concatenation?(range)
+          return true if /.*["']\s*\\\z/.match?(range.source_line)
+
+          processed_source.each_token.any? do |token|
```
bquorning, line 102: `#each_token` is deprecated: https://github.com/rubocop/rubocop-ast/blob/14844626f0c6f87aef0e74198769537b40b21b7b/lib/rubocop/ast/processed_source.rb#L88-L91
```suggestion
          processed_source.tokens.each.any? do |token|
```

lib/rubocop/cop/style/redundant_line_continuation.rb
```diff
@@ -87,16 +88,20 @@ def on_new_investigation
...
+            ALLOWED_STRING_TOKENS.include?(token.type) && token.pos.overlaps?(range)
+          end
```
bquorning, line 104: This loop looks _a bit_ slow (but I have not measured), and it doesn’t really belong inside `#string_concatenation?`. Should it perhaps be a separate method (e.g. `#inside_string_literal?`?) and called _last_ inside `#require_line_continuation?`?
```rb
          !ends_with_backslash_without_comment?(range.source_line) ||
            string_concatenation?(range.source_line) ||
            start_with_arithmetic_operator?(processed_source[range.line]) ||
            inside_string_literal?(range)
```

koic: This looks like good logic ❤️ Thank you for the suggestion!

koic:
Hm, especially `#ends_with_backslash_without_comment?` may have some problem. Any other method would be preferable to the regexp. I think the issue can be clarified and dealt with separately.

PR #11760, opened 2023-04-04
Style/GuardClause could prevent useless else after multiline guard

TSMMark:
eslint has `no-else-return` https://eslint.org/docs/latest/rules/no-else-return

I think Rubocop's `Style/GuardClause` rule could apply a similar style enforcement

---

I have committed a failing test to exemplify what I envision, and will open this PR as Draft. I would like to know if maintainers would support such functionality before I spend the time implementing it.

Thanks!

### DIFF
--- a/spec/rubocop/cop/style/guard_clause_spec.rb
+++ b/spec/rubocop/cop/style/guard_clause_spec.rb
@@ -195,6 +195,31 @@ RSpec.describe RuboCop::Cop::Style::GuardClause, :config do
     expect_no_corrections
   end
 
+  it 'registers an offense when using if / else with multiple lines' do
+    expect_offense(<<~RUBY)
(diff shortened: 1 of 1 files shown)


bbatsov:
@TSMMark Doesn't `Lint/UnreachableCode` flag this already?

TSMMark:
@bbatsov hey sorry I missed your comment from 2023! `Lint/UnreachableCode` does not flag this. All the code is actually reachable, so I think that's intended...

I'm pushing up the same test case, added in spec/rubocop/cop/lint/unreachable_code_spec.rb to prove it is not flagged.

TSMMark:
I'm not sure why the commit is not showing on this PR, but the new commit is:

9f6a3a985

PR #11764, opened 2023-04-05
[Fix #11762] Fix an incorrect autocorrect for `Style/ClassEqualityComparison`

### LINKED ISSUE #11762: AutoCorrect breaks class name comparison
RuboCop auto-correction feature makes invalid correction for class name comparison. For example it replaces `obj.class.name == cls_name` with `obj.instance_of?(cls_name)`.  This is not correct since `instance_of()` method expects a class as a parameter and not a name of the class.  As a result after this replacement code fails with the following error:
```
	1: from test.rb:2:in `<main>'
test.rb:2:in `instance_of?': class or module required (TypeError)
```

## Expected behavior

RuboCop should not perform such replacement, for example `obj.class.name == cls_name` should stay unmodified.

## Actual behavior

RuboCop replaced `obj.class.name == cls_name` with invalid Ruby code: `obj.instance_of?(cls_name)`

Create test.rb file with the following content:
```ruby
cls_name = 'String'
raise 'Failed' unless 'somestr'.class.name == cls_name
puts 'SUCCESS'
```
Verify that created file works by running `ruby test.rb`, it should output `SUCCESS`

Run RubuCop auto-correction: `rubocop --auto-correct`

Now if you run `ruby test.rb` it will fail with error:
```
	1: from test.rb:2:in `<main>'
test.rb:2:in `instance_of?': class or module required (TypeError)
```

## RuboCop version

```
~/.runtimes/Ruby27x/lib/ruby/gems/2.7.0/gems/rubocop-1.49.0/exe/rubocop -V
1.49.0 (using Parser 3.2.2.0, rubocop-ast 1.28.0, running on ruby 2.7.3) [x86_64-darwin21]
```

koic:
Fixes #11762.

This PR fixes an incorrect autocorrect for `Style/ClassEqualityComparison` when comparing a variable or return value for equality.

when a variable or return value of a method is used, it cannot be suggested and safe autocorrected because the type is not known. So this PR tweaks the cop that doesn't suggest `instance_of` argument at those times and doesn't autocorrect. Anyway, detection is possible.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/class_equality_comparison_spec.rb
+++ b/spec/rubocop/cop/style/class_equality_comparison_spec.rb
@@ -69,6 +69,43 @@ RSpec.describe RuboCop::Cop::Style::ClassEqualityComparison, :config do
         var.instance_of?(Date)
       RUBY
     end
+
+    it 'registers an offense and corrects when comparing local variable for equality' do
+      expect_offense(<<~RUBY)
+        class_name = 'Model'
+        var.class.name == class_name
+            ^^^^^^^^^^^^^^^^^^^^^^^^ Use `instance_of?` instead of comparing classes.
+      RUBY
+
+      expect_no_corrections
+    end
+
+    it 'registers an offense and corrects when comparing instance variable for equality' do
+      expect_offense(<<~RUBY)
+        var.class.name == @class_name
+            ^^^^^^^^^^^^^^^^^^^^^^^^^ Use `instance_of?` instead of comparing classes.
+      RUBY
+
+      expect_no_corrections
+    end
+
+    it 'registers an offense and corrects when comparing method call for equality' do
+      expect_offense(<<~RUBY)
+        var.class.name == class_name
+            ^^^^^^^^^^^^^^^^^^^^^^^^ Use `instance_of?` instead of comparing classes.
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

spec/rubocop/cop/style/class_equality_comparison_spec.rb
```diff
@@ -69,6 +69,43 @@
         var.instance_of?(Date)
       RUBY
     end
+
+    it 'registers an offense and corrects when comparing local variable for equality' do
```
tdeo, line 73: Hey @koic I believe most of these tests have a wrong description since they don't correct.

koic: Oops, I'm going to take it later. Thanks!

koic: Fixed in https://github.com/rubocop/rubocop/commit/b1d26463f11f72d0dba8756c2db0f0b8062ef457.

tdeo: Thanks

PR #11769, opened 2023-04-06
Remove redundant `SafeAutoCorrect: false`

ydah:
This PR remove `SafeAutoCorrect: false` from `Style/FileEmpty`, `Style/NumericPredicate` and update the documentation.
It seems that other cops also appear to be choosing one or the other, so I thought it would be best to set only one of them to false for this cop. WDYT?

-----------------

* [-] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -3743,7 +3743,6 @@ Style/FileEmpty:
                  Prefer to use `File.empty?('path/to/file')` when checking if a file is empty.
   Enabled: pending
   Safe: false
-  SafeAutoCorrect: false
   VersionAdded: '1.48'
 
 Style/FileRead:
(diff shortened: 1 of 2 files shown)


koic:
I think `Style/NumericPredicate` can also be followed along with the change reason "Remove redundant `SafeAutoCorrect: false`" instead of updating to a separate cop.

config/default.yml
```diff
@@ -4572,10 +4571,6 @@ Style/NumericPredicate:
                  numeric comparisons.
   StyleGuide: '#predicate-methods'
   Safe: false
-  # This will change to a new method call which isn't guaranteed to be on the
-  # object. Switching these methods has to be done with knowledge of the types
-  # of the variables which rubocop doesn't have.
```
koic, line 4577: Shouldn't this comment be left in `Safe: false`? Because `Safe: false` includes `SafeAutoCorrect: false`.

PR #11773, opened 2023-04-07
Make `Style/ClassStructure` aware of singleton class

koic:
It is necessary to consider how to handle module better, but singleton class (`class << self`) can be handled in the same way as class. Because they are both `class`.

Also this PR fixes the following new offense:

```console
$ bundle exec rubocop
(snip)

Offenses:

lib/rubocop/config_loader.rb:168:7: C: [Correctable] Layout/ClassStructure: constants is supposed to appear before public_methods.
      PENDING_BANNER = <<~BANNER
      ^^^^^^^^^^^^^^^^^^^^^^^^^^

1481 files inspected, 1 offense detected, 1 offe
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_make_layout_class_structure_aware_of_singleton_class.md
+++ b/changelog/new_make_layout_class_structure_aware_of_singleton_class.md
@@ -1,0 +1,1 @@
+* [#11773](https://github.com/rubocop/rubocop/pull/11773): Make `Layout/ClassStructure` aware of singleton class. ([@koic][])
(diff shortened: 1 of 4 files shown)


bbatsov:
Looks good. We should probably check the codebase for other places where we forgot to handle `sclass`, as I assume this is not the only one.

koic:
Yaeh, exactly so. I'm working on it as a follow-up work.

vlad-pisanov:
The same rule can probably be applied to dynamic class definitions too:
```ruby
Class.new do
   ...
end
```

PR #11779, opened 2023-04-09
Make `Lint/RedundantStringCoercion` aware of print method arguments

koic:
This PR makes `Lint/RedundantStringCoercion` aware of print method arguments. The name of this cop is "RedundantStringCoercion", so it makes more sense to extend the feature than to provide a new cop I think.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/redundant_string_coercion.rb
+++ b/lib/rubocop/cop/lint/redundant_string_coercion.rb
@@ -3,7 +3,7 @@
 module RuboCop
   module Cop
     module Lint
-      # Checks for string conversion in string interpolation,
+      # Checks for string conversion in string interpolation, `print`, `puts`, and `warn` arguments,
       # which is redundant.
       #
       # @example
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/lint/redundant_string_coercion.rb
```diff
@@ -3,26 +3,33 @@
...
+      # Checks for string conversion in string interpolation, `print`, `puts`, and `warn` arguments,
       # which is redundant.
       #
       # @example
       #
       #   # bad
       #
       #   "result is #{something.to_s}"
+      #   print something.to_s
+      #   puts something.to_s
```
bbatsov, line 15: Seems you forgot `warn something.to_s` here.

bbatsov:
> The name of this cop is "RedundantStringCoercion", so it makes more sense to extend the feature than to provide a new cop I think.

Totally agree.

lib/rubocop/cop/lint/redundant_string_coercion.rb
koic: Oops, I've added it.

PR #11790, opened 2023-04-12
Add spec for `Style/RescueModifier`

koic:
The following syntax changes in Ruby 2.7, so the offense highlight range of `Style/RescueModifier` was changed.
https://github.com/ruby/ruby/commit/53b3be5d58a9bf1efce229b3dce723f96e820c79

For the role of `Style/RescueModifier`, it seems reasonable to have different offense highlight ranges between Ruby 2.6 and 2.7 against these syntax. So just adding a test should be enough I think.

-----------------

* [x] Added tests.


bbatsov:
Yeah, I think that's enough.

PR #11794, opened 2023-04-12
Enable anonymous forwarding in Style/ArgumentsForwarding for Ruby 3.2+

owst:
Adds support for anonymous [rest/kwrest argument forwarding](https://rubyreferences.github.io/rubychanges/3.2.html#anonymous-arguments-passing-improvements) on Ruby 3.2+ and also rewrites the existing functionality for "forwarding all" (`...`)

Some questions I'd like to answer before fleshing this PR out any further:
1. Should this be a new cop? (I _think_ no, due to the overlap, but e.g. `Naming/BlockForwarding` seems relevant)
1. Should there be configuration to enable/disable anonymous forwarding? (I think yes but haven't done it yet - config name suggestions very welcome)
1. Is there an existing mechanism to alert for now-unnecessary options in later Ruby versions (specifically `AllowOnlyRestArgument` isn't needed in Ruby 3.2+)

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/arguments_forwarding.rb
+++ b/lib/rubocop/cop/style/arguments_forwarding.rb
@@ -10,8 +10,9 @@ module RuboCop
       #
       # In Ruby 3.2, anonymous args/kwargs forwarding has been added.
       #
-      # This cop identifies places where `use_args(*args)`/`use_kwargs(**kwargs)`
-      # can be replaced by `use_args(*)`/`use_kwargs(**)`.
+      # This cop also identifies places where `use_args(*args)`/`use_kwargs(**kwargs)` can be
+      # replaced by `use_args(*)`/`use_kwargs(**)`; if desired, this functionality can be disabled
+      # by setting UseAnonymousForwarding: false.
       #
       # @example
       #   # bad
--- a/lib/rubocop/cop/style/arguments_forwarding.rb
+++ b/lib/rubocop/cop/style/arguments_forwarding.rb
@@ -81,7 +81,7 @@ module RuboCop
 
         FORWARDING_LVAR_TYPES = %i[splat kwsplat block_pass].freeze
 
-        FORWARDING_MSG = 'Use arguments forwarding.'
+        FORWARDING_MSG = 'Use shorthand syntax `...` for arguments forwarding.'
         ARGS_MSG = 'Use anonymous positional arguments forwarding.'
         KWARGS_MSG = 'Use anonymous keyword arguments forwarding.'
 
--- a/lib/rubocop/cop/style/arguments_forwarding.rb
+++ b/lib/rubocop/cop/style/arguments_forwarding.rb
@@ -8,6 +8,11 @@ module RuboCop
       # This cop identifies places where `do_something(*args, &block)`
       # can be replaced by `do_something(...)`.
       #
+      # In Ruby 3.2, anonymous args/kwargs forwarding has been added.
+      #
+      # This cop identifies places where `use_args(*args)`/`use_kwargs(**kwargs)`
+      # can be replaced by `use_args(*)`/`use_kwargs(**)`.
+      #
       # @example
       #   # bad
       #   def foo(*args, &block)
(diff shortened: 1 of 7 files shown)


owst:
Hi @koic, as the original author of this modified Cop I wonder if you'd be able to give a preliminary review of this PR (and hopefully answer the questions in the description)? Thanks in advance!

owst:
Hi @bbatsov, @koic, is there anything I can do to help progress this PR?

bbatsov:
@owst I didn't pay it much attention because I'm was vacation until today and it was marked as draft. :-)

owst:
I hope you had a good vacation @bbatsov! 

I've intentionally left the PR as draft, as there are three questions that I'd like to answer before polishing this into a ready for review PR. N.b. I think the essence of the change is right and can be reviewed/critiqued, it's just a few finer points that need sorting out. Any thoughts/answers you have would be greatfully received 😄

bbatsov:
> Should this be a new cop? (I think no, due to the overlap, but e.g. Naming/BlockForwarding seems relevant)

No need for a new cop IMO. 

> Should there be configuration to enable/disable anonymous forwarding? (I think yes but haven't done it yet - config name suggestions very welcome)

Let's make it configurable. The name will depend on what we want to do exactly - I assume we want to use anonymous forwarding everywhere that's possible, right? Something like `UseAnonymousForwarding` sounds reasonable. 

> Is there an existing mechanism to alert for now-unnecessary options in later Ruby versions (specifically AllowOnlyRestArgument isn't needed in Ruby 3.2+)

I can't think of anything generic that we have. @koic correct me if I'm wrong.

owst:
Thanks @bbatsov - I've polished this PR and I think it's now ready for review 👍

lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -8,6 +8,11 @@ module Style
       # This cop identifies places where `do_something(*args, &block)`
       # can be replaced by `do_something(...)`.
       #
+      # In Ruby 3.2, anonymous args/kwargs forwarding has been added.
+      #
+      # This cop identifies places where `use_args(*args)`/`use_kwargs(**kwargs)`
```
bbatsov, line 13: Probably you should mention the config option controlling this behavior here, otherwise people might think that's something that's always getting checked.

bbatsov:
Looks good to me overall, apart from my small remark. @koic any feedback from you?

lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -53,86 +78,142 @@ class ArgumentsForwarding < Base
...
-              (block-pass {(lvar %2) nil?}))
-            (send _ _
-              (splat (lvar %1))
-              (hash (kwsplat (lvar %3)))
-              (block-pass {(lvar %2) nil?}))
-          }
-        PATTERN
+        FORWARDING_LVAR_TYPES = %i[splat kwsplat block_pass].freeze
+
+        FORWARDING_MSG = 'Use arguments forwarding.'
```
bbatsov, line 83: Might be good to mention the specific syntax for this in the message, as I'm not sure everyone's familiar with those names.

owst: I somewhat agree, but this was the same message as previously used in this cop, so I think I'd favour leaving it as it is, unless you feel strongly about it?

bbatsov: Let's change this, I just didn't notice it when the cop was originally introduced. You can do it in a separate commit or the same. It's all the same to me.

bbatsov:
I see you updated the first message, but I meant updating all messages. Ping me when this is ready to be reviewed again.

owst:
Sorry @bbatsov I hadn't realised that's what you meant - I've pushed another commit with a change for the two remaining messages.

lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -53,86 +79,142 @@ class ArgumentsForwarding < Base
...
-            (send _ _
-              (splat (lvar %1))
-              (hash (kwsplat (lvar %3)))
-              (block-pass {(lvar %2) nil?}))
-          }
-        PATTERN
+        FORWARDING_LVAR_TYPES = %i[splat kwsplat block_pass].freeze
+
+        FORWARDING_MSG = 'Use shorthand syntax `...` for arguments forwarding.'
+        ARGS_MSG = 'Use anonymous positional arguments forwarding: `*`.'
```
bbatsov, line 85: I think it'd be better to use parens here instead of `: *` - (`*`). To me it reads a bit better. Not an expert on the subject though. :-)

owst: Sure thing, I've pushed another a commit 👍

owst: I think the build failure is unrelated to this PR, and unfortunately I can't rerun it

bbatsov (approved):
Looks good to me. Thanks for working on this!

PR #11807, opened 2023-04-17
Do not try to expand nil Include when inheriting.

juanxo:
When inheriting from a parent config file that has nullified an Include in any rule, it will currently fail as it will assume that the Include will always be an non-nullable array

Example:

```
# rubocop.yml
Style/StringLiterals:
  Include:
    - dir/**/*.rb

# child_dir/.rubocop.yml
inherit_from: ../.rubocop.yml
Style/StringLiterals:
  Include: ~
```

I'm currently using this pattern to remove a default Include pattern so it resorts back to the AllCops inclusion list, so setting it to `[]` does not accomplish the same effect.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/config_loader_spec.rb
+++ b/spec/rubocop/config_loader_spec.rb
@@ -286,6 +286,32 @@ RSpec.describe RuboCop::ConfigLoader do
         expect(configuration_from_file['Style/StringLiterals']['Include']).to eq(['dir/**/*.rb'])
       end
 
+      it 'does not expand a nil Include' do
+        create_file(file_path, <<~YAML)
+          inherit_from: ../.rubocop.yml
+          Style/StringLiterals:
+            Include: ~
+        YAML
+
+        expect(configuration_from_file['Style/StringLiterals']['Include'].nil?).to be(true)
+      end
+
+      it 'gets an Include that is relative to the subdirectory when inheriting from a nil Include' do
+        create_file('.rubocop.yml', <<~YAML)
+          Style/StringLiterals:
+            Include: ~
+        YAML
+
+        create_file(file_path, <<~YAML)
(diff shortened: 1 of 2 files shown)


spec/rubocop/config_loader_spec.rb
```diff
@@ -286,6 +286,32 @@
...
+        create_file(file_path, <<~YAML)
+          inherit_from: ../.rubocop.yml
+          Style/StringLiterals:
+            Include: ~
+        YAML
+
+        expect(configuration_from_file['Style/StringLiterals']['Include']).to be_nil
+      end
+
+      it 'gets an Include that is relative to the subdirectory when inheriting from a nil Include' do
```
juanxo, line 299: Not so sure about this test. I wanted to mostly check that going from not nil to nil and vice versa worked

bbatsov:
> I'm currently using this pattern to remove a default Include pattern so it resorts back to the AllCops inclusion list, so setting it to [] does not accomplish the same effect.

I'm OK with your proposal, but this is also something we should document in the configuration section of RuboCop, as generally I don't think we've considered the use of `nil` much there. //cc @koic @jonas054

koic:
Yeah, I agree with @bbatsov.

bbatsov:
@juanxo Ping :-)

jonas054 (approved):
Looks good!

bbatsov:
@juanxo Ping again :-)

bbatsov:
@juanxo Ping again :-)

koic:
@juanxo Ping.

juanxo:
Hey, sorry for the long delay in answering. Will look to rebase and ship this

bbatsov:
@juanxo ping :-)

PR #11824, opened 2023-04-28
Add autocorrection for `Lint/TopLevelReturnWithArgument`

r7kamura:
I have added an autocorrection that simply removes the argument part from top-level `return`.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/top_level_return_with_argument.rb
+++ b/lib/rubocop/cop/lint/top_level_return_with_argument.rb
@@ -8,25 +8,39 @@ module RuboCop
       # always ignored. This is detected automatically since Ruby 2.7.
       #
       # @example
+      #   # bad
+      #   return 1
       #
-      #   # Detected since Ruby 2.7
-      #   return 1 # 1 is always ignored.
+      #   # good
+      #   return
       class TopLevelReturnWithArgument < Base
-        # This cop works by validating the ancestors of the return node. A
-        # top-level return node's ancestors should not be of block, def, or
-        # defs type.
+        extend AutoCorrector
 
         MSG = 'Top level return with argument detected.'
 
         def on_return(return_node)
-          add_offense(return_node) if return_node.arguments? && ancestors_valid?(return_node)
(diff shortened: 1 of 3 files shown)


koic:
I think there is not a fix other than removing arguments, so this logic looks acceptable to me.

lib/rubocop/cop/lint/top_level_return_with_argument.rb
```diff
@@ -8,25 +8,39 @@ module Lint
...
+        # This cop works by validating the ancestors of the return node. A
+        # top-level return node's ancestors should not be of block, def, or
+        # defs type.
+        def top_level_return?(return_node)
+          return_node.each_ancestor(:block, :def, :defs).none?
+        end
+
+        def top_level_return_with_any_argument?(return_node)
+          top_level_return?(return_node) && return_node.arguments?
         end
```
koic, line 44: That's just my two cents. Can you reorder it?

```suggestion
        def top_level_return_with_any_argument?(return_node)
          top_level_return?(return_node) && return_node.arguments?
        end

        def remove_arguments(corrector, return_node)
          corrector.replace(return_node, 'return')
        end

        # This cop works by validating the ancestors of the return node. A
        # top-level return node's ancestors should not be of block, def, or
        # defs type.
        def top_level_return?(return_node)
          return_node.each_ancestor(:block, :def, :defs).none?
        end
```

r7kamura: Okay, I will change it that way 👌 
IMHO it would be less maintainable if the definition order changed with the writer's sense, so I went with alphabetical order because that can be determined statically and similar methods are placed close together, but I have no strong preference about this.

koic: Thank you. This fix has been updated across the board, so there is no significant impact on this change. (In other words, this PR change doesn't impact the order.) As a long time active maintainer I like the newspaper style where Uncle Bob is told in Clean Code, but dislike alphabetical order style and it's not maintainable to me.

PR #11825, opened 2023-04-28
Mark `Lint/TopLevelReturnWithArgument` as unsafe

r7kamura:
For example, the following code may be written in jb:

```ruby
# example.jb
return {} if article.nil?

{
  id: article.id,
  name: article.name
}
```

This cop detects offenses in this case, but in practice this `{}` is used as the correct return value, so it is a false positive. jb is [officially supported by RuboCop](https://github.com/rubocop/rubocop/blob/b980e7cd1b691aa9121b77b6bc4456146eb326e8/config/default.yml#L11-L62), and one can well assume that RuboCop will inspect other Ruby files that are used in this way. Therefore, I think it is reasonable to assume that this cop is unsafe.

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/change_mark_lint_top_level_return_with_argument_as_unsafe.md
+++ b/changelog/change_mark_lint_top_level_return_with_argument_as_unsafe.md
@@ -1,0 +1,1 @@
+* [#11825](https://github.com/rubocop/rubocop/pull/11825): Mark `Lint/TopLevelReturnWithArgument` as unsafe. ([@r7kamura][])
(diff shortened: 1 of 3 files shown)


r7kamura:
As a further improvement idea, since it is possible to predict from the file name or extension whether a file is likely to be used through `eval` or not, adding that pattern to `Exclude` would reduce false positives.

koic:
> adding that pattern to `Exclude` would reduce false positives.


I like the idea. Considering `eval` probably leads to may unsafe cops from some safe cops. It may lose the linter's role. So, it should be excluded unsafe extensions instead of `Safe: false`.

r7kamura:
OK, so I've created another pull request for the idea, so I'll close this one 👌 

- https://github.com/rubocop/rubocop/pull/11826

PR #6293, opened 2018-09-14
Extend cops meta data

Darhazer:
* Merged default.yml, enabled.yml, and disabled.yml into single config
* Added version added/changed/removed for all cops since 0.5 (428/428 cops)

TODO:
* [x] Update documentation generating script to include meta info
* [x] Add Safe & SafeAutocorrect meta-data (currently 6 cops are marked unsafe, 10 with unsafe autocorrect, probably needs more work)
* [x] Add version meta data for older releases 
* [ ] ~Update OBSOLETE_COPS checks to check VersionRemoved meta data~ We don't have VersionRemoved in this release, this should be a follow up for 1.0
* [x] Update manual/configuration.md
* [x] Review EnabledByDefault/DisabledByDefault option

I wonder what should we do with removed cops? Add them in a separate config? Leave them in default.yml in alphabetical order? Just remove them and start providing the VersionRemoved from 1.0 onwards?

Currently they are in a separate file, not required in default.yml, as otherwise config loader breaks.

-----------------

Before submitting the PR make sure the following are checked:

* [ ] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [ ] Feature branch is up-to-date with `master` (if not - rebase it).
* [ ] Squashed related commits together.
* [ ] Added tests.
* [ ] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes. See [changelog entry format](../blob/master/CONTRIBUTING.md#changelog-entry-format).
* [ ] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.
* [ ] Run `rake default`. It executes all tests and RuboCop for itself, and generates the documentation.

[1]: http://chris.beams.io/posts/git-commit/

PR #11826, opened 2023-04-28
Exclude `**/*.jb` from `Lint/TopLevelReturnWithArgument`

r7kamura:
Since non-standard Ruby files such as `*.jb` and so on are expected to be used via `eval` or equivalent, it would be better to exclude them from this Cop's inspection to reduce false positives.

See also the following discussion for more information:

- https://github.com/rubocop/rubocop/pull/11825#issuecomment-1527002005

The patterns specified in `Exclude` are based from `default_configuration['AllCops']['Include']`, excluding the following patterns:

- `**/*.rb`
- `**/*.rbw`
- `**/*.rbx`
- `**/*.ruby`

-----------------

* [ ] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2343,6 +2343,9 @@ Lint/TopLevelReturnWithArgument:
   Description: 'Detects top level return statements with argument.'
   Enabled: true
   VersionAdded: '0.89'
+  # These codes are `eval`-ed in method and their return values may be used.
+  Exclude:
+    - '**/*.jb'
 
 Lint/TrailingCommaInAttributeDeclaration:
   Description: 'Checks for trailing commas in attribute declarations.'
--- a/changelog/change_exclude_non_standard_ruby_files_from.md
+++ b/changelog/change_exclude_non_standard_ruby_files_from.md
@@ -1,0 +1,1 @@
+* [#11826](https://github.com/rubocop/rubocop/pull/11826): Exclude `**/*.jb` from `Lint/TopLevelReturnWithArgument`. ([@r7kamura][])


config/default.yml
```diff
@@ -2343,6 +2343,55 @@ Lint/TopLevelReturnWithArgument:
...
+    - '**/Mavenfile'
+    - '**/Podfile'
+    - '**/Puppetfile'
+    - '**/Rakefile'
+    - '**/rakefile'
+    - '**/Snapfile'
+    - '**/Steepfile'
+    - '**/Thorfile'
+    - '**/Vagabondfile'
+    - '**/Vagrantfile'
```
koic, line 2394: Hm, `Exclude` too many may raise false negatives. It's probably better to target only `jb` that actually have issue.

config/default.yml
```diff
@@ -2343,6 +2343,9 @@ Lint/TopLevelReturnWithArgument:
   Description: 'Detects top level return statements with argument.'
   Enabled: true
   VersionAdded: '0.89'
+  VersionChanged: '<<next>>'
```
koic, line 2346: I think it is unnecessary because this is a bugfix and doesn't imply a default configuration changes.

```suggestion
```

r7kamura: I thought that `VersionChanged` should always be changed when there is any change in the default configuration of the Cop, regardless of the purpose or background of the change.  Shouldn't it be changed in the case of a bug fix?

- https://github.com/rubocop/rubocop/pull/6293#issuecomment-423640987

koic: As a short answer yes for here. it's unnecessary given the current context I think (It's not a one-size-fits-all) .

r7kamura: I see, thank you for explaining 👍

config/default.yml
```diff
@@ -2343,6 +2343,8 @@ Lint/TopLevelReturnWithArgument:
   Description: 'Detects top level return statements with argument.'
   Enabled: true
   VersionAdded: '0.89'
+  Exclude:
+    - '**/*.jb'
```
koic, line 2347: It would be useful to leave a comment with the reason for excluding it by default.

koic:
Thanks!

Issue #11831, opened 2023-04-30
`Naming/InclusiveLanguage`'s `AllowedRegex` does not support free-spacing/extended mode Regexp


sambostock:
[Ruby's Regexp "free-spacing"/"extended" mode (`/.../x`)](https://docs.ruby-lang.org/en/master/Regexp.html#class-Regexp-label-Free-Spacing+Mode+and+Comments) allows the use of additional whitespace and comments, which is useful to express a complex or otherwise long regular expression.

These are (de)serializable as YAML, the format used in RuboCop configuration files.

--------

## Expected behavior

As such, it should be possible to use such a regular expression as configuration for the `Naming/InclusiveLanguage` cop's `AllowedRegex` config (and all other cops, for that matter).

Given the files listed in the reproduction section below:

```sh
$ bundle exec rubocop --only Naming/InclusiveLanguage
Inspecting 2 files
..

2 files inspected, no offenses detected
```

## Actual behavior

```
bundle exec rubocop --only Naming/InclusiveLanguage 
Inspecting 2 files
.C

Offenses:

example.rb:1:2: C: Naming/InclusiveLanguage: Consider replacing 'master' with another term.
:mastercard
 ^^^^^^
example.rb:2:2: C: Naming/InclusiveLanguage: Consider replacing 'master' with another term.
:mastermind
 ^^^^^^

2 files inspected, 2 offenses detected
```

<details><summary><code>--debug</code></summary>

```
bundle exec rubocop --debug --only Naming/InclusiveLanguage
For /Users/sambostock/src/local/rubocop_inclusive_language_free_spacing.rb: configuration from /Users/sambostock/src/local/rubocop_inclusive_language_free_spacing.rb/.rubocop.yml
Default configuration from /Users/sambostock/.gem/ruby/3.2.2/gems/rubocop-1.50.2/config/default.yml
Use parallel by default.
Running parallel inspection
Loading cache from /Users/sambostock/.cache/rubocop_cache/7a2a65ad3891cc55333c8e84fe1fc49849c788ea/98f069e84e9f37415c1544e9767039eaea49508b/3d0d8b164e427f6266892b307ab7c8935e882152
Loading cache from /Users/sambostock/.cache/rubocop_cache/7a2a65ad3891cc55333c8e84fe1fc49849c788ea/98f069e84e9f37415c1544e9767039eaea49508b/f04a6c2e0a12bc07fad2192371c0cdf651355532
Inspecting 2 files
[...]

PR #11832, opened 2023-05-01
Support free-spacing mode `Regexp` in `Naming/InclusiveLanguage`

sambostock:
WIP – See #11831.

This edits the way the `Naming/InclusiveLanguage` cop combines its config into `Regexp` to match terms. Specifically, because we would convert given `Regexp` to `String`s via `Regexp#source`, and then simply combine patterns/strings using

```ruby
Regexp.new(strings.join('|'), Regexp::IGNORECASE)
```

we would lose any mode/options information about the `Regexp`, which in the case of ["free-spacing"/"extended" mode](https://docs.ruby-lang.org/en/master/Regexp.html#class-Regexp-label-Free-Spacing+Mode+and+Comments), would mean the `Regexp` could be corrupted by additional spacing being considered part of the pattern.

This changes the approach to normalize patterns as `Regexp`s instead of `String`s, and then simply combine them using `Regexp.union`. To preserve the existing behaviour of always making the patterns case insensitive, we extract the source and options from any given `Regexp`, and construct a new one where we force case insensitivity using

```ruby
Regexp.new(regexp.source, regexp.options | Regexp::IGNORECASE)
```

This way, we preserve all other flags (`/.../x`, but also others like `/.../m`).

Note that it was previously possible to work around this by applying options at the subexpression level, which remains possible (e.g. to force case **sensitivity**).

### TODO

- [ ] Validate approach
- [ ] Rename identifiers accordingly
- [ ] Add more specs for edge cases and covering existing behaviour, to ensure it is preserved.
- [ ] Better document use of `Regexp` in cop documentation (unclear how `AllowedRegexp` works; maybe separate PR?)

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/naming/inclusive_language.rb
+++ b/lib/rubocop/cop/naming/inclusive_language.rb
@@ -148,7 +148,7 @@ module RuboCop
             next if term_definition.nil?
 
             allowed_strings.concat(process_allowed_regex(term_definition['AllowedRegex']))
-            regex_string = ensure_regex_string(extract_regexp(term, term_definition))
+            regex_string = ensure_case_insensitive_regexp(extract_regexp(term, term_definition))
             flagged_term_strings << regex_string
 
             add_to_flagged_term_hash(regex_string, term, term_definition)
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/naming/inclusive_language.rb
```diff
@@ -182,12 +184,16 @@ def process_allowed_regex(allowed)
...
-            ensure_regex_string(allowed_term)
+            ensure_case_insensitive_regexp(allowed_term)
           end
         end
 
-        def ensure_regex_string(regex)
-          regex.is_a?(Regexp) ? regex.source : regex
+        def ensure_case_insensitive_regexp(object)
+          case object
+          when Regexp then Regexp.new(object.source, object.options | Regexp::IGNORECASE)
```
sambostock, line 193: Could optimize to skip creating duplicates

```suggestion
          when Regexp
            return object if object.options == Regexp::IGNORECASE

            Regexp.new(object.source, object.options | Regexp::IGNORECASE)
```

bbatsov:
@sambostock Do you plan to wrap this up or should we close it?

PR #11833, opened 2023-05-01
Add TargetRailsVersion to default.yml

searls:
After discovering that a valid configuration can't be created programmatically if it includes `TargetRailsVersion` without [patching it in](https://github.com/rubocop/rubocop-rails/blob/master/lib/rubocop/rails/inject.rb#L14) by doing an `instance_variable_set` on the `ConfigLoader`'s `@default_configuration`, it seems appropriate to explicitly list this property even though it's specific to rubocop-rails.

For example, before this commit, running: 

```ruby
RuboCop::Config.create({
  "AllCops" => {
    "TargetRailsVersion" => 5.2
  }
}, "some_path")
```

Will emit this warning:

```
Warning: AllCops does not support TargetRailsVersion parameter.

Supported parameters are:

  - RubyInterpreters
  - Include
  - Exclude
  - DefaultFormatter
  - DisplayCopNames
  - DisplayStyleGuide
  - StyleGuideBaseURL
  - DocumentationBaseURL
  - ExtraDetails
  - StyleGuideCopsOnly
  - EnabledByDefault
  - DisabledByDefault
  - NewCops
  - UseCache
  - MaxFilesInCache
  - CacheRootDirectory
  - AllowSymlinksInCacheRootDirectory
  - TargetRubyVersion
  - SuggestExtensions
  - ActiveSupportExtensionsEnabled
```

Even though this is the domain of a plugin, it seems appropriate to add it since:

1. `ActiveSupportExtensionsEnabled` has similarly been added (perhaps for the same reason?)
2. RuboCop itself (and not rubocop-rails) actually contains the logic handling this setting  in [Config](/lib/rubocop/config.rb#L249-L258)

-----------------

* [ ] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -156,6 +156,8 @@ AllCops:
     rubocop-capybara: [capybara]
   # Enable/Disable checking the methods extended by Active Support.
   ActiveSupportExtensionsEnabled: false
+  # Used by rubocop-rails to set the version of Rails used by the inspected code
+  TargetRailsVersion: ~
 
 #################### Bundler ###############################
 


koic:
This is a tricky part.

> 1. `ActiveSupportExtensionsEnabled` has similarly been added (perhaps for the same reason?)

This option has been introduced for implementation convenience. For example RuboCop Rails sets `ActiveSupportExtensionsEnabled: true`. It's not positive, but it's a realistic solution for now. Recently suggested: https://github.com/rubocop/rubocop-performance/pull/342#issuecomment-1531080588

> RuboCop itself (and not rubocop-rails) actually contains the logic handling this setting in [Config](https://github.com/lib/rubocop/config.rb#L249-L258)

Code around `TargetRailsVersion` should be extracted to RuboCop Rails, but no progress yet. Either way it would be expected to be removed from RuboCop core.

I'm still not sure if it's best to restore the `TargetRailsVersion` to core.

searls:
Agree that this is tricky! The division is not very clear right now which makes it a little hard to work with the API. 

I was able to work around these issues for standard-rails yesterday, so there's no urgency in my request

bbatsov:
> Code around TargetRailsVersion should be extracted to RuboCop Rails, but no progress yet. Either way it would be expected to be removed from RuboCop core in future.

Agreed.

Arlantir:
Hi, are you planning to add `TargetRailsVersion`?

bbatsov:
It seems we won't be moving forward with this.

PR #11841, opened 2023-05-03
Add new `Style/RedundantFilterChain` cop

fatkodima:
This a https://github.com/rubocop/rubocop-performance/pull/342 recreated for rubocop itself.
Closes https://github.com/rubocop/rubocop-performance/pull/342
cc @koic


lib/rubocop/cop/style/redundant_filter_chain.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Identifies usages of `any?`, `empty?` or `none?` predicate methods
+      # chained to `select`/`filter` and change them to use predicate method instead.
```
bbatsov, line 7: Don't forget about `find_all`. :-) Too many aliases!

lib/rubocop/cop/style/redundant_filter_chain.rb
```diff
@@ -0,0 +1,90 @@
...
+        def_node_matcher :select_predicate?, <<~PATTERN
+          (send
+            {
+              (block $(send _ {:select :filter}) ...)
+              $(send _ {:select :filter} block_pass_type?)
+            }
+            ${:#{RESTRICT_ON_SEND.join(' :')}})
+        PATTERN
+
+        REPLACEMENT_METHODS = {
```
bbatsov, line 45: Might be better to make this configurable instead of hardcoded. This would also remove the need to hardcode the support for `many?`.

bbatsov: (whether it's useful to do this depends mostly on how common is for people to encounter other similar methods in the wild)

fatkodima: We need a consensus https://github.com/rubocop/rubocop-performance/pull/342#issuecomment-1531080588 😅

bbatsov: I'm fine with @koic's suggestion.

bbatsov: It seem the PR currently implements it, right?

fatkodima: Yes. Expanded docs with an additional example.

lib/rubocop/cop/style/redundant_filter_chain.rb
bbatsov: @fatkodima Not sure if you saw this comment.

fatkodima: Ah, sorry, somehow missed this. Updated with this suggestion.

bbatsov:
Thanks!

PR #11843, opened 2023-05-03
Simplify `inspect` output

sambostock:
When debugging errors, cops and configs may be inspected as part of error messages.

Extremely long error messages can be problematic when debugging, as they may exhaust the terminal scrollback buffer, without the visible portion necesarily providing useful information.

Because the config typically contains at least the entire default config, its `inspect` output is unusably long.

Because cops contain many instance variables relating to the file being investigated (which may be long), their `inspect` output is often unusably long.

Simplifying the `inspect` output for such objects makes error output more usable, and is a technique used in other gems, such as Rails.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -305,6 +305,10 @@ module RuboCop
         @current_original = original
       end
 
+      def inspect # :nodoc:
+        "#<#{self.class.name}:#{object_id} @config=#{@config} @options=#{@options}>"
+      end
+
       private
 
       ### Reserved for Cop::Cop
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/base.rb
```diff
@@ -305,6 +305,10 @@ def begin_investigation(processed_source, offset: 0, original: processed_source)
         @current_original = original
       end
 
+      def inspect # :nodoc:
+        "#<#{self.class.name}:#{object_id} @config=#{@config} @options=#{@options}>"
```
sambostock, line 309: I tried to retain a useful subset of variables.

sambostock (commented):
I'm not sure if this warrants tests.

There may also be other objects that could benefit from this.

bbatsov:
Can you drop here some output before/after the change?

PR #11847, opened 2023-05-05
[Fix #11846] Fix a false positive for `Lint/RedundantStringCoercion`

### LINKED ISSUE #11846: Lint/RedundantStringCoercion incorrectly flags uses of #to_s that customize the output string
## Expected behavior

The cop `Lint/RedundantStringCoercion` flags unnecessary usage of #to_s when it is implicit (eg. `puts "My name is #{name.to_s}"`

However there are cases where it is desirable to call an object's #to_s method in order to pass in arguments which customize the output string.

Here is an example from some code of mine using libxml:

```ruby
  # format and print xml
  xdoc = XML::Document.string(value)
  puts xdoc.to_s(indent: true)
```
The call to `#to_s` determines how indentation is inserted into the resulting output string.

## Expected behaviour 
This code should not be caught by the Lint/RedundantStringCoercion  cop

## Actual behavior
This code is caught by the cop.

I recommend that calls to `#to_s` be ignored by this cop if they are being passed an argument.

## RuboCop version
1.50.2 (using Parser 3.2.2.1, rubocop-ast 1.28.0, running on ruby 2.7.4) [x86_64-linux-gnu]

koic:
Fixes #11846.

This PR fixes a false positive for `Lint/RedundantStringCoercion` when using `to_s(argument)` in `puts` argument.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_lint_redundant_string_coercion.md
+++ b/changelog/fix_a_false_positive_for_lint_redundant_string_coercion.md
@@ -1,0 +1,1 @@
+* [#11846](https://github.com/rubocop/rubocop/issues/11846): Fix a false positive for `Lint/RedundantStringCoercion` when using `to_s(argument)` in `puts` argument. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #11854, opened 2023-05-06
Add autocorrection for `Style/HashLikeCase`

r7kamura:
Implemented autocorrection that replaces case-when with Hash literal lookup.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/hash_like_case_spec.rb
+++ b/spec/rubocop/cop/style/hash_like_case_spec.rb
@@ -14,6 +14,10 @@ RSpec.describe RuboCop::Cop::Style::HashLikeCase, :config do
           'BAR'
         end
       RUBY
+
+      expect_correction(<<~RUBY)
+        { 'foo' => 'FOO', 'bar' => 'BAR' }[x]
+      RUBY
     end
 
     it 'registers an offense when using `case-when` with symbol conditions and literal bodies of the same type' do
@@ -25,6 +29,10 @@ RSpec.describe RuboCop::Cop::Style::HashLikeCase, :config do
         when :bar
           'BAR'
         end
+      RUBY
+
+      expect_correction(<<~RUBY)
+        { :foo => 'FOO', :bar => 'BAR' }[x]
       RUBY
     end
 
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/hash_like_case_spec.rb
```diff
@@ -14,6 +14,10 @@
           'BAR'
         end
       RUBY
+
+      expect_correction(<<~RUBY)
+        { 'foo' => 'FOO', 'bar' => 'BAR' }[x]
```
koic, line 19: I suspect that replacing it with a Hash instance may degrade instantiation performance and increase memory consumption. Therefore, I think this cop cannot support autocorrection.

r7kamura: I think no matter how it is implemented, as long as it is replaced by Hash, the memory increase will be unavoidable.

> Checks for places where `case-when` represents a simple 1:1 mapping and can be replaced with a hash lookup.

If we believe that the performance degradation caused by the use of Hash should be avoided, then doesn't this cop have no raison d'etre? Or are you saying that if the Hash is stored in a constant, etc., it is OK because the degree of degradation looks small?

koic: > if the Hash is stored in a constant,

Yep, It can be avoided by constantization, but autocorrection cannot be named.

r7kamura: Hmmm. The reason I asked that question is that I think this problem can be neatly solved if we have a new cop that cares about performance, not style, that says "large Hash literals should be stored in constants". What do you think about this idea?
I think it is a good thing to be able to separate the concerns, since the cop in the Style department should not be that strongly concerned about performance.

koic: Released cop should be stable. What's going on with this cop-introduced PR discussion?

koic: Also, autocorrection is only a supplement. This cop was introduced as a style cop and I doubt it should move to performance now.

r7kamura: I thought that if we discuss whether the performance degradation caused by this performance-unsafe autocorrection is likely to be resolved by another cop, it will affect whether the change I am proposing in this pull request is acceptable.
Nevertheless, from what you say, it sounds like we should be conservative about changes to the existing cops and not add autocorrection here 👌

PR #11857, opened 2023-05-06
Server mode: only read $stdin when -s or --stdin argument provided

naveg:
When using `--server` in a non-tty environment, rubocop hangs reading
STDIN. Avoid this by only reading STDIN when the `-s` or `--stdin`
arguments are provided, in which case text on STDIN is expected.

This change allows `rubocop --server` to be used in non-tty contexts,
such as File Watchers or External Tools in RubyMine.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_server_mode_only_read_stdin_when_s_or_stdin.md
+++ b/changelog/fix_server_mode_only_read_stdin_when_s_or_stdin.md
@@ -1,0 +1,1 @@
+* [#11857](https://github.com/rubocop/rubocop/pull/11857): Server mode: only read $stdin when -s or --stdin argument provided. ([@naveg][])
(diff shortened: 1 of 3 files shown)


koic:
It's best if repro tests are written. Can you write it?

naveg:
@koic done. It was a little tricky because `rubocop_server_spec` forks, so we can't mock `$stdin` there. I wrote a new unit test for `ClientCommand::Exec` that makes sure `$stdin.read` is only called with `-s/--stdin`

koic:
Is it possible to write real tests like https://github.com/rubocop/rubocop/blob/master/spec/rubocop/server/rubocop_server_spec.rb instead of mock stub tests?

naveg:
Not really, as explained in my comment above. The `Exec` code runs in a forked process, so we can't mock `$stdin.tty?`, which is true in the test environment but must be false for these tests.

Open to suggestions if you have ideas

bbatsov:
I think that's a reasonable and simple change, so I can live without an integration test for it. Thanks!

PR #11488, opened 2023-01-24
[Fix #11485] Fix a false positive for `Lint/UselessAssignment`

koic:
Fixes #11485.

This PR fixes a false positive for `Lint/UselessAssignment` when using numbered block parameter.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #11858, opened 2023-05-06
Fix `find_end_line` for block nodes.

reitermarkus:
When two block nodes are in the root, e.g.

```ruby
block1 do
  # ...
end

block2 do
  # ...
end
```

calling `find_end_line` on the second will return `1` instead of `7`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_find_end_line_for_blocks.md
+++ b/changelog/fix_find_end_line_for_blocks.md
@@ -1,0 +1,1 @@
+* [#11858](https://github.com/rubocop/rubocop/pull/11858): Fix false positives when using source comments in blocks. ([@reitermarkus][])
--- a/lib/rubocop/cop/mixin/comments_help.rb
+++ b/lib/rubocop/cop/mixin/comments_help.rb
@@ -62,7 +62,7 @@ module RuboCop
       # Returns the end line of a node, which might be a comment and not part of the AST
       # End line is considered either the line at which another node starts, or
       # the line at which the parent node ends.
-      # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity, Lint/DuplicateBranch
       def find_end_line(node)
         if node.if_type? && node.else?
           node.loc.else.line
@@ -70,6 +70,8 @@ module RuboCop
           node.else_branch.loc.line
         elsif node.if_type? && node.elsif?
           node.each_ancestor(:if).find(&:if?).loc.end.line
+        elsif node.block_type? || node.numblock_type?
+          node.loc.end.line
         elsif (next_sibling = node.right_sibling) && next_sibling.is_a?(AST::Node)
           next_sibling.loc.line
         elsif (parent = node.parent)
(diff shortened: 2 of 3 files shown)


koic:
Can you add a test case and a changelog entry?

lib/rubocop/cop/mixin/comments_help.rb
```diff
@@ -62,14 +62,16 @@ def buffer
...
-      # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity, Lint/DuplicateBranch
       def find_end_line(node)
         if node.if_type? && node.else?
           node.loc.else.line
         elsif node.if_type? && node.ternary?
           node.else_branch.loc.line
         elsif node.if_type? && node.elsif?
           node.each_ancestor(:if).find(&:if?).loc.end.line
+        elsif node.block_type?
```
koic, line 73: `numblock_type?` is the same, right?

reitermarkus: I guess so. What is a `numblock_type?` tough?

reitermarkus:
Are there any existing tests for `CommentsHelp`? I only ran into this issue while implementing a custom cop, so not sure about the best way to test this here.

lib/rubocop/cop/mixin/comments_help.rb
koic: e.g. https://github.com/rubocop/rubocop/pull/11488

koic:
It would be better to add a reproducible test case in any cop that mixin `CommentsHelp`:

```console
$ git grep 'include CommentsHelp'
lib/rubocop/cop/layout/class_structure.rb:137:        include CommentsHelp
lib/rubocop/cop/lint/empty_conditional_body.rb:65:        include CommentsHelp
lib/rubocop/cop/lint/empty_in_pattern.rb:47:        include CommentsHelp
lib/rubocop/cop/lint/empty_when.rb:48:        include CommentsHelp
lib/rubocop/cop/style/class_methods_definitions.rb:63:        include CommentsHelp
lib/rubocop/cop/style/if_unless_modifier.rb:68:        include CommentsHelp
lib/rubocop/cop/style/multiline_ternary_operator.rb:37:        include CommentsHelp
lib/rubocop/cop/style/redundant_initialize.rb:98:        include CommentsHelp
lib/rubocop/cop/style/symbol_proc.rb:88:        include CommentsHelp
```

It may not be the best, but having regression tests is essential.

reitermarkus:
Thanks, I found a cop that runs on blocks which actually has this problem.

changelog/fix_find_end_line_for_blocks.md
```diff
@@ -0,0 +1 @@
+* [#11858](https://github.com/rubocop/rubocop/pull/11858): Fix `find_end_line` for blocks. ([@reitermarkus][])
```
koic, line 1:
```suggestion
* [#11858](https://github.com/rubocop/rubocop/pull/11858): Fix false positives when using source comments in blocks. ([@reitermarkus][])
```

spec/rubocop/cop/style/symbol_proc_spec.rb
```diff
@@ -293,6 +293,15 @@
...
       RUBY
 
+      expect_no_offenses(<<~RUBY)
+        something
+
+        something do |e|
+          # comment
+          e.upcase
+        end
+      RUBY
```
koic, line 303: Looks good! Can you also add a test for `numblock` somewhere below L350 in this spec file?

koic:
I left two comments. Finally, please squash the commits into one.

koic:
Thanks!

PR #11870, opened 2023-05-12
Fix a false positive for  when method's receiver/argument is not the same as block key argument

### LINKED ISSUE #11868: Incorrect substitution for Style/HashExcept
Given the following line where record is an ActiveRecord object (or a hash with stringified keys, sans `attributes`):
```
record.attributes.reject { |k, _v| k.include?('id') }
```

The `Style/HashExcept` cop detects an offense and suggests a change to:
```
record.attributes.except('id')
```
which is not equivalent.

--------

## Expected behavior

The `Style/HashExcept` cop should not trigger.

## Actual behavior

```
$ bundle exec rubocop ./app/services/example_service.rb
Inspecting 1 file
C

Offenses:

app/services/example_service.rb:8:23: C: [Correctable] Style/HashExcept: Use except('id') instead.
    record.attributes.reject { |k, _v| k.include?('id') }
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
app/services/example_service.rb:12:25: C: [Correctable] Style/HashExcept: Use except('id') instead.
    hash.stringify_keys.reject { |k, _v| k.include?('id') }
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 2 offenses detected, 2 offenses autocorrectable
```

```
class ExampleService
  def call
    Rails.logger.info with_active_record(User.new)
    Rails.logger.info with_a_hash({ example: 'hash', excluded_id: 123_456 })
  end

  def with_active_record(record)
    record.attributes.reject { |k, _v| k.include?('id') }
  end

  def with_a_hash(hash)
    hash.stringify_keys.reject { |k, _v| k.include?('id') }
  end
end

ExampleService.new.call
{"email"=>"", "password"=> ""}
{"example"=>"hash"}
=> nil
```

## RuboCop version

```
$ bundle exec rubocop -V
1.50.2 (using Parser 3.2.2.1, rubocop-ast 1.28.1, running on ruby 3.2.2) [x86_64-darwin22]
  - rubocop-performance 1.17.1
  - rubocop-rails 2.19.1
  - rubocop-rspec 2.22.0
```

fatkodima:
Fixes #11868.

This PR properly checks method's receiver/argument, before considering if this is an offense.


lib/rubocop/cop/style/hash_except.rb
```diff
@@ -89,13 +89,25 @@ def on_send(node)
...
           if active_support_extensions_enabled?
-            bad_method_with_active_support?(block)
+            bad_method_with_active_support?(block) do |key_arg, send_node|
+              if send_node.method?(:in?) && send_node.receiver&.source != key_arg.source
+                return false
+              end
+              return false if (send_node.method?(:include?) || send_node.method?(:exclude?)) &&
+                              send_node.first_argument&.source != key_arg.source
+
+              true
```
koic, line 102: Can you tweak it?

```suggestion
              return true if !send_node.method?(:include?) && !send_node.method?(:exclude?)

              send_node.first_argument&.source != key_arg.source
```

fatkodima: Tweaked.

PR #11872, opened 2023-05-12
Fix Gemspec/DevelopmentDependencies not trigger when add_development_…

Bhacaz:
…dependency has more then one arguments

I was wondering why this cop was not triggered in my gemspec, to finally found that, the node matcher didn't match if an argument (or more) is passed. 

```ruby
# triggered 
spec.add_development_dependency 'foo'

# not triggered
spec.add_development_dependency 'foo', '>= 1.0'
```

There was a [comment](https://github.com/rubocop/rubocop/pull/11469#issuecomment-1403541120) in the PR that introduce this cop which had no response and I didn't found any related issues about this behaviour.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/gemspec/development_dependencies_spec.rb
+++ b/spec/rubocop/cop/gemspec/development_dependencies_spec.rb
@@ -23,6 +23,28 @@ RSpec.describe RuboCop::Cop::Gemspec::DevelopmentDependencies, :config do
       RUBY
     end
 
+    it 'registers an offense when using `#add_development_dependency` in a gemspec a single version argument' do
+      expect_offense(<<~RUBY, 'example.gemspec', preferred_file: enforced_style)
+        Gem::Specification.new do |spec|
+          spec.name = 'example'
+          spec.add_development_dependency 'foo', '>= 1.0'
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Specify development dependencies in %{preferred_file}.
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/gemspec/development_dependencies_spec.rb
```diff
@@ -23,6 +23,28 @@
       RUBY
     end
 
+    it 'registers an offense when using `#add_development_dependency` in a gemspec with version' do
```
koic, line 26:
```suggestion
    it 'registers an offense when using `#add_development_dependency` in a gemspec with a single version argument' do
```

spec/rubocop/cop/gemspec/development_dependencies_spec.rb
```diff
@@ -23,6 +23,28 @@
...
+        Gem::Specification.new do |spec|
+          spec.name = 'example'
+          spec.add_development_dependency 'foo', '>= 1.0'
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Specify development dependencies in %{preferred_file}.
+          spec.add_development_dependency 'allowed', '>= 1.0'
+        end
+      RUBY
+    end
+
+    it 'registers an offense when using `#add_development_dependency` in a gemspec with many arguments' do
```
koic, line 37:
```suggestion
    it 'registers an offense when using `#add_development_dependency` in a gemspec with two version arguments' do
```

koic:
Thanks!

PR #11873, opened 2023-05-12
Add `MinComparisonsCount` config option to `Style/MultipleComparison`

fatkodima:
I want to use `Style/MultipleComparison`, but only when at least 3 comparisons are made. It enforces this style when there are 2. Personally, I do not see much benefit for myself from this in most of the cases. So I made this value configurable, while preserving the original value of 2.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -4383,6 +4383,7 @@ Style/MultipleComparison:
   VersionAdded: '0.49'
   VersionChanged: '1.1'
   AllowMethodComparison: true
+  ComparisonsThreshold: 2
 
 Style/MutableConstant:
   Description: 'Do not assign mutable objects to constants.'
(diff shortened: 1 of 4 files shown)


config/default.yml
```diff
@@ -4374,6 +4374,7 @@ Style/MultipleComparison:
   VersionAdded: '0.49'
   VersionChanged: '1.1'
   AllowMethodComparison: true
+  MinComparisonsCount: 2
```
bbatsov, line 4377: I'd name this "AllowedComparisonsCount" or "ThresholdComparisonsCount". I don't think that "Min" is the best term in the context. Perhaps the word "count" is a bit redundant.

bbatsov: (obviously the default value would be different depending on the exact name chosen)

fatkodima: Changed to `ComparisonsThreshold`. Let me know what you think.

bbatsov:
Looks good to me. Thanks!

PR #11874, opened 2023-05-13
Add new `Style/RedundantRegexpWrapping` cop

koic:
This PR adds new `Style/RedundantRegexpConstructor` cop that checks for the instantiation of regexp using redundant `Regexp.new` or `Regexp.compile`.

```ruby
# bad
Regexp.new(/regexp/)
Regexp.compile(/regexp/)

# good
/regexp/
Regexp.new('regexp')
Regexp.compile('regexp')
```

Autocorrect replaces to regexp literal which is the simplest and fastest. There is speed benefits as well:

```ruby
% cat regexp_new.rb
#!/usr/local/bin/ruby

require 'benchmark/ips'

Benchmark.ips do |x|
  x.report('Regexp.new(/regexp/)') { Regexp.new(/regexp/) }
  x.report('Regexp.new("regexp")') { Regexp.new('regexp') }
  x.report('/regexp/')             { /regexp/ }

  x.compare!
end
```

```console
$ ruby regexp_new.rb
Regexp.new(/regexp/)    96.864k i/100ms
Regexp.new("regexp")    84.848k i/100ms
            /regexp/     2.755M i/100ms
Regexp.new(/regexp/)      1.011M (± 2.1%) i/s -      5.134M in   5.082604s
Regexp.new("regexp")    839.839k (± 3.2%) i/s -      4.242M in   5.057360s
            /regexp/     26.538M (± 9.6%) i/s -    132.263M in   5.055437s

Comparison:
            /regexp/: 26537811.8 i/s
Regexp.new(/regexp/):  1010544.6 i/s - 26.26x  slower
Regexp.new("regexp"):   839838.6 i/s - 31.60x  slower
```

And I think that it can be a general rule, so `Style` is better than `Performance`.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/redundant_regexp_wrapping.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for wrapping regexp literal with redundant `Regexp.new`.
```
bbatsov, line 6: or `Regexp.compile`

koic: Sure, I've updated it.

bbatsov:
I'm wondering if `RedundantRegexpConstructor` might not be a better name for this, although perhaps there are other methods that might work similarly. 

I'm also wondering if we need a similar cop or should extend this cop to discourage the use of `Regexp.new` with string literals and no options, as it's better to just use regexp literals directly. I think we had some similar cops for array and hash literals.

koic:
`RedundantRegexpConstructor` looks like a better name. I updated this PR with it.

> I'm also wondering if we need a similar cop or should extend this cop to discourage the use of `Regexp.new` with string literals and no options, as it's better to just use regexp literals directly.

Sure. This initial implementation accepts `Regexp.new` with string literals and options, but it could be replaced with a `/regexp/i` literal instead of `Regexp.new('regexp', Regexp::IGNORECASE)`. That decision may come later.

>  I think we had some similar cops for array and hash literals.

There was no cops to detect `Array.new([])` and `Hash.new({})`. So, other cops called `RedundantArrayConstructor` and `RedundantHashConstructor` could be implemented as well. I can implement them if you prefer :-)

bbatsov:
> There was no cops to detect Array.new([]) and Hash.new({}). So, other cops called RedundantArrayConstructor and RedundantHashConstructor could be implemented as well. I can implement them if you prefer :-)

That would be great! 

> Sure. This initial implementation accepts Regexp.new with string literals and options, but it could be replaced with a /regexp/i literal instead of Regexp.new('regexp', Regexp::IGNORECASE). That decision may come later.

Yeah, we can polish the new cop down the road to cover such cases as well.

koic:
`Hash.new` doesn't provide similar functionality, so it doesn't seem necessary to implement `RedundantHashConstructor`.
https://docs.ruby-lang.org/en/3.2/Hash.html#method-c-new

FnControlOption:
Thoughts on [`String.new`](https://docs.ruby-lang.org/en/3.2/String.html#method-c-new)?

koic:
It seems like that case would be handled by `Performance/UnfreezeString`.
https://docs.rubocop.org/rubocop-performance/cops_performance.html#performanceunfreezestring

https://github.com/rubocop/rubocop/pull/11886#issuecomment-1555940907

PR #11886, opened 2023-05-17
Add new `Style/RedundantArrayConstructor` cop

koic:
Follow up https://github.com/rubocop/rubocop/pull/11874#issuecomment-1550911537.

Checks for the instantiation of array using redundant `Array` constructor. Autocorrect replaces to array literal which is the simplest and fastest.

```ruby
# bad
Array.new([])
Array[]
Array([])
Array.new(['foo', 'foo', 'foo'])
Array['foo', 'foo', 'foo']
Array(['foo', 'foo', 'foo'])

# good
[]
['foo', 'foo', 'foo']
Array.new(3, 'foo')
Array.new(3) { 'foo' }
```

-----------------

* [x] Added tests.


fatkodima:
I think a similar cop can be created for a string too?

bbatsov:
Btw, for arrays the constructor is redundant also when you're just enumerating some elements, as you can use a literal directly then. Not sure if it's worth covering this here or separately, though.

koic:
> Btw, for arrays the constructor is redundant also when you're just enumerating some elements, as you can use a literal directly then. 

My understanding hasn't quite caught up 💦  Can you provide an example code?

koic:
> I think a similar cop can be created for a string too?

It seems like that case would be handled by `Performance/UnfreezeString`.
https://docs.rubocop.org/rubocop-performance/cops_performance.html#performanceunfreezestring

bbatsov:
@koic I was thinking of:

```ruby
arr = Array[1, 2, 3]
```

Not exactly a classic constructor, but I guess you get the idea.

koic:
Ah I didn't realize that. I think `Array([1, 2, 3])` can be handled similarly. I updated this PR to handle them.

bbatsov:
Looks great now. Thanks!

PR #11898, opened 2023-05-29
Fix offences in calls inside blocks with braces for Style/MethodCallWithArgsParentheses

gsamokovarov:
For the `omit_parentheses` enforced style, I noticed that calls with parentheses in calls **inside** braced blocks were not linted in a code base of mine, so I investigated. It turned out I had a bad detection of the callers that the braced blocks are attached to, they do **need** the parentheses for the call to be syntactically correct. However, the detection leaked into the calls inside the braces that **don't need** the parentheses.

Example:

```ruby
client.images.map { Image.new _1 }
```

```lisp
(numblock
  (send
    (send
      (send nil :client) :images) :map) 1
  (send
    (const nil :Image) :new
    (lvar :_1)))
```

Here, `(send (const nil :Image) :new (lvar :_1)))`, was considered a **call with braced block**, because the detection used to check for the parent node being a block node. This is not correct. I have used the mechanism defined in `RuboCop::AST::MethodDispatchNode#block_mode` to properly detect only the send node the block is attached to and not other sends in the body.

### DIFF
--- a/spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
+++ b/spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
@@ -695,6 +695,28 @@ RSpec.describe RuboCop::Cop::Style::MethodCallWithArgsParentheses, :config do
       RUBY
     end
 
+    it 'register an offense in calls inside braced blocks' do
+      expect_offense(<<~RUBY)
+        client.images(page: page) { |resource| Image.new(resource) }
+                                                        ^^^^^^^^^^ Omit parentheses for method calls with arguments.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        client.images(page: page) { |resource| Image.new resource }
+      RUBY
+    end
+
+    it 'register an offense in calls inside braced numblocks', :ruby27 do
+      expect_offense(<<~RUBY)
+        client.images(page: page) { Image.new(_1) }
+                                             ^^^^ Omit parentheses for method calls with arguments.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        client.images(page: page) { Image.new _1 }
+      RUBY
+    end
+
     it 'accepts parens in single-line inheritance' do
       expect_no_offenses(<<-RUBY)
         class Point < Struct.new(:x, :y); end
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
```diff
@@ -695,6 +695,20 @@ class Point < Struct.new :x, :y
       RUBY
     end
 
+    it 'register an offense in calls inside braced blocks' do
+      expect_offense(<<~RUBY)
+        client.images(page: page) { |resource| Image.new(resource) }
+                                                        ^^^^^^^^^^ Omit parentheses for method calls with arguments.
+      RUBY
```
koic, line 702: Can you add `expect_correction`?

```suggestion
      RUBY
      
      expect_correction(<<~RUBY)
        client.images(page: page) { |resource| Image.new resource }
      RUBY
```

spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
```diff
@@ -695,6 +695,20 @@ class Point < Struct.new :x, :y
...
+    end
+
+    it 'register an offense in calls inside braced numblocks', :ruby27 do
+      expect_offense(<<~RUBY)
+        client.images(page: page) { Image.new(_1) }
+                                             ^^^^ Omit parentheses for method calls with arguments.
+      RUBY
```
koic, line 709: Ditto :-)

```suggestion
      RUBY
      
      expect_correction(<<~RUBY)
        client.images(page: page) { Image.new _1 }
      RUBY
```

spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
gsamokovarov: Sure, will add it.

spec/rubocop/cop/style/method_call_with_args_parentheses_spec.rb
gsamokovarov: I have amended the commit

gsamokovarov:
Added the corrections to the tests.

koic:
Thanks!

PR #11901, opened 2023-05-29
[Fix #11899] Fix an incorrect autocorrect for `Style/SingleLineMethods`

### LINKED ISSUE #11899: Style/SingleLineMethods should not correct to endless methods when Style/EndlessMethod is disabled
**This is just a question and came up when researching https://github.com/standardrb/standard/issues/489**

I can't help but wonder if the guard clause's boolean logic in [this method](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/style/single_line_methods.rb#L138) in `Style/SingleLineMethods` isn't reversed.

This method is used in determining if a single-line method should be converted to an endless method:

```ruby
def disallow_endless_method_style?
  endless_method_config = config.for_cop('Style/EndlessMethod')
  return false unless endless_method_config['Enabled']
  
  endless_method_config['EnforcedStyle'] == 'disallow'
end
```

Converted to English and with the double-negatives unwound, the guard clause seems to be saying "If Style/EndlessMethod is disabled, then *allow* violations of this cop to be converted to Endless methods".  Am I reading that wrong, or is that logically backwards? If someone has disabled `Style/EndlessMethod`, shouldn't endless method corrections by this cop _be disallowed_?

I expected the line to read:

```ruby
  return true unless endless_method_config['Enabled']
```

Do I have this right? Just a quick spot-check, as I may be totally wrong here—it's a lot of boolean negation to keep straight, and I'm not sure how the two cops interact

koic:
Fixes #11899.

This PR fixes an incorrect autocorrect for `Style/SingleLineMethods` when using Ruby 3.0 and `Style/EndlessMethod` is disabled.

This is an overlooked bug due to the lack of `:ruby30` in the target test code.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_an_incorrect_autocorrect_for_style_single_line_methods.md
+++ b/changelog/fix_an_incorrect_autocorrect_for_style_single_line_methods.md
@@ -1,0 +1,1 @@
+* [#11899](https://github.com/rubocop/rubocop/issues/11899): Fix an incorrect autocorrect for `Style/SingleLineMethods` when using Ruby 3.0 and `Style/EndlessMethod` is disabled. ([@koic][])
--- a/lib/rubocop/cop/style/single_line_methods.rb
+++ b/lib/rubocop/cop/style/single_line_methods.rb
@@ -135,7 +135,7 @@ module RuboCop
 
         def disallow_endless_method_style?
           endless_method_config = config.for_cop('Style/EndlessMethod')
-          return false unless endless_method_config['Enabled']
+          return true unless endless_method_config['Enabled']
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Hmm, why isn't this change affecting the test cases?

koic:
It affects the following test case:

```diff
-  context 'when `Style/EndlessMethod` is disabled' do
+  context 'when `Style/EndlessMethod` is disabled', :ruby30 do
```

Endless method definition was introduced in Ruby 3.0 and the test was missing `:ruby30`.

bbatsov:
Thanks!

PR #11904, opened 2023-05-30
Mark `Layout/ClassStructure` as unsafe to autocorrect

nevans:
Autocorrection is unsafe because class instance methods and module inclusion may behave dynamically, based on which methods or constants have been defined.  ~~Constant definitions may also depend upon pre-existing methods or included modules.~~ _Only literal constants are autocorrected: #7808._

### Alternative approach:

`ClassStructure` could be updated to make autocorrection behave differently for `-a` vs `-A`, i.e: method `def`s can be safely re-ordered if all "macros" are anchored to their current position.  But that would be a more complicated PR.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/layout/class_structure.rb
+++ b/lib/rubocop/cop/layout/class_structure.rb
@@ -68,6 +68,13 @@ module RuboCop
       #        - extend
       # ----
       #
+      # @safety
+      #   Autocorrection is unsafe because class methods and module inclusion
+      #   can behave differently, based on which methods or constants have
+      #   already been defined.
+      #
+      #   Constants will only be moved when they are assigned with literals.
+      #
       # @example
       #   # bad
       #   # Expect extend be before constant
--- a/changelog/change_layout_class_structure_autorrect_safety.md
+++ b/changelog/change_layout_class_structure_autorrect_safety.md
@@ -1,0 +1,1 @@
+* [#11904](https://github.com/rubocop/rubocop/pull/11904): Mark `Layout/ClassStructure` as unsafe to autocorrect. ([@nevans][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -467,7 +467,9 @@ Layout/ClassStructure:
   Description: 'Enforces a configured order of definitions within a class body.'
   StyleGuide: '#consistent-classes'
   Enabled: false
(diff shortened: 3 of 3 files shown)


lib/rubocop/cop/layout/class_structure.rb
```diff
@@ -68,6 +68,12 @@ module Layout
       #        - extend
       # ----
       #
+      # @safety
+      #   Autocorrection is unsafe because class instance methods and module
+      #   inclusion may behave dynamically, based on which methods or constants
```
bbatsov, line 73: I don't understand "may behave dynamically" here. I guess you meant to write "differently".

bbatsov:
> ClassStructure could be updated to make autocorrection behave differently for -a vs -A, i.e: method defs can be safely re-ordered if all "macros" are anchored to their current position. But that would be a more complicated PR.

I'm fine with the simpler approach. I guess you can also leave some separate ticket for additional improvements that we can do in the future. This cop needs a lot of love.

lib/rubocop/cop/layout/class_structure.rb
nevans: I think that I did intend to write "dynamic".  But "different" is both simpler and communicates everything I wanted to communicate.  I'll change it to that.  Thanks!

By "dynamic", I was thinking of `Module` callbacks (`included`, `extended`, etc) which do anything they want to do, behaving differently based on context, as opposed to the default "static" inclusion behavior which is the same thing regardless of what methods or constants are already defined on the class.

bbatsov: My main issue with the term is that "dynamic" is quite overloaded in Ruby (e.g. for types, evaluation, etc), so I'd rather go with something less ambiguous here.

nevans:
@bbatsov Thanks for the feedback.  I updated the text.  Let me know what you think.

> This cop needs a lot of love.

I wouldn't know to make the PR and issues if I didn't already love it, at least a little bit.  😉  Hopefully I'll make some time in the future to learn rubocop internals well enough to "love" it with some more!  😄

bbatsov:
@nevans You can take a look at this unfinished PR that included many improvements https://github.com/rubocop/rubocop/pull/9575

PR #11923, opened 2023-06-03
New cop to find return statements inside memoization blocks

catlee:
A frequent source of caching bugs is returning from inside memoization blocks. This can happen when code is refactored, or when someone unfamiliar with how Ruby `||=` and `begin ... end` interact writes memoization code.

For example, in

```ruby
def foo
  @foo ||= begin
    return qux if baz?
    ruz
  end
end
```

The intent is to return (and cache) `qux` when `baz?` is true. In reality, if `baz?` is true, then `qux` is returned from the method, bypassing the conditional assignment operator. `@foo` is never assigned a value in this case, and the next time the method is called, `baz?` will be reevaluated.

This PR adds a new cop to find `return` statements inside these kinds of memoization blocks.

Returning a value, or returning `nil` (as in `return if/unless baz?`) are equally problematic. In both cases, the value is returned from the current method, bypassing the instance variable assignment.

-----------------

* [x] Added tests.


vlad-pisanov:
I think this case is covered by [Lint/NoReturnInBeginEndBlocks](https://docs.rubocop.org/rubocop/cops_lint.html#lintnoreturninbeginendblocks)

Related: proposed cop [Lint/PredicateMemoization](https://github.com/rubocop/rubocop-performance/issues/326)

catlee:
> I think this case is covered by [Lint/NoReturnInBeginEndBlocks](https://docs.rubocop.org/rubocop/cops_lint.html#lintnoreturninbeginendblocks)

Ah, perfect, I missed that! Thank you!

PR #11925, opened 2023-06-04
[Fix #11924] Add new `Style/RedundantCurrentDirectoryPath` cop

### LINKED ISSUE #11924: require_relative with "same folder" prefix "./"

I sometimes see people use e.g. `require_relative "./foo"` where the shorter, cleaner `require_relative "foo"` would suffice. The same goes for e/g/ `require_relative "./foo/bar/baz"` vs `require_relative "foo/bar/baz"`. I see no reason for ever using the `./` prefix when calling `require_relative`.

I think we should have a cop that disallows the `./` prefix (but of course allows `../`) on `require_relative`.

—

## Additional context

The `./` prefix has its merit when calling shell commands, and you want to distinguish e.g. `./rubocop` from a `rubocop` found in your `$PATH`.

In Ruby we have `require` finding code in `$LOAD_PATH`, whereas `require_relative` only loads code in a path relative to the requiring file’s path. The `./` is superfluous.

koic:
Resolves #11924.

This PR adds new `Style/RedundantCurrentDirectoryPath` cop. It checks for uses a redundant current directory path in `require_relative`.

```ruby
# bad
require_relative './path/to/feature'

# good
require_relative 'path/to/feature'
```

The cop name `Style/RedundantCurrentDirectoryPath` does not contain `RequireRelative` (`require_relative`). Because I thought it could be an abstract name that detects redundant current directory path (`./`). So, if there are methods other than `require_relative` that certainly uses redundant current directory path, this cop could be extended.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/redundant_current_directory_path.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for uses a redundant current directory path in `require_relative`.
```
bbatsov, line 6: Maybe we can make this a configurable list of method names from the beginning?

koic: Good point. This bothered me. Published configuration options must be stable and cannot be renamed later.
So I thought I could expose it as a configuration option when needed.

Of course it can also be made configurable from the beginning. Do you prefer it?

bbatsov: Seems to me a name like `TargetMethods` (or similar) would work fairly well. I made my remark just because you said you wanted to keep the name of the cop generic in case more methods are added down the road. I assume that there are quite a few methods that operate on the current directory (but we'll have to double check this), so this makes sense to me. 

Anyways, not a big deal for me - just a small point of consideration.

koic: `TargetMethods` looks good to me! On the other hand, I thought about it again. The redundant `./` location might be not always the first argument, like `require_relative './path/to/feature`.

So I thought it would be better if the configuration will be extended based on the feedback. What do you think?

bbatsov: Sure.

bbatsov:
Left a bit of small feedback about the naming & description and we can merge this when it is addressed.

Issue #11099, opened 2022-10-21
Lint/RedundantRequireStatement should not remove requiring `pp`


mvz:
Lint/RedundantRequireStatement autocorrects `require "pp"` by removing it. However, `pp` is not loaded by default:

```bash
$ ruby -e 'puts $LOADED_FEATURES' | grep pp
/home/matijs/.rbenv/versions/3.1.2/lib/ruby/3.1.0/rubygems/path_support.rb
$ ruby --version
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]
```

## Expected behavior

RuboCop leaves `require "pp"` in place.

## Actual behavior

RuboCop removes `require "pp"`.

## Steps to reproduce the problem

- Create a file `foo.rb` containing

```
require "pp"
```

- Run `rubocop -a foo.rb --only Lint/RedundantRequireStatement`.

## RuboCop version

```
$ rubocop -V
1.37.0 (using Parser 3.1.2.1, rubocop-ast 1.22.0, running on ruby 3.1.2) [x86_64-linux]
```

koic:
This is a quote from the official release article of Ruby 2.5.

> One of our most loved libraries, pp.rb, is now automatically loaded. You no longer have to write require "pp". [[Feature #14123]](https://bugs.ruby-lang.org/issues/14123)

https://www.ruby-lang.org/en/news/2017/12/25/ruby-2-5-0-released/

So, `require 'pp'` seems unnecessary. Are there any use cases where a problem occurs?

mvz:
Oh, I see it auto-requires `pp` when needed. That explains why it's not in the list of loaded features.

The problem I'm facing is failing tests in this pull request: https://github.com/mvz/live_ast/pull/70

However, I think this actually exposed an issue in that gem: It tries to open the file where `pp` is defined, and fails because it's defined in the prelude.

So this is not really a good reason to remove `pp` from the list of redundant requires.

bobf:
@koic I don't know if my use case justifies removing this cop but please see the following:

```console
$ ruby -e 'require "stringio"; io = StringIO.new; pp("foo", io); puts io.string'
"foo"                                                       
#<StringIO:0x00007f2f25e0b338>   
```

```console
$ ruby -e 'require "stringio"; io = StringIO.new; PP.pp("foo", io); puts io.string'
-e:1:in `<main>': uninitialized constant PP (NameError)
```

```console
[...]

PR #11931, opened 2023-06-06
Fix a false positive for `Lint/RedundantRequireStatement`

koic:
Follow up https://github.com/rubocop/rubocop/issues/11099#issuecomment-1577117258

This PR fixes a false positive for `Lint/RedundantRequireStatement` when using `PP.pp`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_positive_for_lint_redundant_require_statement.md
+++ b/changelog/fix_a_false_positive_for_lint_redundant_require_statement.md
@@ -1,0 +1,1 @@
+* [#11931](https://github.com/rubocop/rubocop/pull/11931): Fix a false positive for `Lint/RedundantRequireStatement` when using `PP.pp`. ([@koic][])
--- a/lib/rubocop/cop/lint/redundant_require_statement.rb
+++ b/lib/rubocop/cop/lint/redundant_require_statement.rb
@@ -49,6 +49,11 @@ module RuboCop
             (str #redundant_feature?))
         PATTERN
 
+        # @!method pp_const?(node)
+        def_node_matcher :pp_const?, <<~PATTERN
+          (const {nil? cbase} :PP)
+        PATTERN
+
         def on_send(node)
           return unless redundant_require_statement?(node)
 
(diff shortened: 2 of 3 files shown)


bobf:
@koic Tried your branch on my project, works when `require 'pp'` is in the same file as `PP.pp` usage, but in my case I require all dependencies for my gem in the gem's root (i.e. `lib/my_gem.rb`, and `PP.pp` is called in `lib/my_gem/my_module/my_class.rb`).

Is this intended behaviour ? I prefer to keep all `require` statements in the same place when building a gem, since all namespace will be impacted regardless of where the `require` call happens.

bbatsov:
@bobf Hmm, that's a bit tricky for that particular cop. Perhaps it's just best to stick to the disable and we can ignore this PR together as it only adds complexity for little added value. I'd suggest enhancing the cop documentation to highlight better its limitations instead. 

@koic Any thoughts?

bbatsov:
I'm leaning towards just ignoring the additional problem, as it's outside of RuboCop's scope. I'm wondering if we should just ignore the special handing for `PP.pp` altogether.

bobf:
@bbatsov It's no bother for me to add a `rubocop:disable` directive on this one use case, it's only the second time I've ever deliberately required `pp` so I can definitely live with this being an exceptional case. If anything it clarifies that I'm doing it deliberately so somebody reading the code might assume that I'm doing something slightly unusual.

koic:
For now, the implementation of this PR appears simple and seems capable of preventing false positives for `PP.pp` within the same file.

>  I'd suggest enhancing the cop documentation to highlight better its limitations instead.

Yes, I can certainly enhance the documentation. However, I'm wondering if it would be better to set `AutoCorrect: false` along with mentioning it in the documentation.

bbatsov:
> Yes, I can certainly enhance the documentation. However, I'm wondering if it would be better to set AutoCorrect: false along with mentioning it in the documentation.

Not a bad idea.

bbatsov:
I'll merge this in its current state, as I'm about the cut a release. The rest of the stuff we discussed can happen later.

PR #11940, opened 2023-06-08
[Fix #8506] Add AllowedParentClasses config to Lint/MissingSuper.

### LINKED ISSUE #8506: Lint/MissingSuper false positive
There are cases where I'm explicitly avoiding calling `super` in an initializer, however this cop complains on all of them. This doesn't seem very useful.

iMacTia:
Add `AllowedParentClasses ` config to `Lint/MissingSuper`.

The default value of `%w[BasicObject Object]` is kept for backwards compatibility, but users can now add to that list using the `AllowedParentClasses ` config value.

It's important to note that the value of `AllowedParentClasses ` will be ADDED to the default value.
This is because I don't foresee scenarios where users would want to remove `BasicObject` or `Object` from this list, and removes the requirement for always specifying them.
I'm open to feedback on this assumption and happy to change it into a full replacement, I simply went with the most logical approach in my opinion.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/missing_super.rb
+++ b/lib/rubocop/cop/lint/missing_super.rb
@@ -14,6 +14,13 @@ module RuboCop
       # Autocorrection is not supported because the position of `super` cannot be
       # determined automatically.
       #
+      # `Object` and `BasicObject` are allowed by this cop because of their
+      # stateless nature. However, sometimes you might want to allow other parent
+      # classes from this cop, for example in the case of an abstract class that is
+      # not meant to be called with `super`. In those cases, you can use the
+      # `AllowedParentClasses` option to specify which classes should be allowed
+      # *in addition to* `Object` and `BasicObject`.
+      #
       # @example
       #   # bad
       #   class Employee < Person
@@ -56,6 +63,21 @@ module RuboCop
       #   class Parent
       #     def self.inherited(base)
       #       super
+      #       do_something
+      #     end
+      #   end
+      #
+      #   # good
+      #   class ClassWithNoParent
+      #     def initialize
+      #       do_something
+      #     end
+      #   end
+      #
+      # @example AllowedParentClasses: [MyAbstractClass]
+      #   # good
+      #   class MyConcreteClass < MyAbstractClass
+      #     def initialize
       #       do_something
       #     end
       #   end
@@ -116,16 +138,20 @@ module RuboCop
           if (block_node = node.each_ancestor(:block, :numblock).first)
             return false unless (super_class = class_new_block(block_node))
 
-            !stateless_class?(super_class)
+            !allowed_class?(super_class)
           elsif (class_node = node.each_ancestor(:class).first)
-            class_node.parent_class && !stateless_class?(class_node.parent_class)
+            class_node.parent_class && !allowed_class?(class_node.parent_class)
           else
             false
           end
         end
 
-        def stateless_class?(node)
-          STATELESS_CLASSES.include?(node.const_name)
+        def allowed_class?(node)
+          allowed_classes.include?(node.const_name)
+        end
+
+        def allowed_classes
+          @allowed_classes ||= STATELESS_CLASSES + cop_config.fetch('AllowedParentClasses', [])
         end
       end
     end
--- a/changelog/change_merge_pull_request_11936_from.md
+++ b/changelog/change_merge_pull_request_11936_from.md
@@ -1,0 +1,1 @@
+* [#8506](https://github.com/rubocop/rubocop/issues/8506): Add `AllowedParentClasses` config to `Lint/MissingSuper`. ([@iMacTia][])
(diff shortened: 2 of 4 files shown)


Drenmi:
Change looks good, but perhaps we can generalize the setting name to divorce it from one particular reason for ignoring a node. E.g. `ExcludedParentClasses` or similar? 🙂

bbatsov:
Agreed - the name should be more generic. Also this settings needs to be documented and some example usages should be added to the docs.

iMacTia:
> Change looks good, but perhaps we can generalize the setting name

That makes sense, `ExcludedParentClasses` conveys the intent rather than the reason, I'll change it to that 👍 

> Also this settings needs to be documented and some example usages should be added to the docs

I wasn't sure about this, I'll add something. Thanks for pointing it out

iMacTia:
Updated and force-pushed to keep the commit history clean.
@Drenmi @bbatsov this should now be ready for a second pass 👍

docs/modules/ROOT/pages/cops_lint.adoc
```diff
@@ -3295,6 +3295,13 @@ challenging or verbose for no actual gain.
 Autocorrection is not supported because the position of `super` cannot be
 determined automatically.
 
+`Object` and `BasicObject` are excluded from this cop because of their
+stateless nature. However, sometimes you might want to exclude other parent
+classes from this cop, for example in the case of an abstract class that is
+not meant to be called with `super`. In those cases, you can use the
+`ExcludedParentClasses` option to specify which classes should be excluced
+*in addition to* `Object` and `BasicObject`.
```
koic, line 3303: This document and examples are auto-generated based on the source code. Please update lib/rubocop/cop/lint/missing_super.rb directly.
There's no need to update the docs/modules/ROOT/pages/cops_lint.adoc doc as it will be generated at the time of release.

iMacTia: Thank you @koic, I've amended the commit to keep the history clean.
I hope I did it right 😄

koic (approved):
Looks good to me 👍

lib/rubocop/cop/lint/missing_super.rb
```diff
@@ -14,6 +14,13 @@ module Lint
       # Autocorrection is not supported because the position of `super` cannot be
       # determined automatically.
       #
+      # `Object` and `BasicObject` are excluded from this cop because of their
```
bbatsov, line 17: excluded from -> allowed by

bbatsov:
Other than my remark about the naming the changes look good as well. I'll merge the PR once the name is updated.

lib/rubocop/cop/lint/missing_super.rb
```diff
@@ -60,6 +67,22 @@ module Lint
       #     end
       #   end
       #
+      # @example ExcludedParentClasses: [MyAbstractClass]
```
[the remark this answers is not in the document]
koic: I agree the new suggested name :-)

iMacTia:
Thank you @bbatsov, the new suggested name makes sense.
I've updated the branch (amended to keep commit history clean), as well as the PR title and description

lib/rubocop/cop/lint/missing_super.rb
```diff
@@ -60,6 +67,22 @@ module Lint
...
+      #       # do_something without calling super
+      #     end
+      #   end
+      #
+      #   class ClassWithNoParent
+      #     def initialize
+      #       # there's no need to specify Object or BasicObject
+      #       # in the AllowedParentClasses option.
+      #     end
+      #   end
```
koic, line 84: The example `ClassWithNoParent` class is irrelevant to the `AllowedParentClasses` configuration and is always allowed, so it looks better to put it in the `# good` case of `@example`, right?

iMacTia: Of course, I can move it there, but I wanted to make it absolutely clear that no matter what the value of `AllowedParentClasses` is, `Object` and `BasicObject` will always be allowed, so this example is just to remark that one more time in case the reader misses it from the description above. Although it can technically go in both places, I feel like it's more useful here. WDYT?

koic: Yeah, I understand the intention. But `@example ConfigurationName` is typically used to describe unique behaviors of `ConfigurationName`. For the sake of consistent documentation structure, it would be better to move this to `@example` I think :-)

iMacTia: Can't really argue with that, I understand the value of consistency 🙂 
I'll move this up as suggested 👍

iMacTia:
I've got a "rubocop server is not running" error while running specs on 2.7, but that might just be a flaky?
I can't re-run the test though, so I'll need your help 🙏

PR #11967, opened 2023-06-21
Add new `Style/ReturnNilInPredicateMethodDefinition` cop

koic:
This PR adds new `Style/ReturnNilInPredicateMethodDefinition` cop that checks for `return` or `return nil` is used in predicate method definitions.

```ruby
# bad
def foo?
  return if condition

  do_something?
end

# bad
def foo?
  return nil if condition

  do_something?
end

# good
def foo?
  return false if condition

  do_something?
end
```

Autocorrection is marked as unsafe because the change of the return value from `nil` to `false` could potentially lead to incompatibility issues.

-----------------

* [x] Added tests.


bbatsov:
Hmm, seems we don't have a style guide entry for the expected return values from predicate methods (ideally `true` or `false`).

As for the cop - overall it looks good. I'm just wondering if we should tackle both return values in the same cop(e.g. flag returns of non-true truthy values as well) or maybe make some other cop for the truthiness case. At any rate - it's clear that returning `nil` is the bigger problem. 

One day we should have some cop that discourages the return of `nil` completely.

koic:
Good point. However, unlike falsey values which are limited to `nil` and `false`, determining whether truthy values are `true` seems quite challenging.

Just as RuboCop wouldn't know whether the following `value` is boolean or not. It also can't determine if it's `true` or truthy.

```console
return value if condition
```

It can't know whether `value` is already boolean or whether `!!value` is necessary.

On the other hand, The following can be confidently changed to `return false`. It is the job of this cop.

```ruby
return if condition
return nil if condition
```

Theoretically, it can be added a warning for cases where non-`nil` and non-`false` literals are used (e.g. `return 42 if condition`), it can be replaced `return true if condition`. However, this usage is quite rare 😅 , so I'm not certain whether it would add worth.

To sum up, detecting expected returning `true` value seems to be extremely difficult, unlike expected detecting returning `false` value.

As the style guide, it might be covered (or can be extended) with the following:
https://rubystyle.guide/#bool-methods-qmark

bbatsov:
> Just as RuboCop wouldn't know whether the following value is boolean or not. It also can't determine if it's true or truthy.

I was just thinking we can imply some simple heuristics - if something is not `true` or a call to some predicate we assume that it's not a boolean. It would yield a lot of false positives for sure and it can't be enabled by default, but it still may be useful to some extent. Anyways, it's not something particularly important and can happen down the road.

AlexWayfer:
Predicate methods return `nil` in stdlib.

This cop is breaking the logic in my library: when `valid?` should not be executed under some conditions and `nil` just a quick return with signal there were no check. I can't tell if it's valid or not, but it's falsey.

Examples from stdlib:

* https://rubyapi.org/3.2/o/string#method-i-casecmp-3F
* https://rubyapi.org/3.2/o/kernel#method-i-autoload-3F

Maybe something else.

klyonrad:
Yeah I also had the instant feeling that this is very opinionated - and I like that style. Had a bunch of colleagues who wouldn't care

PR #11968, opened 2023-06-22
Don't adjust Includes if base and derived configs are in the same folder

naveg:
This fixes an issue caused by #11416, where files no longer match when
rubocop is run from a child directory.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_don_t_adjust_includes_if_base_and_derived_configs.md
+++ b/changelog/fix_don_t_adjust_includes_if_base_and_derived_configs.md
@@ -1,0 +1,1 @@
+* [#11968](https://github.com/rubocop/rubocop/pull/11968): Don't adjust Includes if base and derived configs are in the same folder. ([@naveg][])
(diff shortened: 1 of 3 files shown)


alexevanczuk (commented):
This makes sense to me @naveg . 

I'd also be fine if we wanted to revert the original solution. I realized I really need an *extends* directive that doesn't change path semantics at all, since child rubocop files that inherit from a parent still suffer from the issues that it changes all paths to need to be relative to the child config, when we want paths to not change.

naveg:
Yeah, this change kind of feels like a hacky patch, but I do worry that the use of `Dir.pwd` will break other cases beyond this one.

Using absolute paths everywhere does seem like the best fix, but that's a much more involved change.

If it turns out that your original change doesn't actually meet your use case, maybe we should just revert that one instead. @bbatsov @jonas054 any thoughts?

bbatsov:
@naveg I'd be fine to revert the previous change if it didn't improve anything.

PR #11983, opened 2023-06-26
Add Ridgepole files to default `Include` list

ydah:
With ridgepole, we can write Ruby DSLs in files called `Schemafile` or `.schema`. 
https://github.com/ridgepole/ridgepole

For example:
```ruby
create_table "users", force: :cascade do |t|
  t.string   "last_name"
  t.string   "first_name"
  t.virtual  "full_name", type: :string, as: "concat(`last_name`,' ',`first_name`)", stored: true
end
```

-----------------

* [x] Added tests.


changelog/change_add_schema_and_schemafile_to_default_include_list.md
```diff
@@ -0,0 +1 @@
+* [#11983](https://github.com/rubocop/rubocop/pull/11983): Add `*.schema` and `Schemafile` to default `Include` list. ([@ydah][])
```
koic, line 1: It might be user-friendly to mention that these are Ridgepole files.

config/default.yml
```diff
@@ -30,6 +30,7 @@ AllCops:
     - '**/*.rbx'
     - '**/*.ru'
     - '**/*.ruby'
+    - '**/*.schema'
```
koic, line 33: Note, I'm worried about triggering parse errors when the generic extension name `*.schema` is used in products other than Ridgepole. Expect no conflicts.

changelog/change_add_schema_and_schemafile_to_default_include_list.md
ydah: I updated this PR.

config/default.yml
ydah: Indeed it is. I searched on GitHub, but I don't see anything other than Ridgepole that uses `*.schema` that we would use in the same products. Please point out if I have missed something.
https://github.com/search?q=path%3A*.schema&type=code

bbatsov:
Thanks!

PR #11985, opened 2023-06-26
[Fix #11984] Add configurable server idle timeout

### LINKED ISSUE #11984: Rubocop server auto-shutdown if idle

I run Rubocop in git commit hooks, and I the server feature has been a very welcome addition since I no longer have to wait several seconds on every commit attempt.
However, after I'm done with a task I don't like having leftover background processes idling, and it's easy to forget to kill the running Rubocop server after I'm done doing Ruby things for the day.

I would love to have access to a CLI option or environment variable that Rubocop reads when starting a server and which configures the server to automatically shut down if it doesn't receive any requests for a certain amount of time.

For example, `--idle-timeout <seconds>` or `RUBOCOP_SERVER_IDLE_TIMEOUT`.

I considered writing my own wrapper around Rubocop to handle this use case, but it's impractical to implement from outside of the server process.

Palladinium:
This adds support for a new environment variable when running rubocop in server mode, `RUBOCOP_SERVER_IDLE_TIMEOUT`.

When the variable is set, it's interpreted as a number of seconds, and the server will automatically exit if that many seconds pass without receiving any connections.

I'm hesitant to add tests since they'll likely be timing-sensitive (read: inconsistent) and/or might slow down the suite with `sleep`s.

-----------------

* [ ] Added tests.


bbatsov:
I'll defer to @koic on this one, but I'll mention that it might be nice to be able to set the timeout without with some command-line flag as well. (it would take precedence over the env var)

koic:
> I'm hesitant to add tests since they'll likely be timing-sensitive (read: inconsistent) and/or might slow down the suite with `sleep`s.

It looks good to have a test that uses stubs and mocks. Especially the behavior around the `IO::WaitReadable` exception handling.

lib/rubocop/server/core.rb
```diff
@@ -71,15 +85,18 @@ def server_mode?
...
       end
 
-      def start_server(host, port)
-        @server = TCPServer.open(host, port)
-
+      def output_stream
         # JSON format does not expected output message when IDE integration with server mode.
         # See: https://github.com/rubocop/rubocop/issues/11164
-        return if use_json_format?
+        return nil if use_json_format?
```
koic, line 91: Is this change intentional?

Palladinium: I think it was just force of habit, I forgot the two are equivalent. I'll revert it since the older line is more concise.

bbatsov:

PR #12003, opened 2023-06-29
Fix code length calculator for method calls with heredoc

### LINKED ISSUE #12001: Metrics/MethodLength heredoc and method_call not completely compatible with each other?
## Expected behavior

Should count heredoc lines as one when using the following configuration:

```yaml
Metrics/MethodLength:
  CountAsOne: ['heredoc', 'method_call']
```

## Actual behavior

With some specific conditions, it counts each line in heredoc within a method.

Conditions are: Don't use parentheses on a method call that uses heredoc and call something else after that method call
 
```sh
For /home/x/test_proj: configuration from /home/x/test_proj/.rubocop.yml
Default configuration from /home/x/.local/share/gem/ruby/3.2.0/gems/rubocop-1.53.1/config/default.yml
Use parallel by default.
Inspecting 1 file
Scanning /home/x/test_proj/test.rb
C

Offenses:

test.rb:5:3: C: Metrics/MethodLength: Method has too many lines. [11/10]
  def test_method ...
  ^^^^^^^^^^^^^^^

Finished in 0.2313520830066409 seconds
```

Configure `rubocop.yml`

```yaml
AllCops:
  NewCops: enable
  TargetRubyVersion: 3.1

Metrics/MethodLength:
  CountAsOne: ['heredoc', 'method_call']
```

Create a file `test.rb`

```
# frozen_string_literal: true

# Test class
class Test
  def test_method
    other_method <<~ANY
      2
      3
      4
      5
      6
      7
      8
      9
    ANY
    puts '11'
  end
end
```

Test the file
```sh
rubocop test.rb
```

## RuboCop version

```
[test_proj]$ rubocop -V
1.53.1 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.2.2) [x86_64-linux]
```

fatkodima:
Fixes #12001.

The  problem is that for the code like 
```ruby
other_method <<~ANY
  2
  3
ANY
```

the `.source` method returns just the first line (`other_method <<~ANY`), and so for the example from the linked issue, the total method length included all the lines, but when considering method calls as a single line (`CountAsOne` option), only the first line of the method call was considered and subtracted from the total, leaving all the rest lines from heredoc.

This PR fixes the `CodeLengthCalculator` to correctly calculate sources when the node includes a heredoc in its body.

### DIFF
--- a/changelog/fix_code_length_calculator_for_method_calls_with_heredoc.md
+++ b/changelog/fix_code_length_calculator_for_method_calls_with_heredoc.md
@@ -1,0 +1,1 @@
+* [#12001](https://github.com/rubocop/rubocop/issues/12001): Fix code length calculator for method calls with heredoc. ([@fatkodima][])
--- a/lib/rubocop/cop/metrics/utils/code_length_calculator.rb
+++ b/lib/rubocop/cop/metrics/utils/code_length_calculator.rb
@@ -63,7 +63,7 @@ module RuboCop
             types
           end
 
-          def code_length(node)
+          def code_length(node) # rubocop:disable Metrics/MethodLength
             if classlike_node?(node)
               classlike_code_length(node)
             elsif heredoc_node?(node)
@@ -72,7 +72,14 @@ module RuboCop
               body = extract_body(node)
               return 0 unless body
 
-              body.source.each_line.count { |line| !irrelevant_line?(line) }
+              source =
+                if node_with_heredoc?(body)
+                  source_from_node_with_heredoc(body)
+                else
+                  body.source.lines
+                end
+
+              source.count { |line| !irrelevant_line?(line) }
             end
           end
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Looks good. Thanks!

mobilutz:
@fatkodima 
This change leads to false positives for a code I am maintaining.

This pseudo code, raises a `Metrics/MethodLength` error:
```ruby
def sql
  <<-SQL.squish
  DO $$
  BEGIN
    WHILE (
      SELECT
      count(*)
      FROM
      posts
    ) > 0
  LOOP
      DELETE 
      FROM 
      posts 
      LIMIT 
      10;
  END LOOP;
  END $$;
  SQL
end
```

It seems that the problem is the `DO`, which is also a keyword for some SQL databases.
Example PostgreSQL https://www.postgresql.org/docs/current/sql-do.html

fatkodima:
That heredoc is parsed just as a string. I am not able to reproduce. Can you share your configuration for `Metrics/MethodLength` cop?

mobilutz:
I will try to create a gist to make it work. I don't want to share the complete projects configuration.

mobilutz:
@fatkodima 
Ah, we forgot to add `'heredoc'` to the `CountAsOne` option for this cop.

This now just shows that we need to add it.

PS: I will update my first comment with the error.

PR #12016, opened 2023-07-03
[Fix #12015] Mark `Style/HashConversion` as unsafe autocorrection

### LINKED ISSUE #12015: Style/HashConversion unsafe autofix
I've come across a situation where I was using `Hash[...]` and [`Style/HashConversion`](https://docs.rubocop.org/rubocop/cops_style.html#stylehashconversion) suggested a change to `.to_h` which caused an exception.

There is a subtle difference between `Hash[]` and `Array#to_h` when there is an odd number of elements:

```rb
> Hash[[[1,2],[3,4],[5]]]
=> {1=>2, 3=>4, 5=>nil}
> [[1,2],[3,4],[5]].to_h
ArgumentError: wrong array length at 2 (expected 2, was 1)
from (pry):5:in `to_h'
```

It might seem contrived to use an odd number of elements, however I think the situation reveals why it comes up. This is in a script which takes the `environment:` key of a `docker-compose.yml`, where one of its syntaxes is an array of strings of the the form `KEY=value` _or_ `KEY`.

As such, the code was `Hash[lines.map { |line| line.split("=", 2) }]`. It can be worked around by changing the split to something like `(line.split("=", 2) << nil).first(2)` but it's quite indirect.

--------

## Expected behavior

I'm not really sure that I would expect it to know that it was unsafe. I think perhaps it should not  claim to be safe at autofix, though.

## Actual behavior

Test file:

```rb
lines = %w[
  A=b
  B=c
  C
]

Hash[lines.map { |line| line.split("=", 2) }]
```

This was replaced with the following code which raises an exception for the given `line`:

```
lines.map { |line| line.split("=", 2) }.to_h
```

I am using `standardrb` which enables this cop. However, it distinguishes between `--fix` and `--fix-unsafely`. Because this cop is marked as safe, this fix is automatically applied when formatting the file. I don't believe this is a bug in standard but because of the metadata of this cop marking it as safe unconditionally.

## RuboCop version

```
❯ rubocop -V
1.53.1 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.2.1) [arm64-darwin22]
```

koic:
Fixes #12015.

This PR marks `Style/HashConversion` as unsafe autocorrection.

This cop's autocorrection is unsafe because `ArgumentError` occurs if the number of elements is odd:

```ruby
Hash[[[1, 2], [3, 4], [5]]] #=> {1=>2, 3=>4, 5=>nil}
[[1, 2], [3, 4], [5]].to_h  #=> wrong array length at 2 (expected 2, was 1) (ArgumentError)
```

For example, this cannot be avoided when the arguments are variables.

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/change_mark_style_hash_conversion_as_unsafe_autocorrection.md
+++ b/changelog/change_mark_style_hash_conversion_as_unsafe_autocorrection.md
@@ -1,0 +1,1 @@
+* [#12015](https://github.com/rubocop/rubocop/issues/12015): Mark `Style/HashConversion` as unsafe autocorrection. ([@koic][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -3916,8 +3916,9 @@ Style/HashConversion:
   Description: 'Avoid Hash[] in favor of ary.to_h or literal hashes.'
   StyleGuide: '#avoid-hash-constructor'
   Enabled: pending
+  SafeAutoCorrect: false
   VersionAdded: '1.10'
-  VersionChanged: '1.11'
+  VersionChanged: '<<next>>'
   AllowSplatArgument: true
 
 Style/HashEachMethods:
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #12037, opened 2023-07-08
[Fix #11802] Improve handling of [] and () with percent symbol arrays

### LINKED ISSUE #11802: Symbol/WordArray style "percent" should also consider [ ] chars as "bad"
The current behavior of the Symbol/WordArray cop's default style ("percent") is as follows:

```ruby
# bad
[:foo, :[]=]

# bad (contains spaces)
%w[foo\ bar baz\ quux]

# good (allows [ ] chars)
%i[foo \[\]=]
```

I think we should change this so that `[ ]` trigger the same special handling that space char does, and the preferred style here would be `[:foo, :[]=]`

jasondoc3:
Addresses https://github.com/rubocop/rubocop/issues/11802, but also accounts for `()`

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/symbol_array.rb
+++ b/lib/rubocop/cop/style/symbol_array.rb
@@ -22,6 +22,15 @@ module RuboCop
       #   # bad
       #   [:foo, :bar, :baz]
       #
+      #   # bad (contains spaces)
+      #   %i[foo\ bar baz\ quux]
+      #
+      #   # bad (contains [] with spaces)
+      #   %i[foo \[ \]]
+      #
+      #   # bad (contains () with spaces)
+      #   %i(foo \( \))
+      #
       # @example EnforcedStyle: brackets
       #   # good
       #   [:foo, :bar, :baz]
@@ -40,6 +49,7 @@ module RuboCop
 
         PERCENT_MSG = 'Use `%i` or `%I` for an array of symbols.'
         ARRAY_MSG = 'Use %<prefer>s for an array of symbols.'
+        DELIMITERS = ['[', ']', '(', ')'].freeze
 
         class << self
           attr_accessor :largest_brackets
@@ -47,7 +57,7 @@ module RuboCop
 
         def on_array(node)
           if bracketed_array_of?(:sym, node)
-            return if symbols_contain_spaces?(node)
+            return if complex_content?(node)
 
             check_bracketed_array(node, 'i')
           elsif node.percent_literal?(:symbol)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/internal_affairs/undefined_config.rb
```diff
@@ -8,7 +8,7 @@ class UndefinedConfig < Base
         ALLOWED_CONFIGURATIONS = %w[
           Safe SafeAutoCorrect AutoCorrect Severity StyleGuide Details Reference Include Exclude
         ].freeze
-        RESTRICT_ON_SEND = %i[[] fetch].freeze
+        RESTRICT_ON_SEND = [:[], :fetch].freeze
```
koic, line 11: If bracket do not contain spaces, they should not be changed.

So, only the following case that need escaping can be changed from percent literal style to bracket style:

```ruby
%i[[ ]] #=> [:"[", :"]"]
```

This case where brackets do not contain spaces should not be changed:

```ruby
%i[[]]  #=> [:[]]
```

That is, it must be checked for spaces in brackets. Can you update the entire changes in this PR?

jasondoc3: @koic I think that makes sense to me!

But one thing that I notice is that while `%i[[] one]` does not currently raise an offense, rubocop still autocorrects `[:[], :one]` to `%i[\[\] one]`, which feels inconsistent 🤔 Any thoughts?

koic: When `EnforcedStyle: percent`, `%i[[] one]` (no spaces in the brackets) will need to adjust the implementation if necessary to be accepted.

lib/rubocop/cop/style/symbol_array.rb
```diff
@@ -22,6 +22,15 @@ module Style
       #   # bad
       #   [:foo, :bar, :baz]
       #
+      #   # bad (contains spaces)
+      #   %i[foo\ bar baz\ quux]
+      #
+      #   # bad (contains [] with spaces)
+      #   %i[foo\ \[ \]]
```
bbatsov, line 29: I'd remove the escaped space after `foo` to make this example clearer.

lib/rubocop/cop/style/symbol_array.rb
```diff
@@ -22,6 +22,15 @@ module Style
...
+      #
+      #   # bad (contains () with spaces)
+      #   %i(foo\ \( \))
```
bbatsov, line 32: I'd remove the escaped space after `foo` to make this example clearer.

bbatsov:
@jasondoc3 ping :-)

jasondoc3:
@bbatsov Removed the escaped spaces! I _think_ I addressed the concerns of @koic, but if not I can give another look :eyes:

lib/rubocop/cop/style/symbol_array.rb
```diff
@@ -57,13 +66,24 @@ def on_array(node)
 
         private
 
-        def symbols_contain_spaces?(node)
+        def complex_content?(node)
+          delimiters = ['[', ']', '(', ')']
```
koic, line 70: @jasondoc3 It could be a constant instead of a variable :-)

jasondoc3: Done!

koic:
Thanks!

PR #12048, opened 2023-07-14
[Docs] Add reference to LSP to quickstart section

koic:
This PR adds reference to RuboCop's built-in LSP to quickstart section in the README.

While I recognize that the quickstart section is expanding, I think that due to its potential to dramatically change the development experience, LSP is worth mentioning as a second form of magic.

To keep the quickstart text concise, I have not mentioned the required RuboCop version. It can be easily found by referring to the linked page.

-----------------

* [ ] Added tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,8 @@ $ cd my/cool/ruby/project
 $ rubocop
 ```
 
+You can also use this magic in your favorite editor with RuboCop's [built-in LSP](https://docs.rubocop.org/rubocop/usage/lsp.html).
+
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


bbatsov:
We should probably add some note in the `Server` section that people should prefer the use of the LSP server in editors. And maybe an entry in the FAQ about the differences between the two servers and their use-cases.

koic:
Yeah, the linked [LSP page](https://docs.rubocop.org/rubocop/usage/lsp.html) does provide a brief explanation on the differences from Server mode, but it's something to think about.

PR #12055, opened 2023-07-17
Allow parentheses in single-line match patterns

gsamokovarov:
```ruby
execute(query) => {elapsed:, sql_count:}

# (match-pattern
#   (send nil :execute
#     (send nil :query))
#   (hash-pattern
#     (match-var :elapsed)
#     (match-var :sql_count)))
```

is a single-line pattern-match. The current version of the `omit_parentheses` style in `Style/MethodCallWithArgsParentheses` forces us to omit the parens in the call, making it a call with an options hash that has a string key and a value omitted hash as an value:

```ruby
execute query => {elapsed:, sql_count:}

# (send nil :execute
#   (kwargs
#     (pair
#       (send nil :query)
#       (hash
#         (pair
#           (sym :elapsed)
#           (send nil :elapsed))
#         (pair
#           (sym :sql_count)
#           (send nil :sql_count))))))
```

This is yet another ambiguity and we need to allow the parentheses when we pattern match the value of a call.


koic:
Thanks!

PR #12060, opened 2023-07-21
Fix a false negative for `Style/StringLiteralsInInterpolation` for symbols with interpolation

### LINKED ISSUE #12059: Interpolation in symbols brings about `Style/StringLiterals` rather than `Style/StringLiteralsInInterpolation`
This problem is especially noticeable when using `EnforcedStyle: double_quotes` for `Style/StringLiterals`, as it corrects strings and symbols inconsistently.

## To reproduce

With default settings, run RuboCop on:

```rb
:"foo#{"sar".sub("s", "b")}"
```

## Expected behavior

A `Style/StringLiteralsInInterpolation` warning for each string inside of the interpolation.

## Actual behavior

A `Style/StringLiterals` warning for each string inside of the interpolation.

## RuboCop version

```
1.54.2 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.2.2) [x86_64-linux]
```

fatkodima:
Fixes #12059.

### DIFF
--- a/spec/rubocop/cop/style/string_literals_in_interpolation_spec.rb
+++ b/spec/rubocop/cop/style/string_literals_in_interpolation_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe RuboCop::Cop::Style::StringLiteralsInInterpolation, :config do
   context 'configured with single quotes preferred' do
     let(:cop_config) { { 'EnforcedStyle' => 'single_quotes' } }
 
-    it 'registers an offense for double quotes within embedded expression' do
+    it 'registers an offense for double quotes within embedded expression in a string' do
       expect_offense(<<~'RUBY')
         "#{"A"}"
            ^^^ Prefer single-quoted strings inside interpolations.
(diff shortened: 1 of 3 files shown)


vinnydiehl:
That was fast, thank you!

Now that I think of it, should we include RegEx interpolations here as well, for consistency?

fatkodima:
That is a feature request, for sure, so probably better as a separate PR. Should we do that? I don't know 😄 That is also an interpolation, so probably yes.

vinnydiehl:
Whatever we decide, I just did a double-take and realized that `Style/StringLiterals` is not working at all inside of a Regex interpolation! I'd call that a bug- I'll open a separate issue and raise the question there which error would be appropriate.

spec/rubocop/cop/style/string_literals_in_interpolation_spec.rb
```diff
@@ -8,10 +8,13 @@
       expect_offense(<<~'RUBY')
         "#{"A"}"
            ^^^ Prefer single-quoted strings inside interpolations.
+        :"#{"A"}"
+            ^^^ Prefer single-quoted strings inside interpolations.
```
koic, line 12: Can you add a new test case without directly editing the existing test case?

spec/rubocop/cop/style/string_literals_in_interpolation_spec.rb
```diff
@@ -70,10 +73,13 @@ def m
       expect_offense(<<~'RUBY')
         "#{'A'}"
            ^^^ Prefer double-quoted strings inside interpolations.
+        :"#{'A'}"
+            ^^^ Prefer double-quoted strings inside interpolations.
```
koic, line 77: Ditto :-)

spec/rubocop/cop/style/string_literals_in_interpolation_spec.rb
fatkodima: Done.

koic:
Thanks!

PR #12062, opened 2023-07-22
Fix `LoadError` on Windows when loading RuboCop from symlinks

p0deje:
There is a particular edge case when RuboCop is loaded from a fully symlinked environment. This is the default for Bazel which creates a tree of symlinks for every dependency, mimicking the real file tree structure.

In such a setup, RuboCop would fail to load with an error. Resolving the path to an absolute one fixes it.

```
No such file or directory - //?/D:/_bazel/external/bundle/ruby/3.0.0/gems/rubocop-1.36.0/lib/rubocop/../../exe
...
//?/D:/_bazel/external/bundle/ruby/3.0.0/gems/rubocop-1.36.0/lib/rubocop/result_cache.rb:208:in `each'
```

I am not sure what is the best way to write a test for this, I could use some guidance.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/result_cache.rb
+++ b/lib/rubocop/result_cache.rb
@@ -202,6 +202,10 @@ module RuboCop
       lib_root = File.join(File.dirname(__FILE__), '..')
       exe_root = File.join(lib_root, '..', 'exe')
 
+      # Make sure to use an absolute path to prevent errors on Windows
+      # when traversing the relative paths with symlinks.
+      exe_root = File.absolute_path(exe_root)
+
       # These are all the files we have `require`d plus everything in the
       # exe directory. A change to any of them could affect the cop output
       # so we include them in the cache hash.


bbatsov:
> This is the default for Bazel which creates a tree of symlinks for every dependency, mimicking the real file tree structure.

Does it work differently on Unix-like OSes? I'm just trying to understand if this is truly Windows-specific.

p0deje:
> Does it work differently on Unix-like OSes? I'm just trying to understand if this is truly Windows-specific.

It works the same way on Unix-like OSes, but the issue with Windows is that its symlink support is way different - as far as I know there is no real symlinks at all. You can see the `//?` prefix in the path which denotes this is a UNC path - I suppose that's what is being used internally.

I wish I could find a way configure Windows to avoid these problems rather than patching gems. If you are not willing to merge the PR, I can totally understand that.

bbatsov:
The patch seems harmless enough and it does solve a real problem. I'll merge it as is, given that adding tests for this would be quite complicated.

bbatsov:
Ops, just noticed you forgot to add a changelog entry for this.

PR #12069, opened 2023-07-25
[Fix #12068] Fix a false positive for `Style/ReturnNilInPredicateMethodDefinition`

### LINKED ISSUE #12068: Style/ReturnNilInPredicateMethodDefinition is firing incorrectly
## Expected behavior

`Style/ReturnNilInPredicateMethodDefinition` should only fire when a predicate method returns `nil`.

## Actual behavior

Offences are reported for code that uses `nil` but does not return it:

```
Offenses:

app/models/oauth_token.rb:51:23: C: [Correctable] Style/ReturnNilInPredicateMethodDefinition: Return false instead of nil in predicate methods.
    invalidated_at != nil
                      ^^^
lib/bounding_box.rb:92:19: C: [Correctable] Style/ReturnNilInPredicateMethodDefinition: Return false instead of nil in predicate methods.
    to_a.exclude?(nil)
                  ^^^

2 files inspected, 2 offenses detected, 2 offenses autocorrectable
Finished in 0.807309715077281 seconds
```

Run rubocop on these methods:

```ruby
def invalidated?
  invalidated_at != nil
end

def complete?
  to_a.exclude?(nil)
end
```

## RuboCop version

```
1.55.0 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.2.2) [x86_64-linux]
  - rubocop-capybara 2.18.0
  - rubocop-factory_bot 2.23.1
  - rubocop-minitest 0.31.0
  - rubocop-performance 1.18.0
  - rubocop-rails 2.20.2
  - rubocop-rake 0.6.0
```

koic:
Fixes #12068.

This PR fixes a false positive for `Style/ReturnNilInPredicateMethodDefinition` when the last method argument in method definition is `nil`.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/return_nil_in_predicate_method_definition.rb
+++ b/lib/rubocop/cop/style/return_nil_in_predicate_method_definition.rb
@@ -77,6 +77,8 @@ module RuboCop
         private
 
         def nil_node_at_the_end_of_method_body(body)
+          return body if body.nil_type?
+          return unless body.begin_type?
           return unless (last_child = body.children.last)
 
           last_child if last_child.is_a?(AST::Node) && last_child.nil_type?
--- a/spec/rubocop/cop/style/return_nil_in_predicate_method_definition_spec.rb
+++ b/spec/rubocop/cop/style/return_nil_in_predicate_method_definition_spec.rb
@@ -40,7 +40,7 @@ RSpec.describe RuboCop::Cop::Style::ReturnNilInPredicateMethodDefinition, :confi
       RUBY
     end
 
-    it 'registers an offense when using `nil` at the end of the predicate method definition' do
+    it 'registers an offense when using `nil` at the end of the predicate method definition and using guard condition' do
       expect_offense(<<~RUBY)
         def foo?
           return true if condition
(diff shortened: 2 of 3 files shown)


mobilutz:
@koic Do you know, when this will be merged?

bbatsov:
@mobilutz All in good time. 😉

spec/rubocop/cop/style/return_nil_in_predicate_method_definition_spec.rb
```diff
@@ -59,12 +59,17 @@ def foo?
       RUBY
     end
 
-    it 'does not register an offense when using `return false`' do
```
bbatsov, line 62: Why did you remove this test case?

spec/rubocop/cop/style/return_nil_in_predicate_method_definition_spec.rb
```diff
@@ -101,6 +106,40 @@ def foo?
...
+
+    it 'does not register an offense when assigning `nil` to a variable in predicate method definition' do
+      expect_no_offenses(<<~RUBY)
+        def foo?
+          bar = nil
+        end
+      RUBY
+    end
+
+    it 'does not register an offense when using `return false`' do
```
bbatsov, line 133: Ah, seems it was moved.

lib/rubocop/cop/style/return_nil_in_predicate_method_definition.rb
```diff
@@ -77,6 +77,8 @@ def on_def(node)
         private
 
         def nil_node_at_the_end_of_method_body(body)
+          return if !body.begin_type? && !body.nil_type?
```
bbatsov, line 80: It seems to me the logic in this method became a bit too convoluted. I wonder if there isn't a simpler way to get the final node of method that doesn't use `body.children.last` and would not require to the additional checks. I'm guessing you can also just check that the parent of the last `nil` is either `begin` or `def`.

koic: I haven't come up with a good logic yet 💦, but for now, I've simplified the conditions a bit.

bbatsov: Yeah, it reads better this way. I guess it will do for now.

PR #12092, opened 2023-08-02
[Fix #12061] Support regex in StringLiteralsInInterpolation

### LINKED ISSUE #12061: `Style/StringLiterals` does not work for `String`s interpolated into a `Regex`
## To reproduce

With default settings, run RuboCop on:

```rb
/foo#{"sar".sub("s", "b")}/
```

## Expected behavior

A `Style/StringLiterals` warning, or maybe `Style/StringLiteralsInInterpolation`.

## Actual behavior

No warnings.

## RuboCop version

```
1.54.2 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.2.2) [x86_64-linux]
```

## Other thoughts

Would it be better to use `Style/StringLiteralsInInterpolation` here, rather than `Style/StringLiterals`? This would be consistent with other interpolations, however, the rule isn't quite as important in this case since there are no external `"` to cause confusion, as there would be with a `String` or `Symbol`.

jonas054:
Interpolation can occur in strings, symbols, and regular expressions, so update code to support regex and extend code examples in documentation comments.

### DIFF
--- a/changelog/fix_support_regex_in_string_literals_in_interpolation.md
+++ b/changelog/fix_support_regex_in_string_literals_in_interpolation.md
@@ -1,0 +1,1 @@
+* [#12061](https://github.com/rubocop/rubocop/issues/12061): Support regex in StringLiteralsInInterpolation. ([@jonas054][])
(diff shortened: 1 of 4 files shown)


koic:
Thanks!

Issue #12100, opened 2023-08-07
Style/ArgumentsForwarding breaks on argument destructuring in method signature


herwinw:
```ruby
def forwarded(*args, **kwargs)
  p [args, kwargs].inspect
end

def foo((bar, baz), **kwargs)
  forwarded(bar, baz, **kwargs)
end

foo([1, 2], keyword: 3)
```
The extra set parentheses in the method signature of `foo` deconstructs the array argument and names the first two elements `bar` and `baz` (extra elements in the array are ignored, missing elements result in `nil` values).

Running Rubocop on this file with `TargetRubyVersion: 3.1`, this results in a parsing error:
```
An error occurred while Style/ArgumentsForwarding cop was inspecting poc.rb:8:0.
undefined method `default?' for s(:mlhs,
  s(:arg, :bar),
  s(:arg, :baz)):RuboCop::AST::Node

        methods.each {|method| def_instance_delegator(accessor, method)}
               ^^^^^
/usr/lib/ruby/3.1.0/forwardable.rb:238:in `none?'
/usr/lib/ruby/3.1.0/forwardable.rb:238:in `none?'
vendor/bundle/ruby/3.1.0/gems/rubocop-1.55.1/lib/rubocop/cop/style/arguments_forwarding.rb:325:in `pre_ruby_32_allow_forward_all?'
[... more backtrace ...]
```
Setting the target version to Ruby 3.2 does not result in this error (as hinted by the `pre_ruby_32_allow_forward_all?` in the backtrace)

--------
## RuboCop version

```
$ bundle exec rubocop -V
1.55.1 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.1.2) [x86_64-linux-gnu]
```

Issue #12096, opened 2023-08-04
Style/ArgumentsForwarding finds an offense when **kwargs are extended


kpost:
`Style/ArgumentsForwarding` autocorrect will just replace all arguments, even when **kwargs are extended inside the method.

(My use case actually has the `category` leading argument as well, which I know overlaps with #12087, but the issue described is different)

--------

## Expected behavior

Shouldn't find an offense for this:

```
def my_method(category, *args, **kwargs, &)
  tag.tr(*args, id: category, class: ('selected' if some_boolean), **kwargs, &)
end
```

## Actual behavior
```
def my_method(...)
  tag.tr(...)
end
```

## Steps to reproduce the problem

Autocorrect above example for `Style/ArgumentsForwarding`

## RuboCop version

```
1.55.1 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.1.3) [arm64-darwin22]
  - rubocop-performance 1.18.0
  - rubocop-rails 2.20.2
  - rubocop-rspec 2.22.0
```

Issue #12089, opened 2023-08-01
Style/ArgumentsForwarding incorrectly fixes a case where an argument is used elsewhere in the method


davidenglishmusic:
The auto-correct for the cop can break code by replacing the arguments when one of them is referenced in another part of the method.

--------

## Expected behavior
The following should not elicit a correction:

```
def method_missing(m, *args, **kwargs, &block)
  if @template.respond_to?(m)
    @template.send(m, *args, **kwargs, &block)
  else
    super
  end
end
```

## Actual behavior
It corrects to the following which throws an undefined variable error:

```
def method_missing(...)
  if @template.respond_to?(m)
    @template.send(...)
  else
    super
  end
end
```

## Steps to reproduce the problem

Auto-correct the code listed under the expected behavior heading.

## RuboCop version

```
1.55.1 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.1.1) [x86_64-linux]
  - rubocop-capybara 2.18.0
  - rubocop-factory_bot 2.23.1
  - rubocop-performance 1.18.0
  - rubocop-rails 2.20.2
  - rubocop-rspec 2.23.0
```

mattbrictson:
I noticed a similar bug.

Original:

```ruby
def post(*args, &block)
  future_on(executor, *args, &block)
```

Erroneously corrected by rubocop to:

```ruby
def post(...)
  future_on(...)
```

Expected behavior:

```ruby
def post(...)
  future_on(executor, ...)
```

tmikoss:
Adding another similar example. Produced by `rubocop -a`, version `1.55.1`. `association_name` is used in the fucntion body.

<img width="519" alt="Screenshot 2023-08-07 at 10 31 07" src="https://github.com/rubocop/rubocop/assets/222126/d502549d-eeb8-44d3-9b44-bcec339fc2d7">

koic:
FYI, I've opened #12101 to fix due to many reports about this issue. Thank you.

chulkilee:
[1.56](https://github.com/rubocop/rubocop/releases/tag/v1.56.0) included the fix (#12103) but it did not fix the issue.

Source

```ruby
class Foo
  def bar(*args, &block)
    hello(*args, &block)
    world(*args)
  end
end
```

Script

```ruby
bin/rubocop --version
1.56.0

bin/rubocop -a test.rb
Inspecting 1 file
C

Offenses:

[...]

PR #12103, opened 2023-08-08
[Fix #12087] Fix false positives for Style/ArgumentsForwarding

### LINKED ISSUE #12087: Style/ArgumentsForwarding mis-corrects when the called method has additional arguments.
Style/ArgumentsForwarding will replace all arguments to the called method with `...`, even if there are arguments besides the `*args, **kwargs, &`.

Given the following code:

```ruby
def self.get(*args, **kwargs, &)
  CanvasHttp.request(Net::HTTP::Get, *args, **kwargs, &)
end
```

## Expected behavior

```ruby
def self.get(...)
  CanvasHttp.request(Net::HTTP::Get, ...)
end
```

## Actual behavior

```ruby
def self.get(...)
  CanvasHttp.request(...)
end
```

Note the additional Net::HTTP::Get inline argument has been removed.

Auto-correct the above code with the Style/ArgumentsForwarding cop.

## RuboCop version

```
1.55.1 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.1.3) [arm64-darwin21]
  - rubocop-factory_bot 2.23.1
  - rubocop-graphql 1.3.0
  - rubocop-performance 1.18.0
  - rubocop-rails 2.20.2
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.23.0
```

Note that I'm not sure if adding leading arguments is supported by all Ruby versions. I seem to remember when this syntax was introduced it was very limited. If that's the case, I would expect this cop to not even find an offense if the target Ruby version doesn't support leading arguments.

owst:
Additionally fixes:
* #12089
* #12096
* #12100

In Ruby < 3.2 this cop was overzealous and not considering various additional args/kwargs and other reasons to not allow forward-all (`...`), and was additionally assuming that forward-all meant all arguments were replaced, which was not always the case.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/arguments_forwarding.rb
+++ b/lib/rubocop/cop/style/arguments_forwarding.rb
@@ -100,7 +100,7 @@ module RuboCop
           return if send_classifications.empty?
 
           if only_forwards_all?(send_classifications)
-            add_forward_all_offenses(node, send_classifications)
+            add_forward_all_offenses(node, send_classifications, forwardable_args)
           elsif target_ruby_version >= 3.2
             add_post_ruby_32_offenses(node, send_classifications, forwardable_args)
           end
@@ -118,12 +118,13 @@ module RuboCop
           send_classifications.each_value.all? { |c, _, _| c == :all }
         end
 
-        def add_forward_all_offenses(node, send_classifications)
-          send_classifications.each_key do |send_node|
-            register_forward_all_offense_on_forwarding_method(send_node)
+        def add_forward_all_offenses(node, send_classifications, forwardable_args)
+          send_classifications.each do |send_node, (_c, forward_rest, _forward_kwrest)|
+            register_forward_all_offense(send_node, send_node, forward_rest)
           end
 
-          register_forward_all_offense_on_method_def(node)
+          rest_arg, _kwrest_arg, _block_arg = *forwardable_args
+          register_forward_all_offense(node, node.arguments, rest_arg)
         end
 
         def add_post_ruby_32_offenses(def_node, send_classifications, forwardable_args)
@@ -186,9 +187,7 @@ module RuboCop
 
         def register_forward_args_offense(def_arguments_or_send, rest_arg_or_splat)
           add_offense(rest_arg_or_splat, message: ARGS_MSG) do |corrector|
-            unless parentheses?(def_arguments_or_send)
-              add_parentheses(def_arguments_or_send, corrector)
-            end
+            add_parens_if_missing(def_arguments_or_send, corrector)
 
             corrector.replace(rest_arg_or_splat, '*')
           end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -186,46 +187,36 @@ def classification_and_forwards(def_node, send_node, referenced_lvars, forwardab
...
-            if add_parens && !parentheses?(def_arguments_or_send)
-              add_parentheses(def_arguments_or_send, corrector)
-            end
+            add_parens_if_missing(def_arguments_or_send, corrector) if add_parens
 
             corrector.replace(kwrest_arg_or_splat, '**')
           end
         end
 
-        def register_forward_all_offense_on_forwarding_method(forwarding_method)
```
owst, line 207: `register_forward_all_offense_on_forwarding_method` and `register_forward_all_offense_on_method_def` were essentially doing the same thing, and could be made to call the same method.

lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -186,46 +187,36 @@ def classification_and_forwards(def_node, send_node, referenced_lvars, forwardab
 
         def register_forward_args_offense(def_arguments_or_send, rest_arg_or_splat)
           add_offense(rest_arg_or_splat, message: ARGS_MSG) do |corrector|
-            unless parentheses?(def_arguments_or_send)
-              add_parentheses(def_arguments_or_send, corrector)
-            end
```
owst, line 191: I realised I could reuse this functionality across `register_forward_args_offense`, `register_forward_kwargs_offense` and the new method `register_forward_all_offense` (which replaces the two separate `register_forward_all_offense_on_forwarding_method` and `register_forward_all_offense_on_method_def`)

lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -118,12 +118,13 @@ def only_forwards_all?(send_classifications)
           send_classifications.each_value.all? { |c, _, _| c == :all }
         end
 
-        def add_forward_all_offenses(node, send_classifications)
-          send_classifications.each_key do |send_node|
-            register_forward_all_offense_on_forwarding_method(send_node)
+        def add_forward_all_offenses(node, send_classifications, forwardable_args)
+          send_classifications.each do |send_node, (_c, forward_rest, _forward_kwrest)|
```
owst, line 122: We need to capture the `forward_rest` value so that we can replace only after the splat if leading arguments are allowed e.g. in 
```ruby
def foo(x, *args, **kwargs, &block)
```
vs
```ruby
def foo(x, ...)
```

lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -186,46 +187,36 @@ def classification_and_forwards(def_node, send_node, referenced_lvars, forwardab
...
-            arguments_range = range_with_surrounding_space(
-              method_definition.arguments.source_range, side: :left
-            )
-            corrector.replace(arguments_range, '(...)')
+            corrector.replace(arg_range, '...')
           end
         end
 
-        def arguments_range(node)
+        def arguments_range(node, first_node)
```
owst, line 214: This change enables forward-all-with-leading-args e.g. `foo(x, ...)`

lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -289,6 +286,27 @@ def classification
...
           private
 
+          def can_forward_all?
+            return false if any_arg_referenced?
+            return false if ruby_32_missing_rest_or_kwest?
+            return false unless offensive_block_forwarding?
+            return false if forward_additional_kwargs?
+
+            no_additional_args? || (target_ruby_version >= 3.0 && no_post_splat_args?)
+          end
```
owst, line 296: This is the meat of the change of the PR - with various additional conditions as to when a forward-all is or is not allowed

owst:
//cc @koic @bbatsov - please let me know what you think. I think there's a few points of simplification in the implementation that I've tried to highlight, and of course the bug fixes

changelog/fix_false_positives_for_style_arguments_forwarding_one.md
```diff
@@ -0,0 +1 @@
+* [#12087](https://github.com/rubocop/rubocop/issues/12087): Fix false positive for `Style/ArgumentsForwarding` with additional arg passed in send node. ([@owst][])
```
koic, line 1: I think changelog entries can be merged into this representative issue number entry.

lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -305,25 +323,36 @@ def referenced_block_arg?
...
+            splat_index = arguments.index(forwarded_rest_arg)
+            arg_after_splat = arguments[splat_index + 1]
+
+            [nil, :hash, :block_pass].include?(arg_after_splat&.type)
+          end
+
+          def forward_additional_kwargs?
+            return false unless forwarded_kwrest_arg
+
+            forwarded_kwrest_arg.parent.children.size != 1
```
koic, line 344: Nits :-)

```suggestion
            !forwarded_kwrest_arg.parent.children.one?
```

koic:
@owst This looks good to me. Can you squash your commits into one?

owst:
> @owst This looks good to me. Can you squash your commits into one?

Thanks @koic - done!

koic:
Thanks!

PR #12121, opened 2023-08-15
Fix instance variable not initialized warning

chulkilee:
Fix `warning: instance variable @pairs not initialized` warning

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/utils/regexp_ranges.rb
+++ b/lib/rubocop/cop/utils/regexp_ranges.rb
@@ -6,27 +6,40 @@ module RuboCop
       # Helper to abstract complexity of building range pairs
       # with octal escape reconstruction (needed for regexp_parser < 2.7).
       class RegexpRanges
-        attr_reader :compound_token, :root
+        attr_reader :root
 
         def initialize(root)
           @root = root
           @compound_token = []
+          @pairs = []
+          @populated = false
+        end
+
+        def compound_token
+          populate_all unless @populated
+
+          @compound_token
         end
 
         def pairs
-          unless @pairs
-            @pairs = []
-            populate(root)
-          end
+          populate_all unless @populated
+
+          @pairs
+        end
+
+        private
+
+        def populate_all
+          populate(@root)
 
           # If either bound is a compound the first one is an escape
           # and that's all we need to work with.
           # If there are any cops that wanted to operate on the compound
           # expression we could wrap it with a facade class.
-          @pairs.map { |pair| pair.map(&:first) }
-        end
+          @pairs.map! { |pair| pair.map(&:first) }
 
-        private
+          @populated = true
+        end
 
         def populate(expr)
           expressions = expr.expressions.to_a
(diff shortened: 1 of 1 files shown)


lib/rubocop/cop/utils/regexp_ranges.rb
```diff
@@ -14,7 +14,7 @@ def initialize(root)
         end
 
         def pairs
-          unless @pairs
+          unless defined?(@pairs)
```
koic, line 17: Can you initialize `@pairs` with the `initialize` method instead of using `define?`?

```diff
diff --git a/lib/rubocop/cop/utils/regexp_ranges.rb b/lib/rubocop/cop/utils/regexp_ranges.rb
index aa60666c9..6322091c5 100644
--- a/lib/rubocop/cop/utils/regexp_ranges.rb
+++ b/lib/rubocop/cop/utils/regexp_ranges.rb
@@ -11,6 +11,7 @@ module RuboCop
         def initialize(root)
           @root = root
           @compound_token = []
+          @pairs = nil
         end

         def pairs
```

lib/rubocop/cop/utils/regexp_ranges.rb
```diff
@@ -6,27 +6,41 @@ module Utils
...
+          populate_all unless @populated
+
+          @pairs
+        end
+
+        private
+
+        def populate_all
+          @compound_token = []
+          @pairs = []
```
koic, line 32: It looks like the initialization of these instance variables can be moved to the `initialize` method.

lib/rubocop/cop/utils/regexp_ranges.rb
```diff
@@ -6,27 +6,41 @@ module Utils
...
+
+          populate(@root)
 
           # If either bound is a compound the first one is an escape
           # and that's all we need to work with.
           # If there are any cops that wanted to operate on the compound
           # expression we could wrap it with a facade class.
-          @pairs.map { |pair| pair.map(&:first) }
-        end
+          @pairs = @pairs.map { |pair| pair.map(&:first) }
```
koic, line 40: Using `map!` would be simple.

```suggestion
          @pairs.map! { |pair| pair.map(&:first) }
```

lib/rubocop/cop/utils/regexp_ranges.rb
chulkilee: I intentionally moved this to `#populate_all` since this method is responsible for lifecycle of `@compound_token` and `@pairs`. For instance existing and new code are NOT thread-safe - if we want to add a lock then it should be in this method with initializing this value anyway.

If you still prefer to put them in `#initialize` just let me know - I'll move it with a comment

koic: Yeah, I find it easier to understand when these instance variables are set up and prepared at the point when object is instantiated. Can you move these into `initialize`?

chulkilee: Moved!

chulkilee:
Squashed!

koic:
Thanks!

PR #12133, opened 2023-08-19
Fix `Style/RedundantSelfAssignmentBranch` to handle heredocs

r7kamura:
There is a bug in the autocorrection of `Style/RedundantSelfAssignmentBranch`, which causes the following replacement if `opposite_branch` is a heredoc.

Before:

```ruby
foo =
  if some_condition
    foo
  else
    <<~TEXT
      bar
    TEXT
  end
```

After:

```ruby
foo =
  <<~TEXT unless some_condition
```

So in this pull request, I will modify it to make up for the missing parts in the case of heredoc.

This implementation is roughly the same as other methods of dealing with heredocs, for example, in `Style/FileWrite`:

https://github.com/rubocop/rubocop/blob/dd299948479c15cc7dd254a339631d241938b20b/lib/rubocop/cop/style/file_write.rb#L108-L121

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_style_redundant_self_assignment_branch_to_handle_heredocs.md
+++ b/changelog/fix_style_redundant_self_assignment_branch_to_handle_heredocs.md
@@ -1,0 +1,1 @@
+* [#12133](https://github.com/rubocop/rubocop/pull/12133): Fix `Style/RedundantSelfAssignmentBranch` to handle heredocs. ([@r7kamura][])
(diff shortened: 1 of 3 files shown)


changelog/fix_style_redundant_self_assignment_branch_to_handle_heredocs.md
```diff
@@ -0,0 +1 @@
+* [#12130](https://github.com/rubocop/rubocop/pull/12130): Fix `Style/RedundantSelfAssignmentBranch` to handle heredocs. ([@r7kamura][])
```
koic, line 1:
```suggestion
* [#12133](https://github.com/rubocop/rubocop/pull/12133): Fix `Style/RedundantSelfAssignmentBranch` to handle heredocs. ([@r7kamura][])
```

koic:
Thanks!

PR #12147, opened 2023-08-22
If ~/.rubocop.yml contains a Rails (or Performance) cop, loading excludes causes "`Rails` cops have been extracted" error

martinemde:
When running rubocop in a project that does not load Rails (in this case I'm running rubocop on rubygems), if you have Rails/ cops in your `~/.rubycop.yml` file, then they will be loaded, validated, and the obsolete checker will raise.

However, the docs seem to explicitly say that this file should not be loaded since the project root contains a .rubocop.yml.

This is happening because the tests do not test the common case where projects are in a subdirectory of the user's home dir.

When the `.rubocop.yml` file in $HOME are in an ancestor folder of the current project, the  `ConfigLoader` looks for the "last file upwards" matching the file pattern ".rubocop.yml" and finds `~/.rubocop.yml`.

```ruby
      def add_excludes_from_files(config, config_file)
        exclusion_file = find_last_file_upwards(DOTFILE, config_file, ConfigFinder.project_root) # finds ~/.rubocop.yml
        # **snip**
        config.add_excludes_from_higher_level(load_file(exclusion_file)) # load_file loads and validates, causing obsolete errors.
      end
```

This results in fully loading the `~/.rubocop.yml` and loading the excludes contained therein. However, since the validation and obsolete checks are run, errors are generated for not requiring `rubocop-rails`.

My offered fix is to specifically exclude this file, but I can reduce this PR back to just the failing spec if another solution is desired.

-----------------

* [ ] Added tests.


martinemde:
I can squash when we agree this solution is acceptable.

lib/rubocop/config_loader.rb
```diff
@@ -139,7 +139,10 @@ def add_excludes_from_files(config, config_file)
         exclusion_file = find_last_file_upwards(DOTFILE, config_file, ConfigFinder.project_root)
 
         return unless exclusion_file
-        return if PathUtil.relative_path(exclusion_file) == PathUtil.relative_path(config_file)
+
+        exclusion_relative_path = PathUtil.relative_path(exclusion_file)
+        return if exclusion_relative_path == PathUtil.relative_path(config_file)
+        return if exclusion_relative_path == PathUtil.relative_path(File.join(Dir.home, DOTFILE))
```
koic, line 145: Can you refactor to the below?

```suggestion
        return if exclusion_relative_path == PathUtil.relative_path(config_file) ||
                  exclusion_relative_path == PathUtil.relative_path(File.join(Dir.home, DOTFILE))
```

koic:
I leave one small comment, but overall it's LGTM. Can you squash your commits into one?

martinemde:
Squashed. Thank you!

martinemde:
Hey, I just realize this is not the right fix. Please hold off on merging. We need to instead return the file _before_ the $HOME/.rubocop.yml in the directory hierarchy. As is this code will skip the actual project root yml if there is a nested config deeper in the project.

lib/rubocop/config_loader.rb
```diff
@@ -139,7 +139,10 @@ def add_excludes_from_files(config, config_file)
         exclusion_file = find_last_file_upwards(DOTFILE, config_file, ConfigFinder.project_root)
```
martinemde, line 139: Actual fix needs to modify find_last_file_upwards to not return the home .yml file, but whatever config it finds before that home dotfile.

spec/rubocop/config_loader_spec.rb
```diff
@@ -149,6 +149,40 @@
       end
     end
 
+    context 'when the project is in a subdirectory of $HOME' do
+      let(:file_path) { File.expand_path('~/dir/.rubocop.yml') }
+
+      before do
+        Dir.mkdir(File.expand_path('~/dir'))
+
+        create_file(file_path, <<~YAML)
```
martinemde, line 158: The test here should look more like the nested test below so it would fail correctly if we had merged this code.

martinemde (commented):
I'll get these fixes tomorrow.

martinemde:
Worth noting, @deivid-rodriguez linked me to a similar fix #8176 from a few years ago trying to address a similar problem. Hopefully we can get this specified clearly enough this time.

martinemde:
I chose to attempt forcing all specs in the isolated environment to stay within the directory tree that is allowed. Many of the specs were placing files outside of the tmpdir, then testing that those files weren't loaded (relying on the root_level behavior which is only enabled in tests). This meant that some of the tests only passed because the test environment kept the file from being found. In the real world it would pick up the file (just like it does).

However, I'm having trouble fixing a few of the specs. They should all be caused by this situation of previously loading or relying on files that are now off limits (or assuming they wouldn't be loaded). They might also be outside of the search path, or some other magic is happening that I don't understand. Maybe someone with more experience in RuboCop code could help?

bbatsov:
@martinemde Seems we forgot this one for a while and now it needs to be rebased on top of `master`.

PR #12150, opened 2023-08-24
Fix shadowing of FileFinder stop_dir in isolated tests.

martinemde:
Previously, stop_dir was ignored when FileFinde.root_level was present.

Tests that should have relied on stop_dir being set correctly, instead passed because the root_level happened to be the same as the stop_dir because of how the test was constructed.

Shift where FileFinder.root_level is set so that we use root_level only for protecting the test environment tmpdir. It is now set to the actual root of the tmpdir, above which nothing should be written in isolated environment tests.

This is a partial extraction of fixes in #12147, which had to fix too many problems like this and became too large.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/target_finder_spec.rb
+++ b/spec/rubocop/target_finder_spec.rb
@@ -510,18 +510,16 @@ RSpec.describe RuboCop::TargetFinder, :isolated_environment do
     end
 
     it 'can exclude symlinks as well as directories' do
-      Dir.mktmpdir do |tmpdir|
-        create_empty_file(File.join(tmpdir, 'ruby5.rb'))
-        create_link('link', tmpdir)
+      create_empty_file(File.join(Dir.home, 'ruby5.rb'))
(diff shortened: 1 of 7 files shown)


spec/rubocop/target_finder_spec.rb
```diff
@@ -510,18 +510,16 @@
     end
 
     it 'can exclude symlinks as well as directories' do
-      Dir.mktmpdir do |tmpdir|
-        create_empty_file(File.join(tmpdir, 'ruby5.rb'))
-        create_link('link', tmpdir)
+      create_empty_file(File.join(Dir.home, 'ruby5.rb'))
+      create_link('link', Dir.home)
```
martinemde, line 514: Using Dir.home instead of a second tmpdir, which maintains isolation while staying consistent with the test's intention.

bbatsov:
The proposed changes look good to me. Thanks!

PR #12156, opened 2023-08-27
Add new `Naming/PrivateMethodName` cop

kyoshidajp:
This PR adds new `Naming/PrivateMethodName` cop. It checks that private methods names start with a underscore prefix.

```ruby
# bad
class MyClass
  private
  def foo
    # ...
  end
end

# bad
class MyClass
  def self.foo
    # ...
  end

  private_class_method :foo
end

# good
class MyClass
  private
  def _foo
    # ...
  end
end

# good
class MyClass
  def self._foo
    # ...
  end

  private_class_method :_foo
end
```

Sometimes there are cases where private methods are prefixed with an underscore (e.g. [Active Model](https://github.com/rails/rails/blob/v7.0.7.2/activemodel/lib/active_model/callbacks.rb#L128-L153)). This is a good idea because it makes it easier to determine if a method is private or not.

The following topics may need to be considered.

1. `protected` methods are should be covered?
   - Actually, I am not sure which one it should be since I rarely use the protected method. It might be better without it.
1. `attr` accessors are should be covered?
   - If covered, it also affects instance variables naming. I implemented it anyway.

-----------------

* [x] Added tests.


lib/rubocop/cop/naming/private_method_name.rb
```diff
@@ -0,0 +1,137 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Naming
+      # Checks that private methods names
+      # do not start without a underscore prefix.
```
bbatsov, line 7: Might be good to mention here that the convention is used to some extent (e.g. in Rails), but it's not widely adopted in the community and that's the reason why this cop is disabled by default.

lib/rubocop/cop/naming/private_method_name.rb
```diff
@@ -0,0 +1,137 @@
...
+      #   class MyClass
+      #     def self._foo
+      #       # ...
+      #     end
+      #
+      #     private_class_method :_foo
+      #   end
+      #
+      class PrivateMethodName < Base
+        extend AutoCorrector
```
bbatsov, line 45: I guess you should remove this, as the cop doesn't implement auto-correct.

bbatsov:
Overall I'm fine with the proposed cop, given that it will be disabled. I've left a bit of feedback. @koic Feel free to take a look as well.

koic:
Starting private method names with an underscore sometimes indicates a struggle to find a good name for internal methods in contrast to public methods. In these cases, a more descriptive naming would be the essential solution.

It's questionable whether this rule should be a general guideline supported by RuboCop. For instance, there are numerous private methods without an underscore in the Rails repository.
https://github.com/rails/rails/blob/v7.0.7.2/activemodel/lib/active_model/dirty.rb#L250-L292

kyoshidajp:
@koic Thank you for your comment.

I believe my suggestion is only optional and should be disabled by default. Certainly there are many private methods without underscores in several projects. I believe that this cop is effective in determining if a method is private or not, but I am not trying to enforce it. I hope that can be determined by each project.

Would you still be negative about adopting it?

kyoshidajp:
OK. I do not wish to force the introduction of subtle rules that you do not agree with. Again, thank you for your comments and review.

PR #12169, opened 2023-08-29
[Fix #12168]: Fix bug in `Style/ArgumentsForwarding`

### LINKED ISSUE #12168: Bug in `Style/ArgumentsForwarding`: only last occurrence of repeated method call is autocorrected
Extracted from https://github.com/rubocop/rubocop/issues/12144#issuecomment-1695879179

With the following code, RuboCop only marks the second send as an offense, which leads to the auto-corrected code having an undefined lvar:
```ruby
def foo(bar, *args)
  baz(*args)

  if something?
    baz(*args)
  end
end
```

```
Inspecting 1 file
C

Offenses:

example.rb:1:14: C: [Corrected] Style/ArgumentsForwarding: Use anonymous positional arguments forwarding (*).
def foo(bar, *args)
             ^^^^^
example.rb:5:9: C: [Corrected] Style/ArgumentsForwarding: Use anonymous positional arguments forwarding (*).
    baz(*args)
        ^^^^^

```

Auto-corrected code:

```ruby
def foo(bar, *)
  baz(*args)

  if something?
    baz(*)
  end
end
```

--------

## Expected behavior

There should be an additional offense at the send on line 2 and the auto-corrected file should contain:
```ruby
def foo(bar, *)
  baz(*)

  if something?
    baz(*)
  end
end
```

## Actual behavior

The auto-corrected file contains:
```ruby
def foo(bar, *)
  baz(*args) # Bug here: `args` is no longer defined

  if something?
    baz(*)
  end
end
```

Use the example given above.

## RuboCop version

```
$ bundle exec rubocop -V
1.56.2 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.2.2) [arm64-darwin22]
  - rubocop-performance 1.19.0
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.23.0
```

owst:
We were using nodes as Hash keys, which is not supported - the metadata (source location) is not used to distinguish nodes. Thus the two occurrences of the same send (method name and args) were collapsed into one, causing only one of two offense sites to be reported.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_fix_12168_fix_bug_in_style_arguments_forwarding.md
+++ b/changelog/fix_fix_12168_fix_bug_in_style_arguments_forwarding.md
@@ -1,0 +1,1 @@
+* [#12168](https://github.com/rubocop/rubocop/issues/12168): Fix bug in `Style/ArgumentsForwarding` when there are repeated send nodes. ([@owst][])
--- a/lib/rubocop/cop/style/arguments_forwarding.rb
+++ b/lib/rubocop/cop/style/arguments_forwarding.rb
@@ -116,11 +116,11 @@ module RuboCop
         end
 
         def only_forwards_all?(send_classifications)
-          send_classifications.each_value.all? { |c, _, _| c == :all }
+          send_classifications.all? { |_, c, _, _| c == :all }
         end
 
         def add_forward_all_offenses(node, send_classifications, forwardable_args)
-          send_classifications.each do |send_node, (_c, forward_rest, _forward_kwrest)|
+          send_classifications.each do |send_node, _c, forward_rest, _forward_kwrest|
             register_forward_all_offense(send_node, send_node, forward_rest)
           end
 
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -116,7 +116,7 @@ def extract_forwardable_args(args)
         end
 
         def only_forwards_all?(send_classifications)
-          send_classifications.each_value.all? { |c, _, _| c == :all }
+          send_classifications.each.all? { |_, (c, _, _)| c == :all }
```
koic, line 119: It seems possible to simplify the block arguments.

```suggestion
          send_classifications.to_h.each_values.all? { |c, _, _| c == :all }
```

owst: I ended up with a slightly different simplification - I realised I don't need the nested array structure, so can remove the array destructuring in all of the `send_classifications` iteration block args

koic:

changelog/fix_fix_12168_fix_bug_in_style_arguments_forwarding.md
```diff
@@ -0,0 +1 @@
+* [#12168](https://github.com/rubocop/rubocop/pull/12168): Fix bug in `Style/ArgumentsForwarding` when there are repeated send nodes. ([@owst][])
```
koic, line 1:
```suggestion
* [#12168](https://github.com/rubocop/rubocop/issues/12168): Fix bug in `Style/ArgumentsForwarding` when there are repeated send nodes. ([@owst][])
```

koic:
Thanks!

PR #12174, opened 2023-08-31
[Fix #12172] Fix a false positive for `Style/EmptyCaseCondition`

### LINKED ISSUE #12172: Style/EmptyCaseCondition creates syntax error on autocorrect when following a return
## Expected behavior

I don't know. Certainly rubocop corrections should not lead to syntax errors.

## Actual behavior

```ruby
def foo(bar)
  return case
         when bar == 1
           2
         when bar != 12
           3
         else
           4
         end
end
```

is corrected to

```ruby
def foo(bar)
  return if bar == 1
           2
         elsif bar != 12
           3
         else
           4
         end
end
```

which is a syntax error, as the "return if bar == 1" is parsed as a postfix if.

see above

## RuboCop version

```
$ [bundle exec] rubocop -V
1.42.0 (using Parser 3.1.3.0, rubocop-ast 1.24.1, running on ruby 3.1.3) +server [x86_64-linux] 
```

koic:
Fixes #12172.

This PR fixes a false positive for `Style/EmptyCaseCondition` when using `return`, `break`, `next` or method call before empty case condition.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/empty_case_condition_spec.rb
+++ b/spec/rubocop/cop/style/empty_case_condition_spec.rb
@@ -240,22 +240,22 @@ RSpec.describe RuboCop::Cop::Style::EmptyCaseCondition, :config do
     context 'when used as an argument of a method without comment' do
       let(:source) do
         <<~RUBY
-          do_some_work case
-                       ^^^^ Do not use empty `case` condition, instead use an `if` expression.
-                       when object.nil?
-                         Object.new
-                       else
-                         object
-                       end
+          case
+          ^^^^ Do not use empty `case` condition, instead use an `if` expression.
+          when object.nil?
+            Object.new
+          else
+            object
+          end
         RUBY
       end
       let(:corrected_source) do
         <<~RUBY
-          do_some_work if object.nil?
-                         Object.new
-                       else
-                         object
-                       end
+          if object.nil?
+            Object.new
+          else
+            object
+          end
         RUBY
       end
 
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/empty_case_condition_spec.rb
```diff
@@ -266,67 +266,23 @@ def example
...
-                 else object
-                 end
-        RUBY
-      end
-      let(:corrected_source) do
-        <<~RUBY
-          return if object.nil?
-           Object.new
-                 else object
-                 end
```
koic, line 307: This test case is removed because this autocorrected syntax is invalid. In other words, it's a bug.

spec/rubocop/cop/style/empty_case_condition_spec.rb
```diff
@@ -266,67 +266,23 @@ def example
...
-                       else object
-                       end
-        RUBY
-      end
-      let(:corrected_source) do
-        <<~RUBY
-          do_some_work if object.nil?
-           Object.new
-                       else object
-                       end
```
koic, line 329: Ditto.

spec/rubocop/cop/style/empty_case_condition_spec.rb
```diff
@@ -240,22 +240,22 @@ def example
...
-                       else
-                         object
-                       end
+          case
+          ^^^^ Do not use empty `case` condition, instead use an `if` expression.
+          when object.nil?
+            Object.new
+          else
+            object
+          end
```
koic, line 249: NOTE: These tests have been updated to use valid syntax.

PR #12176, opened 2023-09-01
Add support for `Style/RedundantConditional` when redundant returning of true/false only if/unless block

ydah:
Although detected  https://github.com/rubocop/rubocop-rspec/pull/1692#discussion_r1288432020 , the following cases could also be supported. WDYT?

```ruby
# bad
true if x == y

# bad
if x == y
  true
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_add_support_for_style_redundant_conditional.md
+++ b/changelog/change_add_support_for_style_redundant_conditional.md
@@ -1,0 +1,1 @@
+* [#12176](https://github.com/rubocop/rubocop/pull/12176): Add support for `Style/RedundantConditional` when redundant returning of true/false only if/unless block. ([@ydah][])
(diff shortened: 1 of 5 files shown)


koic:
Hm, the return value of `x == y` is `true` or `false`. On the other hand, the return value of `true if x == y` is `true` or `nil`. 
Furthermore, the return value of `false if x = y` is either `false` or `nil`. It's always falsey. That is, they are not compatible and probably an unexpected behavior at all.

ydah:
Ah, I see. That's for sure. How about changing it to `SafeAutoCorrect: false` in that case? Or how about making only this part a separate cop?

koic:
I feel like at least this cop safety should not be sacrificed. And I'm not sure if it's worth extracting into a separate cop.

ydah:
> I feel like at least this cop safety should not be sacrificed. And I'm not sure if it's worth extracting into a separate cop.

Indeed it is. Thank you for replying. I'll close this PR.

bquorning:
We could keep the proposed changes to lib/rubocop/cop/layout/indentation_width.rb and lib/rubocop/cop/layout/space_inside_parens.rb though.

Issue #11920, opened 2023-06-02
Add API for checking gem versions


amomchilov:
There are many gem-specific Rubocop gems:

* https://github.com/rubocop/rubocop-rails
* https://github.com/rubocop/rubocop-rspec
* https://github.com/rubocop/rubocop-minitest
* https://github.com/rubocop/rubocop-factory_bot
* https://github.com/rubocop/rubocop-rake
* https://github.com/DmitryTsepelev/rubocop-graphql
* https://github.com/Shopify/rubocop-sorbet
* ... and so on

Some cops are only applicable to new versions of a gem, e.g. to replace a deprecated API or pattern with its modern replacement. https://github.com/rubocop/rubocop-rails does this [particularly often](https://github.com/search?q=repo%3Arubocop%2Frubocop-rails+minimum_target_rails_version&type=code).

The challenge is that getting the user's version of a gem is quite tricky. You might think to try something like this:

```ruby
Bundler.load.specs["some_gem"].map(&:version).max
```

But it often won't work, because users frequently execute `rubocop` directly (and not with `bundle exec rubocop`).

Interestingly, Rubocop resorts to scanning the `Gemfile.lock` file directly:

https://github.com/rubocop/rubocop/blob/1181d4ebad5f71c586f9514d9c341cdfffc1957d/lib/rubocop/config.rb#L293-L308

Even more strangely, this rails-specific logic is in the base `rubocop` gem, and not `rubocop-rails`:

https://github.com/rubocop/rubocop/blob/1181d4ebad5f71c586f9514d9c341cdfffc1957d/lib/rubocop/cop/team.rb#L159-L168

It would be nice to have API that lets you declare a Cop's gem requirement, e.g.

```ruby
class MyCop < RuboCop::Cop::Base
  requires_gem "rails", "1.2.3" # Only apply this cop to 1.2.3 and newer
end
```

amomchilov:
Here's an example of the kind of implementation required to achieve this today: https://github.com/Shopify/rubocop-sorbet/commit/94b6816dc8ce8497fa29dc8c6ba6f75019f81e18

PR #12180, opened 2023-09-02
Replace regex with `Bundler::LockfileParser`

amomchilov:
Depends on #12186, see the difference: https://github.com/amomchilov/rubocop/compare/add-requires_gem-api...parse-lockfile-with-bundler

Now that we have a Hash of all of the target's gems, we can just use that `target_rails_version_from_bundler_lock_file`, and remove the manual regex parsing.

-----------------

* [ ] ~Added tests.~

### DIFF
--- a/lib/rubocop/config.rb
+++ b/lib/rubocop/config.rb
@@ -306,16 +306,22 @@ module RuboCop
 
     # @return [Float, nil] The Rails version as a `major.minor` Float.
     def read_rails_version_from_bundler_lock_file
-      lock_file_path = bundler_lock_file_path
-      return nil unless lock_file_path
+      return nil unless gem_versions_in_target
 
-      File.foreach(lock_file_path) do |line|
-        # If Rails (or one of its frameworks) is in Gemfile.lock or gems.lock, there should be
-        # a line like:
-        #         railties (X.X.X)
-        result = line.match(/^\s+railties\s+\((\d+\.\d+)/)
-        return result.captures.first.to_f if result
-      end
+      # Look for `railties` instead of `rails`, to support apps that only use a subset of `rails`
+      # See https://github.com/rubocop/rubocop/pull/11289
+      rails_version_in_target = gem_versions_in_target['railties']
+      return nil unless rails_version_in_target
+
+      gem_version_to_major_minor_float(rails_version_in_target)
+    end
+
+    # @param [Gem::Version] gem_version an object like `Gem::Version.new("7.1.2.3")`
+    # @return [Float] The major and minor version, like `7.1`
+    def gem_version_to_major_minor_float(gem_version)
+      segments = gem_version.canonical_segments
+      # segments.fetch(0).to_f + (segments.fetch(1, 0.0).to_f / 10)
+      Float("#{segments.fetch(0)}.#{segments.fetch(1, 0)}")
     end
 
     # @returns [Hash{String => Gem::Version}] The locked gem versions, keyed by the gems' names.
--- a/changelog/change_replace_regex_with_BundlerLockfileParser.md
+++ b/changelog/change_replace_regex_with_BundlerLockfileParser.md
@@ -1,0 +1,1 @@
+* [#12180](https://github.com/rubocop/rubocop/pull/12180): Replace regex with `Bundler::LockfileParser`. ([@amomchilov][])


koic:
Is the `Bundler::LockfileParser` a stable published API that can be used safely? I'm concerned about whether it's a private API intended for internal use within Bundler, which might introduce breaking changes in future Bundler versions.

amomchilov:
Hey @koic!

I don't see any firm evidence that it's intended to be public (e.g. YARD docs), but I think it's very likely to be the case.

* It's [mentioned once in the changelog](https://github.com/rubygems/rubygems/blob/313bf1eb3b86888e60a524c027cd23a8bf028b32/bundler/CHANGELOG.md?plain=1#L69C63-L69C109) (which would be unusual if it was intended to be only an internal implementation detail)
* There are [a lot of usages on GitHub](https://github.com/search?q=%22Bundler%3A%3ALockFileParser%22+-path%3A**%2Fbundler.rb+-path%3A**%2Flockfile_parser.rb+language%3ARuby&type=code&p=1&l=Ruby), including some pretty prominant gems
    * Ruby/ruby: https://github.com/ruby/ruby/blob/3602e253e79b796661168b458e22ba1694ec06e5/lib/bundler/self_manager.rb#L171
    * Rails/spring: https://github.com/rails/spring/pull/492
    
Package management is central to an ecosystem, and being able to introspect dependencies is pretty core to a bunch of tooling. If it's not public, we should perhaps try to make it public, like how `Gem::Version` is used throughout the ecosystem today.

Being dependant on `Bundler::LockfileParser` isn't well justified by this PR on its own (because it's just for one gem, for an API that already works today). I'll have a PR out soon which uses it to read _all_ gem versions of a target, where this parser becomes pretty well justified (compared to trying to roll our own, just for Rubocop). I'll show you my draft when I'm done :)

lib/rubocop/config.rb
```diff
@@ -284,27 +284,45 @@ def pending_cops
...
+    end
+
     def inspect # :nodoc:
       "#<#{self.class.name}:#{object_id} @loaded_path=#{loaded_path}>"
     end
 
     private
 
     def target_rails_version_from_bundler_lock_file
-      @target_rails_version_from_bundler_lock_file ||= read_rails_version_from_bundler_lock_file
```
amomchilov, line 294: I removed this memoization because:

1. If `read_rails_version_from_bundler_lock_file` returned `nil`, it wouldn't memoize correctly and would be trying to parse the lockfile on every call anyway
2. This value is only accessed from `#target_rails_version`, which is itself memoized

amomchilov:
Hey @koic!

I don't see any firm evidence that it's intended to be public (e.g. YARD docs), but I think it's very likely to be the case.

* It's [mentioned once in the changelog](https://github.com/rubygems/rubygems/blob/313bf1eb3b86888e60a524c027cd23a8bf028b32/bundler/CHANGELOG.md?plain=1#L69C63-L69C109) (which would be unusual if it was intended to be only an internal implementation detail)
* There are [a lot of usages on GitHub](https://github.com/search?q=%22Bundler%3A%3ALockFileParser%22+-path%3A**%2Fbundler.rb+-path%3A**%2Flockfile_parser.rb+language%3ARuby&type=code&p=1&l=Ruby), including some pretty prominant gems:
    * Ruby/ruby: https://github.com/ruby/ruby/blob/3602e253e79b796661168b458e22ba1694ec06e5/lib/bundler/self_manager.rb#L171
    * Rails/spring: https://github.com/rails/spring/pull/492
    
Package management is central to an ecosystem, and being able to introspect dependencies is pretty core to a bunch of tooling. If it's not public, we should perhaps try to make it public, like how `Gem::Version` is used throughout the ecosystem today.

Being dependant on `Bundler::LockfileParser` isn't well justified by this PR on its own, but I needed it for `#12186`. Once that's implemented, we can use it here for "free".

koic:
@deivid-rodriguez @hsbt May I ask the Bundler development team, Is `Bundler::LockfileParser` intended to be used as a published API? I'm concerned about potential incompatibilities arising from the use of private API.

deivid-rodriguez:
Hei @koic. `Bundler::LockfileParser` is not documented anywhere as a public API, but I've seen many cases of people using it. So We don't plan to intentionally break things for people. That said, given it's not documented at the moment, it should still be "used at your own risk". We could probably formalize an API at some point.

deivid-rodriguez:
No problem!

bbatsov:
I don't mind using `Bundler::LockfileParser`. @amomchilov Let us know when you feel the PR is ready for review.

amomchilov:
@bbatsov добър вечер!

I've just rebased it. This PR builds on top of #12186, which needs to be reviewed and shipped first. After that, I think this PR is also ready for review/merge (just look at the last commit)

koic:
The commits related to the changes in #12180 are mixed in. Can you leave the change for only "Replace regex with `Bundler::LockfileParser`"?

koic:
It just occurred to me that RuboCop doesn't have a runtime dependency on Bundler. This means to utilize Bundler's API, a runtime dependency on Bundler would be required. However, introducing such a dependency solely for the sake of using `Bundler::LockfileParser` seems like an overkill.

Perhaps continuing to resolve this with regexp would be a more appropriate approach. In other words, it might be better if this change were not made.

amomchilov:
> The commits related to the changes in https://github.com/rubocop/rubocop/pull/12180 are mixed in.

Hey @koic, did you mean #12186? This "Replace regex with `Bundler::LockfileParser`" change _depends on_ the code introduced in the other PR. The other PR would need to be merged first. Could you please take a look?

> However, introducing such a dependency solely for the sake of using `Bundler::LockfileParser` seems like an overkill.
> 
> Perhaps continuing to resolve this with regexp would be a more appropriate approach. In other words, it might be better if this change were not made.

@koic For this change on its own, I would agree. But the real benefit here is new `requires_gem` feature I've proposed and implemented in #12186. Once that PR introduces the ability to parse the lockfile, this PR just becomes a natural next step: to replace this manual lockfile parsing with the results we've already got from #12186.

bbatsov:
I agree it's not prudent to add a runtime dependency on Bundler, so I see three paths ahead:

- Using this parser only when available
- Inlining something similar into RuboCop (I'm not sure how complex its implementation is)
- Keeping things as they are

amomchilov:
@bbatsov You raise an interesting point, I'll think on it!

In the mean time, could I get your thoughts on this [feature request](https://github.com/rubocop/rubocop/issues/11920) and [the implementation I proposed for it](https://github.com/rubocop/rubocop/pull/12186)?

Is the `requires_gem` API useful enough to justify the bundler dependency? Or perhaps I should extract it into a module that you mix into your cops as-needed, and then make the bundler dependency only required for that.

amomchilov:
Just ran into this, apparently this is already used by RuboCop

https://github.com/rubocop/rubocop/blob/master/lib/rubocop/lockfile.rb#L30-L38

bbatsov:
Ah, excellent - so I guess you know how to proceed now. :-)

bbatsov:
@amomchilov Any updates here?

amomchilov:
Hey @bbatsov I had a question about how to proceed. Imessaged you in the RuboCop discord (I figured that was a good spot, because it showed you as online). Cross posting:

——-

Добър вечер @bbatsov !

I'm taking a look at using [the `RuboCop::Lockfile` abstraction](https://github.com/rubocop/rubocop/blob/main/lib/rubocop/lockfile.rb) that already exists in the codebase. It has its own logic to find the relevant lockfile for the repo (`Bundler.default_lockfile`) rather than the custom logic that exists in [`RuboCop::Config#bundler_lock_file_path`](https://github.com/rubocop/rubocop/blob/d3fb79b24c03b1b2ee1afe8e87b66564761e21f1/lib/rubocop/config.rb#L264-L273).

They both do some directory traversal to find a gemfile in a parent dir, if it's not found in the cwd. Not sure if they're exactly the same, though :thinking:

I'm not sure how to proceed. Do you think I should modify `RuboCop::Lockfile`  to take in the path as a param (as determined by the existing `RuboCopConfig#bundler_lock_file_path` logic, or should I just use whatever `RuboCop::Lockfile` does already?

bbatsov:
> I'm not sure how to proceed. Do you think I should modify RuboCop::Lockfile to take in the path as a param (as determined by the existing RuboCopConfig#bundler_lock_file_path logic), or should I just use whatever RuboCop::Lockfile does already?

I'd be fine with both approaches, but the first one sounds slightly better to me.

amomchilov:
@bbatsov Can we move the lock-file parsing related discussion to the underlying PR? That's what introduces that logic, it's an impl detail separate from the changes in this PR.

I responded here: https://github.com/rubocop/rubocop/pull/12186#pullrequestreview-1913106732

bbatsov:
I guess you can rebase this one now that #12186 is finally merged.

amomchilov:
@bbatsov Rebased, and ready to merge!

Thanks for jumping in to merge all these pieces :)

PR #12186, opened 2023-09-03
Add new `requires_gem` API for declaring which gems a Cop needs

### LINKED ISSUE #11920: Add API for checking gem versions

amomchilov:
Fixes #11920.

This PR is used by:

1. https://github.com/rubocop/rubocop/pull/12180
2. https://github.com/rubocop/rubocop-rails/pull/1137

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -269,6 +269,7 @@ module RuboCop
       end
 
       def relevant_file?(file)
+        return false unless target_satisfies_all_gem_version_requirements?
         return true unless @config.clusivity_config_for_badge?(self.class.badge)
 
         file == RuboCop::AST::ProcessedSource::STRING_SOURCE_NAME ||
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -70,6 +70,7 @@ module RuboCop
 
       def self.inherited(subclass)
         super
+        subclass.instance_variable_set(:@gem_requirements, gem_requirements.dup)
         Registry.global.enlist(subclass)
       end
 
@@ -124,6 +125,29 @@ module RuboCop
       # multiple times with the same `processed_source.path` but different content.
       def self.support_multiple_source?
         false
+      end
+
+      ## Gem requirements
+
+      @gem_requirements = {}
+
+      class << self
+        attr_reader :gem_requirements
+
+        # Register a version requirement for the given gem name.
+        # This cop will be skipped unless the target satisfies *all* requirements.
+        # @param [String] gem_name
+        # @param [Array<String>] version_requirements The version requirements,
+        #   using the same syntax as a Gemfile, e.g. ">= 1.2.3"
+        #
+        #   If omitted, any version of the gem will be accepted.
+        #
+        #   https://guides.rubygems.org/patterns/#declaring-dependencies
+        #
+        # @api public
+        def requires_gem(gem_name, *version_requirements)
+          @gem_requirements[gem_name] = Gem::Requirement.new(version_requirements)
+        end
       end
 
       def initialize(config = nil, options = nil)
@@ -495,6 +519,18 @@ module RuboCop
           range.begin_pos + @current_offset,
           range.end_pos + @current_offset
         )
+      end
+
+      def target_satisfies_all_gem_version_requirements?
+        self.class.gem_requirements.all? do |gem_name, version_req|
+          all_gem_versions_in_target = @config.gem_versions_in_target
+          next false unless all_gem_versions_in_target
+
+          gem_version_in_target = all_gem_versions_in_target[gem_name]
+          next false unless gem_version_in_target
+
+          version_req.satisfied_by?(gem_version_in_target)
+        end
       end
     end
   end
--- a/lib/rubocop/cop/team.rb
+++ b/lib/rubocop/cop/team.rb
@@ -174,6 +174,9 @@ module RuboCop
       end
 
       def support_target_rails_version?(cop)
+        # In this case, the rails version was already checked by `#excluded_file?`
+        return true if defined?(RuboCop::Rails::TargetRailsVersion::USES_REQUIRES_GEM_API)
+
         return true unless cop.class.respond_to?(:support_target_rails_version?)
 
         cop.class.support_target_rails_version?(cop.target_rails_version)
--- a/lib/rubocop/lockfile.rb
+++ b/lib/rubocop/lockfile.rb
@@ -5,14 +5,16 @@ module RuboCop
   # Does not actually resolve gems, just parses the lockfile.
   # @api private
   class Lockfile
-    # Gems that the bundle depends on
+    # Gems that the bundle directly depends on.
+    # @return [Array<Bundler::Dependency>, nil]
     def dependencies
       return [] unless parser
 
       parser.dependencies.values
     end
 
-    # All activated gems, including transitive dependencies
+    # All activated gems, including transitive dependencies.
+    # @return [Array<Bundler::Dependency>, nil]
     def gems
       return [] unless parser
 
@@ -21,12 +23,16 @@ module RuboCop
       parser.dependencies.values.concat(parser.specs.flat_map(&:dependencies))
     end
 
+    # Whether this lockfile includes the named gem, directly or indirectly.
+    # @param [String] name
+    # @return [Boolean]
     def includes_gem?(name)
       gems.any? { |gem| gem.name == name }
     end
 
     private
 
+    # @return [Bundler::LockfileParser, nil]
     def parser
       return unless defined?(Bundler) && Bundler.default_lockfile
       return @parser if defined?(@parser)
--- a/lib/rubocop/lockfile.rb
+++ b/lib/rubocop/lockfile.rb
@@ -30,6 +30,23 @@ module RuboCop
       parser.dependencies.values.concat(parser.specs.flat_map(&:dependencies))
     end
 
+    # Returns the locked versions of gems from this lockfile.
+    # @param [Boolean] include_transitive_dependencies: When false, only direct dependencies
+    #   are returned, i.e. those listed explicitly in the `Gemfile`.
+    # @returns [Hash{String => Gem::Version}] The locked gem versions, keyed by the gems' names.
+    def gem_versions(include_transitive_dependencies: true)
+      return {} unless parser
+
+      all_gem_versions = parser.specs.to_h { |spec| [spec.name, spec.version] }
+
+      if include_transitive_dependencies
+        all_gem_versions
+      else
+        direct_dep_names = parser.dependencies.keys
+        all_gem_versions.slice(*direct_dep_names)
+      end
+    end
+
     # Whether this lockfile includes the named gem, directly or indirectly.
     # @param [String] name
     # @return [Boolean]
--- a/lib/rubocop/lockfile.rb
+++ b/lib/rubocop/lockfile.rb
@@ -5,6 +5,13 @@ module RuboCop
   # Does not actually resolve gems, just parses the lockfile.
   # @api private
   class Lockfile
+    # @param [String, Pathname, nil] lockfile_path
+    def initialize(lockfile_path = nil)
+      lockfile_path ||= defined?(Bundler) ? Bundler.default_lockfile : nil
+
+      @lockfile_path = lockfile_path
+    end
+
     # Gems that the bundle directly depends on.
     # @return [Array<Bundler::Dependency>, nil]
     def dependencies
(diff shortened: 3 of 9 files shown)


lib/rubocop/cop/base.rb
```diff
@@ -126,6 +127,36 @@ def self.support_multiple_source?
         false
       end
 
+      ## Gem requirements
+
+      # rubocop:disable Layout/ClassStructure
+      class << self
+        attr_reader :gem_requirements
```
amomchilov, line 134: I'm not quite sure what the correct layout is.

I also see `class << self` is pretty rare in this codebase, but `singleton_class.attr_reader(:gem_requirements)` would be weird.

lib/rubocop/cop/base.rb
```diff
@@ -70,6 +70,7 @@ def self.documentation_url
 
       def self.inherited(subclass)
         super
+        subclass.instance_variable_set(:@gem_requirements, {})
```
amomchilov, line 73: Would it be better to put `@gem_requirements ||= {}` everywhere it's referenced, instead?

Or replace the `attr_reader` with an explicit method that does this?

```diff
-attr_reader :gem_requirements
+def gem_requirements
+  @gem_requirements ||= {}
+end
```

lib/rubocop/cop/base.rb
```diff
@@ -126,6 +127,36 @@ def self.support_multiple_source?
...
+        attr_reader :gem_requirements
+
+        # Register a version requirement for the given gem name.
+        # This cop will be skipped unless the target satisfies *all* requirements.
+        # @param [String] gem_name
+        # @param [String] version_requirement The version requirement,
+        #   using the same syntax as a Gemfile, e.g. ">= 1.2.3"
+        #
+        # @api public
+        def requires_gem(gem_name, version_requirement)
```
amomchilov, line 143: Should this be in a mixin, like [`rubocop-rails`'s `TargetRailsVersion`](https://github.com/rubocop/rubocop-rails/blob/master/lib/rubocop/cop/mixin/target_rails_version.rb)?

It'll keep the logic nicely separated and tested in its own file, but it means [we couldn't use it from `target_rails_version_from_bundler_lock_file`](https://github.com/rubocop/rubocop/pull/12180) which is already in the Base.

spec/rubocop/config_spec.rb
```diff
@@ -825,6 +825,56 @@ def cop_enabled(cop_class)
...
+
+          create_file(lockfile_path, content)
+          expect(configuration.gem_versions_in_target).to eq expected
+        end
+      end
+    end
+
+    context 'and neither Gemfile.lock nor gems.locked exist' do
+      it 'returns nil' do
+        expect(configuration.gem_versions_in_target.nil?).to be(true)
```
amomchilov, line 873: I'm not used to RSpec, is this correct?

I originally had:

```suggestion
        expect(configuration.gem_versions_in_target).to be_nil
```

But rubocop auto-corrected to this. It feels a bit... funky

lib/rubocop/cop/team.rb
```diff
@@ -174,6 +174,12 @@ def support_target_ruby_version?(cop)
       end
 
       def support_target_rails_version?(cop)
+        if cop.class.respond_to?(:gem_requirements)
```
amomchilov, line 177: It looks like the versioning between `rubocop` and `rubocop-rails` gets a bit tricky here.

I'd love to delete this method entirely, but I think it still needs to support calling `support_target_rails_version?` on older versions that don't yet have the `requires_gem` and `gem_requirements` API

amomchilov:
Hey @koic, I got all the tests to pass, but I'm having some issues with the class layout. I think this PR is ready for an early review.

I'd appreciate any early thoughts or feedback you might have!

lib/rubocop/cop/team.rb
```diff
@@ -174,11 +174,28 @@ def support_target_ruby_version?(cop)
...
       end
 
+      # Returns true if the current version of rubocop-rails uses `requires_gem` to specify
+      # its gem requirements.
+      def rubocop_rails_uses_requires_gem?
+        @rubocop_rails_uses_requires_gem ||= begin
+          rubo_rails_version = RuboCop::Rails::Version::STRING if defined?(RuboCop::Rails)
+
+          if rubo_rails_version
+            Gem::Version.new('2.21.0') <= Gem::Version.new(rubo_rails_version)
```
amomchilov, line 192: It looks like the versioning between `rubocop` and `rubocop-rails` gets a bit tricky here. 

In the future, `TargetRailsVersion` can just be a wrapper for `requires_gem`. When that happens, any cops that don't support the target's rails version would have already be filtered out by `#excluded_file?` called by `#roundup_relevant_cops` above. I'm assuming this will be done in version `2.21.0`.

For older versions of rubocop-rails, that won't be correct, and we'll still need to call their `support_target_rails_version?` method.

spec/rubocop/config_spec.rb
abzuar9658: That correct! ` expect(configuration.gem_versions_in_target).to be_nil`

lib/rubocop/cop/team.rb
```diff
@@ -174,6 +174,9 @@ def support_target_ruby_version?(cop)
       end
 
       def support_target_rails_version?(cop)
+        # In this case, the rails version was already checked by `#excluded_file?`
+        return true if defined?(RuboCop::Rails::TargetRailsVersion::USES_REQUIRES_GEM_API)
```
amomchilov, line 178: It looks like the versioning between `rubocop` and `rubocop-rails` gets a bit tricky here.

1. This new version of `rubocop` needs to support older versions of `rubocop-rails`.
2. Inversely, the new version of `rubocop-rails` needs to support older versions of `rubocop`.

I've tested both scenarios, and they both work well.

I added this `USES_REQUIRES_GEM_API` constant to detect new `rubocop-rails` versions, because it was more explicit/clear than checking against some specific version number.

In the future, `TargetRailsVersion` can just be a wrapper for `requires_gem`. When that happens, any cops that don't support the target's rails version would have already be filtered out by `#excluded_file?` called by `#roundup_relevant_cops` above.

For older versions of `rubocop-rails`, that won't be correct, and we'll still need to call their `support_target_rails_version?` method.

amomchilov:
Hey @bbatsov and @koic, I think this is ready for review. This was a bit of a tricky change for my, especially because it spans both `rubocop` and `rubocop-rails`.

I've tested it pretty thoroughly, and it all works correctly, but I would still be open to any feedback!

Cheers

bbatsov:
Sorry for the slow turnaround here - I'm fine with the general direction, but I'll need some time before I can take a closer look at the code, as I've been very busy lately at work and wrapping up releases of some other OSS projects. 

At a glance the one things that's currently missing is some user documentation for the proposed new feature.

lib/rubocop/cop/base.rb
bbatsov: I'll have to think about this more, but I can live with the current proposal. 

There's also the question whether we want this to be able to work with multiple gems, but I'm guessing the likelihood that a cop depends on multiple gems is pretty tiny, so let's keep it simple for now.

bbatsov:
It'd also be nice to benchmark the impact of using `require_gem` on several cops, as speed is something quite important for a tool like RuboCop.

lib/rubocop/lockfile.rb
```diff
@@ -5,14 +5,16 @@ module RuboCop
   # Does not actually resolve gems, just parses the lockfile.
   # @api private
   class Lockfile
-    # Gems that the bundle depends on
+    # Gems that the bundle directly depends on.
+    # @return [Array<Bundler::Dependency>, nil]
```
amomchilov, line 16: These methods' semantics aren't obvious (in particular, whether they include indirect dependencies or not, and what type of results they vend), so I documented them all.

The whole class is marked `@api private`, which get inherited by all of these methods, so there's no risk of people confusing them for public API (even if they're documented here).

lib/rubocop/lockfile.rb
```diff
@@ -5,6 +5,13 @@ module RuboCop
   # Does not actually resolve gems, just parses the lockfile.
   # @api private
   class Lockfile
+    # @param [String, Pathname, nil] lockfile_path
+    def initialize(lockfile_path: nil)
+      lockfile_path ||= defined?(Bundler) ? Bundler.default_lockfile : nil
```
amomchilov, line 10: I parameterized this class to let you specify your own `lockfile_path`, but kept the default behaviour the same as before.

This let's use it with the result from `RuboCop::Config#bundler_lock_file_path`.

lib/rubocop/lockfile.rb
```diff
@@ -30,6 +30,23 @@ def gems
       parser.dependencies.values.concat(parser.specs.flat_map(&:dependencies))
     end
 
+    # Returns the locked versions of gems from this lockfile.
+    # @param [Boolean] include_transitive_dependencies: When false, only direct dependencies
+    #   are returned, i.e. those listed explicitly in the `Gemfile`.
+    # @returns [Hash{String => Gem::Version}] The locked gem versions, keyed by the gems' names.
+    def gem_versions(include_transitive_dependencies: true)
```
amomchilov, line 37: The pre-existing methods (`#dependencies` and `#spec`) return `Gem::Dependency` objects, which describe the version _requirements_, but not the resolved versions.

This method returns the actual resolved versions, locked by the lock file.

This is what we check gem requirements against.

amomchilov (commented):
@bbatsov I had to do a little bit of surgery, but I got this to work with the existing `RuboCop::Lockfile` wrapper for `Bundler::LockfileParser`.

I left some comments to explain my process, let me know what you think!

Still TODO:

- [x] Write user docs – I can do this
- [ ] Benchmark this for perf regression – I'm not sure what the best way to do this is. Let me know if you have any suggestions

bbatsov:
> Benchmark this for perf regression – I'm not sure what the best way to do this is. Let me know if you have any suggestions

You can simply test on some bigger Ruby repo (e.g. Rails). 

@jonas054 @koic Please, take a look as well.

amomchilov:
Rebased to fix the unrelated style failure, which was fixed by #12737

jonas054:
Looks good to me. Some of it goes over my head a bit, but it looks like well-written and quite readable code. Good job on the documentation too.

amomchilov:
@jonas054 Would you be okay to share which parts "go over your head a bit"? I'd love to clarify anything that might be confusing.

jonas054:
No, don't worry about that comment, @amomchilov. That was much more about my abilities to review code in parts of RuboCop that I'm only partly familiar with. I don't see any specific opportunities for improvement when it comes to clarity.

amomchilov:
Ran a quick benchmark against Rails, which shows no discernable performance difference:

On the master branch:

```zsh
$ time ../rubocop/rubocop/exe/rubocop --cache false
Inspecting 3265 files
...

3265 files inspected, no offenses detected
../rubocop/rubocop/exe/rubocop --cache false  61.37s user 0.77s system 95% cpu 1:04.84 total
```

On this branch:

```zsh
$ time ../rubocop/rubocop/exe/rubocop --cache false
Inspecting 3265 files
...
3265 files inspected, no offenses detected
../rubocop/rubocop/exe/rubocop --cache false  61.69s user 0.74s system 97% cpu 1:04.02 total
```

@koic I think this is for a final review and merge. Could you please have a look?

bbatsov:
If @koic doesn't share any feedback I'll merge this after the next bug-fix release.

PR #12211, opened 2023-09-20
[Fix #12210] Mark `Style/RedundantFilterChain` as unsafe autocorrect

### LINKED ISSUE #12210: Style/RedundantFilterChain is not safe
Autocorrect safe rubocop offences are not expected to cause breaking behaviour. Replacing the `select` with `any?` breaks behaviour on each loop iteration because `any?` stops iterating over the loop on true and `select` iterates over every item. 

--------

## Expected behavior

Rubocop autocorrects `Style/RedundantFilterChain` cop which breaks code functionality. 

## Actual behavior

Autocorrecting `Style/RedundantFilterChain` cop should not break code functionality. 

Rubocop autocorrects from:

```
module Test
  def test
    %w[1 2].select do |number|
      number.present? && action(number)
    end.any?
  end

  def action(number)
    # do action with number
    true
  end
end
```

into this:

```
module Test
  def test
    %w[1 2].any? do |number|
      number.present? && action(number)
    end
  end

  def action(number)
    # do action with number
    true
  end
end
```

These are not equal behaviours because the `any?` would stop iterating on the loop after processing `'1'`. Meaning that `action('2')` is never called. The `select` on the before behaviour ensures that the loop iterates over every item. 

** I realize that this code example is probably breaking some best code practices, but this is still valid code. 

## RuboCop version

```
$ [bundle exec] rubocop -V
1.56.3 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 2.7.7) [arm64-darwin22]
  - rubocop-rspec 2.24.0
```

koic:
Fixes #12210.

This PR marks `Style/RedundantFilterChain` as unsafe autocorrect.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_mark_style_redundant_filter_chain_as_unsafe_autocorrect.md
+++ b/changelog/change_mark_style_redundant_filter_chain_as_unsafe_autocorrect.md
@@ -1,0 +1,1 @@
+* [#12210](https://github.com/rubocop/rubocop/issues/12210): Mark `Style/RedundantFilterChain` as unsafe autocorrect. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

PR #12227, opened 2023-09-29
Add new `Style/SingleLineDoEndBlock` cop

koic:
This PR adds new `Style/SingleLineDoEndBlock` cop that checks for single-line `do`...`end` blocks.

```ruby
# bad
foo do |arg| bar(arg) end

# good
foo do |arg|
  bar(arg)
end

# bad
->(arg) do bar(arg) end

# good
->(arg) { bar(arg) }
```

In practice a single line `do`...`end` is autocorrected when `EnforcedStyle: semantic` in `Style/BlockDelimiters`. It can also be detected by this cop if it is written by handcraft. So I decided to introduce this new cop instead of an autocorrect extension of `Style/BlockDelimiters`.

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -5186,6 +5186,11 @@ Style/SingleLineBlockParams:
         - acc
         - elem
 
+Style/SingleLineDoEnd:
```
bbatsov, line 5189: I think there's a style guide entry that you can refer here.

lib/rubocop/cop/style/single_line_do_end.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for single-line `do`...`end` blocks.
```
bbatsov, line 6: You might also add a note about why the auto-correct doesn't replace `do/end` with `{}`, as perhaps this won't be clear for some people.

config/default.yml
koic: Sure! I've opened https://github.com/rubocop/ruby-style-guide/pull/922 to add a new rule for this one.

lib/rubocop/cop/style/single_line_do_end.rb
koic: Indeed, I've added an explain for that.

bbatsov:
Thanks!

PR #12264, opened 2023-10-11
Fix `Style/RedundantDoubleSplatHashBraces` to handle nested hashes

fatkodima:
While tried to run on a `rails/rails` `Style/RedundantDoubleSplatHashBraces` cop, I got a false positive for https://github.com/rails/rails/blob/3064a255c53e04a2f4885aa5d5134786ac3904e9/actionpack/test/dispatch/ssl_test.rb#L13.


koic:
Unfortunately, this doesn't provide a sufficient solution for the similar issue in #11263. I have opened #12268 to address these concerns. This PR will be closed, but thank you for tackling the issue.

PR #12272, opened 2023-10-13
[Fix #12271] Fix a false positive for `Lint/RedundantSafeNavigation`

### LINKED ISSUE #12271: RedundantSafeNavigation does not consider constants may be nil.
## Expected behavior

```
FOO = nil

self::FOO&.dig(...)
      ^^^^^
[Correctable] Lint/RedundantSafeNavigation: Redundant safe navigation detected.
```

## Actual behavior

I don't believe we should flag constants as not needing safe navigation.

## RuboCop version

```
1.57.0 (using Parser 3.2.2.4, rubocop-ast 1.29.0, running on ruby 3.2.2) [x86_64-linux]
  - rubocop-rspec 2.24.1

```

## Potential Fix

Consider deleting this line? https://github.com/rubocop/rubocop/blob/128618c7a8e3336bbe9ff8f800a73ff04dfc7a81/lib/rubocop/cop/lint/redundant_safe_navigation.rb#L76

koic:
Fixes #12271.

This PR fixes a false positive for `Lint/RedundantSafeNavigation` when using snake case constant receiver.

Camel case is used for naming classes and modules, while snake case is used for all other constants. This naming conforms with `Naming/ConstantName` cop.

Since constants that turn into classes or modules are normally not `nil`, they will continue to be detected.
However, this PR will update to allow safe navigation for constants in snake case.

This change resolves both issue https://github.com/rubocop/rubocop-rails/issues/1104 and #12271.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/redundant_safe_navigation.rb
+++ b/lib/rubocop/cop/lint/redundant_safe_navigation.rb
@@ -4,8 +4,8 @@ module RuboCop
   module Cop
     module Lint
       # Checks for redundant safe navigation calls.
-      # Use cases where a constant is `nil` are rare and an offense is detected
-      # when the receiver is a constant.
+      # Use cases where a constant, named in camel case for classes and modules is `nil` are rare,
+      # and an offense is not detected when the receiver is a snake case constant.
       #
       # For all receivers, the `instance_of?`, `kind_of?`, `is_a?`, `eql?`, `respond_to?`,
       # and `equal?` methods are checked by default.
@@ -26,7 +26,7 @@ module RuboCop
       #
       # @example
       #   # bad
-      #   Const&.do_something
+      #   CamelCaseConst&.do_something
       #
       #   # bad
       #   do_something if attrs&.respond_to?(:[])
@@ -40,7 +40,7 @@ module RuboCop
       #   end
       #
       #   # good
-      #   Const.do_something
+      #   CamelCaseConst.do_something
       #
       #   # good
       #   while node.is_a?(BeginNode)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/redundant_safe_navigation.rb
```diff
@@ -4,8 +4,8 @@ module RuboCop
   module Cop
     module Lint
       # Checks for redundant safe navigation calls.
-      # Use cases where a constant is `nil` are rare and an offense is detected
-      # when the receiver is a constant.
+      # Use cases where a constant, named in camel case for classes and modules is `nil` are rare,
+      # and an offense is detected when the receiver is a snake case constant.
```
bbatsov, line 8: I think you meant to write "is not detected" here.

bbatsov: Ah, you mean to target just `SCREAMING_SNAKE_CASE`. I see.

bbatsov:
Frankly, I'm not sure what's the best course of action here, as I don't understand why anyone would have a `nil` constant regardless of whether it's a class/module reference or not.

bbatsov:
P.S. Probably we need some cop about assigning `nil` to constants, as it makes very little sense to me.

lib/rubocop/cop/lint/redundant_safe_navigation.rb
koic: Oops! I've updated the text.

koic:
> P.S. Probably we need some cop about assigning `nil` to constants, as it makes very little sense to me. 

Hm, naming for magic literal, like magic number? 🤔 While I also don't have the use case for assigning `nil` to a constant, So I agree that it's worth considering.

spec/rubocop/cop/lint/redundant_safe_navigation_spec.rb
```diff
@@ -14,6 +14,12 @@
     RUBY
   end
 
+  it 'does not register an offense and corrects when `&.` is used for snake case const receiver' do
+    expect_no_offenses(<<~RUBY)
+      FOO&.do_something
```
bbatsov, line 19: Btw, I guess there's a small change that a class name will be one word, all caps - e.g. `URL`, `URI`, etc. I guess we can leave it as is, but are test cases for `CONST_NAME` and `Const_name`, so it's clearer what we are testing for exactly.

koic: Sure! I've updated the test cases.

bbatsov:
Thanks!

PR #12282, opened 2023-10-14
Do not accept abbreviated options for cli

### LINKED ISSUE #12280: option --disable-pending-cops is activated with as little as --disable-p
## Expected behavior

`rubocop --help` lists command: `--disable-pending-cops`
And invoking rubocop with that option does what it says it does.

## Actual behavior

`rubocop --disable-p` will also disable the pending cops.
`rubocop --disable-` or less will yield `ambiguous option: --disable-`

It looks like rubocop is using substring matches for all command line options?
Its kinda convenient to only type `--disable-pending` instead of the full option.
But this seems weird, I would assume any unrecognized options should throw an error, but I guess maybe this is intended and not a bug?

## RuboCop version

```
rubocop -V
1.57.1 (using Parser 3.2.2.4, rubocop-ast 1.29.0, running on ruby 2.7.8) [x86_64-darwin21]
  - rubocop-performance 1.19.1
  - rubocop-rails 2.21.2
```

fatkodima:
Fixes #12280.

Was also surprised that rubocop's cli accepts abbreviated options. I do not think this is a convenience, but more like a bug if you do not specify the correct option.

### DIFF
--- a/changelog/fix_do_not_accept_abbreviated_options_for_cli.md
+++ b/changelog/fix_do_not_accept_abbreviated_options_for_cli.md
@@ -1,0 +1,1 @@
+* [#12280](https://github.com/rubocop/rubocop/issues/12280): Do not accept abbreviated options for cli. ([@fatkodima][])
--- a/lib/rubocop/options.rb
+++ b/lib/rubocop/options.rb
@@ -53,6 +53,7 @@ module RuboCop
     def define_options
       OptionParser.new do |opts|
         opts.banner = rainbow.wrap('Usage: rubocop [options] [file1, file2, ...]').bright
+        opts.require_exact = true
 
         add_check_options(opts)
         add_cache_options(opts)
--- a/spec/rubocop/cli/options_spec.rb
+++ b/spec/rubocop/cli/options_spec.rb
@@ -2268,12 +2268,22 @@ RSpec.describe 'RuboCop::CLI options', :isolated_environment do # rubocop:disabl
   end
 
   describe 'option is invalid' do
-    it 'suggests to use the --help flag' do
+    it 'suggests to use the --help flag when non existing option' do
       invalid_option = '--invalid-option'
 
       expect(cli.run([invalid_option])).to eq(2)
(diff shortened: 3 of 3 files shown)


bbatsov:
Today I learned. 😆 I agree that's more of a bug, as I never meant for this to be the actual behavior.

koic:
Um, this change could potentially be a breaking change for some users. Whether the shorthand name is appropriate or not may be a matter of perspective.

bbatsov:
> Um, this change could potentially be a breaking change for some users. Whether the shorthand name is appropriate or not may be a matter of perspective.

I also thought of this, but I can't think of a single time I've seen someone use those shorthand options (in any application), so I don't think that's a big deal. Normally they are useful for commands that don't have a shorthand in the first place, but I think it RuboCop pretty much every command has an explicit shorthand.

changelog/fix_do_not_accept_abbreviated_options_for_cli.md
```diff
@@ -0,0 +1 @@
+* [#12280](https://github.com/rubocop/rubocop/issues/12280): Do not accept abbreviated options for cli. ([@fatkodima][])
```
koic, line 1:
```suggestion
* [#12280](https://github.com/rubocop/rubocop/issues/12280): **(Breaking)** Do not accept abbreviated options for cli. ([@fatkodima][])
```

If this change is made, it seems prudent to explicitly note `**(Breaking)**` in the changelog entry.

bbatsov:
We can also simply bucket this to the list of breaking changes liked up for a potential RuboCop 2.0 down the road.

koic:
Ah, might be overthinking on my part. Releasing it without waiting for RuboCop 2.0 could probably be tried. (e.g. next minor release)

fatkodima:
Agreed with both of you. Do not think anyone uses it. 

CI is broken, because seems like I have found a bug in `optionparser` - it does not handle options specified with 
`--[no-]something` with `require_exact = true`. Will investigate more.

PR #12303, opened 2023-10-21
[Fix #11652] Generate inherit_from correctly also with ERB

### LINKED ISSUE #11652: Regenerating `.rubocop_todo.yml` can modify `.rubocop.yml`
I'm ignoring rubocop violations in CI by only including the rubocop todo file when an environment variable is set.
I'm still interested in these failures locally, but don't want them to cause the jobs to fail.

When I recreate the todo file it shifts the include out of the conditional, which changes the meaning.

--------

## Expected behavior

It should leave the existing file alone.

## Actual behavior

The file gets modified to the following:

```yml
inherit_from: .rubocop_todo.yml

<% if ENV["CI"] %>

<% end %>
```

Have the following at the top in your `.rubocop.yml` and run `rubocop --auto-gen-config`

```yml
<% if ENV["CI"] %>
inherit_from: .rubocop_todo.yml
<% end %>
```

## RuboCop version

```
$ [bundle exec] rubocop -V
1.46.0 (using Parser 3.2.1.0, rubocop-ast 1.26.0, running on ruby 3.1.3) [x86_64-linux-musl]
  - rubocop-erb 0.2.3
  - rubocop-rails 2.17.4
```

jonas054:
The problem with the existing algorithm, for generating the `inherit_from` line in `--auto-gen-config` runs, was that we started by removing the `inherit_from` (if present) and then inserted it later.

The things we must deal with in the algorithm are
* there could be a pre-existing `inherit_from` in `.rubocop.yml`, or not
* there could be a YAML doc start `---`, or not
* there could be inheritance from one or multiple files

A more robust approach to the removing, generating, and re-inserting is to assume that any pre-existing `inherit_from` directive is already in the right place. By replacing it with a placeholder and later replacing the placeholder with the new inheritance directive, we will get it right without much effort.

### DIFF
--- a/changelog/fix_auto_gen_config_with_erb.md
+++ b/changelog/fix_auto_gen_config_with_erb.md
@@ -1,0 +1,1 @@
+* [#11652](https://github.com/rubocop/rubocop/issues/11652): Make `--auto-gen-config` generate `inherit_from` correctly inside ERB `if`. ([@jonas054][])
--- a/lib/rubocop/cli/command/auto_generate_config.rb
+++ b/lib/rubocop/cli/command/auto_generate_config.rb
@@ -10,6 +10,7 @@ module RuboCop
 
         AUTO_GENERATED_FILE = '.rubocop_todo.yml'
         YAML_OPTIONAL_DOC_START = /\A---(\s+#|\s*\z)/.freeze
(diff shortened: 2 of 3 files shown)


bbatsov:
Great fix!

PR #12306, opened 2023-10-24
[Fix #12274] Fix a false positive for `Lint/Void`

### LINKED ISSUE #12274: Lint/Void removing valid block content
Before:
```
    e.each do |_k, v|
      v == 'b'
    end
```
After autocorecction:
```
    e.each do |_k, v|
    end
```
In my case `e` is a enumerator that filters. Of course, that fails after autocorrection. But also, removing block contents and leaving a empty block is not so useful either.

koic:
Fixes #12274.

This PR fixes a false positive for `Lint/Void` when `each`'s receiver is an object of `Enumerator` to which `filter` has been applied.

`each` blocks are allowed to prevent false positives. The expression inside the following block is not void:

```ruby
enumerator = [1, 2, 3].filter
enumerator.each { |item| item >= 2 } #=> [2, 3]
```

This might introduce false negatives, but higher severity Lint cop should aim for as safe detections as possible, I think.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/void.rb
+++ b/lib/rubocop/cop/lint/void.rb
@@ -6,6 +6,16 @@ module RuboCop
       # Checks for operators, variables, literals, lambda, proc and nonmutating
       # methods used in void context.
       #
+      # `each` blocks are allowed to prevent false positives.
+      # For example, the expression inside the `each` block below.
+      # It's not void, especially when the receiver is an `Enumerator`:
+      #
+      # [source,ruby]
+      # ----
+      # enumerator = [1, 2, 3].filter
+      # enumerator.each { |item| item >= 2 } #=> [2, 3]
+      # ----
+      #
       # @example CheckForMethodsWithNoSideEffects: false (default)
       #   # bad
       #   def some_method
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/void.rb
```diff
@@ -6,6 +6,15 @@ module Lint
       # Checks for operators, variables, literals, lambda, proc and nonmutating
       # methods used in void context.
       #
+      # `each` blocks are allowed to prevent false positives.
```
bbatsov, line 9: You might expand on this a bit here - e.g. saying that the receiver might be an enumerator.

bbatsov:
Perhaps we should still flag an offense if the receiver is clearly not an enumerator - e.g. it's some literal or a constant (yeah, it can technically be an enumerator but this seems quite unlikely).

koic:
> it's some literal or a constant (yeah, it can technically be an enumerator but this seems quite unlikely).

Already supported by the implementation, but I added missing specs for these. And I've also tweaked the offense message I discovered with this additional spec. There was a message that should be a "Constant" rather than a "Variable". This intentionally separates the commits.

bbatsov:
Thanks!

PR #12313, opened 2023-10-25
[Fix #12310] Drop `base64` gem from runtime dependency

### LINKED ISSUE #12310: bundler 1.17 issue
The pr change to set a base64 version of 0.1.1 cause load failures with bundler 1.17

with the base bundler as 1.15 to less than 3, the base64 change causes ruby 3.0.3 with bundler 1.17.XXX to fail to load, with message
Can we change this to '~> 0.1' that should allow for both and cover ruby 3.3 should it not?

--------

## Expected behavior

it should load the gem

## Actual behavior

You have already activated base64 0.1.0, but your Gemfile requires base64 0.1.1. Since base64 is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports base64 as a default gem. (Gem::LoadError)

setup rubocop in your gemfile and use bundler 1.17 and try to run it.

## RuboCop version

1.15.7

koic:
Fixes #12310.

This PR drops `base64` gem from runtime dependency.

RuboCop only uses `Base64.encode64` from the `base64` gem. Therefore, there's no need to depend on the entire `base64` gem. The implementation of `Base64.encode64` is quite simple: https://github.com/ruby/base64/blob/v0.1.1/lib/base64.rb#L27-L40

This change is a better approach that has also been adopted by https://github.com/rack/rack/pull/2110.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_drop_base64_gem_from_runtime_dependency.md
+++ b/changelog/fix_drop_base64_gem_from_runtime_dependency.md
@@ -1,0 +1,1 @@
+* [#12310](https://github.com/rubocop/rubocop/issues/12310): Drop `base64` gem from runtime dependency. ([@koic][])
--- a/lib/rubocop/formatter/html_formatter.rb
+++ b/lib/rubocop/formatter/html_formatter.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'base64'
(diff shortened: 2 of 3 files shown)


bbatsov:
Great change!

PR #12314, opened 2023-10-25
add cop for standardizing json serialization to to_json

exterm:
Ruby codebases often use `JSON.generate` and `.to_json` interchangeably. The `Style/ToJson` cop helps unify to `to_json` to reduce confusion, and includes a safe (I hope) autocorrect.

```ruby
# bad
JSON.generate(something)

# good
something.to_json
```
-----------------

* [X] Added tests.


lib/rubocop/cop/style/to_json.rb
```diff
@@ -0,0 +1,55 @@
...
+      #
+      #   # good
+      #   something.to_json
+      class ToJson < Base
+        extend AutoCorrector
+
+        MSG = 'Use `.to_json` instead of `JSON.generate`.'
+
+        # @!method json_generate?(node)
+        def_node_matcher :json_generate?, '(send (const nil? :JSON) :generate ...)'
```
sambostock, line 22: This would allow us to also detect `::JSON.generate`:

```suggestion
        def_node_matcher :json_generate?, '(send (const {nil? chase} :JSON) :generate ...)'
```

lib/rubocop/cop/style/to_json.rb
```diff
@@ -0,0 +1,55 @@
...
+
+        def on_send(node)
+          return unless json_generate?(node)
+
+          add_offense(node) do |corrector|
+            _receiver, _method_name, *args = *node
```
sambostock, line 28: I could swear there's an `InternalAffairs` cop that's supposed to disallow deducting nodes (in favor of using the accessors), and I suspect you just found a bug in it (using `*` splat on the right hand side).

lib/rubocop/cop/style/to_json.rb
```diff
@@ -0,0 +1,55 @@
...
+            replacement = "#{data}.to_json"
+            replacement += "(#{opts})" if opts
+            corrector.replace(node, replacement)
+          end
+        end
+
+        private
+
+        def handle_args(args)
+          raise ArgumentError if args.size > 2
```
sambostock, line 39: I'm not sure we should raise in this scenario, as a host app could have something

```ruby
module MySerializer
  module JSON
    class << self
      def generate(foo, bar, biz)
```

and while false positives would be acceptable for that contrived case, I don't think raising an error if they attempt correction would be.

I think, instead, we'd want to detect non-standard arguments and not consider it an offense.

lib/rubocop/cop/style/to_json.rb
```diff
@@ -0,0 +1,55 @@
...
+            data, opts = handle_args(args)
+            replacement = "#{data}.to_json"
+            replacement += "(#{opts})" if opts
+            corrector.replace(node, replacement)
+          end
+        end
+
+        private
+
+        def handle_args(args)
```
sambostock, line 38: If you pass the `node` in instead, there are some useful helpers that could be leveraged (e.g. `first_argument`).

sambostock (commented):
👋 Hey @exterm!

lib/rubocop/cop/style/to_json.rb
exterm: Good point - we can just ignore cases where we have more than two arguments.

exterm:
I don't have a ton of experience writing cops, so thank you for your suggestions @sambostock , these are super helpful.

lib/rubocop/cop/style/to_json.rb
sambostock: Just noticed I have a typo there: it should be `cbase` not ~`chase`~.

lib/rubocop/cop/style/to_json.rb
sambostock: Found the [cop](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/internal_affairs/node_destructuring.rb) I was thinking of.

What's weird is that it should catch this, based on [this spec](https://github.com/rubocop/rubocop/blob/f981a1ad439a23a6d24c3bcda76764f14529f9cd/spec/rubocop/cop/internal_affairs/node_destructuring_spec.rb#L20-L25).

spec/rubocop/cop/style/to_json_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  it 'correctly autocorrects calls with serialization options' do
+    expect_offense(<<~RUBY)
+      JSON.generate(foo, space_before: '  ', space_after: ' ')
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `.to_json` instead of `JSON.generate`.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      foo.to_json(space_before: '  ', space_after: ' ')
+    RUBY
+  end
```
sambostock, line 37: I think we'll want to add specs for some or all of the following non-offending edge cases where we don't understand the arguments:

- `JSON.generate` (no args)
- `JSON.generate(one, two, three)`
- `JSON.generate(one, two, three: true)`

As well as `JSON.generate(one, two)` where `two` is not a `Hash` literal. e.g.
```ruby
options = ...
JSON.generate(blah, options)
```

lib/rubocop/cop/style/to_json.rb
sambostock: What's even weirder is that if I add _this exact file_ as an example in `node_destructuring_spec.rb`, and expect the offense, the spec passes.

```diff
diff --git a/spec/rubocop/cop/internal_affairs/node_destructuring_spec.rb b/spec/rubocop/cop/internal_affairs/node_destructuring_spec.rb
index 72236e565..3a6ae8f87 100644
--- a/spec/rubocop/cop/internal_affairs/node_destructuring_spec.rb
+++ b/spec/rubocop/cop/internal_affairs/node_destructuring_spec.rb
@@ -29,4 +29,67 @@ RSpec.describe RuboCop::Cop::InternalAffairs::NodeDestructuring, :config do
       lhs, rhs = array.children
     RUBY
   end
+
+  context 'example from #12314' do
+    it 'registers an offense' do
+      expect_offense(<<~'RUBY')
+        # frozen_string_literal: true
+
+        module RuboCop
+          ...
+                  add_offense(node) do |corrector|
+                    _receiver, _method_name, *args = *node
+                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use the methods provided with the node extensions instead of manually destructuring nodes.
+                    data, opts = handle_args(args)
+          ...
+        end
+      RUBY
+    end
+  end
 end
```

cc. @koic Any idea why CI didn't catch this?

lib/rubocop/cop/style/to_json.rb
sambostock: What's even weirder is that if I add _this exact file_ as an example in `node_destructuring_spec.rb`, and expect the offense, the spec passes.

```diff
diff --git a/spec/rubocop/cop/internal_affairs/node_destructuring_spec.rb b/spec/rubocop/cop/internal_affairs/node_destructuring_spec.rb
index 72236e565..3a6ae8f87 100644
--- a/spec/rubocop/cop/internal_affairs/node_destructuring_spec.rb
+++ b/spec/rubocop/cop/internal_affairs/node_destructuring_spec.rb
@@ -29,4 +29,67 @@ RSpec.describe RuboCop::Cop::InternalAffairs::NodeDestructuring, :config do
       lhs, rhs = array.children
     RUBY
   end
+
+  context 'example from #12314' do
+    it 'registers an offense' do
+      expect_offense(<<~'RUBY')
+        # frozen_string_literal: true
+
+        module RuboCop
+          ...
+                  add_offense(node) do |corrector|
+                    _receiver, _method_name, *args = *node
+                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use the methods provided with the node extensions instead of manually destructuring nodes.
+                    data, opts = handle_args(args)
+          ...
+        end
+      RUBY
+    end
+  end
 end
```

cc. @koic Any idea why CI didn't catch this?

spec/rubocop/cop/style/to_json_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Style::ToJson, :config do
+  let(:config) { RuboCop::Config.new('Layout/IndentationWidth' => { 'Width' => 2 }) }
```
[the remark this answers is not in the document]
exterm: Ah Sorry, probably copy-pasta. I'll check.

lib/rubocop/cop/style/to_json.rb
```diff
@@ -0,0 +1,55 @@
...
+          data = if args.first.hash_type? && !data_source.start_with?('{')
+                   "{ #{data_source} }"
+                 else
+                   data_source
+                 end
+
+          opts = (args[1].source if args.size == 2)
+
+          [data, opts]
+        end
```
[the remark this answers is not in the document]
exterm: Had to make a slight modification (the second argument isn't always a Hash) but this is great, thank you Sam!

exterm:
Ready for re-review, @sambostock

spec/rubocop/cop/style/to_json_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      JSON.generate(foo, opts)
+      ^^^^^^^^^^^^^^^^^^^^^^^^ Use `.to_json` instead of `JSON.generate`.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      foo.to_json(opts)
+    RUBY
+  end
+
+  it 'ignores calls to JSON.generate with unexpected arguments' do
```
sambostock, line 55: Unlikely edge case, but should we check passing a block?

```ruby
JSON.generate(foo) { nil }
```

config/default.yml
```diff
@@ -5379,6 +5379,11 @@ Style/TernaryParentheses:
     - require_parentheses_when_complex
   AllowSafeAssignment: true
 
+Style/ToJson:
+  Description: 'Standardize on `to_json` over `JSON.generate`.'
+  Enabled: false
```
sambostock, line 5384: Do we mean for this to be disabled by default, or do we think this is something the community could agree on (and therefore should be `pending`, to be enabled by default in the next major release, IIRC)?

sambostock:
Oh, and I think you'll want to squash this into a single commit matching the changelog entry.

spec/rubocop/cop/style/to_json_spec.rb
exterm: What behavior would we expect? I think I'm OK with either the cop ignoring this other transforming to `foo.to_json { nil }`.

exterm: On the other hand, if that code is valid, it probably means that someone has overridden `JSON.generate`, in which case we shouldn't transform it...?

Not sure it's worth investing time in this edge case, but I will if you think it is.

exterm:
Yeah, I kept commits separate for now to make it easier to follow along with the changes I'm making based on your review. I'll squash once we're done iterating.

config/default.yml
exterm: I'd rather not become blocked on the inevitable bikeshed 😆

exterm:
Squashed. This is ready for final review @sambostock @koic

config/default.yml
sambostock: Fair. It can always be changed from `false` to `pending` in a follow up PR.

spec/rubocop/cop/style/to_json_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    expect_correction(<<~RUBY)
+      foo.to_json
+    RUBY
+  end
+
+  it 'recognizes fully qualified reference to JSON' do
+    expect_offense(<<~RUBY)
+      ::JSON.generate(foo)
+      ^^^^^^^^^^^^^^^^^^^^ Use `.to_json` instead of `JSON.generate`.
+    RUBY
```
sambostock, line 19: I notice this doesn't have an `expect_correction`. Probably not a huge deal, since the case above basically covers it, but I thought I'd mention it, since I can't recall seeing specs that don't also expect the correction.

spec/rubocop/cop/style/to_json_spec.rb
sambostock: > What behavior would we expect?

Probably just ignoring it, but yeah, it's a crazy edge case, so it's probably fine to consider out of scope.

sambostock (commented):
✅ LGTM, but I'm not a maintainer, so I can't approve 😅

cc. @koic

koic:
The context provided in your explanation is clear, so it appears to be a style unification for `ActiveModel::Serializer`. Since `to_json` is not a method by Ruby's core but rather context-specific, it might be more appropriate to place it under the `Rails` department. Can you  reopen this PR as a cop under rubocop-rails? I will review this one there.
https://github.com/rubocop/rubocop-rails

sambostock:
@koic [`to_json` appears to be provided by the `json` gem](https://github.com/search?q=repo%3Aflori%2Fjson+%22def+to_json%22&type=code) that ships with Ruby; Rails just extends that support to classes such as `ActiveRecord::Base`.

koic:
Yes, I understand that, and I am referring to the context in which this style should be enforced.

exterm:
I had thought about that, but I decided that rubocop core is more appropriate.

I may be missing something, but it seems that in all Ruby code we have the choice between `JSON.generate` and `.to_json`. It makes sense to standardize on one of them.

The only reason why I brought up Rails is that in Rails applications we can't (at least not easily) standardize on `JSON.generate`, so the version that makes more sense there is to standardize on `.to_json`.

Why would this not make sense for non-Rails Ruby code?

koic:
Users might intentionally use `JSON.generate`, and it's not clear why `to_json` always should be preferred in this case. However, within the context of Rails, there seem to be distinct advantages to using `to_json`. This is why it appears more appropriate for rubocop-rails. In that sense, whether to_json being standardized is a universally accepted view in the Ruby community is not entirely clear. For instance, if the [Ruby Style Guide](https://github.com/rubocop/ruby-style-guide) explicitly states a rule preferring `to_json`, that could serve as a basis for its inclusion in the `Style` department of RuboCop.

exterm:
Okay! I think I may have misunderstood the semantics of including a cop in Rubocop core. I didn't intent to enable this one by default. I intended this cop to provide a choice, so that users / teams / projects that want to standardize on `to_json` can do so, and everyone else is unaffected.

However, it seems that doesn't fit with established semantics, so I'm happy to move it over.

sambostock:
I still think this cop could be valuable to projects not using Rails, as one may still want to standardize on one or the other, and **`to_json` is a non-Rails API**.

I notice that there's precedent for `rubocop-rails` changing the config of non-`Rails/...` cops.

@koic, what do you think of providing this cop in `rubocop`, but keeping it **disabled by default** and then enabling it by default in `rubocop-rails`?

```yml
# rubocop/config/default.yml
Style/ToJson:
  Enabled: false
```

```yml
# rubocop-rails/config/default.yml
Style/ToJson:
  Enabled: true
```

This would make the cop available to the **entire Ruby community**, **and** allow us to standardize on the `to_json` convention in Rails projects.

lib/rubocop/cop/style/to_json.rb
```diff
@@ -0,0 +1,50 @@
...
+              end
+              corrector.replace(node, replacement)
+            end
+          end
+        end
+
+        private
+
+        def braceless_hash?(node)
+          node.hash_type? && !node.source.start_with?('{')
```
exterm, line 45: We should be able to use `HashNode#braceless?` here.

exterm: Done.

exterm:
Any additional thoughts, @koic ?

PR #12319, opened 2023-10-28
Add a project spec

koic:
Follow up https://github.com/rubocop/rubocop/commit/d995cb4

This PR adds a project spec to prevent an unexpected changelog entry file path. For instance, if an unexpected file named `changelog/changelog/fix_foo.md` exists, it will fail as follows:

```console
$ bundle exec rspec spec/project_spec.rb:309
(snip)

  1) RuboCop Project Changelog future entries will not have a directory
     Failure/Error: expect(Dir[dir].all? { |path| File.file?(path) }).to be(true)

       expected true
            got false
     # ./spec/project_spec.rb:309:in `block (4 levels) in <top (required)>'

Finished in 0.02034 seconds (files took 1.71 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/project_spec.rb:308 # RuboCop Project Changelog future entries will not have a directory
```

-----------------

* [x] Added tests.

### DIFF
--- a/spec/project_spec.rb
+++ b/spec/project_spec.rb
@@ -305,6 +305,10 @@ RSpec.describe 'RuboCop Project', type: :feature do
 
       dir = File.expand_path('../changelog', __dir__)
 
+      it 'will not have a directory' do
+        expect(Dir["#{dir}/*"].none? { |path| File.directory?(path) }).to be(true)
+      end
+
       Dir["#{dir}/*.md"].each do |path|
         context "For #{path}" do
           let(:path) { path }


spec/project_spec.rb
```diff
@@ -305,7 +305,13 @@ def expected
 
       dir = File.expand_path('../changelog', __dir__)
 
-      Dir["#{dir}/*.md"].each do |path|
+      Dir["#{dir}/*"].each do |path|
+        it 'has a file' do
```
sambostock, line 309: I think this would be better as

```suggestion
        it 'is a file' do
```
Although, doesn't this only overwrite a single spec every iteration? Doesn't the path need to appear in the spec name to generate a unique spec?

PR #12366, opened 2023-11-06
Don't assume first and last comments are aligned

eugeneius:
Currently we use `each_cons(3)` to find comments that align with their neighbours, but this means the first and last comments aren't checked. We treat them as always aligned, but this leads to false negatives.

We can instead use `each_cons(2)` to look at each pair of comments and add both if they're aligned. This does mean that sometimes lines are added twice, but this is fine since `Set#<<` is idempotent.

-----------------

* [x] Added tests.


koic:
Thank you!

PR #12375, opened 2023-11-09
[Fix #12372] Fix a false negative for `Lint/Debugger`

### LINKED ISSUE #12372: Lint/Debugger not detected in begin/end blocks
The current implementation of `Lint/Debbuger` doesn't detect offenses inside `begin` / `end` blocks.

I was able to reproduce this issue in this small reproducible example:

```ruby
require 'byebug'

byebug

puts(
  begin
    byebug # This one is not detected
    "example"
  end
)
```

I checked this with the latest version of RuboCop.

## Expected behavior

The expected behaviour for this case is to detect both offenses. The one at top level and the one inside the `begin` `end` block.

## Actual behavior

The current behaviour is that RuboCop only detects the one at top level.

Create a `example.rb` file with the previous mentioned content, and run RuboCop like this:

```
bundle exec rubocop --only Lint/Debugger example.rb
```

Gemfile:
```
source "https://rubygems.org"

gem 'byebug', '~> 11.1', '>= 11.1.3'
gem 'rubocop', '1.57.2'
```

## RuboCop version

```
$ [bundle exec] rubocop -V
1.57.2 (using Parser 3.2.2.4, rubocop-ast 1.30.0, running on ruby 3.0.6) [x86_64-linux]
```

koic:
Fixes #12372.

This PR fixes a false negative for `Lint/Debugger` when used within method arguments a `begin`...`end` block.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_a_false_negative_for_lint_debugger.md
+++ b/changelog/fix_a_false_negative_for_lint_debugger.md
@@ -1,0 +1,1 @@
+* [#12372](https://github.com/rubocop/rubocop/issues/12372): Fix a false negative for `Lint/Debugger` when used within method arguments a `begin`...`end` block. ([@koic][])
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

fcsonline:
Thank you so much!!! 💯

PR #12384, opened 2023-11-10
[Fix #12377] Fix false positives for `Lint/Void`

### LINKED ISSUE #12377: Lint/Void and Lint/UselessAssignment removed my code with side effect
In spec we had the following code 

```rb
invoices = [FactoryBot.create(:invoice, customer_unit_id: @user.current_unit.id),
            FactoryBot.create(:invoice, supplier_unit_id: @user.current_unit.id),]
user = FactoryBot.create(:user)
```

During work to enable Lint/UselessAssignment. Rubocop found that invoices was an unused variable. I ran `be rubocop path.rb  --only Lint/UselessAssignment -A` and it replaced the code with the following code. which is fine. Lint/UselessAssignment is unsafe autocorrect so I ran it with -A.

```rb
[FactoryBot.create(:invoice, customer_unit_id: @user.current_unit.id),
 FactoryBot.create(:invoice, supplier_unit_id: @user.current_unit.id),]
user = FactoryBot.create(:user)
```
When I lint this code I get errors from Lint/Void. Running autofix safe version (-a) rubocop removed the array definition, so my two invoices are not created any more. 

```rb
user = FactoryBot.create(:user)
```
## Expected behavior

My suggestion is to don't modify the array if it has anything which could have side effect

## Actual behavior

Removes by factory bot calls

Explained above 

## RuboCop version

```
⋊> ~/d/w/d/f/repo on st-bugsnag-w40 ⨯ bundle exec rubocop -V                                                                                                                              12:57:13
1.57.2 (using Parser 3.2.2.4, rubocop-ast 1.30.0, running on ruby 3.2.1) [arm64-darwin22]
  - rubocop-capybara 2.19.0
  - rubocop-factory_bot 2.24.0
  - rubocop-performance 1.19.1
  - rubocop-rails 2.22.1
  - rubocop-rspec 2.25.0
```

koic:
Fixes #12377.

This PR fixes false positives for `Lint/Void` when a collection literal that includes non-literal elements in a method definition.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/void.rb
+++ b/lib/rubocop/cop/lint/void.rb
@@ -144,7 +144,7 @@ module RuboCop
         end
 
         def check_literal(node)
-          return if !node.literal? || node.xstr_type? || node.range_type?
+          return if !entirely_literal?(node) || node.xstr_type? || node.range_type?
 
           add_offense(node, message: format(LIT_MSG, lit: node.source)) do |corrector|
             autocorrect_void_expression(corrector, node)
@@ -216,6 +216,19 @@ module RuboCop
                         node.send_node
                       end
           corrector.replace(send_node.loc.selector, suggestion)
+        end
+
+        def entirely_literal?(node)
+          case node.type
+          when :array
+            node.each_value.all? { |value| entirely_literal?(value) }
+          when :hash
+            return false unless node.each_key.all? { |key| entirely_literal?(key) }
+
+            node.each_value.all? { |value| entirely_literal?(value) }
+          else
+            node.literal?
+          end
         end
       end
     end
--- a/changelog/fix_false_positives_for_lint_void.md
+++ b/changelog/fix_false_positives_for_lint_void.md
@@ -1,0 +1,1 @@
+* [#12377](https://github.com/rubocop/rubocop/issues/12377): Fix false positives for `Lint/Void` when a collection literal that includes non-literal elements in a method definition. ([@koic][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/lint/void.rb
```diff
@@ -217,6 +217,17 @@ def autocorrect_nonmutating_send(corrector, node, suggestion)
                       end
           corrector.replace(send_node.loc.selector, suggestion)
         end
+
+        def entirely_literal?(node)
+          case node.type
+          when :array
+            node.each_value.all?(&:literal?)
+          when :hash
+            node.each_value.all?(&:literal?) && node.each_key.all?(&:literal?)
```
vlad-pisanov, line 226: should this be recursive to handle nested literals like `[1, [2, 3]]`?

koic: Considering the possibility, it should have been addressed. I've updated this PR. Thank you!

lib/rubocop/cop/lint/void.rb
```diff
@@ -217,6 +217,19 @@ def autocorrect_nonmutating_send(corrector, node, suggestion)
                       end
           corrector.replace(send_node.loc.selector, suggestion)
         end
+
+        def entirely_literal?(node)
+          case node.type
+          when :array
+            node.each_value.all? { |value| entirely_literal?(value) }
+          when :hash
+            return false unless node.each_key.all?(&:literal?)
```
vlad-pisanov, line 226: technically, keys can be complex literals themselves, but I don't know if it's worth handling since it's not super common 😅  

```ruby
{
  [1, 2] => :foo
}
```

koic: You are right! I've updated this PR. Thanks!

PR #12385, opened 2023-11-12
[Fix #12242] Support `AllowModifiersOnAttrs` option for `Style/AccessModifierDeclarations`

### LINKED ISSUE #12242: Cop idea: `Style/AllowModifiersOnAttrs`

In Ruby >=3.0, we can write the modifier and `attr_*` helpers inline like:

``` ruby
class Foo
  public attr_accessor :a
  protected attr_reader :b
  protected attr_writer :c
  private attr :d, true
end
```

Today, this is an offend on `Style/AccessModifierDeclarations`. However, just for `attr*` helpers this style looks much cleaner than the default group style.

After all the default `group` style of `Style/AccessModifierDeclarations` works well for normal `def` methods, that I don't want to change that to `inline` style. Therefore I think the best approach is to add a new option called `Style/AllowModifiersOnAttrs` similar to `Style/AllowModifiersOnSymbols` to allow inlined modifier on `attr*` methods.

Use the group style everywhere - the problem is that I think most of the developers would prefer to put the attr declarations together, and then normal methods together, in that case we need to toggle public and private back and forth like below: 

``` ruby
class Foo
  attr_accessor :a

  private

  attr_accessor :b

  public # <-- extra "public" to toggle group back and forth

  def c; end

  private

  def d; end
  def e; end
end
```

Use the inline style everywhere, the problem would be simply we don't want to add the modifier to each normal method.

``` ruby
class Foo
  attr_accessor :a
  private attr_accessor :b

  def c; end
  private def d; end 
  private def e; end
  private def f; end
end
```

With a `Style/AllowModifiersOnAttrs` rule similar to `Style/AllowModifiersOnSymbols`, we would be able to write:

``` ruby
class Foo
  attr_accessor :a
  private attr_accessor :b

  def c; end

  private

  def d; end
  def e; end
  def f; end
end
```

krororo:
Fixes #12242

This PR supports `AllowModifiersOnAttrs` options for `Style/AccessModifierDeclarations`.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -3090,6 +3090,7 @@ Style/AccessModifierDeclarations:
     - inline
     - group
   AllowModifiersOnSymbols: true
+  AllowModifiersOnAttrs: true
   SafeAutoCorrect: false
 
 Style/AccessorGrouping:
--- a/lib/rubocop/cop/style/access_modifier_declarations.rb
+++ b/lib/rubocop/cop/style/access_modifier_declarations.rb
@@ -8,6 +8,17 @@ module RuboCop
       # EnforcedStyle config covers only method definitions.
       # Applications of visibility methods to symbols can be controlled
       # using AllowModifiersOnSymbols config.
+      # Also, the visibility of `attr*` methods can be controlled using
+      # AllowModifiersOnAttrs config.
+      #
+      # In Ruby 3.0, `attr*` methods now return an array of defined method names
+      # as symbols. So we can write the modifier and `attr*` in inline style.
+      # AllowModifiersOnAttrs config allows `attr*` methods to be written in
+      # inline style without modifying applications that have been maintained
+      # for a long time in group style. Furthermore, developers who are not very
+      # familiar with Ruby may know that the modifier applies to `def`, but they
+      # may not know that it also applies to `attr*` methods. It would be easier
+      # to understand if we could write `attr*` methods in inline style.
       #
       # @safety
       #   Autocorrection is not safe, because the visibility of dynamically
@@ -67,6 +78,34 @@ module RuboCop
       #     private :bar, :baz
       #
       #   end
+      #
+      # @example AllowModifiersOnAttrs: true (default)
+      #   # good
+      #   class Foo
+      #
+      #     public attr_reader :bar
+      #     protected attr_writer :baz
+      #     private attr_accessor :qux
+      #     private attr :quux
+      #
+      #     def public_method; end
+      #
+      #     private
+      #
+      #     def private_method; end
+      #
+      #   end
+      #
+      # @example AllowModifiersOnAttrs: false
+      #   # bad
+      #   class Foo
+      #
+      #     public attr_reader :bar
+      #     protected attr_writer :baz
+      #     private attr_accessor :qux
+      #     private attr :quux
+      #
+      #   end
       class AccessModifierDeclarations < Base
         extend AutoCorrector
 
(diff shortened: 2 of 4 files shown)


bbatsov:
Hmm, I'm a bit puzzled by the proposal as I imagine that almost everyone will prefer to use the `inline` style in this case. Is there something I'm missing here?

krororo:
I agree that some people prefer the `inline `style. However, as noted in the issue, there is a case where we want to keep the `group` style and write only `attr_*` in `inline` style.

For example, in applications that have been maintained for a long time, much of the code is already written in the `group` style. Changing the setting to `inline` style requires modifying a lot of code. Also, developers who are not very familiar with Ruby may know that the modifier applies to `def`, but they may not know that it also applies to `attr_*`. It would be easier to understand if we could write `attr_*` in `inline` style.

This option is similar to the `Style/AllowModifiersOnSymbols` option that is enabled by default. So we could write the code as follows, but it would be simpler to write the `attr_*` in `inline `style.

```rb
class Foo
  attr_reader :a
  private :a
  attr_accessor :b
  private :b, :b=
end
```

config/default.yml
```diff
@@ -3045,6 +3045,7 @@ Style/AccessModifierDeclarations:
     - inline
     - group
   AllowModifiersOnSymbols: true
+  AllowModifiersOnAttrs: false
```
krororo, line 3048: I wondered what the default value should be, but decided to set it to `false` for compatibility.

lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -92,10 +112,17 @@ class AccessModifierDeclarations < Base
           (send nil? {:private :protected :public :module_function} (sym _))
         PATTERN
 
+        # @!method access_modifier_with_attr?(node)
+        def_node_matcher :access_modifier_with_attr?, <<~PATTERN
+          (send nil? {:private :protected :public :module_function}
+            (send nil? {:attr_reader :attr_writer :attr_accessor} _))
```
ntkme, line 118: Also `:attr` (https://ruby-doc.org/3.2.2/Module.html#method-i-attr)

```suggestion
            (send nil? {:attr :attr_reader :attr_writer :attr_accessor} _))
```

config/default.yml
ntkme: Set it to `true` won't cause any existing code to suddenly have a violation, because this is not really a new rule but just an exception to an existing rule. So I think it can default to `true` to align with `AllowModifiersOnSymbols`.

lib/rubocop/cop/style/access_modifier_declarations.rb
krororo: Good catch! I forgot it.

config/default.yml
krororo: Indeed. Updated.

krororo:
Could you give us an update? Can I do something to merge this PR?

lib/rubocop/cop/style/access_modifier_declarations.rb
```diff
@@ -67,6 +67,28 @@ module Style
       #     private :bar, :baz
       #
       #   end
+      #
+      # @example AllowModifiersOnAttrs: true (default)
```
bbatsov, line 71: I think those examples should be extended to feature something else besides attributes, so people would understand the idea is to augment whatever the primary enforced style of the cop is.

krororo:
@bbatsov No problem. Thank you for the review.

> Makes sense to me. Just reflect this also it cop's documentation, so it's clearer for people why this option exists.

OK. I will update cop's documentation.

lib/rubocop/cop/style/access_modifier_declarations.rb
krororo: I would like to confirm that my understanding is correct, should I add method definitions like this?

``` ruby
# good
class Foo

  public attr_reader :bar
  protected attr_writer :baz
  private attr_accessor :qux
  private attr :quux

  def public_method; end

  private

  def private_method; end

end
```

bbatsov: Yep.

krororo:
@bbatsov Updated cop's documentation and rebased on top of the current master branch. Could you please review again?

PR #12387, opened 2023-11-14
[Fix #12146] Fix a false positive for `Lint/FloatComparison`

### LINKED ISSUE #12146: `Lint/FloatComparison`, `Style/NumericPredicate` and comparing by 0
I have the following rubocop config:

```yaml
Style/NumericPredicate:
  EnforcedStyle: comparison
```

`Lint/FloatComparison` discourages comparing by absolute values. I know why but zero seems like something that should be made an exception for.

```rb
if value.to_f != 0.0 # Lint/FloatComparison: Avoid (in)equality comparisons of floats as they are unreliable
  10 / value.to_f
end
```

If I switch to `EnforcedStyle: predicate` for `Style/NumericPredicate` I can write the following without offence:

```rb
unless value.to_f.zero? # All good!
  10 / value.to_f
end
```
`value` is user-provided and I want to prevent `NaN` from being returned, falling back to some other logic in that case. Alternatively, how would I write this without triggering an offense?

## Expected behavior

Rubocop should either allow `value.to_f != 0.0` or disallow `value.to_f.zero?`

## RuboCop version

```
$ [bundle exec] rubocop -V
1.55.1 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.2.2) [x86_64-linux-musl]
  - rubocop-erb 0.3.0
  - rubocop-rails 2.20.2
```

Earlopain:
Allow comparing against 0. See https://github.com/rubocop/rubocop/pull/12387#issuecomment-1818508562

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/float_comparison.rb
+++ b/lib/rubocop/cop/lint/float_comparison.rb
@@ -18,6 +18,10 @@ module RuboCop
       #   # good - using BigDecimal
       #   x.to_d == 0.1.to_d
       #
+      #  # good - comparing against zero
+      #   x == 0.0
+      #   x != 0.0
+      #
       #   # good
       #   (x - 0.1).abs < Float::EPSILON
       #
@@ -39,6 +43,8 @@ module RuboCop
 
         def on_send(node)
           lhs, _method, rhs = *node
+          return if literal_zero?(lhs) || literal_zero?(rhs)
+
           add_offense(node) if float?(lhs) || float?(rhs)
         end
 
@@ -57,6 +63,10 @@ module RuboCop
           else
             false
           end
+        end
+
+        def literal_zero?(node)
+          node&.numeric_type? && node.value.zero?
         end
 
         # rubocop:disable Metrics/PerceivedComplexity
--- a/changelog/fix_a_false_positive_for_float_comparision_with_zero_nonzero.md
+++ b/changelog/fix_a_false_positive_for_float_comparision_with_zero_nonzero.md
@@ -1,0 +1,1 @@
+* [#12146](https://github.com/rubocop/rubocop/issues/12146): Fix a false positive for `Lint/FloatComparison` when comparing against zero. ([@earlopain][])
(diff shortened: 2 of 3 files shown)


koic:
This cop is categorized under `Lint`, not `Style`. `0.zero?`, `0.0.zero?`, and `0 == 0.0` all return true. Therefore, it doesn't seem like bad code that should be marked as bad in `Lint`.

It appears that this issue should rather be addressed under `Style/NumericPredicate` to ensure consistency, I think.

Earlopain:
`0.0 == 0.0` returns true as well but is still flagged by `Lint/FloatComparison`. That is kind of the point that I made in my initial issue. 0 has special meaning and sometimes it is wanted to check for it. Basically https://github.com/rubocop/rubocop/pull/8432#issuecomment-682931742

My understanding is that `Style/NumericPredicate` enforces uniform usage of `x == 0` and `x.zero?` while `Lint/FloatComparison` makes you aware that you are dealing with floating point numbers and should be careful. This change seems appropriate, at least for how it currently works.

However, if you want to change `Lint/FloatComparison` to not flag comparision by zero, especially with  `x == 0.0` which is currently an offense, I can give that a shot as well.

koic:
> `0.0 == 0.0` returns true as well but is still flagged by `Lint/FloatComparison`.

This looks like a false positive. In any case, registering an offense for `zero?` by `Lint` cop seems strict for the users.

> However, if you want to change `Lint/FloatComparison` to not flag comparision by zero, especially with  `x.to_f == 0.0` which is currently an offense, I can give that a shot as well.

Yeah, that seems reasonable to me. I think zero is special.

Earlopain:
Changed as requested. I'm not entirely sure if I only need to check in `on_send` but I couldn't think of an example that doesn't work with this

lib/rubocop/cop/lint/float_comparison.rb
```diff
@@ -15,6 +15,10 @@ module Lint
       #   x == 0.1
       #   x != 0.1
       #
+      #  # good - comparing against zero
+      #   x == 0.0
+      #   x != 0.0
```
koic, line 20: Nits: Can you swap the positions of the `# good - using BigDecimal` example? It's to contrast it with the bad case, as `# good - using BigDecimal` is more illustrative when placed opposite.

koic:
Thanks!

PR #12388, opened 2023-11-14
Include expanded EnforcedStyle options when --no-auto-gen-enforced-style is given

kpost:
It seems to me that when the `--no-auto-gen-enforced-style` is given, not only `EnforcedStyle` should be rejected, but also all 'expanded' `EnforcedStyle` options as well, like for example `EnforcedStyleForEmptyBraces`. In order to achieve this I used a Regex when checking which keys should be rejected.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/formatter/disabled_config_formatter.rb
+++ b/lib/rubocop/formatter/disabled_config_formatter.rb
@@ -192,8 +192,8 @@ module RuboCop
         # 'Enabled' option will be put into file only if exclude
         # limit is exceeded.
         rejected_keys = ['Enabled']
-        rejected_keys << 'EnforcedStyle' unless auto_gen_enforced_style?
-        cfg.reject { |key| rejected_keys.include?(key) }
+        rejected_keys << /^EnforcedStyle\w*/ unless auto_gen_enforced_style?
+        cfg.reject { |key| include_or_match?(rejected_keys, key) }
       end
 
       def output_offending_files(output_buffer, cfg, cop_name)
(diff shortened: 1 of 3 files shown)


jonas054 (approved):
Looks good! 👍

lib/rubocop/formatter/disabled_config_formatter.rb
```diff
@@ -192,8 +192,8 @@ def filtered_config(cfg)
         # 'Enabled' option will be put into file only if exclude
         # limit is exceeded.
         rejected_keys = ['Enabled']
-        rejected_keys << 'EnforcedStyle' unless auto_gen_enforced_style?
-        cfg.reject { |key| rejected_keys.include?(key) }
+        rejected_keys << /^EnforcedStyle\w*/ unless auto_gen_enforced_style?
+        cfg.reject { |key| include_or_match?(rejected_keys, key) }
```
koic, line 196: Is there a logic by which the following would work, instead of creating the new `include_or_match` method?

```suggestion
        rejected_keys << 'EnforcedStyle' unless auto_gen_enforced_style?
        cfg.reject do |key|
          rejected_keys.any? { |rejected_key| key.start_with?(rejected_key) }
        end
```

kpost: Yes, this would work too! However then it only works for prefixed strings of course. I tried to anticipate for possible use cases where a key would not necessarily just start with the rejected key.

kpost: Yes, but any regexp could be used. :)

bbatsov:
I'm fine with the proposed change. Thanks!

PR #12397, opened 2023-11-17
[Fix #12296] Support `Redundant*Names` options for `Style/ArgumentsForwarding`

### LINKED ISSUE #12296: Style/ArgumentsForwarding` should allow or forbid specific argument names

As it is today, `Style/ArgumentsForwarding` is great for cutting down on the redundant noise of argument passing.  Using anonymous argument names clearly signals that the arguments are either ignored or forwarded.  It was the first autocorrection I ran after upgrading to ruby 3.2.

_However_, argument names can be used to document more than just "this is forwarded".  E.g: these names can document what they are represent, and where they are forwarded.  This is especially important for gem authors and public APIs.  As ruby language servers mature and achieve greater adoption, this documentation becomes more useful even on internal or private methods.

If the names are simple `*rest`, `*args`, `**kwargs`, `**opts`, and similar, then the names are (generally) not really serving any purpose but noise.  But if the names are `**scope_options` or `*authenticator_args` and so on, now they might be worth preserving.

I would like (at least) two new configurable attributes: `AllowedNames` and `ForbiddenNames`.  Both options should be an array of either regular expressions or strings.  The defaults could be the same as today: forbid all and allow none.

We _might_ add separate configuration options for each of the `rest`, `kwrest`, and `block` parameters, but I personally would prefer to keep it simpler with only these two options.

1. Just let rubocop autocorrect names that are unique and helpful... but only a little bit helpful.
2. Add `# rubocop:disable Style/ArgumentForwarding` every time I feel the names are important to the documentation.

## Additional context

When I upgraded to ruby 3.2, I used rubocop to autocorrect almost every single `Style/ArgumentsForwarding` offense in my codebase.  For at least 9 out of ten changes, I agreed completely.  For almost all of the rest, it wasn't really worth the effort to deal with them on a case-by-case basis... but if I'd had these configuration attributes, it would have been worth my time to configure.

Thanks!

koic:
Fixes #12296.

This PR supports `RedundantRestArgumentNames`, `RedundantKeywordRestArgumentNames`, and `RedundantBlockArgumentNames` options for `Style/ArgumentsForwarding`.

Meaningless names that are commonly used can be anonymized by default as part of the new `Redundant*Names` defaults. e.g., `*args`, `**options`, `&block`, and some argument names.

Names not on this list are likely to be meaningful and are allowed by default.

This approach is designed to only detect generally meaningless names without the need for users to edit the configuration for each meaningful name, with the expectation that the default settings will work well for most cases.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -3089,7 +3089,19 @@ Style/ArgumentsForwarding:
   Enabled: pending
   AllowOnlyRestArgument: true
   UseAnonymousForwarding: true
+  RedundantRestArgumentNames:
+    - args
+    - arguments
+  RedundantKeywordRestArgumentNames:
+    - kwargs
+    - options
+    - opts
+  RedundantBlockArgumentNames:
+    - blk
+    - block
+    - proc
   VersionAdded: '1.1'
+  VersionChanged: '<<next>>'
 
 Style/ArrayCoercion:
   Description: >-
--- a/changelog/change_support_redundant_options_for_style_arguments_forwarding.md
+++ b/changelog/change_support_redundant_options_for_style_arguments_forwarding.md
@@ -1,0 +1,1 @@
+* [#12296](https://github.com/rubocop/rubocop/issues/12296): Support `RedundantRestArgumentNames`, `RedundantKeywordRestArgumentNames`, and `RedundantBlockArgumentNames` options for `Style/ArgumentsForwarding`. ([@koic][])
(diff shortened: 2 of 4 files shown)


nevans:
@koic this is great, thanks. 😀

config/default.yml
```diff
@@ -3089,7 +3089,19 @@ Style/ArgumentsForwarding:
   Enabled: pending
   AllowOnlyRestArgument: true
   UseAnonymousForwarding: true
+  ForbiddenRestArgumentNames:
```
bbatsov, line 3092: I think "Redundant" is probably a better word than "Forbidden" for the purpose of this cop.

Issue #12410, opened 2023-11-24
Extend `Style/SelfAssignment` beyond variables


vlad-pisanov:
Is there a reason why `Style/SelfAssignment` only works with variables, as opposed to general expressions? I.e. could we lint for something like:

```ruby
# bad
hash['foo'] = hash['foo'] + 42
user.name = user.name + " Jr."

# good
hash['foo'] += 42
user.name += " Jr."
```

ydakuka:
And

```ruby
# bad

x = x % 2
x = x ^ 2
x = x << 2
x = x >> 2
```

```ruby
# good

x %= 2
x ^= 2
x <<= 2
x >>= 2
```

koic:
The main issue was resolved by #12415, and the derivative will be resolved by #12421.

PR #12415, opened 2023-11-25
Enhance `Lint/SelfAssignment` to check attribute assignment and key assignment

fatkodima:
This is an implementation of #12410 for `Lint/SelfAssignment`.

```ruby
# bad
foo.bar = foo.bar
foo["bar"] = foo["bar"]
```

### DIFF
--- a/lib/rubocop/cop/lint/self_assignment.rb
+++ b/lib/rubocop/cop/lint/self_assignment.rb
@@ -10,11 +10,18 @@ module RuboCop
       #   foo = foo
       #   foo, bar = foo, bar
       #   Foo = Foo
+      #   hash['foo'] = hash['foo']
+      #   obj.attr = obj.attr
       #
       #   # good
       #   foo = bar
       #   foo, bar = bar, foo
       #   Foo = Bar
+      #   hash['foo'] = hash['bar']
+      #   obj.attr = obj.attr2
+      #
+      #   # good (method calls possibly can return different results)
+      #   hash[foo] = hash[foo]
       #
       class SelfAssignment < Base
         MSG = 'Self-assignment detected.'
@@ -25,6 +32,15 @@ module RuboCop
           cvasgn: :cvar,
           gvasgn: :gvar
         }.freeze
+
+        def on_send(node)
+          if node.method?(:[]=)
+            handle_key_assignment(node) if node.arguments.size == 2
+          elsif node.assignment_method?
+            handle_attribute_assignment(node) if node.arguments.size == 1
+          end
+        end
+        alias on_csend on_send
 
         def on_lvasgn(node)
           lhs, rhs = *node
@@ -71,6 +87,27 @@ module RuboCop
         def rhs_matches_lhs?(rhs, lhs)
           rhs.type == ASSIGNMENT_TYPE_TO_RHS_TYPE[lhs.type] &&
             rhs.children.first == lhs.children.first
+        end
+
+        def handle_key_assignment(node)
+          value_node = node.arguments[1]
+
+          if value_node.send_type? && value_node.method?(:[]) &&
+             node.receiver == value_node.receiver &&
+             !node.first_argument.call_type? &&
+             node.first_argument == value_node.first_argument
+            add_offense(node)
+          end
+        end
+
+        def handle_attribute_assignment(node)
+          first_argument = node.first_argument
+
+          if first_argument.call_type? &&
+             node.receiver == first_argument.receiver &&
+             first_argument.method_name.to_s == node.method_name.to_s.delete_suffix('=')
+            add_offense(node)
+          end
         end
       end
     end
--- a/spec/rubocop/cop/lint/self_assignment_spec.rb
+++ b/spec/rubocop/cop/lint/self_assignment_spec.rb
@@ -129,4 +129,181 @@ RSpec.describe RuboCop::Cop::Lint::SelfAssignment, :config do
       foo &&= bar
     RUBY
   end
+
+  it 'registers an offense when using attribute self-assignment' do
+    expect_offense(<<~RUBY)
+      foo.bar = foo.bar
+      ^^^^^^^^^^^^^^^^^ Self-assignment detected.
+    RUBY
+  end
+
+  it 'registers an offense when using attribute self-assignment with a safe navigation call' do
+    expect_offense(<<~RUBY)
+      foo&.bar = foo&.bar
+      ^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
+    RUBY
+  end
+
+  it 'does not register an offense when using attribute assignment with different attributes' do
+    expect_no_offenses(<<~RUBY)
+      foo.bar = foo.baz
+    RUBY
+  end
+
+  it 'does not register an offense when using attribute assignment with different receivers' do
+    expect_no_offenses(<<~RUBY)
+      bar.foo = baz.foo
+    RUBY
+  end
+
+  it 'does not register an offense when using attribute assignment with extra expression' do
+    expect_no_offenses(<<~RUBY)
+      foo.bar = foo.bar + 1
+    RUBY
+  end
+
+  it 'registers an offense when using []= self-assignment with same string literals' do
+    expect_offense(<<~RUBY)
+      foo["bar"] = foo["bar"]
+      ^^^^^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
+    RUBY
+  end
+
+  it 'does not register an offense when using []= self-assignment with different string literals' do
+    expect_no_offenses(<<~RUBY)
+      foo["bar"] = foo["baz"]
+    RUBY
+  end
+
+  it 'registers an offense when using []= self-assignment with same integer literals' do
+    expect_offense(<<~RUBY)
+      foo[1] = foo[1]
+      ^^^^^^^^^^^^^^^ Self-assignment detected.
+    RUBY
+  end
+
+  it 'does not register an offense when using []= self-assignment with different integer literals' do
+    expect_no_offenses(<<~RUBY)
+      foo[1] = foo[2]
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/lint/self_assignment.rb
```diff
@@ -26,6 +26,15 @@ class SelfAssignment < Base
           gvasgn: :gvar
         }.freeze
 
+        def on_send(node)
+          case node.method_name
+          when :[]=
+            handle_key_assignment(node) if node.arguments.size == 2
+          when /=\z/
+            handle_attribute_assignment(node) if node.arguments.size == 1
+          end
```
koic, line 35: Can you tweak to the below?

```suggestion
          if node.method?(:[]=)
            handle_key_assignment(node) if node.arguments.size == 2
          elsif node.assignment_method?
            handle_attribute_assignment(node) if node.arguments.size == 1
          end
```

lib/rubocop/cop/lint/self_assignment.rb
```diff
@@ -72,6 +81,26 @@ def rhs_matches_lhs?(rhs, lhs)
           rhs.type == ASSIGNMENT_TYPE_TO_RHS_TYPE[lhs.type] &&
             rhs.children.first == lhs.children.first
         end
+
+        def handle_key_assignment(node)
+          value_node = node.arguments[1]
+
+          if value_node.send_type? && value_node.method?(:[]) &&
+             node.receiver == value_node.receiver &&
```
koic, line 89: Can you add a test case for the receiver condition?

lib/rubocop/cop/lint/self_assignment.rb
```diff
@@ -72,6 +81,26 @@ def rhs_matches_lhs?(rhs, lhs)
...
+             node.first_argument == value_node.first_argument
+            add_offense(node)
+          end
+        end
+
+        def handle_attribute_assignment(node)
+          first_argument = node.first_argument
+
+          if first_argument.send_type? &&
+             node.receiver == first_argument.receiver &&
```
koic, line 99: Ditto.

fatkodima:
Updated.

spec/rubocop/cop/lint/self_assignment_spec.rb
```diff
@@ -129,4 +129,36 @@
...
+      ^^^^^^^^^^^^^^^^^ Self-assignment detected.
+    RUBY
+  end
+
+  it 'does not register an offense when using attribute assignment' do
+    expect_no_offenses(<<~RUBY)
+      foo.bar = foo.baz
+      foo.bar
+      bar.foo = baz.foo
+      foo.bar = foo.bar + 1
```
koic, line 145: Can you split the `it` blocks for each testing content?

spec/rubocop/cop/lint/self_assignment_spec.rb
```diff
@@ -129,4 +129,36 @@
...
+      ^^^^^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
+    RUBY
+  end
+
+  it 'does not register an offense when using []= assignment' do
+    expect_no_offenses(<<~RUBY)
+      foo["bar"] = foo["baz"]
+      bar["foo"] = baz["foo"]
+      foo["bar"]
+      foo["bar"] = foo["bar"] + 1
```
koic, line 161: Ditto.

spec/rubocop/cop/lint/self_assignment_spec.rb
fatkodima: Done for both.

koic: `foo.bar = foo.bar + 1` is the same as L155. Is it left out intentionally?

spec/rubocop/cop/lint/self_assignment_spec.rb
```diff
@@ -129,4 +129,56 @@
...
+  it 'does not register an offense when using []= assignment with different keys' do
+    expect_no_offenses(<<~RUBY)
+      foo["bar"] = foo["baz"]
+    RUBY
+  end
+
+  it 'does not register an offense when using []= assignment with different receivers' do
+    expect_no_offenses(<<~RUBY)
+      bar["foo"] = baz["foo"]
+      foo["bar"] = foo["bar"] + 1
```
koic, line 175: `foo["bar"] = foo["bar"] + 1` is the same as L181. Is it left out intentionally?

spec/rubocop/cop/lint/self_assignment_spec.rb
fatkodima: Copy paste errors 🤦 Updated

lib/rubocop/cop/lint/self_assignment.rb
```diff
@@ -72,6 +80,26 @@ def rhs_matches_lhs?(rhs, lhs)
           rhs.type == ASSIGNMENT_TYPE_TO_RHS_TYPE[lhs.type] &&
             rhs.children.first == lhs.children.first
         end
+
+        def handle_key_assignment(node)
+          value_node = node.arguments[1]
+
+          if value_node.send_type? && value_node.method?(:[]) &&
+             node.receiver == value_node.receiver &&
+             node.first_argument == value_node.first_argument
```
vlad-pisanov, line 89: would this pick up cases where the hash index is non-deterministic, which (while extremely rare) would also render this cop unsafe? i.e.
```ruby
hash = {}
hash[rand] = hash[rand]
###
{ 0.03256159230386502=>nil } 
```

fatkodima: Yeah, this would pick up such cases. That is indeed extremely rare and I am not sure if it is worth marking this cop as unsafe because of this.

koic: Good catch. To ensure safety and certainty, it would be better to register an offense only when the key is some kind of literal.

fatkodima: Updated with this suggestion.

koic: It could be beneficial to include in the documentation the reason for not registering offenses when the key is a method call.

spec/rubocop/cop/lint/self_assignment_spec.rb
```diff
@@ -129,4 +129,60 @@
...
+      foo.bar = foo.bar + 1
+    RUBY
+  end
+
+  it 'registers an offense when using []= self-assignment with a literal' do
+    expect_offense(<<~RUBY)
+      foo["bar"] = foo["bar"]
+      ^^^^^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
+    RUBY
+  end
```
koic, line 163: Can you add test cases for other literal types and variables are used as keys?

fatkodima:
Updated.

spec/rubocop/cop/lint/self_assignment_spec.rb
```diff
@@ -129,4 +129,117 @@
...
+      foo["bar"] = foo["bar"]
+      ^^^^^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
+      foo[1] = foo[1]
+      ^^^^^^^^^^^^^^^ Self-assignment detected.
+      foo[1.2] = foo[1.2]
+      ^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
+      foo[Foo] = foo[Foo]
+      ^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
+      foo[:bar] = foo[:bar]
+      ^^^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
```
koic, line 176: Can you please separate the tests for each type of literal?

spec/rubocop/cop/lint/self_assignment_spec.rb
```diff
@@ -129,4 +129,117 @@
...
+    RUBY
+  end
+
+  it 'does not register an offense when using []= self-assignment with different literals' do
+    expect_no_offenses(<<~RUBY)
+      foo["bar"] = foo["baz"]
+      foo[1] = foo[2]
+      foo[1.2] = foo[2.2]
+      foo[Foo] = foo[Bar]
+      foo[:bar] = foo[:baz]
```
koic, line 186: Ditto.

spec/rubocop/cop/lint/self_assignment_spec.rb
```diff
@@ -129,4 +129,117 @@
...
+    expect_offense(<<~RUBY)
+      var = 1
+      foo[var] = foo[var]
+      ^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
+      foo[@var] = foo[@var]
+      ^^^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
+      foo[@@var] = foo[@@var]
+      ^^^^^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
+      foo[$var] = foo[$var]
+      ^^^^^^^^^^^^^^^^^^^^^ Self-assignment detected.
```
koic, line 200: Can you please separate the tests for each type of variable?

spec/rubocop/cop/lint/self_assignment_spec.rb
```diff
@@ -129,4 +129,117 @@
...
+  end
+
+  it 'does not register an offense when using []= self-assignment with different variables' do
+    expect_no_offenses(<<~RUBY)
+      var1 = 1
+      var2 = 2
+      foo[var1] = foo[var2]
+      foo[@var1] = foo[@var2]
+      foo[@@var1] = foo[@@var2]
+      foo[$var1] = foo[$var2]
```
koic, line 211: Ditto.

fatkodima:
Separated tests.

koic:
Thanks!

PR #12416, opened 2023-11-25
Add new `Style/ArrayFirstLast` cop

### LINKED ISSUE #12406: Cop idea: Prefer `first` and `last` over `[0]` and `[-1]`
https://rubystyle.guide/#first-and-last

When accessing the first or last element from an array, prefer `first` or `last` over `[0]` or `[-1]`.
```ruby
arr = [1, 2, 3]

# bad
arr[0]  # => 1
arr[-1] # => 3

# good
arr.first # => 1
arr.last  # => 3
```

## Rubocop

```console
ydakuka@yauhenid:~/Work/project$ bin/rails_docker rubocop -V
1.57.2 (using Parser 3.2.2.4, rubocop-ast 1.29.0, running on ruby 2.7.8) [x86_64-linux]
  - rubocop-capybara 2.19.0
  - rubocop-factory_bot 2.24.0
  - rubocop-performance 1.19.1
  - rubocop-rails 2.22.0
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.25.0
  - rubocop-thread_safety 0.5.1
```

fatkodima:
Fixes #12406.

I already implemented this cop (so that's why still decided to open a PR) when realised that `.first`/`.last` is only suited for arrays. While `[0]` and `[-1]` can be also used with strings or hashes. So this can probably generate many false positives. I made it disabled by default.


bbatsov:
If the methods are specific only to arrays, I guess we can name the cop "ArrayFirstLast" or something more generic like "ArrayElementAccess". Might be good to make this configurable for whoever prefers indices as well, although I don't feel strongly about this. 

 > I made it disabled by default.
 
 Makes sense.

koic:
Documenting the reasons for being disabled by default will help users understand.

fatkodima:
Added a line regarding why it is disabled.

lib/rubocop/cop/style/array_first_last.rb
```diff
@@ -0,0 +1,64 @@
...
+module RuboCop
+  module Cop
+    module Style
+      # Identifies usages of `arr[0]` and `arr[-1]` and suggests to change
+      # them to use `arr.first` and `arr.instead`.
+      #
+      # The cop is disabled by default due to safety concerns.
+      #
+      # @safety
+      #   Autocorrection is unsafe because `[0]` or `[-1]` can be called on a Hash,
```
bbatsov, line 12: Not just the autocorrection is unsafe - the cop is fundamentally unsafe. :-)

PR #12420, opened 2023-11-29
Add new `Lint/LiteralAssignmentInCondition` cop

koic:
## Summary

This PR adds a new `Lint/LiteralAssignmentInCondition` cop that checks for literal assignments in the conditions of `if`, `while`, and `until`. It emulates the following Ruby warning:

```console
$ ruby -we 'if x = true; end'
-e:1: warning: found `= literal' in conditional, should be ==
```

As a lint cop, it cannot be determined if `==` is appropriate as intended, therefore this cop does not provide autocorrection.

```ruby
# bad
if x = 42
  do_something
end

# good
if x == 42
  do_something
end

# good
if x = y
  do_something
end
```

Literal assignment used within a condition always triggers a Ruby warning, regardless of parentheses. This PR prevents the Ruby warning from being overlooked due to the `ruby -w` option.

## Additional Information

Here, the differences between similar cops, `Lint/LiteralAssignmentInCondition` cop and `Lint/LiteralAsCondition` cop are discussed.

### `Lint/LiteralAssignmentInCondition` cop

A similar `Lint/LiteralAssignmentInCondition` cop already exists, but as mentioned above, literal assignments are Ruby's warned about regardless of the presence of parentheses:

```ruby
% ruby -we 'if (x = 42); end'
-e:1: warning: found `= literal' in conditional, should be ==
```

Therefore, a new cop has been created separate from `Lint/LiteralAssignmentInCondition`.

Furthermore, as documented for `Lint/LiteralAssignmentInCondition` cop, literal assignments are never considered `good` case, so `= true` has been replaced with `= value`.

## `Lint/LiteralAsCondition` cop

Additionally, the existing `Lint/LiteralAsCondition` cop addresses a slightly different Ruby warning:

```ruby
$ ruby -we 'if 42; end'
-e:1: warning: literal in condition
```

This is another reason for choosing to implement a new cop.

-----------------

* [x] Added tests.


bbatsov:
Looks good to me. Thanks!

PR #12427, opened 2023-12-01
[Fix #12309] Add new `Style/RedundantSuperArguments` cop

### LINKED ISSUE #12309: Cop idea: `RedundantSuperArguments`

Consider the following code:

```rb
def input(attribute_name, options = {}, &)
  value = @options[attribute_name]
  options[:value] = value
  super(attribute_name, options, &)
end
```

Ruby delegates all arguments to super if called without anything. Rubocop should detect this case and autocorrect to a plain super call.

Autocorrect to the following if the arguments to super matches the method signature:

```rb
def input(attribute_name, options = {}, &)
  value = @options[attribute_name]
  options[:value] = value
  super
end
```

## Additional context

Calling super with parents like `super()` has special meaning and will not pass any arguments to the inherited method. There should be no autocorrect in that case.

Earlopain:
#12309 Implements the following:

```rb
def method(a, b, c:, d: &e)
  super(a, b, c:, d:, &e)
  # => super
end
```

There are 57 offenses of this kind in Rails master.

Looks like I picked quite the doozy for my first cop. I managed but I'm open to feedback on how to write better code here.
I tried to think of all the constellations you can use `super` in and handling those appropriately in the cop but I'm not certain if I got it all. Let me know if there are other edge-cases that need to be considered, like calling `super` inside `define_singleton_method`.

The cop could be more intelligent with keyword arguments when they appear in different orders like in the following example but I'll leave that be for now since the cop seems quite complex already. Something to improve upon in the future.
```rb
def test(a:, b:)
  super(b:, a:)
end
```

* [x] Added tests.


koic:
There might be users who intentionally write explicit arguments. So, I wonder if the cop name "Redundant" is suitable. 
And, it would be beneficial to first suggest a rationale in the style guide for using zero arity super (zsuper) when `super` can have its arguments omitted:
https://github.com/rubocop/ruby-style-guide

bbatsov:
@Earlopain ping :-)

Earlopain:
Oh, I forgot I even did this.

I renamed the cop to the suggestion from koic and disabled it by default. I don't fell strongly about this anymore to discuss about this in the styleguide, so if disabled by default is not good then feel free to close this.

bbatsov:
I think the cop is useful, and I even thought we already had something like it. 

As far as the names go - I'm not sure I like `ZeroAritySuper`, as I find it a bit vague. Perhaps it could just be named `SuperArguments`, which would also leave the door to enforce removing or providing any redundant arguments. Anyways, I don't feel strongly about the name, but I'll leave this open for a bit more in case someone proposes a name that is clearer. 

@rubocop/rubocop-core Any opinions on this would be welcome!

bbatsov:
> I renamed the cop to the suggestion from koic and disabled it by default. I don't fell strongly about this anymore to discuss about this in the styleguide, so if disabled by default is not good then feel free to close this.

I don't think there's much to discuss about such a guideline as I thought that most people omit the redundant args, so you can safely file a PR in the style guide. I'm also not sure this should be disabled by default - probably we can leave it to the users to decide if they'd fine it useful or not.

koic:
`Style/SuperArguments` seems like a better name because it is appropriately abstract. I also think this cop can start with the default `Enabled: pending`.

Earlopain:
Alright, I've openend https://github.com/rubocop/ruby-style-guide/pull/941,  set it back to pending and renamed to `SuperArguments`. I also agree that name is better.

I stumbled upon an interesting interaction I wasn't really aware of with blocks, see the styleguide PR for what I'm talking about. I can think of any changes this cop would need because of this.

lib/rubocop/cop/style/super_arguments.rb
```diff
@@ -0,0 +1,144 @@
...
+          return unless def_node
+          return unless arguments_identical?(def_node.arguments.argument_list, super_node.arguments)
+
+          add_offense(super_node) { |corrector| corrector.replace(super_node, 'super') }
+        end
+
+        private
+
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/PerceivedComplexity
```
koic, line 52: Can you tweak to the following?

```suggestion
        # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
```

lib/rubocop/cop/style/super_arguments.rb
```diff
@@ -0,0 +1,144 @@
...
+            next if keyword_rest_arg_same?(def_arg, super_arg)
+            next if block_arg_same?(def_arg, super_arg)
+            next if forward_arg_same?(def_arg, super_arg)
+
+            return false
+          end
+          true
+        end
+        # rubocop:enable Metrics/CyclomaticComplexity
+        # rubocop:enable Metrics/PerceivedComplexity
```
koic, line 70:
```suggestion
        # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
```

lib/rubocop/cop/style/super_arguments.rb
```diff
@@ -0,0 +1,144 @@
...
+        def positional_arg_same?(def_arg, super_arg)
+          return false unless def_arg.arg_type? || def_arg.optarg_type?
+          return false unless super_arg.lvar_type?
+
+          def_arg.name == super_arg.children.first
+        end
+
+        def positional_rest_arg_same(def_arg, super_arg)
+          return false unless def_arg.restarg_type?
+          # anon forwarding
```
koic, line 81: Can you replace `anon` with `anonymous` in this PR?

```suggestion
          # anonymous forwarding.
```

lib/rubocop/cop/style/super_arguments.rb
```diff
@@ -0,0 +1,144 @@
...
+
+          def_arg.name == lvar_node.children.first
+        end
+
+        def keyword_arg_same?(def_arg, super_arg)
+          return false unless def_arg.kwarg_type? || def_arg.kwoptarg_type?
+          return false unless (pair_node = super_arg).pair_type?
+          return false unless (sym_node = pair_node.key).sym_type?
+          return false unless (lvar_node = pair_node.value).lvar_type?
+          return false unless sym_node.value == lvar_node.children.first
```
koic, line 94:
```suggestion
          return false unless sym_node.source == lvar_node.source
```

lib/rubocop/cop/style/super_arguments.rb
```diff
@@ -0,0 +1,144 @@
...
+          # anon forwarding
+          return true if (block_pass_child = super_arg.children.first).nil? && def_arg.name.nil?
+
+          def_arg.name == block_pass_child.children.first
+        end
+
+        def forward_arg_same?(def_arg, super_arg)
+          return false unless def_arg.forward_arg_type? && super_arg.forwarded_args_type?
+
+          true
```
koic, line 120:
```suggestion
          def_arg.forward_arg_type? && super_arg.forwarded_args_type?
```

lib/rubocop/cop/style/super_arguments.rb
```diff
@@ -0,0 +1,144 @@
...
+        end
+
+        def define_method?(node)
+          return false unless node.block_type?
+
+          child = node.child_nodes.first
+          return false unless child.send_type?
+
+          child.method?(:define_method) || child.method?(:define_singleton_method)
```
koic, line 129:
```suggestion
          return false unless node.block_type?

          node.method?(:define_method) || node.method?(:define_singleton_method)
```

lib/rubocop/cop/style/super_arguments.rb
```diff
@@ -0,0 +1,144 @@
...
+        end
+
+        def preprocess_super_args(super_args)
+          super_args.map do |node|
+            if node.hash_type? && !node.braces?
+              node.children
+            else
+              node
+            end
+          end.flatten
```
koic, line 139:
```suggestion
          super_args.flat_map do |node|
            if node.hash_type? && !node.braces?
              node.children
            else
              node
            end
          end
```

Earlopain:
Thanks for your review, I appreciate it. I should have addresses all your points

spec/rubocop/cop/style/super_arguments_spec.rb
```diff
@@ -0,0 +1,220 @@
...
+  it_behaves_like 'no offense', 'block argument with different name', '&blk', '&other_blk'
+  it_behaves_like 'no offense', 'keyword arguments and hash', 'a:', '{ a: a }'
+  it_behaves_like 'no offense', 'keyword arguments with send node', 'a:, b:', 'a: a, b: c'
+  it_behaves_like 'no offense', 'tripple dot forwarding with extra param', '...', 'a, ...'
+  it_behaves_like 'no offense', 'tripple dot forwarding with different param', 'a, ...', 'b, ...'
+  it_behaves_like 'no offense', 'keyword forwarding with extra keyword', 'a, **kwargs', 'a: a, **kwargs'
+
+  context 'Ruby >= 3.1', :ruby31 do
+    it_behaves_like 'offense', 'hash value omission', 'a:'
+    it_behaves_like 'offense', 'anon block forwarding', '&'
```
koic, line 84: Can you also use "anonymous" for the terminology in the test code, just like in the product code?
https://github.com/rubocop/rubocop/pull/12427#discussion_r1610340732

```suggestion
    it_behaves_like 'offense', 'anonymous block forwarding', '&'
```

spec/rubocop/cop/style/super_arguments_spec.rb
```diff
@@ -0,0 +1,220 @@
...
+  end
+
+  context 'Ruby >= 3.2', :ruby32 do
+    it_behaves_like 'offense', 'anon positional forwarding', '*'
+    it_behaves_like 'offense', 'anon keyword forwarding', '**'
+
+    it_behaves_like 'no offense', 'mixed anon forwarding', '*, **', '*'
+    it_behaves_like 'no offense', 'mixed anon forwarding', '*, **', '**'
```
koic, line 92: Some of the above are the same.

bbatsov:
Thanks!

PR #12518, opened 2023-12-07
Add new `Lint/ItWithoutArgumentsInBlock` cop

koic:
Follow up https://bugs.ruby-lang.org/issues/18980 and https://github.com/ruby/ruby/pull/9152.

Emulates the following Ruby warnings in Ruby 3.3.

```ruby
$ ruby -e '0.times { it }'
-e:1: warning: `it` calls without arguments will refer to the first block param in Ruby 3.4;
use it() or self.it
```

`it` calls without arguments will refer to the first block param in Ruby 3.4.
So use `it()` or `self.it` to ensure compatibility.

```ruby
# bad
do_something { it }

# good
do_something { it() }
do_something { self.it }
```

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

Issue #12516, opened 2023-12-05
Regression: Infinite Loop between `Style/RedundantParentheses` and `Lint/AmbiguousRange`


jtannas:
Rubocop v1.58.0 introduces a regression that causes an infinite loop due to conflicting recommendations from  `Style/RedundantParentheses` and `Lint/AmbiguousRange`.

--------

## Expected behavior

I would expect one rule to take priority over the other so that no infinite loop occurs.

## Actual behavior
![image](https://github.com/rubocop/rubocop/assets/22785056/2a1e8204-b297-4958-ad5e-fd454019d2e7)
![image](https://github.com/rubocop/rubocop/assets/22785056/a64d1bbe-e31d-4a8a-9aa6-ca55427284be)

<details>
<summary>Stack Dump</summary>

```
186 files inspected, 4 offenses detected, 4 offenses corrected
Infinite loop detected in /home/jtannas/my-project/backend/app/models/foo.rb and caused by Lint/AmbiguousRange, Style/RedundantParentheses -> Lint/AmbiguousRange, Style/RedundantParentheses
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-1.58.0/lib/rubocop/runner.rb:332:in `check_for_infinite_loop'
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-1.58.0/lib/rubocop/runner.rb:315:in `block in iterate_until_no_changes'
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-1.58.0/lib/rubocop/runner.rb:314:in `loop'
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-1.58.0/lib/rubocop/runner.rb:314:in `iterate_until_no_changes'
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-1.58.0/lib/rubocop/runner.rb:283:in `do_inspection_loop'
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-1.58.0/lib/rubocop/runner.rb:164:in `block in file_offenses'
[...]

PR #12531, opened 2023-12-11
Fix pr reference to the new `Lint/ItWithoutArgumentsInBlock` cop in CHANGELOG

brkn:
Reference to the new cop pr was incorrect #12516 is unrelated to the correct #12518


koic:
Thanks!

PR #12535, opened 2023-12-12
Allow autocorrect with display-only-fail-level-offenses

naveg:
The `--display-only-fail-level-offenses` option uses
`considered_failure?` to determine what to print. The same method is
used to determine the exit code of `rubocop`, so the two options should
be entirely safe to use together.

This change is useful when using rubocop as part of a pre-commit hook.
When the hook fails becuase rubocop exited with a non-zero status, we
can use `--display-only-fail-level-offenses` to avoid having the output
cluttered up with non-blocking offenses.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/options_spec.rb
+++ b/spec/rubocop/options_spec.rb
@@ -356,10 +356,10 @@ RSpec.describe RuboCop::Options, :isolated_environment do
     end
 
     describe '--display-only-fail-level-offenses' do
-      it 'fails if given with an autocorrect argument' do
-        %w[--fix-layout -x --autocorrect -a --autocorrect-all -A].each do |o|
-          expect { options.parse ['--display-only-correctable', o] }
-            .to raise_error(RuboCop::OptionArgumentError)
+      %w[--fix-layout -x --autocorrect -a --autocorrect-all -A].each do |o|
+        it 'fails if given with an autocorrect argument' do
+          expect { options.parse ['--display-only-fail-level-offenses', o] }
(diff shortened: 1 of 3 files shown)


spec/rubocop/options_spec.rb
```diff
@@ -355,15 +355,6 @@ def abs(path)
       end
     end
 
-    describe '--display-only-fail-level-offenses' do
-      it 'fails if given with an autocorrect argument' do
-        %w[--fix-layout -x --autocorrect -a --autocorrect-all -A].each do |o|
-          expect { options.parse ['--display-only-correctable', o] }
-            .to raise_error(RuboCop::OptionArgumentError)
```
koic, line 362: It seems that instead of being removed, it should be updated to reflect an expected behavior, rather than resulting in the `RuboCop::OptionArgumentError` error.

naveg: I'm not sure I follow. This tests the validation that I removed in order to allow the two arguments to be used together. I could change it to `not_to` expect the error, but did you have something else in mind?

naveg: Done. The test was actually not doing the right thing before, because it was re-using the same `Options` instance and generating errors for a different reason (using -a and -A together)

naveg:
@koic would appreciate another look when you have a chance, thanks! 🙏

naveg:
@koic @bbatsov just checking in on this one. Let me know if I should make any other adjustments

bbatsov:
The change looks good to me. Thanks!

PR #12536, opened 2023-12-13
Fix BufPreWrite pattern in nvim-lspconfig rubocop formatting example

pjg:
The pattern for `Buf*` commands in neovim looks to be a file names pattern, not file types. When set to `ruby`, automatic formatting doesn't work. It starts working only when I set it to `*.rb`.

Reference:
https://neovim.io/doc/user/api.html#:~:text=%22BufWinEnter%22%7D%2C%20%7B-,pattern%20%3D%20%7B%22*.c%22%2C%20%22*.h%22%7D,-%2C%0A%20%20callback%20%3D%20function


bbatsov:
Thanks!

PR #12541, opened 2023-12-14
 Fix false negative in `Style/ArgumentsForwarding` when a block is forwarded but other args aren't

dvandersluis:
The following code gets registered as a false negative by `Style/ArgumentsForwarding`

```ruby
def foo(*args, &block)
  bar(*args, &block)
  baz(&block)
end
```

Which results in the following invalid autocorrection (`&block` is now longer defined):
```ruby
def foo(...)
  bar(...)
  baz(&block)
end
```

This change fixes that case by considering block arguments when determining if forwarding is possible. 

This had two side effects:
1. The test for different argument names failed due to the block parameter now being considered when determining if there is an offense. To fix this, we now check that the rest and kwrest arguments have the same name.
2. This revealed a few offenses in rubocop code where only a block argument should have been replaced with forwarding.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_fix_false_negative_in_style_arguments_forwarding.md
+++ b/changelog/fix_fix_false_negative_in_style_arguments_forwarding.md
@@ -1,0 +1,1 @@
+* [#12541](https://github.com/rubocop/rubocop/pull/12541): Fix false negative in `Style/ArgumentsForwarding` when a block is forwarded but other args aren't. ([@dvandersluis][])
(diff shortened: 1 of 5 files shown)


bbatsov:
Looks good! Thanks for tackling this!

mvz:
The suggested correction seems to change the method's arity, which could be the wrong thing to do.

mvz:
See #12618.

PR #12546, opened 2023-12-15
Teach `Lint/AssignmentInCondition` about `case`

sambostock:
This teaches `Lint/AssignmentInCondition` about `case`, to match the behavior with other conditionals.

```diff
 case test = 10
+          ^ Use `==` if you meant to do a comparison or wrap the expression in parentheses to indicate you meant to assign in a condition.
 when nil
 end
```

```ruby
case (test = 10)
when nil
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_teach_lint_assignment_in_condition_about_case.md
+++ b/changelog/change_teach_lint_assignment_in_condition_about_case.md
@@ -1,0 +1,1 @@
+* [#12546](https://github.com/rubocop/rubocop/pull/12546): Teach `Lint/AssignmentInCondition` about `case`. ([@sambostock][])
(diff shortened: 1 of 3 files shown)


koic:
This cop emulates the following Ruby warning:

```console
$ ruby -ce 'if (test = 10); end'
-e:1: warning: found `= literal' in conditional, should be ==
Syntax OK
```

It does not display the warning for `case`:

```
$ cat example.rb
case (test = 10)
when nil
end

$ ruby -wc example.rb
Syntax OK
```

koic:
Ah, I made a mistake in my previous comment about `Lint/LiteralAssignmentInCondition` 😓 

However, `case`...`when` differs from `if` and `while`, as it does not independently form a condition. This might call for an initial discussion in the style guide.
https://rubystyle.guide/#safe-assignment-in-condition

PR #12559, opened 2023-12-19
Document `Style/SymbolProc` protected method incompatibility

sambostock:
This documents the false positive for `Style/SymbolProc` when used with a `protected` method. `public` and `private` method behaviour matches, but `protected` methods which work in the normal block style break when using `Symbol#to_proc`.

The motivation for this is that it's not something that comes up very often (`protected` is not as common as `public` or `private`, and even less so used in iteration/blocks), so reviewers will often need to ask why a `rubocop:disable Style/SymbolProc` comment is being added. Documenting this makes it easier to discover the reason.

-----------------

* ~Added tests.~

### DIFF
--- a/lib/rubocop/cop/style/symbol_proc.rb
+++ b/lib/rubocop/cop/style/symbol_proc.rb
@@ -37,6 +37,42 @@ module RuboCop
       #   # ArgumentError: wrong number of arguments (given 1, expected 0)
       #   ----
       #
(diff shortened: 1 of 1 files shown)


sambostock:
Since this doesn't add or change any _behaviour_, I didn't add a changelog file. I can add one if necessary.

bbatsov:
Today I learned! :-) Thanks for documenting this.

PR #12562, opened 2023-12-20
Improve invertible unless message

sambostock:
The existing message is ambiguous about how to invert the condition, leading many to assume the cop is suggesting replacing code like

```ruby
unless number.odd?
^^^^^^^^^^^^^^^^^^ Favor `if` with inverted condition over `unless`.
```

with

```ruby
if !number.odd?
```

which would conflict with `Style/NegatedIf`.

This updates the message to clearly describe the change being recommended:

```ruby
unless number.odd?
^^^^^^^^^^^^^^^^^^ Prefer `if number.even?` over `unless number.odd?`.
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/invertible_unless_condition.rb
+++ b/lib/rubocop/cop/style/invertible_unless_condition.rb
@@ -51,7 +51,7 @@ module RuboCop
       class InvertibleUnlessCondition < Base
         extend AutoCorrector
 
-        MSG = 'Favor `if` with inverted condition over `unless`.'
+        MSG = 'Prefer `%<prefer>s` over `%<current>s`.'
 
         def on_if(node)
           return unless node.unless?
@@ -59,7 +59,10 @@ module RuboCop
           condition = node.condition
           return unless invertible?(condition)
 
-          add_offense(node) do |corrector|
+          message = format(MSG, prefer: "#{node.inverse_keyword} #{preferred_condition(condition)}",
+                                current: "#{node.keyword} #{condition.source}")
+
+          add_offense(node, message: message) do |corrector|
             corrector.replace(node.loc.keyword, node.inverse_keyword)
             autocorrect(corrector, condition)
           end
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/style/invertible_unless_condition.rb
```diff
@@ -88,6 +91,47 @@ def inheritance_check?(node)
...
+            argument_list = node.arguments.map(&:source).join(', ')
+            if node.parenthesized?
+              "#{receiver_source}.#{inverse_method_name}(#{argument_list})"
+            else
+              "#{receiver_source}.#{inverse_method_name} #{argument_list}"
+            end
+          else
+            "#{receiver_source}.#{inverse_method_name}"
+          end
+        end
```
sambostock, line 126: I kept this version of this method as a separate commit, in case we prefer the one which does not introduce an offense:

```ruby
def preferred_send_condition(node)
  receiver_source = node.receiver.source
  return receiver_source if node.method?(:!)

  inverse_method_name = inverse_methods[node.method_name]
  dot_or_space = node.dot? ? '.' : ' '
  args = node.arguments.map(&:source).join(', ')
  args = node.parenthesized? ? "(#{args})" : " #{args}" if node.arguments?

  "#{receiver_source}#{dot_or_space}#{inverse_method_name}#{args}"
end
```

While this approach is shorter, I don't know if it's actually clearer, as it condenses the logic into fewer more complex lines.

Another approach would be to extract more methods, but I feel like that would also hurt the overall readability.

The longer approach makes it simpler to see what the outcome is:

Scenario|Preferred Implementation
-|-
Unary `!`|`receiver_source`
Operator|`"#{receiver_source} #{inverse_method_name} #{node.first_argument.source}"`
Parentheses|`"#{receiver_source}.#{inverse_method_name}(#{argument_list})"`
No Parentheses|`"#{receiver_source}.#{inverse_method_name} #{argument_list}"`
No arguments|`"#{receiver_source}.#{inverse_method_name}"`

lib/rubocop/cop/style/invertible_unless_condition.rb
```diff
@@ -88,6 +91,47 @@ def inheritance_check?(node)
             (argument.const_type? && argument.short_name.to_s.upcase != argument.short_name.to_s)
         end
 
+        def preferred_condition(node)
+          case node.type
+          when :begin
+            if node.parenthesized_call?
+              "(#{preferred_condition(node.children.first)})"
+            else # TODO: What about other kinds of begin nodes?
```
sambostock, line 99: I forgot about this comment, but maybe it still applies.

Do we care about things like this?

```ruby
if begin; dangerous_condition; rescue => SomeError; fallback_condition; end;
  # ...
end
```

sambostock: Okay, so I tried

```ruby
foo unless begin x != y end
foo unless (begin x != y end)
```

but neither detected an offense, so I'll just assume the parenthesized case is the only one we care about, and simplify the implementation.

lib/rubocop/cop/style/invertible_unless_condition.rb
sambostock: I managed to put together a similar implementation which doesn't trigger an offense.

Gemfile
```diff
@@ -31,8 +31,8 @@ group :test do
   gem 'webmock', require: false
 end
 
-local_ast = File.expand_path('../rubocop-ast', __dir__)
-gem 'rubocop-ast', path: local_ast if Dir.exist? local_ast
+# local_ast = File.expand_path('../rubocop-ast', __dir__)
+# gem 'rubocop-ast', path: local_ast if Dir.exist? local_ast
```
koic, line 35: Please revert this to its original state.

changelog/change_improve_style_invertible_unless_condition_message.md
```diff
@@ -0,0 +1 @@
+* [#12562](https://github.com/rubocop/rubocop/pull/12562): Improve `Style/InvertibleUnlessCondition` message. ([@sambostock][])
```
koic, line 1: There is no need for a changelog entry regarding this change.

Gemfile
sambostock: Whoops, that wasn't meant to be committed.

bbatsov:
Looks good to me. Thanks!

PR #12584, opened 2023-12-30
[Fix #11878] Add new `Style/MapIntoArray` cop

### LINKED ISSUE #11878: Style cop idea: prefer `map` over `each` + `<<`
I often encounter this pattern, especially amongst junior devs; they start with the `[]` literal, and then build up a list by repeatedly pushing elements into it. It's messy and inefficient.

```ruby
# bad
result = []
arr.each { |x| result << x + 42 }

# good
result = arr.map { |x| x + 42 }
```

ymap:
Fixes #11878.

This PR adds new `Style/MapIntoArray` cop.
This cop checks for usages of `each` with `<<`, `push`, or `append` which can be replaced by `map`. It is unsafe, because it can produce false positives if the receiver is not an `Enumerable`.

```ruby
# bad
dest = []
src.each { |e| dest << e * 2 }
dest

# good
dest = src.map { |e| e * 2 }

# good - contains another operation
dest = []
src.each { |e| dest << e * 2; puts e }
dest
```

-----------------

* [x] Added tests.


lib/rubocop/cop/style/map_into_array.rb
```diff
@@ -0,0 +1,176 @@
...
+      #   # good - contains another operation
+      #   dest = []
+      #   src.each { |e| dest << e * 2; puts e  }
+      #   dest
+      #
+      class MapIntoArray < Base
+        include RangeHelp
+        extend AutoCorrector
+
+        MSG = 'Use `map` instead of `each` to map elements into an array.'
```
bbatsov, line 51: `map/collect`

Perhaps this can be extract from their configuration.

bbatsov:
I like the cop overall, but I'm a bit unsure about the name. @rubocop/rubocop-core Suggestions welcome!

spec/rubocop/cop/style/map_into_array_spec.rb
```diff
@@ -0,0 +1,345 @@
...
+RSpec.describe RuboCop::Cop::Style::MapIntoArray, :config do
+  it 'registers an offense and corrects when using `each` with `<<` to build an array' do
+    expect_offense(<<~RUBY)
+      dest = []
+      src.each { |e| dest << e * 2 }
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `map` instead of `each` to map elements into an array.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      src.map { |e| e * 2 }
```
bquorning, line 12: Shouldn’t the expected correction be
```suggestion
      dest = src.map { |e| e * 2 }
```
Removing a local variable is bound to break people’s code.

spec/rubocop/cop/style/map_into_array_spec.rb
```diff
@@ -0,0 +1,395 @@
...
+    expect_offense(<<~RUBY)
+      dest = []
+      do_something
+      src.each { |e| dest << e * 2 }
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `map` instead of `each` to map elements into an array.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      do_something
+      src.map { |e| e * 2 }
```
mvz, line 50: This autocorrect is extra tricky since if `do_something` is more code it might modify `dest`.

lib/rubocop/cop/style/map_into_array.rb
ymap: Thanks. I've fixed it by using the `Style/CollectionMethods` configuration. How does it look?

spec/rubocop/cop/style/map_into_array_spec.rb
ymap: Fixed and amended!

bbatsov:
I guess it's time to finally merge this. Thanks for the PR, @ymap!

amomchilov:
Ironic that your handle can be read as "why map?" 😄

Great contribution, thanks @ymap!

PR #12586, opened 2023-12-31
[Fix #11033] Change warning message for `Lint/Syntax` when using LSP

### LINKED ISSUE #11033: Every linter diagnostic message is appended with: Using Ruby #{ruby_version} parser; configure using 'TargetRubyVersion' parameter, under 'AllCops'
When using solargraph language server with rubocop each linter diagnostic message has "(Using Ruby 3.0 parser; configure using TargetRubyVersion parameter, under AllCops)" appended to it. 

189960705-36572249-2c56-48e6-9de8-0fbe852e35fc

189960859-58160019-3c35-43d5-9439-af3867912b47

189961527-b4a4305c-eab6-4c34-bc43-88f47463bfef

189965427-442f8963-68c3-40c6-8eeb-ca72570aed41

Above setup is also using neovim in iTerm2 on MacOS.

## Expected behavior

Linter diagnostic messages from rubocop should only show linter info only. Notice how formatter diagnostic messages (in blue) do not have any configuration info appended. Appending information about rubocop configurations if there is a configuration issue would be appropriate however. 

## Actual behavior

Each linter diagnostic message has info about the ruby version parser. This message is distracting and removing it would make intuitive sense. [I initially posted this issue to the solargraph repo](https://github.com/castwide/solargraph/issues/597) who was helpful enough to show me the probable culprit located in [lib/rubocop/cop/lint/syntax.rb](https://github.com/rubocop/rubocop/blob/00e7ac78fb83383ef537e9e08ecf2301a1427578/lib/rubocop/cop/lint/syntax.rb)
Screen Shot 2022-09-30 at 7 10 49 PM

If anyone uses solargraph lsp with vscode or vim with rubocop configured I would expect this to happened to them as well (does it?). If adding rubocop config info to linting diagnostics is a useful feature to them please help me understand why.

## RuboCop version
Screen Shot 2022-09-30 at 7 05 03 PM

I was unsure to post this as a feature request or a bug, forgive me.

koic:
Fixes #11033.

## Summary

This PR changes warning message for `Lint/Syntax` between LSP and the command line.

The redundancy of the offense message probably depends on the context. The context in which the issue has been reported is LSP.

In LSP, source code being typed may still be evaluated as invalid syntax. On the other hand, when executed via the `rubocop` command line, the input is usually complete and evaluated as valid syntax in most cases. This means that suggesting a possible misconfiguration in .rubocop.yml would be sufficient only when using the command line.

So, this is because the difference in evaluation timing of source code. Therefore, it is implemented as a distinction based on the execution context, rather than as a configuration option.

## Additional Information

To resolve this, it seems reasonable to switch the context of the message between LSP and the `rubocop` command line.

As for implementation, while the issue in Solargraph won't be directly resolved, considering a switch to RuboCop's built-in LSP could be an option.

This PR adds the `RuboCop::Base#lsp_mode?` method, which is intended to be used for the LSP mode expansion I'm currently considering. The LSP still can be carved, and I think there is potential to make RuboCop's LSP more user-friendly and improve it.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -481,6 +481,12 @@ module RuboCop
           range.end_pos + @current_offset
         )
       end
+
+      # This experimental feature has been under consideration for a while.
+      # @api private
+      def lsp_mode?
+        ARGV.include?('--lsp')
+      end
     end
   end
 end
--- a/lib/rubocop/cop/lint/syntax.rb
+++ b/lib/rubocop/cop/lint/syntax.rb
@@ -17,9 +17,12 @@ module RuboCop
         private
 
         def add_offense_from_diagnostic(diagnostic, ruby_version)
-          message =
-            "#{diagnostic.message}\n(Using Ruby #{ruby_version} parser; " \
-            'configure using `TargetRubyVersion` parameter, under `AllCops`)'
+          message = if lsp_mode?
+                      diagnostic.message
+                    else
+                      "#{diagnostic.message}\n(Using Ruby #{ruby_version} parser; " \
+                        'configure using `TargetRubyVersion` parameter, under `AllCops`)'
+                    end
           add_offense(diagnostic.location, message: message, severity: diagnostic.level)
         end
 
--- a/lib/rubocop/rspec/shared_contexts.rb
+++ b/lib/rubocop/rspec/shared_contexts.rb
@@ -128,6 +128,12 @@ RSpec.shared_context 'mock console output' do
   end
 end
 
+RSpec.shared_context 'lsp mode' do
+  before do
+    allow(cop).to receive(:lsp_mode?).and_return(true)
+  end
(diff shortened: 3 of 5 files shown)


bbatsov:
> In LSP, source code being typed may still be evaluated as invalid syntax.

But isn't this a standard issue with any linter integration in an editor? I was under the impression that lint checks get triggered when an user stops typing or something like this.

koic:
Yes, that is correct, and it is a common issue in editor integration.

For editor integration, it is likely that control will be achieved through a new command-line argument or an environment variable, but that new parameter is not added in this PR.

Initially, support can be started with the already RuboCop's built-in LSP mode (i. e., the `--lsp` command-line option), and later, expansion can be followed, I think.

bbatsov:
Okay, I guess that will do for now.

PR #12605, opened 2024-01-10
[Fix #12400] Add new cop `Layout/EmptyLineAfterRequires`

### LINKED ISSUE #12400: New cop idea: `Layout/EmptyLineAfterLeadingRequires`

I often see these two patterns of empty lines at the beginning of ruby files, and I feel it would be better if the style of empty lines were unified:

### Pattern 1

```ruby
# frozen_string_literal: true

require 'foo'
bar
baz
```

### Pattern 2

```ruby
# frozen_string_literal: true

require 'foo'

bar
baz
```

How about adding a new cop like `Layout/EmptyLineAfterLeadingRequires` ?

```ruby
# bad
require 'foo'
bar
baz
```

```ruby
# good
require 'foo'

bar
baz
```

The implementation itself is probably easy, so if it looks good, I will create a pull request.

I once considered whether this cop should also target other `require`s not located at the beginning of the file, but it seems that it would be better not to do so.

r7kamura:
Fix https://github.com/rubocop/rubocop/issues/12400.

-----------------

* [x] Added tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,6 +9,7 @@ task release: 'changelog:check_clean' # Before task is required
 
 require 'bundler'
 require 'bundler/gem_tasks'
+
 begin
   Bundler.setup(:default, :development)
 rescue Bundler::BundlerError => e
@@ -16,6 +17,7 @@ rescue Bundler::BundlerError => e
   warn 'Run `bundle install` to install missing gems'
   exit e.status_code
 end
+
 require 'rake'
 require 'rubocop/rake_task'
 
@@ -27,6 +29,7 @@ RuboCop::RakeTask.new(:internal_investigation)
 task default: %i[documentation_syntax_check spec ascii_spec internal_investigation]
 
 require 'yard'
+
 YARD::Rake::YardocTask.new
 
 desc 'Benchmark a cop on given source file/dir'
--- a/config/default.yml
+++ b/config/default.yml
@@ -584,10 +584,10 @@ Layout/EmptyLineAfterRequires:
   Description: 'Add an empty line after requires.'
   Enabled: pending
   VersionAdded: '<<next>>'
-  EnforcedStyle: beginning_of_file_only
+  EnforcedStyle: top_level
   SupportedStyles:
-    - all
-    - beginning_of_file_only
+    - always
+    - top_level
   RequireMethodNames:
     - require
     - require_relative
--- a/config/default.yml
+++ b/config/default.yml
@@ -580,6 +580,18 @@ Layout/EmptyLineAfterMultilineCondition:
   Reference:
     - https://github.com/airbnb/ruby#multiline-if-newline
 
+Layout/EmptyLineAfterRequires:
+  Description: 'Add an empty line after requires.'
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  EnforcedStyle: beginning_of_file_only
+  SupportedStyles:
+    - all
+    - beginning_of_file_only
+  RequireMethodNames:
+    - require
+    - require_relative
+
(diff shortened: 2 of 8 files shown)


config/default.yml
```diff
@@ -580,6 +580,18 @@ Layout/EmptyLineAfterMultilineCondition:
   Reference:
     - https://github.com/airbnb/ruby#multiline-if-newline
 
+Layout/EmptyLineAfterRequires:
+  Description: 'Add an empty line after requires.'
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  EnforcedStyle: beginning_of_file_only
+  SupportedStyles:
```
bbatsov, line 588: Are those names in line with other similar cops? 

I haven't checked them, but it seems to me that something like `always` and `beginning_of_file` might be better options. I'm also wondering if the beginning of file style makes sense in general, as I've rarely seen this in the wild.

lib/rubocop/cop/layout/empty_line_after_requires.rb
```diff
@@ -0,0 +1,122 @@
...
+      #   bar
+      #
+      #   # good
+      #   foo
+      #
+      #   require 'a'
+      #
+      #   bar
+      #
+      # @example RequireMethodNames: ['require', 'require_relative'] (default)
```
bbatsov, line 41: Will the cop handle properly several `require` groups with blank lines between them? This happens from time to time when the ordering of the requires matters.

r7kamura: Yes. Successive `require`s are treated as a single group, and this cop checks for an empty line after each group.

```ruby
require 'a'
require 'b'

foo
```

Another example:

```ruby
require 'a'
require 'b'

require 'c'
require 'd'

require 'e'
require 'f'
```

config/default.yml
r7kamura: > something like `always` and `beginning_of_file` might be better options

Those names sounds better. I will change them that way.

r7kamura: > I'm also wondering if the beginning of file style makes sense in general, as I've rarely seen this in the wild.

I too think this is a bit controversial.

For example, this would make the following code valid:

```ruby
require 'a'
require 'b'

require 'c'
require 'd'
class Foo
  # ...
end
```

Ideally, I would like the above code to look like this:

```ruby
require 'a'
require 'b'

require 'c'
require 'd'

class Foo
  # ...
end
```

So it might be better to nuance it as `top_level` instead of `beginning_of_file` and change the implementation accordingly 🤔

bbatsov: Yeah, that seems better to me.

Rakefile
```diff
@@ -27,6 +29,7 @@ RuboCop::RakeTask.new(:internal_investigation)
 task default: %i[documentation_syntax_check spec ascii_spec internal_investigation]
 
 require 'yard'
+
 YARD::Rake::YardocTask.new
```
r7kamura, line 33: hmm, I'm sorry to suggest this myself, but I don't seem to want to add an empty line to this kind of code, and there may be too many cases where this Cop won't work well.

This rule might have worked in a language like JavaScript, where there is an advantage to grouping `import` at the beginning of a file for static imports, but in Ruby, `require` often appears abruptly in the middle of a file, like this code.

r7kamura:
As commented at https://github.com/rubocop/rubocop/pull/12605#discussion_r1449040049, I found that there are too many cases where this cop will not work, so I am going to withdraw this proposal, along with the original issue. (Close https://github.com/rubocop/rubocop/issues/12400)

Sorry I went to the trouble of getting you to review it. If I come up with a better idea, I'll suggest it again.

PR #12610, opened 2024-01-11
Accept parentheses in argument calls with blocks

gsamokovarov:
In `Style/MethodCallWithArgsParentheses` with `omit_parentheses` enforced style we need to allow parentheses in method calls with blocks, when their value is used as an arguments in method dispatch calls. We have syntax errors when we enforce their omissions:

```ruby
foo(
  Class.new(Base) do
    # valid Ruby
  end
)
```

Should be allowed as the example below is invalid Ruby:

```ruby
foo(
  Class.new Base do
    # invalid Ruby
  end
)
```

### DIFF
--- a/lib/rubocop/cop/style/method_call_with_args_parentheses/omit_parentheses.rb
+++ b/lib/rubocop/cop/style/method_call_with_args_parentheses/omit_parentheses.rb
@@ -127,23 +127,31 @@ module RuboCop
 
           def call_with_ambiguous_arguments?(node) # rubocop:disable Metrics/PerceivedComplexity
             call_with_braced_block?(node) ||
+              call_in_argument_with_block?(node) ||
               call_as_argument_or_chain?(node) ||
               call_in_match_pattern?(node) ||
               hash_literal_in_arguments?(node) ||
               node.descendants.any? do |n|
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/method_call_with_args_parentheses/omit_parentheses.rb
```diff
@@ -127,23 +127,31 @@ def call_in_single_line_inheritance?(node)
...
-                n.forwarded_args_type? || ambiguous_literal?(n) || logical_operator?(n) ||
-                  call_with_braced_block?(n)
+                n.forwarded_args_type? || n.block_type? ||
+                  ambiguous_literal?(n) || logical_operator?(n)
               end
           end
 
           def call_with_braced_block?(node)
-            (node.send_type? || node.super_type?) && node.block_node&.braces?
+            (node.call_type? || node.super_type?) && node.block_node&.braces?
```
koic, line 141: Can you add a spec for `csend` type?

gsamokovarov: Sure, just did. Converted the tests to foo-bar language for them to make sense with the csend.

koic:
Thanks!

Issue #12621, opened 2024-01-15
Style/ArgumentsForwarding is detecting false positives


dylan-hoefsloot:
I am getting false positives, which if autocorrected break the code

Here is an example:
```ruby
def before_transition(options = {}, &block)
  add_callback(callback_type: :before, callback_class: Callback, from: options[:from], to: options[:to], &block)
end
```

when I run `rubocop` I get the warning:
```
Style/ArgumentsForwarding: Use shorthand syntax ... for arguments forwarding.
```

Which seemed incorrect to me, but just in case I was missing something I tried running autocorrect which produced:
```ruby
def before_transition(...)
  add_callback(...)
end
```

Which is completely broken and a change of behaviour.

--------

## Expected behavior

No warning raised

## Actual behavior

Raised a false positive

## Steps to reproduce the problem

As above

## RuboCop version

```
$ [bundle exec] rubocop -V
1.60.0 (using Parser 3.3.0.4, rubocop-ast 1.30.0, running on ruby 3.2.1) [arm64-darwin22]
  - rubocop-performance 1.20.2
  - rubocop-rspec 2.26.1
```

AlexWayfer:
I have kind of similar problem:

```ruby
def view(path = nil, options = {}, &block)
  cache = options.delete(:cache)
  cache = config[:environment] == 'production' if cache.nil?
  template = Flame::Render.new(
    self,
    (path || caller_locations(1, 1)[0].label.to_sym),
    options
  )
  template.render(cache: cache, &block)
end
```

```
lib/flame/controller.rb:128:12: C: [Correctable] Style/ArgumentsForwarding: Use shorthand syntax ... for arguments forwarding.
		def view(path = nil, options = {}, &block)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/flame/controller.rb:136:20: C: [Correctable] Style/ArgumentsForwarding: Use shorthand syntax ... for arguments forwarding.
			template.render(cache: cache, &block)
                   ^^^^^^^^^^^^^^^^^^^^
```

![image](https://github.com/rubocop/rubocop/assets/6510020/5f1afad9-9cc6-477c-81bf-6d05c69198f9)

How should I replace there something via `...`?

RuboCop v1.60, Target Ruby version is 3.0.

koic:
The issue is caused by the same reason as #12618 and is planned to be resolved in #12624.

Issue #12620, opened 2024-01-15
`Style/ArgumentsForwarding` unsafe autocorrect example


Zajn:
Hi, we updated to `rubocop 1.60.0` today and had several auto-correctable offenses for `Style/ArgumentsForwarding`. Running `bundle exec rubocop --only Style/ArgumentsForwarding -A` yielded updated code that does not do the same thing as it did before auto-correction. 

The corrections that caused issues in our case were examples in which a block was forwarded to another method, along with additional arguments. 

If this is already handled as a configuration option, I apologize for not searching hard enough.

--------

## Expected behavior

```ruby
def use_time_zone(&)
  Time.use_zone(current_identity.time_zone, &)
end
```
This code should either require manual updating, or be corrected to not clobber the non-forwarded arguments (`current_identity.time_zone`, in this case)

## Actual behavior

The above code is auto-corrected to:
```ruby
def use_time_zone(...)
  Time.use_zone(...)
end
```
which does not perform the same.

## Steps to reproduce the problem
Run `bundle exec rubocop --only Style/ArgumentsForwarding -A` on above code

## RuboCop version
```
1.60.0 (using Parser 3.3.0.3, rubocop-ast 1.30.0, running on ruby 3.2.2) [arm64-darwin21]
  - rubocop-performance 1.20.2
  - rubocop-rails 2.23.1
  - rubocop-rspec 2.26.1
```

koic:
The issue is caused by the same reason as #12618 and is planned to be resolved in #12624.

Issue #12619, opened 2024-01-15
Bad autocorrect by Style/ArgumentsForwarding


mvz:
Style/ArgumentsForwarding removes optional arguments when autocorrecting while not removing their use.

## Expected behavior

Given the following code:

```ruby
def local_nodes(type, ignored = [], &blk)
  ignored |= [:class, :module]
  exp.each_node(type, ignored, &blk)
end
```

I expect RuboCop to not change this code, or replace `&blk` with `&`, or replace `&blk` with `...` (but see #12618).

## Actual behavior

RuboCop changes the code to:

```ruby
def local_nodes(type, ...)
  ignored |= [:class, :module]
  exp.each_node(type, ignored, ...)
end
```

This makes `ignored` always equal to `true` when passed to the `#each_node` method.

## Steps to reproduce the problem

1. Create a file `foo.rb` with the content given in the first code block above.
2.  Run `rubocop -a`

## RuboCop version

```
$ bundle exec rubocop -V
1.60.0 (using Parser 3.3.0.3, rubocop-ast 1.30.0, running on ruby 3.3.0) [x86_64-linux]
  - rubocop-performance 1.20.2
  - rubocop-rspec 2.26.1
```

znz:
I can reproduce a similar issue with `TargetRubyVersion: '3.1'`.
And I cannot reproduce with `TargetRubyVersion: '3.2'`.

```
% docker run --rm -it ruby:3.1 /bin/bash
root@71a1d3a91de0:/# gem i rubocop
Fetching ast-2.4.2.gem
Fetching unicode-display_width-2.5.0.gem
Fetching parallel-1.24.0.gem
Fetching parser-3.3.0.4.gem
Fetching rubocop-ast-1.30.0.gem
Fetching rainbow-3.1.1.gem
Fetching ruby-progressbar-1.13.0.gem
Fetching rubocop-1.60.0.gem
Fetching regexp_parser-2.9.0.gem
Fetching language_server-protocol-3.17.0.3.gem
Successfully installed unicode-display_width-2.5.0
Successfully installed ruby-progressbar-1.13.0
Successfully installed ast-2.4.2
Successfully installed parser-3.3.0.4
Successfully installed rubocop-ast-1.30.0
Successfully installed regexp_parser-2.9.0
Successfully installed rainbow-3.1.1
Successfully installed parallel-1.24.0
Successfully installed language_server-protocol-3.17.0.3
Successfully installed rubocop-1.60.0
[...]

PR #12624, opened 2024-01-17
[Fix #12618] Fix false positives for `Style/ArgumentsForwarding`

### LINKED ISSUE #12618: False positive for Style/ArgumentsForwarding when only a block argument can be forwarded anonymously
In the new RuboCop version 1.60.0, the Style/ArgumentsForwarding cop now autocorrects argument lists with only a single block argument, or positional arguments combined with a block argument, to use `...` instead of the block argument. While this does not change things when the method is called in the original way, it does change the arity and will now forward (additional) positional and keyword arguments.

## Expected behavior

Given the following methods:
```ruby
def foo(&block)
  bar(&block)
end

def baz(qux, quuz, &block)
  bar(qux, quuz, &block)
end
```

I expect RuboCop _not_ to autocorrect when targeting Ruby before 3.1.

When targeting 3.1+, I expect it to autocorrect to:

```ruby
def foo(&)
  bar(&)
end

def baz(qux, quuz, &)
  bar(qux, quuz, &)
end
```

## Actual behavior

RuboCop autocorrects as follows, regardless of whether Ruby 3.0 or 3.1 is targeted.

```ruby
def foo(...)
  bar(...)
end

def baz(qux, quuz, ...)
  bar(qux, quuz, ...)
end
```

1. Create a file `foo.rb` with the content given in the first code block above.
2. Run `rubocop -A`

## RuboCop version

```
$ bundle exec rubocop -V
1.60.0 (using Parser 3.3.0.3, rubocop-ast 1.30.0, running on ruby 3.3.0) [x86_64-linux]
  - rubocop-performance 1.20.2
  - rubocop-rspec 2.26.1
```

koic:
Fixes #12618, #12619, #12620, and #12621.

This PR fixes false positives for `Style/ArgumentsForwarding` when using block argument fowarding with other arguments.

And this PR makes `Style/ArgumentsForwarding` aware only block argument fowarding when using Ruby 3.1+.

-----------------

* [x] Added tests.


lib/rubocop/cop/registry.rb
```diff
@@ -240,12 +240,12 @@ def sort!
...
-      def select(...)
-        cops.select(...)
+      def select(&block)
+        cops.select(&block)
       end
 
-      def each(...)
-        cops.each(...)
+      def each(&block)
+        cops.each(&block)
```
koic, line 248: Note: As mentioned in #12618, it revers the incorrect changes made in #12541.

koic:
This is a relatively critical bug with many reports coming in. It would be good to release a fix for it as soon as possible.

spec/rubocop/cop/style/arguments_forwarding_spec.rb
```diff
@@ -173,6 +173,30 @@ def foo(...)
       RUBY
     end
 
+    it 'registers an offense when using block arg', :ruby30 do
```
dvandersluis, line 176:
```suggestion
    it 'does not register an offense when using block arg', :ruby30 do
```

dvandersluis:
Thanks for fixing this @koic!

bbatsov:
I'll cut a bugfix release later today.

PR #12641, opened 2024-01-22
Make error message clearer when the namespace is incorrect

maruth-stripe:
Updates the error message when the namespace is incorrect to specify the exact action the developer must take to fix the error. 

The error message as it exists today does not immediately signal an action to fix it to developers unfamiliar with Rubocop terminology.

-----------------

* [X] Added tests.

### DIFF
--- a/changelog/change_error_message_on_incorrect_namespace.md
+++ b/changelog/change_error_message_on_incorrect_namespace.md
@@ -1,0 +1,1 @@
+* [#12641](https://github.com/rubocop/rubocop/pull/12641): Make error message clearer when the namespace is incorrect. ([@maruth-stripe][])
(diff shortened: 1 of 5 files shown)


bbatsov:
I'm on the fence about the proposed message change, as I don't think the old one was unclear. @rubocop/rubocop-core any thoughts?

bquorning (approved):
I find the new text easier to understand.

Please squash commits though 😅

koic:
There are pros and cons to both the before and after changes of the message. I don't have a strong opinion on this. It's left to discretion for this.

maruth-stripe:
@bquorning squashed the commits! 

@bbatsov Thanks for the feedback! While the old message may not be unclear, developers have told us they would find a message like the new one more clear as it tells them precisely what change to make. Hence, just trying to improve that experience for them :D

bbatsov:
Fair enough. I'm OK with the proposed change.

PR #12647, opened 2024-01-24
LSP mode configurable from external source

koic:
The original LSP mode determination relied on the `--lsp` command-line argument, making it hard to set the mode from 3rd party LSP(-like) tools. This PR refines the API to resolve this issue. Additionally, as a side effect, it improves the mocking part of spec.

However, the APIs are still experimental and not yet mature.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -84,6 +84,24 @@ module RuboCop
         false
       end
 
+      ### LSP
+
+      # This experimental feature has been under consideration for a while.
+      # @api private
+      def self.lsp_mode?
+        !!@lsp_mode
+      end
+
+      # This experimental feature has been under consideration for a while.
+      def self.enable_lsp_mode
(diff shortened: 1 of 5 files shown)


bbatsov:
> making it hard to set the mode from 3rd party LSP(-like) tools

Can you share some example use-cases these changes target?

koic:
There are no specific use cases as examples yet. However, the root cause of something like https://github.com/rubocop/rubocop/issues/11033 was discovered in Solargraph. To resolve this, the API provided in this PR could be useful. I'm not familiar with Solargraph in detail, but it seems that applying a patch to the following could solve https://github.com/rubocop/rubocop/issues/11033.
https://github.com/castwide/solargraph/blob/v0.50.0/lib/solargraph/diagnostics/rubocop.rb#L25-L40

bbatsov:
Got it. Let's try it out then.

PR #12648, opened 2024-01-24
Fix numblock regressions in `omit_parentheses` Style/MethodCallWithArgsParentheses`

gsamokovarov:
My [last change](https://github.com/rubocop/rubocop/pull/12610) was trying to fix an edge case where RuboCop wanted us to remove parentheses in the following example, but removing the parens was resulting in a `SyntaxError`:

```ruby
AnyCable.middleware.use(
  Class.new(AnyCable::Middleware) do
           ^^^^^^^^^^^^^^^^^^^^^^ Omit parentheses for method calls with arguments.
    pass
  end
)
```

As it often happens, though, I broke a few other cases where we now want to remove parens, while the removal can result in ambiguous code:

**Case 1**

```ruby
Foo::Bar.find(pending.things.map { _1['code'] })
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Omit parentheses for method calls with arguments.
```

While we can remove the parentheses in *Case 1*, the author has to be well aware of the difference between the do/end and braced blocks method bounding semantics. Allowing the author to put the parens can remove this ambiguity and we used to allow it.

**Case 2**

```ruby
[a, b].map { _1.call 'something' }.uniq.join(' - ')
                                            ^^^^^^^ Omit parentheses for method calls with arguments.
```

If we set `AllowParenthesesInChaining: true`, we should allow parentheses in chained calls. However, this is broken in current RuboCop.

Both of the issues were caused by a 'refactoring' that replaced a braced block check that handled `numblock`s with a more general `block` check while forgetting about `numblock`s.

### DIFF
--- a/changelog/fix_numblock_regressions_in_omit_parentheses.md
+++ b/changelog/fix_numblock_regressions_in_omit_parentheses.md
@@ -1,0 +1,1 @@
+* [#12648](https://github.com/rubocop/rubocop/pull/12648): Fix numblock regressions in `omit_parentheses` `Style/MethodCallWithArgsParentheses`. ([@gsamokovarov][])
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

PR #12657, opened 2024-01-28
Support `AutoCorrect: contextual` option for LSP

koic:
## Summary

This PR introduces `AutoCorrect: contextual` option that prevents autocorrection during typing in LSP.

This option extends the existing `AutoCorrect` parameter.

- Before: `AutoCorrect: true` or `AutoCorrect: false`
- After: `AutoCorrect: always`, `AutoCorrect: disabled`, or `AutoCorrect: contextual`

Note, `AutoCorrect: always` maintains compatibility with `AutoCorrect: true`,
and `AutoCorrect: disabled` maintains compatibility with `AutoCorrect: false`.

## Details

For example, `Style/EmptyMethod` should not autocorrect to one-liner while in the process of writing the body of the method.
This isn't an issue with autocorrection itself, but rather due to differing requirements in contexts such as editing (LSP) and command line (e.g., CI).

So, `AutoCorrect: contextual` is set for some cops like this. There may be other cases, but for the obvious following cops. The classification is as follows:

### Cops that might remove code being edited

- `Layout/EmptyComment` cop
- `Lint/EmptyConditionalBody` cop
- `Lint/EmptyEnsure` cop
- `Lint/EmptyInterpolation` cop
- `Lint/TrailingCommaInAttributeDeclaration` cop
- `Lint/UselessAccessModifier` cop
- `Lint/UselessAssignment` cop
- `Lint/UselessMethodDefinition` cop
- `Lint/UselessTimes` cop
- `Lint/Void` cop
- `Style/EmptyElse` cop
- `Style/RedundantInitialize` cop

### Cops that might adjust code being edited

- `Lint/UnusedBlockArgument` cop
- `Lint/UnusedMethodArgument` cop
- `Style/EmptyHeredoc` cop
- `Style/EmptyMethod` cop

As a result, it allows for a distinction in the use of autocorrection between the LSP and command-line contexts.

## Other Information

Here is why the parameter name was chosen as `contextual`.

In this PR, autocorrection from LSP is always not applied in the case of `contextual`. However, even when `contextual` is set, I was considering a possibility for autocorrection to be forcibly applied from LSP under certain conditions.

Therefore, the initial thought was to name it `contextual`, but it might have been too abstract in context, so it was changed to `contextual`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_extend_autocorrect_option_for_lsp.md
+++ b/changelog/new_extend_autocorrect_option_for_lsp.md
@@ -1,0 +1,1 @@
+* [#12657](https://github.com/rubocop/rubocop/pull/12657): Support `AutoCorrect: contextual` option for LSP. ([@koic][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -556,7 +556,9 @@ Layout/ElseAlignment:
 Layout/EmptyComment:
   Description: 'Checks empty comment.'
   Enabled: true
+  AutoCorrect: contextual
   VersionAdded: '0.53'
+  VersionChanged: '<<next>>'
   AllowBorderComment: true
   AllowMarginComment: true
 
@@ -1825,16 +1827,18 @@ Lint/EmptyClass:
 Lint/EmptyConditionalBody:
   Description: 'Checks for the presence of `if`, `elsif` and `unless` branches without a body.'
   Enabled: true
+  AutoCorrect: contextual
   SafeAutoCorrect: false
   AllowComments: true
   VersionAdded: '0.89'
-  VersionChanged: '1.34'
+  VersionChanged: '<<next>>'
 
 Lint/EmptyEnsure:
   Description: 'Checks for empty ensure block.'
   Enabled: true
+  AutoCorrect: contextual
   VersionAdded: '0.10'
-  VersionChanged: '0.48'
+  VersionChanged: '<<next>>'
 
 Lint/EmptyExpression:
   Description: 'Checks for empty expressions.'
@@ -1856,8 +1860,9 @@ Lint/EmptyInPattern:
 Lint/EmptyInterpolation:
   Description: 'Checks for empty string interpolation.'
   Enabled: true
+  AutoCorrect: contextual
   VersionAdded: '0.20'
-  VersionChanged: '0.45'
+  VersionChanged: '<<next>>'
 
 Lint/EmptyWhen:
   Description: 'Checks for `when` branches with empty bodies.'
@@ -2395,7 +2400,9 @@ Lint/TopLevelReturnWithArgument:
 Lint/TrailingCommaInAttributeDeclaration:
   Description: 'Checks for trailing commas in attribute declarations.'
   Enabled: true
+  AutoCorrect: contextual
   VersionAdded: '0.90'
+  VersionChanged: '<<next>>'
 
 Lint/TripleQuotes:
   Description: 'Checks for useless triple quote constructs.'
@@ -2455,8 +2462,9 @@ Lint/UnusedBlockArgument:
   Description: 'Checks for unused block arguments.'
   StyleGuide: '#underscore-unused-vars'
   Enabled: true
+  AutoCorrect: contextual
   VersionAdded: '0.21'
-  VersionChanged: '0.22'
+  VersionChanged: '<<next>>'
   IgnoreEmptyBlocks: true
   AllowUnusedKeywordArguments: false
 
@@ -2464,8 +2472,9 @@ Lint/UnusedMethodArgument:
   Description: 'Checks for unused method arguments.'
   StyleGuide: '#underscore-unused-vars'
   Enabled: true
+  AutoCorrect: contextual
   VersionAdded: '0.21'
-  VersionChanged: '0.81'
+  VersionChanged: '<<next>>'
   AllowUnusedKeywordArguments: false
   IgnoreEmptyMethods: true
   IgnoreNotImplementedMethods: true
@@ -2489,8 +2498,9 @@ Lint/UriRegexp:
 Lint/UselessAccessModifier:
   Description: 'Checks for useless access modifiers.'
   Enabled: true
+  AutoCorrect: contextual
   VersionAdded: '0.20'
-  VersionChanged: '0.83'
+  VersionChanged: '<<next>>'
   ContextCreatingMethods: []
   MethodCreatingMethods: []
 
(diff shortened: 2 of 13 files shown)


bbatsov:
I like the spirit of the suggestion, but I'm not sure about the proposed names. E.g. it's not clear even to me what `except_on_edit` is supposed to do. 

I'm always wondering if `enabled` and `disabled` (or `on`/`off`) are not better synonyms for `true`/`false`. @rubocop/rubocop-core Feedback and ideas welcome!

bbatsov:
> For example, Style/EmptyMethod should not autocorrect to one-liner while in the process of writing the body of the method.
This isn't an issue with autocorrection itself, but rather due to differing requirements in contexts such as editing (LSP) and command line (e.g., CI).

It's not clear to me what this will mean in practice - is the idea that the new setting will apply when LSP mode is active? How do we decide if the user is running RuboCop while editing the code?

koic:
> It's not clear to me what this will mean in practice - is the idea that the new setting will apply when LSP mode is active? How do we decide if the user is running RuboCop while editing the code?

When using LSP in an editor, having autocorrect executed can sometimes be problematic. For instance, consider autocorrect being triggered upon saving a file.

The definition is written before the body of a method:

```ruby
def do_something
end
```

Upon saving, it gets autocorrected to a one-liner:

```ruby
def do_something; end
```

However, during the implementation of this method, it is preferred not to have it autocorrected, allowing for the continuation of writing the method body.

Another interesting example, with RuboCop RSpec's `RSpec/Focus`, if it's autocorrected by LSP, `fit` gets automatically changed to `it`, preventing the execution of specific `fit` test during development.

In essence, this feature modifies the behavior of autocorrect on editors utilizing LSP. By setting `expect_on_edit`, RuboCop will register offenses only and refrain from executing autocorrections on LSP editor.

koic:
And, I'm still thinking about a better name to replace `expect_on_edit`, but I haven't come up with one yet.

bbatsov:
> When using LSP in an editor, having autocorrect executed can sometimes be problematic. For instance, consider autocorrect being triggered upon saving a file.

I get the problem, but I don't get how are we going to know if something is being edited or not, so this new option would be considered.

koic:
Yeah, this is the main point. The explanation is a bit long, but please stay with me.

It's hard to tell if some source code is being edited or not just by looking at it. So, here's what was think about.

## State of Source Code

Source code can be in the following states: either the input is complete, or it's not.

```mermaid
stateDiagram-v2
    [*] --> InvalidSyntax: Start Editing
    InvalidSyntax --> ValidSyntax: While Editing
    ValidSyntax --> InvalidSyntax: While Editing
    ValidSyntax --> Completed: Finish Editing
    Completed --> [*]: End
```

### Completed (Finish Editing)

When the source code input is complete, it's clear and has no syntax errors, normally. So, I think the status of the completed source code is clear.

### Valid Syntax and Invalid Syntax (While Editing)

The hard part is when the source code is not complete. There are two cases:

- Valid Syntax: The syntax is valid, but it's still being written.
- Invalid Syntax: The syntax is invalid and still being written.

This PR doesn't handle with "Invalid Syntax". Hopefully, a future error-tolerant parser will handle this. Currently always handled by `Lint/Syntax` cop.

So, the focus is on "Valid Syntax" with valid syntax. The question is whether the editing is finished or not in these cases.

## Deciding if Valid Syntax Source Code is Being Edited

The feature in this PR decides as follows:

- If it's via LSP, it's always seen as source code being edited.
- If it's via the command line, it's seen as source code where editing is completed.

```mermaid
stateDiagram-v2
    [*] --> EditingSourceCode: (Handle with LSP)
    EditingSourceCode --> CompletedSourceCode: Finish Editing
    [*] --> CompletedSourceCode: (Handle with Command Line)
```

Now, `RuboCop::LSP.enabled?`, added in #12679, is important. This flag helps decide if the check is via LSP.

If the check is via LSP, it's seen as source code being edited, so cops with `AutoCorrect: except_on_edit` won't autocorrect.

```ruby
# Logic to determine if auto-correction is enabled in the `RuboCop::Cop::Base` class.
RuboCop::LSP.enabled? && support_autocorrect_except_on_edit?
```

That's how this feature figures out if source code is being edited or not.

koic:
Regarding the naming of the `AutoCorrect` parameter, how about updating the names as follows?

## Before

- `AutoCorrect: always` ... Always support autocorrection
- `AutoCorrect: expect_on_edit` ... Support autocorrection except when using LSP
- `AutoCorrect: off` ... Never support autocorrection

## After

- `AutoCorrect: enable` ... Always support autocorrection
- `AutoCorrect: contextual` ... Support autocorrection except when using LSP
- `AutoCorrect: disable` ... Never support autocorrection

I'm not sure if `contextual` is the best name, but it's an attempt to express that the availability of autocorrection depends on the context.

bbatsov:
I like `contextual`, as it also gives us some flexibility. As word pairings - it probably goes better with `always` and `off`/`disabled`. 

Btw, I'm wondering if we can also do something about editors that are running RuboCop directly (e.g. by shelling out) as they are going to have more or less the same problem as with LSP when editing code. Perhaps we can also add some command-line flag to reflect such usage?

koic:
Then, the parameter names will be updated to `always`, `context`, `off` (or `disabled`).

> Perhaps we can also add some command-line flag to reflect such usage?

That's a great idea! A good option name hasn't been thought of yet, but adding that command-line flag can certainly be considered. Because calling the Ruby code `RuboCop::LSP.enable` is not always easy.

koic:
I've updated this PR. `AutoCorrect` parameter has been set to `always`, `contextual`, `disabled`. For backward compatibility, `true` is treated the same as `always`, and `false` is treated the same as `disabled`.

This is a note. `'off'` is a boolean literal in `YAML` that defaults to returning `false`:

```ruby
$ ruby -ryaml -e "p YAML.load('off')"
false
```

Therefore, while `off` can also be used to disable, `disabled` has been chosen as the official setting to convey a more specific intention.

koic:
I haven't thought of a good name for the command-line flag yet. I plan to continue developing this feature, but it probably would be better to implement it separately from this PR.

bbatsov:
> I haven't thought of a good name for the command-line flag yet. I plan to continue developing this feature, but it probably would be better to implement it separately from this PR.

Perhaps something like `--editor-mode`?

koic:
`--editor-mode` looks like a good name! Thank you!

andyw8:
Issue for supporting this in Ruby LSP: https://github.com/Shopify/ruby-lsp/issues/1449

PR #12658, opened 2024-01-29
Fixes bug in `Metrics/MethodLength` cop. The cop does not count block comments even though count comments setting is turned off.

KarlHeitmann:

The other day I was trying to reduce the number of lines in the methods of my project. I am not a big fan of block comments but I usually use them to toggle code snippets I want to test later on. I discovered my rubocop LSP instance was complaining that the method had a lot of lines, that was because I commented with `=begin ... =end` _a lot_ of lines.

The code changes on this PR attempts to fix the issue, so it will not count block comments. The main problem is this "bad API" [method](https://github.com/rubocop/rubocop/blob/b6ee1c20a553e13c404b0b48137a17120a64e176/lib/rubocop/cop/util.rb#L17) was used by [CodeLengthCalculator#irrelevant_line?](https://github.com/rubocop/rubocop/blob/b6ee1c20a553e13c404b0b48137a17120a64e176/lib/rubocop/cop/metrics/utils/code_length_calculator.rb#L160). So, block comments were not considered by the `CodeLengthCalculator` (because the regex in the "bad API" method does not catch multiline comments (ie, block comments)).

My first approach to solve the issue consisted in using [@processed_source](https://github.com/rubocop/rubocop/blob/b6ee1c20a553e13c404b0b48137a17120a64e176/lib/rubocop/cop/metrics/utils/code_length_calculator.rb#L18) to analyze its tokens array. But I found out that the tokens array value is set using [tokenise(create_parser(ruby_version))](https://github.com/rubocop/rubocop-ast/blob/fa414056753a0682aa4ce18270f1c9263e902e0f/lib/rubocop/ast/processed_source.rb#L210). `create_parser` must use a specific ruby version class depending on the ruby version used. After realizing that, I thought "Mmmhh... maybe this `create_parser` method is very expensive in terms of computation and maybe I should not use it..."

Then, I decided to follow a simpler approach: find all the places [irrelevant_line?](https://github.com/rubocop/rubocop/blob/b6ee1c20a553e13c404b0b48137a17120a64e176/lib/rubocop/cop/metrics/utils/code_length_calculator.rb#L159) method is called, and add a filter stage just before the method is called. The filter stage consists in calling a [recursive method](https://github.com/KarlHeitmann/rubocop/blob/2efb69bdf89914f018b86f5333cefc965509cef3/lib/rubocop/cop/metrics/utils/code_length_calculator.rb#L87) that will remove all lines between a `"=begin"` tag, and `"=end"` tag.

For example, by using this [CodeLengthCalculator#clean_block_comments](https://github.com/KarlHeitmann/rubocop/blob/2efb69bdf89914f018b86f5333cefc965509cef3/lib/rubocop/cop/metrics/utils/code_length_calculator.rb#L87) recursive method, you will transform this array:

```ruby
["a = 1\n", "  # This is a comment\n", "=begin\n", "kkkkkk = 92\n", "=end\n", "  # ANOTHER comment\n", "=begin\n", "a = 6\n", "=end\n", "  b = 2"]
```

and convert that array into this one:

```ruby
["a = 1\n", "  # This is a comment\n", "  # ANOTHER comment\n", "  b = 2"]
```

My solution seems to work, I can add more tests to it if you want more. But I'd love to make a more efficient solution to this bug. Any feedback is welcome.

### Notes on the checklist item:

- I don't know if is worth to squash at least the 2 first commits of this PR. If you'd like me to squash the first 2 or all the commits into one, just ping me here.
- I've introduced some alerts: `Metrics/PerceivedComplexity` and `Metrics/AbcSize`. I don't know how to solve these two issues.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_adds_count_comments_flag_and_cleanup.md
+++ b/changelog/fix_adds_count_comments_flag_and_cleanup.md
@@ -1,0 +1,1 @@
+* [#12658](https://github.com/rubocop/rubocop/pull/12658): Fixes bug in `Metrics/MethodLength` cop: the cop does not count block comments even though count comments setting is turned on. ([@KarlHeitmann][])
--- a/lib/rubocop/cop/metrics/utils/code_length_calculator.rb
+++ b/lib/rubocop/cop/metrics/utils/code_length_calculator.rb
@@ -79,8 +79,8 @@ module RuboCop
                   body.source.lines
                 end
 
-              source = clean_block_comments(source, false) if true
-              source.count { |line| !irrelevant_line?(line) } # This filters out the comments
+              source = clean_block_comments(source, false) unless count_comments?
+              source.count { |line| !irrelevant_line?(line) }
             end
           end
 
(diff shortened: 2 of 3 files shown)


KarlHeitmann:
I completed the items on the checklist. I don't know if I need to squash more commits, and I don't know either how to fix de rubocop metrics related to high complexity.

PR #12660, opened 2024-01-29
Update `crack` to v0.4.6

andyw8:
https://github.com/jnunemaker/crack/pull/75 was merged so this should allow the `bigdecimal` dependency to be removed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,9 +25,7 @@ group :test do
   # FIXME: This `base64` dependency can be removed when https://github.com/bblimke/webmock/pull/1041
   # is merged and released. It's a workaround until then.
   gem 'base64'
-  # FIXME: This `bigdecimal` dependency can be removed when https://github.com/jnunemaker/crack/pull/75
-  # is merged and released. It's a workaround until then.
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -25,9 +25,7 @@ group :test do
   # FIXME: This `base64` dependency can be removed when https://github.com/bblimke/webmock/pull/1041
   # is merged and released. It's a workaround until then.
   gem 'base64'
-  # FIXME: This `bigdecimal` dependency can be removed when https://github.com/jnunemaker/crack/pull/75
-  # is merged and released. It's a workaround until then.
-  gem 'bigdecimal', platform: %i[mri windows]
+  gem 'crack', '>= 0.4.6' # Needed for webmock on Ruby 3.3
```
koic, line 28: Please remove it and update the commit title.

```suggestion
```

koic:
#12659 was opened first, so I merged that.

Gemfile
andyw8: Initially I did intend to just remove this line, but since `Gemfile.lock` isn't committed, I was concerned it would cause problems for people who already have a checked-out copy of the repo.

PR #12664, opened 2024-01-31
Fix handling of textDocument/diagnostic

muxcmux:
The `textDocument` parameter for the `textDocument/diagnostic` request is of type `TextDocumentIdentifier`, which has a single property `uri` of type `DocumentUri`.

Ref: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_diagnostic

According to the spec the `DocumentUri` type does not hold the contents of the document, which the handler currently relies on in order to publish diagnostics to the client.

This breaks the LSP mode in Neovim 0.10, which now supports pull diagnostics, and possibly other editors too. The same problem exists in `standardrb` (https://github.com/standardrb/standard/issues/575)

I have made the `textDocument/diagnostic` hander a no-op, since diagnostics are already being pushed to clients on `textDocument/didChange`.

~~Let me know if you want to merge this in, and I'll add a changelog entry 👍~~

- [x] Add changelog entry

### DIFF
--- a/changelog/fix_handling_of_text_document_diagnostic.md
+++ b/changelog/fix_handling_of_text_document_diagnostic.md
@@ -1,0 +1,1 @@
+* [#12664](https://github.com/rubocop/rubocop/pull/12664): Fix handling of `textDocument/diagnostic`. ([@muxcmux][])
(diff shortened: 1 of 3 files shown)


koic:
Thank you for opening this PR. Whether to merge or not will be determined based on the verification of https://github.com/standardrb/standard/pull/601, but for now, please add a changelog entry file and squash the commits into one.

muxcmux:
Cheers mate, I've added a changelog entry.

PR #12671, opened 2024-02-02
[feedback requested] Detect void value expressions

pablobm:
This is a WIP 🚧  hoping to get some feedback before I commit more effort. I know it doesn't have documentation, changelog, etc. I hope this is fine for the moment 🙏 

Some interesting questions here would be:
- Is there a reason why Rubocop doesn't have a cop for this?
- Is there a desire to have a cop for this?
- Any tips on how to handle this?

### Detail

Recently I came across a new (to me) error in Ruby code similar to the following:

```ruby
def void_assignment
  a = return 1
end
```

This code would fail with the following syntax error:
```
run.rb: run.rb:2: void value expression (SyntaxError)
```

So that's ok, I can understand it. Also it is an interesting one: despite being a syntax error, it's not flagged by Rubocop or ruby-parse. I'm assuming the interpreter catches it after building the AST, as opposed to during the parsing process itself? Whichever way it is, the tools I use (including Rubocop) don't flag it so I don't get to know it's happening until I run the code.

I suspect this is well known scenario and there's a good reason why Rubocop is not catching it. Is this correct?

Just in case, it piqued my curiosity and I had a go at it :-) This PR is a first version of what a cop for this might look like, if there's desire for one. I'm sure I'm missing something. For one, I know that `next` and `break` can also cause a void value expression, so that'd be something to add. And of course documentation.

I have also noticed that `case` statements appear immune to it, so I'm not raising offences on them. However I wonder if that's something that should be an option to the rule.

Additionally, I have noticed that different MRI and JRuby treat this differently, which makes sense. There are edge cases (noted in the spec) where JRuby is better at catching these cases. Also there are questions of what to do in cases that look wrong, but the interpreters I have tried don't complain (see also marked in the spec).

## Checklist

Not fully checked as this is a WIP asking for feedback:

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2668,6 +2668,11 @@ Lint/Void:
 
 #################### Metrics ###############################
 
+Lint/VoidValueExpression:
+  Description: 'TODO: Write a description of the cop.'
+  Enabled: true
+  VersionAdded: '<<next>>'
+
 Metrics/AbcSize:
   Description: >-
                  A calculated magnitude based on number of assignments,
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -437,6 +437,7 @@ require_relative 'rubocop/cop/lint/useless_ruby2_keywords'
 require_relative 'rubocop/cop/lint/useless_setter_call'
 require_relative 'rubocop/cop/lint/useless_times'
 require_relative 'rubocop/cop/lint/void'
+require_relative 'rubocop/cop/lint/void_value_expression'
 
 require_relative 'rubocop/cop/metrics/utils/iterating_block'
 require_relative 'rubocop/cop/metrics/cyclomatic_complexity'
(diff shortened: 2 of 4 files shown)


pablobm:
@koic :wave: Not sure what the correct way to ask for feedback is :sweat_smile: I hope a ping is ok! Thank you in advance.

bbatsov:
> Is there a reason why Rubocop doesn't have a cop for this?

What about `Lint/Void`? :-)

pablobm:
Thank you for having a look @bbatsov! My understanding is that `Lint/Void` does not check for the same thing, at least at the moment. For one, `Lint/Void` doesn't see anything wrong with the example that led me to look into this: `a = return 1`.

Whereas related, I see `Lint/Void` as a check for "abandoned values", whereas what I am trying to detect is void values that would be "mixed" with non-void values.

Perhaps a different way to express this would be: `Lint/Void` checks for non-void where void is expected, whereas this new rule checks for the opposite: void where non-void is expected.

dvandersluis:
Void value expressions not being handled is a known issue of parser as documented here: https://github.com/whitequark/parser?tab=readme-ov-file#void-value-expressions

It seems to me that this should be handled upstream so that it can be flagged by `Lint/Syntax` rather than introducing a new cop for it.

PR #12677, opened 2024-02-06
Fix an error for `Style/HashSyntax` when using `Prism::Translation::Parser`

koic:
This PR fixes the following error `Style/HashSyntax` when using `Prism::Translation::Parser` as a parser:

```console
$ bundle exec ruby -rprism/translation/parser/rubocop $(bundle exec which rubocop) \
  --only Style/HashSyntax -d lib/rubocop/cop/style/not.rb
(snip)

An error occurred while Style/HashSyntax cop was inspecting
/Users/koic/src/github.com/rubocop/rubocop/lib/rubocop/cop/style/not.rb:23:27.
undefined method `loc' for an instance of AST::Node
/Users/koic/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/parser-3.3.0.5/lib/parser/builders/default.rb:1875:
in `join_exprs'
```

Other ways to resolve this might exist, but using `RuboCop::Cop::Base#parser` is not necessary.

Currently, no repro tests have been added as it is sufficient to verify that it is not broken with the Parser gem. And, this is for refactoring aimed at integration with `Prism`, no changelog entry will be added.

cf. https://github.com/ruby/prism/blob/v0.21.0/docs/parser_translation.md#rubocop

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #12678, opened 2024-02-07
Fix an error for `Style/RedundantLineContinuation` when using `Prism::Translation::Parser`

koic:
This PR fixes the following error `Style/RedundantLineContinuation` when using `Prism::Translation::Parser` as a parser:

```console
$ bundle exec ruby -rprism/translation/parser/rubocop $(bundle exec which rubocop)
(snip)

An error occurred while Style/RedundantLineContinuation cop was inspecting
/Users/koic/src/github.com/rubocop/rubocop/lib/rubocop/cop/lint/out_of_range_regexp_ref.rb.
undefined method `complete?' for an instance of AST::Node
/Users/koic/src/github.com/rubocop/rubocop-ast/lib/rubocop/ast/node.rb:104:in `block in initialize'
```

This is a similar error to #12677.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_line_continuation.rb
+++ b/lib/rubocop/cop/style/redundant_line_continuation.rb
@@ -124,8 +124,10 @@ module RuboCop
           return true unless (node = find_node_for_line(range.line))
           return false if argument_newline?(node)
 
-          source = node.parent ? node.parent.source : node.source
-          parse(source.gsub("\\\n", "\n")).valid_syntax?
+          continuation_node = node.parent || node
+          return false if allowed_type?(node) || allowed_type?(continuation_node)
+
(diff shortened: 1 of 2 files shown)


bbatsov:
To be clear - is the a bug in Prism that will be addressed down the road? (I'm asking if we need to leave to some TODO comments)

koic:
This error that occurred with Prism 0.21.0 has been resolved in Prism 0.22.0, as per the comment below:
https://github.com/rubocop/rubocop/pull/12677#issuecomment-1932189221

On the other hand, since there is no need to parse an already parsed target again, I think this change could be merged as a refactoring to avoid using `parse`. So, If necessary, I will edit the commit message and the PR title.

bbatsov:
Got it. No need for further changes.

PR #12679, opened 2024-02-08
Publish `RuboCop::LSP.enable` to enable LSP mode

koic:
`RuboCop::Cop::Base.enable_lsp_mode` will be renamed to `RuboCop::LSP.enable`.

Before:

```ruby
RuboCop::Cop::Base.enable_lsp_mode
```

After:

```ruby
RuboCop::LSP.enable
```

This is because whether or not it is LSP should not be set per cop but should be a state controlled for the entire RuboCop process.

Since `RuboCop::Cop::Base.enable_lsp_mode` was an experimental private API, it will be renamed without a deprecation warning.

Along with this, the module name has been changed from `RuboCop::Lsp` to `RuboCop::LSP`. This change is made to align with Ruby's standard API, like `RubyVM::YJIT.enable`, making it `RuboCop::LSP.enable` instead. `RuboCop::Lsp.enable` does not look as pretty name.

With the renaming to `RuboCop::LSP.enable`, the usage of this API will be documented. This allows libraries like Ruby LSP and Solargraph, which use internal modules other than RuboCop's built-in `RuboCop::LSP::Server#start`, to easily utilize the features in #12586 and #12657.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_publish_lsp_enable_api.md
+++ b/changelog/new_publish_lsp_enable_api.md
@@ -1,0 +1,1 @@
+* [#12679](https://github.com/rubocop/rubocop/pull/12679): Publish `RuboCop::LSP.enable` API to enable LSP mode. ([@koic][])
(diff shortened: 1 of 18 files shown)


bbatsov:
Good change!

PR #12682, opened 2024-02-11
Add `--editor-mode` CLI option

koic:
Follow up https://github.com/rubocop/rubocop/pull/12657#issuecomment-1937084864.

This PR adds `--editor-mode` CLI option, which optimize real-time feedback in editors, adjusting behaviors for editing experience.

Editors that run RuboCop directly (e.g., by shelling out) encounter the same issues as with `--lsp`. This option is designed for such editors.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/options.rb
+++ b/lib/rubocop/options.rb
@@ -95,6 +95,7 @@ module RuboCop
         option(opts, '--ignore-unrecognized-cops')
         option(opts, '--force-default-config')
         option(opts, '-s', '--stdin FILE')
+        option(opts, '--editor-mode')
         option(opts, '-P', '--[no-]parallel')
         option(opts, '--raise-cop-error')
         add_severity_option(opts)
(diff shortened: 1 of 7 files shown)


lib/rubocop/options.rb
```diff
@@ -416,6 +418,13 @@ def validate_display_only_failed_and_display_only_correctable
             format('--display-only-failed cannot be used together with other display options.')
     end
 
+    def validate_lsp_and_editor_mode
+      return if !@options.key?(:lsp) || !@options.key?(:editor_mode)
```
mvz, line 422: It may be clearer to use `return unless options.key(:lsp) && @options.key(:editor_mode)`?

koic: Following De Morgan's laws, it's questionable whether `unless` is appropriate in this case. Anyway, the current condition will be maintained here.

bbatsov:
Looks good to me. Thanks!

PR #12688, opened 2024-02-15
[Fix #12687] Fix a false positive for `Lint/Void`

### LINKED ISSUE #12687: Lint/Void false positive and unsafe autocorrect
Hi, I noticed a bug while using [Parallel)](https://github.com/grosser/parallel/tree/master) .

sample.rb
```ruby
# frozen_string_literal: true
require 'parallel'
items = [1, 2, 3]
Parallel.each(items,
              finish: lambda { |_item, _index, result|
                puts "puts result: #{result}"
              }) do |i|
  puts i * 10
  i * 10
end
```

## Expected behavior
```shell
bundle exec rubocop -a sample.rb
```
No changes should occur, and the following result should be obtained:

```
30
10
20
puts result: 10
puts result: 20
puts result: 30
```

## Actual behavior
The following error occurs:

```
Inspecting 1 file
W

Offenses:

sample.rb:11:3: W: [Corrected] Lint/Void: Variable i used in void context.
  i
  ^
sample.rb:11:5: W: [Corrected] Lint/Void: Operator * used in void context.
  i * 10
    ^
sample.rb:12:1: C: [Corrected] Layout/IndentationConsistency: Inconsistent indentation detected.
10
^^
sample.rb:12:1: W: [Corrected] Lint/Void: Literal 10 used in void context.
10
^^
sample.rb:12:3: W: [Corrected] Lint/Void: Literal 10 used in void context.
  10
  ^^

1 file inspected, 5 offenses detected, 5 offenses corrected
```

This code and results are changed.

```ruby
# frozen_string_literal: true
require 'parallel'
items = [1, 2, 3]
Parallel.each(items,
              finish: lambda { |_item, _index, result|
                puts "puts result: #{result}"
              }) do |i|
  puts i * 10
end
```

```
20
10
puts result:
puts result:
30
puts result:
```

I think `check_void_op` runs when 'each' is detected.
https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/lint/void.rb#L86

Additionally, If you change the return value from `i * 10` to `(i * 10)`, nothing happens. which is storange.

Test case above.
You can pull it from [this gist](https://gist.github.com/yayamochi/d9accbcf1c28e44274333d9725ce6273) and execute rubocop.

## RuboCop version

```
$ [bundle exec] rubocop -V
1.60.2 (using Parser 3.3.0.5, rubocop-ast 1.30.0, running on ruby 3.0.4) [arm64-darwin22]
```

koic:
Fixes #12687.

This PR fixes a false positive for `Lint/Void` when `each` block with conditional expressions that has multiple statements.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/void_spec.rb
+++ b/spec/rubocop/cop/lint/void_spec.rb
@@ -614,6 +614,18 @@ RSpec.describe RuboCop::Cop::Lint::Void, :config do
     RUBY
   end
 
+  it 'does not register `#each` block with conditional expressions that has multiple statements' do
+    expect_no_offenses(<<~RUBY)
+      enumerator_as_filter.each do |item|
+        puts item
+
+        # The `filter` method is used to filter for matches with `42`.
+        # In this case, it's not void.
+        item == 42
+      end
+    RUBY
+  end
+
   context 'Ruby 2.7', :ruby27 do
     it 'registers two offenses for void literals in `#tap` method' do
       expect_offense(<<~RUBY)
--- a/changelog/fix_false_positive_for_lint_void.md
+++ b/changelog/fix_false_positive_for_lint_void.md
@@ -1,0 +1,1 @@
+* [#12687](https://github.com/rubocop/rubocop/issues/12687): Fix a false positive for `Lint/Void` when `each` block with conditional expressions that has multiple statements. ([@koic][])
--- a/lib/rubocop/cop/lint/void.rb
+++ b/lib/rubocop/cop/lint/void.rb
@@ -100,7 +100,12 @@ module RuboCop
           expressions = *node
           expressions.pop unless in_void_context?(node)
           expressions.each do |expr|
-            check_void_op(expr)
+            check_void_op(expr) do
+              block_node = node.each_ancestor(:block).first
+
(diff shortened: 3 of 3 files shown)


spec/rubocop/cop/lint/void_spec.rb
```diff
@@ -614,6 +614,18 @@ def foo=(rhs)
     RUBY
   end
 
+  it 'does not register `#each` block with conditional expressions that has multiple statements' do
+    expect_no_offenses(<<~RUBY)
+      enumerator_as_filter.each do |item|
```
bbatsov, line 619: Won't this be different for this different enumerators, though? Maybe some Allowed list would make more sense than just assuming that multiple expressions are OK all the time.

koic: Initially, I also considered Allowed list approach, but it appears that the issue was simply that the following logic for `on_block` was not accounted for in the implementation of `on_begin`.
https://github.com/rubocop/rubocop/blob/v1.60.2/lib/rubocop/cop/lint/void.rb#L86

So, I think this approach could provide a more generic solution instead of Allowed list.

PR #12708, opened 2024-02-25
[Fix #12695] Handle Include from inherited file in parent directory

### LINKED ISSUE #12695: Unexpected glob behavior when including files
In the brew [repo](https://github.com/Homebrew/brew), I would like to our non-vendored .rbi files (Sorbet's "Ruby Interface" [files](https://sorbet.org/docs/rbi), written in Ruby) to be covered by RuboCop. However, I just noticed that our configuration wasn't behaving as expected.

In particular, I'm seeing that this configuration:
```
AllCops:
  Include:
    - "**/*.rbi"
```
is less inclusive than this one:
```
AllCops:
  Include:
    - "Homebrew/**/*.rbi"
```

As a demonstration, [this](https://github.com/Homebrew/brew/pull/16673/files) PR leads to [these](https://github.com/Homebrew/brew/actions/runs/7924742511/job/21636825927?pr=16673) violations, although there are no violations on the base branch.

--------

## Expected behavior

`**/*.rbi` should be a superset of `Homebrew/**/*.rbi`

## Actual behavior

`Homebrew/**/*.rbi` includes files that `**/*.rbi` does not (I'm not actually sure the latter is including any files)

See the config file and PR run linked above. I don't have a more minimal repro available, but could attempt one if needed.

## RuboCop version

```
1.60.2 (using Parser 3.3.0.5, rubocop-ast 1.30.0, running on ruby 3.1.4) [arm64-darwin20]
with:
    rubocop-ast (1.30.0)
    rubocop-capybara (2.20.0)
    rubocop-factory_bot (2.25.1)
    rubocop-md (1.2.2)
    rubocop-performance (1.20.2)
    rubocop-rspec (2.26.1)
    rubocop-sorbet (0.7.7)
```

jonas054:
When we inherit configuration from a file in a parent or ancestor directory, and that file is called `.rubocop.yml`, the paths in its Include parameters are interpreted as being relative to the file, or rather its directory. This is specific for file names starting with `.rubocop`. The problem is that file name matching that we do when searching for target files to inspect doesn't work for patterns that start with `..`. Solve the problem by matching absolute paths in these corner cases.

### DIFF
--- a/lib/rubocop/path_util.rb
+++ b/lib/rubocop/path_util.rb
@@ -44,7 +44,7 @@ module RuboCop
       end
     end
 
-    # rubocop:disable Metrics/MethodLength, Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
     def match_path?(pattern, path)
       case pattern
       when String
@@ -52,6 +52,10 @@ module RuboCop
           if pattern == path
             true
           elsif glob?(pattern)
+            # File name matching doesn't really work with relative patterns the start with "..". We
+            # get around that problem by converting the pattern to an absolute path.
+            pattern = File.expand_path(pattern) if pattern.start_with?('..')
+
             File.fnmatch?(pattern, path, File::FNM_PATHNAME | File::FNM_EXTGLOB)
           end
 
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

lib/rubocop/path_util.rb
```diff
@@ -44,14 +44,18 @@ def smart_path(path)
...
-    # rubocop:disable Metrics/MethodLength, Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
     def match_path?(pattern, path)
       case pattern
       when String
         matches =
           if pattern == path
             true
           elsif glob?(pattern)
+            # File name matching doesn't really work with relative patterns the start with "..". We
```
dduugg, line 55: Should be "…that start with"

jonas054: Thanks! I'll fix that.

PR #12709, opened 2024-02-25
[Fix #12646] Handle opposite styles for Layout/SpaceBeforeBlockBraces

### LINKED ISSUE #12646: auto-gen-config generates incorrect config for Layout/SpaceBeforeBlockBraces
When running `rubocop --auto-gen-config` with default settings on a large legacy project, the generated `.rubocop-todo.yml` does not correctly suppress all found offenses, so that a subsequent RuboCop inspection using the generated config does not pass.

The problem seems to be with the `Layout/SpaceBeforeBlockBraces` cop, for which the following unexpected config entry is generated:

```yaml
# Offense count: 210
# This cop supports safe autocorrection (--autocorrect).
# Configuration parameters: EnforcedStyle.
# SupportedStyles: space, no_space
# SupportedStylesForEmptyBraces: space, no_space
Layout/SpaceBeforeBlockBraces:
  EnforcedStyleForEmptyBraces: no_space
```

This generated config is particularly strange since our recommended (and _mostly_ followed) Ruby code style in fact mandates spaces before both empty and non-empty block braces, although we do have a few dozen Ruby code files (out of more than 2000) that currently violate this rule.

Manually editing `.rubocop-todo.yml` to replace `EnforcedStyleForEmptyBraces: no_space` with `Enabled: false` does make a subsequent `rubocop` inspection pass, but `rubocop --regenerate-todo` still undoes this manual change.

## RuboCop version

I am using RuboCop 1.60.1 on JRuby 9.4.5.0. Running `rubocop -V` prints:

```
1.60.1 (using Parser 3.3.0.4, rubocop-ast 1.30.0, running on jruby 3.1.4) [java]
```

jonas054:
The `Layout/SpaceBeforeBlockBraces` cop has two separate style parameters, `EnforcedStyle` and `EnforcedStyleForEmptyBraces`. The logic for generating `Enabled: false` when opposite styles are detected during `--auto-gen-config` was implemented for `EnforcedStyle`, but it was missing for `EnforcedStyleForEmptyBraces`.

### DIFF
--- a/changelog/fix_auto_gen_config_SpaceBeforeBlockBraces.md
+++ b/changelog/fix_auto_gen_config_SpaceBeforeBlockBraces.md
@@ -1,0 +1,1 @@
+* [#12646](https://github.com/rubocop/rubocop/issues/12646): Fix `--auto-gen-config` bug for `Layout/SpaceBeforeBlockBraces`. ([@jonas054][])
--- a/lib/rubocop/cop/layout/space_before_block_braces.rb
+++ b/lib/rubocop/cop/layout/space_before_block_braces.rb
@@ -80,10 +80,18 @@ module RuboCop
 
         private
 
+        # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
         def check_empty(left_brace, space_plus_brace, used_style)
-          return if style_for_empty_braces == used_style
-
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #12711, opened 2024-02-25
Handle implicit receivers in `Style/InvertibleUnlessCondition`

sambostock:
Previously, we would attempt to call `node.receiver.source`, but `node.receiver` could be `nil`.

```
An error occurred while Style/InvertibleUnlessCondition cop was inspecting /path/to/file.rb:123:4.
undefined method `source' for nil
```

This teaches the cop to handle implicit receivers.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/invertible_unless_condition.rb
+++ b/lib/rubocop/cop/style/invertible_unless_condition.rb
@@ -99,23 +99,23 @@ module RuboCop
           end
         end
 
-        def preferred_send_condition(node)
-          receiver_source = node.receiver.source
+        def preferred_send_condition(node) # rubocop:disable Metrics/CyclomaticComplexity
+          receiver_source = node.receiver&.source
           return receiver_source if node.method?(:!)
 
+          receive = receiver_source ? "#{receiver_source}." : '' # receiver may be implicit (self)
+
           inverse_method_name = inverse_methods[node.method_name]
-          return "#{receiver_source}.#{inverse_method_name}" unless node.arguments?
+          return "#{receive}#{inverse_method_name}" unless node.arguments?
 
           argument_list = node.arguments.map(&:source).join(', ')
           if node.operator_method?
             return "#{receiver_source} #{inverse_method_name} #{argument_list}"
           end
 
-          if node.parenthesized?
-            return "#{receiver_source}.#{inverse_method_name}(#{argument_list})"
-          end
+          return "#{receive}#{inverse_method_name}(#{argument_list})" if node.parenthesized?
 
-          "#{receiver_source}.#{inverse_method_name} #{argument_list}"
+          "#{receive}#{inverse_method_name} #{argument_list}"
         end
 
         def preferred_logical_condition(node)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/invertible_unless_condition.rb
```diff
@@ -99,23 +99,23 @@ def preferred_condition(node)
           end
         end
 
-        def preferred_send_condition(node)
-          receiver_source = node.receiver.source
+        def preferred_send_condition(node) # rubocop:disable Metrics/CyclomaticComplexity
```
sambostock, line 102: I didn't see a good way to avoid disabling this cop. It would require breaking apart this method, but I don't know that that would actually improve understandability.

Open to suggestions though.

lib/rubocop/cop/style/invertible_unless_condition.rb
```diff
@@ -99,23 +99,23 @@ def preferred_condition(node)
...
+          receiver_source = node.receiver&.source
           return receiver_source if node.method?(:!)
 
+          receive = receiver_source ? "#{receiver_source}." : '' # receiver may be implicit (self)
```
sambostock, line 106: I chose to do this instead of repeating the conditional

```ruby
if receiver_source
  # ...
else
  # ...
end
```

in three places, since it's always the same logic.

I also considered `receipt` as the variable name. Not sure.

spec/rubocop/cop/style/invertible_unless_condition_spec.rb
sambostock, line 1: I didn't add a case for an operator method, as the receiver would explicitly be `self`.

lib/rubocop/cop/style/invertible_unless_condition.rb
sambostock: Another option might be to explicitly fallback to `self` as the `receiver_source` and then leave all the rest the same, and rely on the redundant self correction to clean it up, but I don't know if that would actually fix the complexity offense, as it still introduces the conditional, even if it's inline.

lib/rubocop/cop/style/invertible_unless_condition.rb
bbatsov: Any reason why you didn't go with the name `receiver`?

bbatsov:
Might also be good to add an example in the cop's description with an implicit receiver.

lib/rubocop/cop/style/invertible_unless_condition.rb
sambostock: I wasn't sure if that would be confusing because we would then have `receiver_source` and `receiver`, but `receiver` would be `receiver_source + '.'` or `nil`, not the "receiver" node, as one might expect.

I suppose another option could be `dotted_receiver` or `dotted_receiver_source`.

sambostock: Yeah. I did switch it to `dotted_receiver_source` though, since that's clearer than `receive`.

sambostock:
I've switched to `dotted_receiver_source`, and added an implicit receiver example to the documentation. 👍

bbatsov (approved):

sambostock:
@bbatsov Sorry, it looks like I didn't actually push my commit doing the rename and adding the examples. 🤦 

I've made the changes in #12722.

PR #12730, opened 2024-03-01
Skip `LineLength` phase on `--auto-gen-only-exclude`

sambostock:
Phase 1 of config auto-generation runs the `Layout/LineLength` cop on all files to discover the `Max` line length, so it can be used by cops which depend on it.

The `--auto-gen-only-exclude` config skips generating any `Max` configs, meaning phase 1 is redundant, as we won't generate the `Max` for use by phase 2.

Therefore, we can skip phase 1 entirely. Even though we're only running `Layout/LineLength`, parsing the entire codebase twice can get quite time consuming on large repositories.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cli/auto_gen_config_spec.rb
+++ b/spec/rubocop/cli/auto_gen_config_spec.rb
@@ -1309,8 +1309,13 @@ RSpec.describe 'RuboCop::CLI --auto-gen-config', :isolated_environment do # rubo
       # absolute Exclude paths will point into this example's work directory.
       RuboCop::ConfigLoader.default_configuration = nil
 
+      $stdout = StringIO.new
       expect(cli.run(['--auto-gen-config', '--auto-gen-only-exclude',
                       '--exclude-limit', '1'])).to eq(0)
+      expect($stdout.string.include?(<<~STRING)).to be(true)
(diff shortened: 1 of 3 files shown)


spec/rubocop/cli/auto_gen_config_spec.rb
```diff
@@ -1309,8 +1309,10 @@ def function(arg1, arg2, arg3, arg4, arg5, arg6, arg7)
       # absolute Exclude paths will point into this example's work directory.
       RuboCop::ConfigLoader.default_configuration = nil
 
+      $stdout = StringIO.new
       expect(cli.run(['--auto-gen-config', '--auto-gen-only-exclude',
                       '--exclude-limit', '1'])).to eq(0)
+      expect($stdout.string.include?('run Layout/LineLength cop (skipped')).to be(true)
```
jonas054, line 1315: I think you should test the whole message about why `Layout/LineLength` is skipped. Don't cut it after `(skipped` because that could potentially be a different message.

jonas054 (changes requested):
Found one thing I would like changed. Other than that, looks good!

spec/rubocop/cli/auto_gen_config_spec.rb
sambostock: Good call. Done!

jonas054 (approved):
Nice! Approved.

Issue #12527, opened 2023-12-10
undefined method `name` while accessing to `RuboCop::Cop::Offense::highlighted_area`


snutij:
Hello 👋  I'm trying to access to the `highlighted_area` method of an offense, which is on the public api, but it's raising an error.

--------

## Expected behavior

Should return a `Parser::Source::Range` as documented here: 

https://github.com/rubocop/rubocop/blob/b1bcb3134d49a60061d08ef168ae996d70d8ce5c/lib/rubocop/cop/offense.rb#L138-L144

## Actual behavior

If I'm adding a debug in the `clang_formatter`:
https://github.com/rubocop/rubocop/blob/b1bcb3134d49a60061d08ef168ae996d70d8ce5c/lib/rubocop/formatter/clang_style_formatter.rb#L51-L56

Accessing to `highlighted_area` raise a `NoMethodError`

```ruby
(ruby) highlighted_area
#<NoMethodError: undefined method `name' for "  def dummy(node)":String> rescued during inspection
(ruby) highlighted_area.source_buffer
"  def dummy(node)"
(ruby) highlighted_area.begin_pos
2
(ruby) highlighted_area.source
"def dummy(node)"
``` 

## Steps to reproduce the problem

- Create a file `foo.rb`

```ruby
class Foo
  def dummy(node)
    puts "foo"
    puts "foo"
    puts "foo"
  end
end
``` 

- Enable the rule `Metrics/MethodLength` with `Max: 2`
- Add a breakpoint into `lib/rubocop/formatter/clang_style_formatter.rb`
- Run `bin/rubocop -d --only Metrics/MethodLength foo.rb`

## RuboCop version

```
bundle exec rubocop -V
1.58.0 (using Parser 3.2.2.4, rubocop-ast 1.30.0, running on ruby 3.2.2) [arm64-darwin23]
  - rubocop-minitest 0.33.0
  - rubocop-rake 0.6.0
  - rubocop-sorbet 0.7.5
```

snutij:
[...]

PR #12748, opened 2024-03-06
Attempt to fix offense.highlighted_area returning an incorrectly constructed range

Earlopain:
Attempt at #12527. This is not correct yet because tests are not passing.

To be honest, I'm not sure what's wrong here. Sometimes it starts one character to the left, adding 1 all the time is not the fix since it then starts one to the right in some cases. There's some offset fuckery going on that I'm not getting behind.

If someone can give insight into why that might be, that would be appreciated. I will add changleog/tests when I've got this figured out.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/offense.rb
+++ b/lib/rubocop/cop/offense.rb
@@ -140,7 +140,8 @@ module RuboCop
       # @return [Parser::Source::Range]
       #   the range of the code that is highlighted
       def highlighted_area
-        Parser::Source::Range.new(source_line, column, column + column_length)
+        range = location.source_buffer.line_range(line)
+        range.with(begin_pos: column, end_pos: column + column_length)
       end
 
       # @api private


Earlopain:
Sample test case `bundle exec rspec ./spec/rubocop/cli/autocorrect_spec.rb:1714`. I have added a bit of debugging output in the method:

```rb
def highlighted_area
  correct_range = Parser::Source::Range.new(source_line, column, column + column_length)
  range = location.source_buffer.line_range(line)
  offset = location.begin_pos - column
  range = range.with(begin_pos: column, end_pos: column + column_length)

  pp "loc.begin_pos: #{location.begin_pos}, loc.end_pos #{location.end_pos}, loc.column: #{location.column}, loc.column_length: #{column_length}"
  pp correct_range.begin_pos - range.begin_pos
  pp correct_range.end_pos - range.end_pos
  pp correct_range.source, range.source

  range
end
```

```
example.rb:1:12: C: [Corrected] Style/MethodDefParentheses: Use def with parentheses when there are parameters.
def primes limit
+"loc.begin_pos: 11, loc.end_pos 16, loc.column: 11, loc.column_length: 5"
+0
+0
+"limit"
+"limit"
           ^^^^^
example.rb:2:24: C: [Corrected] Style/SymbolProc: Pass &:even? as an argument to select instead of a block.
  1.upto(limit).select { |i| i.even? }
-                       ^^^^^^^^^^^^^^^
+"loc.begin_pos: 40, loc.end_pos 55, loc.column: 23, loc.column_length: 15"
+0
+0
+"{ |i| i.even? }"
+"to(limit).selec"
+                      ^^^^^^^^^^^^^^^
```

I can tweak the positions that the source is correct but then the positions are all wrong. Haven't figured out how to get both to line up yet.

bbatsov:
@koic Can you take a look here?

PR #12753, opened 2024-03-06
[Fix #12665] Fix infinite loop for `InternalAffairs/NodeMatcherDirective` with class methods

### LINKED ISSUE #12665: InternalAffairs/NodeMatcherDirective should support class methods
In https://github.com/rubocop/rubocop-rspec/pull/1793, we are using `def_node_matcher` to define class methods. On this bit of code

```rb
# @!method self.foo(node)
def_node_matcher 'self.foo', '(send ...)'
```

I get the following result:

```
[...] C: [Correctable] InternalAffairs/NodeMatcherDirective: @!method YARD directive has invalid method name, use self.foo instead of self.
def_node_matcher 'self.foo', '(send ...)'
```

Note the “self.foo instead of self.” part.

Running `rubocop --autocorrect` changes the YARD comment into this monstrosity:

```
# @!method self.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo.foo(node)
```

Actually, I think the correct YARD comment for a class method might be two lines:

```rb
# @!method foo(node)
# @!scope class
```

## RuboCop version

```
$ bundle exec rubocop -V
1.60.2 (using Parser 3.3.0.5, rubocop-ast 1.30.0, running on ruby 2.7.8) [arm64-darwin22]
  - rubocop-performance 1.20.2
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.26.1
```

Earlopain:
Closes #12665

Allow `self.` as a prefix for the method name.

I've added tests for both symbol and string arguments. All tests were for symbols but the original issues made use of strings, so just double-checking that that also works.

Since the regex would have gone over the line limit I've enabled free-spacing mode for it to have it span across multiple lines.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/internal_affairs/node_matcher_directive_spec.rb
+++ b/spec/rubocop/cop/internal_affairs/node_matcher_directive_spec.rb
@@ -222,5 +222,177 @@ RSpec.describe RuboCop::Cop::InternalAffairs::NodeMatcherDirective, :config do
         PATTERN
       RUBY
     end
+
+    it 'removes `@!scope class` YARD directive if it is not a class method' do
+      expect_offense(<<~RUBY, method: method)
+        # @!method foo?(node)
+        # @!scope class
+        #{method} :foo?, <<~PATTERN
+        ^{method}^^^^^^^^^^^^^^^^^^ Do not use the `@!scope class` YARD directive if it is not a class method.
+          (str)
+        PATTERN
+      RUBY
+
+      expect_correction(<<~RUBY)
+        # @!method foo?(node)
+        #{method} :foo?, <<~PATTERN
+          (str)
+        PATTERN
+      RUBY
+    end
+
+    it 'removes the receiver from the YARD directive if it is not a class method' do
+      expect_offense(<<~RUBY, method: method)
+        # @!method self.foo?(node)
+        #{method} :foo?, <<~PATTERN
+        ^{method}^^^^^^^^^^^^^^^^^^ `@!method` YARD directive has invalid method name, use `foo?` instead of `self.foo?`.
+          (str)
+        PATTERN
+      RUBY
+
+      expect_correction(<<~RUBY)
+        # @!method foo?(node)
+        #{method} :foo?, <<~PATTERN
+          (str)
+        PATTERN
+      RUBY
+    end
+
+    it 'removes the receiver from the YARD directive and the scope directive if it is not a class method' do
+      expect_offense(<<~RUBY, method: method)
+        # @!method self.foo?(node)
+        # @!scope class
+        #{method} :foo?, <<~PATTERN
+        ^{method}^^^^^^^^^^^^^^^^^^ Do not use the `@!scope class` YARD directive if it is not a class method.
+          (str)
+        PATTERN
+      RUBY
+
+      expect_correction(<<~RUBY)
+        # @!method foo?(node)
+        #{method} :foo?, <<~PATTERN
+          (str)
+        PATTERN
+      RUBY
+    end
+
+    it 'registers an offense when the directive has the wrong name without self' do
+      expect_offense(<<~RUBY, method: method)
+        # @!method self.bar?(node)
+        #{method} :foo?, <<~PATTERN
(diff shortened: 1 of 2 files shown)


spec/rubocop/cop/internal_affairs/node_matcher_directive_spec.rb
```diff
@@ -222,5 +222,31 @@ class MyCop
...
+        PATTERN
+
+        #{method} "self.bar?", <<~PATTERN
+        ^{method}^^^^^^^^^^^^^^^^^^^^^^^^ Precede `#{method}` with a `@!method` YARD directive.
+          (str)
+        PATTERN
+      RUBY
+
+      expect_correction(<<~RUBY)
+        # @!method self.foo?(node)
```
koic, line 240: Isn't the expected behavior described in #12665 as follows?

```suggestion
        # @!method foo?(node)
        # @!scope class
```

It also seems necessary to implement the detection and autocorrection in the opposite direction:

```ruby
# @!method self.foo?(node)
# @!scope class
#{method} :"foo?", <<~PATTERN
  (str)
PATTERN
```

Earlopain: You're right, I missed that part. I just looked at the PR in that repo. I changed it to correct to this syntax instead.

If there are variations I've missed to test against let me know.

koic:
Writing a specific value instead of `self` will lead to `InfiniteCorrectionLoop` error:

```ruby
def_node_matcher :'x.foo', <<~PATTERN
  ...
PATTERN
```

Earlopain:
Good point. I fixed this by just ignoring any matchers that don't use `self`. Handling this case would have added a bunch more complexity, hope that's ok.

bquorning:
Thank you @Earlopain. If I’m allowed to add an edge case; this spec example is currently failing (add it inside the 'when using class methods' context):
```
      it 'gives the correct message for "self." prefix on @!method when @!scope is correctly "class"' do
        expect_offense(<<~RUBY, method: method)
          # @!method self.foo?(node)
          # @!scope class
          #{method} 'self.foo?', <<~PATTERN
          ^{method}^^^^^^^^^^^^^^^^^^^^^^^^ `@!method` YARD directive has invalid method name, use `foo?` instead of `self.foo?`.
            (str)
          PATTERN
        RUBY

        expect_correction(<<~RUBY)
          # @!method foo?(node)
          # @!scope class
          #{method} 'self.foo?', <<~PATTERN
            (str)
          PATTERN
        RUBY
      end
```
with the offense message reading

> `@!method` YARD directive has invalid method name, use `self.foo?` instead of `self.foo?`.

PR #12732, opened 2024-03-02
Call .value to get gemspec required_ruby_version only if str_type?

davidrunger:
fixes #12744

This change avoids an exception (NoMethodError: undefined method \`value' for an instance of RuboCop::AST::SendNode) that currently occurs in `rubocop` version 1.61.0 when the project has a gemspec file and that gemspec specifies a `required_ruby_version` by reading from another file (such as `.ruby-version`).

**Example of an affected gemspec file:**

```rb
Gem::Specification.new do |spec|
  # ...
  spec.required_ruby_version = Gem::Requirement.new(">= #{File.read('.ruby-version').rstrip}")
  # ...
end
```

This bug began occurring in several of my projects as a result of #12645 (d1746bedd) ("Check gemspec `required_ruby_version` before `.ruby-version` and other sources"). My projects have both a `.ruby-version` file and also a gemspec that (as in the example above) sets the `required_ruby_version` by reading from that `.ruby-version` file. Prior to #12645, the latent bug that this change fixes was not occurring for me, because the GemspecFile TargetRuby finder was never running, because the RubyVersionFile TargetRuby finder had higher precedence (and a target Ruby version was found using this finder), and so the GemspecFile TargetRubyVersion finder was never invoked.

With #12645, because the GemspecFile finder now runs before the RubyVersionFile finder, this exception has started occurring in these projects, when I try to run rubocop 1.61.0.

This change fixes/avoids this exception by checking that the `right_hand_side` is a `str_type?` before attempting to call `right_hand_side.value`. In a setup like mentioned in my projects above, `right_hand_side` will be a `RuboCop::AST::SendNode` and `str_type?` will be `false`, so we avoid the NoMethodError exception that otherwise will occur if we attempt to call `.value` on that node.

With this change, for projects with such a gemspec file that reads from a `.ruby-version` file, the GemspecFile finder will now return `nil` for the target Ruby version (rather than raising an unhandled exception), and so we will continue through the other TargetRuby finder classes, until the RubyVersionFile finder returns a Ruby version based on the one specified in the `.ruby-version` file.

[...]

PR #12645, opened 2024-01-23
Check gemspec `required_ruby_version` before `.ruby-version` and other sources

jenshenny:
Currently, if I have a gem that has both a `.ruby-version` file and a `required_ruby_version` in the gemspec, Rubocop would default to the ruby version in the `.ruby-version` file.

However, it would be more correct to take the ruby version in the gemspec as the `TargetRubyVersion` [should be the oldest version of Ruby which your project supports with](https://docs.rubocop.org/rubocop/configuration.html#setting-the-target-ruby-version). `required_ruby_version` would specify any minimum Ruby version. The `.ruby-version` and other sources would get the development version in the project and might not be the oldest supported version. 

### Changes
Moved the GemspecFile source to be under the rubocop config.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

Issue #12579, opened 2023-12-28
Gem::Requirement.new in required_ruby_version raises BadRequirementError when version is read from elsewhere


CamJN:
## Expected behavior

Rubocop should evaluate the gemspec file rather than try to parse the unevaluated Ruby.

## Actual behavior

```
rake aborted!
Gem::Requirement::BadRequirementError: Illformed requirement [">= \#{ruby_version}.0"] (Gem::Requirement::BadRequirementError)

      raise BadRequirementError, "Illformed requirement [#{obj.inspect}]"
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Users/runner/work/getargv.rb/getargv.rb/getargv_ruby/vendor/bundle/ruby/3.3.0/gems/rubocop-1.59.0/lib/rubocop/target_ruby.rb:216:in `new'
/Users/runner/work/getargv.rb/getargv.rb/getargv_ruby/vendor/bundle/ruby/3.3.0/gems/rubocop-1.59.0/lib/rubocop/target_ruby.rb:216:in `find_default_minimal_known_ruby'
/Users/runner/work/getargv.rb/getargv.rb/getargv_ruby/vendor/bundle/ruby/3.3.0/gems/rubocop-1.59.0/lib/rubocop/target_ruby.rb:178:in `find_version'
/Users/runner/work/getargv.rb/getargv.rb/getargv_ruby/vendor/bundle/ruby/3.3.0/gems/rubocop-1.59.0/lib/rubocop/target_ruby.rb:29:in `initialize'
/Users/runner/work/getargv.rb/getargv.rb/getargv_ruby/vendor/bundle/ruby/3.3.0/gems/rubocop-1.59.0/lib/rubocop/target_ruby.rb:257:in `new'
/Users/runner/work/getargv.rb/getargv.rb/getargv_ruby/vendor/bundle/ruby/3.3.0/gems/rubocop-1.59.0/lib/rubocop/target_ruby.rb:257:in `block in source'
/Users/runner/work/getargv.rb/getargv.rb/getargv_ruby/vendor/bundle/ruby/3.3.0/gems/rubocop-1.59.0/lib/rubocop/target_ruby.rb:257:in `each'
[...]

PR #12756, opened 2024-03-07
Only parse target Ruby from gemspec if array elements are strings

davidrunger:
This fixes an issue mentioned in [this comment](https://github.com/rubocop/rubocop/issues/12579#issuecomment-1974916930) on issue #12579. (I believe that the main/original issue mentioned in issue #12579 has essentially already been fixed by #12732.)

This change is a related/similar followup to #12732 (50263938f). That PR avoided an exception that was occurring in the
`RuboCop::TargetRuby::GemspecFile` target Ruby finder when a project's gemspec determines the value of `required_ruby_version` in some dynamic way (such as by reading from a `.ruby-version` file).

However, #12732 did not handle a case where the dynamic value for the gemspec's `required_ruby_version` is wrapped within an array, and, in such a case, currently, an exception like the following will occur:

```
Gem::Requirement::BadRequirementError:
  Illformed requirement [">= \#{File.read('.ruby-version').rstrip}"]
  ./lib/rubocop/target_ruby.rb:123:in `new'
  ./lib/rubocop/target_ruby.rb:123:in `find_default_minimal_known_ruby'
  ./lib/rubocop/target_ruby.rb:83:in `find_version'
  ./lib/rubocop/target_ruby.rb:29:in `initialize'
  ./lib/rubocop/target_ruby.rb:259:in `new'
```

(As with the scenario that _was_ fixed by #12732, I believe that this is essentially a "latent" issue that is now occurring in more projects than previously, due to #12645 having increased the precedence of the `GemspecFile` target Ruby finder in the `RuboCop::TargetRuby::SOURCES` list.)

This change avoids this exception by only attempting to parse a target Ruby version from an array value for a gemspec's `required_ruby_version` if every value in the array is a string. When the values within the `required_ruby_version` array are _not_ all strings, now, rather than raising an unhandled exception, the `GemspecFile` finder will return `nil` for the target Ruby version, and so we will continue on to the other TargetRuby finder classes, in order to find a target Ruby version.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/target_ruby_spec.rb
+++ b/spec/rubocop/target_ruby_spec.rb
@@ -188,10 +188,25 @@ RSpec.describe RuboCop::TargetRuby, :isolated_environment do
 
       context 'when file reads the required_ruby_version from another file' do
         it 'uses the default target ruby version' do
-          content = <<~HEREDOC
+          content = <<~'HEREDOC'
             Gem::Specification.new do |s|
               s.name = 'test'
-              s.required_ruby_version = Gem::Requirement.new(">= \#{File.read('.ruby-version').rstrip}")
+              s.required_ruby_version = Gem::Requirement.new(">= #{File.read('.ruby-version').rstrip}")
+              s.licenses = ['MIT']
+            end
+          HEREDOC
+
+          create_file(gemspec_file_path, content)
+          expect(target_ruby.version).to eq default_version
+        end
+      end
+
+      context 'when file reads the required_ruby_version from another file in an array' do
+        it 'uses the default target ruby version' do
+          content = <<~'HEREDOC'
+            Gem::Specification.new do |s|
+              s.name = 'test'
+              s.required_ruby_version = [">= #{File.read('.ruby-version').rstrip}"]
               s.licenses = ['MIT']
             end
           HEREDOC
--- a/changelog/fix_only_parse_target_ruby_from_gemspec_if_array.md
+++ b/changelog/fix_only_parse_target_ruby_from_gemspec_if_array.md
@@ -1,0 +1,1 @@
+* [#12756](https://github.com/rubocop/rubocop/pull/12756): Only parse target Ruby from gemspec if array elements are strings. ([@davidrunger][])
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -647,11 +647,33 @@ AllCops:
 
 NOTE: When `ParserEngine: parser_prism` is specified, the values that can be assigned to `TargetRubyVersion` must be `3.3` or higher.
 
-Otherwise, RuboCop will then check your project for a series of files where
-the version may be specified already. The files that will be looked for are
+If a `TargetRubyVersion` is not specified in your config, then RuboCop will
+check your project for a series of other files where the Ruby version may be
+specified already. The files that will be checked are (in this order):
 `*.gemspec`, `.ruby-version`, `.tool-versions`, and `Gemfile.lock`.
-If Gemspec file has an array for `required_ruby_version`, the lowest version will be used.
-If none of the files are found a default version value will be used.
+
+If a target Ruby version cannot be found via any of the above sources, then a
+default target Ruby version will be used.
+
+=== Finding target Ruby in a `*.gemspec` file
+
+In order for RuboCop to parse a `*.gemspec` file's `required_ruby_version`, the
+Ruby version must be specified using one of these syntaxes:
+
+1. a string range, e.g. `'~> 3.2.0'` or `'>= 3.2.2'`
+2. an array of strings, e.g. `['>= 3.0.0', '< 3.4.0']`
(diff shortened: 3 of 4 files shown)


davidrunger:
cc @Earlopain (Thank you for connecting the dots [here](https://github.com/rubocop/rubocop/issues/12579#issuecomment-1981500920) between that issue #12579 and my recent PR #12732 !)

spec/rubocop/target_ruby_spec.rb
```diff
@@ -201,6 +201,21 @@
...
 
+      context 'when file reads the required_ruby_version from another file in an array' do
+        it 'uses the default target ruby version' do
+          content = <<~HEREDOC
+            Gem::Specification.new do |s|
+              s.name = 'test'
+              s.required_ruby_version = [">= \#{File.read('.ruby-version').rstrip}"]
+              s.licenses = ['MIT']
+            end
+          HEREDOC
```
koic, line 212:
```suggestion
          content = <<~'HEREDOC'
            Gem::Specification.new do |s|
              s.name = 'test'
              s.required_ruby_version = [">= #{File.read('.ruby-version').rstrip}"]
              s.licenses = ['MIT']
            end
          HEREDOC
```

davidrunger:
cc @capripot ; I believe that this will fix the issue that you mentioned [here](https://github.com/rubocop/rubocop/issues/12579#issuecomment-1974916930).

bbatsov:
I'd also suggest expanding the docs here a bit https://www.gnu.org/software/emacs/manual/html_node/elisp/Declaring-Functions.html (the section about the target version discovery is pretty short, so it might not be clear to people how this is expected to work exactly)

spec/rubocop/target_ruby_spec.rb
davidrunger: @koic : Thank you for the review and the good suggestion! I have applied your suggestion both to the spec that I am adding in this PR (as you suggested) and also to the similar spec just above, which I added in #12732. (Unfortunately, this makes the diff in GitHub look a little bit more complicated than necessary.)

spec/rubocop/target_ruby_spec.rb
```diff
@@ -188,10 +188,25 @@
 
       context 'when file reads the required_ruby_version from another file' do
         it 'uses the default target ruby version' do
```
davidrunger, line 190: (This is an existing spec that I only modified to have a heredoc without interpolation; it is otherwise unchanged.)

spec/rubocop/target_ruby_spec.rb
```diff
@@ -188,10 +188,25 @@
...
+            end
+          HEREDOC
+
+          create_file(gemspec_file_path, content)
+          expect(target_ruby.version).to eq default_version
+        end
+      end
+
+      context 'when file reads the required_ruby_version from another file in an array' do
+        it 'uses the default target ruby version' do
```
davidrunger, line 205: This is the new spec that I am adding for this change. Unfortunately (due to how it looks similar to the spec changed above), GitHub is actually not even including the whole new spec in its rendering of this diff. You'll have to click the down arrow to see the last few lines of this new spec, which GitHub is rendering as if they are preexisting lines.

bbatsov:
I had pasted the wrong link (updated) - the right one is https://docs.rubocop.org/rubocop/1.62/configuration.html#setting-the-target-ruby-version Basically I'm referring to:

> Otherwise, RuboCop will then check your project for a series of files where the version may be specified already. The files that will be looked for are *.gemspec, .ruby-version, .tool-versions, and Gemfile.lock. If Gemspec file has an array for required_ruby_version, the lowest version will be used. If none of the files are found a default version value will be used.

Would be good to clarify when can RuboCop actually extract a version from the gemspec or the Gemfile.lock. We can also clarify which is the default value that will be used if inference fails.

davidrunger:
> I had pasted the wrong link (updated) - the right one is https://docs.rubocop.org/rubocop/1.62/configuration.html#setting-the-target-ruby-version Basically I'm referring to:
> 
> > Otherwise, RuboCop will then check your project for a series of files where the version may be specified already. The files that will be looked for are *.gemspec, .ruby-version, .tool-versions, and Gemfile.lock. If Gemspec file has an array for required_ruby_version, the lowest version will be used. If none of the files are found a default version value will be used.
> 
> Would be good to clarify when can RuboCop actually extract a version from the gemspec or the Gemfile.lock. We can also clarify which is the default value that will be used if inference fails.

@bbatsov : I made an effort to expand the documentation a bit in 64ca11bd360fa2084232e0b2dd985c0580c5fcfa (before reading your latest comment). I have a tendency to be too verbose, and I'm not sure I touched on the key points in the best way. Feel free to suggest any additions, deletions, or changes, or to simply make those changes on this branch yourself, if that's easier/preferable. (I believe that I've granted you permission to do so.)

davidrunger:
> We can also clarify which is the default value that will be used if inference fails.

@bbatsov : I had thought about adding that, but I didn't want to put that as a maintenance burden on you or to have that be something that we might forget to update and which would be prone to maybe becoming outdated and inaccurate. I can add that, if you'd like me to, though (or, again, feel free to add it yourself, if that's easier).

PR #12760, opened 2024-03-07
[Fix #12663] Prevent `Lint/Syntax` being disabled by directive comments

### LINKED ISSUE #12663: `Lint/Syntax` is able to be disabled by directive comments
Consider this very simple example file:

```rb
foo*
```

This isn't valid syntax and RuboCop complains. If I change this snippet to the following no offense is detected anymore:

```rb
# rubocop:disable Lint/Syntax
foo*
```

My understanding is that `Lint/Syntax` should not be able to be disabled under any circumstance. https://docs.rubocop.org/rubocop/1.60/configuration.html
> The cop enabling process can be altered by setting DisabledByDefault or EnabledByDefault (but not both) to true. These settings override the default for all cops to disabled or enabled, except Lint/Syntax which is always enabled, regardless of the cops' default values (whether enabled, disabled or pending).

> There is one exception from the general rule above and that is Lint/Syntax, a special cop that checks for syntax errors before the other cops are invoked. It cannot be disabled and its severity (fatal) cannot be changed in configuration.

If this is considered a bug then `RedundantCopDisableDirective` and `RedundantCopEnableDirective` should probably detect this as an offense. (actually, probably not? It's not valid syntax after all, don't think that would work)

## RuboCop version

```
$ [bundle exec] rubocop -V
1.60.2 (using Parser 3.3.0.5, rubocop-ast 1.30.0, running on ruby 3.2.2) [x86_64-linux]
  - rubocop-performance 1.20.2
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.26.1
```

Earlopain:
Fixes #12663

This would swallow the syntax error, making it look like the file is all good.
In reality all other cops are not being run, even if `Lint/Syntax` is "enabled" again later on.

This code for example would show as having no offenses.
```rb
# rubocop:disable Lint/Syntax
1 /// 2
# rubocop:enable Lint/Syntax

a = ""
b = ''
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_lint_syntax_being_disabled_by_directive_comments.md
+++ b/changelog/fix_lint_syntax_being_disabled_by_directive_comments.md
@@ -1,0 +1,1 @@
+* [#12663](https://github.com/rubocop/rubocop/issues/12663): Fix `Lint/Syntax` getting disabled by `rubocop:disable Lint/Syntax`. ([@earlopain][])
--- a/lib/rubocop/directive_comment.rb
+++ b/lib/rubocop/directive_comment.rb
@@ -6,9 +6,11 @@ module RuboCop
   # cops it contains.
   class DirectiveComment
     # @api private
-    REDUNDANT_DIRECTIVE_COP_DEPARTMENT = 'Lint'
+    LINT_DEPARTMENT = 'Lint'
     # @api private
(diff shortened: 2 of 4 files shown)


jonas054 (approved):
Looks good!

koic:
Thanks!

PR #11768, opened 2023-04-06
[Fix #11765] Fix an error for `Style/MultilineMethodSignature`

koic:
Fixes #11765.

This PR fixes an error for `Style/MultilineMethodSignature` when line break after `def` keyword.

It seems preferable that the following cases can be detected by a new different cop:

```ruby
def
method_name arg; end
```

This is probably not a common line break, the new cop will be able to handle `class`, `module`, class method definitions, and others as well.
And this `Style/MultilineMethodSignature` cop is disabled by default, but I have a feeling that the new cop can be enabled by default.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [ ] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #12763, opened 2024-03-09
Fix an infinite loop for `Style/MultilineMethodSignature` when line break after def

Earlopain:
This undoes a part of #8090, specifically at https://github.com/rubocop/rubocop/pull/8090/commits/448eb9662d2e5067b65c1e8de9b4dd8019cc9e0c. There don't seem to be any consequences of doing so except for uncovering this bug.

A test case for this issue has already been added in #11768 but after the result has been reached there are still correctors that result in an infinite loop. In practice some other cop comes in first and makes this situation unlikely to actually encounter, unless it is disabled of course.

While I was working on #12753 I got into a similar situation, where my assertions where passing but actually running it on the codebase resulted in an infinite loop. 

This PR fixes the ininite loop and makes these situations impossible by only considering a correction done if there are no correctors left.

-----------------

* [x] Added tests.


bbatsov:
Looks good to me. Thanks!

PR #12766, opened 2024-03-09
[Fix #12721] Add `DebuggerRequires` to `Lint/Debugger`

### LINKED ISSUE #12721: Lint/Debugger does not catch require "debug/start"
There is a [feature](https://github.com/ruby/debug?tab=readme-ov-file#start-by-requiring-a-library) in the [debug gem](https://rubygems.org/gems/debug) where you can require the gem and enter the debugger in a single statement. I don't think other debuggers have this so it is being missed by the `Lint/Debugger` cop.

The feature allows the following statements to start a debugger:
```rb
require 'debug/start' # Same as rdbg.
require 'debug/open' # Same as rdbg --open.
require 'debug/open_nonstop' # Same as rdbg --open --nonstop.
```

I don't think we want any of these to accidentally make it to production.

## Expected behavior

`Lint/Debugger` should complain about this code, since it is a debugger breakpoint the same way as putting `debugger` in your code is:

```rb
def a_method
  x = 2 + 2

  require "debug/start"

  x * 2
end
```

## Actual behavior

`1 file inspected, no offenses detected`

Put `require "debug/start"` in your code. `$ rubocop `

## Attempted workaround

This has no effect:

```yml
Lint/Debugger:
  DebuggerMethods:
    requires:
      "require \"debug/start\""
```

I would assume this cop configuration does not support checking for the specific argument to require. Maybe I'm missing something?

## RuboCop version

```
$ bundle exec rubocop -V
1.56.3 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.2.2) [x86_64-linux]
  - rubocop-performance 1.19.0
  - rubocop-rspec 2.24.0
```

and

```
$ bundle exec rubocop -V
1.56.3 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.2.1) [x86_64-linux]
  - rubocop-capybara 2.18.0
  - rubocop-factory_bot 2.23.1
  - rubocop-performance 1.19.0
  - rubocop-rails 2.21.0
  - rubocop-rspec 2.24.0
```

Earlopain:
I debated adding requires like `debug` or `pry` but believe that would do more harm than good.
RuboCops own spec_helper requires pry. This seems to come from some sort of template.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1657,7 +1657,7 @@ Lint/Debugger:
   Description: 'Check for debugger calls.'
   Enabled: true
   VersionAdded: '0.14'
-  VersionChanged: '1.46'
+  VersionChanged: '<<next>>'
   DebuggerMethods:
     # Groups are available so that a specific group can be disabled in
     # a user's configuration, but are otherwise not significant.
@@ -1693,6 +1693,11 @@ Lint/Debugger:
       - jard
     WebConsole:
       - binding.console
+  DebuggerRequires:
+    debug.rb:
+      - debug/open
+      - debug/open_nonstop
+      - debug/start
 
 Lint/DeprecatedClassMethods:
   Description: 'Check for deprecated class method calls.'
--- a/lib/rubocop/cop/lint/debugger.rb
+++ b/lib/rubocop/cop/lint/debugger.rb
@@ -29,6 +29,11 @@ module RuboCop
       #       MyDebugger.debug_this
       # ----
       #
+      # Some gems also ship files that will start a debugging session when required,
+      # for example `require 'debug/start'` from `ruby/debug`. These requires can
+      # be configured through `DebuggerRequires`. It has the same structure as
+      # `DebuggerMethods`, which you can read about above.
+      #
       # @example
       #
       #   # bad (ok during development)
@@ -64,14 +69,20 @@ module RuboCop
       #   def some_method
       #     my_debugger
       #   end
+      #
+      # @example DebuggerRequires: [my_debugger/start]
+      #
+      #   # bad (ok during development)
+      #
+      #   require 'my_debugger/start'
       class Debugger < Base
         MSG = 'Remove debugger entry point `%<source>s`.'
         BLOCK_TYPES = %i[block numblock kwbegin].freeze
 
         def on_send(node)
-          return if !debugger_method?(node) || assumed_usage_context?(node)
+          return if assumed_usage_context?(node)
 
-          add_offense(node)
+          add_offense(node) if debugger_method?(node) || debugger_require?(node)
         end
 
         private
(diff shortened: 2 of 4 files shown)


bbatsov:
> RuboCops own spec_helper requires pry. This seems to come from some sort of template.

I'm looking at that and even I can't remember why we've done it. :D I see it was added with the following commit message:

>  Allow `pry` to be activated in rspec if in the local gemfile.

Perhaps it was just for debugging purposes? @dvandersluis probably remembers better.

Earlopain:
I did a github code search ([link here](https://github.com/search?q=%22require+%27pry%27%22+language%3ARuby&type=code&l=Ruby&p=1)) and its telling me there are just about 60k results for that. You can't actually go past page 5 but the majority seem to come from spec helpers as well.

bbatsov:
Got it. I we shouldn't bother with this indeed.

dvandersluis:
> Perhaps it was just for debugging purposes? @dvandersluis probably remembers better.

I use pry extensively when debugging in rspec, so that was the intention here. I have pry in my `Gemfile.local` file so it's not actually a dependency, but if it's present it'll be activated. I don't think there's necessarily a cleaner way to add it to the spec_helper.

config/default.yml
```diff
@@ -1693,6 +1693,11 @@ Lint/Debugger:
       - jard
     WebConsole:
       - binding.console
+  DebuggerRequires:
+    debug.rb:
+      - debug/start
+      - debug/open
+      - debug/open_nonstop
```
koic, line 1700:
```suggestion
      - debug/open
      - debug/open_nonstop
      - debug/start
```

lib/rubocop/cop/lint/debugger.rb
```diff
@@ -87,12 +98,26 @@ def debugger_methods
...
+        end
+
         def debugger_method?(send_node)
           return false if send_node.parent&.send_type? && send_node.parent.receiver == send_node
 
           debugger_methods.include?(chained_method_name(send_node))
         end
 
+        def debugger_require?(send_node)
+          return false unless send_node.method?(:require) && send_node.arguments.count == 1
```
koic, line 115:
```suggestion
          return false unless send_node.method?(:require) && send_node.arguments.one?
```

koic:
FYI, the reason why `pry` is required in this repository's `spec_helper` may be explained by the following link:
https://github.com/rubocop/rubocop/pull/9341

mvz:
FWIW, I used to just put the following code wherever I want to start a pry session (using mostly IRB now):

```ruby
require "pry"
binding.pry
```

Are there uses cases for pry where this doesn't work?

Earlopain:
I don't understand what you mean with that, can you elaborate?

mvz:
@Earlopain I mean it is not necessary to permanently have `require "pry"` in your spec helper in order to use pry when debugging specs. Whenever you need to start a pry session, instead of just typing

```ruby
binding.pry
```

you can type the following two lines right where you want to start the session:

```ruby
require "pry"
binding.pry
```

I have a shortcut in my editor for that to save typing.

koic:
> For now, users can add these to their config themselves

I agree with @Earlopain. Detecting `require 'pry'` could lead to some false positives. It would be better for users who need it to configure it themselves. In other words, it should not be provided by default, I think.

PR #12767, opened 2024-03-10
[Fix #12731] Add cop `Lint/PossibleAssignmentToNil`

### LINKED ISSUE #12731: `AssignmentInCondition` didn't find the style problem
## Expected behavior

RuboCop should find the style problem in the `if` statement

## Actual behavior

RuboCop didn't find the style problem

test.rb

```rb
a = 1

if a&.to_i = 1
  puts '111'
end
```

Run:

```bash
rubocop --only Lint/AssignmentInCondition -a test.rb
```

Result:

```
Inspecting 1 file
.

```

## RuboCop version

```bash
root@localhost:~$ rubocop -v
1.61.0
```

jonas054:
The cop detects assignments, or actually calls to setter methods, where the receiver of the assignment can be `nil`, which would result in an exception.

Having expressions containing `&.` to the left of the `=` never makes sense, as far as I can see. I've tried to illustrate, in the cop documentation examples, a couple of situations where it's possible to make this mistake.

One is a conditional assignment `a&.value ||= 0` where the author might have thought that they want to set `value` to `0` if `a` is `nil`. That's not the way to do it, though.

The other is assignment in an `if` statement, `do_something if a&.value = 0`. Here the author just forgot to use the double equals for comparison. Normally we'd expect `Lint/AssignmentInCondition` to handle this, but since that cop does auto-correction, I think it's better to let this new cop handle this particular case. It requires some analysis by the author, and there's really no auto-correction that can be made.


Earlopain:
> The cop detects assignments, or actually calls to setter methods, where the receiver of the assignment can be `nil`, which would result in an exception.

I don't think I understand what you mean there, the following code will pring `Saving foo!` and not error (on Ruby 3.3). Or is there a different case you mean with this?

```rb
class Test
  def setter=(value)
    @value = value
  end

  def save!
    puts "Saving #{@value}!"
  end
end

not_nil = Test.new
is_nil = nil

not_nil&.setter = "foo"
not_nil&.save!

is_nil&.setter = "bar"
is_nil&.save!
```

jonas054:
@Earlopain Oh! You're right. There is indeed a legitimate case when we want to use `&.` for setting something only if the object is not `nil`.

Back to the drawing board!

PR #12768, opened 2024-03-10
Fix a false positive for `Style/NilComparison` without receiver and `EnforcedStyle: comparison`

Earlopain:
This would previously show an offense and get into an infinite loop during correction.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/nil_comparison.rb
+++ b/lib/rubocop/cop/style/nil_comparison.rb
@@ -44,6 +44,8 @@ module RuboCop
         def_node_matcher :nil_check?, '(send _ :nil?)'
 
         def on_send(node)
+          return unless node.receiver
+
           style_check?(node) do
             add_offense(node.loc.selector) do |corrector|
               new_code = if prefer_comparison?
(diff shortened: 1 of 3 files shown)


Earlopain:
This is broken on prism 0.24. I'm assuming this is the same as all the other tests that have this comment. I confirmed that the test passes against prism main.

lib/rubocop/cop/style/nil_comparison.rb
```diff
@@ -41,7 +41,7 @@ class NilComparison < Base
         def_node_matcher :nil_comparison?, '(send _ {:== :===} nil)'
 
         # @!method nil_check?(node)
-        def_node_matcher :nil_check?, '(send _ :nil?)'
+        def_node_matcher :nil_check?, '(send ^_ :nil?)'
 
         def on_send(node)
           style_check?(node) do
```
koic, line 47: I think it might be clearer to write `return unless node.receiver` at the beginning of `def on_send` rather than using `^`.

```suggestion
        def_node_matcher :nil_check?, '(send _ :nil?)'

        def on_send(node)
          return unless node.receiver

          style_check?(node) do
```

Earlopain: 👍

PR #12770, opened 2024-03-10
Fix an error for `Lint/RedundantWithIndex` when there is no receiver

Earlopain:
Did not validate that there even is a receiver.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/redundant_with_index.rb
+++ b/lib/rubocop/cop/lint/redundant_with_index.rb
@@ -34,6 +34,7 @@ module RuboCop
         MSG_WITH_INDEX = 'Remove redundant `with_index`.'
 
         def on_block(node)
+          return unless node.receiver
           return unless (send = redundant_with_index?(node))
 
           range = with_index_range(send)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/redundant_with_index.rb
```diff
@@ -56,10 +56,10 @@ def on_block(node)
         def_node_matcher :redundant_with_index?, <<~PATTERN
           {
             (block
-              $(call _ {:each_with_index :with_index} ...)
+              $(call ^_ {:each_with_index :with_index} ...)
               (args (arg _)) ...)
             (numblock
-              $(call _ {:each_with_index :with_index} ...) 1 ...)
+              $(call ^_ {:each_with_index :with_index} ...) 1 ...)
```
koic, line 62: I think it might be clearer to write `return unless node.receiver` at the beginning of `def on_block` rather than using `^`.
Ref: https://github.com/rubocop/rubocop/pull/12768#discussion_r1518852124

PR #12774, opened 2024-03-10
Fix an infinite loop for `Style/RaiseArgs` with `EnforcedStyle: compact` when passing more than 2 arguments to `raise`

Earlopain:
The following test causes the loop: https://github.com/rubocop/rubocop/blob/5ee786dcadfa3c91096d9e81dcd16e4776ca6186/spec/rubocop/cop/style/raise_args_spec.rb#L130-L140

This changes `expect_no_corrections`  to always raise if there are correctors. In this case the corrector replaced nothing, resulting in the same cop being invocted again for the same reason.

Undoing the cop change while keeping the assertion change will produce a failure. The source assertion is left in because it produces nice diffs if there are actually changes.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/raise_args.rb
+++ b/lib/rubocop/cop/style/raise_args.rb
@@ -80,7 +80,7 @@ module RuboCop
 
         def correction_exploded_to_compact(node)
           exception_node, *message_nodes = *node.arguments
-          return node.source if message_nodes.size > 1
+          return if message_nodes.size > 1
 
           argument = message_nodes.first.source
           exception_class = exception_node.receiver&.source || exception_node.source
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/raise_args.rb
```diff
@@ -98,9 +98,10 @@ def check_compact(node)
             return if exception.send_type? && exception.first_argument&.hash_type?
 
             add_offense(node, message: format(COMPACT_MSG, method: node.method_name)) do |corrector|
-              replacement = correction_exploded_to_compact(node)
+              if (replacement = correction_exploded_to_compact(node))
+                corrector.replace(node, replacement)
+              end
 
-              corrector.replace(node, replacement)
```
koic, line 103: It seems that the tests pass without the changes in L101-L103. If this change is necessary, can you add a test for it?

Earlopain: That is indeed not needed, I was just defensive about passing nil to the corrector but that seems to work just fine. Thank you.

PR #12775, opened 2024-03-11
Fix an error for `Lint/UselessTimes` when no block is present

Earlopain:
Another thing that's more common with an LSP.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/useless_times_spec.rb
+++ b/spec/rubocop/cop/lint/useless_times_spec.rb
@@ -292,6 +292,15 @@ RSpec.describe RuboCop::Cop::Lint::UselessTimes, :config do
 
         expect_no_corrections
       end
+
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/lint/useless_times_spec.rb
```diff
@@ -292,6 +292,15 @@ def my_method
 
         expect_no_corrections
       end
+
+      it 'registers no offense for 1.times without block' do
```
fynsta, line 296: It does register an offense, though?

Earlopain: Yeah, right. Initially my intention was to not register something but I switched it up without updating this as well.

I'll try my hand at maybe making/extending some internal affair for this, there are already things that check for correct spec titles.

PR #12776, opened 2024-03-11
Add tests for `expect_no_corrections` and `expect_correction` loop behaviour

Earlopain:
This relates to #12763 and #12774, where behaviour for these two methods was changed. Undoing changes from these PRs will make these new tests fail.

I thought it would be nice to explicitly test against since there shouldn't ever be code here where this behaviour is actually triggered. I haven't found a place where the expectations were already tested so I made a new file for that.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/rspec/expect_offense_spec.rb
+++ b/spec/rubocop/rspec/expect_offense_spec.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::RSpec::ExpectOffense, :config do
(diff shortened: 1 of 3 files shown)


bbatsov:
I'm OK with the proposed changes. @koic are you okay with the suggested location for the tests?

koic:
Considering spec/rubocop/rspec_expect_offense_spec.rb, it seems better to place it as spec/rubocop/rspec/expect_offense_spec.rb to form the testing pair with lib/rubocop/rspec/expect_offense.rb.

PR #12782, opened 2024-03-11
Fix an error for `Style/Alias` when calling `alias_method` with fewer than 2 arguments.

Earlopain:
This happens in every context. Classes, Modules, etc. I don't believe it necessary to add one for each. I've also added tests for the other EnforcedStyle but it doesn't raise currently.

I've added no tests for `alias` calls with fewer arguments call since that is just a syntax error.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/alias_spec.rb
+++ b/spec/rubocop/cop/style/alias_spec.rb
@@ -56,6 +56,14 @@ RSpec.describe RuboCop::Cop::Style::Alias, :config do
         end
       RUBY
     end
+
+    it 'register an offense for alias_method when calling with no arguments' do
+      expect_no_offenses('alias_method')
+    end
+
+    it 'registers no offense for alias_method when calling with one argument' do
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/alias_spec.rb
```diff
@@ -208,5 +216,23 @@ def foo
...
+      expect_no_corrections
+    end
+
+    it 'registers an offense for alias_method when calling with one argument' do
+      expect_offense(<<~RUBY)
+        alias_method :foo
+        ^^^^^^^^^^^^ Use `alias` instead of `alias_method` at the top level.
+      RUBY
+
+      expect_no_corrections
```
koic, line 235: Since their usage as `alias_method(new, original)` is not completed, I think they should be ignored instead of registering an offense.

Earlopain: That works for me.

PR #12789, opened 2024-03-13
Make Style/RedundantPercentQ handle multiline source safely

boardfish:
The regex used in `Style/RedundantPercentQ` is not resilient to multiline input, meaning that its autocorrect would unsafely operate on strings defined across multiple lines with `%Q` where these used quotes. This PR changes the anchors on the regex used by the cop so that its autocorrect determines the correct delimiter safely.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_make_redundant_percent_q_safe_on_multiline_strings.md
+++ b/changelog/fix_make_redundant_percent_q_safe_on_multiline_strings.md
@@ -1,0 +1,1 @@
+* [#12789](https://github.com/rubocop/rubocop/pull/12789): Make `Style/RedundantPercentQ` safe on multiline strings. ([@boardfish][])
--- a/lib/rubocop/cop/style/redundant_percent_q.rb
+++ b/lib/rubocop/cop/style/redundant_percent_q.rb
@@ -53,7 +53,7 @@ module RuboCop
           return if interpolated_quotes?(node) || allowed_percent_q?(node)
 
           add_offense(node) do |corrector|
-            delimiter = /^%Q[^"]+$|'/.match?(node.source) ? QUOTE : SINGLE_QUOTE
+            delimiter = /\A%Q[^"]+\z|'/.match?(node.source) ? QUOTE : SINGLE_QUOTE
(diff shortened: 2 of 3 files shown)


changelog/fix_make_redundant_percent_q_safe_on_multiline_strings.md
```diff
@@ -0,0 +1 @@
+* [#12789](https://github.com/rubocop/rubocop/pull/12789): Make `Style/RedundantPercentQ` safe on multiline strings. ([@boardfish][])
```
koic, line 1:
```suggestion
* [#12789](https://github.com/rubocop/rubocop/pull/12789): Fix a false positive for `Style/RedundantPercentQ` when multiline strings containing hash rocket syntax. ([@boardfish][])
```

lib/rubocop/cop/style/redundant_percent_q.rb
```diff
@@ -53,13 +53,17 @@ def check(node)
           return if interpolated_quotes?(node) || allowed_percent_q?(node)
 
           add_offense(node) do |corrector|
-            delimiter = /^%Q[^"]+$|'/.match?(node.source) ? QUOTE : SINGLE_QUOTE
+            delimiter = delimiter_for(node.source)
```
koic, line 56: Can you update to the following instead of creating the new `delimiter_for` method:

```suggestion
            delimiter = /\A%Q[^"]+\z|'/.match?(node.source) ? QUOTE : SINGLE_QUOTE
```

changelog/fix_make_redundant_percent_q_safe_on_multiline_strings.md
boardfish: Hash rocket syntax isn't specifically the problem here – the issue is with multiline strings in general. I experienced it on a string with hash rocket syntax, but it actually affects any multiline use of %Q.

```diff
-%Q(
+"
  "Something is wrong"
-)
+"
```

PR #12792, opened 2024-03-14
Add ExemptHashRocketPairs config option to Style/RedundantDoubleSplatHashBraces

jarednielson:
Our team currently uses the Style/RedundantDoubleSplatHashBraces cop, but found that it does not prevent the double splatting of hash rocket hash literals. While this may be advantage to some, we prefer to dissallow it.

This PR introduces a config option for the cop ExemptHashRocketPairs. This option defaults to true to maintain existing behavior.
When the config option is set to false most redundant double splats on hash rocket hashes where all keys are literals will fail.

```yml
ExemptHashRocketPairs: true
```

```rb
#bad
do_something(**{ foo: 'bar' })

#good
do_something(**{ :foo => 'bar' })
```

```yml
ExemptHashRocketPairs: false
```

```rb
#bad
do_something(**{ foo: 'bar' })

#bad
do_something(**{ :foo => 'bar' })

#good
foo = :my_sym
do_something(** { foo => 'bar')
```

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -4952,7 +4952,7 @@ Style/RedundantDoubleSplatHashBraces:
   Description: 'Checks for redundant uses of double splat hash braces.'
   Enabled: pending
   VersionAdded: '1.41'
-  ExemptHashRocketPairs: false
+  ExemptHashRocketPairs: true
 
 Style/RedundantEach:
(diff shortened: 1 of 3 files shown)


jarednielson:
This PR needs a bit of cleanup

PR #12802, opened 2024-03-20
Return global offenses for `Naming/FileName` and `Naming/InclusiveLanguage`

Earlopain:
When checking against filenames, a global offense should be added. The current behaviour causes problems in things like LSPs that inspect the offense for location information since some operations on these offenses will raise a `IndexError`.

I've tweaked the specs for `InclusiveLanguage` since they just passed without any changes. With the annotation expectations, it actually checks the offense location (or lack thereof).

```rb
require "rubocop"

processed_source = RuboCop::ProcessedSource.new("", 3.3)
range = Class.new.extend(RuboCop::Cop::RangeHelp).send(:source_range, processed_source.buffer, 1, 0)
offense = RuboCop::Cop::Offense.new(:info, range, "global offense", "Naming/FileName", :unsupported)
offense.highlighted_area.begin_pos
=> # .../gems/parser-3.3.0.5/lib/parser/source/buffer.rb:274:in `fetch': index 0 outside of array bounds: 0...0 (IndexError)
```

As a note, `Style/Copyright` can add a offense in exactly the same way. However, since it supports autocorrect just switching to `add_global_offense` is not possible.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_return_global_offenses_for_inclusive_naming_and_file_name.md
+++ b/changelog/fix_return_global_offenses_for_inclusive_naming_and_file_name.md
@@ -1,0 +1,1 @@
+* [#12802](https://github.com/rubocop/rubocop/pull/12802): Return global offenses for `Naming/FileName` and `Naming/InclusiveLanguage` for empty files. ([@earlopain][])
--- a/lib/rubocop/cop/naming/file_name.rb
+++ b/lib/rubocop/cop/naming/file_name.rb
@@ -57,7 +57,7 @@ module RuboCop
           file_path = processed_source.file_path
           return if config.file_to_exclude?(file_path) || config.allowed_camel_case_file?(file_path)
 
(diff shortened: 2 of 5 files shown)


config/default.yml
```diff
@@ -2858,7 +2858,7 @@ Naming/InclusiveLanguage:
   Description: 'Recommend the use of inclusive language instead of problematic terms.'
   Enabled: false
   VersionAdded: '1.18'
-  VersionChanged: '1.49'
+  VersionChanged: <<next>>
```
koic, line 2861: These seem more like bug fixes rather than specification changes. If so, this update should be unnecessary.

Earlopain: I wasn't so sure about if this actually qualifies for a change. Moved to bugfix instead

koic:
Thanks!

PR #12804, opened 2024-03-21
Return global offenses for `Style/Copyright` when the file is empty

Earlopain:
Very similar to #12802

This adds a check that the returned offenses have a location that formatters can display. While the standard formatter catched the `IndexError`, barely any others did.

Since it is not possible to autocorrect global offenses, just don't do that. Adding autocorrect support would be a bunch more work. I've kept autocorrect for cases where the file isn't empty since, while the offense location isn't technically correct, it works.

---

Note that I removed the catching of `IndexError` for the default formatter. Barely any others, except for Tap did that. Formatters like html and markdown would raise when inspecting a empty file with `Style/Copyright`.

The added extra expectations would have caught the `Style/Copyright` and `Naming/FileName` cases, and it also raises for `Naming/InclusiveLanguage` with the test modifications I did in my previous PR.

If this change is considered too dangerous I can remove it. The added assertions should already be pretty good. 

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_return_global_offenses_for_style_copyright.md
+++ b/changelog/fix_return_global_offenses_for_style_copyright.md
@@ -1,0 +1,1 @@
+* [#12804](https://github.com/rubocop/rubocop/pull/12804): Return global offenses for `Style/Copyright` when the file is empty. ([@earlopain][])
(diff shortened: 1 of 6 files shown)


bbatsov:
Thanks!

Issue #11514, opened 2023-01-27
Cop to catch unescaped `]` in RegExp


sambostock:
## Is your feature request related to a problem? Please describe.

Ruby emits a warning for the following:

```sh
ruby -c -e '/abc]123/'
-e:1: warning: regular expression has ']' without escape: /abc]123/
Syntax OK
```

## Describe the solution you'd like

It would be nice if Rubocop could catch this statically, as that integrates better into CI.

## Describe alternatives you've considered

Tools like [Deprecation Toolkit](https://github.com/Shopify/deprecation_toolkit) can be configured to catch warnings, but in the case of Deprecation Toolkit, it will only do so while running tests, whereas we encountered a case where the file had already been read by then.

imskr:
Can I work on this?

bbatsov:
Sure.

sonalinavlakhe:
@bbatsov, if no one picked up this, I wanted to work on it.

imskr:
I was about to make a PR. I will let you know if I will not be working on this. Sorry for the delay. @sonalinavlakhe

TechTarun:
@bbatsov can I take this up?

bbatsov:
Sure!

yamakoud:
any update?

Is someone working for this?

TechTarun:
Yess I am! Will raise MR soon

rimshach2:
If no one is working on this, can I take this issue?

TechTarun:
I am working on it!

Thanks
Tarun

bbatsov:
See also https://github.com/rubocop/rubocop/pull/12805

petrenkorf:
@TechTarun any updates on this?

PR #12805, opened 2024-03-21
#11514 New cop to catch unescaped ']' in RegExp

Andy-Stack:
This pull request adds a new cop that will catch and provide an auto-fix for unescaped closing square bracket meta characters in a regular expression.


bbatsov:
I wonder if the cop is not way too specific in its current form and shouldn't be expand to account for other similar scenarios.

Andy-Stack:
Happy to try and expand this. Did you have anything in mind?

bbatsov:
@owst may have some suggestions. I recall he used to work on some Regexp related checks.

owst:
> @owst may have some suggestions. I recall he used to work on some Regexp related checks.

Yeah, this cop feels closely related to `Style/RedundantRegexpEscape`.

Having poked about a bit, I think Ruby [will warn for an unescaped `]`](https://github.com/ruby/ruby/blob/2f654588d9e0cefff1c23529d2f2672029e1bd21/regparse.c#L4158-L4160) _unless_ it appears as the first character in the pattern (tested with 3.3.0):

```
$ ruby -e 'p /]/'
/]/
$ ruby -e 'p /]X/'
/]X/
$ ruby -e 'p /X]/'
-e:1: warning: regular expression has ']' without escape: /X]/
/X]/
```

I agree with @bbatsov - this change feels overly specific at least as-is. 

However, perhaps the behaviour in this PR could be achieved by adding a style to the existing cop (perhaps after renaming it?) that allows/requires `]` to be escaped even when it is redundant to do so (at the start of the pattern). Similarly, as per https://github.com/rubocop/rubocop/pull/11152 escaping a `-` is redudant at the start or end of a character class, but I think a similar argument could be made to always require `-` to be escaped within a character class. There are possibly other cases, too.

As a name to improve on, how about a style name of `require_position_dependent` to require escapes that are only redundant in certain positions?

bbatsov:

PR #12809, opened 2024-03-23
[Fix #12731] Treat csend the same way as send for setter methods

### LINKED ISSUE #12731: `AssignmentInCondition` didn't find the style problem

jonas054:
In order to detect code like `if test&.method = 10` we need to find not only `send` nodes but also `csend` nodes.

### DIFF
--- a/lib/rubocop/cop/mixin/safe_assignment.rb
+++ b/lib/rubocop/cop/mixin/safe_assignment.rb
@@ -14,7 +14,7 @@ module RuboCop
       def_node_matcher :empty_condition?, '(begin)'
 
       # @!method setter_method?(node)
-      def_node_matcher :setter_method?, '[(send ...) setter_method?]'
+      def_node_matcher :setter_method?, '[({send csend} ...) setter_method?]'
 
       # @!method safe_assignment?(node)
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/mixin/safe_assignment.rb
```diff
@@ -14,7 +14,7 @@ module SafeAssignment
       def_node_matcher :empty_condition?, '(begin)'
 
       # @!method setter_method?(node)
-      def_node_matcher :setter_method?, '[(send ...) setter_method?]'
+      def_node_matcher :setter_method?, '[({send csend} ...) setter_method?]'
```
koic, line 17: Can you use `call` instead?

```suggestion
      def_node_matcher :setter_method?, '[(call ...) setter_method?]'
```

bbatsov:
Thanks!

PR #9156, opened 2020-12-02
[Fix #9151] Don't suggest extensions that are installed but not direct dependencies

dvandersluis:
`SuggestExtensions` checks against your gemfile for installed gems in order to make suggestions, but it did not account for rubocop extensions that were installed as dependencies of dependencies (ie. in Gemfile.lock but not Gemfile). This changes that so that it continues to look at only `Gemfile` to determine what to suggest, but it's filtered down by all activated gems.

Also I figured out a way that I could test without having to mock out bundler, so it uses bundler to ensure this works (and also I added some test cases that were missing from #9149).

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [x] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop-hq/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop-hq/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

Issue #9151, opened 2020-12-02
RuboCop suggests inherited extensions


haines:
## Expected behavior

RuboCop does not suggest to add extensions that have already been configured in an inherited gem.

## Actual behavior

RuboCop suggests extensions that are already configured.

## Steps to reproduce the problem

- set up a project that depends on `rspec`
- put `inherit_gem` in its .rubocop.yml, and inherit from a gem that depends on `rubocop-rspec` and has `require: rubocop-rspec` in its RuboCop config
- run RuboCop and receive an extension suggestion for `rubocop-rspec`

Minimal repro here: https://github.com/haines/rubocop-inherited-extensions-bug

```console
$ git clone https://github.com/haines/rubocop-inherited-extensions-bug.git
$ cd rubocop-inherited-extensions-bug/project
$ bundle install
$ bundle exec rubocop
Inspecting 2 files
.C

Offenses:

spec/some_spec.rb:3:16: C: RSpec/DescribeClass: The first argument to describe should be the class or module being tested.
RSpec.describe 'offensive' do
               ^^^^^^^^^^^

2 files inspected, 1 offense detected

Tip: Based on detected gems, the following RuboCop extension libraries might be helpful:
  * rubocop-rspec (http://github.com/rubocop-hq/rubocop-rspec)

You can opt out of this message by adding the following to your config (see https://docs.rubocop.org/rubocop/extensions.html#extension-suggestions for more options):
  AllCops:
    SuggestExtensions: false
```

From the offense that's detected, it's clear that rubocop-rspec is configured, so the "Tip" shouldn't be displayed.

## RuboCop version

```
$ bundle exec rubocop -V
1.5.1 (using Parser 2.7.2.0, rubocop-ast 1.3.0, running on ruby 2.7.2 x86_64-darwin19)
  - rubocop-rspec 2.0.1
```

bbatsov:
Sounds like a task for @dvandersluis.

dvandersluis:
Yeah it's not currently considering what's activated in rubocop.yml, just what's found from gems. I'll take a look!

dvandersluis:
@haines In the meantime this configuration should work for you:

```yaml
AllCops:
  SuggestExtensions:
    rubocop-rspec: false
```

dvandersluis:
@haines can you confirm that `rubocop-rspec` is in your `Gemfile.lock`?

haines:
@dvandersluis thanks for looking into this!

[...]

PR #12813, opened 2024-04-01
Add rubocop-rspec_rails to suggested extensions and extension doc

ydah:
This PR adds [rubocop-rspec_rails](https://github.com/rubocop/rubocop-rspec_rails) to the `SuggestExtensions` list and extension doc.

-----------------

* [-] Added tests.

### DIFF
--- a/changelog/change_add_rubocop_rspec_rails_to_suggested_extensions.md
+++ b/changelog/change_add_rubocop_rspec_rails_to_suggested_extensions.md
@@ -1,0 +1,1 @@
+* [#12813](https://github.com/rubocop/rubocop/pull/12813): Add rubocop-rspec_rails to suggested extensions and extension doc. ([@ydah][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -156,13 +156,14 @@ AllCops:
   # included.
   SuggestExtensions:
     rubocop-rails: [rails]
-    rubocop-rspec: [rspec, rspec-rails]
+    rubocop-rspec: [rspec]
     rubocop-minitest: [minitest]
     rubocop-sequel: [sequel]
     rubocop-rake: [rake]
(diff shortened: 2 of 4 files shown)


aprescott:
This appears to have resulted in the following suggestion message:

```
The following RuboCop extension libraries are installed but not loaded in config:
  * rubocop-rspec_rails
```

I assume this is related to https://github.com/rubocop/rubocop-rspec/pull/1830 which extracted RSpec Rails cops to their own gem.

From reading https://github.com/rubocop/rubocop/pull/9156 and https://github.com/rubocop/rubocop/issues/9151#issuecomment-737512700 I would have thought this suggestion message wouldn't appear? rubocop-rspec_rails is indeed a (transitive) dependency for the app generating above the suggestion, and https://github.com/rubocop/rubocop/pull/9156 seems to suggest that in this case no such message should appear:

```
    rubocop-rspec (2.29.1)
      rubocop (~> 1.40)
      rubocop-capybara (~> 2.17)
      rubocop-factory_bot (~> 2.22)
      rubocop-rspec_rails (~> 2.28)
    rubocop-rspec_rails (2.28.2)
      rubocop (~> 1.40)
```

I can of course fix the suggestion message, but I thought I'd mention it in case this is a sign of some other bug. The presence of the suggestion message implies there's missing config, which I don't think is the case, and addressing the extra suggestion message noise feels like unnecessary work given that there's no actual change happening yet.

aprescott:
Maybe also worth calling out: [rubocop-rspec's 2.28.0 changelog](https://github.com/rubocop/rubocop-rspec/blob/863bafacd1a3313d707728fbba2623f8fcd9dd6b/CHANGELOG.md#2280-2024-03-30) indicates the extraction is intended to be invisible:

> the change will be invisible to users until then

PR #12814, opened 2024-04-01
[Fix #12676] Adjust offending range in LSP

### LINKED ISSUE #12676: Allow specifying the location to report on for `check_code_length`

"I'm always frustrated when..." I'm working on a complex test and suddenly my IDE flags _every single line_ in the block because it's too long:

image

In the above, the first test is being flagged by `RSpec/MultipleExpectations` while the second is being flagged by `RSpec/ExampleLength` - while these cops belong to `rubocop-rspec`, the latter is literally just calling `check_code_length` which is a `rubocop` core method hence I'm opening this issue.

I'd like `RSpec/ExampleLength` to report on the sending node rather than the whole block, so by extension I'd like to be able to somehow control the node `check_code_length` reports as the location of the offense.

Forking `check_code_length` in `rubocop-rspec` - there's not that much code so this doesn't seem like the worst thing in the world, but I expect it would be harder to convince the team to accept such a PR and really this issue could apply to any cop using `check_code_length`.

I could also disable the cop on the test while I'm working, but that's a bit annoying to have to manage - I've got to do it each time, and then remove it if I manage to shrink the example again.

## Additional context

I've got three ideas on how this could be supported without a breaking change:
  1. support taking a block which would get called instead of `add_offense`, and passed the values currently being passed to `add_offense`; this would allow the most flexibility
      * (I think this is probably the best way to go since its not an error to pass a block to a method that doesn't `yield` so it'd gracefully fallback on older versions of Rubocop)
  2. make `location` a param with a default value; I think that should be completely not breaking, though it would be annoying having such a complex default value and technically the method would do a bit more work than it does now when `node.line_count <= max_length`
  3. make an internal `check_code_length` and then expose a couple of public methods with different signatures (including the current "original", + `check_code_length_with_location` etc); effectively 2. but in a way that makes it easier to avoid breaking changes if the signature needs to change in future

koic:
Fixes #12676.

Highlighting offenses for entire classes, modules, and methods is considered noisy in LSP. Instead, highlighting will be applied only to the class names or method names for offenses.

To maintain compatibility for uses outside of LSP, such as with `--disable-uncorrectable`, the behavior will be changed only when `RuboCop::LSP.enabled?` is true.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_adjust_offending_range_in_lsp.md
+++ b/changelog/change_adjust_offending_range_in_lsp.md
@@ -1,0 +1,1 @@
+* [#12676](https://github.com/rubocop/rubocop/issues/12676): Adjust offending range in LSP. ([@koic][])
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -602,6 +602,8 @@ Style/PerlBackrefs:
 ----
 
 This setting prevents autocorrection during editing in the editor.
+Additionally, for cases like `Metrics` cops where the highlight range extends over the entire body of classes, modules, methods, or blocks
+offending range will be confined to only name. This approach helps to avoid redundant and noisy offenses in editor display.
 
 ==== `disabled`
 
--- a/lib/rubocop/cop/mixin/code_length.rb
+++ b/lib/rubocop/cop/mixin/code_length.rb
@@ -36,7 +36,7 @@ module RuboCop
         length = calculator.calculate
         return if length <= max_length
 
-        location = node.casgn_type? ? node.loc.name : node.source_range
+        location = location(node)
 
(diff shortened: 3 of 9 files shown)


bbatsov:
I think we should document somewhere the difference in behavior. 

> Highlighting offenses for entire classes, modules, and methods is considered noisy in LSP.

I'm guessing this applies to using RuboCop in an editor in general, not just to LSP.

koic:
> I think we should document somewhere the difference in behavior.

Indeed. Regarding this, I have added a description to `docs/modules/ROOT/pages/configuration.adoc`.

> I'm guessing this applies to using RuboCop in an editor in general, not just to LSP.

Yes, `RuboCop::LSP.enabled?` also becomes true when `--editor-mode` is enabled. This means that the behavior will be covered in editor mode as well.

PR #12820, opened 2024-04-05
Add support more Capybara debugger entry points for `Lint/Debugger`

ydah:
This PR makes add support more Capybara debugger entry points for `Lint/Debugger`

- page.save_and_open_page
- page.save_and_open_screenshot
- page.save_page
- page.save_screenshot
- save_page
  - https://www.rubydoc.info/gems/capybara/Capybara%2FSession:save_page
- save_screenshot
  - https://www.rubydoc.info/gems/capybara/Capybara%2FSession:save_screenshot

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_add_support_more_capybara_debugger_entry_points.md
+++ b/changelog/change_add_support_more_capybara_debugger_entry_points.md
@@ -1,0 +1,1 @@
+* [#12820](https://github.com/rubocop/rubocop/pull/12820): Add support more Capybara debugger entry points for `Lint/Debugger`. ([@ydah][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -1671,8 +1671,14 @@ Lint/Debugger:
       - Kernel.byebug
       - Kernel.remote_byebug
     Capybara:
+      - page.save_and_open_page
+      - page.save_and_open_screenshot
+      - page.save_page
+      - page.save_screenshot
       - save_and_open_page
       - save_and_open_screenshot
+      - save_page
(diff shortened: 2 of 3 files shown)


ydah:
One thing that bothers me is that I think one of the future improvements would be to create a cop in rubocop-capybara to detect the debugger entry point for capybara and remove the entry point for capybara from rubocop. I don't have a particularly strong opinion, but the benefit of extracting to rubocop-capybara is that `Lint/Debugger` doesn't allow you to specify the path to exclude for each entry point type, and this cop check is enabled by default even if you're not using Capybara. However, since method names are quite distinctive, the same method name might not be defined and there might not be many false positives. WDYT?

koic:
> One thing that bothers me is that I think one of the future improvements would be to create a cop in [rubocop-capybara](https://github.com/rubocop/rubocop-capybara) to detect the debugger entry point for capybara and remove the entry point for capybara from rubocop

I generally agree. However, instead of creating a new Cop for RuboCop Capybara, it might be smarter to move the `Lint/Debugger` configuration specifically for Capybara to RuboCop Capybara. The following RuboCop Rails's PR may serve as a reference:
https://github.com/rubocop/rubocop-rails/pull/1097

bbatsov:
As there's already some Capybara config in RuboCop I think it's fine to merge the configuration changes. We can discuss separately injection of configuration from plugins, as that's a broader topic.

koic:
Thanks!

PR #12822, opened 2024-04-07
Run tests with Prism 0.25+

koic:
The tests are being updated to use Prism 0.25+ from Prism 0.24. Prism 0.25 has fixed many issues, allowing for the execution of tests that were previously skipped due to `broken_on: :prism`. For future purposes, `broken_on: :prism` is intentionally left in the spec_helper.rb.

Furthermore, Prism 0.25 appropriately addresses the following syntax as an error:

```console
$ ruby -ve "break"
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-darwin22]
-e:1: Invalid break
break
-e: compile error (SyntaxError)
```

However, the Parser gem incorrectly parses it:

```console
$ ruby-parse --33 -e "break"
(break)
```

This issue is expected to be reported to the Parser gem separately.

-----------------

* [x] Added tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,11 +8,7 @@ gem 'asciidoctor'
 gem 'bump', require: false
 gem 'bundler', '>= 1.15.0', '< 3.0'
 gem 'memory_profiler', platform: :mri
-# FIXME: This is a workaround for incompatibilities between Prism 0.24.0 and 0.25.0.
-# To upgrade to Prism 0.25+, it is necessary to investigate the following build error
-# and provide feedback to Prism:
-# https://github.com/rubocop/rubocop/actions/runs/8578707777/job/23512878899
-gem 'prism', '0.24.0'
+gem 'prism', '>= 0.25.0'
 gem 'rake', '~> 13.0'
 gem 'rspec', '~> 3.7'
 gem 'rubocop-performance', '~> 1.21.0'
--- a/spec/rubocop/cop/alignment_corrector_spec.rb
+++ b/spec/rubocop/cop/alignment_corrector_spec.rb
@@ -69,8 +69,7 @@ RSpec.describe RuboCop::Cop::AlignmentCorrector, :config do
         it_behaves_like 'heredoc indenter', '<<DOC', 20
       end
 
-      # FIXME: https://github.com/ruby/prism/issues/2498
-      context 'with heredoc in backticks (<<``)', broken_on: :prism do
+      context 'with heredoc in backticks (<<``)' do
         it_behaves_like 'heredoc indenter', '<<`DOC`', 20
       end
     end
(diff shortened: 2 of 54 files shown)


Earlopain:
Would it not be more correct to put the invalid syntax ones in a context where these keywords would be valid syntax? I believe most of these were already invalid code in earlier rubies, just not at parse time:

```sh
# test.rb
foo = (yield) ? a : b

$ ruby test.rb
test.rb: --> test.rb
Invalid yield
> 1  foo = yield ? a : b

test.rb:1: Invalid yield (SyntaxError)
foo = yield ? a : b
      ^~~~~

$ rbenv local 3.2.2
$ ruby test.rb
test.rb:1: Invalid yield
test.rb: compile error (SyntaxError)
```

I tested quite a few and they all gave similar errors on both 3.3 and 3.2 when trying to run.

koic:
What does this mean? The following is valid syntax in Ruby 3.2.

## Ruby 3.3

```console
$ ruby -cve 'foo = (yield) ? a : b'
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-darwin22]
-e:1: Invalid yield
foo = (yield) ? a : b
-e: compile error (SyntaxError)
```

## Ruby 3.2

```console
$ ruby -cve 'foo = (yield) ? a : b'
ruby 3.2.3 (2024-01-18 revision 52bb2ac0a6) [x86_64-darwin22]
Syntax OK
```

Earlopain:
The parser on 3.2 doesn't complain but when running it throws a `SyntaxError` error anyways. I'm just saying, the `:ruby32, unsupported_on: :prism` tests can be modified into code that both versions understand and is actually able to be interpreted. For example, the following code is parser/handled correctly but will not be tested by prism now:

```rb
def foo
  foo = (yield) ? a : b
end
```

koic:
Hmm, I'm not quite sure I understand. In Ruby 3.3, `foo = (yield) ? a : b` is a syntax error at the top level, but not within a method body. This PR makes existing test cases testable with Prism 0.25. Which test case in this PR are you referring to?

Additionally, Prism parses Ruby 3.3 and above. That means syntax parsing for Ruby 3.2 and below is not supported, and such tests should be skipped, which is what has been done.

bbatsov:
The PR looks good to me. Thanks!

Earlopain:
> Which test case in this PR are you referring to?

This here for example: https://github.com/rubocop/rubocop/pull/12822/files#diff-cbedb596a9e288e54ae57df6b0110c55194521811c147de4a0dbb9f6749af655R133-R149

The syntax is always valid on 3.2 on parse but needs to be wrapped in a block in order to parse on 3.3. Only prism honors that behaviour, the parser gem has not support for that like you said. Still, it should not register an offense when wrapped in a block on prism but that is not tested anymore. Similarly, if the parser gem fixes the behaviour and aligns with CRuby it will not be tested that no offense occurs for newer rubies.

PR #12836, opened 2024-04-12
Fix a build error

koic:
This PR prevents the following build error:

```console
NoMethodError:
  undefined method `cop_name' for nil
# ./lib/rubocop/cop/badge.rb:52:in `match?'
# ./lib/rubocop/cop/registry.rb:300:in `resolve_badge'
# ./vendor/bundle/ruby/3.3.0/gems/rubocop-rspec_rails-2.28.3/lib/rubocop-rspec_rails.rb:30:in `qualified_cop_name'
```

https://github.com/rubocop/rubocop/actions/runs/8656558784/job/23737409762

This error is due to a monkey patch in rubocop-rails_rails: https://github.com/rubocop/rubocop-rspec_rails/pull/14

The monkey patch in rubocop-rspec_rails should also be fixed, but this patch prevents RuboCop from failing due to dependencies. RuboCop does not depend on Rails, so the ideal solution would be to remove the dependency on rubocop-rspec_rails gem. However, this is not possible currently because rubocop-rspec is still dependent on it.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/badge.rb
+++ b/lib/rubocop/cop/badge.rb
@@ -49,6 +49,10 @@ module RuboCop
       end
 
       def match?(other)
+        # Prevents the following error:
+        # https://github.com/rubocop/rubocop/actions/runs/8656558784/job/23737409762
+        return false if other.nil?
+
         cop_name == other.cop_name && (!qualified? || department == other.department)
       end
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks!

koic:
This was not an appropriate workaround as a development dependency because it involved updating the production code. Instead, #12837 has been opened.

PR #12838, opened 2024-04-12
[Fix #12835] Allow global offenses to be disabled by directive comments

### LINKED ISSUE #12835: custom rule derived from FileName rule can no longer be disabled in 1.63
If you have a custom FileName rule defined, say `Acme/FileName` that is derived from the `FileName` rule, you are now unable to disable it with `# rubocop:disable Acme/FileName` at the top of the file in version 1.63+. Rubocop is insisting that the violation must be fixed.

--------

## Expected behavior

`# rubocop:disable Acme/FileName` should continue to disable the custom rule on a per file basis as it did in 1.62 and prior.

## Actual behavior

It does not disable the rule and insists we fix the violation.

Create a custom rule based on FileName:

```rb
module Acme
  class FileName < RuboCop::Cop::Naming::FileName
  # it's enough just to define the class and not actually override or add any behaviour  
  end
end
```

then create a file that violates the naming rule and try adding:
```rb
# rubocop:disable Acme/FileName
# code here
# rubocop enable Acme/FileName
```

run: `bundle exec rubocop --only Acme/FileName <filename>`

Notice the disable does not work.

## RuboCop version

```
1.63.1 (using Parser 3.3.0.5, rubocop-ast 1.31.2, running on ruby 3.3.0) +server [arm64-darwin23]
  - rubocop-factory_bot 2.25.1
  - rubocop-minitest 0.35.0
  - rubocop-performance 1.21.0
  - rubocop-rake 0.6.0
  - rubocop-sequel 0.3.4
```

Earlopain:
Fix #12835

`Naming/FileName` now returns global offenses but there are quite a few projects that disable the cop with a comment, see https://github.com/search?q=%22rubocop%3Adisable+Naming%2FFileName%22&type=code. The change to global offenses in #12802 broke that behaviour.

While this is technically a new feature I classify this as a bugfix to restore previous behaviour.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_global_offenses_directive_comments.md
+++ b/changelog/fix_global_offenses_directive_comments.md
@@ -1,0 +1,1 @@
+* [#12835](https://github.com/rubocop/rubocop/issues/12835): Allow global offenses to be disabled by directive comments. ([@earlopain][])
--- a/lib/rubocop/cop/base.rb
+++ b/lib/rubocop/cop/base.rb
@@ -186,7 +186,9 @@ module RuboCop
       def add_global_offense(message = nil, severity: nil)
         severity = find_severity(nil, severity)
         message = find_message(nil, message)
-        current_offenses << Offense.new(severity, Offense::NO_LOCATION, message, name, :unsupported)
+        range = Offense::NO_LOCATION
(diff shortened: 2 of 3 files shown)


bbatsov:
Thanks!

PR #12839, opened 2024-04-12
Fix misleading spec

bquorning:
Our users can only configure the 'Regex' key as a String object, not as a Regexp instance.

-----------------

* [ ] ~Added tests.~
* [ ] ~Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.~

### DIFF
--- a/spec/rubocop/cop/naming/file_name_spec.rb
+++ b/spec/rubocop/cop/naming/file_name_spec.rb
@@ -392,7 +392,7 @@ RSpec.describe RuboCop::Cop::Naming::FileName, :config do
   end
 
   context 'when Regex is set' do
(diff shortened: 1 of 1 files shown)


koic:
A Regexp instance can be configured with the following setting:

```yaml
Naming/FileName:
  Regex:
    - !ruby/regexp /\A[aeiou]\z/i
```

bquorning:
You are right. And I just noticed that we also mention this in the docs – at https://docs.rubocop.org/rubocop/1.63/configuration.html

PR #12850, opened 2024-04-16
[Fix #12842] Add new `Style/SendWithLiteralMethodName` cop

### LINKED ISSUE #12842: Cop Idea: Replace #send(LITERAL, ...) with method call

In projects I've worked on, I've seen cases where `#send` was called with a literal symbol or string. This can come about due to heavy refactoring where the send was once more dynamic.

For example:

```rb
# bad
send(:puts)
send('puts')
obj.send(:nil?)
obj.send('nil?')

# good
puts
obj.nil?
```

A new cop that detects `#send(LITERAL, ...)` and replaces it with the method call named by the literal.

The alternative is not implementing the cop and closing this issue. As I discover these by inspection, I manually fix them.

## Additional context

N/A

koic:
Fixes #12842.

## Summary

Detects the use of the `public_send` method with a literal method name argument. Since the `send` method can be used to call private methods, by default, only the `public_send` method is detected.

```ruby
# bad
obj.public_send(:method_name)
obj.public_send('method_name')

# good
obj.method_name
```

## Safety

This cop is not safe because it can incorrectly detect based on the receiver. Additionally, when `AllowSend` is set to `true`, it cannot determine whether the `send` method being detected is calling a private method.

## `AllowSend` option

This cop has `AllowSend` option.

### AllowSend: true (default)

```ruby
# good
obj.send(:method_name)
obj.send('method_name')
obj.__send__(:method_name)
obj.__send__('method_name')
```

### AllowSend: false

```ruby
# bad
obj.send(:method_name)
obj.send('method_name')
obj.__send__(:method_name)
obj.__send__('method_name')

# good
obj.method_name
```

-----------------

* [x] Added tests.


Earlopain:
I don't think this cop should concern itself with constants (at least by default). Surely I used a constant to intentionally factor it out, unless the constant is referenced only once I'd want to keep it.

koic:
Since a constant is generally immutable by definition, I think it can always be targeted without needing options. Are there any valid cases for excluding a constant?

Earlopain:
There are barely any cases I could find that use `public_send` over `send` for this (send is just so much more used). This is basically what I was thinking about: https://github.com/hopsoft/turbo_boost-commands/blob/9b4535ca0fe17481e70ae292d03db5b3e984d376/test/application_system_test_case.rb#L11

It uses a constant to have one place that holds which browser is being tested. I don't believe the code will improve by calling the method directly.

koic:
> I don't believe the code will improve by calling the method directly.

I see! Since using constants is intentional in some cases, it seems better to respect the user's intent and continue to target only string literals and symbol literals as is.

> I'm not sure about the name.

The cop name `Style/StaticSend` has been updated to `Style/SendWithLiteralMethodName`.

PR #12881, opened 2024-05-03
Fix incorrect autocorrect when `Style/NumericPredicate` is used with negations

fatkodima:
When using `Style/NumericPredicate` cop with `EnforcedStyle: comparison` and having some negations, rubocop suggests invalid autocorrections. 

For example:
```ruby
!foo.zero?

# Suggests to convert to
!foo == 0

# But should be
foo != 0
```

In this PR, I did the simplest change possible to just wrap the autocorrected code in the parentheses and it is assumed then after this other cops will convert it from `!(foo == 0)` to something like `foo != 0`. Which is actually the case.

### DIFF
--- a/lib/rubocop/cop/style/numeric_predicate.rb
+++ b/lib/rubocop/cop/style/numeric_predicate.rb
@@ -118,12 +118,14 @@ module RuboCop
 
           return unless numeric && operator && replacement_supported?(operator)
 
-          [numeric, replacement(numeric, operator)]
+          [numeric, replacement(node, numeric, operator)]
         end
 
-        def replacement(numeric, operation)
+        def replacement(node, numeric, operation)
           if style == :predicate
             [parenthesized_source(numeric), REPLACEMENTS.invert[operation.to_s]].join('.')
+          elsif negated?(node)
+            "(#{numeric.source} #{REPLACEMENTS[operation.to_s]} 0)"
           else
             [numeric.source, REPLACEMENTS[operation.to_s], 0].join(' ')
           end
@@ -155,6 +157,12 @@ module RuboCop
 
             [numeric, comparison]
           end
+        end
+
+        def negated?(node)
+          return false unless (parent = node.parent)
+
+          parent.send_type? && parent.method?(:!)
         end
 
         # @!method predicate(node)
--- a/spec/rubocop/cop/style/numeric_predicate_spec.rb
+++ b/spec/rubocop/cop/style/numeric_predicate_spec.rb
@@ -215,6 +215,17 @@ RSpec.describe RuboCop::Cop::Style::NumericPredicate, :config do
       RUBY
     end
 
+    it 'registers an offense for checking if a number is not zero' do
+      expect_offense(<<~RUBY)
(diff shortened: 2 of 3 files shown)


spec/rubocop/cop/style/numeric_predicate_spec.rb
```diff
@@ -207,11 +207,22 @@ def m(foo)
     it 'registers an offense for checking if a number is zero' do
       expect_offense(<<~RUBY)
         number.zero?
-        ^^^^^^^^^^^^ Use `number == 0` instead of `number.zero?`.
+        ^^^^^^^^^^^^ Use `(number == 0)` instead of `number.zero?`.
       RUBY
 
       expect_correction(<<~RUBY)
-        number == 0
+        (number == 0)
```
koic, line 214: Parentheses should not be added when it is not needed.

spec/rubocop/cop/style/numeric_predicate_spec.rb
```diff
@@ -222,22 +233,22 @@ def m(foo)
     it 'registers an offense for checking if a number is positive' do
       expect_offense(<<~RUBY)
         number.positive?
-        ^^^^^^^^^^^^^^^^ Use `number > 0` instead of `number.positive?`.
+        ^^^^^^^^^^^^^^^^ Use `(number > 0)` instead of `number.positive?`.
       RUBY
 
       expect_correction(<<~RUBY)
-        number > 0
+        (number > 0)
```
koic, line 240: Ditto.

spec/rubocop/cop/style/numeric_predicate_spec.rb
```diff
@@ -222,22 +233,22 @@ def m(foo)
...
     it 'registers an offense for checking if a number is negative' do
       expect_offense(<<~RUBY)
         number.negative?
-        ^^^^^^^^^^^^^^^^ Use `number < 0` instead of `number.negative?`.
+        ^^^^^^^^^^^^^^^^ Use `(number < 0)` instead of `number.negative?`.
       RUBY
 
       expect_correction(<<~RUBY)
-        number < 0
+        (number < 0)
```
koic, line 251: Ditto.

spec/rubocop/cop/style/numeric_predicate_spec.rb
fatkodima: Made updates.

lib/rubocop/cop/style/numeric_predicate.rb
```diff
@@ -157,6 +159,11 @@ def invert
           end
         end
 
+        def negated?(node)
+          parent = node.parent
+          parent&.send_type? && parent&.method?(:!)
```
koic, line 164: Can you refactor to the following?

```suggestion
          return false unless (parent = node.parent)

          parent.send_type? && parent.method?(:!)
```

fatkodima: Updated.

koic:
Thanks!

Issue #12511, opened 2023-12-04
Style/RedundantParentheses should allow parenthesis for compound logical statements


postmodern:
Recently `rubocop` began flagging code with the [Style/RedundantParentheses](https://docs.rubocop.org/rubocop/cops_style.html#styleredundantparentheses) rule. However, it was flagging complex compound logical statements where the parenthesis help with showing the beginning and end of said statement, which helps with code readability.

```
lib/ronin/app/schemas/params_schema.rb:55:14: C: [Correctable] Style/RedundantParentheses: Don't use parentheses around a logical expression.
10
          if (param.required? && !param.has_default?)
11
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
12
lib/ronin/app/validations/nmap_params.rb:156:14: C: [Correctable] Style/RedundantParentheses: Don't use parentheses around a logical expression.
13
          if (value && !(value >= 0.0 && value <= 1.0))
14
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
15
lib/ronin/app/validations/nmap_params.rb:162:14: C: [Correctable] Style/RedundantParentheses: Don't use parentheses around a logical expression.
16
          if (value && !(value >= 0 && value <= 9))
17
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I do still believe `Style/RedundantParentheses` can be useful for catching parenthesis around singular logical statements (ex: `if (variable.nil?)`).

I believe `Style/RedundantParentheses` should make an exception for compound logical statements, or should at least be configurable to make an exception for compound logical statements. Disabling the rule entirely would allow needless parenthesis around singular logical statements to slip by.

bbatsov:
> I do still believe Style/RedundantParentheses can be useful for catching parenthesis around singular logical statements (ex: if (variable.nil?)).

Those are super rare in practice, though, so providing special support for them seems like an overkill to me. Might be best to just implement a custom cop targeting those.

postmodern:
@bbatsov if you check the documentation for [Style/RedundantParentheses](https://docs.rubocop.org/rubocop/cops_style.html#styleredundantparentheses) it shows:

```ruby
# bad
(x) if ((y.z).nil?)

# good
x if y.z.nil?
```

[...]

PR #12892, opened 2024-05-08
Make `Style/RedundantParentheses` allow parentheses in assignment

koic:
Fixes https://github.com/rubocop/rubocop/issues/12511#issuecomment-1852817131.

This PR makes `Style/RedundantParentheses` allow parentheses around conditional expressions in assignment to respect user's intentions for readability:

```ruby
var = (foo || bar)
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_parentheses.rb
+++ b/lib/rubocop/cop/style/redundant_parentheses.rb
@@ -5,6 +5,13 @@ module RuboCop
     module Style
       # Checks for redundant parentheses.
       #
+      # Parentheses are allowed in the following cases for readability:
+      #
+      # [source,ruby]
+      # ----
+      # var ||= (foo || bar)
+      # ----
+      #
       # @example
       #
       #   # bad
@@ -160,6 +167,7 @@ module RuboCop
             return if node.semantic_operator? && begin_node.parent
             return if node.multiline? && allow_in_multiline_conditions?
             return if ALLOWED_NODE_TYPES.include?(begin_node.parent&.type)
+            return if begin_node.parent&.assignment?
             return if begin_node.parent&.if_type? && begin_node.parent&.ternary?
 
             'a logical expression'
--- a/changelog/change_make_style_redundant_parentheses_allow_parentheses_in_assignment.md
+++ b/changelog/change_make_style_redundant_parentheses_allow_parentheses_in_assignment.md
@@ -1,0 +1,1 @@
+* [#12892](https://github.com/rubocop/rubocop/pull/12892): Make `Style/RedundantParentheses` allow parentheses around conditional expressions in assignment. ([@koic][])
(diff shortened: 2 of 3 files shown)


bbatsov:
I think this should be some configuration option, not a default behavior.

koic:
I think that by default user's intent can be respected with `var = (a || b)`. If user don't like it, intentionally writing `var = foo || bar` will not register an offense. 
How about starting with that approach, and if issues become noticeable, consider adding a configuration option?

bbatsov:
My point is that if a cop is supposed to register redundant parentheses anything that's redundant, but someone might want to keep should be considered some special case. In general it's hard to know if some redundant parentheses where placed on purpose or accidentally.

koic:
That reasoning makes a lot of sense. OTOH, I apologize for sharing my personal opinion, but for instance, the operator precedence differs in the following example:

```console
$ ruby-parse -e "var = foo || bar"
(lvasgn :var
  (or
    (send nil :foo)
    (send nil :bar)))

$ ruby-parse -e "var = foo or bar"
(or
  (lvasgn :var
    (send nil :foo))
  (send nil :bar))
```

I understand that allowing parentheses in these cases provides consistent readability. But, the current behavior forces parentheses to be removed when using `||`. (which is, of course, expected behavior).
For a use case, see https://github.com/rubygems/rubygems/commit/7cc647c8f34da70caff27b3090db30e37791a72b#diff-6e1c7ff50326d1315e87ff73c66c05ba3897487bcbac0ed1290ad3a3bd889efc, though I doubt that this improved readability, from `var ||= (foo || bar)` to `var ||= foo || bar`.

So, I've added this special case to the documentation. How about starting with this?

bbatsov:
> So, I've added this special case to the documentation. How about starting with this?

Yeah, I agree the parentheses should be preserved when removing them would change the AST.

lib/rubocop/cop/style/redundant_parentheses.rb
```diff
@@ -5,6 +5,13 @@ module Cop
     module Style
       # Checks for redundant parentheses.
       #
+      # Parentheses are allowed in the following cases for readability:
```
bbatsov, line 8: Looking at the tests we've allowed them everywhere where there is assignment, and I don't think this always improves the readability. 

I still think that by default they should allowed only where removing them would change the AST, and that there should be an option to allow them in assignments more broadly.

bbatsov:
@rubocop/rubocop-core Anyone else has an opinion on how to proceed here?

bquorning:
I would expect a Style/RedundantParentheses in its default configuration to flag not just _some_ redundant parentheses but _all_ redundant parentheses.

If that results in my code looking “inconsistent” because `a = (b or c)` has parens while `a = b || c` has not, it is fine. It actually underlines the difference between the two operators. (And in my personal opinion I think many people could use the reminder that `||` and `or` _are not the same thing_.)

bbatsov:
@bquorning I totally agree with your perspective!

bbatsov:
@koic friendly ping :-) It'd be nice to wrap up this PR.

bbatsov:
@koic Do you plan to revisit this PR or should we close it?

koic:
Ah, I will retake later.

PR #12899, opened 2024-05-13
[Fix #12898] Fix parsing of `TargetRailsVersion` for prerelases

### LINKED ISSUE #12898: Rubocop::Config#gem_version_to_major_minor_float doesn't work for Rails 8.0
See https://buildkite.com/rails/rails/builds/107101#018f72e2-502d-4242-9c8f-117e744182a3 for example.

The problem is that Rails 8.0.0.alpha is being interpreted as `8.alpha` that is invalid Float. It should be interpreted as `8.0`

Earlopain:
`canonical_segments` removes trailing zeroes, which is also why there's this `fetch(1, 0)`.

```rb
Gem::Version.new("7.0.0").canonical_segments
=> [7]
Gem::Version.new("8.0.0.alpha").canonical_segments
=> [8, "alpha"]
```

The `segments` method seems to do exactly what we want:

```rb
Gem::Version.new("7.0.0").segments
=> [7, 0, 0]
Gem::Version.new("8.0.0.alpha").segments
=> [8, 0, 0, "alpha"]
```

Taking the first two entries will always work.

-------

I've started adding some tests for this since there were absolutely none here beforehand. I understand that this rails-specific stuff should eventually be moved to `rubocop-rails`, but until then lets ensure everything works as expected.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_target_rails_version_for_prereleases.md
+++ b/changelog/fix_target_rails_version_for_prereleases.md
@@ -1,0 +1,1 @@
+* [#12898](https://github.com/rubocop/rubocop/issues/12898): Fix an error for `TargetRailsVersion` when parsing from the lockfile with prerelease rails. ([@earlopain][])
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

bbatsov:
Likely early next week.

PR #12900, opened 2024-05-13
Add `Style/CStyleIncrementDecrement` cop

nvasilevski:
### Summary 
This PR introduces a new  `Style/CStyleIncrementDecrement`, aimed at identifying and correcting the usage of C-style increment (++) and decrement (--) operators. Unlike many other programming languages, Ruby does not support these operators for incrementing or decrementing values. Instead, these operators behave as double unary operators and in most cases do nothing. For instance, `++x` translates to `+(+x)`, and `--x` translates to `-(-x)`, which simply affects the sign, not the magnitude. 

```ruby
# bad
++counter
--counter

# good
counter += 1
counter -= 1
```

-----------------

* [x] Added tests.


koic:
This concerns a fundamental syntax of Ruby, and I've reservations about introducing mechanical detection as a cop. First, this cop doesn't distinguish between redundant expressions like `++str` and `--str`. These are redundant expressions of `+str` and `-str`. Secondly, it is unclear to me whether the true positives of inexperienced Ruby developers using `++int`, or the false positives where such redundant expressions occur, are more valuable. Lastly, `var += 1` and `++1` are entirely incompatible syntaxes. It is questionable to introduce this as a cop, especially since C-style `var++` or `var--` would result in a syntax error in Ruby, yet `++var` and `--var` are considered valid syntax. 

If this case becomes a cop, rather than comparing it to C style, it would be better to detect redundant `+` or `-` in `++str` or `--str`, thereby guiding users towards the correct usage of `+str` or `-str`.

For these reasons, I'll close this proposal. Thank you.

nvasilevski:
I agree, it looks like autocorrection is not an option for this kind of cop and implying that `++` was mistakenly used as an increment is too specific and doesn't account for other scenarios

>If this case becomes a cop, rather than comparing it to C-style, it would be better to detect redundant + or - in ++str or --str, thereby guiding users towards the correct usage of +str or -str.

Yeah, I think I'll let this idea to soak but if agree if such cop was to exist it should only hint on potential redundancy of `--` or `++` in any context without guessing what motivated this usage (c-style increment vs double string freeze/unfreeze)

Thanks for the feedback!

PR #12902, opened 2024-05-14
[Fix #12888] Fix `--no-exclude-limit` for cops with `Max` config option

### LINKED ISSUE #12888: Metrics/AbcSize with `auto-gen-only-exclude` option
When comparing the `.rubocop_todo.yml` file generated with: 
```sh
rubocop --auto-gen-config --auto-gen-only-exclude --no-exclude-limit
``` 

versus 

```
rubocop --auto-gen-config --auto-gen-only-exclude --exclude-limit=100000
```

 the exclude section for `Metrics/AbcSize` is absent with the former and a `Max` value is given instead.

--------

## Expected behavior

I would expect the two versions of `.rubocop_todo.yml` to be the same since the number of excluded files for `Metrics/AbcSize` is only 52.

## Actual behavior

Using the `--no-exclude-limit` option gives:

```
# Offense count: 69
# Configuration parameters: AllowedMethods, AllowedPatterns, CountRepeatedAttributes.
Metrics/AbcSize:
  Max: 43
```

The problem only occurred when the offense count was 25 or above but it could be the number of excluded files is also a factor.

## RuboCop version

```
1.63.4 (using Parser 3.3.1.0, rubocop-ast 1.31.3, running on ruby 3.3.1) [arm64-darwin23]
  - rubocop-capybara 2.20.0
  - rubocop-factory_bot 2.25.1
  - rubocop-graphql 1.5.1
  - rubocop-rails 2.24.1
  - rubocop-rspec 2.29.2
  - rubocop-rspec_rails 2.28.3
```

Earlopain:
Fixes #12888. Basically, the problem here was a missing `no_exclude_limit?` when deciding if `Max` should be set. Factored out into its own method because of Metrics cops.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_no_exclude_limit_max.md
+++ b/changelog/fix_no_exclude_limit_max.md
@@ -1,0 +1,1 @@
+* [#12888](https://github.com/rubocop/rubocop/issues/12888): Fix `--no-exclude-limit` generating a todo with `Max` config instead of listing everything out with `Exclude`. ([@earlopain][])
--- a/lib/rubocop/formatter/disabled_config_formatter.rb
+++ b/lib/rubocop/formatter/disabled_config_formatter.rb
@@ -116,18 +116,22 @@ module RuboCop
       def set_max(cfg, cop_name)
         return unless cfg[:exclude_limit]
 
-        max_set_in_user_config =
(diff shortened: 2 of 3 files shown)


koic:
Thanks!

PR #12905, opened 2024-05-15
Support `AllCops:ActiveSupportExtensionsEnabled` for `Style/SymbolProc`

koic:
Follow up https://github.com/rubocop/rubocop/commit/d39f07376d7928f7b6ee1df12cdba086cb552e5e.

This PR supports `AllCops:ActiveSupportExtensionsEnabled` instead of a TODO comment in `Style/SymbolProc` cop.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_support_active_support_extensions_enabled_for_style_symbol_proc.md
+++ b/changelog/change_support_active_support_extensions_enabled_for_style_symbol_proc.md
@@ -1,0 +1,1 @@
+* [#12890](https://github.com/rubocop/rubocop/issues/12890): Support `ActiveSupportExtensionsEnabled` for `Style/SymbolProc`. ([@koic][])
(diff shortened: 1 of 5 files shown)


bbatsov:
Thanks!

etiennebarrie:
I'm seeing a false positive here with CSV converters. They also check the arity of the proc, so auto-correcting from `->(x) { x.upcase }` to `lambda(&:upcase)` breaks. I guess it's fine because the autocorrect is not safe, but just wanted to point it out in case someone else hits this.

I was able to prevent Rubocop from reporting the offense by using the `ActiveSupportExtensionsEnabled` configuration option: https://github.com/Shopify/maintenance_tasks/commit/d24e07a9087b8159502198ce439b6e8a08ac6149. Although I don't think this applies to everybody because a Ruby gem without Active Support could hit the same issue, but my case is a Rails engine so it's fine.

PR #12907, opened 2024-05-16
[Fix #12884] Allow `cop_class.documentation_url` to take a config

### LINKED ISSUE #12884: `cop.documentation_url` is nil for extension cops

This one here: https://github.com/rubocop/rubocop/blob/7f9595aa3ca3524f888b1464c4eeb7e15aabfc22/lib/rubocop/cop/base.rb#L63-L69

It only returns something for cops in this repository. The Ruby LSP makes use of this method to link to a cops documentation in code lenses:

image
The blue link takes me to https://docs.rubocop.org/rubocop/cops_style.html#stylestringliterals

Here is how it looks for rubocop-rails (and basically all other extension) cops:
image

The method has the following comment:
> Cops (other than builtin) are encouraged to implement this

This hasn't happenend yet. I'd like to do so, however I don't really know how I would do that since the Documentation class is private.

`rubocop --show-docs-url Rails/Exit` prints the correct url, so I wanted to just copy how it does that over. However I don't know how the get to the config/config store in the context of that method without requiring users to pass it in themselves.

Earlopain:
Fix #12884 (almost). It requires passing in a config object but that seems ok to require.

The current implementation is not very useful to consumers since it requires every extension to overwrite this method.
No official extension currenty does that.

The behaviour of `--show-docs-url` is almost what's required. It takes a config to construct this url and returns correct results for all official extensions.

The url it provides falls back to the default RuboCop docs url which is undesirable in all cases I can think about. That behaviour has been removed. No 3rd-party extension cops will be found on the RuboCop docs website. Only buildin cops will be found there.

All extensions that provide `DocumentationBaseURL` will now have result returned by `documentation_url` without needing to do something themselves (as long as proper config object is passed in)

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_allow_cop_documentation_url_to_take_config.md
+++ b/changelog/change_allow_cop_documentation_url_to_take_config.md
@@ -1,0 +1,1 @@
+* [#12884](https://github.com/rubocop/rubocop/issues/12884): Align output from `cop.documentation_url` with `--show-docs-url` when passing a config as argument. ([@earlopain][])
--- a/lib/rubocop/cli/command/show_docs_url.rb
+++ b/lib/rubocop/cli/command/show_docs_url.rb
@@ -26,10 +26,10 @@ module RuboCop
 
           cops_array.each do |cop_name|
             cop = registry_hash[cop_name]
-
             next if cop.empty?
 
-            puts Cop::Documentation.url_for(cop.first, @config)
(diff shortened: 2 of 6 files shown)


bbatsov:
Seems like a reasonable change to me. Thanks!

PR #12913, opened 2024-05-19
Update `Lint/Void` to recognize more nonmutating methods

vlad-pisanov:
Update `Lint/Void` to recognize more nonmutating methods when the `CheckForMethodsWithNoSideEffects` option is used.

This PR adds one (now)-native `transform_keys` method, and about a dozen common ActiveSupport methods such as `stringify_keys`, `compact_blank` and `squish`.

The ActiveSupport methods are only recognized when `AllCops/ActiveSupportExtensionsEnabled: true` (such as in a Rails project)

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/void.rb
+++ b/lib/rubocop/cop/lint/void.rb
@@ -67,17 +67,28 @@ module RuboCop
         UNARY_OPERATORS = %i[+@ -@ ~ !].freeze
         OPERATORS = (BINARY_OPERATORS + UNARY_OPERATORS).freeze
         VOID_CONTEXT_TYPES = %i[def for block].freeze
-        NONMUTATING_METHODS_WITH_BANG_VERSION = %i[capitalize chomp chop compact
-                                                   delete_prefix delete_suffix downcase
-                                                   encode flatten gsub lstrip merge next
-                                                   reject reverse rotate rstrip scrub select
-                                                   shuffle slice sort sort_by squeeze strip sub
-                                                   succ swapcase tr tr_s transform_values
-                                                   unicode_normalize uniq upcase].freeze
+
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/void.rb
```diff
@@ -67,17 +67,28 @@ class Void < Base
...
-                                                   succ swapcase tr tr_s transform_values
-                                                   unicode_normalize uniq upcase].freeze
+
+        NONMUTATING_NATIVE_METHODS_WITH_BANG_VERSION = %i[
+          capitalize chomp chop compact
+          delete_prefix delete_suffix downcase
+          encode filter flatten gsub lstrip merge next
+          reject reverse rotate rstrip scrub select
+          shuffle slice sort sort_by squeeze strip sub
+          succ swapcase tr tr_s transform_keys
```
vlad-pisanov, line 77: added `filter` and `transform_keys`

koic:
The goal seems reasonable, but it might be preferable to adjust the approach slightly. Instead of using `ActiveSupportExtensionsEnabled`, it might be better to provide a user config parameter like `NonmutatingMethods`. (There might be a better name than `NonmutatingMethods`.)
After that, In RuboCop Rails, configure `NonmutatingMethods` as follows:

```yaml
# config/default.yml in the RuboCop Rails
NonmutatingMethods:
  - compact_blank
  - deep_merge
  - deep_stringify_keys
  - ...
```

This approach allows for the elimination of the `NONMUTATING_ACTIVE_SUPPORT_METHODS_WITH_BANG_VERSION` constant, ensuring that changes in Active Support do not have an impact.

koic:
ping @vlad-pisanov

Issue #12910, opened 2024-05-16
"too many arguments for format string" warning


andyw8:
It doesn't cause a problem, but when running with warnings enabled, the following is emitted:

```
/home/runner/work/.../vendor/bundle/ruby/3.3.0/gems/rubocop-1.63.5/lib/rubocop/cop/mixin/surrounding_space.rb:87: warning: too many arguments for format string
```

(I think it may be a clash between the `format` defined by RuboCop and `format` defined in `Kernel`?)

The same can be seen when running RuboCop's own tests with `config.warnings = true`.

Earlopain:
There are quite a few of these. Some cops define multiple offense format strings with different "arity" and they just all get the same amount of arguments passed.

I looked at this in the past and some are trivial to fix, some require a bit more finesse.

andyw8:
Ah, I see, thanks. Feel free to close if it's already been discussed previously.

Earlopain:
I haven't talked about this with anyone. It's not a very common usecase since it's mostly a cli tool but I think it would be nice to have RuboCop warning-free anyways for those that use it in other ways.

You probably won't be able to raise on warnings since RuboCop inspects some rather strange-looking code but most looked fixable from what I remember.

Earlopain:
I'm trying to some of these off:
* https://github.com/whitequark/parser/pull/1013 (lets see how that goes)
* #12911

Remaining:
* if_with_boolean_literal_branches
* erb_new_arguments
* space_inside_string_interpolation (the one this issue is about)
* Setting `Encoding.default_external`. This one needs to be ignored and is test only.

Wasn't quite as much as I thought it would be. The regex warnings made up the bulk of all the warnings in the test suite

PR #12915, opened 2024-05-20
Remove warnings during tests when overwriting encoding

Earlopain:
Just setting this is enough to emit a warning:
> warning: setting Encoding.default_internal

Use helpers to set internal/external encoding for the duration of a block without emitting these warnings.

Part of #12910

-----------------

* [ ] Added tests.

### DIFF
--- a/spec/support/encoding_helper.rb
+++ b/spec/support/encoding_helper.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+# Overwrite the default internal/external encoding for the duration of a block.
+# Ruby discourages this by emitting a warning when modifying, these helpers exist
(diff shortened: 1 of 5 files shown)


spec/support/encoding_helper.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module EncodingHelper
```
bbatsov, line 3: Might be good to document the module a bit, so it's clearer what we're aiming to achieve with it.

Earlopain: 👍

bbatsov:
Thanks!

PR #12918, opened 2024-05-21
Suggest correct formatter name

koic:
This PR suggests correct formatter name when wrong similar format name is specified.

## Before

```console
$ bundle exec rubocop --format=quite
No formatter for "quite"
(snip)
```

## After

```console
$ bundle exec rubocop --format=quite
Formatter "quite" not formatter. Did you mean? "quiet"
(snip)
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/formatter/formatter_set.rb
+++ b/lib/rubocop/formatter/formatter_set.rb
@@ -27,6 +27,7 @@ module RuboCop
         '[t]ap'         => 'TapFormatter',
         '[w]orst'       => 'WorstOffendersFormatter'
       }.freeze
+      BUILTIN_FORMATTER_NAMES = BUILTIN_FORMATTERS_FOR_KEYS.keys.map { |key| key.delete('[]') }
 
       FORMATTER_APIS = %i[started finished].freeze
 
@@ -88,7 +89,12 @@ module RuboCop
           /^\[#{specified_key}\]/.match?(key) || specified_key == key.delete('[]')
         end
 
-        raise %(No formatter for "#{specified_key}") if matching_keys.empty?
+        if matching_keys.empty?
+          similar_name = NameSimilarity.find_similar_name(specified_key, BUILTIN_FORMATTER_NAMES)
+          suggestion = %( Did you mean? "#{similar_name}") if similar_name
+
+          raise Rainbow(%(Formatter "#{specified_key}" not found.#{suggestion})).red
+        end
 
         raise %(Cannot determine formatter for "#{specified_key}") if matching_keys.size > 1
 
--- a/changelog/new_suggest_correct_formatter_name.md
+++ b/changelog/new_suggest_correct_formatter_name.md
@@ -1,0 +1,1 @@
+* [#12917](https://github.com/rubocop/rubocop/pull/12917): Suggest correct formatter name for `--format` command line option. ([@koic][])
--- a/spec/rubocop/cli/options_spec.rb
+++ b/spec/rubocop/cli/options_spec.rb
@@ -1715,7 +1715,16 @@ RSpec.describe 'RuboCop::CLI options', :isolated_environment do # rubocop:disabl
       context 'when unknown format name is specified' do
         it 'aborts with error message' do
           expect(cli.run(['--format', 'unknown', 'example.rb'])).to eq(2)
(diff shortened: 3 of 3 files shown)


lib/rubocop/formatter/formatter_set.rb
```diff
@@ -88,7 +89,12 @@ def builtin_formatter_class(specified_key)
           /^\[#{specified_key}\]/.match?(key) || specified_key == key.delete('[]')
         end
 
-        raise %(No formatter for "#{specified_key}") if matching_keys.empty?
+        if matching_keys.empty?
+          similar_name = NameSimilarity.find_similar_name(specified_key, BUILTIN_FORMATTER_NAMES)
+          suggestion = %( Did you mean? "#{similar_name}") if similar_name
+
+          raise Rainbow(%(No formatter for "#{specified_key}".#{suggestion})).red
```
bbatsov, line 96: Maybe change this to "Formatter x not found/does not exist. ..."? The current message looks a bit off to me.

koic: Thanks! I've updated the message.

bbatsov:
@koic One test is currently failing:

```
RuboCop::CLI options -f/--format builtin formatters when unknown format name is specified aborts with error message
     Failure/Error: expect($stderr.string.include?('No formatter for "unknown"')).to be(true)
```

lib/rubocop/formatter/formatter_set.rb
```diff
@@ -88,7 +89,12 @@ def builtin_formatter_class(specified_key)
...
 
-        raise %(No formatter for "#{specified_key}") if matching_keys.empty?
+        if matching_keys.empty?
+          similar_name = NameSimilarity.find_similar_name(specified_key, BUILTIN_FORMATTER_NAMES)
+          suggestion = %( Did you mean? "#{similar_name}") if similar_name
+
+          raise Rainbow(%(Formatter "#{specified_key}" not found.#{suggestion})).red
+        end
 
         raise %(Cannot determine formatter for "#{specified_key}") if matching_keys.size > 1
```
bbatsov, line 99: Maybe in this message we should clarify that multiple formatters were matched and users need to be more specific.

lib/rubocop/formatter/formatter_set.rb
```diff
@@ -88,7 +89,12 @@ def builtin_formatter_class(specified_key)
           /^\[#{specified_key}\]/.match?(key) || specified_key == key.delete('[]')
         end
 
-        raise %(No formatter for "#{specified_key}") if matching_keys.empty?
+        if matching_keys.empty?
+          similar_name = NameSimilarity.find_similar_name(specified_key, BUILTIN_FORMATTER_NAMES)
+          suggestion = %( Did you mean? "#{similar_name}") if similar_name
+
+          raise Rainbow(%(Formatter "#{specified_key}" not found.#{suggestion})).red
```
bbatsov, line 96: Btw, reading the code more carefully I see that we're matching the command-line param using regex instead of doing so literally, which was the reason for the old slightly confusing message.

I guess the reason here is to type less, but it seems to me it'd be better if we were explicit and don't have to deal with ambiguity.

koic: I apologize if I'm mistaken about the issue being pointed out.
In this PR, `NameSimilarity.find_similar_name` is used to find approximations for the command-line parameter (format name) entered. This involves calculating the Levenshtein distance provided by DidYouMean. As a result, it suggests the most similar correct value for incorrect inputs.

lib/rubocop/formatter/formatter_set.rb
koic: It seems that multiple formatters are never specified, and if that's the case, this logic will not be called. I will investigate separately whether it can be removed as unnecessary logic.

lib/rubocop/formatter/formatter_set.rb
bbatsov: Your solution is fine, the issue that I'm talking about is in the existing code. Basically what bothers me is the regexp we're using to find the potential name(s) of a formatter. It seems to me we should just assume that whatever was passes is the name of the formatter, and in this case your proposal will still be just as useful.

PR #12927, opened 2024-05-23
[Fix #12926] Fix false positive for `Style/SuperArguments`

### LINKED ISSUE #12926: Style/SuperArguments is probably not safe to autocorrect
## Expected behavior

Safe autofix of `Style/SuperArguments` offenses does not break CI

## Actual behavior

Safe autofix breaks CI

https://github.com/ifad/chronomodel/actions/runs/9206966951/job/25325982139?pr=294#step:6:1076

If we autofix

https://github.com/ifad/chronomodel/blob/804afe24cc5cbf049822e56076c81e4614ec3497/lib/chrono_model/adapter/migrations.rb#L83

then

```
rspec ./spec/chrono_model/adapter/migrations_spec.rb:125
```

fails

```
     Failure/Error: super
     
     LocalJumpError:
       no block given (yield)
```

## RuboCop version

```
$ rubocop -V
1.64.0 (using Parser 3.3.1.0, rubocop-ast 1.31.3, running on ruby 3.3.1) [x86_64-darwin23]
  - rubocop-packaging 0.5.2
  - rubocop-performance 1.21.0
  - rubocop-rails 2.25.0
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.29.2
```

Earlopain:
Fixes #12926. The following snippet demonstrates the issue:

```rb
class A
  def positional_arg(a)
    puts a
  end

  def block_arg(&block)
    yield
  end
end

class B < A
  def positional_arg(a = nil)
    a = 'b'
    super
  end

  def block_arg(&block)
    block = proc { puts 'b' }
    super
  end
end

B.new.positional_arg('a')
B.new.positional_arg
B.new.block_arg { puts 'a' }
B.new.block_arg
```

Modifying a methods block argument passes along the original block to `super`.

This fix is overly defensive and will miss some cases that could be autocorrected. This however would make this very complicated:

```rb
def foo(&blk)
  super
  blk = proc {} # Technically fine
end
```

The same goes for nested method definitions

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positive_for_super_arguments.md
+++ b/changelog/fix_false_positive_for_super_arguments.md
@@ -1,0 +1,1 @@
+* [#12926](https://github.com/rubocop/rubocop/issues/12926): Fix a false positive for `Style/SuperArguments` when the methods block argument is reassigned before `super`. ([@earlopain][])
--- a/lib/rubocop/cop/style/super_arguments.rb
+++ b/lib/rubocop/cop/style/super_arguments.rb
@@ -26,10 +26,19 @@ module RuboCop
       #   def method(*args, **kwargs)
       #     super()
       #   end
+      #
+      #   # good - assigning to the block variable before calling super
+      #   def method(&block)
+      #     # Assigning to the block variable would pass the old value to super,
+      #     # under this circumstance the block must be referenced explicitly.
+      #     block ||= proc { 'fallback behavior' }
+      #     super(&block)
+      #   end
       class SuperArguments < Base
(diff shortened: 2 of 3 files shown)


bbatsov:
Probably this should be mentioned in the cop's docs, so people are not surprised by this conservative behavior.

Earlopain:
Ok, sure. I've also added a link to the ruby bugtracker issue I made

bbatsov:
Thanks!

PR #12940, opened 2024-05-28
Add Exclude per constant settings for Lint/DeprecatedConstants cop

r4do:
Adds `Exclude` settings per specific deprecated constant for `Lint/DeprecatedConstants`.
Usecase example: monorepo in decomposition state.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_add_exclude_settings_for_lint_deprecated_constants.md
+++ b/changelog/change_add_exclude_settings_for_lint_deprecated_constants.md
@@ -1,0 +1,1 @@
+* [#12940](https://github.com/rubocop/rubocop/pull/12940): Add `Exclude` settings for `Lint/DeprecatedConstants` per constant. ([@r4do][])
--- a/lib/rubocop/cop/lint/deprecated_constants.rb
+++ b/lib/rubocop/cop/lint/deprecated_constants.rb
@@ -13,6 +13,8 @@ module RuboCop
       #     'DEPRECATED_CONSTANT':
       #       Alternative: 'alternative_value'
       #       DeprecatedVersion: 'deprecated_version'
+      #       Exclude:
+      #         - 'path/to/exclude'
       #
       # By default, `NIL`, `TRUE`, `FALSE`, `Net::HTTPServerException, `Random::DEFAULT`,
       # `Struct::Group`, and `Struct::Passwd` are configured.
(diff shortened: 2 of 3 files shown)


Earlopain:
Would creating a separate `.rubocop.yml` work for you? There are many cops with extra config which users may want to tweak for specific folders (monorepo or not) and I don't think adding support for `Exclude` to each of them is the solution (unless it would be something generic).

You make one common `.rubocop.yml` in the root folder and have more specific ones in subfolders:

```yml
# .rubocop.yml
Lint/DeprecatedConstants:
  'Common':
    Alternative: Deprecations
  'Go':
    Alternative: Here
```

```yml
# subfolder/.rubocop.yml
inherit_from:
  - '../.rubocop.yml'

Lint/DeprecatedConstants:
  'NotDeprecatedEverywhere':
    Alternative: Whatever
```

Does that look good? If you have a group of common deprecations for multiple folders you can also make a config file to inherit from.

r4do:
There are use cases where separate configs can't work.
E.g. constant `Foo` could be allowed for file `app/services/foo_service.rb`, and deprecated for the rest. OTOH, `Bar` should be allowed for `app/services/bar_service.rb`, and deprecated for the rest. In this case config won't help. Could possibly sound like a bad app design, so I'd like to see other opinion on this.

Speaking about the generic solution, I'm not sure how can this be done, so open for suggestions. The best suggestion I have is to extract `exclude_file?` to some base class.

Earlopain:
> E.g. constant Foo could be allowed for file app/services/foo_service.rb, and deprecated for the rest. ... Bar should be allowed for app/services/bar_service.rb, and deprecated for the rest

In that case I would just use an inline directive to disable the cop for lines that do access it. RuboCop does something similar with its internal cops that encourage one api over the other: https://github.com/rubocop/rubocop/blob/49c6e92964e87435af7fee6d87bc6c4012fc8f04/lib/rubocop/cop/internal_affairs/create_empty_file.rb and https://github.com/rubocop/rubocop/blob/49c6e92964e87435af7fee6d87bc6c4012fc8f04/spec/support/file_helper.rb#L26-L30. That snippet is the only part where the discouraged method is used since it is implemented in terms of that method. The `ruby-lsp` does something similar where it creates shorthand names for constants that should always be used (except of course in the place where these aliases are made in the first place).

-------

Generic solution was just a thought. There are no other cops that implement specific excludes like propsed here which is why I mentioned it

bbatsov:
> Generic solution was just a thought. There are no other cops that implement specific excludes like propsed here which is why I mentioned it

Indeed. That's why I don't like this proposal and would suggest using some other way to achieve more or less the same result (e.g. nested configs, todos, etc).

r4do:
Ok, got it, thanks for your opinions, guys.
Closing then.

Issue #8363, opened 2020-07-18
CI


marcandre:
1) Are we going to switch to github actions, or will we remain on CircleCI / whatever else we are using?

2) There seems to be an issue with the internal investigation not using the up to date source, as [we broke rubocop-rspec](https://github.com/rubocop-hq/rubocop-rspec/issues/974) recently even though the CI was all green

3) We should consider also running the spec suites of the main dependencies like `rubocop-rspec`

bbatsov:
> Are we going to switch to github actions, or will we remain on CircleCI / whatever else we are using?

I've been thinking about moving everything to GHA for the sake of consistency. I like CircleCI, but I'd rather use 1 CI instead of 2. 

> There seems to be an issue with the internal investigation not using the up to date source, as we broke rubocop-rspec recently even though the CI was all green

Is there any ticket about this?

> We should consider also running the spec suites of the main dependencies like rubocop-rspec

That's a good idea, although the CI is already a bit slow.

marcandre:
> > Are we going to switch to github actions, or will we remain on CircleCI / whatever else we are using?
> 
> I've been thinking about moving everything to GHA for the sake of consistency. I like CircleCI, but I'd rather use 1 CI instead of 2.

Good. I could check it out, I have a bunch of things I'd like to do first though. Let me know if you want to do it.

> > There seems to be an issue with the internal investigation not using the up to date source, as we broke rubocop-rspec recently even though the CI was all green
> 
> Is there any ticket about this?

I linked to it: https://github.com/rubocop-hq/rubocop-rspec/issues/974

> That's a good idea, although the CI is already a bit slow.

True. I want to modify the AST one so that internal investigation is done separately and on a single combination of platforms; only the spec suite should really run on different versions of Ruby. It will also make it clear if we are not green if it's style only, or specs. If we do the same for RuboCop it should make things faster a bit I think.

bquorning:
We are no longer testing against ruby head, because the builds of that custom Docker image started failing over 3 months ago and nobody noticed [until now](https://github.com/rubocop/circleci-ruby-snapshot-image/issues/21).

There is (what I assume to be) an official and maintained build of ruby head for GitHub Actions at https://github.com/ruby/setup-ruby / https://github.com/ruby/ruby-dev-builder which we can use instead, if and when we switch over.

If nobody else is on the task, I would like to take a stab at it. (I think most of the work was actually done in #7893)

Earlopain:
[...]

PR #12941, opened 2024-05-28
Move `documentation-check` to GitHub actions

Earlopain:
Working towards https://github.com/rubocop/rubocop/issues/8363

I'll do this in smaller steps since the workflow is somewhat large. Should be overall rather simple, though the codeclimate integration will need some extra work.

Newly added job output: https://github.com/rubocop/rubocop/actions/runs/9273549841/job/25513896170?pr=12941

------------------

Question: Is CodeClimate even still in use? The badges in the readme show some stats but the link gives 404.

-----------------

* [ ] Added tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -215,25 +215,10 @@ jobs:
             ./tmp/cc-test-reporter sum-coverage tmp/codeclimate.*.json --parts 5 --output tmp/codeclimate.total.json
             ./tmp/cc-test-reporter upload-coverage --input tmp/codeclimate.total.json
 
-  # Miscellaneous tasks
-  documentation-checks:
-    docker:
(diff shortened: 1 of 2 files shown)


bbatsov:
> Question: Is CodeClimate even still in use? The badges in the readme show some stats but the link gives 404.

I thought it was, although I have to admit I haven't seen reports from it recently. Perhaps something changed/broke and we didn't notice it?

Earlopain:
> I thought it was, although I have to admit I haven't seen reports from it recently. Perhaps something changed/broke and we didn't notice it?

I found https://codeclimate.com/github/rubocop-hq/rubocop (old org name) but that says the last master build was 18 days ago which also doesn't look right. Maybe you can rename it? Otherwise the simplest thing would just be to point these to the correct location.

PR #12943, opened 2024-05-29
Move `ascii_spec`, `internal_investigation` to GitHub actions

Earlopain:
The only thing left after this are the "normal" specs.

Unfortunatly GitHub doesn't support yml anchors so the matrix is duplicated. There are ways to have the matrix generated/reusable but that results in complex/unreadable actions from my experience. This ain't so bad, CircleCI didn't use/have matrix, you win some you loose some.

---------------

I'm moving the jruby job to the different jobs. I don't see why Jurby should have its own section instead of being part of the matrix. This also should speed up green CI by a bit. JRuby does take the longest and now it can execute these in parallel if there are enough available workers.

-----------------

* [ ] Added tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,11 +22,6 @@ spec_steps: &spec_steps
       paths:
         - codeclimate.*.json
 
-ascii_spec_steps: &ascii_spec_steps
-  - checkout
-  - run: bundle install
-  - run: bundle exec rake ascii_spec
-
 jobs:
   # Ruby 2.7
   ruby-2.7-spec:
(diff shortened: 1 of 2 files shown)


bbatsov:
> I'm moving the jruby job to the different jobs. I don't see why Jurby should have its own section instead of being part of the matrix. This also should speed up green CI by a bit. JRuby does take the longest and now it can execute these in parallel if there are enough available workers.

Good idea!

> Unfortunatly GitHub doesn't support yml anchors so the matrix is duplicated. There are ways to have the matrix generated/reusable but that results in complex/unreadable actions from my experience. This ain't so bad, CircleCI didn't use/have matrix, you win some you loose some.

That's fine by me.

Earlopain:
This seems to work quite nicely actually, the windows jobs also benefit from the split. CI went from the longest job taking ~17 minutes (jruby) to ~9minutes overall

PR #12946, opened 2024-05-30
Slightly speed up CI runs

Earlopain:
GitHub actions can run 4 workers at once. There's also no need to explicitly set this variable as these are all vms. I presume that CircleCI reported a value too high.

Also a good chance to see what the coverage upload is up to now.

-----------------

* [ ] Added tests.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -18,9 +18,6 @@ jobs:
   spec:
     name: Spec - ${{ matrix.os }} ${{ matrix.ruby }}
     runs-on: ${{ matrix.os }}-latest
-    env:
-      # See https://github.com/tmm1/test-queue#environment-variables
-      TEST_QUEUE_WORKERS: 2
     strategy:
       fail-fast: false
       matrix:


Earlopain:
Is it possible that `CC_TEST_REPORTER_ID` was overwritten in CircleCI project settings? I've copied the value from the circleci config here but I suspect that that is not the actual value:
image

If that is indeed the case then you'll need to create a repository secret https://github.com/rubocop/rubocop/settings/secrets/actions/new with the correct value to reference in the job. The correct value is here: https://codeclimate.com/repos/5ee70a8d0cdfc901a1002d23/settings/test_reporter

koic:
There were concerns that not limiting the number of parallel processes might cause memory shortages in JRuby, but it seems there are no issues. The CI failures appear to be unrelated to this change, so I will merge it and investigate the failures separately.

bbatsov:
Something is messed up in CodeClimate as the old links to the repos are there (those from `rubocop-hq`) and they seem to be active, but their settings can't be changed. Also `rubocop/rubocop` is there twice for some reason... I've added some `CC_TEST_REPORTER_ID` from one of them, but I think we'll need to contact CC's support to clean up the legacy stuff and this duplication (when I try to delete one of the two RuboCop repos it blows up). 

image

bbatsov:
Okay, it seems I managed to delete one `rubocop/rubocop`, so the ID in the GH repo should be correct now.

Earlopain:
Fascinating, this does seem to have fixed the badge links in the readme. The coverage upload still fails though, is it really the same id still?

bbatsov:
No idea. That was the only such ID I could find in the repo's settings on CodeClimate.

bbatsov:
Might be a good idea to create a separate issue about the code coverage issue, so we can track all investigation in a single place.

PR #12947, opened 2024-05-30
Rearrange CI config

bquorning:
Work in progress.

-----------------

* [ ] Added tests.


bquorning:
cc @Earlopain

Earlopain:
This looks reasonable to me, the cleanup is nice.

--------------

One thing I was asking myself while working on CI was if that many jobs for checking that are even needed. They take up a few minutes during runs and if the the only expectation is that the code is formatted correctly then that doesn't need to be confirmed 14 times. Would the test suite not be enough?

bquorning:
I agree that we use the internal_investigation task as a fix of linting _and_ verification on top of the specs. I don’t know if that extra verification is necessary (did it ever catch anything the specs didn’t?) but hopefully someone else will remember.

bbatsov:
> One thing I was asking myself while working on CI was if that many jobs for checking that are even needed. They take up a few minutes during runs and if the the only expectation is that the code is formatted correctly then that doesn't need to be confirmed 14 times. Would the test suite not be enough? I guess what I'm saying is it either is testing something and should be part of the other file, or it really is just linting and doing that once is ok

Likely on oversight on our part when the CI got migrated in the past. That's now the 3rd CI Platform (Travis, Circle, GHA)...

I agree that things like linting and formatting should be done just once, everything else is just a waste of time.

PR #12950, opened 2024-05-30
[Fix #12949] Use the secret for the coverage reporter id

### LINKED ISSUE #12949: CodeClimate coverage upload not working
Ref #12946

> Something is messed up in CodeClimate as the old links to the repos are there (those from rubocop-hq) and they seem to be active, but their settings can't be changed. Also rubocop/rubocop is there twice for some reason... I've added some CC_TEST_REPORTER_ID from one of them, but I think we'll need to contact CC's support to clean up the legacy stuff and this duplication (when I try to delete one of the two RuboCop repos it blows up).

Sample job: https://github.com/rubocop/rubocop/actions/runs/9302007461/job/25601715895

@bquorning, maybe you have an idea?

I guess, if all fails contact support. The duplication above gives me an inkling that this just won't work regardless of what we try.

Earlopain:
Fixes #12949, this uses the configured secret (https://github.com/rubocop/rubocop/issues/12949#issuecomment-2139496033) instead of the hardcoded value which probably hasn't worked for quite some time.

As secrets are not available to forks the job will be skipped for this PR but it should hopefully work once merged. I tested both scenarios in my fork and that went ok. cc @bquorning

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -70,7 +70,8 @@ jobs:
           pattern: coverage-*
       - uses: paambaati/codeclimate-action@v6
         env:
-          CC_TEST_REPORTER_ID: a11b66bfbb1acdf220d5cb317b2e945a986fd85adebe29a76d411ad6d74ec31f
+          CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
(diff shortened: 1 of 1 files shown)


bbatsov:
Ah, I didn't see the hardcoded value at all. 🤦

PR #12958, opened 2024-06-03
[Fix #12957] Mark `Style/SuperArguments` as unsafe

### LINKED ISSUE #12957: Style/SuperArguments incompatible with define_method
## Expected behavior

The recently inntroduced `Style/SuperArguments` should "just work". I assumed that, by just autocorrecting [the errors from this test suite](https://gitlab.com/os85/rodauth-oauth/-/jobs/6944724946), the errors would go away. 

## Actual behavior

Replacing `super` calls broke the code. `rodauth` defines auth methods using `define_method`, which does not support implicit `super` calls:

```
RodauthOauthDynamicClientRegistrationTest#test_oauth_dynamic_client_response_types:
RuntimeError: implicit argument passing of super from method defined by define_method() is not supported. Specify all arguments explicitly.
```

clone the rodauth-oauth repository, enable the cop, run `rubocop -a lib test` and run `bundle exec rake test` to find the offending tests.

## RuboCop version

`rubocop-1.64.1`

Earlopain:
Here is a somewhat convoluted case in which autocomplete breaks the code:

```rb
class A
  def method(a)
  end

  def other_method(a)
    puts a
  end
end

class B < A
  def method(a)
    context_change { super(a) }
    other_method
  end

  def context_change(&)
    define_singleton_method(:other_method, &)
  end
end

B.new.method('foo')
#=> foo
```

With autocorrect:

```rb
class A
  def method(a)
  end

  def other_method(a)
    puts a
  end
end

class B < A
  def method(a)
    context_change { super }
    other_method
  end

  def context_change(&)
    define_singleton_method(:other_method, &)
  end
end

B.new.method('foo')
#=> test.rb:12:in 'block in method': implicit argument passing of super from method defined by define_method() is not supported. Specify all arguments explicitly. (RuntimeError)
```

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/change_mark_super_arguments_as_unsafe.md
+++ b/changelog/change_mark_super_arguments_as_unsafe.md
@@ -1,0 +1,1 @@
+* [#12957](https://github.com/rubocop/rubocop/issues/12957): Mark `Style/SuperArguments` as unsafe. ([@earlopain][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -5427,8 +5427,10 @@ Style/StructInheritance:
 
 Style/SuperArguments:
   Description: 'Call `super` without arguments and parentheses when the signature is identical.'
+  Safe: false
   Enabled: pending
   VersionAdded: '1.64'
+  VersionChanged: <<next>>
 
 Style/SuperWithArgsParentheses:
   Description: 'Use parentheses for `super` with arguments.'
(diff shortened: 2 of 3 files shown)


koic:
I'm going to close this to let it be handled safely. I'll take up the issue later. Thank you.

Earlopain:
I'm interested to see how you're going to handle this. I don't see a way unless you're just going to stop at all blocks

PR #12959, opened 2024-06-04
[Fix #12957] Fix false positives for `Style/SuperArguments`

### LINKED ISSUE #12957: Style/SuperArguments incompatible with define_method

koic:
Fixes #12957.

This PR fixes false positives for `Style/SuperArguments` when calling super in a block.

Using zero arity `super` within a `define_method` block results in `RuntimeError`:

```ruby
def m
  define_method(:foo) { super() } # => OK
end

def m
  define_method(:foo) { super }   # => RuntimeError
end
```

Furthermore, any arguments accompanied by a block may potentially be delegating to `define_method`, therefore, `super` used within these blocks will be allowed. This approach might result in false negatives, yet ensuring safe detection takes precedence.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positives_for_style_super_arguments.md
+++ b/changelog/fix_false_positives_for_style_super_arguments.md
@@ -1,0 +1,1 @@
+* [#12957](https://github.com/rubocop/rubocop/issues/12957): Fix false positives for `Style/SuperArguments` when calling super in a block. ([@koic][])
--- a/lib/rubocop/cop/style/super_arguments.rb
+++ b/lib/rubocop/cop/style/super_arguments.rb
@@ -3,8 +3,25 @@
 module RuboCop
   module Cop
     module Style
-      # Checks for redundant argument forwarding when calling super
-      # with arguments identical to the method definition.
+      # Checks for redundant argument forwarding when calling super with arguments identical to
+      # the method definition.
+      #
+      # Using zero arity `super` within a `define_method` block results in `RuntimeError`:
+      #
+      # [source,ruby]
+      # ----
+      # def m
+      #   define_method(:foo) { super() } # => OK
+      # end
+      #
+      # def m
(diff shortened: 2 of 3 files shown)


bbatsov:
The proposed changes look good to me. 

//cc @Earlopain

Earlopain:
Either I'm overestimating the amound of times `super` is called within a block or underestimating the times when a block with `super` gets passed to one of these two methods. Is this not what Unsafe is for?

koic:
Fundamentally, unsafe should only be chosen as a last resort. Opting for unsafe readily can lead to a gradual disappearance of safe cops in static analysis. This issue is a relatively minor case regarding style consistency, and I think that being able to safely detect and correct it adds value to the user's experience.

Earlopain:
Fair, I guess. Would a configurationfor this be accepted? I can put that in after this PR is merged

koic:
I have also considered the configuration, but given that it compromises safety, I think not adopting such a configuration would maintain simplicity.

bbatsov:
We can discuss a configuration option down the road if there's demand for it, but I think for now the proposed solution is fine.

PR #12964, opened 2024-06-04
Ci lint once / actually run `internal_investigation`

Earlopain:
As per https://github.com/rubocop/rubocop/pull/12947#issuecomment-2147256882

This also fixes an issue with that PR where rubocop would only lint itself when a yaml file got changed. I only noticed this after opening this file, the diff unfortunately didn't make that very visible. I've copied the conditions for the main workflow file which means yaml will be linted every time now, that seems fine. cc @bquorning

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -9,33 +9,14 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
+
 jobs:
   lint-ruby:
-    name: Ruby - ${{ matrix.os }} ${{ matrix.ruby }}
-    runs-on: ${{ matrix.os }}-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu, windows]
-        ruby: ['2.7', '3.0', '3.1', '3.2', '3.3', 'head']
-        include:
-          - os: ubuntu
-            ruby: jruby-9.4
-          - os: windows
-            ruby: mingw
-        exclude:
-          - os: windows
-            ruby: head
     steps:
-      - name: Windows Specific
-        if: matrix.os == 'windows'
-        run: |
-          # Work around `Layout/EndOfLine: Carriage return character detected`
-          git config --system core.autocrlf false
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ matrix.ruby }}
+          ruby-version: '3.3'
           bundler-cache: true
       - name: internal_investigation
         run: bundle exec rake internal_investigation
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: '3.3'
+          ruby-version: ruby # Latest stable CRuby version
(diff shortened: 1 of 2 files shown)


Earlopain:
Sigh, this probably needs merged before the changes to `on` are respected. Next time I'll just let someone else handle this

.github/workflows/linting.yml
```diff
@@ -1,40 +1,22 @@
...
-      - name: Windows Specific
-        if: matrix.os == 'windows'
-        run: |
-          # Work around `Layout/EndOfLine: Carriage return character detected`
-          git config --system core.autocrlf false
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ matrix.ruby }}
+          ruby-version: '3.3'
```
koic, line 19: Can you add a comment to make the intention clearer?

```suggestion
          ruby-version: '3.3' # Use the latest Ruby version.
```

bbatsov: Or maybe this can be declared as some variable (e.g. `latest_ruby_version`)

bquorning, line 19: https://github.com/ruby/setup-ruby/blob/master/README.md#supported-version-syntax mentions that you can use
> - engine only like `ruby` and `truffleruby`, uses the latest stable release of that implementation
```suggestion
          ruby-version: ruby
```

.github/workflows/linting.yml
```diff
@@ -1,40 +1,22 @@
 name: Linting
 on:
+  push:
+    branches:
+      - master
   pull_request:
-    paths:
-      - '**/*.yaml'
-      - '**/*.yml'
```
bquorning, line 6: Sorry about that, I should have noticed when I moved the configuration around.

Let me just have another look at the documentation before I approve this PR.

.github/workflows/linting.yml
Earlopain: https://github.com/rubocop/rubocop/pull/12964#discussion_r1625852504

> engine only like ruby and truffleruby, uses the latest stable release of that implementation

`ruby-version: ruby`

So, no need to change this every year. Quite nice I think

Earlopain: Oh, thats nice. I was looking at docs but didn't find this

.github/workflows/linting.yml
bquorning: Ok, so the intention with these two lines was to not run Yamllint unnecessarily. From what I can see, the Yamllint check usually takes about 20 seconds to run, so it’s no big problem to merge the config as-is.

The alternative I was looking into involves using e.g. [the dorny/paths-filter action](https://github.com/dorny/paths-filter) to look at changed paths _inside_ the `lint-yaml` job. But I’m not sure that it will actually be faster than 20 seconds anyway 😅

bquorning (approved):

.github/workflows/linting.yml
Earlopain: Huh, 20s seems quite long for what it does

Earlopain:
😓 here's probably the reason why it didn't run: https://github.com/rubocop/rubocop/actions/runs/9366642349. Quite sad that GitHub doesn't show this here in the PR overview.

I can make a PR in an hour or so for this. Feel free to do so beforehand yourself

bquorning:
Yeah, I also find it hard to validate changes to GitHub Actions configuration before merging. So much so that I have a separate repo that I use just for that 😄

PR #12965, opened 2024-06-04
CI: Fix config for linting.yml

bquorning:
Fixing the configuration for linting.yml on CI.

Currently CI is failing ([e.g. here](https://github.com/rubocop/rubocop/actions/runs/9366635974)) with
```
Invalid workflow file: .github/workflows/linting.yml#L15
The workflow is not valid. .github/workflows/linting.yml (Line: 15, Col: 5): Required property is missing: runs-on
```

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -12,6 +12,8 @@ concurrency:
 
 jobs:
   lint-ruby:
+    name: Ruby
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1


bquorning:
cc @Earlopain

bbatsov (approved):
I had just noticed the syntax issue. The fix looks good to me.

bbatsov:
Hmm, for some reason all jobs seem stuck - waiting for workers to pick them up.

bquorning:
> Hmm, for some reason all jobs seem stuck - waiting for workers to pick them up.

The jobs are slowly being picked up.

I wonder if we are already running into a parallelization limit of GitHub Actions, or maybe GitHub is having a issue?

bbatsov:
Yeah, just noticed this myself. So far, so good.

Next stop - maybe remove the Ascii specs, as I don't remember the last time they found something useful. @koic what do you think about their usefulness?

PR #12967, opened 2024-06-04
CI: Optimize for faster results and lower resource use

bquorning:
Three commits:

1. Split specs running on ubuntu and windows.
   We'll have some duplication, but it allows for some optimization.
2. Reduce some test matrixes
   Reduce the test matrix for ascii_spec from 13 combinations to just four, and for windows specs from six to three.
   Based on discussion in https://github.com/rubocop/rubocop/pull/12967#discussion_r1626082395 and https://github.com/rubocop/rubocop/pull/12965#issuecomment-2147419092.
3. Only run windows specs after ubuntu specs pass.
   We don't _require_ windows specs to pass, and windows specs are very slow. By only running them _after_ ubuntu specs pass, we get the most important results first - and in case of failing specs, we'll save 30 minutes of CI use.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -54,7 +54,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.7', '3.0', '3.1', '3.2', '3.3', 'mingw']
+        ruby:
+          - '2.7'  # Oldest supported version
+          - 'ruby' # Latest stable CRuby version
 
     steps:
       - uses: actions/checkout@v4
@@ -97,15 +99,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, windows]
-        ruby: ['2.7', '3.0', '3.1', '3.2', '3.3', 'head']
-        include:
-          - os: ubuntu
-            ruby: jruby-9.4
-          - os: windows
-            ruby: mingw
-        exclude:
-          - os: windows
-            ruby: head
+        ruby:
+          - '2.7'  # Oldest supported version
+          - 'ruby' # Latest stable CRuby version
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -49,6 +49,7 @@ jobs:
           if-no-files-found: error
 
   spec-windows:
+    needs: spec-ubuntu # Don't spend CI resources on slow Windows specs if CI won't pass anyway.
     name: Spec - windows ${{ matrix.ruby }}
     runs-on: windows-latest
     strategy:
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -15,22 +15,13 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  spec:
-    name: Spec - ${{ matrix.os }} ${{ matrix.ruby }}
-    runs-on: ${{ matrix.os }}-latest
+  spec-ubuntu:
+    name: Spec - ubuntu ${{ matrix.ruby }}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu, windows]
-        ruby: ['2.7', '3.0', '3.1', '3.2', '3.3', 'head']
-        include:
-          - os: ubuntu
-            ruby: jruby-9.4
-          - os: windows
-            ruby: mingw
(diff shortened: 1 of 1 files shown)


.github/workflows/rubocop.yml
```diff
@@ -43,23 +34,46 @@ jobs:
...
-        if: matrix.os == 'ubuntu' && !startsWith(matrix.ruby, 'jruby-')
+        if: ${{ !startsWith(matrix.ruby, 'jruby-') }}
         uses: actions/upload-artifact@v4
         with:
-          name: coverage-${{ matrix.os }}-${{ matrix.ruby }}
+          name: coverage-ubuntu-${{ matrix.ruby }}
           path: coverage/.resultset.json
           if-no-files-found: error
 
+  spec-windows:
```
bbatsov, line 51: I'd add some comments here about the reasoning for running the Windows tests after the Ubuntu tests, so someone won't accidentally change this in the future.

.github/workflows/rubocop.yml
```diff
@@ -43,23 +34,46 @@ jobs:
...
+    needs: spec-ubuntu
+    name: Spec - windows ${{ matrix.ruby }}
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: ['2.7', '3.0', '3.1', '3.2', '3.3', 'mingw']
```
bbatsov, line 58: I'm also wondering if we need to run all the Ruby versions here or we can again just go for the oldest/newest combo. I'm struggling to come up with a scenario where on Windows we'd have failures that are both Windows-specific and specific to a specific Ruby version.

.github/workflows/rubocop.yml
```diff
@@ -84,15 +98,7 @@ jobs:
...
-        ruby: ['2.7', '3.0', '3.1', '3.2', '3.3', 'head']
-        include:
-          - os: ubuntu
-            ruby: jruby-9.4
-          - os: windows
-            ruby: mingw
-        exclude:
-          - os: windows
-            ruby: head
+        ruby: ['2.7', '3.3'] # Oldest and newest supported stable version
```
Earlopain, line 101: Maybe you can also just use `ruby` here? Since it'll be bumped straight away anyways when a new one releases

.github/workflows/rubocop.yml
bquorning: I agree, and I’ve reduced the matrix to just 2.7, 3.3 and mingw.

I am not sure what mingw is and whether we need it.

.github/workflows/rubocop.yml
bquorning: Good idea, I’ll do that.

.github/workflows/rubocop.yml
Earlopain: That seems to be a differently build version of ruby. Just specifying a version pulls the `ucrt` variant. Considering that RuboCop doesn't do C extensions I fail to see the value in testing against that.

bquorning:
> It looks like the few specs in `spec/rubocop/cli*` and `spec/rubocop/server/` make up about 2/3 of the total runtime of our specs. If we could either 1) make them faster or 2) run them less often, our CI would become quite a bit faster.

We could use the `:isolated_environment` tag to separate fast and slow specs:
```
$ bundle exec rspec --tag=isolated_environment
Finished in 1 minute 14.25 seconds (files took 3.12 seconds to load)
974 examples, 0 failures
```
```
$ bundle exec rspec --tag=~isolated_environment
Finished in 22.79 seconds (files took 3.21 seconds to load)
21427 examples, 0 failures, 1 pending
```

Earlopain:
> We could use the :isolated_environment tag to separate fast and slow specs:

Have these two runs be part of the matrix for windows, already getting some gains from splitting it up. Since they run as the last step there are likely a few free runners available: actions has a concurrency of 20 https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

Earlopain:
Windows tests (to my knowledge) didn't run on CircleCI: they were moved from AppVeyor to GitHub Actions a few years ago

.github/workflows/rubocop.yml
bquorning: I will remove `mingw` then.

bbatsov:
Great work!

> Windows tests (to my knowledge) didn't run on CircleCI: they were moved from AppVeyor to GitHub Actions a few years ago

Yep, that's how I remember it as well. For a very long time we didn't have a Windows build at all, btw.

PR #12969, opened 2024-06-05
CI: Run JRuby job outside of spec-ubuntu

Earlopain:
It was actually like that before all the CI restructuring but I moved it into the matrix to speed things up a bit by separating out the individual rake tasks (`ìnternal_investigation`, `spec`, `ascii_spec`) but now having it there is actually a bother since windows specs wait on completion of it (about 4 minutes)

I specified the version as `jruby` instead of `jruby-9.4` to keep the theme of always getting the latest stable version.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.7', '3.0', '3.1', '3.2', '3.3', 'head', 'jruby-9.4']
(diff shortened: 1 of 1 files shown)


Earlopain:
Another option is to just remove that dependency again. There are not nearly as much jobs being generated where this doesn't matter as much anymore, I think

bbatsov:
I think it's fine as you propose - we run the Linux MRI and JRuby jobs in parallel and Windows at the end. Probably we should also look into some profiling of the specs and see if we can't optimize all builds by making the slowest specs faster.

PR #12971, opened 2024-06-06
Implement `NoReturnFromBlock` rule

### LINKED ISSUE #4064: please add a cop that warns if there is a return from within a block
I've seen many times where someone writes code akin to

```
def foo(lst)
   lst.map do |item|
       return 'even' if item % 2 == 0
       'odd'
   end
end
```

instead of foo([0,1]) returning ['even','odd'], it returns 'even', which is unexpected.

Even the official documentation for a rubocop rule demonstrates this

http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Lint/NonLocalExitFromIterator

the return that is commented with ```#allowed ``` within the map would actually fail as a LocalJumpError if it ran because the return doesn't run within a method, it runs from within a block.

clearly the author didn't want to cause LocalJumpError.

```
  rescue_from ValidationError do |e| # non-iteration block with arg
    return message: 'validation error' unless e.errors # allowed
    error_array = e.errors.map do |error| # block with method chain
      return if error.suppress? # warned
      return "#{error.param}: invalid" unless error.message # allowed
      "#{error.param}: #{error.message}"
    end
    message: 'validation error', errors: error_array
  end

```

so can we please add a rule that warns on returns from within blocks?

toddnestor:
Fixes #4064.

This PR adds a `NoReturnFromBlock` rule that warns when `return` is called from inside a block.  This rule is turned off by default so as not to cause any breaking changes.  It can be enabled by setting `AllowReturnFromBlock` to `false` in the config.

This rule is important as a `return` in a block can often cause bugs.

-----------------

* [X] Added tests.


lib/rubocop/cop/lint/no_return_from_block.rb
```diff
@@ -0,0 +1,47 @@
...
+
+          node.each_node(:return) do |return_node|
+            add_offense(return_node)
+          end
+        end
+
+        private
+
+        def allow_return_from_block?
+          cop_config['AllowReturnFromBlock']
```
bbatsov, line 42: I don't see the point of this config, as it's essentially the same as disabling the cop. Also - what about procs? They are essentially blocks for all practical purposes.

koic:
The bad cases noted in the sample of this PR can already be detected by `Lint/NonLocalExitFromIterator`, so I have concerns about the necessity of providing this new cop.

```ruby
      # @example
      #   # bad
      #   items.each do |item|
      #     return if item.nil?
      #     puts item.some_attribute
      #   end
```

https://docs.rubocop.org/rubocop/1.64/cops_lint.html#lintnonlocalexitfromiterator

bbatsov:
Ah, great point! I had forgotten about it.

toddnestor:
> The bad case described in the example of this PR can already be detected by `Lint/NonLocalExitFromIterator`, so I have concerns about the necessity of providing this new cop.
> 
> ```ruby
>       # @example
>       #   # bad
>       #   items.each do |item|
>       #     return if item.nil?
>       #     puts item.some_attribute
>       #   end
> ```
> 
> https://docs.rubocop.org/rubocop/1.64/cops_lint.html#lintnonlocalexitfromiterator

Ah I wasn't aware of this one, let me test that rule out and see if why it didn't catch the bugs we had recently, I'll close this PR if it is indeed redundant.

dimakura:
> The bad case described in the example of this PR can already be detected by `Lint/NonLocalExitFromIterator`, so I have concerns about the necessity of providing this new cop.
> 
> ```ruby
>       # @example
>       #   # bad
>       #   items.each do |item|
>       #     return if item.nil?
>       #     puts item.some_attribute
>       #   end
> ```
> 
> https://docs.rubocop.org/rubocop/1.64/cops_lint.html#lintnonlocalexitfromiterator

@koic how about return with a value, which is allowed by `Lint/NonLocalExitFromIterator`?

```rb
items.each do |item|
  return "empty" if item.nil? # Lint/NonLocalExitFromIterator doesn't complain about it
  # ...
end
```

stevetsanders:
Agreed, I'm looking for this as well to be merged. Same scenario, production bug caused by `return` in `do/end` block

Bertg:
Not stale dear bot...

@bbatsov what should we do to get this merged in? 👆 there is quite a bit of support for it above.

PR #12972, opened 2024-06-07
[Fix #12970] Add `CountModifierForms` option to `Metrics/BlockNesting`

### LINKED ISSUE #12970: Add configuration for Metrics/BlockNesting to ignore inline if statements
In my understanding, the purpose of Metrics/BlockNesting is to avoid "indentation hell" with conditionals and loops.
Rubocop is currently flagging code like this:
```
x = nil
y = true
z = false

if x
  if y
    if z
      x = 3 if true # <<<< flagged as error
    end
  end
end
```

It would be nice to be able to exclude trailing conditionals from this metrics cop.
Something similar to `AllowIfModifier` for `Style/IfInsideElse` cop.

Observed in rubocop v1.31

koic:
Resolves #12970.

This PR adds `CountModifierForms` option to `Metrics/BlockNesting`.

Similar to the existing `CountBlocks`, the default is `false`. This means that existing modifier form code will not be detected by default, but this seems reasonable as part of the block nesting functionality. If user require the previous behavior, user can achieve it by setting `CountBlocks` to `true`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_add_count_blocks_to_metrics_block_nesting.md
+++ b/changelog/change_add_count_blocks_to_metrics_block_nesting.md
@@ -1,0 +1,1 @@
+* [#12970](https://github.com/rubocop/rubocop/issues/12970): Add `CountModifierForms` option to `Metrics/BlockNesting` and set it to `false` by default. ([@koic][])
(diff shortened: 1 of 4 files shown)


bbatsov:
Fine by me. Thanks!

PR #12974, opened 2024-06-08
[Fix #11957] Add cop `Lint/ExplicitPublicAccessModifier`

### LINKED ISSUE #11957: Flag explicit `public` access modifier usage

```ruby
class Foo
  def foo
  end

  private

  def bar
  end

  public # This might be surprising to those who got used to add private methods to the bottom of the class

  def baz
  end

  # Any method added here will be public
end
```

Add a cop that flags the explicit usage of `public`

[`Layout/ClassStructure`](https://docs.rubocop.org/rubocop/cops_layout.html#layoutclassstructure) may or may not cover this, it's [not covered in its specs](https://github.com/rubocop/rubocop/blob/35d5286ae52ea89826732b6a05df23625a6432eb/spec/rubocop/cop/layout/class_structure_spec.rb#L428). Some may prefer to disable `Layout/ClassStructure`.

The purpose of this cop would be to restrict the confusing usage completely, not just rearrange the methods.

yoshiyoshiharu:
Fix #11957 

This PR adds a `Lint/ExplicitPublicAccessModifier`.
It warns explicit usage of `public` which might be surprising for developers accustomed to placing private methods at the bottom of a class.

-----------------

* [x] Added tests.


koic:
The bad case described in the example of this PR can already be detected by `Lint/UselessAccessModifier`, so I have concerns about the necessity of providing this new cop.

```ruby
      #   class Test
      #     public
      #
      #     def test; end
      #   end
```

https://docs.rubocop.org/rubocop/1.64/cops_lint.html#lintuselessaccessmodifier

yoshiyoshiharu:
@koic 
As mentioned in #11957, using the `public` access modifier below the `private` access modifier doesn't be warned.
`ClassStructure` may detect this; However,  if it is disabled (which is default setting), the following example will not be detected
```rb
      #   # bad
      #   class Test
      #     private
      #
      #     def private_method;
      #
      #     public
      #
      #     def public_method; end
      #   end
```

PR #12984, opened 2024-06-12
[Fix #12983] Fix false positives for `Style/SendWithLiteralMethodName`

### LINKED ISSUE #12983: Additional caution for new cop: Style/SendWithLiteralMethodName

I believe some additional caution should be urged for the new cop: `Style/SendWithLiteralMethodName` https://github.com/rubocop/rubocop/pull/12850

Presently the cop is marked unsafe for "receiver" reasons (which I don't completely understand).
However, there are other cases where simply converting from `send` (or `public_send`) to actual method invocation is not safe.

Writer methods in particular behave differently when invoked directly than with `send`.

```rb
class WeirdRuby
  attr_reader :wat
  def wat=(w)
    @wat = w
    42
  end
end
```

```irb
>> fun = WeirdRuby.new
=> #<WeirdRuby:0x000000011d9fa8d0>
>> fun.wat = 7
=> 7
>> fun.public_send(:wat=, 7)
=> 42
```

The exact same method, invoked via send vs invoked directly results in completely different return value. (IIRC it's because the ruby compiler special-cases `=` syntax to ensure the expression evaluates to the assigned value, even if the writer method returns something else. However, with `send` invocation, no such protection is in place.

I'm not really sure. Is it possible (or reasonable) for the cop to detect writer methods and refuse autocorrect them? Or should that be an additional option? Or should this just be an extra example in the docs? Or a test case in rubocop's test suite?

I would be happy to provide a PR depending on the approach desired. But I think _something_ should be done since this could be a huge footgun. (Not huge in frequency, but in debugging. No one would ever think to confirm that send vs direct-method could ever behave differently.)

## Additional context

This case may seem obscure (and it very likely is), but considering that if ever a writer method is indeed implemented directly instead of via the `attr_writer` macro, the odds that the return value differs from the "assigned" value go up greatly. (The only reason to override a writer is to do something odd anyway. And it's super common to have an "accidental" return value during a side-effect operation by not being careful of the last expression in the method. I've seen custom writers like this act as setters that are calling off to a service somehow and return nil or status codes or things completely unrelated to "assignment" value.)

koic:
Fixes #12983.

This PR fixes false positives for `Style/SendWithLiteralMethodName` using `send` with writer method name.

Writer methods with names ending in `=` are always permitted because their behavior differs as follows:

```ruby
def foo=(foo)
  @foo = foo
  42
end

foo(1)         # => 1
send(:foo=, 1) # => 42
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positives_for_style_send_with_literal_method_name.md
+++ b/changelog/fix_false_positives_for_style_send_with_literal_method_name.md
@@ -1,0 +1,1 @@
+* [#12983](https://github.com/rubocop/rubocop/issues/12983): Fix false positives for `Style/SendWithLiteralMethodName` using `send` with writer method name. ([@koic][])
--- a/lib/rubocop/cop/style/send_with_literal_method_name.rb
+++ b/lib/rubocop/cop/style/send_with_literal_method_name.rb
@@ -7,6 +7,20 @@ module RuboCop
       # Since the `send` method can be used to call private methods, by default,
       # only the `public_send` method is detected.
       #
+      # NOTE: Writer methods with names ending in `=` are always permitted because their
+      # behavior differs as follows:
+      #
+      # [source,ruby]
+      # ----
+      # def foo=(foo)
+      #   @foo = foo
+      #   42
+      # end
+      #
+      # foo(1)         # => 1
+      # send(:foo=, 1) # => 42
+      # ----
+      #
       # @safety
       #   This cop is not safe because it can incorrectly detect based on the receiver.
       #   Additionally, when `AllowSend` is set to `true`, it cannot determine whether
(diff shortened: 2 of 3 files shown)


Earlopain:
I don't think the example code is correct.

`send(:foo=, 1)` would get corrected to `foo=(1)` which is a variable declaration! So that method doesn't even get called anymore and in that sense it makes sense for the return value to be 1.

koic:
This indicates a method call on `self`, not a variable declaration:

```console
$ ruby -e "send(:foo=, 1)"
-e:1:in `<main>': undefined method `foo=' for main (NoMethodError)

send(:foo=, 1)
^^^^

$ ruby -e "def foo=(foo); 42 end; p send(:foo=, 1)"
42
```

Earlopain:
Yes, but the previous behaviour of autocorrect would have produced `foo=(1)` which is not. The example code isn't executable because `foo` isn't defined.

koic:
Ah, I get it. That is a mistake in the example. It has been fixed in https://github.com/rubocop/rubocop/commit/967fe88135188c9d74494cbd4355180c140c4a1a.

Earlopain:
Yes, that looks good. Thank!

PR #12991, opened 2024-06-13
[Fix #12989] Fix an error for the `inherit_gem` config when the Gemfile contains an uninstalled git gem

### LINKED ISSUE #12989: improve compatibility with pre-commit when using inherit_gem in a project with git dependencies

I would like to use rubocop with pre-commit (which seems to be supported, since you ship a `.pre-commit-hooks.yaml`). Unfortunately, rubocop attempts to load extensions by calling `Bundler.load`; this attempts to load the gems for the calling application (as opposed to the gems in the pre-commit / rubocop isolated environment), which is not desirable -- at best, it will pollute global gem state by installing all of the dependencies from the containing application's gemfile; if any of the dependencies are git dependencies, it'll just fail

I've been told that this has been addressed before, but whatever was done doesn't seem to be comprehensive.

My thought was that rubocop could add a way to signal that we don't want to load from bundler (perhaps a new command-line flag like `--no-i-really-dont-want-to-use-bundler`) and pass it in the `.pre-commit-hooks.yaml`. When that flag is passed, rubocop would skip the `Bundler.load` call entirely in `Rubocop::ConfigLoaderResolver#gem_config_path`.

I sent `pre-commit` a patch (https://github.com/pre-commit/pre-commit/pull/3232) to disable loading the Gemfile from their side, but they are adamantly opposed to making that change. Their stance is "a linter blindly trying to access an unrelated environment is a poorly behaving linter" and that this is a rubocop bug. 

## Additional context

https://github.com/Roguelazer/rubocop-pre-commit-breakage-demo is a repository which has minimal reproduction steps to show that this is broken.

Earlopain:
Fixes #12989

I also move the bundler isolation into the shared context and add a test for it

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_error_for_inherit_gem_git_gem.md
+++ b/changelog/fix_error_for_inherit_gem_git_gem.md
@@ -1,0 +1,1 @@
+* [#12989](https://github.com/rubocop/rubocop/issues/12989): Fix an error for the `inherit_gem` config when the Gemfile contains an uninstalled git gem. ([@earlopain][])
--- a/lib/rubocop/config_loader_resolver.rb
+++ b/lib/rubocop/config_loader_resolver.rb
@@ -272,6 +272,8 @@ module RuboCop
           gem_path = gem.full_gem_path if gem
         rescue Bundler::GemfileNotFound
           # No Gemfile found. Bundler may be loaded manually
+        rescue Bundler::GitError
+          # The Gemfile exists but contains an uninstalled git source
         end
       end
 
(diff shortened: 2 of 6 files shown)


Earlopain:
Let's wait for feedback from OP in the issue (or here). I don't know how to test this in the context of that tool

Earlopain:
Seems fine: https://github.com/rubocop/rubocop/issues/12989#issuecomment-2166064469

koic:
Thanks!

PR #13001, opened 2024-06-17
Reconfigure `RSpec/PredicateMatcher` for better expectation failure messages

Earlopain:
This is something that has bothered me for a while. There is an rspec cop for this to enforce better failure messages. I changed the config an let RuboCop autocorrect.

Changed expectations would previously output messages like this:
```
     Failure/Error: expect(options_keys.include?(:aautocorrect)).to be(true)

       expected true
            got false
```

The new expectations print the actual vs expected value:
```
     Failure/Error: expect(options_keys).to include(:aautocorrect)

       expected [:autocorrect_all, :autocorrect] to include :aautocorrect
       Diff:
       @@ -1 +1 @@
       -[:aautocorrect]
       +[:autocorrect_all, :autocorrect]
```

The same improvement also applies to something like `expect($stderr.string.include?("whatever).to be(true)` or `expect(foo.nil?).to be(true)` (and probably more)

-----------------

* [ ] Added tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -142,7 +142,8 @@ RSpec:
       - expect_offense
 
 RSpec/PredicateMatcher:
-  EnforcedStyle: explicit
+  # Also consider `be(true)` and `be(false)` as offenses, instead of only `be_truthy` and `be_falsy`
+  Strict: false
 
 RSpec/SpecFilePathFormat:
   CustomTransform:
--- a/lib/rubocop/cop/util.rb
+++ b/lib/rubocop/cop/util.rb
@@ -173,7 +173,9 @@ module RuboCop
       def same_line?(node1, node2)
         line1 = line(node1)
         line2 = line(node2)
-        line1 && line2 && line1 == line2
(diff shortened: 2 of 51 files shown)


.rubocop.yml
```diff
@@ -142,7 +142,7 @@ RSpec:
       - expect_offense
 
 RSpec/PredicateMatcher:
-  EnforcedStyle: explicit
+  Strict: false
```
bbatsov, line 145: I'd add a comment explaining the need for this.

Earlopain:
Had to change some application code here: https://github.com/rubocop/rubocop/pull/13001/files#diff-fd0332e5689d0ab642df39f706a915663f5d8764d4fa4442d36a291b16de70fb

For some reason the test for that would fail only with rspec 4 if this returns nil instead of false.

.rubocop.yml
Earlopain: Added:
> Also consider `be(true)` and `be(false)` as offenses, instead of only `be_truthy` and `be_falsy`

PR #13004, opened 2024-06-18
[Fix #12557] Make server mode aware of auto-restart for `bundle update`

### LINKED ISSUE #12557: Wrong version of extensions loaded with 1.59.0?
## Expected behavior

With same exact Gemfile and Gemfile.lock, with updated rubocop extensions, rubocop is loading the wrong version dependencies. (see version info below)

This is true even if the other versions are fully uninstalled (gem uninstall rubocop-performance and bundle exec gem uninstall rubocop-performance before bundle install again)

## Actual behavior

Gemfile has this in it:

```
  gem "rubocop", "~> 1.59.0"
  gem "rubocop-capybara", "~> 2.19.0"
  gem "rubocop-graphql", "~> 1.4.0"
  gem "rubocop-minitest", "~> 0.34.1"
  gem "rubocop-performance", "~> 1.20.0"
  gem "rubocop-rails", "~> 2.23.0"
  gem "rubocop-rake", "~> 0.6.0"
 ```

and this in .rubocop.yml:

```
require:
  - rubocop-rails
  - rubocop-minitest
  - rubocop-performance
  - rubocop-rake
  - rubocop-graphql
  - rubocop-capybara
```

with above gemfile, have multiple people run:

```
bundle install
bundle exec rubocop -V
```

## RuboCop version

This is what it should be per Gemfile.lock and there on a number of our machines:

```
1.59.0 (using Parser 3.2.2.4, rubocop-ast 1.30.0, running on ruby 3.0.6) [arm64-darwin22]
  - rubocop-capybara 2.19.0
  - rubocop-graphql 1.4.0
  - rubocop-minitest 0.34.1
  - rubocop-performance 1.20.0
  - rubocop-rails 2.23.0
  - rubocop-rake 0.6.0
 ```

But others are seeing:

```
1.59.0 (using Parser 3.2.2.4, rubocop-ast 1.30.0, running on ruby 3.0.6) +server [arm64-darwin22]
  - rubocop-capybara 2.19.0
  - rubocop-graphql 1.4.0
  - rubocop-minitest 0.33.0
  - rubocop-performance 1.19.1
  - rubocop-rails 2.22.2
  - rubocop-rake 0.6.0
```

You can further see this in debug mode:

When working, it states:

```
For /Users/REDACTED/REDACTED/app: configuration from /Users/REDACTED/REDACTED/app/.rubocop.yml
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-rails-2.23.0/config/default.yml
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-rails-2.23.0/config/default.yml
Default configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-1.59.0/config/default.yml
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-minitest-0.34.1/config/default.yml
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-minitest-0.34.1/config/default.yml
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-performance-1.20.0/config/default.yml
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-performance-1.20.0/config/default.yml
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-rake-0.6.0/config/default.yml
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-rake-0.6.0/config/default.yml
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-graphql-1.4.0/config/default.yml
configuration from #{path}
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-capybara-2.19.0/config/default.yml
configuration from /Users/REDACTED/.rbenv/versions/3.0.6/lib/ruby/gems/3.0.0/gems/rubocop-capybara-2.19.0/lib/../config/default.yml
```

But when not, it only has:

```
For /Users/REDACTED/REDACTED/app: configuration from /Users/REDACTED/REDACTED/app/.rubocop.yml
```

koic:
Resolves #12557.

Currently, the server mode only auto-restarts due to incompatible versions of RuboCop core. This PR will enable auto-restart when the SHA1 value of Gemfile.lock becomes incompatible. If Gemfile.lock is missing, it will continue to refer to the RuboCop core version as before.

The advantage is that it can cover cases of version incompatibility caused by `bundle update`, which will likely cover many scenarios like those in #12557.

The disadvantage is that the server mode may become slightly slower. However, it is unlikely to be slow enough for users to notice. The trade-off is considered acceptable.

In the future, it might be possible to detect changes in rubocop configuration as well, but for now, I'd like to see how things go with this change.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/server/cache.rb
+++ b/lib/rubocop/server/cache.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'digest'
 require 'pathname'
 require_relative '../cache_config'
 require_relative '../config_finder'
@@ -19,6 +20,7 @@ module RuboCop
     # @api private
     class Cache
       GEMFILE_NAMES = %w[Gemfile gems.rb].freeze
+      LOCKFILE_NAMES = %w[Gemfile.lock gems.locked].freeze
 
       class << self
         attr_accessor :cache_root_path
@@ -39,6 +41,14 @@ module RuboCop
 
         def project_dir_cache_key
           @project_dir_cache_key ||= project_dir[1..].tr('/', '+')
+        end
+
+        def restart_key
+          lockfile_path = LOCKFILE_NAMES.map do |lockfile_name|
+            Pathname(project_dir).join(lockfile_name)
+          end.find(&:exist?)
+
+          Digest::SHA1.hexdigest(lockfile_path&.read || RuboCop::Version::STRING)
         end
 
         def dir
--- a/changelog/change_make_server_mode_aware-of_auto_restart_for_bundle_update.md
+++ b/changelog/change_make_server_mode_aware-of_auto_restart_for_bundle_update.md
@@ -1,0 +1,1 @@
+* [#12557](https://github.com/rubocop/rubocop/issues/12557): Make server mode aware of auto-restart for `bundle update`. ([@koic][])
--- a/docs/modules/ROOT/pages/usage/server.adoc
+++ b/docs/modules/ROOT/pages/usage/server.adoc
@@ -49,7 +49,7 @@ user             16060   0.0  0.0  5078568   2264   ??  S     7:54AM   0:00.00 r
 user             16337   0.0  0.0  5331560   2396   ??  S    23:51PM   0:00.00 rubocop --server /Users/user/src/github.com/rubocop/rubocop-rails
 ```
 
-When you update and run `rubocop`, the server process will restart automatically.
+When you `bundle update` and run `rubocop`, the server process will restart automatically.
 
 ```console
 $ rubocop --server
--- a/lib/rubocop/server/client_command/exec.rb
+++ b/lib/rubocop/server/client_command/exec.rb
@@ -41,7 +41,7 @@ module RuboCop
         end
 
         def incompatible_version?
-          Cache.version_path.read != RuboCop::Version::STRING
+          Cache.version_path.read != Cache.restart_key
         end
 
         def stderr
--- a/lib/rubocop/server/client_command/start.rb
+++ b/lib/rubocop/server/client_command/start.rb
@@ -34,7 +34,7 @@ module RuboCop
               exit 0
             end
 
-            Cache.write_version_file(RuboCop::Version::STRING)
+            Cache.write_version_file(Cache.restart_key)
 
             host = ENV.fetch('RUBOCOP_SERVER_HOST', '127.0.0.1')
             port = ENV.fetch('RUBOCOP_SERVER_PORT', 0)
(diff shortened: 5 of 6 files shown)


Earlopain:
Does this actually consider `Gemfile.lock`? From looking at the code I'd say that file is not used in that check.

koic:
Ah, indeed. I've updated this PR.

lib/rubocop/server/cache.rb
```diff
@@ -41,6 +43,14 @@ def project_dir_cache_key
           @project_dir_cache_key ||= project_dir[1..].tr('/', '+')
         end
 
+        def restart_key
+          lockfile_path = LOCKFILE_NAMES.map do |lockfile_name|
+            Pathname(project_dir).join(lockfile_name)
+          end.find(&:exist?)
+
+          Digest::SHA1.hexdigest(lockfile_path&.read || RuboCop::Version::STRING)
```
pirj, line 51: Does it make sense to `read.grep(/rubocop/)` to avoid restarting on unrelated updates?

pirj (approved):
Fantastic, thank you! 🙏

lib/rubocop/server/cache.rb
Earlopain: Unrelated is a bit fuzzy here. For example, if I conservatively update the `parser` gem because a new ruby version released I'd still want it to restart rubocop.

Personally I think its safer to keep it like that, having it run more closely to how it would do without server mode.

koic: Yes. It is prioritize that the environment can be aligned even when updating dependent gems that don't match `/rubocop/`, to prevent user confusion. Especially since it is not possible to know in advance which gems 3rd party gem depends on. Also, since `bundle update` occurs less frequently than `rubocop` is run, it has been decided that updating the entire Gemfile.lock to SHA1 is a trade-off that makes sense.

Issue #11786, opened 2023-04-11
Possible `Lint/UselessAssignment` false negative


vlad-pisanov:
Consider:

```ruby
unused = 1
arr.each { unused = 2 }
```

Either line in isolation triggers `Lint/UselessAssignment`, but when taken together, no offense is registered.

In contrast, this triggers `Lint/UselessAssignment` correctly:
```ruby
unused = 1   # 👮‍♂️ Useless assignment to variable - unused 
if cond
  unused = 2 # 👮‍♂️ Useless assignment to variable - unused
end
```

### Rubocop version
```
1.48.1 (using Parser 3.2.1.1, rubocop-ast 1.27.0, running on ruby 2.6.9) [x86_64-darwin21]
  - rubocop-minitest 0.29.0
  - rubocop-performance 1.16.0
  - rubocop-rails 2.18.0
```

bbatsov:
Reopened — still reproduces on master (1.88): `unused = 1; arr.each { unused = 2 }` registers no offense, even though `unused = 1` on its own is flagged. The dead store is missed once the variable is reassigned inside a block.

PR #13020, opened 2024-06-25
Reassignment in a block shouldn't silence UselessAssignment

casperisfine:
```ruby
foo = 1
1.times do
  foo = 2
end
```

The above code does trigger Ruby parser's warning:

```
$ ruby -c -W /tmp/foo.rb
/tmp/foo.rb:1: warning: assigned but unused variable - foo
Syntax OK
```

I found quite a bit of dead code in our app when looking at Ruby warnings that rubocop didn't find, but enforcing this with rubocop would be preferable.

I'd love to also submit a fix, but I couldn't figure out how to fix this without breaking many other tests :/

### DIFF
--- a/spec/rubocop/cop/lint/useless_assignment_spec.rb
+++ b/spec/rubocop/cop/lint/useless_assignment_spec.rb
@@ -415,10 +415,11 @@ RSpec.describe RuboCop::Cop::Lint::UselessAssignment, :config do
   end
 
   context 'when a referenced variable is reassigned in a block' do
-    it 'accepts' do
-      expect_no_offenses(<<~RUBY)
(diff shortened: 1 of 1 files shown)


Earlopain:
Yeah, it doesn't consider assignments in blocks. Looks pretty much like https://github.com/rubocop/rubocop/issues/11786. Would be nice if it did that but the existing specs that then fail do make sense.

Try out https://github.com/rubocop/rubocop-rubycw? I never used it myself but that seems pretty near to what you are looking for:

```
test.rb:1:1: C: Rubycw/Rubycw: assigned but unused variable - foo
foo = 1
```

PR #13028, opened 2024-07-04
[Fix #13022] Handle relative path handling in TargetFinder

### LINKED ISSUE #13022: Behavior difference of `rubocop` vs `rubocop .`
In our project, we noticed that `rubocop .` took 5 times longer to execute than `rubocop`, even though it's reasonable to expect that both should behave the same.

We traced the difference to the `AllCops/Exclude` directive. Unlike `rubocop`, `rubocop .` fails to prune excluded directories in `TargetFinder#wanted_dir_patterns` early on, and ends up traversing ALL sub-directories in the current folder:

This guard clause never gets matched with `rubocop .`
https://github.com/rubocop/rubocop/blob/2b8ef67b3287bf1346b16d5551bb466238f42486/lib/rubocop/target_finder.rb#L91

because:

`rubocop` exclude_pattern is `"/Users/main/project/{.git/**/*,tmp/**/*}"`, while
`rubocop .` exclude_pattern is the invalid `"./{/Users/main/project/.git/**/*,/Users/main/project/tmp/**/*}"`

vlmonk:
- Address issue where combined_exclude_glob_patterns produced incorrect patterns for relative paths.
- Convert base_dir to an absolute path to handle relative paths correctly.

Resolves: 13022

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/target_finder.rb
+++ b/lib/rubocop/target_finder.rb
@@ -105,6 +105,9 @@ module RuboCop
 
     def combined_exclude_glob_patterns(base_dir)
       exclude = @config_store.for(base_dir).for_all_cops['Exclude'] || []
+      # base_dir can be relative path, so convert it to absoute here.
+      # see issue https://github.com/rubocop/rubocop/issues/13022
+      base_dir = File.expand_path(base_dir)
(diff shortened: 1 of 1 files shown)


koic:
I did some investigation a few days ago, and there is a possibility that this implementation might be insufficient. Please add test code that reproduces the issue and ensure it passes.

PR #13030, opened 2024-07-05
Add new `Gemspec/AddRuntimeDependency` cop

koic:
Follow up https://github.com/rubygems/rubygems/issues/7799#issuecomment-2192720316.

Prefer `add_dependency` over `add_runtime_dependency` as the latter is considered soft-deprecated.

```ruby
# bad
  spec.add_runtime_dependency('rubocop')
end

# good
  spec.add_dependency('rubocop')
end
```

https://github.com/rubocop/ruby-style-guide/pull/943 has been opened for the Style Guide.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -262,6 +262,15 @@ Bundler/OrderedGems:
 
 #################### Gemspec ###############################
 
+Gemspec/AddRuntimeDependency:
+  Description: 'Prefer `add_dependency` over `add_runtime_dependency`.'
+  StyleGuide: '#add_dependency_vs_add_runtime_dependency'
+  Reference: https://github.com/rubygems/rubygems/issues/7799#issuecomment-2192720316
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  Include:
+    - '**/*.gemspec'
+
 Gemspec/DependencyVersion:
   Description: 'Requires or forbids specifying gem dependency versions.'
   Enabled: false
--- a/lib/rubocop/cop/gemspec/add_runtime_dependency.rb
+++ b/lib/rubocop/cop/gemspec/add_runtime_dependency.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Gemspec
+      # Prefer `add_dependency` over `add_runtime_dependency` as the latter is
+      # considered soft-deprecated.
+      #
+      # @example
+      #
+      #   # bad
+      #   Gem::Specification.new do |spec|
+      #     spec.add_runtime_dependency('rubocop')
+      #   end
+      #
+      #   # good
+      #   Gem::Specification.new do |spec|
+      #     spec.add_dependency('rubocop')
+      #   end
+      #
+      class AddRuntimeDependency < Base
+        extend AutoCorrector
+
+        MSG = 'Use `add_dependency` instead of `add_runtime_dependency`.'
+
+        RESTRICT_ON_SEND = %i[add_runtime_dependency].freeze
+
+        def on_send(node)
+          return if !node.receiver || node.arguments.empty?
+
+          add_offense(node.loc.selector) do |corrector|
+            corrector.replace(node.loc.selector, 'add_dependency')
+          end
+        end
+      end
+    end
+  end
+end
--- a/changelog/new_add_new_gemspec_add_runtime_dependency_cop.md
+++ b/changelog/new_add_new_gemspec_add_runtime_dependency_cop.md
@@ -1,0 +1,1 @@
+* [#13030](https://github.com/rubocop/rubocop/pull/13030): Add new `Gemspec/AddRuntimeDependency` cop. ([@koic][])
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -169,6 +169,7 @@ require_relative 'rubocop/cop/bundler/gem_version'
 require_relative 'rubocop/cop/bundler/insecure_protocol_source'
 require_relative 'rubocop/cop/bundler/ordered_gems'
 
+require_relative 'rubocop/cop/gemspec/add_runtime_dependency'
 require_relative 'rubocop/cop/gemspec/dependency_version'
 require_relative 'rubocop/cop/gemspec/deprecated_attribute_assignment'
 require_relative 'rubocop/cop/gemspec/development_dependencies'
--- a/rubocop.gemspec
+++ b/rubocop.gemspec
@@ -31,14 +31,14 @@ Gem::Specification.new do |s|
     'rubygems_mfa_required' => 'true'
   }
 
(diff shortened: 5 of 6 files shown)


bbatsov:
Won't it be better to have some more generic name that we can use for other deprecated methods in the gemspec down the road?

lib/rubocop/cop/gemspec/add_runtime_dependency.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Gemspec
+      # Prefer `add_dependency` over `add_runtime_dependency` because
+      # `add_dependency` is considered soft-deprecated.
```
Earlopain, line 7:
```suggestion
      # `add_runtime_dependency` is considered soft-deprecated.
```

koic:
Yeah, I have made this an independent cop for the following reasons:

- `add_runtime_dependency` is soft-deprecated to the extent that it is not even marked as deprecated in the documentation.
- This cop is mapped to the intent of a specific rule in the style guide.
- In case it becomes not deprecated in the future, it can be reclassified from a Lint-type cop to a Style-type cop. (If it remains a Lint-type cop, its `Severity` should be changed to `warning` by default.)

Therefore, it is not a general-purpose cop supporting deprecated methods. Of course, I'm not sure if this is the best choice. If there are any issues, I will rename it.

bbatsov:
Let's keep it as a separate cop then.

config/default.yml
```diff
@@ -262,6 +262,15 @@ Bundler/OrderedGems:
 
 #################### Gemspec ###############################
 
+Gemspec/AddRuntimeDependency:
```
noraj, line 265: > Yeah, I have made this an independent cop for the following reasons:
> 
> - `add_runtime_dependency` is soft-deprecated to the extent that it is not even marked as deprecated in the documentation.
> - This cop is mapped to the intent of a specific rule in the style guide.
> - In case it becomes not deprecated in the future, it can be reclassified from a Lint-type cop to a Style-type cop. (If it remains a Lint-type cop, its `Severity` should be changed to `warning` by default.)
> 
> 
> Therefore, it is not a general-purpose cop supporting deprecated methods. Of course, I'm not sure if this is the best choice. If there are any issues, I will rename it.

Since `add_runtime_dependency` is not deprecated, but it's just that `add_dependency` is “preferred”, I think it should be a style cop. Also, `add_runtime_dependency` is an alias that has been existing for 10+ years and that is massively used so it's not planned to be deprecated anytime soon in the future.

Also, I believe this makes a precedent: should we make a cop per alias?

- `Hash#select` ↔️ `Hash#filter`
- `Hash#each_pair` ↔️ `Hash#each`
- `Hash#include?` ↔️ `Hash#has_key?`
- `Hash#value?` ↔️ `Hash#has_value?`
- `Array#push` ↔️ `Array#append`
- `Array#collect` ↔️ `Array#map`
- `Array#find_index` ↔️ `Array#index`
- etc. just to name a few


I looked for existing alias cops, and it seems there are no alias cops for Ruby and 2 alias cops for RoR [Rails/ActiveRecordAliases](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsactiverecordaliases) and [Rails/ActiveSupportAliases](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsactivesupportaliases).

So to keep it coherent, this cop should actually be renamed from `AddRuntimeDependency` to `AddRuntimeDependencyAlias`.

But then, a new question arises: are we going to make a group cop with all aliases, or one cop per alias?

We could group everything in two cops:

- `Gempsec/Aliases` or `Style/GemspecAliases` for all [Gemspec specific aliases](https://github.com/search?q=repo%3Arubygems%2Frubygems+alias_method+language%3ARuby&type=code)
- `Style/Aliases` for Ruby standard library aliases (or maybe grouping them by class: `Style/ArrayAliases`, `Style/HashAliases`, etc.)

koic: I also considered adopting a consistent style between `add_dependency` and `add_runtime_dependency` when implementing cop, but I hesitate to make `add_runtime_dependency`, which the Bundler and RubyGems teams consider soft-deprecated, an option:

> My personal take is that option 2 is the one consistent with the last ten years of maintainer actions. The `add_runtime_dependency` was added long ago, before Bundler existed, while adding `add_development_dependency`, which is (to my understanding) soft-deprecated by the Bundler and RubyGems team.

https://github.com/rubygems/rubygems/issues/7799#issuecomment-2192720316

Considering the thoughts of the developer linked, the case is different from the aliases mentioned as examples (e.g., `select` and `filter`), so it doesn't seem appropriate to consider them as such aliases.

noraj: On one hand, there is only one person saying:

> which is (**to my understanding**) soft-deprecated

So it's what he thinks, not a proven fact.

Also, the definition of _soft-deprecated_ is:

> Soft deprecated allows a package to change its interface to encourage package developers to update their code before their users are forced to change.

So _soft-deprecated_ means that is was tagged an officially deprecated with a warning label as opposed to _hard-deprecated_ or _defunct_ when the code is removed.

`add_runtime_dependency` is then not soft-deprecated as it's not tagged as it in the code.

The person also says:

> The Bundler team has consistently **recommended** add_dependency [...], and I **believe** it should be the "**blessed**" option.

Vocabulary: recommended, believe, blessed.

Also:

> I **don't think** we need to actively **deprecate** the other methods [...].

So to me, it's pretty clear `add_runtime_dependency` is officially an alias like any other and is not deprecated.

@indirect What are your thoughts about this? Could you enlighten us?

koic: For reference, there is also an internal spec update based on the issue at https://github.com/rubygems/rubygems/pull/7834. In any case, let's wait for a response from the Bundler and RubyGems team.

indirect: I made my understanding clear because I am just one person and do not want to speak for other team members without permission. As @koic rightfully points out, the consensus position that emerged from the team after the comment you quote is that we all recommend and prefer `add_dependency`. I doubt we will remove the other method, since it is required for existing gemspecs to evaluate correctly.

bbatsov: I think we're overthinking this. I also wondered what's the best place for this suggestion, but at the end I decided it doesn't matter that much and I decided to go with @koic suggestion. In practice I don't see any potential downsides for our users based on this. We can make this configurable if someone wants to enforce the old name, but I don't see any reason to go farther than this.

noraj: Okay, let's forget everything we said about this cop and let it as is. Would that still be a good idea to implement the style Alias cops I mentioned earlier for people who want to implement a consistent naming convention? Having several styles configurable: original only, alias only or mixed.

If you think it's a good idea I can open a separate discussion dedicated to it.

PR #13031, opened 2024-07-07
Enable YJIT by default in server mode

koic:
YJIT tends to increase memory consumption as a trade-off for speed, but `--server` mode is usually used to speed up the execution of `rubocop` command locally. This patch is based on the hypothesis that it is acceptable to always enable YJIT by default in server mode to prioritize performance.

Therefore, this patch doesn't provide an option to disable YJIT in server mode. Whether this is truly necessary can be considered based on user feedback.

-----------------

* [ ] Added tests.


bbatsov:
Makes sense to me. Did you manage to measure the performance gain when using YJIT in `--server` mode?

koic:
Unfortunately, no measurements have been taken. Theoretically, it is considered the same as https://github.com/rubocop/rubocop/pull/12879.

PR #13038, opened 2024-07-12
Consider Sorbet signatures in `Style/DocumentationMethod` cop

wpolicarpo:
The `Style/DocumentationMethod` cop does not support Sorbet's `sig` blocks when both sig and docs are present, making it incompatible with [Sorbet/EmptyLineAfterSig](https://github.com/Shopify/rubocop-sorbet/blob/main/manual/cops_sorbet.md#sorbetemptylineaftersig) cop.

I think it makes more sense for the docs to be above the signature than the opposite. 

```ruby
# This should be good
class Foo
  extend T::Sig

  # Documentation
  sig { void }
  def bar
    do_something
  end
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/documentation_method.rb
+++ b/lib/rubocop/cop/style/documentation_method.rb
@@ -117,6 +117,25 @@ module RuboCop
           (send nil? {:module_function :ruby2_keywords} ...)
         PATTERN
 
+        # @!method sig_node?(node)
+        def_node_matcher(:sig_node?, <<~PATTERN)
+          (block (send
+            {nil? #with_runtime? #without_runtime?}
+            :sig
+            (sym :final)?
+          ) (args) ...)
+        PATTERN
+
+        # @!method with_runtime?(node)
+        def_node_matcher(:with_runtime?, <<~PATTERN)
+          (const (const {nil? cbase} :T) :Sig)
+        PATTERN
+
+        # @!method without_runtime?(node)
+        def_node_matcher(:without_runtime?, <<~PATTERN)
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/style/documentation_method.rb
```diff
@@ -117,6 +117,25 @@ class DocumentationMethod < Base
...
+
+        # @!method with_runtime?(node)
+        def_node_matcher(:with_runtime?, <<~PATTERN)
+          (const (const {nil? cbase} :T) :Sig)
+        PATTERN
+
+        # @!method without_runtime?(node)
+        def_node_matcher(:without_runtime?, <<~PATTERN)
+          (const (const (const {nil? cbase} :T) :Sig) :WithoutRuntime)
+        PATTERN
```
wpolicarpo, line 137: Borrowed from rubocop-sorbet itself.

bbatsov:
I'm fine with the proposal, but I think this change in behavior should be mentioned somewhere in the docs as well, as I don't think it'd be obvious out-of-the-box that the cop has some special considerations for Sorbet.

bbatsov:
@wpolicarpo ping :-)

wpolicarpo:
@bbatsov I will see if I can add something to the docs this week, sorry for the delay :)

wpolicarpo:
@bbatsov I added a note to the docs and rebased my branch to pull latest master. Let me know what you think.

bbatsov:
@wpolicarpo The cop docs are generated automatically from the leading comments in the cop files, so you'll need to move your changes there.

bbatsov:
(the wording and the example themselves are fine)

PR #13053, opened 2024-07-17
Added ability to skip some lines while calculating length

AsbahIshaq:
The PR adds the ability to skip certain lines while calculating lengths of class, method, module or block.

cops affectted:
```
Metrics/BlockLength
Metrics/ClassLength
Metrics/MethodLength
Metrics/ModuleLength
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_added_ability_to_skip_some_lines_while_calculating.md
+++ b/changelog/change_added_ability_to_skip_some_lines_while_calculating.md
@@ -1,0 +1,1 @@
+* [#13053](https://github.com/rubocop/rubocop/pull/13053): Added ability to skip some lines while calculating length. ([@AsbahIshaq][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2613,6 +2613,7 @@ Metrics/BlockLength:
   CountComments: false  # count full line comments?
   Max: 25
   CountAsOne: []
+  SkipableLines: []
   AllowedMethods:
     # By default, exclude the `#refine` method, as it tends to have larger
     # associated blocks.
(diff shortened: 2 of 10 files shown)


Earlopain:
I don't believe applying regex to code is the way to go. Can you explain with a concrete example why you'd want that and why inline disabling doesn't work for you?

AsbahIshaq:
It will be helpful in skipping log lines and detailed error messages in the codebase. Inline disabling will have to be used everywhere separately. 

I have an example of a user asking for it in [stackoverflow](https://stackoverflow.com/questions/78753404/how-to-use-rubocop-metrics-methodlength-allowedpatterns-allowedmethods/78761110#78761110)

Earlopain:
Thanks, this context makes sense to me. I think, this is basically `AllowedPatterns` is for. As you noticed, one of the metrics cop already makes use of that. There's a module that encapsulates that logic `(AllowedPattern)` and you should be able to reuse that here pretty easily.

AsbahIshaq:
``AllowedPatterns`` is for method names. In my understanding it will allow methods that have names matching ``AllowedPatterns`` but the PR allows to skip the matching lines in the total count.
For example:
``SkipableLines =[\/bputs/b\]`` will count the length of the below method to be 1 instead of 2.

```
def method_1
  puts 'a'
  x = 2
end
```

but ``AllowedPatterns = [\/bmethod_1/b\]`` will allow ``method_1`` to have any number of lines

AsbahIshaq:
sorry I might have misunderstood you. Did you mean to implement this feature using ``AllowedPattern``?

Earlopain:
Yeah, that's what I meant. `AllowedPatterns`, to my understanding, is a generic mechanism to ignore certain lines. Quite a few cops already implement it. What this option actually means in the end is up to the individual cops. Though, I think, this would clash with the already existing option from `BlockLength`, so perhaps not so good of an idea after all.

AsbahIshaq:
yes I think it will create a clash, also I think the name (``AllowedPatterns``) would be a bit confusing in the context of calculating length of a block, method etc

AsbahIshaq:
can anyone look at this one please? @koic @Earlopain

PR #13054, opened 2024-07-18
[Fix #13050] Add new `Style/BitwisePredicate` cop

### LINKED ISSUE #13050: Cop idea: Style/BitPredicates

Ruby 2.5 (!) [introduced](https://rubyreferences.github.io/rubychanges/2.5.html#allbits-anybits-nobits) `Integer#anybits?`/`#allbits?`/`#nobits?` predicates. Accordingly, it would be cool if Rubocop suggested those corrections:
```ruby
# 1. anybits?
# bad
(variable & MY_FLAGS).positive?
# good
variable.anybits?(MY_FLAGS)

# 2. allbits?
# bad
(variable & MY_FLAGS) == MY_FLAGS
# good
variable.allbits?(MY_FLAGS)

# 3. nobits?
# bad
(variable & MY_FLAGS).zero?
# good
variable.nobits?(MY_FLAGS)
```

I’ve seen a lot of codebases that use the bit-math version, and in many cases, predicates version seems to be clearer. It also has an educational value: I constantly see people writing bit-math to check flags seven years after methods introduction (because they simply don’t remember methods exist).

Haven’t yet time enough to work on the cop myself (well, I happen to have a draft branch, and I feel guilty for never finishing it, maybe I’ll do... but it is what it is) :cry:

koic:
Closes #13050.

This PR adds new `Style/BitwisePredicates` cop, which checks for usage of `&` operator in bitwise operations involving comparisons.

```ruby
# bad - checks any set bits
(variable & flags).positive?

# good
variable.anybits?(flags)

# bad - checks all set bits
(variable & flags) == flags

# good
variable.allbits?(flags)

# bad - checks no set bits
(variable & flags).zero?
(variable & flags) == 0

# good
variable.nobits?(flags)
```

I've opened a proposal in the style guide at https://github.com/rubocop/ruby-style-guide/pull/944.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/bitwise_predicate.rb
```diff
@@ -0,0 +1,95 @@
...
+
+        minimum_target_ruby_version 2.5
+
+        # @!method anybits?(node)
+        def_node_matcher :anybits?, <<~PATTERN
+          {
+            (send #bit_operation? :positive?)
+            (send #bit_operation? :> (int 0))
+            (send #bit_operation? :>= (int 1))
+          }
```
vlad-pisanov, line 43: another common one is
```ruby
if (variable & flags) != 0
```

lib/rubocop/cop/style/bitwise_predicate.rb
```diff
@@ -0,0 +1,95 @@
...
+      # @example
+      #
+      #   # bad - checks any set bits
+      #   (variable & flags).positive?
+      #
+      #   # good
+      #   variable.anybits?(flags)
+      #
+      #   # bad - checks all set bits
+      #   (variable & flags) == flags
```
vlad-pisanov, line 17: This cop should be marked unsafe since such expressions could involve arrays/sets.

lib/rubocop/cop/style/bitwise_predicate.rb
```diff
@@ -0,0 +1,95 @@
...
+      #   (variable & flags).zero?
+      #
+      #   # good
+      #   variable.nobits?(flags)
+      #
+      class BitwisePredicate < Base
+        extend AutoCorrector
+        extend TargetRubyVersion
+
+        MSG = 'Replace with `%<preferred>s`.'
```
zverok, line 32: A small thing (and I am not sure whether Rubocop has a guidelines for that — BTW, has it?..), but I think that the message could be:
1. A bit softer in tone
2. A bit more educational (why are we proposing that)

Like
> Prefer `%<preferred>s` for comparison with bit flags.

...or something like that.

koic: Style guide proposal https://github.com/rubocop/ruby-style-guide/pull/944 has been merged. But `Prefer %<preferred>s for comparison with bit flags.` is better, I think. Thank you.

koic:
Thank you for feedback. I updated this PR.

lib/rubocop/cop/style/bitwise_predicate.rb
zverok: Cool :+1: 

(When I asked about the guideline, I meant “whether there is some Rubocop guide about the preferred way of offense phrasing”? I believe they aren’t always consistent and sometimes lack the “educational” merit, especially valuable for newer Ruby features. 

Not the fault of cops authors, of course, but maybe some kind of official recommendation about phrasing/level of details might help.

cc @bbatsov)

bbatsov: We never wrote anything official, but most cops stick to more or less the same language, so probably we should standardize it at some point and add something here https://docs.rubocop.org/rubocop/1.65/development.html#create-a-new-cop.

Might also be good to have some internal cops checking for the message consistency (see https://github.com/rubocop/rubocop/tree/master/lib/rubocop/cop/internal_affairs)

But I guess that's all a topic for a dedicated conversation/discussion.

zverok:
@koic @bbatsov Is there a chance of merging this? I have a codebase that _begs_ for this cop dearly :grin:

bbatsov:
@zverok Sure!

Issue #12225, opened 2023-09-28
Rubocop 1.56.3 consuming high CPU and memory on CircleCI


akhilgkrishnan:
We recently upgraded our RuboCop version from 1.42.0 to 1.56.3, and following the upgrade, we observed occasional crashes in our CircleCI environment, characterized by high CPU and memory usage.

With the previous RuboCop version, everything was running smoothly without any such issues. In our CircleCI configuration, we have been utilizing resource_class: small to allocate resources.

## Expected behavior

We anticipate that the CircleCI RuboCop tests should execute successfully without encountering any problems.

## Actual behavior

![Screenshot 2023-09-28 at 6 53 50 PM](https://github.com/rubocop/rubocop/assets/22231095/9f6ad91e-6929-4f82-b428-8f9f96df1eb0)

## RuboCop version

1.56.3

fatkodima:
It is hard to tell what is wrong.

1. Is upgrading rubocop version the only thing you did before observing the mentioned errors?
2. Do you see the same behavior when running locally?
3. Can you profile rubocop on CircleCI (see https://github.com/rubocop/rubocop/pull/11396)?

koic:
I will close this issue due to the lack of sufficient information for analysis. Please feel free to reopen it once you have gathered information to analyze the problem. Thank you.

letrungkien:
For those who have the same issue. Here's how to fix it:
```
PARALLEL_PROCESSOR_COUNT=2 bundle exec rubocop --config .rubocop.yml --parallel
```
by adding `PARALLEL_PROCESSOR_COUNT`.
ref from Parallel gem: https://github.com/grosser/parallel/blob/f66f10552e32681bf47f7dfd2f8183d7e0de442e/lib/parallel.rb#L347

In circleci, I checked that it returned 36 processors (???). That's why CircleCI runner couldn't handle and kill the rubocop command because of OOO
```
Etc.nprocessors
36
```

PR #13056, opened 2024-07-19
Clarify how to specify the number of parallel processes

koic:
Follow up https://github.com/rubocop/rubocop/issues/12225#issuecomment-2097859065.

This PR clarifies how to specify the number of parallel processes to adjust settings when there is excessive CPU and memory usage.

-----------------

* [x] Added tests.


spec/rubocop/options_spec.rb
```diff
@@ -63,6 +63,8 @@ def abs(path)
                                                adjusting behaviors for editing experience.
               -P, --[no-]parallel              Use available CPUs to execute inspection in
                                                parallel. Default is true.
+                                               You can specify the number of parallel processes
```
bbatsov, line 66: I guess you missed something like `with/via` here.

lib/rubocop/options.rb
```diff
@@ -615,7 +615,9 @@ module OptionsHelp
       version:                          'Display version.',
       verbose_version:                  'Display verbose version.',
       parallel:                         ['Use available CPUs to execute inspection in',
-                                         'parallel. Default is true.'],
+                                         'parallel. Default is true.',
+                                         'You can specify the number of parallel processes',
```
bbatsov, line 619: Same. I use "using" being used in the longer text. :-) Maybe that's what you meant to put here as well.

koic:
Thanks! I've updated this PR.

PR #13067, opened 2024-07-25
[Fix #13061] Add new `Style/RedundantInterpolationUnfreeze` cop

### LINKED ISSUE #13061: Cop idea: When Ruby >= 3.0, don't unfreeze interpolated strings when frozen magic comment is true
For compatibility with Ruby 2.7, it may be necessary to write the following code:

```rb
# frozen_string_literal: true

foo = +"#{bar}baz"
# or foo = "#{bar}baz".dup
foo << "bat"
```

This is because interpolated strings follow the frozen string literal comment on Ruby below 2.7. This behaviour got changed with Ruby 3.0 which means that this unfreezing is not necessary anymore. It got discussed in https://bugs.ruby-lang.org/issues/17104

Earlopain:
Closes #13061

In Ruby >= 3.0, interpolated strings are always unfrozen: https://bugs.ruby-lang.org/issues/17104

On Ruby 2.7 they followed the magic comment literal which made it necessary to sometimes unfreeze them manually.

This cop could also apply on Ruby 2.7 by looking at the magic comment but it doesn't seem worth the effort. Personally I'm not interested in that.

I handle a few special cases I'm aware of like `'foo' \\n'bar'` (https://github.com/rubocop/rubocop/issues/10134). Let me know if you can can think of more cases like that that I missed.

RuboCop itself has 5 offenses (if it would target 3.0):

<details><summary>Details</summary>
<p>

```
Offenses:

lib/rubocop/cop/lint/unused_method_argument.rb:99:21: C: [Correctable] Style/RedundantInterpolationUnfreeze: Don't unfreeze interpolated strings as they are already unfrozen
          message = +"Unused method argument - `#{variable.name}`."
                    ^
lib/rubocop/cops_documentation_generator.rb:133:15: C: [Correctable] Style/RedundantInterpolationUnfreeze: Don't unfreeze interpolated strings as they are already unfrozen
    content = +"==== #{title}\n"
              ^
lib/rubocop/cops_documentation_generator.rb:259:15: C: [Correctable] Style/RedundantInterpolationUnfreeze: Don't unfreeze interpolated strings as they are already unfrozen
    content = +<<~HEADER
              ^
lib/rubocop/cops_documentation_generator.rb:307:15: C: [Correctable] Style/RedundantInterpolationUnfreeze: Don't unfreeze interpolated strings as they are already unfrozen
    content = +"=== Department xref:#{filename}[#{type_title}]\n\n"
              ^
spec/rubocop/cop/layout/end_of_line_spec.rb:21:16: C: [Correctable] Style/RedundantInterpolationUnfreeze: Don't unfreeze interpolated strings as they are already unfrozen
      input = (+<<~RUBY).force_encoding(encoding)
               ^

1546 files inspected, 5 offenses detected, 5 offenses autocorrectable
```

</p>
</details> 

-----------------

* [x] Added tests.


lib/rubocop/cop/mixin/frozen_string_literal.rb
```diff
@@ -32,11 +32,12 @@ def uninterpolated_string?(node)
         node.str_type? || (node.dstr_type? && node.each_descendant(:begin).none?)
       end
 
-      def frozen_heredoc?(node)
+      def uninterpolated_heredoc?(node)
```
Earlopain, line 35: This name seems more accurate. I left an alias since this is technically public api but after a github code search I doubt anyone actually uses this right now

bbatsov:
Seems more of a Lint cop to me. Perhaps we should change its department? Is there any particular reason you decided to go with Style?

Earlopain:
The lint description says:
>  Lint cops check for ambiguities and possible errors in your code.

That doesn't apply here so I didn't think it should go in that category.

bbatsov:
One can say that meaningless operations are kind of an error, but fair enough. I've never been particularly dogmatic about the separation between the two, and originally Lint cops mostly replicated Ruby's own warnings.

Earlopain:
I don't have an intuition about if a cop should be style or lint so I defer to what I can read in the docs about that. I'd follow your opinion here, many cops already are in a different department that I would expect.

`Lint/DeprecatedConstants` for example, wouldn't count that as an error. I can see the other perspective why it's Lint anyways but I would probably have chucked that into Style as well. It can be difficult to decide.

PR #13068, opened 2024-07-25
Clarify docs in `Naming/PredicateName`

maxjacobson:
- Remove references to autocorrect (it does not autocorrect)
- Add some more examples

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/naming/predicate_name.rb
+++ b/lib/rubocop/cop/naming/predicate_name.rb
@@ -93,6 +93,16 @@ module RuboCop
         end
         alias on_defs on_def
 
+        def validate_config
+          forbidden_prefixes.each do |forbidden_prefix|
+            next if predicate_prefixes.include?(forbidden_prefix)
+
+            raise ValidationError, <<~MSG.chomp
+              The `Naming/PredicateName` cop is misconfigured. Prefix #{forbidden_prefix} must be included in NamePrefix because it is included in ForbiddenPrefixes.
+            MSG
+          end
+        end
+
         private
 
         def allowed_method_name?(method_name, prefix)
--- a/lib/rubocop/cop/naming/predicate_name.rb
+++ b/lib/rubocop/cop/naming/predicate_name.rb
@@ -3,58 +3,74 @@
 module RuboCop
   module Cop
     module Naming
-      # Checks that predicate methods names end with a question mark and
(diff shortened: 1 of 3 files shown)


changelog/change_documentation_of_naming_predicate_name.md
```diff
@@ -0,0 +1 @@
+* [#13068](https://github.com/rubocop/rubocop/pull/13068): Clarify `Naming/PredicateName` docs to remove erroneous references to autocorrection and add more examples. ([@maxjacobson][])
```
koic, line 1: Can you remove this file? Because documentation change doesn't require changelog entry.

maxjacobson: Ah, sounds good. Done

lib/rubocop/cop/naming/predicate_name.rb
```diff
@@ -3,59 +3,107 @@
...
       #
-      # @example
+      # @example NamePrefix: ['is_', 'has_', 'have_'] (default)
       #   # bad
       #   def is_even(value)
       #   end
       #
-      #   def is_even?(value)
+      #   # good
+      #   def even?(value)
```
bbatsov, line 38: Btw, without matching "ForbiddenPrefixes" `is_even?` will be good as well, which I think we should mention for clarity.

lib/rubocop/cop/naming/predicate_name.rb
```diff
@@ -3,59 +3,107 @@
...
+      #   end
+      #
+      # @example NamePrefix: ['seems_to_be_']
+      #   # bad
+      #   def seems_to_be_even(value)
```
bbatsov, line 43: Same here for `seems_to_be_even?`.

lib/rubocop/cop/naming/predicate_name.rb
maxjacobson: Yeah, it's a little tricky to demonstrate the combination of configurations. I pushed up an update which I hope makes it more clear.

While doing that, I also thought it would be nice to add a config validation to automatically check the config rather than trust that people actually read the docs. It seems possible to really streamline the prose at the top and lean on the examples and the config validation to demonstrate the nuances.

Feedback welcome of course.

Issue #11191, opened 2022-11-19
New cop that detects operations with constant results


fatkodima:
Inspired by https://hexdocs.pm/credo/Credo.Check.Warning.OperationWithConstantResult.html

It should check for the code like
```ruby
x * 1
x * 0
x / 1
x ** 0
x ** 1
x % 1
```

This is almost always a mistake (or just left after debugging, probably) or just redundant code.

PR #13074, opened 2024-07-29
Add new `Lint/UselessNumericOperation` cop

Zopolis4:
> [!NOTE]
> The operations with constant results (i.e. `x * 0` will be implemented in a different cop, as they include patterns like `x - x` which will require modifications to `Lint/BinaryOperatorWithIdenticalOperands`)

Makes progress on #11191, taking care of the inconsequential operations.

Certain numeric operations have no impact, being:
Adding or subtracting 0, multiplying or dividing by 1 or raising to the power of 1.

These are probably mistakes or leftover from debugging, so flag and optionally remove them.

* [x] Added tests.


lib/rubocop/cop/lint/useless_numeric_operation.rb
```diff
@@ -0,0 +1,77 @@
...
+      #
+      #   # bad
+      #   x += 0
+      #   x -= 0
+      #   x *= 1
+      #   x /= 1
+      #   x **= 1
+      #
+      #   # good
+      #   x = x
```
bbatsov, line 30: Perhaps you meant to write just `x` here, as I don't see how `x = x` makes sense in this context.

Zopolis4: My thinking was to only remove the numeric component-- the assignment can be handled by `Lint/UselessAssignment`.

bbatsov:
What about doing the same things with floats? The results will be similar, even if often they won't be totally accurate due to the way floats are implemented internally.

Zopolis4:
I guess? To be honest, I'm not confident enough in my understanding of floats to implement that.

bbatsov:
Okay, I guess we can merge this as is. Thanks for working on it!

PR #13077, opened 2024-07-29
[Fix #11977] Add new global `StringLiteralsFrozenByDefault` option

### LINKED ISSUE #11977: Style/EmptyLiteral with --enable=frozen-string-literal flag
## Expected behavior

`String.new` should not be auto-corrected with `''` when have `Style/EmptyLiteral` enabled and run with the `--enable=frozen-string-literal` flag.

## Actual behavior

`String.new` will be replaced with `''`

1. add a ruby file with a line `String.new`.
2. run `export RUBYOPT="--enable-frozen-string-literal"` in command line.
3. run auto-correct on the ruby file.
4. `String.new` will be replaced with `''`

## RuboCop version

```
$ [bundle exec] rubocop -V
1.53.0 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 3.0.0) [x86_64-linux]
  - rubocop-performance 1.15.0
  - rubocop-rails 2.16.1
```

## Possible fixes:

We can check `RUBYOPT` in `frozen_string_literals_enabled?` method.
Ex:
```rb
return true if ENV['RUBYOPT'].include?('--enable-frozen-string-literal')
```
https://github.com/rubocop/rubocop/blob/a7b7c5f95ac32612b0f9c0a9661281b5633ae07b/lib/rubocop/cop/mixin/frozen_string_literal.rb#L21-L41

Earlopain:
While Ruby continues to not have this the default, user can opt into it using the `RUBYOPT` environment variable. To make RuboCop behave correctly when doing that, offer a config value to say how literals behave by default.

The most egregious cop `Style/FrozenStringLiteralComment` can already be configured appropriately by setting `EnforcedStyle` to `never`. This makes if possible to also change behaviour about cops that concern themselves with string literals (primarily `MutableConstant`).

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/empty_literal.rb
+++ b/lib/rubocop/cop/style/empty_literal.rb
@@ -129,8 +129,10 @@ module RuboCop
         def frozen_strings?
           return true if frozen_string_literals_enabled?
 
-          frozen_string_cop_enabled = config.for_cop('Style/FrozenStringLiteral')['Enabled']
-          frozen_string_cop_enabled && !frozen_string_literals_disabled?
+          frozen_string_cop_enabled = config.for_cop('Style/FrozenStringLiteralComment')['Enabled']
+          frozen_string_cop_enabled &&
+            !frozen_string_literals_disabled? &&
+            string_literals_frozen_by_default?.nil?
         end
       end
     end
--- a/changelog/new_all_cops_string_literals_frozen_by_default.md
+++ b/changelog/new_all_cops_string_literals_frozen_by_default.md
@@ -1,0 +1,1 @@
+* [#13077](https://github.com/rubocop/rubocop/pull/13077): Add new global `StringLiteralsFrozenByDefault` option for correct analysis with `RUBYOPT=--enable=frozen-string-literal`. ([@earlopain][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -166,6 +166,11 @@ AllCops:
     rubocop-rspec_rails: [rspec-rails]
   # Enable/Disable checking the methods extended by Active Support.
   ActiveSupportExtensionsEnabled: false
+  # Future version of Ruby will freeze string literals by default.
+  # This allows to opt in early, for example when enabled through RUBYOPT.
+  # For now this will behave as if set to false but in future ruby versions
+  # (likely 4.0) it will be true by default.
+  StringLiteralsFrozenByDefault: ~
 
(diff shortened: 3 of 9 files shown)


bbatsov:
> While Ruby continues to not have this the default, user can opt into it using the RUBYOPT environment variable. To make RuboCop behave correctly when doing that, offer a config value to say how literals behave by default.

Alternatively we can check the env variable and remove the need for some manual configuration. But in general I'm OK with your proposal.

Earlopain:
Draft for now, found an issue with `Style/EmptyLiteral` that needs to be addressed first.

lib/rubocop/cop/style/empty_literal.rb
```diff
@@ -129,8 +129,10 @@ def correction(node)
         def frozen_strings?
           return true if frozen_string_literals_enabled?
 
-          frozen_string_cop_enabled = config.for_cop('Style/FrozenStringLiteral')['Enabled']
```
Earlopain, line 132: Typo since introduction, was always true for most users since `Enabled` came from the department.

lib/rubocop/cop/style/empty_literal.rb
```diff
@@ -129,8 +129,10 @@ def correction(node)
         def frozen_strings?
```
Earlopain, line 129: This function says strings are frozen in preparation of `Style/FrozenStringLiteralComment` doing its autocorrect. Introduced in #9773. I don't love this since it now gets complicated when the default changes. The change to the function works but I feel pretty meh about it.

Why no `autocorrect_incompatible_with`? `EmptyLiteral` comes before `FrozenStringLiteralComment` alphabetically. I played around with letting cops say their autocorrect is more important than others but didn't like how that turned out.

I would rather get rid of this workaround entirely. The user runs unsafe autocorrect and should not be surprised when things break.

Earlopain:
Is this good or do I need to change anything?

lib/rubocop/cop/style/empty_literal.rb
bbatsov: No argument from me.

bbatsov:
Looks good to me. Thanks!

Earlopain:
Can't wait to yeet these magic comments in a few of my projects (:

Really nice that this is possible, most libraries seem well-prepared already.

PR #13080, opened 2024-07-30
Add new `DocumentationFormat` to serve documentation with extensions different than `html`

Earlopain:
The `rubocop-sorbet` extension hosts its documentation on GitHub in markdown format. This allows this extension to also benefit from providing documentation urls through RuboCop. It follows all the conventions that rubocop expects here, except for the file extension. GitHub doesn't render html files inline (for obvious reasons).

Documentation is available here as md: https://github.com/Shopify/rubocop-sorbet/blob/main/manual/cops_sorbet.md

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -82,6 +82,8 @@ AllCops:
   StyleGuideBaseURL: https://rubystyle.guide
   # Documentation URLs will be constructed using the base URL.
   DocumentationBaseURL: https://docs.rubocop.org/rubocop
+  # Documentation URLs will end with this extension.
+  DocumentationExtension: .html
   # Extra details are not displayed in offense messages by default. Change
   # behavior by overriding ExtraDetails, or by giving the
   # `-E/--extra-details` option.
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -976,6 +976,21 @@ Rails:
   DocumentationBaseURL: https://docs.rubocop.org/rubocop-rails
 ----
 
+By default, documentation is expected to be served as HTML but if you prefer
+to use something else like markdown you can set `DocumentationExtension`.
+
+With markdown as the documentation format you are able to host it directly through
+GitHub without having to own a domain or using GitHub Pages. The `rubocop-sorbet`
+extension is an example of this, its docs are available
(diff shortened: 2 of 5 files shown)


config/default.yml
```diff
@@ -82,6 +82,8 @@ AllCops:
   StyleGuideBaseURL: https://rubystyle.guide
   # Documentation URLs will be constructed using the base URL.
   DocumentationBaseURL: https://docs.rubocop.org/rubocop
+  # Documentation URLs will end with this extension
+  DocumentationFormat: .html
```
bbatsov, line 86: Won't it be better to specify the actual format here and just infer the extension from it?

Earlopain: You mean something like `application/html`? I don't think I'm following.

bbatsov: I thought about "markdown" instead of "md", but this was before I actual saw what needed to be changed. :-)

koic: Because the name is `DocumentationFormat` rather than `DocumentationExtension`, `markdown` felt like a more appropriate value than `.md`. It would be better to match the name and the value.

Earlopain: This makes sense, thanks. I renamed the option.

docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -976,6 +976,15 @@ Rails:
   DocumentationBaseURL: https://docs.rubocop.org/rubocop-rails
 ----
 
+By default, documentation will be expected to be served as HTML but if you need
```
bbatsov, line 979: need -> prefer

bbatsov:
Given that this is a bit niche you might also want to mention specific examples, so people would have an easier time to figure out how to use this in practice. (not the config itself, but to have their document be served is some other format) And perhaps mention why someone would want to do this in the first place.

Earlopain:
Sure, I can do that. Is it fine to explicitly point to `rubocop-sorbet` even though its not part of the official extensions?

bbatsov:
Yeah, it's fine.

Earlopain:
Added an extra paragraph for this, let me know what you think. Also made the config demonstration more useful with an  example would produce working links.

bbatsov:
Looks good. Thanks!

PR #13087, opened 2024-08-01
Fix an incorrect autocorrect for `Style/MultipleComparison` when statement with more comparisons precedes a statement with less comparisons

fatkodima:
I caught an edge case for `Style/MultipleComparison` cop, when we have an expression with more comparisons that is followed by an expression with less comparisons.

```ruby
x = 1
x == 1 || x == 2 || x == 3  # <=== more comparisons
x == 1 || x == 2            # <=== less comparisons
```

Is corrected to
```ruby
x = 1
[1, 2, 3].include?(x)
[1, 2, 3, 1, 2, 3, 1, 2].include?(x)
```

But should be
```ruby
x = 1
[1, 2, 3].include?(x)
[1, 2].include?(x)
```


bbatsov:
Nice catch!

PR #13112, opened 2024-08-11
Fix detection of ruby versions through the gemfile below 2.7

Earlopain:
Some things going on here:

If I have a gemspec saying `target_ruby_version ">= 2.6"`, what RuboCop will actually use is 2.7 because it is the default version. All versions prior are simply discarded, even if rubocop is able to provide accurate results for these older versions.

There were tests for this but I believe that they lost what they actually were testing against during some refactor, somewhere here: #9515, #10629, and the PRs removing and reading analysis for older rubies.

This test for example used to check that the returned ruby version is 2.4 and not 2.3 which it didn't know how to analyse at one point (i.e. don't select versions below what we are currently supporting):

```rb

          it 'sets target_ruby from required_ruby_version from inclusive requirement range' do
            content = <<~HEREDOC
              Gem::Specification.new do |s|
                s.name = 'test'
                s.required_ruby_version = Gem::Requirement.new('>= 2.3.1')
                s.licenses = ['MIT']
              end
            HEREDOC

            create_file(gemspec_file_path, content)
            expect(target_ruby.version).to eq default_version
          end
```

Now it checks that 2.7 is the returned version, which doesn't make much sense since older versions are able to be analysed.

For the `< 3.0.0` case, I think its ok to select the earliest version supported (currently 2.0). Selecting 2.7 (or whatever the default may be in the future) surely isn't the intention.

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/fix_ruby_version_detection_gemfile.md
+++ b/changelog/fix_ruby_version_detection_gemfile.md
@@ -1,0 +1,1 @@
+* [#13112](https://github.com/rubocop/rubocop/pull/13112): Fix detection of `TargetRubyVersion` through the gemfile if the gemfile ruby version is below 2.7. ([@earlopain][])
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

PR #13115, opened 2024-08-13
Add a configuration option to NonLocalExitFromIterator to handle return values.

### LINKED ISSUE #4064: please add a cop that warns if there is a return from within a block

sloanesturz:
Fixes #4064.

This solves a long-standing (August 2017) issue. An explicit `return` in a block has unintuitive behavior where the `yield`-ing code is skipped.

NonLocalExitFromIterator _almost_ catches this error, it just needs a setting to make it more aggressive.

Example of the type of bug I'm trying to catch:

```
def execute_with_block
  result = yield
  puts "The result was #{result}"
  result
end

def handle_thing
  execute_with_block do 
    return 3
  end

foo = handle_thing # foo is 3, but the `puts` line never executes
```

Related: https://github.com/rubocop/rubocop/pull/12971

This is my first PR to Rubocop, so please let me know if I need additional work on this change.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/non_local_exit_from_iterator.rb
+++ b/lib/rubocop/cop/lint/non_local_exit_from_iterator.rb
@@ -14,7 +14,10 @@ module RuboCop
       # * the return is not contained in an inner scope, e.g. a lambda or a
       # method definition.
       #
-      # @example
+      # Setting `AllowReturnValue` to `false` overrides the default behavior
+      # for the first condition.
+      #
+      # @example AllowReturnValue: true (default)
       #
       #   class ItemApi
       #     rescue_from ValidationError do |e| # non-iteration block with arg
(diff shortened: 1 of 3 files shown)


Earlopain (commented):
I agree this makes more sense as a config instead of a new cop. What do you think @koic?

lib/rubocop/cop/lint/non_local_exit_from_iterator.rb
```diff
@@ -14,7 +14,10 @@ module Lint
       # * the return is not contained in an inner scope, e.g. a lambda or a
       # method definition.
       #
-      # @example
+      # Setting `AllowReturnValue` to `false` overrides the default behavior
+      # for the first condition.
+      #
+      # @example AllowReturnValue: true (default)
```
sloanesturz, line 20: @Earlopain there are a lot of rules here -- in my mind, a setting called `Strict` should turn off all of the exceptions. That's why I went for a more narrow setting name to disable just "No value is returned".

I just ran a version of that "strictest" setting against my ~5000 line ruby project and I think it probably helps with correctness. I got about 25 new errors there: mostly around "returning from a block that doesn't have any arguments".

This does change the meaning of the Cop, a bit: it's no longer "non local exit from iterator" but "you should `break` instead of `return` from a block". Do you think that makes it worthwhile to add an entirely different Cop?

Issue #12840, opened 2024-04-12
Cop(s) Idea: remove redundant booleans from control flow


jayqui:
### Basic problem & use case

It is possible to run across cases where there are booleans in the antecedents of conditionals (`true ? foo : bar`). This kind of case can be simplified to return the logical branch that the statement will always results in (`foo`). A cop that could take care of this might be called `Lint/BooleanCondition`.

Perhaps sometimes a well-intentioned developer could straightforwardly make this kind of mistake. Anyway, another use case in production code is with what my team calls "fully-enabled feature flags": a bit of conditional code that will forever more evaluate to `true`, which we can reasonably `gsub` to just `true`. If we had `Lint/BooleanCondition` cop, then we could automate the feature flags away:
```ruby
# original
feature_flags(:foo_flag_enabled, user) ? render(:new_page) : render(:legacy_page)

# substitute `true` for fully enabled flag
true ? render(:new_page) : render(:legacy_page)

# apply `Lint/BooleanCondition`
render(:new_page)
```

#### antecendents that always evaluate to true

Sometimes a conditional can have an antecedent that will always evaluate to true, especially when it contains no variables, e.g. `2 > 1 ? foo : bar`, which will always evaluate to `foo`. Perhaps that kind of case could also be taken care of by a `Lint/BooleanCondition` cop.

#### conjunctions and disjunctions

`true && foo` will always evaluate to `foo`. `true || foo` will always evaluate to `true`. Perhaps such cases should also fit into the scope of `Lint/BooleanCondition`. If not, then those cases could have their own cops, `Lint/BooleanInConjunction` and `Lint/BooleanInDisjucntion`.

#### `unless`, `case`, etc.

Perhaps it goes without saying, but we would want `Lint/BooleanCondition` to handle the various kinds of conditionals Ruby allows for, including `if`, the ternary operator, `unless`, `case`, and possibly `and` and `or`.

## Describe the solution you'd like

A solution that could auto-correct away such cases as I will post in the first comment.

jayqui:
## 1. Boolean as antecedent of a conditional (`Lint/BooleanCondition`)

### 1.1 if/else with `true`

```ruby
# Bad
true ? foo : bar

# Bad
if true
  foo
elsif anything
  bar
else
  "never going to get here"
end

# Good 
foo
```

### 1.2 if/else with `false`

```ruby
# Bad
false ? foo : bar

# Bad
if false
  foo
else
  bar
end

# Good
bar
```

[...]

PR #13117, opened 2024-08-14
Add partial autocorrect support for `Lint/LiteralAsCondition`

Zopolis4:
Makes progress on #12840.

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/literal_as_condition.rb
+++ b/lib/rubocop/cop/lint/literal_as_condition.rb
@@ -34,27 +34,100 @@ module RuboCop
       #   end
       class LiteralAsCondition < Base
         include RangeHelp
+        extend AutoCorrector
 
         MSG = 'Literal `%<literal>s` appeared as a condition.'
         RESTRICT_ON_SEND = [:!].freeze
 
+        # rubocop:disable Metrics/AbcSize
+        def on_and(node)
+          if node.lhs.truthy_literal? && node.rhs.truthy_literal?
+            add_offense(node) do |corrector|
+              corrector.replace(node, 'true')
+            end
+          elsif node.lhs.truthy_literal?
+            add_offense(node.lhs) do |corrector|
+              corrector.replace(node, node.rhs.source)
+            end
+          elsif node.rhs.truthy_literal?
+            add_offense(node.rhs) do |corrector|
+              corrector.replace(node, node.lhs.source)
+            end
+          end
+        end
+        # rubocop:enable Metrics/AbcSize
+
         def on_if(node)
-          check_for_literal(node)
+          cond = condition(node)
+
+          if node.unless?
+            correct_if_node(node, cond, true) if cond.falsey_literal?
+            correct_if_node(node, cond, false) if cond.truthy_literal?
+          else
+            correct_if_node(node, cond, true) if cond.truthy_literal?
+            correct_if_node(node, cond, false) if cond.falsey_literal?
+          end
         end
 
         def on_while(node)
-          return if condition(node).true_type?
+          return if node.condition.source == 'true'
 
-          check_for_literal(node)
+          if node.condition.truthy_literal?
+            add_offense(node.condition) do |corrector|
+              corrector.replace(node.condition, 'true')
+            end
+          elsif node.condition.falsey_literal?
+            add_offense(node.condition) do |corrector|
+              corrector.remove(node)
+            end
+          end
         end
-        alias on_while_post on_while
+
+        # rubocop:disable Metrics/AbcSize
+        def on_while_post(node)
+          return if node.condition.source == 'true'
+
+          if node.condition.truthy_literal?
+            add_offense(node.condition) do |corrector|
+              corrector.replace(node, node.source.sub(node.condition.source, 'true'))
+            end
+          elsif node.condition.falsey_literal?
+            add_offense(node.condition) do |corrector|
+              corrector.replace(node, node.body.child_nodes.map(&:source).join("\n"))
+            end
+          end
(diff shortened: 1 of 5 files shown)


Zopolis4:
Tests do not cover the post-loop while case as I was unable to prevent it from also checking post-loop until's, which do not autocorrect as all the literals tested by the spec are truthy literals.

spec/rubocop/cop/lint/literal_as_condition_spec.rb
```diff
@@ -9,6 +9,10 @@
           top
         end
       RUBY
+
+      expect_correction(<<~RUBY)
```
Earlopain, line 13: Let's add `expect_no_corrections` to the other cases to confirm nothing unexpected is going on.

Earlopain (commented):
I believe it makes sense to mark this one as `Autocorrect: contextual` in order to prevent autosave in editors removing the code. I do this relativly often while testing things and I don't want things to shift around during that.

Zopolis4:
Good point-- I've added that.

Zopolis4:
The current autocorrection for code which has a redundant conditional check is just to remove the check-- right now nothing is being done to code which has a conditional check that will never pass (i.e. `if false`, `unless true`).

Thinking about it, that is unreachable code, and so the autocorrection would be just to delete the code entirely. Or maybe it should fall under the purview of `Lint/UnreachableCode`?

lib/rubocop/cop/lint/literal_as_condition.rb
```diff
@@ -98,13 +99,43 @@ def message(node)
...
+          if cond.truthy_literal?
+            autocorrect_redundant_truthy_condition(node, cond)
+          elsif cond.falsey_literal?
+            autocorrect_redundant_falsey_condition(node, cond)
           else
             check_node(cond)
           end
         end
 
+        def autocorrect_redundant_truthy_condition(node, cond)
```
Zopolis4, line 111: Splitting the autocorrection into `autocorrect_redundant_truthy_condition` and `autocorrect_redundant_falsey_condition` means there's some duplicate code shared between them, but unifying them triggers a lot of metrics and line length failures.
```ruby
def autocorrect_redundant_condition(node, cond)
  if (cond.truthy_literal? && node.if_type?) || (cond.falsey_literal? && node.if_type? && node.unless?)
    add_offense(cond) do |corrector|
      corrector.replace(node, node.if_branch.source)
    end
  elsif (cond.truthy_literal? && (node.while_type? || node.while_post_type?)) || (cond.falsey_literal? && (node.until_type? || node.until_post_type?))
    add_offense(cond) do |corrector|
      corrector.replace(cond, cond.truthy_literal?.to_s)
    end
  else
    add_offense(cond)
  end
end
```

lib/rubocop/cop/lint/literal_as_condition.rb
```diff
@@ -98,13 +99,43 @@ def message(node)
 
         def check_for_literal(node)
           cond = condition(node)
-          if cond.literal?
-            add_offense(cond)
+          if cond.truthy_literal?
+            autocorrect_redundant_truthy_condition(node, cond)
+          elsif cond.falsey_literal?
+            autocorrect_redundant_falsey_condition(node, cond)
```
koic, line 105: Can you add test cases for handling falsey literals?

Zopolis4: Added.

bbatsov:
> Thinking about it, that is unreachable code, and so the autocorrection would be just to delete the code entirely. Or maybe it should fall under the purview of Lint/UnreachableCode?

I'd say - `Lint/UnreachableCode`. It's good to keep cop focus relatively narrow.

Zopolis4:
Ok, I'll handle that cop in a second PR.

Zopolis4:
Although, implementing this further, it does seem to make more sense to have the autocorrect logic in the same place-- right now I'm just raising an offense to maintain behaviour, but the testing and the logic setup are much better suited here than they are in `Lint/UnreachableCode`, even if the goals of the cops line up a little less.

As it stands I'll keep things as they are, but unless anyone has strong objections I'll probably add autocorrect for `if false` and the like in a few days here.

spec/rubocop/cop/lint/literal_as_condition_spec.rb
```diff
@@ -2,49 +2,119 @@
...
         if %{lit}
            ^{lit} Literal `#{lit}` appeared as a condition.
           top
         end
       RUBY
+
+      expect_correction(<<~RUBY)
+        top
+      RUBY
+    end
```
dvandersluis, line 16: I think that `#{a}` should not be included in the correctable cases because unlike the other literals, removing it will potentially stop calling a method which could have side effects.

spec/rubocop/cop/lint/literal_as_condition_spec.rb
```diff
@@ -2,49 +2,119 @@
...
-        end while(%{lit})
+        end while %{lit}
                   ^{lit} Literal `#{lit}` appeared as a condition.
       RUBY
+
+      expect_correction(<<~RUBY)
+        while true
+          top
+        end
+      RUBY
```
dvandersluis, line 91: Should the form of loop change here? I think that maybe should be the domain of a separate Style cop.

Zopolis4: Fair enough. Fixed.

bbatsov:
I had kind of forgotten about this PR, but it seems to be in a good shape. Thanks, @Zopolis4!

PR #13119, opened 2024-08-14
[Fix #13118] Fix a false positive for `Style/MapIntoArray` when splatting

### LINKED ISSUE #13118: "Style/MapIntoArray" false positive with push(*ary)
I have the following code:

```ruby
a = [[1,2,3],[4,5,6]]
b = []

a.each do |x|
  b.push(*x)
end

puts a.inspect  #=> [[1, 2, 3], [4, 5, 6]]
puts b.inspect  #=> [1, 2, 3, 4, 5, 6]
```

RuboCop complains about this with `Style/MapIntoArray`, and auto-correct creates invalid code:

```
b = a.map do |x|
  *x
end
```

--------

## Expected behavior

Have `Style/MapIntoArray` enabled but not report this as an offense (or to produce correct code on auto-correct).

## Actual behavior

Reports as an offense.

(Explained above.)

## RuboCop version

```
$ rubocop -V
1.65.1 (using Parser 3.3.4.2, rubocop-ast 1.32.0, running on ruby 3.0.2) [x86_64-linux]

$ ruby -v
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux]
```

Earlopain:
Fixes #13118

A proper correction would look something like this:
```rb
b = a.map do |x|
  x
end.flatten
```

which I don't think makes much sense for this cop.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positive_for_style_map_into_array.md
+++ b/changelog/fix_false_positive_for_style_map_into_array.md
@@ -1,0 +1,1 @@
+* [#13118](https://github.com/rubocop/rubocop/issues/13118): Fix a false positive for `Style/MapIntoArray` when splatting. ([@earlopain][])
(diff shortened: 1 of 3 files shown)


koic:
I agree with this approach. Thanks!

Issue #13063, opened 2024-07-24
Cop idea: Implicit return in predicate method


Earlopain:
Ruby returns the value of the last evaluated expression. Predicate methods should return `true`/`false` but implicit returns can result in `nil` being the result.

## Expected behavior

Offense for the following

### case/when
```rb
# bad
def foo?
  case foo
  when bar
     true
  end
end

# good
def foo?
  case foo
  when bar
     true
  else
    false
  end
end
```

### inline if/unless

```rb
# bad
def foo?
  false if bar?
end


# good
def foo?
  return false if bar?

  true
end
```

### multiline if/unless

```rb
# bad
def foo?
  if bar?
    baz
    true
  end
end


# good
def foo?
  if bar?
    baz
    true
  else
    false
  end
end
```

koic:
`Style/ReturnNilInPredicateMethodDefinition` is intended to check only the `return` keyword. So, this proposal doesn't align with the role of `Style/ReturnNilInPredicateMethodDefinition`. Checking the last evaluated value should be the role of a different cop.

Earlopain:
I don't think I agree with your opinion. While the cop may only check for explicit returns right now, the effect is the same: a predicate method returns nil with this code.

koic:
I understand what you are trying to convey, but as the designer of this cop, I can say that each cop should be simple and should not be assigned more roles than necessary. This cop is specifically focused on the style of `return` (not the last evaluated value) , and its cop name is intentionally reflective of that purpose.
https://docs.rubocop.org/rubocop/1.65/cops_style.html#stylereturnnilinpredicatemethoddefinition

Earlopain:
Ok, I merged this with the other one and made it a feature request instead.

dvandersluis:
I didn't realize this issue existed but I opened #13305 to handle one implicit case.

PR #13125, opened 2024-08-15
Update `Style/ReturnNilInPredicateMethodDefinition` to recognize implicit `nil` returns in `rescue` and `ensure`

vlad-pisanov:
`Style/ReturnNilInPredicateMethodDefinition` already recognizes implicit `nil` returns from predicates:
```ruby
def foo?
  nil # ⚠️ existing offense: Return false instead of nil in predicate methods
end
```

This PR expands implicit `nil` detection to predicates with `rescue` and `ensure` blocks. New cases detected:
```ruby
def foo?
  nil # ⚠️ new offense: Return false instead of nil in predicate methods
rescue
end
```

```ruby
def foo?
  nil # ⚠️ new offense: Return false instead of nil in predicate methods
ensure
end
```

```ruby
def foo?
rescue
  nil # ⚠️ new offense: Return false instead of nil in predicate methods
end
```
-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/return_nil_in_predicate_method_definition_spec.rb
+++ b/spec/rubocop/cop/style/return_nil_in_predicate_method_definition_spec.rb
@@ -74,6 +74,89 @@ RSpec.describe RuboCop::Cop::Style::ReturnNilInPredicateMethodDefinition, :confi
       RUBY
     end
 
+    it 'registers an offense when using `nil` before `rescue`' do
+      expect_offense(<<~RUBY)
+        def foo?
+          bar
+          nil
+          ^^^ Return `false` instead of `nil` in predicate methods.
+        rescue
+        end
+      RUBY
+
+      expect_correction(<<~RUBY)
+        def foo?
+          bar
+          false
+        rescue
+        end
+      RUBY
+    end
+
+    it 'registers an offense when using `nil` before `ensure`' do
+      expect_offense(<<~RUBY)
+        def foo?
+          bar
+          nil
+          ^^^ Return `false` instead of `nil` in predicate methods.
+        ensure
+        end
+      RUBY
+
+      expect_correction(<<~RUBY)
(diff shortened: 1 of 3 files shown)


Earlopain:
I openend #13063 as a feature request for this cop but @koic felt like it would be more appropriate as a new cop instead. I'm not sure of the name (maybe just `ImplicitReturnNilInPredicateMethodDefinition`?) but conceptually this looks like a good start.

spec/rubocop/cop/style/return_nil_in_predicate_method_definition_spec.rb
```diff
@@ -74,6 +74,108 @@ def foo?
...
+      RUBY
+    end
+
+    it 'registers an offense when using `nil` inside `ensure`' do
+      expect_offense(<<~RUBY)
+        def foo?
+        ensure
+          bar
+          nil
+          ^^^ Return `false` instead of `nil` in predicate methods.
```
Earlopain, line 121: I don't believe that ensure will ever return a value without an explicit return. This method would return `nil` from the empty block above.

vlad-pisanov: good catch! I'm surprised `Lint/Void` doesn't pick up on this, sounds like a false negative 👀

vlad-pisanov:
@Earlopain I'm with you -- since this cop already detects the basic implicit `nil` return scenario, it makes sense to me to expand this cop to handle more implicit return scenarios.

spec/rubocop/cop/style/return_nil_in_predicate_method_definition_spec.rb
bbatsov: Indeed! We need to improve `Lint/Void` a bit. :-)

PR #13134, opened 2024-08-18
Fix false negative for `Lint/Void` when using using frozen literals

vlad-pisanov:
Currently, `Lint/Void` looks for literals, and arrays/hashes composed of literals. However, returning a _frozen_ literal into a void context is just as pointless as returning a plain literal (and more expensive). This PR adds support for frozen literal detection.

```ruby
def foo
  'hello'.freeze      # ⚠️ new offense: Literal 'hello'.freeze used in a void context
  [1,2,3].freeze      # ⚠️ new offense: Literal [1,2,3].freeze used in a void context
  [1, 'hello'.freeze] # ⚠️ new offense: Literal [1,'hello'.freeze] used in a void context
  baz
end
```

NOTE: this isn't the same as `Style/RedundantFreeze` which doesn't detect nested structures

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/fix_frozen_literal_detection_in_lint_void.md
+++ b/changelog/fix_frozen_literal_detection_in_lint_void.md
@@ -1,0 +1,1 @@
+* [#13134](https://github.com/rubocop/rubocop/pull/13134): Fix false negative for `Lint/Void` when using using frozen literals. ([@vlad-pisanov][])
--- a/lib/rubocop/cop/lint/void.rb
+++ b/lib/rubocop/cop/lint/void.rb
@@ -245,14 +245,22 @@ module RuboCop
         def entirely_literal?(node)
           case node.type
           when :array
-            node.each_value.all? { |value| entirely_literal?(value) }
+            all_values_entirely_literal?(node)
           when :hash
-            return false unless node.each_key.all? { |key| entirely_literal?(key) }
-
-            node.each_value.all? { |value| entirely_literal?(value) }
+            all_keys_entirely_literal?(node) && all_values_entirely_literal?(node)
+          when :send, :csend
+            node.method?(:freeze) && node.receiver && entirely_literal?(node.receiver)
           else
             node.literal?
           end
+        end
+
+        def all_keys_entirely_literal?(node)
+          node.each_key.all? { |key| entirely_literal?(key) }
+        end
+
+        def all_values_entirely_literal?(node)
+          node.each_value.all? { |value| entirely_literal?(value) }
         end
       end
     end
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/lint/void.rb
```diff
@@ -245,15 +245,23 @@ def autocorrect_nonmutating_send(corrector, node, suggestion)
         def entirely_literal?(node)
           case node.type
           when :array
-            node.each_value.all? { |value| entirely_literal?(value) }
+            all_values_entirely_literal?(node)
```
vlad-pisanov, line 248: Some refactoring to keep `Metrics` happy

lib/rubocop/cop/lint/void.rb
```diff
@@ -245,15 +245,23 @@ def autocorrect_nonmutating_send(corrector, node, suggestion)
...
           when :hash
-            return false unless node.each_key.all? { |key| entirely_literal?(key) }
-
-            node.each_value.all? { |value| entirely_literal?(value) }
+            all_keys_entirely_literal?(node) && all_values_entirely_literal?(node)
+          when :send
```
koic, line 251: The same applies to the safe navigation operator. Can you add a test case for it as well?

```suggestion
          when :send, :csend
```

lib/rubocop/cop/lint/void.rb
```diff
@@ -245,15 +245,23 @@ def autocorrect_nonmutating_send(corrector, node, suggestion)
...
+            node.method?(:freeze) && node.receiver && entirely_literal?(node.receiver)
```
Earlopain, line 252: [itself](https://docs.ruby-lang.org/en/3.3/Object.html#method-i-itself) looks like a good conditate to add as well. There are others I would reasonably include as well, like `hash`, but they might have side-effects

changelog/fix_frozen_literal_detection_in_lint_void.md
```diff
@@ -0,0 +1 @@
+* [#13134](https://github.com/rubocop/rubocop/pull/13134): Fix false negative for `Lint/Void` when using using frozen literals. ([@vlad-pisanov][])
```
Earlopain, line 1: I don't think this is entirely correct. These are not literals anymore, I think better would be something like "when calling `freeze` on literals`.

The offense message for these is also a bit meh but I wouldn't bother changing that.

lib/rubocop/cop/lint/void.rb
vlad-pisanov: Actually, almost any instance method applied to a literal yields a "throwaway" result in this context (`true.to_s`, `123.to_f`, `[1,2,3].sum`, etc.)

This could be a whole new PR 👀

bbatsov:
Thanks!

PR #13135, opened 2024-08-19
[Fix #13103] Make `Lint/UselessAssignment` autocorrection safe

### LINKED ISSUE #13103: Do not do unsafe autocorrection for `Lint/UselessAssignment`
The cop is already marked as unsafe with the example using `||=` provided and the problematic variable declaration. 
But, there is another example I just encountered.

```ruby
foo += bar ? 1 : 2 
```

Which is incorrectly gets autocorrected to `foo + bar ? 1 : 2` (which is equivalent to `(foo + bar) ? 1 : 2`, but should be autocorrected to something like `foo + (bar ? 1 : 2)` (note the operator precedences in both examples).

What if we stop autocorrecting for such cases (and for cases like mentioned in the "safety" doc-block)?

koic:
Fixes #13103.

This PR makes `Lint/UselessAssignment` autocorrection safe.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/useless_assignment.rb
+++ b/lib/rubocop/cop/lint/useless_assignment.rb
@@ -16,14 +16,13 @@ module RuboCop
       # reassignments and properly handles varied cases such as branch, loop,
       # rescue, ensure, etc.
       #
+      # This cop's autocorrection avoids cases like `a ||= 1` because removing assignment from
+      # operator assignment can cause NameError if this assignment has been used to declare
+      # a local variable. For example, replacing `a ||= 1` with `a || 1` may cause
+      # "undefined local variable or method `a' for main:Object (NameError)".
+      #
       # NOTE: Given the assignment `foo = 1, bar = 2`, removing unused variables
       # can lead to a syntax error, so this case is not autocorrected.
-      #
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/lint/useless_assignment.rb
```diff
@@ -53,24 +52,27 @@ def after_leaving_scope(scope, _variable_table)
...
-            next if assignment.used? || part_of_ignored_node?(assignment.node)
+            assignment_node = assignment.node
+            next if assignment.used? || part_of_ignored_node?(assignment_node)
 
             message = message_for_useless_assignment(assignment)
             range = offense_range(assignment)
 
             add_offense(range, message: message) do |corrector|
-              autocorrect(corrector, assignment) unless sequential_assignment?(assignment.node)
+              if !sequential_assignment?(assignment_node) && !assignment_node.parent&.or_asgn_type?
```
bbatsov, line 67: Might be good to add some comment here explaining what we're skipping for the auto-correction.

koic: Sure! I've added code comment for the logic.

PR #13141, opened 2024-08-19
Load only project specific config for `InternalAffairs/UndefinedConfig`

Earlopain:
Going through the default config is fine for RuboCop itself but extensions need to inject their config for RuboCop to pick them up. This makes it necessary to load the extension, even if the extension itself doesn't make use of its own cops, like `rubocop-rails` or `rubocop-capybara`.

Instead just load `config/default.yml` from `Dir.pwd` in isolation and use just that. This would make https://github.com/rubocop/rubocop-capybara/pull/133 and https://github.com/rubocop/rubocop-rails/commit/288c7ce265d7fd0f5a244c6887224fc975ff2eb9 unnecessary. I also tested with https://github.com/rubocop/rubocop-rails/commit/9b450a8da5d5ff3fb73447a31ea09738f88dd6e4 and observed no issues with `bundle exec rake`.

-----------------

Not sure how to write proper tests for this but I tested manually against most first party extensions and it behaves as expected.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/internal_affairs/undefined_config.rb
+++ b/lib/rubocop/cop/internal_affairs/undefined_config.rb
@@ -5,12 +5,22 @@ module RuboCop
     module InternalAffairs
       # Looks for references to a cop configuration key that isn't defined in config/default.yml.
       class UndefinedConfig < Base
+        # `FileFinder` is a private API not intended to be used by cops,
+        # but it's fine for `InternalAffairs`.
+        extend FileFinder
+
         ALLOWED_CONFIGURATIONS = %w[
           Safe SafeAutoCorrect AutoCorrect Severity StyleGuide Details Reference Include Exclude
         ].freeze
         RESTRICT_ON_SEND = %i[[] fetch].freeze
         MSG = '`%<name>s` is not defined in the configuration for `%<cop>s` ' \
               'in `config/default.yml`.'
+        CONFIG_PATH = find_file_upwards('config/default.yml', Dir.pwd)
+        CONFIG = if CONFIG_PATH
+                   ConfigLoader.load_yaml_configuration(CONFIG_PATH)
+                 else
+                   {}
+                 end
 
         # @!method cop_class_def(node)
         def_node_search :cop_class_def, <<~PATTERN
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/internal_affairs/undefined_config.rb
```diff
@@ -5,12 +5,22 @@ module Cop
     module InternalAffairs
       # Looks for references to a cop configuration key that isn't defined in config/default.yml.
       class UndefinedConfig < Base
+        class << self
+          include FileFinder
+        end
```
koic, line 10:
```suggestion
        extend FileFinder
```

koic: `FileFinder` is a private API:
https://github.com/Earlopain/rubocop/blob/58eeb7a29ed6c35d111fb4f264c364261dc7ecbb/lib/rubocop/file_finder.rb#L7

Although this is an `InternalAffairs` cop, so the risk of issues arising might be low, it might be better to leave a comment in the source code noting that a private API is being used.

Earlopain: I added a new test to `file_finder` to exercise it a bit more, I don't think it was being used in that way yet (looking for a file in a directory)

lib/rubocop/cop/internal_affairs/undefined_config.rb
```diff
@@ -5,12 +5,21 @@ module Cop
     module InternalAffairs
       # Looks for references to a cop configuration key that isn't defined in config/default.yml.
       class UndefinedConfig < Base
+        # FileFinder is private API not expected to be used by cops but its fine for InternalAffairs
```
koic, line 8:
```suggestion
        # `FileFinder` is a private API not intended to be used by cops, but it's fine for `InternalAffairs`.
```

Earlopain: Put it on two lines because of LineLength

koic:
Let's see how it goes with this. Thanks!

PR #13151, opened 2024-08-22
[Fix #13150] Allow `get_!`, `set_!`, `get_?`, `set_?`, `get_=`, and `set_=` in `Naming/AccessorMethodName`

### LINKED ISSUE #13150: Naming/AccessorMethodName could be more tunable
Hi!

We have a method named `get_min!`.  It has a bang because it's not just an accessor: it has side-effects (and does return a result).

I think that Naming/AccessorMethodName should not fire in this case (names ending with a bang).

We also have other `get_foo` methods, where we use `get_` as a hint that they are costly.  They are not just as cheap as accessors.  It would be nice if Naming/AccessorMethodName had some `AllowedMethods` or `AllowedNames` support for such cases.

Of course we can use `# rubocop:disable`, but:
- for `get_min!`, it does not look like a genuine accessor, IMHO it should not fire at all
- for other the cases, it's just nicer to avoid cluttering the source with magic comments, but it's also less precise (it is project-wise), so I can see reason to reject this.

Thanks!

koic:
Fixes #13150.

This PR allows `get_!`, `set_!`, `get_?`, `set_?`, `get_=`, and `set_=` method names in `Naming/AccessorMethodName`.

Names ending with `!`, `?`, or `=` are not appropriate as attribute names.

```console
$ ruby -e 'class X attr_reader :x! end'
-e:1:in 'Module#attr_reader': invalid attribute name 'x!' (NameError)

class X attr_reader :x! end
        ^^^^^^^^^^^
        from -e:1:in '<class:X>'
        from -e:1:in '<main>'
```

Therefore, they don't need to be recognized as intended accessor method names.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_allow_invalid_attr_name_in_naming_accessor_method_name.md
+++ b/changelog/change_allow_invalid_attr_name_in_naming_accessor_method_name.md
@@ -1,0 +1,1 @@
+* [#13050](https://github.com/rubocop/rubocop/issues/13050): Allow `get_!`, `set_!`, `get_?`, `set_?`, `get_=`, and `set_=` in `Naming/AccessorMethodName`. ([@koic][])
--- a/lib/rubocop/cop/naming/accessor_method_name.rb
+++ b/lib/rubocop/cop/naming/accessor_method_name.rb
@@ -40,6 +40,7 @@ module RuboCop
         MSG_WRITER = 'Do not prefix writer method names with `set_`.'
 
         def on_def(node)
+          return unless proper_attribute_name?(node)
           return unless bad_reader_name?(node) || bad_writer_name?(node)
 
           message = message(node)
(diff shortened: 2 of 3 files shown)


akimd:
FTR, I think you should also include `=`.

koic:
@akimd Sure! I've updated this PR.

bbatsov:
I get allowing `!` for assignment and `=` for setters, but I don't get the other cases. Why would someone name methods in such a way - it doesn't make any sense to me.

akimd:
I think we wouldn't use them.  But the point here is that this cop promotes using regular attribute syntax for accessors, and these special cases aren't accessors.  For some reason someone wants to flag them for some specific weirdness (like my original `get_min!` which is costly and I didn't want to look innocuous).

In fact I think the title of this MR is slightly misleading, it is not about "allowing them as accessor names", but rather "stop considering them as accessor names".

bbatsov:
> In fact I think the title of this MR is slightly misleading, it is not about "allowing them as accessor names", but rather "stop considering them as accessor names".

Good point - admittedly I was mislead by the title of the PR.

PR #13157, opened 2024-08-23
Resolve ruby-head uri deprecation warning

Earlopain:
> URI::RFC3986_PARSER.make_regexp is obsoleted. Use URI::RFC2396_PARSER.make_regexp explicitly

The `uri` gem is switching over its default parser from RFC 2396 to 3986. The new parser doesn't have `make_regexp` available but for compatibility a fallback was added.

Additionally, old versions of `uri` don't have the RFC2396 parser available. See https://github.com/ruby/uri/issues/118

I guess, for Ruby 3.4, `Lint/UriRegexp` needs a change because as it is now it provides bad directions.

### DIFF
--- a/lib/rubocop/config_loader_resolver.rb
+++ b/lib/rubocop/config_loader_resolver.rb
@@ -239,8 +239,7 @@ module RuboCop
     end
 
     def remote_file?(uri)
-      regex = URI::DEFAULT_PARSER.make_regexp(%w[http https])
-      /\A#{regex}\z/.match?(uri)
+      uri.start_with?('http://', 'https://')
     end
 
     def handle_disabled_by_default(config, new_default_configuration)
--- a/lib/rubocop/cop/mixin/line_length_help.rb
+++ b/lib/rubocop/cop/mixin/line_length_help.rb
@@ -91,8 +91,13 @@ module RuboCop
       end
 
       def uri_regexp
-        @uri_regexp ||=
-          URI::DEFAULT_PARSER.make_regexp(config.for_cop('Layout/LineLength')['URISchemes'])
+        @uri_regexp ||= begin
+          # Ruby 3.4 changes the default parser to RFC3986 which warns on make_regexp.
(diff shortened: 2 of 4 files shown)


lib/rubocop/config_loader_resolver.rb
```diff
@@ -239,8 +239,7 @@ def inherited_file(path, inherit_from, file)
     end
 
     def remote_file?(uri)
-      regex = URI::DEFAULT_PARSER.make_regexp(%w[http https])
-      /\A#{regex}\z/.match?(uri)
+      uri.start_with?('http://', 'https://')
```
Earlopain, line 242: Seems like a simpler and sufficient solution, though I can do the `defined` check here as well.

bbatsov: Might be good to check what prompted the more complex solution here in the first place. :-) I assume the simpler option will work for most people, but I want to be careful not to break someone's setup.

Earlopain: I can do some spelunking in a bit. My initial assumption is that it would be broken anyways, since uri-like but invalid uris would fall through being loaded from disk. I haven't confirmed it though.

Earlopain: Right, so initially this was a bit simpler with just `URI.regexp` but over time evolved into this because of Ruby changes and a bugfix https://github.com/rubocop/rubocop/commit/45a06913a56e4704a4e1269216d87613a6cb131e

It does indeed fall down to trying to load it from disk, which then fails. These files will never be valid because of file naming restrictions around `:` and `/`, so the only change will be a different error message if you do end up writing an invalid uri.

On that note, I improved the error message and added a test for it:
Previous:
> Configuration file not found: /home/user/code/rubocop/https:/invalid_üri/.rubocop.yml

New:
> Failed to resolve configuration: 'https://invalid_üri/.rubocop.yml' is not a valid URI

Because it now no longer falls through.

bbatsov:
OK, you have me convinced! Thanks!

PR #13163, opened 2024-08-28
Fix readme CI badge

Earlopain:
It's always green, even when CI fails. This new one is copied from the actions tab

And also the opencollective badges. I broke those by changing the heading text


bbatsov:
Thanks!

PR #13166, opened 2024-08-29
Add `MixedOffenseSafety` mixin to support both safe and unsafe offenses in a cop, extend `Style/FileEmpty` cop to take advantage of this.

Zopolis4:
In some cops, we know that certain offenses are guaranteed to be safe even if the overall cop produces unsafe results. 

Adding the `MixedOffenseSafety` mixin allows us to access these safe offenses in contexts where this is required, while still maintaining the overall safety of the cop.

I'm using the `Style/FileEmpty` cop as a simple example of how to support mixed safety.

-----------------

* [x] Added tests.


koic:
Safe is a boolean, so I don't think it should be extended to a new three-valued type like true, false, and another value. 
Also, when considering a global value, it is necessary to take into account the overall context, not just a specific cop.
I understand the issue at hand, but it may be necessary to consider a better design.

Zopolis4:
> Safe is a boolean, so I don't think it should be extended to a new three-valued type like true, false, and another value.

I agree the configuration could be improved, this was just the best way I could think of. Open to suggestions.

> Also, when considering a global value, it is necessary to take into account the overall context, not just a specific cop.

I believe there are additional cops that could benefit from this, so its not just for `Style/FileEmpty`.

koic:
This PR helps to understand the issues as a proof of concept (PoC), but unfortunately, it is difficult to integration under the current approach. I cannot immediately address this issue, but when I do, I will consider a solution policy from scratch (this PR helped to understand the issue). I will close this PR, but please feel free to open an issue outlining the problem, or if you come up with a better approach, you can reopen and continue. Thank you.

PR #13170, opened 2024-08-29
Remove `benchmark` dependency

Earlopain:
Ruby wants to bundle it: https://github.com/ruby/ruby/pull/11492

This is pretty much all the gem does anyways (that's being used here): https://github.com/ruby/benchmark/blob/8d2c8a0d7ca72bff7e97a4800e16c72f232867e4/lib/benchmark.rb#L311-L315

I removed the `bench_cop` rake task since it uses `benchmark` and I believe it just isn't very useful. Dedicated tools like `vernier` or `stackprof` (which has buildin support in RuboCop these days) are much more suited for the task.

-----------------

* [ ] Added tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -32,49 +32,6 @@ task default: %i[documentation_syntax_check spec prism_spec internal_investigati
 require 'yard'
 YARD::Rake::YardocTask.new
 
-desc 'Benchmark a cop on given source file/dir'
-task :bench_cop, %i[cop srcpath times] do |_task, args|
-  require 'benchmark'
-  require 'rubocop'
-  include RuboCop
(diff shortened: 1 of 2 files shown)


bbatsov:
Did you make sure the tasks isn't mentioned somewhere in the docs? I recall it was added to simplify profiling cops, but I'm not sure if we ever documented it.

Earlopain:
Yeah, I did a search and found no results: https://github.com/search?q=org%3Arubocop%20bench_cop&type=code

bbatsov:
Ah, yeah - I just realized I was thinking about the memory profiler functionality. (`rubocop --profile`)

PR #13173, opened 2024-08-30
Fix false negatives for `Style/EmptyLiteral`

vlad-pisanov:
Improve `Style/EmptyLiteral` to detect other common ways of creating empty literals.

```ruby
Array.new      # ✔️ existing offense
Hash.new       # ✔️ existing offense

Array[]        # ⚠️ new offense
Array([])      # ⚠️ new offense
Array.new([])  # ⚠️ new offense
Hash[]         # ⚠️ new offense
Hash([])       # ⚠️ new offense
Hash.new([])   # ⚠️ new offense
```

Also: improve offense message by including the actual offense source

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/empty_literal.rb
+++ b/lib/rubocop/cop/style/empty_literal.rb
@@ -9,7 +9,9 @@ module RuboCop
       # @example
       #   # bad
       #   a = Array.new
+      #   a = Array[]
       #   h = Hash.new
+      #   h = Hash[]
       #   s = String.new
       #
       #   # good
@@ -22,17 +24,17 @@ module RuboCop
         include StringLiteralsHelp
         extend AutoCorrector
 
-        ARR_MSG = 'Use array literal `[]` instead of `Array.new`.'
-        HASH_MSG = 'Use hash literal `{}` instead of `Hash.new`.'
+        ARR_MSG = 'Use array literal `[]` instead of `%<current>s`.'
+        HASH_MSG = 'Use hash literal `{}` instead of `%<current>s`.'
         STR_MSG = 'Use string literal `%<prefer>s` instead of `String.new`.'
 
-        RESTRICT_ON_SEND = %i[new].freeze
+        RESTRICT_ON_SEND = %i[new [] Array Hash].freeze
 
         # @!method array_node(node)
-        def_node_matcher :array_node, '(send (const {nil? cbase} :Array) :new)'
+        def_node_matcher :array_node, '(send (const {nil? cbase} :Array) :new (array)?)'
 
         # @!method hash_node(node)
-        def_node_matcher :hash_node, '(send (const {nil? cbase} :Hash) :new)'
+        def_node_matcher :hash_node, '(send (const {nil? cbase} :Hash) :new (array)?)'
 
         # @!method str_node(node)
         def_node_matcher :str_node, '(send (const {nil? cbase} :String) :new)'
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/style/empty_literal.rb
```diff
@@ -60,9 +78,11 @@ def on_send(node)
 
         def offense_message(node)
           if offense_array_node?(node)
-            ARR_MSG
+            current = array_with_index(node) ? 'Array[]' : 'Array.new'
+            format(ARR_MSG, current: current)
```
koic, line 82: Can you update as below and adjust the specs to have more appropriate message?

```suggestion
            format(ARR_MSG, current: node.source)
```

lib/rubocop/cop/style/empty_literal.rb
```diff
@@ -60,9 +78,11 @@ def on_send(node)
...
           elsif offense_hash_node?(node)
-            HASH_MSG
+            current = hash_with_index(node) ? 'Hash[]' : 'Hash.new'
+            format(HASH_MSG, current: current)
```
koic, line 85: Ditto.

```suggestion
            format(HASH_MSG, current: node.source)
```

spec/rubocop/cop/style/empty_literal_spec.rb
```diff
@@ -2,37 +2,26 @@
...
-    it 'registers an offense for ::Array.new' do
-      expect_offense(<<~RUBY)
-        test = ::Array.new
-               ^^^^^^^^^^^ Use array literal `[]` instead of `Array.new`.
-      RUBY
-
-      expect_correction(<<~RUBY)
-        test = []
-      RUBY
+    context 'initializer resulting in an empty array literal' do
```
vlad-pisanov, line 18: combined existing and new cases into shared examples for readability

spec/rubocop/cop/style/empty_literal_spec.rb
```diff
@@ -2,37 +2,26 @@
...
+                 #{'^' * initializer.length} Use array literal `[]` instead of `#{initializer}`.
+        RUBY
 
-      expect_correction(<<~RUBY)
-        test = []
-      RUBY
+        expect_correction(<<~RUBY)
+          test = []
+        RUBY
+      end
```
koic, line 15: `shared_examples` don't always improve readability and reproducibility. Could you expand them into individual tests, as in the original specs?

spec/rubocop/cop/style/empty_literal_spec.rb
koic: Thank you for being considerate. However, as I mentioned in the comment below, `shared_examples` do not always improve readability. Can you respect the existing spec descriptions?

bbatsov: I think in that particular case the shared examples are OK, given that we're essentially testing 6 times for the same thing.

franzliedke:
@vlad-pisanov @bbatsov Why is `Hash.new([])` an offense? That parameter is the *default value* for unknown keys. (It's better to do this with the block form, but that's a different discussion.) That's *not* the same as a literal `{}`, or what am I missing?

bbatsov:
@franzliedke That's an oversight that's fixed on `master`.

bbatsov:
See https://github.com/rubocop/rubocop/pull/13178

PR #13175, opened 2024-08-30
Replace forwardable with explicit delegation for performance

Earlopain:
I'm doing performance profiling and was surprised for these to appear at all in the samples. Here are some numbers (3 warmup runs, 20 measured runs, cache used)::

```
# with forwardable in both rubocop + rubocop-ast
$ hyperfine -w 3 -r 20 "bundle exec rubocop Rakefile"
  Time (mean ± σ):      1.169 s ±  0.016 s    [User: 0.976 s, System: 0.190 s]
  Range (min … max):    1.138 s …  1.198 s    20 runs

# with forwardable only in rubocop-ast
$ hyperfine -w 3 -r 20 "bundle exec rubocop Rakefile"
  Time (mean ± σ):      1.145 s ±  0.011 s    [User: 0.959 s, System: 0.185 s]
  Range (min … max):    1.131 s …  1.170 s    20 runs

# with forwardable in neither
$ hyperfine -w 3 -r 20 "bundle exec rubocop Rakefile"
  Time (mean ± σ):      1.136 s ±  0.020 s    [User: 0.950 s, System: 0.183 s]
  Range (min … max):    1.098 s …  1.168 s    20 runs
```

Overall ~ 3% faster

-------

Full run, also about 3%:

```
# with forwardable in both rubocop + rubocop-ast
$ hyperfine -w 3 -r 20 "bundle exec rubocop"
Benchmark 1: bundle exec rubocop
  Time (mean ± σ):      1.704 s ±  0.018 s    [User: 1.903 s, System: 1.004 s]
  Range (min … max):    1.669 s …  1.742 s    20 runs

# with forwardable only in rubocop-ast
$ hyperfine -w 3 -r 20 "bundle exec rubocop"
Benchmark 1: bundle exec rubocop
  Time (mean ± σ):      1.677 s ±  0.025 s    [User: 1.847 s, System: 0.988 s]
  Range (min … max):    1.640 s …  1.763 s    20 runs

# with forwardable in neither
$ hyperfine -w 3 -r 20 "bundle exec rubocop"
Benchmark 1: bundle exec rubocop
  Time (mean ± σ):      1.656 s ±  0.020 s    [User: 1.817 s, System: 0.984 s]
  Range (min … max):    1.630 s …  1.700 s    20 runs
```

Companion PR for `rubocop-ast`: https://github.com/rubocop/rubocop-ast/pull/312

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -9,7 +9,6 @@ require 'fileutils'
 before_us = $LOADED_FEATURES.dup
 require 'rainbow'
 
-require 'forwardable'
 require 'regexp_parser'
 require 'set'
 require 'unicode/display_width'
--- a/lib/rubocop/comment_config.rb
+++ b/lib/rubocop/comment_config.rb
@@ -4,8 +4,6 @@ module RuboCop
   # This class parses the special `rubocop:disable` comments in a source
   # and provides a way to check if each cop is enabled at arbitrary line.
   class CommentConfig
-    extend Forwardable
-
     CONFIG_DISABLED_LINE_RANGE_MIN = -Float::INFINITY
 
     # This class provides an API compatible with RuboCop::DirectiveComment
@@ -29,11 +27,17 @@ module RuboCop
 
     attr_reader :processed_source
 
-    def_delegators :@processed_source, :config, :registry
-
     def initialize(processed_source)
       @processed_source = processed_source
       @no_directives = !processed_source.raw_source.include?('rubocop')
+    end
+
+    def config
+      @processed_source.config
+    end
+
+    def registry
+      @processed_source.registry
     end
 
     def cop_enabled_at_line?(cop, line_number)
--- a/lib/rubocop/config.rb
+++ b/lib/rubocop/config.rb
@@ -12,7 +12,6 @@ module RuboCop
   class Config
     include PathUtil
     include FileFinder
-    extend Forwardable
 
     CopConfig = Struct.new(:name, :metadata)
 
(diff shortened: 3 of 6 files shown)


koic:
This might be sufficient if the abstraction provided for unknown methods is not needed.

marcandre:
So indeed it looks like `forwardable` [executes extra code](https://github.com/ruby/ruby/blob/master/lib/forwardable.rb#L203) than what is necessary.

1) Has there ever been a feature request opened on ruby-core for a `minimal: true` option or similar to `def_delegate`?
2) Why don't we instead make our own `forwardable` clone that only does the forwarding?

Earlopain:
> This might be sufficient if the abstraction provided for unknown methods is not needed.

Can you elaborate? I don't know what you mean with that.

> Has there ever been a feature request...

I searched through the bugtracker and haven't found something. It did remind me that ruby has thoughs about making `forwardable` a bundled gem though... I commented on the issue for that. Not a fan

> Why don't we instead make our own...

I honestly don't see much value in that. Few places really _need_ it, delegating numerous methods. Most just do one or two which I think is fine to expliciy type out. But if you insist I will amend the PR.

marcandre:
> suffici



> > Why don't we instead make our own...
> 
> I honestly don't see much value in that. Few places really _need_ it, delegating numerous methods. Most just do one or two which I think is fine to expliciy type out. But if you insist I will amend the PR.

I do find the older code clearer than the new one. It was using a simple abstraction explicitly. We already have the code (the `class_eval`), we just need to extract it to a helper function.

Earlopain:
> I do find the older code clearer

That's fair, I don't feel too strongly myself. I will amend the ast PR later when I got a bit more time and offer something similar here

marcandre:
Great, thank you so much @Earlopain . Let me know if you'd prefer I do it.

Earlopain:
Please do, you probably have something in mind already. I'm not sure where to put. You should be able to edit my branch if you want, or just make it from scratch if you prefer.

PR #13178, opened 2024-09-01
Fix false positive for `Style/EmptyLiteral` with `Hash.new([])`

Earlopain:
The first parameter for Hash.new is the default value:

```rb
Hash.new([])[:foo]
# => []
```

Side note: this usage is most likely a programming error. You want the block form where different keys _don't_ share the same underlying array.

-----------------

* [x] Added tests.


koic:
Yes, the compatibility with the behavior of `{}[:key]` needs to be maintained. Thank you!

bbatsov:
> Side note: this usage is most likely a programming error. You want the block form where different keys don't share the same underlying array.

Indeed. We should probably add a Lint check for this.

PR #13179, opened 2024-09-01
Fix an error for `InternalAffairs/NodeMatcherDirective` when no second argument

Earlopain:
There is code below that assumes the existance of two arguments

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/internal_affairs/node_matcher_directive.rb
+++ b/lib/rubocop/cop/internal_affairs/node_matcher_directive.rb
@@ -45,7 +45,7 @@ module RuboCop
         PATTERN
 
         def on_send(node)
-          return if node.arguments.none?
+          return unless node.arguments.count == 2
           return unless valid_method_name?(node)
 
           actual_name = node.first_argument.value.to_s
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/internal_affairs/node_matcher_directive.rb
```diff
@@ -103,6 +103,7 @@ def too_many_directives(node)
 
         def process_directive(node, actual_name, directive)
           return unless (offense_type = directive_offense_type(directive, actual_name))
+          return unless node.arguments.count == 2
```
bbatsov, line 106: I'd add a comment explaining this, as it's quite cryptic without one.

Earlopain: I didn't even add it add the proper place, oops. Right after `on_send` it doesn't need one I think but let me know if you still think otherwise.

koic: The intent of the modified code is clearly readable after this comment.

PR #13184, opened 2024-09-02
Fix some false positives in `Lint/UnreachableCode`.

isuckatcs:
If a developer decides to either redefine builtin methods like `exit`, `throw`, etc. or defines methods with the same name inside a `class`, `Lint/UnreachableCode` will treat these methods as if they were still the builtin ones, which leads to false positive reports.

E.g.:
```ruby
def exit;end

exit
puts "this is reachable"
^^^^^^^^^^^^^^^^^^^^^^^^
# RuboCop: Unreachable code detected. [Lint/ UnreachableCode]
```
```ruby
class Dummy
  def exit; end
end

d = Dummy.new
d.instance_eval do
  exit
  puts "this is reachable"
  ^^^^^^^^^^^^^^^^^^^^^^^^
  # RuboCop: Unreachable code detected. [Lint/ UnreachableCode]
end
```

This patch aims to address these cases by also checking if the called builtin has been redefined or not. The only exception is inside `instance_eval`, as we have no way to check the type of `self` and decide which methods are called, so in that case coverage is dropped instead. As a result, in the following case, we get a false negative.
```ruby
o = Object.new
o.instance_eval do
  exit
  puts "this should be reported as unreachable"
end
```
If introduing a false negative as a drawback of removing false positives is not desired, the `instance_eval` related part of the patch can be removed.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/unreachable_code.rb
+++ b/lib/rubocop/cop/lint/unreachable_code.rb
@@ -32,6 +32,22 @@ module RuboCop
       class UnreachableCode < Base
         MSG = 'Unreachable code detected.'
 
+        def initialize(config = nil, options = nil)
+          super
+          @redefined = []
+          @instance_eval_count = 0
+        end
+
+        def on_block(node)
+          @instance_eval_count += 1 if instance_eval_block?(node)
+        end
+
+        alias on_numblock on_block
+
+        def after_block(node)
+          @instance_eval_count -= 1 if instance_eval_block?(node)
+        end
+
         def on_begin(node)
           expressions = *node
 
@@ -46,19 +62,23 @@ module RuboCop
 
         private
 
+        def redefinable_flow_method?(method)
+          %i[raise fail throw exit exit! abort].include? method
+        end
+
         # @!method flow_command?(node)
         def_node_matcher :flow_command?, <<~PATTERN
           {
             return next break retry redo
             (send
              {nil? (const {nil? cbase} :Kernel)}
-             {:raise :fail :throw :exit :exit! :abort}
+             #redefinable_flow_method?
              ...)
           }
         PATTERN
 
         def flow_expression?(node)
-          return true if flow_command?(node)
+          return report_on_flow_command?(node) if flow_command?(node)
 
           case node.type
           when :begin, :kwbegin
@@ -68,6 +88,8 @@ module RuboCop
             check_if(node)
           when :case, :case_match
             check_case(node)
+          when :def
+            register_redefinition(node)
           else
             false
           end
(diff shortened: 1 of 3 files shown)


bbatsov:
> If introduing a false negative as a drawback of removing false positives is not desired, the instance_eval related part of the patch can be removed.

I'm on the fence about this part. @dvandersluis @koic What do you think?

dvandersluis:
It seems to me that the method definition being used inside the `instance_eval` wouldn't be in the file being processed in many cases anyways, so it's of low value. 

Would this offense still be registered?

```ruby
def exit
  # ...
end

Kernel.exit
foo
^^^ Unreachable code detected.
```

lib/rubocop/cop/lint/unreachable_code.rb
```diff
@@ -88,6 +106,33 @@ def check_case(node)
...
+          node.block_type? && node.method?(:instance_eval)
+        end
+
+        def report_on_flow_command?(node)
+          return true unless node.send_type?
+
+          # By the contract of this function, this case means that
+          # the method is called on `Kernel` in which case we
+          # always want to report a warning.
+          return true if node.children[0] != nil
```
dvandersluis, line 125: Since `node` is guaranteed to be a `send` node here, we can use node methods here.

```suggestion
          return true if node.receiver
```

lib/rubocop/cop/lint/unreachable_code.rb
```diff
@@ -88,6 +106,33 @@ def check_case(node)
...
+
+          # Inside an `instance_eval` we have no way to tell the
+          # type of `self` just by looking at the AST, so we can't
+          # tell if the give function that's called has been
+          # redefined or not, so to avoid false positives, we silence
+          # the warning.
+          return false if @instance_eval_count.positive?
+
+          !(@redefined.include? node.children[1])
```
dvandersluis, line 134: Same thing here:

```suggestion
          !@redefined.include?(node.method_name)
```

lib/rubocop/cop/lint/unreachable_code.rb
```diff
@@ -88,6 +106,33 @@ def check_case(node)
 
           branches.all? { |branch| branch.body && flow_expression?(branch.body) }
         end
+
+        def handle_def(node)
+          @redefined << node.method_name
+          false
+        end
```
dvandersluis, line 113: This should probably only add the method to `@redefined` if it's one of the flow control methods.

dvandersluis:
Thanks! Please fix up the rubocop failures and then squash your commits.

isuckatcs:
@dvandersluis Thanks for the review! I squashed the commits and fixed the failures.

dvandersluis:
Thank you!

PR #13185, opened 2024-09-03
Fix false negatives in `Style/MapIntoArray` autocorrection

vlad-pisanov:
Small improvement to `Style/MapIntoArray`'s autocorrection detection.

This cop can only autocorrect in a void context, and currently it only checks (kw)begin/(num)block contexts. This PR adapts some of the new logic from `Lint/Void` to expand void context detection to `ensure`, `def`, `defs`, and `for` contexts by delegating the responsibility to `rubocop-ast`.

e.g.
```ruby
def initialize
  dest = []
  list.each { dest << _1 } # 🆕 can now be autocorrected since it's in constructor
end
```
```ruby
def self.name=(value)
  dest = []
  list.each { dest << _1 } # 🆕 can now be autocorrected since it's in an assignment method
end
```
```ruby
def foo
  bar
ensure
  dest = []
  list.each { dest << _1 } # 🆕 can now be autocorrected since ensure is always a void context
end
```
```ruby
for i in list
  dest = []
  list.each { dest << _1 } # 🆕 can now be autocorrected since it's in a for loop
end
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_style_map_into_array_autocorrect_detection.md
+++ b/changelog/fix_style_map_into_array_autocorrect_detection.md
@@ -1,0 +1,1 @@
+* [#13185](https://github.com/rubocop/rubocop/pull/13185): Fix false negatives in `Style/MapIntoArray` autocorrection when using `ensure`, `def`, `defs` and `for`. ([@vlad-pisanov][])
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

PR #13188, opened 2024-09-03
Use `rubocop-ast` `SimpleForwardable`

Earlopain:
Introduced in https://github.com/rubocop/rubocop-ast/pull/312
Also see https://github.com/rubocop/rubocop/pull/13175

I performance tested this change and found no significant difference, which makes sense since it's basically the same thing.

This is pretty much a revert of the previous PR, changing `Forwardable` to `SimpleForwardable` (except for one class, which just didn't use it so I left that out)

cc @marcandre


bbatsov:
Looks good. Thanks!

marcandre:
Nice, thank you ❤️

PR #13190, opened 2024-09-03
Don't parse YAML twice

Earlopain:
Once for the duplication check, and once again to convert it to a ruby hash

We can instead use the yaml ast from the duplication check and convert that into a hash directly.

```
# Rubocop
# Base:
$ hyperfine -w 5 -r 20 "bundle exec rubocop Rakefile"
  Time (mean ± σ):      1.041 s ±  0.014 s    [User: 0.898 s, System: 0.140 s]
  Range (min … max):    1.013 s …  1.061 s    20 runs

# Patch:
$ hyperfine -w 5 -r 20 "bundle exec rubocop Rakefile"
  Time (mean ± σ):      1.013 s ±  0.008 s    [User: 0.865 s, System: 0.146 s]
  Range (min … max):    0.998 s …  1.027 s    20 runs
```

~2.7%  Faster

```
# Gitlab
# Base:
$ hyperfine -w 5 -r 20 "bundle exec rubocop Rakefile"
  Time (mean ± σ):      3.485 s ±  0.027 s    [User: 2.565 s, System: 0.909 s]
  Range (min … max):    3.445 s …  3.529 s    20 runs

# Applied:
$ hyperfine -w 5 -r 20 "bundle exec rubocop Rakefile"
  Time (mean ± σ):      3.401 s ±  0.029 s    [User: 2.483 s, System: 0.906 s]
  Range (min … max):    3.356 s …  3.466 s    20 runs
```

~2.5% faster. I expected more since they have so many config files but hey, still something.

-----

I checked https://docs.ruby-lang.org/en/3.3/table_of_contents.html#classes and they all seem to be documented and have been around since 2.1

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/config_loader.rb
+++ b/lib/rubocop/config_loader.rb
@@ -67,8 +67,8 @@ module RuboCop
       def load_yaml_configuration(absolute_path)
         file_contents = read_file(absolute_path)
         yaml_code = Dir.chdir(File.dirname(absolute_path)) { ERB.new(file_contents).result }
-        check_duplication(yaml_code, absolute_path)
-        hash = yaml_safe_load(yaml_code, absolute_path) || {}
+        yaml_tree = check_duplication(yaml_code, absolute_path)
+        hash = yaml_tree_to_hash(yaml_tree) || {}
 
         puts "configuration from #{absolute_path}" if debug?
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/config_loader.rb
```diff
@@ -245,10 +245,16 @@ def yaml_safe_load(yaml_code, filename)
         raise
       end
 
-      def yaml_safe_load!(yaml_code, filename)
-        YAML.safe_load(
-          yaml_code, permitted_classes: [Regexp, Symbol], aliases: true, filename: filename
-        )
+      def yaml_safe_load!(yaml_tree)
```
bbatsov, line 248: I've got a slight concern that the name is pretty similar to `yaml_safe_load`, but the signature now is different. Maybe it should just be a method named differently?

Earlopain: Yes, I suppose I should have renamed it. It's a private method, so no need to keep the old one around. New name is `yaml_tree_to_hash`, I guess? I did want to do some refactor to `YAMLDuplicationChecker`, and while it is marked private api, extensions use it to lint their own config file through the `project_spec` so I left that alone.

Issue #13193, opened 2024-09-04
Regression on Style/MultipleComparison: false positives


akimd:
In the following example, there are several "multiple comparisons", but each one of them has only two comparisons, not more.  I have:
```
Style/MultipleComparison:
  ComparisonsThreshold: 3
```

so they are all right.  However there must be some state kept somewhere and the cop believes we reached the threshold.

```ruby
$ cat multiple-comparison.rb
def foo(stage)
  puts if stage == 1 || stage == 2
end

def bar(safe)
  puts if safe == 'false' || safe == 'true'
end

def baz(x, y)
  puts if x == 1 || x == 2
  puts if y == 1 || y == 2
end
```

--------

## Expected behavior

No complaints.   That was the case with 1.65.1, it's a regression in 1.66 and 1.66.1.

## Actual behavior

```
$ bundle exec rubocop --only Style/MultipleComparison multiple-comparison.rb
Inspecting 1 file
C

Offenses:

multiple-comparison.rb:6:11: C: [Correctable] Style/MultipleComparison: Avoid comparing a variable with multiple items in a conditional, use Array#include? instead.
  puts if safe == 'false' || safe == 'true'
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
multiple-comparison.rb:11:11: C: [Correctable] Style/MultipleComparison: Avoid comparing a variable with multiple items in a conditional, use Array#include? instead.
  puts if y == 1 || y == 2
          ^^^^^^^^^^^^^^^^

1 file inspected, 2 offenses detected, 2 offenses autocorrectable
```

If I remove `foo`, then `bar` has no violation.

## RuboCop version

```
$ bundle exec rubocop -V
1.66.1 (using Parser 3.3.5.0, rubocop-ast 1.32.2, running on ruby 3.3.4) [arm64-darwin23]
  - rubocop-performance 1.21.1
```

Earlopain:
@fatkodima, do you want to take a look?

fatkodima:
Yes, I will take a look.

vlad-pisanov:
@fatkodima I took a crack at [fixing the regression](https://github.com/rubocop/rubocop/pull/13195), can you take a look when you get a chance? 🙏

PR #13195, opened 2024-09-04
Fix false positive in `Style/MultipleComparison`

vlad-pisanov:
Fix for https://github.com/rubocop/rubocop/issues/13193

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/multiple_comparison.rb
+++ b/lib/rubocop/cop/style/multiple_comparison.rb
@@ -61,8 +61,13 @@ module RuboCop
 
         def on_or(node)
           root_of_or_node = root_of_or_node(node)
-
           return unless node == root_of_or_node
+
+          if or_root_changed?(root_of_or_node)
+            reset_comparison
+            @current_or_root = root_of_or_node
+          end
+
           return unless nested_variable_comparison?(root_of_or_node)
           return if @allowed_method_comparison
           return if @compared_elements.size < comparisons_threshold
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/multiple_comparison.rb
```diff
@@ -61,8 +61,13 @@ def on_new_investigation
 
         def on_or(node)
           root_of_or_node = root_of_or_node(node)
-
           return unless node == root_of_or_node
+
+          if or_root_changed?(root_of_or_node)
+            reset_comparison
+            @current_or_root = root_of_or_node
+          end
```
vlad-pisanov, line 69: We need to reset the context every time we find a new OR root expression, not just when we start a new investigation. Otherwise multiple short OR expressions add up into a false offense 😬

fatkodima:
This became even more complicated. I was thinking if it is possible to get rid of instance variables (and so state) in this cop?

vlad-pisanov:
Unless rubocop has some kind of hook for "new root OR expression", I don't think it's possible to eliminate state tracking. 🤷‍♂️  We have to listen to every `on_or` and check for context changes. That's how the cop [was originally implemented](https://github.com/rubocop/rubocop/pull/13087/files#diff-1039f41d8fda246e269cb3b972990d287a7ebbaf24b29544dbcf85998730ed93L63) so this approach doesn't add anything new.

fatkodima:
I will try to implement an alternative fix using that approach. Will report here if was successful.

fatkodima:
https://github.com/rubocop/rubocop/pull/13196

PR #13200, opened 2024-09-05
Add new `IgnoreDuplicateElseBranch` option to `Lint/DuplicateBranch`

### LINKED ISSUE #13146: Ability to ignore branches repeated only by `else` in `Lint/DuplicateBranch`
I would like to propose a new configuration option (or do this by default) to ignore repeated branches when we have some branch which is repeated only in `else`. Sometimes it is better (more descriptive) to be explicit by listing all the conditions/options and fallback to `else`.

```ruby
# good
case option
when "bar"
  do_something
when "baz"
  do_something_else
else
  do_something
end

# bad, the same is repeated in both branches
if condition
  do_something
else
  do_something
end
```

fatkodima:
Closes #13146.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1785,6 +1785,7 @@ Lint/DuplicateBranch:
   VersionChanged: '1.7'
   IgnoreLiteralBranches: false
   IgnoreConstantBranches: false
+  IgnoreDuplicateElseBranch: false
 
 Lint/DuplicateCaseCondition:
   Description: 'Do not repeat values in case conditionals.'
--- a/lib/rubocop/cop/lint/duplicate_branch.rb
+++ b/lib/rubocop/cop/lint/duplicate_branch.rb
@@ -15,6 +15,9 @@ module RuboCop
       # With `IgnoreConstantBranches: true`, branches are not registered
       # as offenses if they return a constant value.
       #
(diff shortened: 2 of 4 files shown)


lib/rubocop/cop/lint/duplicate_branch.rb
```diff
@@ -83,21 +86,37 @@ module Lint
       #   else MEDIUM_SIZE
       #   end
       #
+      # @example IgnoreDuplicateElseBranches: true
```
bbatsov, line 89: I think a better name would be `IgnoreDuplicateElseBranch`, as the `else` branch is just one.

fatkodima: Agreed, changed.

config/default.yml
```diff
@@ -1782,9 +1782,10 @@ Lint/DuplicateBranch:
   Description: Checks that there are no repeated bodies within `if/unless`, `case-when` and `rescue` constructs.
   Enabled: pending
   VersionAdded: '1.3'
-  VersionChanged: '1.7'
+  VersionChanged: '<<next>>'
```
koic, line 1785: Although a configuration parameter has been added, the default behavior has not changed, so updating `VersionChanged` is not necessary.

```suggestion
  VersionChanged: '1.7'
```

PR #13206, opened 2024-09-07
Remove `yard` workarounds

Earlopain:
This is now resolved

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,15 +7,7 @@ gemspec
 gem 'asciidoctor'
 gem 'bump', require: false
 gem 'bundler', '>= 1.15.0', '< 3.0'
-# FIXME: This is a workaround to prevent the following warning in YARD:
-# https://github.com/lsegal/yard/pull/1546
-# Please remove this dependency when the issue is resolved.
-gem 'logger'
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -25,7 +17,7 @@ gem 'rubocop-rspec', '~> 3.0.0'
 gem 'simplecov', '~> 0.20'
 gem 'stackprof', platform: :mri
 gem 'test-queue'
-gem 'yard', '~> 0.9'
+gem 'yard', '~> 0.9', '>= 0.9.37'
```
koic, line 20: Can you tweak to the following?

```suggestion
gem 'yard', '>= 0.9.37', '< 1.0'
```

koic: The above suggestion might have been a bit too complex.

Rather, it would be better to keep it as is since it will likely be resolved during `bundle update`. This configuration shouldn't cause any issues and is simple enough.

```suggestion
gem 'yard', '~> 0.9'
```

This is because issues with updating YARD are rare.

Earlopain: I changed this to ensure local copies are updated but I guess it doesn't terribly matter

PR #13209, opened 2024-09-09
Call super in `relevant_file?` for InternalAffairs cops

KNejad:
This overwritten `relevant_file?` methods are preventing `Exclude` from working for these cops, since the behaviour is defined in the parent class `RuboCop::Cop::Base`.

You can verify this bug by applying the following patch and running `bundle exec rubocop`:

```patch
diff --git a/.rubocop_todo.yml b/.rubocop_todo.yml
index 8c0c585df..27293b8ff 100644
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -10,6 +10,10 @@
 InternalAffairs/NodeDestructuring:
   Enabled: false

+InternalAffairs/CopDescription:
+  Exclude:
+    - 'lib/rubocop/cop/internal_affairs/cop_description.rb'
+
 # Offense count: 55
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
diff --git a/lib/rubocop/cop/internal_affairs/cop_description.rb b/lib/rubocop/cop/internal_affairs/cop_description.rb
index d5097d4d3..bf1814d90 100644
--- a/lib/rubocop/cop/internal_affairs/cop_description.rb
+++ b/lib/rubocop/cop/internal_affairs/cop_description.rb
@@ -3,6 +3,7 @@
 module RuboCop
   module Cop
     module InternalAffairs
+      #
       # Enforces the cop description to start with a word such as verb.
       #
       # @example
```

You'll notice that on my branch it's green, whilst on master there is an offense which should have been ignored.

* [x] Added tests.
  * I don't believe this is needed since it only affects internal affairs cops.

### DIFF
--- a/lib/rubocop/cop/internal_affairs/cop_description.rb
+++ b/lib/rubocop/cop/internal_affairs/cop_description.rb
@@ -114,7 +114,7 @@ module RuboCop
         end
 
         def relevant_file?(file)
-          file.match?(%r{/cop/.*\.rb\z})
+          file.match?(%r{/cop/.*\.rb\z}) && super
         end
       end
     end
--- a/lib/rubocop/cop/internal_affairs/useless_message_assertion.rb
+++ b/lib/rubocop/cop/internal_affairs/useless_message_assertion.rb
@@ -46,7 +46,7 @@ module RuboCop
 
         # Only process spec files
         def relevant_file?(file)
-          file.end_with?('_spec.rb')
+          file.end_with?('_spec.rb') && super
         end
       end
     end
--- a/spec/rubocop/cop/internal_affairs/cop_description_spec.rb
+++ b/spec/rubocop/cop/internal_affairs/cop_description_spec.rb
@@ -152,4 +152,33 @@ RSpec.describe RuboCop::Cop::InternalAffairs::CopDescription, :config do
       RUBY
     end
   end
+
+  context 'when the file is excluded' do
+    before do
+      allow_any_instance_of(described_class).to receive(:relevant_file?).and_call_original # rubocop:disable RSpec/AnyInstance
+    end
+
+    let(:config) do
+      RuboCop::Config.new(
+        'InternalAffairs/CopDescription' => { 'Exclude' => ['**/example_cop.rb'] }
+      )
+    end
+
+    it 'does not register an offense' do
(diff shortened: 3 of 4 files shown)


lib/rubocop/cop/internal_affairs/cop_description.rb
```diff
@@ -114,7 +114,7 @@ def comment_index(node, comment_line)
         end
 
         def relevant_file?(file)
-          file.match?(%r{/cop/.*\.rb\z})
+          file.match?(%r{/cop/.*\.rb\z}) && super
```
koic, line 117: After reverting the implementation by removing `super`, the newly added test case still pass. There might be some oversight in the test case.

```console
$ git diff .
diff --git a/lib/rubocop/cop/internal_affairs/cop_description.rb b/lib/rubocop/cop/internal_affairs/cop_description.rb
index d5097d4d3..769be81ac 100644
--- a/lib/rubocop/cop/internal_affairs/cop_description.rb
+++ b/lib/rubocop/cop/internal_affairs/cop_description.rb
@@ -114,7 +114,7 @@ module RuboCop
         end

         def relevant_file?(file)
-          file.match?(%r{/cop/.*\.rb\z}) && super
+          file.match?(%r{/cop/.*\.rb\z})
         end
       end
     end

$ bundle exec rspec spec/rubocop/cop/internal_affairs/cop_description_spec.rb

Randomized with seed 41144
........

Finished in 0.20845 seconds (files took 1.8 seconds to load)
8 examples, 0 failures

Randomized with seed 41144
```

koic:
Thanks!

KNejad:
Thanks for the review @koic :slightly_smiling_face: 

@splattael, yeah that all sounds good!

Earlopain:
> I also wonder if we should move the configuration for the InternalAffairs department from .rubocop.yml to internal_affairs.yml config

I think that would be nice. There are a few cops that either only apply to tests or library code (or something else) but currently they run through the whole codebase anyways, even here.

splattael:
https://gitlab.com/gitlab-org/ruby/gems/gitlab-styles/-/merge_requests/238/diffs#779d8015085a503ca370502f8d28b7dd367027fa_98_97 to demonstrate how `Include` could work.

PR #13224, opened 2024-09-11
Fix false positives for `Style/OperatorMethodCall` with named forwarding

Earlopain:
These cases are syntax errors as well, in addition to anonymous forwarding

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positives_for_operator_method_call.md
+++ b/changelog/fix_false_positives_for_operator_method_call.md
@@ -1,0 +1,1 @@
+* [#13224](https://github.com/rubocop/rubocop/pull/13224): Fix false positives for `Style/OperatorMethodCall` with named forwarding. ([@earlopain][])
--- a/lib/rubocop/cop/style/operator_method_call.rb
+++ b/lib/rubocop/cop/style/operator_method_call.rb
@@ -22,6 +22,9 @@ module RuboCop
 
         MSG = 'Redundant dot detected.'
         RESTRICT_ON_SEND = %i[| ^ & <=> == === =~ > >= < <= << >> + - * / % ** ~ ! != !~].freeze
+        INVALID_SYNTAX_ARG_TYPES = %i[
+          splat kwsplat forwarded_args forwarded_restarg forwarded_kwrestarg block_pass
+        ].freeze
 
         # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity
         def on_send(node)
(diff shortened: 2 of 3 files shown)


changelog/fix_false_positives_for_operator_method_call.md
```diff
@@ -0,0 +1 @@
+* [#13218](https://github.com/rubocop/rubocop/pull/13218): Fix false positives for `Style/OperatorMethodCall` with named forwarding. ([@earlopain][])
```
koic, line 1:
```suggestion
* [#13224](https://github.com/rubocop/rubocop/pull/13224): Fix false positives for `Style/OperatorMethodCall` with named forwarding. ([@earlopain][])
```

Earlopain: :roll_eyes:  I _really_ want this checked in CI, somehow

lib/rubocop/cop/style/operator_method_call.rb
```diff
@@ -50,11 +53,10 @@ def method_call_with_parenthesized_arg?(argument)
...
         end
 
-        def anonymous_forwarding?(argument)
-          return true if argument.forwarded_args_type? || argument.forwarded_restarg_type?
-          return true if argument.hash_type? && argument.children.first&.forwarded_kwrestarg_type?
+        def forwarding?(argument)
+          return DISALLOWED_ARG_TYPES.include?(argument.children.first&.type) if argument.hash_type?
 
-          argument.block_pass_type? && argument.source == '&'
+          DISALLOWED_ARG_TYPES.include?(argument.type)
```
koic, line 59: It seems that the duplication with `DISALLOWED_ARG_TYPES.include?` below can be eliminated. However, the combination of the method name `forwarding?` and the constant name `DISALLOWED_ARG_TYPES` might still be a bit unclear.

```ruby
         def forwarding?(argument)
           type = argument.hash_type? ? argument.children.first&.type : argument.type

           DISALLOWED_ARG_TYPES.include?(type)
         end
```

Earlopain: Nice suggestion, I like it. Renamed things a bit, so hopefully its clearer now

lib/rubocop/cop/style/operator_method_call.rb
```diff
@@ -50,11 +53,10 @@ def method_call_with_parenthesized_arg?(argument)
           argument.children.first && argument.parent.parenthesized?
         end
 
-        def anonymous_forwarding?(argument)
-          return true if argument.forwarded_args_type? || argument.forwarded_restarg_type?
-          return true if argument.hash_type? && argument.children.first&.forwarded_kwrestarg_type?
+        def disallow_argument?(argument)
```
koic, line 56: Hm, `disallow_argument?` might be a bit ambiguous. What do you think about the names `invalid_syntax_argument?` method and `INVALID_SYNTAX_ARG_TYPES` constant?

```suggestion
        def invalid_syntax_argument?(argument)
```

Or `malformed_argument?` and `MALFORMED_ARG_TYPES`, etc...

Earlopain: `invalid_syntax_argument` sounds fine

PR #13228, opened 2024-09-11
Add new `Style/SafeNavigationChainLength` cop

### LINKED ISSUE #13171: Cop to limit the number of safe navigation calls in a chain
It is painful to see many `&.` calls in a chain.

If I want to have at most 2 safe navigation calls in a row, for example, then:
```ruby
# bad
foo&.bar&.baz&.zoo

# good
foo&.bar&.baz
```

fatkodima:
Closes #13171.


fatkodima:
@bbatsov Can you please also consider this for the next rubocop release? 🙏

bbatsov:
I will.

config/default.yml
```diff
@@ -5232,6 +5232,13 @@ Style/SafeNavigation:
   # Maximum length of method chains for register an offense.
   MaxChainLength: 2
 
+Style/SafeNavigationChainLength:
+  Description: 'Enforces safe navigation chains length to not exceed the configured maximum.'
+  StyleGuide: '#long-safe-navigation-chain'
```
bbatsov, line 5237: I think this link needs to be updated.

lib/rubocop/cop/style/safe_navigation_chain_length.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Enforces safe navigation chains length to not exceed the configured maximum.
```
bbatsov, line 6: You might want to add a bit of rationale here as well, as one of the most common feedback we receive from users is that they don't understand from the cop docs alone what's the reasoning behind some of our rules.

fatkodima:
Applied suggestions.

bbatsov:
I think in the docs we should also mention some potential interplay with `Style/SafeNavigation` - e.g. if both are enabled and their settings are "incompatible", one of the cops will complain about what the other proposes.

fatkodima:
Updated with the suggestion. I can expand with a little more sentences about the "incompatible" term, if needed.

bbatsov:
@fatkodima Yeah, that would be nice.

lib/rubocop/cop/style/safe_navigation_chain_length.rb
```diff
@@ -0,0 +1,52 @@
...
+      # The longer the chain is, the harder it becomes to track what on it could be
+      # returning a `nil`.
```
Earlopain, line 8:
```suggestion
      # returning `nil`.
```

spec/rubocop/cop/style/safe_navigation_chain_length_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Style::SafeNavigationChainLength, :config do
+  let(:cop_config) { { 'Max' => max } }
+  let(:max) { 2 }
```
koic, line 5: Can you add some test cases such as `1` or `3`, in addition to case `2`?

fatkodima:
Added more tests and updated implementation.

dvandersluis:
We documented the potential incompatibility between this cop and `Style/SafeNavigation`, but wouldn't it make more sense to just define the maximum chain length across rubocop once? I'm not sure I understand why it would make sense to have different values. If nothing else, shouldn't the default at least be `SafeNavigation`'s `MaxChainLength`?

/cc @bbatsov

PR #13231, opened 2024-09-12
Don't use `loop: false` for `AlignmentDirective` tests

Earlopain:
I'd like to get rid of the `loop` option. This test has a semi-good reason to use this option. But it's also easy not to rely on it.

Simply remove the directive comment in the correction

-----------------

* [ ] Added tests.

### DIFF
--- a/spec/rubocop/cop/alignment_corrector_spec.rb
+++ b/spec/rubocop/cop/alignment_corrector_spec.rb
@@ -13,8 +13,8 @@ RSpec.describe RuboCop::Cop::AlignmentCorrector, :config do
               ^^ Indent this node.
           RUBY
 
-          expect_correction(<<~RUBY, loop: false)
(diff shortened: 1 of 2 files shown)


bbatsov:
> I'd like to get rid of the loop option.

Any particular reason for this?

Earlopain:
> Any particular reason for this?

Not looping is special handling for tests and not how it usually works. A few older tests still - do `loop: false`, hiding infinite loops. I'd like to take a look at these some time.

This is of course not related to this PR but my understanding is that the `loop` option was more used as an escape hatch from when tests  used to not loop at all. So trying to reduce usage and see where that leaves it.

bbatsov:
Your understanding is correct, I was mostly pondering if the updated tests are a bit less readable, but I agree it's a nice goal to remove the need for `loop: false`.

PR #11815, opened 2023-04-22
Make `Style/CollectionCompact` aware of `delete_if`

koic:
This PR makes `Style/CollectionCompact` aware of `delete_if`.

Note, `Enumerator` does not have `delete_if`, the following use case are allowed:

```ruby
% ruby -ve '[1, 2, 3].to_enum.delete_if { |e| e.nil? }'
ruby 3.3.0dev (2023-03-17T00:50:41Z master f29c9d6d36) [x86_64-darwin19]
-e:1:in `<main>': undefined method `delete_if' for an instance of Enumerator (NoMethodError)

[1, 2, 3].to_enum.delete_if { |e| e.nil? }
                 ^^^^^^^^^^
```

It doesn't register an offense as well as `to_enum` when `lazy` is used.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #13239, opened 2024-09-14
Fix false positive for `Style/CollectionCompact` when using `delete_if`

masato-bkn:
This PR fixes `Style/CollectionCompact` to ignore `delete_if`. 

`Style/CollectionCompact` supports `delete_if` in https://github.com/rubocop/rubocop/pull/11815 and https://github.com/rubocop/rubocop/pull/12803.
but, `delete_if` always returns self, whereas `compact!` returns `nil` if no changes are made. Therefore, the two methods are not compatible.
By ignoring `delete_if`, we can prevent false positives and ensure more appropriate code style checks.

```ruby
irb(main):001> {:a => 1, :b => 2}.delete_if { |k, v| v.nil? }
=> {:a=>1, :b=>2}
irb(main):002> {:a => 1, :b => 2}.compact!
=> nil

irb(main):003> [1, 2].delete_if { |v| v.nil? }
=> [1, 2]
irb(main):004> [1, 2].compact!
=> nil
```

### docs
- [Array#delete_if](https://docs.ruby-lang.org/en/3.3/Array.html#method-i-delete_if)
- [Array#compact!](https://docs.ruby-lang.org/en/3.3/Array.html#method-i-compact)
- [Hash#delete_if](https://docs.ruby-lang.org/en/3.3/Hash.html#method-i-delete_if)
- [Hash#compact!](https://docs.ruby-lang.org/en/3.3/Hash.html#method-i-compact)

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/collection_compact_spec.rb
+++ b/spec/rubocop/cop/style/collection_compact_spec.rb
@@ -5,15 +5,12 @@ RSpec.describe RuboCop::Cop::Style::CollectionCompact, :config, :ruby24 do
     expect_offense(<<~RUBY)
       array.reject { |e| e.nil? }
             ^^^^^^^^^^^^^^^^^^^^^ Use `compact` instead of `reject { |e| e.nil? }`.
-      array.delete_if { |e| e.nil? }
-            ^^^^^^^^^^^^^^^^^^^^^^^^ Use `compact!` instead of `delete_if { |e| e.nil? }`.
       array.reject! { |e| e.nil? }
             ^^^^^^^^^^^^^^^^^^^^^^ Use `compact!` instead of `reject! { |e| e.nil? }`.
     RUBY
 
     expect_correction(<<~RUBY)
       array.compact
-      array.compact!
       array.compact!
     RUBY
   end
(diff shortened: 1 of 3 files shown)


Earlopain:
Can you add a test that no offense is registered for documentation purposes?

masato-bkn:
Thanks! I've added the test case.

spec/rubocop/cop/style/collection_compact_spec.rb
```diff
@@ -199,11 +177,17 @@ def foo(params)
     RUBY
   end
 
+  it 'does not register an offense when using `delete_if`' do
```
koic, line 180: It may be helpful to add a comment like the following:

```suggestion
  # NOTE: Unlike `reject` or `compact` which return a new collection object,
  # `delete_if` always returns its own object `self` making them incompatible.
  it 'does not register an offense when using `delete_if`' do
```

masato-bkn: Thank you for the review! I've made the changes.

koic:
Thanks!

PR #13241, opened 2024-09-17
Add basic support for conditional returns that always return to `Lint/UnreachableCode`

Zopolis4:
Following up on https://github.com/rubocop/rubocop/pull/13117#issuecomment-2325419197. More could be done here, in particular for more traditional if statements like so:
```ruby
# bad
if false
  foo
end
```
But that would require a more significant rework of this cop, so for now I just covered the easy stuff.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/unreachable_code.rb
+++ b/lib/rubocop/cop/lint/unreachable_code.rb
@@ -74,9 +74,19 @@ module RuboCop
         end
 
         def check_if(node)
+          return true if check_literal_if(node)
+
           if_branch = node.if_branch
           else_branch = node.else_branch
           if_branch && else_branch && flow_expression?(if_branch) && flow_expression?(else_branch)
+        end
+
+        def check_literal_if(node)
+          return false unless flow_expression?(node.if_branch)
+
+          return true if !node.unless? && node.condition.truthy_literal?
+
+          node.unless? && node.condition.falsey_literal?
         end
 
         def check_case(node)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/unreachable_code.rb
```diff
@@ -74,11 +74,19 @@ def flow_expression?(node)
         end
 
         def check_if(node)
+          return true if check_literal_if(node)
```
vlad-pisanov, line 77: It seems this doesn't check whether the outcome of the `if` is a `flow_expression?` (return, exit, etc.) so you would get a false positive for something like:

```ruby
foo if true
bar              # wrongly marked as unreachable
```

Zopolis4: I missed that, good catch! I've fixed that and added a test case to ensure it doesn't get missed again.

lib/rubocop/cop/lint/unreachable_code.rb
```diff
@@ -74,11 +74,21 @@ def flow_expression?(node)
...
           else_branch = node.else_branch
           if_branch && else_branch && flow_expression?(if_branch) && flow_expression?(else_branch)
         end
 
+        def check_literal_if(node)
+          return false unless flow_expression?(node.if_branch)
+
+          return true if !node.unless? && node.condition.truthy_literal?
+
+          node.unless? && node.condition.falsey_literal?
```
vlad-pisanov, line 89: I think you could easily extend this to also support `else` branches. Something like:

```ruby
def check_literal_if(node)
  branch = unconditional_branch(node)
  branch && flow_expression?(branch)
end

def unconditional_branch(node)
  return node.if_branch   if !node.unless? && node.condition.truthy_literal?
  return node.else_branch if node.unless?  && node.condition.falsey_literal?
  nil
end
```

This would catch cases like
```ruby
if false
  foo
else
  return
end
```

Zopolis4:
Closing in favor of #13117.

PR #13243, opened 2024-09-17
`require 'language-server_protocol'` on-demand

Earlopain:
It's a heavy gem to require for something that may not even be used:

Old:
```
$ hyperfine -w 5 -r 20 "bundle exec rubocop Rakefile"
  Time (mean ± σ):      3.061 s ±  0.087 s    [User: 2.494 s, System: 0.321 s]
  Range (min … max):    2.947 s …  3.275 s    20 runs
```

New:
```
$ hyperfine -w 5 -r 20 "bundle exec rubocop Rakefile"
  Time (mean ± σ):      2.801 s ±  0.064 s    [User: 2.317 s, System: 0.235 s]
  Range (min … max):    2.710 s …  2.958 s    20 runs
```

~9% faster startup time for me

Of course, most time is still spent requiring all the various cops of rubocop itself (there are quite a few)

-----------------

* [ ] Added tests.


lib/rubocop.rb
```diff
@@ -750,6 +750,11 @@
 require_relative 'rubocop/config_validator'
 require_relative 'rubocop/feature_loader'
 require_relative 'rubocop/lockfile'
+require_relative 'rubocop/lsp'
+require_relative 'rubocop/lsp/logger'
+require_relative 'rubocop/lsp/routes'
+require_relative 'rubocop/lsp/runtime'
+require_relative 'rubocop/lsp/severity'
```
koic, line 757: These `require_relative` codes should only be necessary when running `rubocop --lsp` command. Is it really true that requiring these LSP-related codes globally in `lib/rubocop.rb` speeds up the regular `rubocop` command as well?

Earlopain: Indeed, the only one necessary is the one for `RuboCop::LSP.enabled?` and similar that some cops are using. These don't really matter since they are such small files, the real gain is from not requiring `language-server_protocol` every time.

These requires where only needed for some other approach I tried and I had them mixed now. Reran the benchmark with the revised change and its basically the same still.

Earlopain: For reference, here is the flamegraph:

grafik

Most of the small ones to the left are cop requires, and the one I'm hovering over is the gem (notice `module:LanguageServer` a bit on top)

lib/rubocop/cli/command/lsp.rb
```diff
@@ -11,6 +9,8 @@ class LSP < Base
         self.command_name = :lsp
 
         def run
+          # Load on demand, `languge-server-protocol` is heavy to require
```
koic, line 12:
```suggestion
          # Load on demand, `languge-server-protocol` is heavy to require.
```

PR #13032, opened 2024-07-08
Display warning messages for deprecated APIs

koic:
This PR displays warning messages for deprecated APIs.

The deprecated APIs are only soft-deprecated as explicitly mentioned in the documentation. Over three years have passed since RuboCop 1.0 was released: https://rubygems.org/gems/rubocop/versions/1.0.0

So it seems appropriate to start displaying deprecation warnings.

This aims to prepare for RuboCop 2.0 by providing users with the opportunity to become aware of the deprecated APIs and encouraging them to take action to avoid deprecated APIs.

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [ ] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [ ] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #8088, opened 2020-06-02
[Fixes #8077] Restore compatibility for Team.new

marcandre:

PR #13244, opened 2024-09-17
Emit a deprecation for `Team.new([Cop], config)`

Earlopain:
This should not be carried over to RuboCop v2

Followup to https://github.com/rubocop/rubocop/pull/13032. I don't think this single one needs a changelog entry?

-----------------

* [ ] Added tests.


lib/rubocop/cop/team.rb
```diff
@@ -9,11 +9,16 @@ module Cop
     # For performance reasons, Team will first dispatch cops & forces in two groups,
     # first the ones needed for autocorrection (if any), then the rest
     # (unless autocorrections happened).
-    class Team
+    class Team # rubocop:disable Metrics/ClassLength
```
koic, line 12: Since this `rubocop:disable` directive is expected to become unnecessary once the warning is resolved in the future, it is preferable to specify it as a block rather than at the end of the line. This prevents cluttering `git blame`. 

```suggestion
    # rubocop:disable Metrics/ClassLength
    class Team
```

Please place the `rubocop:enable` directive after the `end` corresponding to the `class`.

koic:
> I don't think this single one needs a changelog entry?

Yep, a changelog entry is probably unnecessary.

lib/rubocop/cop/team.rb
```diff
@@ -9,11 +9,17 @@ module Cop
...
     # (unless autocorrections happened).
+    # rubocop:disable Metrics/ClassLength
     class Team
       # @return [Team]
       def self.new(cop_or_classes, config, options = {})
         # Support v0 api:
-        return mobilize(cop_or_classes, config, options) if cop_or_classes.first.is_a?(Class)
+        if cop_or_classes.first.is_a?(Class)
+          warn Rainbow(<<~WARNING).yellow, uplevel: 1
+            `new` with cop classes is deprecated. Use `mobilize` instead.
```
koic, line 19: Can you tweak this message to the following?

```suggestion
            `Team.new` with cop classes is deprecated. Use `Team.mobilize` instead.
```

lib/rubocop/cop/team.rb
```diff
@@ -9,11 +9,17 @@ module Cop
     # For performance reasons, Team will first dispatch cops & forces in two groups,
     # first the ones needed for autocorrection (if any), then the rest
     # (unless autocorrections happened).
+    # rubocop:disable Metrics/ClassLength
     class Team
       # @return [Team]
```
koic, line 14: By the way, this API was not marked as soft-deprecated in the documentation. Is there any basis for deprecating this API?

Earlopain: It was added in https://github.com/rubocop/rubocop/pull/8088 as a workaround for breakage, and the comment seemed to suggest to me that it would be removed at a later date again.

I guess, if you prefer I can make a note for rubocop v2 or remove the comment altogether if it should stay around?

koic: Ah, this is fine. Thank you for the explanation.

PR #13245, opened 2024-09-17
Support `filter/filter!` in `Style/CollectionCompact`

masato-bkn:
This PR changes `Style/CollectionCompact` to support `filter/filter!`.

`filter/filter!` was introduced in Ruby 2.6 as an alias for `select/select!`, so I think it can be handled as the same way in this cop.

- https://docs.ruby-lang.org/en/3.3/Array.html#method-i-filter
- https://docs.ruby-lang.org/en/3.3/Hash.html#method-i-filter

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/collection_compact.rb
+++ b/lib/rubocop/cop/style/collection_compact.rb
@@ -21,6 +21,7 @@ module RuboCop
       #   array.reject(&:nil?)
       #   array.reject { |e| e.nil? }
       #   array.select { |e| !e.nil? }
+      #   array.filter { |e| !e.nil? }
       #   array.grep_v(nil)
       #   array.grep_v(NilClass)
       #
@@ -31,6 +32,7 @@ module RuboCop
       #   hash.reject!(&:nil?)
       #   hash.reject! { |k, v| v.nil? }
       #   hash.select! { |k, v| !v.nil? }
+      #   hash.filter! { |k, v| !v.nil? }
       #
       #   # good
       #   hash.compact!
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/collection_compact.rb
```diff
@@ -85,6 +87,7 @@ class CollectionCompact < Base
         PATTERN
 
         def on_send(node)
+          return if target_ruby_version < 2.6 && %i[filter filter!].include?(node.method_name)
```
koic, line 90: Can you define `%i[filter filter!]` as a constant? (e.g., `FILTER_METHODS`)

```suggestion
          return if target_ruby_version < 2.6 && FILTER_METHODS.include?(node.method_name)
```

spec/rubocop/cop/style/collection_compact_spec.rb
```diff
@@ -249,6 +249,52 @@ def foo(params)
...
+
+    it 'does not register an offense when using `filter/filter!` to reject nils without a receiver' do
+      expect_no_offenses(<<~RUBY)
+        filter { |e| !e.nil? }
+        filter! { |k, v| !v.nil? }
+      RUBY
+    end
+  end
+
+  context 'Ruby < 2.6', :ruby25, unsupported_on: :prism do
```
koic, line 289: Can you tweak it?

```suggestion
  context 'Ruby <= 2.5', :ruby25, unsupported_on: :prism do
```

masato-bkn:
Thank you for the review! I've fixed it.

PR #13247, opened 2024-09-17
Improve yaml duplication checker performance

Earlopain:
The first commit is an attempt at an optimization, where I made it not traverse the yaml ast. This is not necessary since psych offers an api to get tokens as they are emitted. There must be some improvement with this change but it's not measurable. I left it in anyways since I think the code is neater with that approach. I'm not attached, I can drop that commit if you want.

---------

The second commit does improve performance: The keys array can grow quite large and checking for the existance of a key becomes expensive. The worst case scenario is also the most common one, where the key is not found. Instead, use a hash:

Current:
```
$ hyperfine -w 10 -r 25 "bundle exec rubocop Rakefile"
  Time (mean ± σ):     962.2 ms ±  11.4 ms    [User: 845.0 ms, System: 116.6 ms]
  Range (min … max):   944.9 ms … 985.0 ms    25 runs
```

Without any checks at all:
```
$ hyperfine -w 10 -r 25 "bundle exec rubocop Rakefile"
  Time (mean ± σ):     922.5 ms ±  13.8 ms    [User: 807.8 ms, System: 113.4 ms]
  Range (min … max):   899.8 ms … 951.5 ms    25 runs
```

PR:
```
$ hyperfine -w 10 -r 25 "bundle exec rubocop Rakefile"
  Time (mean ± σ):     929.8 ms ±  13.1 ms    [User: 810.1 ms, System: 119.3 ms]
  Range (min … max):   912.0 ms … 959.9 ms    25 runs
```

So, about 3.5% faster for RuboCop itself. Or in other words, ~40ms checking duplication previously, now only ~7ms.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/yaml_duplication_checker.rb
+++ b/lib/rubocop/yaml_duplication_checker.rb
@@ -19,10 +19,13 @@ module RuboCop
 
       def end_mapping
         mapping_node = super
-        mapping_node.children.each_slice(2).with_object([]) do |(key, _value), keys|
-          exist = keys.find { |key2| key2.value == key.value }
-          @block.call(exist, key) if exist
-          keys << key
+        # OPTIMIZE: Use a hash for faster lookup since there can
+        # be quite a few keys at the top-level.
+        keys = {}
+        mapping_node.children.each_slice(2) do |key, _value|
+          duplicate = keys[key.value]
+          @block.call(duplicate, key) if duplicate
+          keys[key.value] = key
         end
         mapping_node
       end
--- a/lib/rubocop/yaml_duplication_checker.rb
+++ b/lib/rubocop/yaml_duplication_checker.rb
@@ -5,37 +5,27 @@ module RuboCop
   # @api private
   module YAMLDuplicationChecker
     def self.check(yaml_string, filename, &on_duplicated)
-      # Ruby 2.6+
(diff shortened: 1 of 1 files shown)


lib/rubocop/yaml_duplication_checker.rb
```diff
@@ -5,37 +5,30 @@ module RuboCop
...
-               YAML.parse(yaml_string, filename: filename)
-             else
-               YAML.parse(yaml_string, filename)
-             end
-      return unless tree
-
-      traverse(tree, &on_duplicated)
-      tree
+      handler = DuplicationCheckHandler.new(&on_duplicated)
+      parser = Psych::Parser.new(handler)
```
Earlopain, line 9: This api never moved to keyword arguments so no need for a version check

lib/rubocop/yaml_duplication_checker.rb
```diff
@@ -5,37 +5,30 @@ module RuboCop
...
+      parser.parse(yaml_string, filename)
+      parser.handler.root.children[0]
```
Earlopain, line 11: `Psych::Handlers::DocumentStream` would make this a bit nicer  (it's what `Psych.parse` uses under the hood) but it's marked nodoc so I didn't check it out further.

bbatsov:
Looks good to me. Thanks!

PR #13255, opened 2024-09-22
Fix false negatives for `Style/MapIntoArray` when using non-splatted arguments

vlad-pisanov:
Bugfix https://github.com/rubocop/rubocop/pull/13119 fixed the splat bug in `Style/MapIntoArray`, but it also limited the types of offenses the cop recognized. For example,

```ruby
dest = []
arr.each { |e| dest << [1, e] }
arr.each { |e| dest << true }
arr.each { |e| dest << (cond? ? 1 : e) }
```

and several others are no longer detected as of 1.66.1

This PR restores the original cop behavior while preserving the splat fix.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/map_into_array.rb
+++ b/lib/rubocop/cop/style/map_into_array.rb
@@ -53,12 +53,17 @@ module RuboCop
 
         MSG = 'Use `%<new_method_name>s` instead of `each` to map elements into an array.'
 
+        # @!method suitable_argument_node?(node)
+        def_node_matcher :suitable_argument_node?, <<-PATTERN
+          !{splat forwarded-restarg forwarded-args (hash (forwarded-kwrestarg)) (block-pass nil?)}
+        PATTERN
+
         # @!method each_block_with_push?(node)
         def_node_matcher :each_block_with_push?, <<-PATTERN
           [
             ^({begin kwbegin} ...)
             ({block numblock} (send !{nil? self} :each) _
-              (send (lvar _) {:<< :push :append} {send lvar begin}))
+              (send (lvar _) {:<< :push :append} #suitable_argument_node?))
           ]
         PATTERN
 
--- a/spec/rubocop/cop/style/map_into_array_spec.rb
+++ b/spec/rubocop/cop/style/map_into_array_spec.rb
@@ -65,6 +65,102 @@ RSpec.describe RuboCop::Cop::Style::MapIntoArray, :config do
     RUBY
   end
 
+  it 'registers an offense and corrects when using an if statement' do
+    expect_offense(<<~RUBY)
+      dest = []
+      src.each do |e|
+      ^^^^^^^^^^^^^^^ Use `map` instead of `each` to map elements into an array.
+        dest << if cond?
+          e
+        else
+          e * 2
+        end
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      dest = src.map do |e|
+        if cond?
+          e
+        else
+          e * 2
+        end
+      end
+    RUBY
+  end
+
+  it 'registers an offense and corrects when using a case statement' do
+    expect_offense(<<~RUBY)
+      dest = []
+      src.each do |e|
+      ^^^^^^^^^^^^^^^ Use `map` instead of `each` to map elements into an array.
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/style/map_into_array.rb
```diff
@@ -58,7 +58,7 @@ class MapIntoArray < Base
           [
             ^({begin kwbegin} ...)
             ({block numblock} (send !{nil? self} :each) _
-              (send (lvar _) {:<< :push :append} {send lvar begin}))
+              (send (lvar _) {:<< :push :append} !splat))
```
Earlopain, line 61: You need to consider `kwsplat` as well here.

I guess anonymous forwarding too? I doubt anyone would write such code though.

vlad-pisanov: `kwsplat` gets wrapped in a `hash` (since rubocop-ast runs in legacy mode), so we're covered there

```
ruby-parse -e "arr.push(**opts)" --legacy
(send
  (send nil :arr) :push
  (hash
    (kwsplat
      (send nil :opts))))

```

anonymous forwarding is a good callout, will add

lib/rubocop/cop/style/map_into_array.rb
```diff
@@ -53,12 +53,17 @@ class MapIntoArray < Base
 
         MSG = 'Use `%<new_method_name>s` instead of `each` to map elements into an array.'
 
+        # @!method suitable_element?(node)
+        def_node_matcher :suitable_element?, <<-PATTERN
+          !{splat forwarded-restarg forwarded-args (block-pass nil?)}
```
vlad-pisanov, line 58: `map { |x| ... }`, `map { |x| * }`, `map { |x| & }` aren't valid Ruby syntax, so we don't want to correct this type of forwarding

spec/rubocop/cop/style/map_into_array_spec.rb
```diff
@@ -215,6 +311,36 @@
...
+
+  it 'does not register an offense when pushing anonymously-forwarded restarg', :ruby32 do
+    expect_no_offenses(<<~RUBY)
+      def foo(*)
+        dest = []
+        src.each { |e| dest.push(*) }
+        do_something
+      end
+    RUBY
+  end
```
koic, line 332: It should also be ignored for anonymously forwarded kwrestarg (`**`)'.

spec/rubocop/cop/style/map_into_array_spec.rb
```diff
@@ -215,6 +311,36 @@
     RUBY
   end
 
+  it 'does not register an offense when pushing anonymously-forwarded block', :ruby31 do
+    expect_no_offenses(<<~RUBY)
+      def foo(&)
+        dest = []
+        src.each { |e| dest.push(&) }
+        do_something
```
koic, line 319: Does this `do_something` have any meaning? If not, it can probably be removed.

lib/rubocop/cop/style/map_into_array.rb
```diff
@@ -53,12 +53,17 @@ class MapIntoArray < Base
 
         MSG = 'Use `%<new_method_name>s` instead of `each` to map elements into an array.'
 
+        # @!method suitable_element?(node)
+        def_node_matcher :suitable_element?, <<-PATTERN
```
koic, line 57: How about a name like this?

```suggestion
        def_node_matcher :suitable_argument_node?, <<-PATTERN
```

bbatsov:
@vlad-pisanov Ping :-)

koic:
@vlad-pisanov Can you take a look at this?

spec/rubocop/cop/style/map_into_array_spec.rb
vlad-pisanov: Added ✔️

spec/rubocop/cop/style/map_into_array_spec.rb
vlad-pisanov: Good point, that was only relevant for autocorrection (to create an empty context.)

koic:
Thanks!

PR #13264, opened 2024-09-26
[Fix #13259] Add new duplicate set enumerables in set cop

### LINKED ISSUE #13259: Cop idea: detect duplicated set elements
I was a little bit surprised there's no built-in cop for things like

```ruby
UNRECOVERABLE_FAILURES = Set[:card_declined, :card_frozen, ..., :card_declined].freeze
                                                                 ^^^^^^^^^^^^^^ Duplicated set element.
```

Is there a reason it wasn't implemented yet?

For sets with symbols and numerics it should be safe to assume that
```
(send
  (const nil :Set) :[]
  (sym :foo)
  (sym :bar)
  (sym :foo))
```

is an offense. 

There might be room for improvements, like detecting `Set.new(*%i[foo bar foo])` or even `foo = 1; bar = 2; Set[foo, bar, foo]` (when `foo` and `bar` are of type `lvar`)

## Additional context

I made a PoC implementation for the baseline case and it seems to work fine.

ally1002:
Fix #13259 

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/duplicate_set_enumerables.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Checks for duplicate enumerables in a Set.
+      # This cop detects arguments and use them to check duplacity.
```
dvandersluis, line 7: It'd be good to include in the documentation a list of what type of objects this cop handles. Also duplacity isn't a word :)

lib/rubocop/cop/lint/duplicate_set_enumerables.rb
```diff
@@ -0,0 +1,49 @@
...
+
+        MSG = 'Duplicate enumerables found in Set.'
+
+        def on_send(node)
+          return unless duplicates?(node.arguments)
+
+          duplicates(node.arguments).each do |duplicate|
+            add_offense(duplicate)
+          end
+        end
```
dvandersluis, line 45: This `on_send` method will be called for every `send`. You can use `RESTRICT_ON_SEND` to restrict to specific methods (`:new` and `:[]` is a good start; note that `Set[1,2]` also creates a Set), but you should also check the receiver to make sure it's `Set`). See https://docs.rubocop.org/rubocop/development.html#implementation

lib/rubocop/cop/lint/duplicate_set_enumerables.rb
ally1002: Sure! I'll change that later, thanks

lib/rubocop/cop/lint/duplicate_set_enumerables.rb
```diff
@@ -0,0 +1,49 @@
...
+      #
+      #   # bad
+      #   Set.new[true, true]
+      #
+      #   # good
+      #   Set.new[true, false]
+      class DuplicateSetEnumerables < Base
+        include Duplication
+
+        MSG = 'Duplicate enumerables found in Set.'
```
dvandersluis, line 37:
```suggestion
        MSG = 'Duplicate items found in Set.'
```

spec/rubocop/cop/lint/duplicate_set_enumerables_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Lint::DuplicateSetEnumerables, :config do
+  context 'when there is duplicated symbols in set' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        Set.new[:otherkey, :key, :key]
+                                 ^^^^ Duplicate enumerables found in Set.
+                           ^^^^ Duplicate enumerables found in Set.
```
vlad-pisanov, line 9: consider using `consecutive_duplicates` instead of `duplicates` like [here](https://github.com/rubocop/rubocop/blob/b03a1edd08037f61641ca76d81e7804ca3f404d5/lib/rubocop/cop/lint/duplicate_hash_key.rb#L28)

this would produce a much [cleaner error message](https://github.com/rubocop/rubocop/blob/b03a1edd08037f61641ca76d81e7804ca3f404d5/spec/rubocop/cop/lint/duplicate_hash_key_spec.rb#L7-L8) where only the first duplicate is highlighted.

In general, I think this cop is extremely similar to `Lint/DuplicateHashKey`, so we should probably follow its conventions to keep things consistent 🙂

koic:
It seems that the work in this PR was duplicated. I will close this to maintain #13265, which supports some cases and autocorrection. Thank you for the contribution.

PR #13265, opened 2024-09-27
[Fix #13259] Add new `Lint/DuplicateSetElement` cop

### LINKED ISSUE #13259: Cop idea: detect duplicated set elements

koic:
Fixes #13259.

This PR adds new `Lint/DuplicateSetElement` cop that checks for duplicate elements in Set.

```ruby
# bad
Set[:foo, :bar, :foo]

# good
Set[:foo, :bar]

# bad
Set.new([:foo, :bar, :foo])

# good
Set.new([:foo, :bar])
```

-----------------

* [x] Added tests.


lib/rubocop/cop/lint/duplicate_set_element.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        def on_send(node)
+          return unless (set_elements = set_init_elements(node))
+
+          seen_elements = Set[]
+
+          set_elements.each_with_index do |set_element, index|
+            # NOTE: Ignore due to the possibility of the corner case where Set elements
+            # may have changing return values through method calls.
+            next if set_element.call_type?
```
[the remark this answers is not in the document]
koic: Specifically, what kind of code are you referring to?

spec/rubocop/cop/lint/duplicate_set_element_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+      ::Set.new([:foo, :bar, :foo])
+                             ^^^^ Remove the duplicate element in Set.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      ::Set.new([:foo, :bar])
+    RUBY
+  end
+
+  it 'does not register an offense when not using duplicate element in Set' do
```
Earlopain, line 122: This tests method calls, so will not register an offense because of that if I understand correctly.

lib/rubocop/cop/lint/duplicate_set_element.rb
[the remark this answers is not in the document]
Earlopain: If you ignore method calls as values, then something like `Set.new([1, 2, 3]) { |x| foo(x) }` should also be considered.

spec/rubocop/cop/lint/duplicate_set_element_spec.rb
koic: The test is written that way, but is there any problem with it?

Earlopain: If it is intended that way then the test description should mention method calls, since it would not register an offense even if there was duplication with method calls, I think.

lib/rubocop/cop/lint/duplicate_set_element.rb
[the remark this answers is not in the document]
koic: Hmm. Duplicated elements listed in `Set[...]` are most likely mistakes. On the other hand, changing Set elements within a block seems to be quite a corner case.

Let's see how this works with feedback from the real world. For now, I'll leave it as it is.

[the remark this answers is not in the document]
Earlopain: Yeah, fair. Block form itself is probably already pretty rare. https://github.com/search?q=lang%3Aruby+%2FSet%5C.new%5C%28%5C%5B.*%5C%5D%5C%29%5Cs%5C%7B%2F&type=code, and impure method calls inside even moreso

lib/rubocop/cop/lint/duplicate_set_element.rb
```diff
@@ -0,0 +1,66 @@
...
+          }
+        PATTERN
+
+        def on_send(node)
+          return unless (set_elements = set_init_elements(node))
+
+          seen_elements = Set[]
+
+          set_elements.each_with_index do |set_element, index|
+            # NOTE: Ignore due to the possibility of the corner case where Set elements
```
vlad-pisanov, line 42: checking for `call_type?` is probably not enough, since this would still wrongly flag and correct cases where the method call is buried in some expression, like:
```ruby
Set[(1 + rand), (1 + rand)]
Set[[rand], [rand]]
Set[rand > 0.5 ? 1 : 2, rand > 0.5 ? 1 : 2]
```

it might be safest to just restrict this cop to literals, constants, and variables than try to detect every edge case

lib/rubocop/cop/lint/duplicate_set_element.rb
```diff
@@ -0,0 +1,66 @@
...
+
+module RuboCop
+  module Cop
+    module Lint
+      # Checks for duplicate elements in Set.
+      #
+      # @example
+      #
+      #   # bad
+      #   Set[:foo, :bar, :foo]
```
vlad-pisanov, line 11: should this cop also detect `[:foo, :bar, :foo].to_set`?

I know a few codebases where this is preferred to `Set[...]`

lib/rubocop/cop/lint/duplicate_set_element.rb
koic: Ah, this was a point I was also considering. I've updated this PR. Thank you!

lib/rubocop/cop/lint/duplicate_set_element.rb
koic: This case seems to be reasonable. I've updated this PR.

bbatsov:
Nice addition! Thanks!

PR #11349, opened 2022-12-29
[#8227] Change Style/FormatStringToken style to `template` internally

pirj:
Related: #8227

Extracted from #10723

-----------------

Before submitting the PR make sure the following are checked:

* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [x] Wrote [good commit messages][1].
* [-] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [-] Added tests.
* [x] Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [-] Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

[1]: https://chris.beams.io/posts/git-commit/

PR #13266, opened 2024-09-27
[Fix #13121] Add new `Style/EmptyStringInsideInterpolation` cop.

### LINKED ISSUE #13121: Cop idea: detect redundant setting to empty string inside string interpolation

I sometimes encounter code with an if/else condition that sets a variable to an empty string as one of the outcomes. 

This isn't needed if the variable is only being used in string interpolation, and although thats not always something rubocop can detect, I assume it can detect if it the condition is inside of a string.

```ruby
# bad, but rubocop isn't able to know that.
example = [true, false].sample ? '--example-arg' : ''
puts "program #{example} --bonus"

# manually, this can be corrected if the user can ensure this is the only usage of example.
puts "program #{'--example-arg' if [true, false].sample} --bonus"
```

```ruby
# bad
puts "program #{example = [true, false].sample ? '--example-arg' : ''} --bonus"

# good
puts "program #{'--example-arg' if [true, false].sample} --bonus"
```

For the following code patterns to be flagged as offenses and autocorrectable:
```ruby
# bad
"bar #{condition ? 'foo' : ''}"
"bar #{if condition; 'foo' else '' end}"

# good
"bar #{'foo' if condition}"
```

(and also the same for cases where the ternary operator is reversed and it should correct to `unless` instead of `if`, and also for empty strings in the form of `""`-- not sure about whether flagging assigning things to `nil` is safe, i think it is?)

Zopolis4:
Fixes #13121.

The new `Style/EmptyStringInsideInterpolation` checks for variables being assigned to empty strings inside of string interpolation, and removes this path from conditionals.

Additionally, I extended the `Lint/EmptyInterpolation` cop to detect basic empty string interpolations, such as `"#{''}"` and `"#{nil}"`.

this registered an offense in the `Lint/Void` spec, so I let it autocorrect that.

Currently `Lint/EmptyInterpolation` does not support cases such as `"#{''''}"` or ` "#{'' + "" '' ""'' + nil.to_s}"`, although these cases of multiple empty strings as the only content of a string interpolation would presumably be rare.

Admittedly, `Style/EmptyStringInsideInterpolation` does not support them either, so I need to at some point find a good way of going about this. 

As for now, though, the majority of common occurrences are covered, so I'll work on the more obscure ones in a follow-up PR.

-----------------

* [x] Added tests.


Zopolis4:
Just realized I forgot to change the description in `config/default.yml` when copying the block over from the last cop-- would a basic sanity check along the lines of "no two cops have the same description in the configuration" make sense, or is this a mistake other people wouldn't make?

(for what its worth, i did do a quick check locally and no two cops have the same description in the configuration, at least those with one-line descriptions)

Earlopain:
This should probably be a style cop. I understand this comes a bit late, but can you open a PR at https://github.com/rubocop/ruby-style-guide/? Both versions seem perfectly acceptable to me (and I do have a strong preference for using empty strings).

> Just realized I forgot to change the description in config/default.yml when copying the block over

You're supposed to use `rake new_cop[name]` (: It does most things for you.

koic:
First of all, this is definitely not Lint. Since this is a matter of preference, it falls under Style.

Also, I'm not sure if the rule enforced by this cop is appropriate.
`#{cond? ? 'str' : ''}` can be considered reasonable in that the result of `cond?` is always a string. And it would be clear.
`#{'str' if cond?}` uses the feature where `nil` becomes an empty string through string interpolation. Some people might prefer this, but I'm not sure if it would be universally accepted.

Zopolis4:
Ok, I'll move it to Style and open a pull request over at the style guide to see whether this style is preferred or not.

pirj:
I’d argue that this style is significantly more frequently used and should be enforced.
We strive for consistency. Is there a reason to bother projects/teams that used to the ternary style?

Zopolis4:
> We strive for consistency. Is there a reason to bother projects/teams that used to the ternary style?


I don't really understand what you mean here-- the point of every `Style` cop in rubocop is to enforce a particular way of programming. 'bothering' projects/teams that are using a different style is what an opinionated linter/formatter does.

pirj:
Do you think RuboCop is opinionated?

Zopolis4:
Yeah? Cops like `Style/Next`, `Style/RedundantReturn`, `Style/StringLiterals` are all opinionated, as they're promoting a specific way of writing code as opposed to an valid alternative. 

I get that most of them are configurable to support an alternative style, or can just otherwise be disabled, but the default choice of a style is still opinionated. If the issue is that this cop doesn't support an alternative style, I can add that as a configuration option.

pirj:
We strive to make default styles match the usage, and for this, before picking the default style, we consult existing open codebases, and provide options for styles that are used less frequently. I can’t think of a more democratic way, and I wouldn’t call this “opinionated”.
At least this is the case for rspec-rubocop, -capybara and -factory_bot extensions that I feel responsible for.

There are opposite examples, of course, https://github.com/rubocop/rubocop/pull/11349#issuecomment-1367752612, but I believe this is a rare exception (and an edge case, so I didn’t insist).

Speaking of this cop - I’d suggest to add an alternative style. I don’t insist on doing this in this initial pr, it will be fine in a follow-up, or never, but it’s worth noting our intent of doing so.

And yes, no doubt that many teams still perceive the default RuboCop settings as “the standard”, but it’s not, and we’ve tried to conduct this message in different ways.

Zopolis4:
I recognize that the default styles match real-world usage and are chosen with consideration-- at the end of the day, if I run rubocop on a codebase for the first time, the default configuration may enforce a style of programming that is different to the one I am currently using. I view that as opinionated, even though there is frequently a way to ensure rubocop follows an individual's preferred style.

At any rate, I believe this cop follows a similar thought process to other `Style` cops, whether you view that as opinionated or not. 

That being said, I'll work on adding an alternative style so it better matches other cops.

bbatsov:
I get the point of the suggested cop, but the name seems a bit misleading to me as it seems it only checks empty literals returns in conditionals, which is a much narrower scope than the name of the cop may suggest (especially given how close the name is to the existing `Lint` check). If I didn't know how the two cops differ, I'd struggle to figure this out from their names alone.

Zopolis4:
Fair enough-- I'm open to alternative name suggestions, as this was the best I could think of at the moment. What about something like `Style/EmptyLiteralInsideInterpolation`? `Style/EmptyReturnInsideInterpolation`?

bbatsov:
I haven't forgotten about this PR, but I'm still not sure about how to name it exactly. I've also been thinking today that probably it'd be good if the style was configurable, as I don't think there will an agreement about the explicit mentions of empty literals.

lib/rubocop/cop/style/empty_string_inside_interpolation.rb
```diff
@@ -0,0 +1,61 @@
...
+      #   # bad
+      #   "#{condition ? '' : 'foo'}"
+      #
+      #   # good
+      #   "#{'foo' unless condition}"
+      #
+      class EmptyStringInsideInterpolation < Base
+        include Interpolation
+        extend AutoCorrector
+        MSG = 'Do not assign empty strings to variables inside string interpolation.'
```
bbatsov, line 29: I find the mention of "variables" here confusing.

Zopolis4: Would `Do not assign empty strings inside string interpolation.` be a better way of phrasing it?

Zopolis4: Fixed.

Zopolis4:
Added alternative style. Not 100% sure about naming and wording of the initial explanation of the cops purpose, but the functionality is all there.

PR #13267, opened 2024-09-27
[Fix #13256] Report and correct more `Style/SafeNavigation` offenses.

### LINKED ISSUE #13256: False negative for `Style/SafeNavigation` when safe navigation is not used exclusivly
## Expected behavior

Offense for the following:

```rb
foo? && bar && bar.end_with?('baz')
```

## Actual behavior

No offense.

When it comes after, it works:

```rb
bar && bar.end_with?('baz') && foo?
```

## RuboCop version

```
$ [bundle exec] rubocop -V
1.66.1 (using Parser 3.3.5.0, rubocop-ast 1.32.3, running on ruby 3.3.5) [x86_64-linux]
  - rubocop-performance 1.22.1
  - rubocop-rake 0.6.0
  - rubocop-rspec 3.0.4
```

dvandersluis:
This change adds the ability for `Style/SafeNavigation` to detect offenses such as

```ruby
foo? && bar && bar.end_with?('baz')
```

`and` nodes nest leftward, ie. the above code becomes 

```
(and
  (and
    (send nil :foo?)
    (send nil :bar))
  (send
    (send nil :bar) :end_with?
    (send nil :baz)))
```

Because of this, the previously version of the cop was unable to find an offense for the above code because it was comparing `foo? && bar` with `bar.end_with?`. In order to support this, I had to greatly rewrite how `on_and` processes nodes. The new approach collects all the clauses of a (nested) `and` statement, and evaluates them in pairs. If a pair has an object detection on the LHS and a method call of that object on the RHS, it will be picked up by the cop as an offense and corrected. I also handled `begin` nodes within `and` nodes so code like `foo && (bar && bar.x?)` will be caught as well.

I mostly did not change the `on_if` code, other than using my new methods where appropriate. The core of it that finds which elements to operate on has not changed.

As part of my refactor, I added tests for a number of cases that were being covered by the cop but not tested for, to ensure the behaviour remains the same. I have run this new version of the cop against a number of codebases and do not get any crashes or false positives that I've seen.

A few new offenses were detected in the RuboCop codebase by the changes, which are fixed in the second commit.

Fixes #13256.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_report_and_correct_more_style_safe_navigation.md
+++ b/changelog/change_report_and_correct_more_style_safe_navigation.md
@@ -1,0 +1,1 @@
+* [#13256](https://github.com/rubocop/rubocop/issues/13256): Report and correct more `Style/SafeNavigation` offenses. ([@dvandersluis][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2332,7 +2332,7 @@ Lint/SafeNavigationConsistency:
                  for all method calls on that same object.
   Enabled: true
   VersionAdded: '0.55'
-  VersionChanged: '0.77'
+  VersionChanged: '<<next>>'
   AllowedMethods:
     - present?
     - blank?
--- a/lib/rubocop/config_loader_resolver.rb
+++ b/lib/rubocop/config_loader_resolver.rb
@@ -166,7 +166,7 @@ module RuboCop
       return unless duplicate_setting?(base_hash, derived_hash, key, opts[:inherited_file])
 
       inherit_mode = opts[:inherit_mode]['merge'] || opts[:inherit_mode]['override']
-      return if base_hash[key].is_a?(Array) && inherit_mode && inherit_mode.include?(key)
+      return if base_hash[key].is_a?(Array) && inherit_mode&.include?(key)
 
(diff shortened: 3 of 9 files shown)


Earlopain:
I ran this over real-world-rails and found some erroring (simplified for clarity):

```rb
foo? && begin
  bar? && baz?
end
# undefined method `receiver' for an instance of Parser::Source::Range
# lib/rubocop/cop/style/safe_navigation.rb:295:in `find_matching_receiver_invocation'
```

```rb
return if foo? && bar = baz && bat != bar
# undefined method `receiver' for an instance of Parser::Source::Range
# lib/rubocop/cop/style/safe_navigation.rb:295:in `find_matching_receiver_invocation'
```

```rb
foo? && (bar? || baz? && bat?)
# undefined method `receiver' for an instance of Parser::Source::Range
# lib/rubocop/cop/style/safe_navigation.rb:295:in `find_matching_receiver_invocation'
```

Although they look quite different, seems like one fix will take care of all.

dvandersluis:
Thanks @Earlopain I'll check those out!

bbatsov:
Excellent improvement!

PR #13269, opened 2024-09-28
Refactor `style/select_by_regexp_spec` to prevent overwriting `ruby_version` within contexts

masato-bkn:
This PR refactors `style/select_by_regexp_spec` to prevent overwriting `ruby_version` within contexts.
I have split the changes into multiple commits to make the review process easier. If you prefer to squash commits, I will do so.

### Background
In `style/select_by_regexp_spec`, the `ruby_version` is being overwritten within the context of the shared_examples. As a result, the test is executed with a different `ruby_version` than the `ruby_version` specified in the caller of shared_examples.

For example, in `context 'when Ruby <= 2.2', :ruby22`, `ruby_version` 2.2 is specified, but due to a different `ruby_version` being set within the following contexts inside the shared_examples, the tests are actually being run with a different version than intended:

- https://github.com/rubocop/rubocop/blob/master/spec/rubocop/cop/style/select_by_regexp_spec.rb#L330
- https://github.com/rubocop/rubocop/blob/master/spec/rubocop/cop/style/select_by_regexp_spec.rb#L443

### Motivation
This mismatch between test case names and the actual behavior is not ideal, as it can lead to confusion. It may also result in bugs or omissions in the specifications.
For maintainability, I believe this is a good opportunity to refactor the test structure.

In fact, when considering adding support for `filter` in `Style/SelectByRegexp`, I found it difficult to determine how to modify the tests due to this issue.

-----------------

* [ ] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/select_by_regexp_spec.rb
+++ b/spec/rubocop/cop/style/select_by_regexp_spec.rb
@@ -261,113 +261,117 @@ RSpec.describe RuboCop::Cop::Style::SelectByRegexp, :config do
           ::ENV.#{method} { |x| x.match? /regexp/ }
         RUBY
       end
+    end
+  end
 
-      context 'with `numblock`s', :ruby27 do
-        it 'registers an offense and corrects for `match?`' do
-          expect_offense(<<~RUBY, method: method)
-            array.#{method} { _1.match? /regexp/ }
-            ^^^^^^^{method}^^^^^^^^^^^^^^^^^^^^^^^ #{message}
-          RUBY
+  shared_examples 'regexp match with `numblock`s' do |method, correction|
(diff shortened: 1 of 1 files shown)


koic:
In reality, the working procedures vary from person to person and are not necessary. What matters is the level of granularity required. Personally, And I don't prefer `shared_examples`, but let's proceed with this. Please squash the commits into one.

masato-bkn:
I squashed the associated commits.

At first, I wasn’t sure if removing `shared_examples` all at once was the right approach, so I followed the pattern used in other specs. Therefore, in this PR, I used `shared_examples`. However, ultimately, I think it would be better to write the tests directly without generalization, even if it increases the amount of code.

koic:
Yup, I understand that it follows the existing style in the file. That’s fine. Thank you!

PR #13273, opened 2024-09-29
Only raise for warnings in CI

Earlopain:
It's OK to only print these during development for convenience. It's how rails does it as well.

### DIFF
--- a/spec/support/strict_warnings.rb
+++ b/spec/support/strict_warnings.rb
@@ -33,6 +33,9 @@ module StrictWarnings
     # Ignore warnings from dev/rc ruby versions. Things are subject to change and
     # contributors should not be bothered by them with red CI.
     return if RUBY_PATCHLEVEL == -1
+    # Don't raise for warnings during development. It's expected that
+    # some code will warn like "unused variable" while iterating.
+    return unless ENV['CI']
 
     raise WarningError, message
   end


koic:
> It's how rails does it as well.

Can you provide a reference URL?

Earlopain:
Yeah, sure: https://github.com/rails/rails/blob/3780c138694091140c52601f6214f12bc7221eac/activesupport/lib/active_support/testing/strict_warnings.rb#L36

Since they use buildkite it looks a bit different and they also have a different env to force enable it locally but I don't think that's necessarily.

koic:
Thank you for the reference. However, my understanding is that this module is used in test code. This means that the behavior during test execution would differ between local and CI environments with this patch.

1. Wouldn’t this lead to cases where issues unnoticed locally are first caught in the CI environment? It seems more reasonable to be able to catch such issues during local test execution.
2. Ultimately, it is expected that there will be no warnings, such as unused variables, during the final test execution. Since tests should be run when the coding has settled to some extent, it seems more reasonable to be able to catch issues locally as things currently stand.

In the end, introducing behavior differences based on `ENV['CI']` involves a trade-off. Honestly, I think it's better to maintain the consistent behavior between local and CI environments, rather than introducing this trade-off.

Earlopain:
> Wouldn’t this lead to cases where issues unnoticed locally are first caught in the CI environment? 

Theoretically, yes. The warnings however are not hidden and still appear in your output, it will just not raise. This is especially important when the warning doesn't happen at runtime but during parsing, like "unused variable" warnings. Those prevent the tests from running at all.

> Ultimately, it is expected that there will be no warnings, such as unused variables, during the final test execution

If I'm almost done, yes. But I run tests often, and code is not always in a finished state and sometimes I want to test some things out and (for example) add an unconditional `return` somewhere which then complains about unused code:
```rb
$ bundle exec rspec
/home/user/code/rubocop/lib/rubocop/cop/base.rb:61: warning: statement not reached

An error occurred while loading spec_helper.
Failure/Error: raise WarningError, message

StrictWarnings::WarningError:
  /home/user/code/rubocop/lib/rubocop/cop/base.rb:61: warning: statement not reached
# ./spec/support/strict_warnings.rb:34:in 'StrictWarnings#warn'
# ./lib/rubocop.rb:49:in 'Kernel#require_relative'
# ./lib/rubocop.rb:49:in '<top (required)>'
# ./spec/spec_helper.rb:10:in '<top (required)>'
No examples found.


Finished in 0.00007 seconds (files took 0.31768 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

I guess it kinda depends on your workflow. Both I and @dvandersluis have encountered this.

With this change the warnings are mixed into test output which still should be plenty noticable:
```rb
$ bundle exec rspec
lib/rubocop/cop/base.rb:61: warning: statement not reached

Randomized with seed 49864
..................................................
etc.etc.
```

koic:
>  The warnings however are not hidden and still appear in your output, it will just not raise.

OK. Let's proceed with this for now, following the practice in the Rails repository.

PR #13276, opened 2024-09-30
Tweak docs for `Offense.correctable?`

Earlopain:
Pretty much every offense can be disabled through a todo comment. This method will only consider those when the user actually requested that.

At least that is how I understand this to work.

### DIFF
--- a/lib/rubocop/cop/offense.rb
+++ b/lib/rubocop/cop/offense.rb
@@ -97,8 +97,8 @@ module RuboCop
       # @!attribute [r] correctable?
       #
       # @return [Boolean]
-      #   whether this offense can be automatically corrected via
-      #   autocorrect or a todo.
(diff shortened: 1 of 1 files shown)


lib/rubocop/cop/offense.rb
```diff
@@ -98,7 +98,7 @@ def initialize(severity, location, message, cop_name, # rubocop:disable Metrics/
       #
       # @return [Boolean]
       #   whether this offense can be automatically corrected via
-      #   autocorrect or a todo.
+      #   autocorrect. This includes todo comments, when requested.
```
koic, line 101: It would be better to add a bit more explanation about "when requested".

Earlopain: 👍 I had that in there initially but thought it was too much

PR #13277, opened 2024-09-30
Generate predictable header ids for duplicate headers

Earlopain:
Instead of:
* https://docs.rubocop.org/rubocop/1.66/cops_bundler.html#examples
* https://docs.rubocop.org/rubocop/1.66/cops_bundler.html#examples-2
* https://docs.rubocop.org/rubocop/1.66/cops_bundler.html#examples-3

Do this:
* https://docs.rubocop.org/rubocop/1.66/cops_bundler.html#examples-bundlerduplicatedgem
* https://docs.rubocop.org/rubocop/1.66/cops_bundler.html#examples-bundlerduplicatedgroup
* https://docs.rubocop.org/rubocop/1.66/cops_bundler.html#examples-bundlergemcomment

Links like this will continue to work when new cops are added. Anchors to cops themselves are kept as is, since these already consistently work between versions.

Right now, adding a new cop potentially shifts these numbers around which causes links like that to go to a different place than expected when linking to non-versioned doc links like https://docs.rubocop.org/rubocop/cops_style.html#safety-54

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cops_documentation_generator.rb
+++ b/lib/rubocop/cops_documentation_generator.rb
@@ -11,13 +11,13 @@ class CopsDocumentationGenerator # rubocop:disable Metrics/ClassLength
   )
 
   STRUCTURE = {
-    name:                  ->(data) { h2(data.cop.cop_name) },
+    name:                  ->(data) { cop_header(data.cop) },
     required_ruby_version: ->(data) { required_ruby_version(data.cop) },
     properties:            ->(data) { properties(data.cop) },
     description:           ->(data) { "#{data.description}\n" },
-    safety:                ->(data) { safety_object(data.safety_objects) },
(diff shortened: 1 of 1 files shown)


bbatsov:
> https://docs.rubocop.org/rubocop/1.66/cops_bundler.html#examples-bundlerduplicatedgem

In the future we might consider adding some dash between the department and the cop name (for better readability), but that's a minor thingy.

Earlopain:
That would need more careful consideration. For example, the docs urls produced by rubocop are currently unversioned (and I think that's good), the anchors would then break for all previous versions. I'm not sure if you can have two anchors to the same place (you probably can).

If it does get changed, the best time is now, at least for the newly added anchors so that they don't have compatibility concerns like that.

Let me know if I should tweak the new anchors like that and I'll put up a PR.

PR #13281, opened 2024-10-01
Support Ruby 3.4 for `Lint/UriRegexp`

koic:
This PR supports Ruby 3.4 for `Lint/UriRegexp` to avoid the following warning:

```console
$ ruby -ruri -vwe "URI::DEFAULT_PARSER.make_regexp('http://example.com')"
ruby 3.4.0dev (2024-09-21T18:00:23Z master 9f574fa12f) +PRISM [x86_64-darwin23]
-e:1: warning: URI::RFC3986_PARSER.make_regexp is obsoleted. Use URI::RFC2396_PARSER.make_regexp explicitly.
```

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_support_ruby34_for_lint_uri_regexp.md
+++ b/changelog/change_support_ruby34_for_lint_uri_regexp.md
@@ -1,0 +1,1 @@
+* [#13281](https://github.com/rubocop/rubocop/pull/13281): Support Ruby 3.4 for `Lint/UriRegexp` to avoid obsolete API. ([@koic][])
--- a/lib/rubocop/cop/lint/uri_regexp.rb
+++ b/lib/rubocop/cop/lint/uri_regexp.rb
@@ -3,29 +3,47 @@
 module RuboCop
   module Cop
     module Lint
-      # Identifies places where `URI.regexp` is obsolete and should
-      # not be used. Instead, use `URI::DEFAULT_PARSER.make_regexp`.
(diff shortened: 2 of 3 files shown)


Earlopain:
I have some concerns about how default/bundled gems are handled at the moment. `faraday` for example transitivly depends on `uri` which decouples it from the ruby version for a large number of users. Rails also recently added a dependency on `uri` (0.13.1, ensuring `RFC2396_PARSER` always exists)

I think, people on latest ruby that also don't update their dependencies are rare. It also only registers an offense for `URI.regexp` and not `URI::DEFAULT_PARSER.regexp` so for most this won't change anything.

Something like `target_gem_version(:uri)` to mirror `target_ruby_version`. But for now, this seems fine since it still only concerns itself with `URI.regexp`.

koic:
Ideally, something like `target_gem_version(:uri)` would exist, but it currently does not. The goal of this PR is to address the current situation where incorrect 'good' code is suggested despite autocorrection. In doing so, the most useful API at the moment is `target_ruby_version`. If `target_gem_version` is introduced in the future, it can be replaced accordingly.

The gemification of Ruby has progressed in recent years, from the perspective of RuboCop's history. An API like `target_gem_version` would respond to this trend (though it is beyond the scope of this PR).

bbatsov:
The proposed solution seems reasonable to me. Thanks!

PR #13287, opened 2024-10-02
Add new `Lint/ObjectEqualityOverride` cop

### LINKED ISSUE #12289: New cop to detect unsafe comparison method definitions
When defining comparison methods (`==` and `eql?`), it is easy to miss that the `other` should be of the same class (or some other class which we should also check explicitly) and write a non safe method.

Example: 
```ruby
# bad
def ==(other)
  attributes == other.attributes
end

# good
def ==(other)
  self.class == other.class && # or kind_of? or is_a? or instance_of?
    attributes == other.attributes
end

# also good
def ==(other)
  attributes == other&.attributes
end
```

Inspired by https://github.com/rails/rails/issues/49670.

viralpraxis:
This PR adds new `Lint/UnsafeEqualityMethodDefinition` cop that checks for `eql?` and `==` methods without argument class check.

```ruby
# bad
def eql?(other)
  title == other.title
end

#good
def eql?(other)
  other.is_a?(self.class) && title == other.title
end
```

NOTE: currently there are exactly 5 offenses (some of them belong to public classes AFAICS) in the rubocop's codebase itself:

```
== lib/rubocop/cop/offense.rb ==
W:211:  7: Lint/UnsafeEqualityMethodDefinition: Check against argument class within custom equality methods.
== lib/rubocop/cop/registry.rb ==
W:232:  7: Lint/UnsafeEqualityMethodDefinition: Check against argument class within custom equality methods.
== lib/rubocop/cop/severity.rb ==
W: 50:  7: Lint/UnsafeEqualityMethodDefinition: Check against argument class within custom equality methods.
== lib/rubocop/cop/variable_force/branch.rb ==
W:121: 11: Lint/UnsafeEqualityMethodDefinition: Check against argument class within custom equality methods.
== lib/rubocop/cop/variable_force/scope.rb ==
W: 35:  9: Lint/UnsafeEqualityMethodDefinition: Check against argument class within custom equality methods.
```

I'm not sure what to do with them.

Fixes #12289.

-----------------

* [x] Added tests.


koic:
Before diving into the implementation details, here is some feedback on the requirements:

1. I don't have a strong objection to performing class checks for object identity inspected with `eql?`, but can you provide a URL as a basis for this? I'm curious whether always including class comparisons in the overrides, as shown in the good examples, is logically consistent with the language specifications.
2. `==` checks for object equality. I think performing class checks for `==` is excessive. For example, in cases like `1 == 1.0 #=> true`, equality via `==` should be delegated to the context of the class. Therefore, please exclude this from the scope of the cop’s inspection.
3. The name `Lint/UnsafeEqualityMethodDefinition` feels off. `Lint/ObjectIdentificationDefinition` seems to be a better fit.

viralpraxis:
> Before diving into the implementation details, here is some feedback on the requirements:
> 
> 1. I don't have a strong objection to performing class checks for object identity inspected with `eql?`, but can you provide a URL as a basis for this? I'm curious whether _always_ including class comparisons in the overrides, as shown in the good examples, is logically consistent with the language specification.

I don’t have anything ruby-specific at hand, but it is widely accepted that the `:==` binary relation (setting aside the minor differences between `:==` and `:eql?`) should be an equivalence relation, meaning it must be reflexive, transitive, and symmetric. Checking whether the argument passed to `:==` is an instance of the same class (via `other.instance_of?`) is a necessary, though not sufficient, condition to ensure these properties (it doesn't work in the case of `kind_of?` but people use it everywhere anyway). 

Although the original motivation for this cop was to prevent runtime errors related to object comparisons (see the [original issue](https://github.com/rubocop/rubocop/issues/12289)), I firmly believe that the absence of class checks can lead to difficult-to-diagnose bugs. I recall struggling with a ruby-i18n-related bug after upgrading from Ruby 3.0.0 to 3.0.1, which was ultimately caused by a missing class check (see https://stackoverflow.com/questions/77830896/hash-related-behavior-mismatch-between-ruby-mri-3-0-0-and-3-0-1/77841475#77841475 for details)

koic:
Well, `==` and `eql?` are expected to be overridden for equality checks based on the context. In other words, whether they maintain identity like `equal?` by default or are used for equality depends on the nature of the class. Considering the convenience for users, I have concerns about enforcing such context-dependent usage uniformly through a Lint rule.

For example, I think linter that enforces overriding `hash` method whenever `eql?` method is overridden can be uniformly implemented. This is because it is specified in the language specification as follows.
https://ruby-doc.org/3.3.5/Object.html#method-i-eql-3F

viralpraxis:
Thanks, I understand your point now. Unfortunately, ATM I don't see anything related in the specification. Also, some built-in classes [use this approach](https://github.com/ruby/ruby/blob/a8a921aef3e3f5a91425e53e4faf44d032a32a97/struct.c#L1484)

viralpraxis:
Fixed the issues you mentioned (`#==` and naming) just in case you think it worth merging. Lint now passes.

bbatsov:
I understand the proposal, but something like this needs an iron-clad rationale before we can merge it. (mostly because I don't think such code is particularly common in the wild)

@rubocop/rubocop-core As usual, in such tricky situations your thoughts would be most appreciated!

spec/rubocop/cop/lint/object_equality_override_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    RUBY
+  end
+
+  it 'does not register an offense when classes are checked with `<=` method' do
+    expect_no_offenses(<<~RUBY)
+      def eql?(other)
+        other.class <= self.class && self.title == other.title
+      end
+    RUBY
+  end
```
dvandersluis, line 20: Shouldn't the same thing apply for the `<` operator?

config/default.yml
```diff
@@ -2138,6 +2138,11 @@ Lint/NumberedParameterAssignment:
   Enabled: pending
   VersionAdded: '1.9'
 
+Lint/ObjectEqualityOverride:
+  Description: 'Check against argument class within overridden equality methods.'
```
dvandersluis, line 2142:
```suggestion
  Description: 'Check against argument class within overridden `eql?` methods.'
```

lib/rubocop/cop/lint/object_equality_override.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Identifies `eql?` method definitions which do not
+      # check for argument class which might lead to runtime errors.
```
dvandersluis, line 7: There's a number of places in the PR where wording like this pops up but I don't think it's clear. Something like this might be better:

> Identifies `eql?` method definitions which do not check the argument's class is the same as, or a descendant of, the current class.

Since there are more "good" possibilities than are enumerated in `@example`, we should be more comprehensive.

Additionally "might lead to runtime errors" isn't a strong rationale IMO. What sort of runtime errors, and how does checking the class improve that?

lib/rubocop/cop/lint/object_equality_override.rb
```diff
@@ -0,0 +1,57 @@
...
+      #   def eql?(other)
+      #     other.is_a?(self.class) && other.name == name
+      #   end
+      #
+      #   # good
+      #   def eql?(other)
+      #     other.instance_of?(self.class) && to_s == other.to_s
+      #   end
+      class ObjectEqualityOverride < Base
+        MSG = 'Check against argument class within custom equality methods.'
```
dvandersluis, line 31: A thought for a more concise message:

```suggestion
        MSG = 'Check the class of `%<parameter_name>s` in `eql?`'
```

dvandersluis:
This will likely register a false positive for methods along the lines of (`&&` could be `||`, etc.):

```ruby
def eql?(other)
  super && self.title == other.title
end
```

dvandersluis:
Another class of false positive:

```ruby
def eql?(other)
  # where CLASS_NAME is the actual class constant
  other.is_a?(CLASS_NAME) && self.title == other.title
end
```

lib/rubocop/cop/lint/object_equality_override.rb
viralpraxis: Thanks.

config/default.yml
viralpraxis: Thanks.

spec/rubocop/cop/lint/object_equality_override_spec.rb
viralpraxis: Yep, that's true. Fixed.

PR #13288, opened 2024-10-02
[Fix #12988] Add new `Style/AmbiguousEndlessMethodDefinition` cop

### LINKED ISSUE #12988: Endless Method Keyword/Operator Precedence Check
## Is your feature request related to a problem?

Consider the following code

```ruby
def foo = true and false
def bar = true && false

foo # => true
bar # => false
```

Because the `=` operator has higher precedence than `and` or `or`, they will not be included as part the method.

A rubocop check to detect operators/keywords with lower precedence than `=` on the same line as an endless expression would be incredibly helpful in avoiding this sort of footgun.

dvandersluis:
Adds a new `Style/AmbiguousEndlessMethodDefinition` cop to look for endless methods that are wrapped in `and`, `or`, `if`, `unless`, `while` or `until` nodes but likely cause confusion due to operator precedence.

For example:

```ruby
def foo = true if bar
```

may appear to be `def foo = (true if bar)` but is actually parsed as `(def foo = true) if bar`. The cop corrects by wrapping the clause in parentheses to make the former explicit.

Autocorrection is not safe because the implicit precedence may actually be desired.

-----------------

* [x] Added tests.


koic:
> **Question**: should this be in the `Lint` namespace instead, to match `Lint/AmbiguousOperatorPrecedence`? Should the name be changed to match?

As ongoing in https://github.com/rubocop/rubocop/pull/11111, this should be categorized under Style department, not Lint department, at the least. This is just a note, TBH, I have doubts about whether it’s worth providing this as a cop. It seems like the type of issue that should be addressed by LSP's Inlay Hints.

lib/rubocop/cop/style/endless_method_operator_precedence.rb
```diff
@@ -0,0 +1,80 @@
...
+      #   def foo = true and false
+      #
+      #   # good
+      #   def foo = (true and false)
+      #
+      #   # bad
+      #   def foo = true if bar
+      #
+      #   # good
+      #   def foo = (true if bar)
```
koic, line 29: I feel uncomfortable with presenting something that is not compatible with the bad example as good. This could be misleading for developers who properly understand the precedence of operators.

lib/rubocop/cop/style/endless_method_operator_precedence.rb
```diff
@@ -0,0 +1,80 @@
...
+        include RangeHelp
+
+        minimum_target_ruby_version 3.0
+
+        MSG = 'Use parentheses around `%<keyword>s` statement in endless method body.'
+
+        # @!method ambiguous_endless_method_body(node)
+        def_node_matcher :ambiguous_endless_method_body, <<~PATTERN
+          ^${
+            (if _ <def _>)
```
phene, line 42: What about `unless` ?

dvandersluis: There is no specific `unless` node, the unless modifier parses into an `if` node with only an "else" clause.

```
$ ruby-parse --legacy -e 'def foo = true unless bar'
(if
  (send nil :bar) nil
  (def :foo
    (args)
    (true)))
```

lib/rubocop/cop/style/endless_method_operator_precedence.rb
dvandersluis: Yes this is a good point that the good case has different behaviour than the bad case. I was thinking that the parens in the good case would be what was actually intended for the majority of cases like this, but making the parens explicit for the actual precedence is possible too. 

I am not sure that it is very likely that someone would write code along these lines and actually mean for the and/or/if etc. to modify the `def` rather than the `def` body, but I could be wrong. Maybe @phene or @vlad-pisanov can weigh in given their participation in the issue.

koic: It doesn't seem appropriate to use endless method definitions in cases where they can cause confusion. Methods that can lead to ambiguity without specifying operator precedence are likely not suited for endless method definitions in the first place.
In such cases, I think it's better to suggest using the regular method definition style. For example, the following would be bad and good examples:

```ruby
      #   # bad
      #   def foo = true and false
      #
      #   # good - When the method body is the intended focus. This is incompatible behavior.
      #   def foo
      #     true and false
      #   end
      #
      #   # good - When the intended meaning is clear. This is compatible behavior.
      #   def foo
      #     true
      #   end and false
```

Also, because the user intent can't be made completely clear, I think autocorrect should not be supported.

dvandersluis:
@koic thank you for your input. What do you think about having the autocorrect change to your "compatible" form?

dvandersluis:
I have updated the cop examples as per @koic's suggestion, and changed autocorrect to replace with a normal method definition instead of an endless method, in order to reduce ambiguity but retain the behaviour as is. @koic: I think this autocorrect is appropriate now, but if you still feel strongly about not having autocorrect I can remove it!

dvandersluis:
@bbatsov thank you for the review! I've updated the name as you suggested.

dvandersluis:
@bbatsov I've opened a ruby-style-guide PR here: https://github.com/rubocop/ruby-style-guide/pull/949

PR #13290, opened 2024-10-03
Allow Style/WordArray to match words with non-whitespace characters

floriandejonckheere:
Allow `Style/WordArray` cop to match words that include non-whitespace characters. Before this change, words could only include letters, digits, underscores, and hyphens. After this change, words can include any non-whitespace characters. The word regex now uses the `\P{Z}` character class, which matches any non-whitespace character.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/word_array.rb
+++ b/lib/rubocop/cop/style/word_array.rb
@@ -68,6 +68,13 @@ module RuboCop
       #     %w[two Two]
       #   ]
       #
+      # @example UseExtendedWords: true
+      #   # good
+      #   %w(foo bar foo-bar foo/bar foo#bar foo?bar foo@bar)
+      #
+      #   # bad
+      #   ['foo', 'bar', 'foo-bar', 'foo/bar', 'foo#bar', 'foo?bar', 'foo@bar']
+      #
       class WordArray < Base
         include ArrayMinSize
         include ArraySyntax
@@ -132,7 +139,11 @@ module RuboCop
         end
 
         def word_regex
-          Regexp.new(cop_config['WordRegex'])
+          if cop_config['UseExtendedWords']
+            Regexp.new(cop_config['ExtendedWordRegex'])
+          else
+            Regexp.new(cop_config['WordRegex'])
+          end
         end
 
         def build_bracketed_array(node)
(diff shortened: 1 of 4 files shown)


bbatsov:
I'm OK with the proposed change, but I'm wondering if we shouldn't make this configurable as it will result in quite a few new offenses for our end users. @rubocop/rubocop-core Any thoughts?

koic:
Yeah, I'm also primarily concerned about the impact on users. Moreover, the name of this Cop is `Style/WordArray`, which includes the word "Word"' not "Element"' I'm not confident whether it’s appropriate to consider something like `Department/Cop` (units containing `/`) as a "Word" in this context.

floriandejonckheere:
I actually started this PR exactly because some string containing slashes were not matched by `Style/WordArray`. While this change may result in a number of offenses on existing code, they should not have any further impact, as it's only about syntax.

Perhaps we could introduce a configuration or even a separate cop for "extended" word arrays?

bbatsov:
Yeah, I think at the bare minimum we need a configuration whether to consider spaces in words or not. But a new cop is also an option, as I'd be wary to add nuances to something as simple as "word".

bbatsov:
So, what are we doing here - config option or a new cop? @rubocop/rubocop-core Your opinion on the topic would be welcome!

docs/modules/ROOT/pages/cops_style.adoc
```diff
@@ -18307,7 +18307,16 @@ array of 2 or fewer elements.
 
 | WordRegex
 | `(?-mix:\A(?:\p{Word}\|\p{Word}-\p{Word}\|\n\|\t)+\z)`
-|
```
dvandersluis, line 18310: The adoc files are automatically generated from the configuration and comments in the class, could you revert this file?

lib/rubocop/cop/style/word_array.rb
```diff
@@ -68,6 +68,13 @@ module Style
       #     %w[two Two]
       #   ]
       #
+      # @example UseExtendedWords: true
+      #   # good
+      #   %w(foo bar foo-bar foo/bar foso#bar foo?bar foo@bar)
```
dvandersluis, line 73:
```suggestion
      #   %w(foo bar foo-bar foo/bar foo#bar foo?bar foo@bar)
```

dvandersluis:
My concern here is that we may end up with situations where users configure the "wrong" one of `WordRegexp` and `ExtendedWordRegexp` without realizing the use of it. It'd be best to only have a single configuration but I'm not exactly sure how we can do this. @bbatsov any thoughts?

bbatsov:
Yeah, I share your concerns and that's why I haven't pulled the trigger on this PR yet. I've been thinking that perhaps all we need is a bit of documentation for people to update the existing regexp accordingly if they need to use extended words. 

We can even have it commenting out in the default config or something like this.

PR #13296, opened 2024-10-05
Add `Bundler/DebugRequire` cop

sambostock:
The debug gem can create unnecessary overhead for applications, especially when it's not explicitly required. This can lead to unintended debugger activation in local development environments, potentially impacting performance and causing unexpected behaviour.

This cop enforces that if `gem "debug"` is present, one of the following is specified:

    gem "debug", ... require: "debug/prelude"
    gem "debug", ... require: false

Thanks to @alexcrocha for originally implementing a version of this cop in our codebase!

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/new_add_bundler_debug_require_cop.md
+++ b/changelog/new_add_bundler_debug_require_cop.md
@@ -1,0 +1,1 @@
+* [#13296](https://github.com/rubocop/rubocop/pull/13296): Add `Bundler/DebugRequire` cop. ([@sambostock][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -176,6 +176,15 @@ AllCops:
 
 #################### Bundler ###############################
 
+Bundler/DebugRequire:
+  Description: 'Enforces that if `gem "debug"` is present, `require: "debug/prelude"` or `require: false` is specified.'
+  Enabled: pending
+  VersionAdded: '<<next>>'
+  Include:
+    - '**/*.gemfile'
+    - '**/Gemfile'
+    - '**/gems.rb'
+
 Bundler/DuplicatedGem:
   Description: 'Checks for duplicate gem entries in Gemfile.'
   Enabled: true
--- a/lib/rubocop.rb
+++ b/lib/rubocop.rb
@@ -165,6 +165,7 @@ require_relative 'rubocop/cop/correctors/space_corrector'
 require_relative 'rubocop/cop/correctors/string_literal_corrector'
 require_relative 'rubocop/cop/correctors/unused_arg_corrector'
 
+require_relative 'rubocop/cop/bundler/debug_require'
 require_relative 'rubocop/cop/bundler/duplicated_gem'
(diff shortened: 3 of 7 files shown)


sambostock:
I've extracted the first commit into #13297, which fixes Ruby HEAD compatiblity on `master`. I can rebase once it's merged.

Earlopain:
I feel like this is something that should instead be addresses upstream, if at all possible. https://github.com/ruby/debug/issues/797 hasn't been rejected and the cop could become irrelevant tomorrow.

koic:
I agree with @Earlopain's opinion. Moreover, `require: false` helps reduce overhead for tools beyond just `debug`, such as RuboCop and gems. Therefore, it may not be appropriate to consider this a `debug` gem-specific issue.

As an off-topic note, a cop could be considered to check the necessity of `require: false`, but supporting the ever-growing number of gems would involve considerable maintenance, raising doubts about whether the cost is justified.

For these reasons, I'll close this suggestion, but thank you for proposing it.

st0012:
> the cop could become irrelevant tomorrow.

As the author of the issue, I think it's extremely unlikely to happen. Changing the activation behaviour requires extensive testing on a few niche/edge usages, which I don't think anyone with enough context have time to do at the moment.

> it may not be appropriate to consider this a debug gem-specific issue.

While I agree that using RuboCop rules for gem-specific optimization isn't scalable, I'd argue that `debug` is a bundled gem of the language and a Rails default, it can have some special treatment.

st0012:
> I believe you are one yourself

No I'm not 🙂

st0012:
> Though the default rails gemfile now generates with require: "debug/prelude"

That's because I added it in https://github.com/rails/rails/pull/51692. But that doesn't help any existing applications.

I don't believe this is the best approach either as creating side-effects upon a gem is required is not an ideal behaviour, which I already pointed out in my issue.

But given the stale state of the issue and that people are spending 10% more time running tests on CI everyday, I believe there could be ways to stop the bleeding in other places. And a RuboCop rule is one of them.

That said, I also completely understand any opposition from RuboCop maintainers as it'd inevitably shift the responsibility of explaining the issue to them, while they don't have the necessary context.

koic:
> While I agree that using RuboCop rules for gem-specific optimization isn't scalable, I'd argue that `debug` is a bundled gem of the language and a Rails default, it can have some special treatment.

A more ideal cop would be one that allows users to customize which gem names should have `require: false` added, rather than treating the `debug` gem as a special case. Of course, the `debug` gem could be set as the default because, as you mentioned, it's bundled with Ruby. In any case, I think similar cops shouldn't be added every time a gem requiring `require: false` is introduced. Therefore, instead of a cop specific to `debug`, it should be a more generic cop.

Furthermore, considering the use case for the `debug` gem, it would be preferable to allow the `require` value to be set not only as `false`, but also as any custom value such as `"debug/prelude"`.

PR #13308, opened 2024-10-08
[Fix #12875] Fix false positive for `Style/ArgumentsForwarding` when argument is used inside a block

### LINKED ISSUE #12875: `Style/ArgumentsForwarding` enforce anonymous bad autocorrect
The following code:

```ruby
    def validators_on(*args)
      [
        super,
        resource_class.validators_on(*args),
        *@auxiliary_resource_classes.map do |klass|
          klass.validators_on(*args)
        end,
      ].flatten
    end
  end
```

results in invalid autocorrect output with

```yml
Style/ArgumentsForwarding:
  UseAnonymousForwarding: true
  Enabled: true
```

--------

## Expected behavior

```ruby
    def validators_on(*args)
      [
        super,
        resource_class.validators_on(*args),
        *@auxiliary_resource_classes.map do |klass|
          klass.validators_on(*args)
        end,
      ].flatten
    end
  end
```

## Actual behavior

```ruby
    def validators_on(*)
      [
        super,
        resource_class.validators_on(*),
        *@auxiliary_resource_classes.map do |klass|
          klass.validators_on(*args)
        end,
      ].flatten
    end
  end
```

`*args` will no longer be defined.

Example code above.

## RuboCop version

```
1.63.4 (using Parser 3.3.1.0, rubocop-ast 1.31.2, running on ruby 3.3.1) [arm64-darwin23]
  - rubocop-capybara 2.20.0
  - rubocop-factory_bot 2.25.1
  - rubocop-performance 1.20.2
  - rubocop-rails 2.24.0
  - rubocop-rspec 2.27.1
```

dvandersluis:
Previously:

```ruby
def foo(*args)
        ^^^^^ Style/ArgumentsForwarding: Use anonymous positional arguments forwarding (*).
  bar(*args)
      ^^^^^ Style/ArgumentsForwarding: Use anonymous positional arguments forwarding (*).
  baz do |x|
    x.foo(*args)
  end
end

# corrects to...

def foo(*)
  bar(*)
  baz do |x|
    x.foo(*args)
  end
end
```

This is problematic because `args` within the block was not changed but was changed outside of the block.

This change prevents offenses from being registered if the argument is used inside a block.

Fixes #12875.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/arguments_forwarding.rb
```diff
@@ -356,6 +357,12 @@ def use_anonymous_forwarding?
           cop_config.fetch('UseAnonymousForwarding', false)
         end
 
+        def sent_inside_block?(send_classifications)
```
koic, line 360: It seems to refer to `send`, so how about `send_inside_block?`, or if it refers to a method call in general, what do you think about names like `called_inside_block?` or `invoked_inside_block?` (or others...)?

dvandersluis: Sure thing, updated!

PR #13312, opened 2024-10-09
[Fix #9816] Refine `Lint/SafeNavigationConsistency`

### LINKED ISSUE #9816: Request: Lint/SafeNavigationConsistency short-circuit option for &&

koic:
Fixes #9816.

As highlighted in user feedback in #9816, the current implementation excessively requires the use of the safe navigation operator. This PR refines `Lint/SafeNavigationConsistency` cop to check that the safe navigation operator is applied consistently and without excess or deficiency.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/safe_navigation_consistency_spec.rb
+++ b/spec/rubocop/cop/lint/safe_navigation_consistency_spec.rb
@@ -3,6 +3,12 @@
 RSpec.describe RuboCop::Cop::Lint::SafeNavigationConsistency, :config do
   let(:cop_config) { { 'AllowedMethods' => %w[present? blank? try presence] } }
 
+  it 'allows && without receiver' do
+    expect_no_offenses(<<~RUBY)
+      foo && bar
+    RUBY
+  end
+
   it 'allows && without safe navigation' do
     expect_no_offenses(<<~RUBY)
       foo.bar && foo.baz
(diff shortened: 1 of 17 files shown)


spec/rubocop/cop/lint/safe_navigation_consistency_spec.rb
```diff
@@ -15,44 +21,74 @@
...
+  end
+
   it 'allows safe navigation when different variables are used' do
     expect_no_offenses(<<~RUBY)
       foo&.bar || foobar.baz
     RUBY
   end
 
-  it 'allows calls to methods that nil responds to' do
+  it 'allows an offense using calling `nil?` after safe navigation with `||`' do
```
dvandersluis, line 36: For this test and the next one it'd make more sense to say "does not register an offense" instead of "allows an offense" IMO

PR #13313, opened 2024-10-09
[Fix #13286] Add `AllowedMethods` and `AllowedPatterns` configuration to `Layout/FirstMethodArgumentLineBreak`

### LINKED ISSUE #13286: Don't warn about Layout/FirstMethodArgumentLineBreak for single-character method names
[Layout/FirstMethodArgumentLineBreak](https://docs.rubocop.org/rubocop/cops_layout.html#layoutfirstmethodargumentlinebreak) checks for a line break before the first argument in a multi-line method call. But, I argue that there should be exceptions. e.g.

```ruby
notice =
  t(   # <--- this line break is required by Layout/FirstMethodArgumentLineBreak
    "flash.successful_update_to",
    type: "username",
    to: current_user.display_name)
```
I often have to break my I18n.translate calls into multiple lines, but the `t(` having to be on its own line is just a waste of vertical space.

Either:
- Just always allow single-character methods
- Add an option to allow single-character methods
- Or, add an option to allow a list of method names, to which I can add `t`, `l`, and the like.

Sometimes I'll try to disable Layout/FirstMethodArgumentLineBreak, but it makes for an extra long line (and I try to stick to 80 chars):

```ruby
notice =
  t("flash.successful_update_to", # rubocop:disable Layout/FirstMethodArgumentLineBreak
    type: "username",
    to: current_user.display_name)
```

And wrapping the method is just extra noisy:

```ruby
# rubocop:disable Layout/FirstMethodArgumentLineBreak
notice =
  t("flash.successful_update_to",
    type: "username",
    to: current_user.display_name)
# rubocop:enable Layout/FirstMethodArgumentLineBreak
```

So I usually just punt and obey the cop:

```ruby
notice =
  t(
    "flash.successful_update_to",
    type: "username",
    to: current_user.display_name)

```

dvandersluis:
Add `AllowedMethods` and `AllowedPatterns` to `Layout/FirstMethodArgumentLineBreak` allow for users to configure for methods they do not want to apply to the cop.

Fixes #13286.

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -861,6 +861,7 @@ Layout/FirstMethodArgumentLineBreak:
   Enabled: false
   VersionAdded: '0.49'
   AllowMultilineFinalElement: false
+  AllowedMethods: []
 
 Layout/FirstMethodParameterLineBreak:
   Description: >-
--- a/lib/rubocop/cop/layout/first_method_argument_line_break.rb
+++ b/lib/rubocop/cop/layout/first_method_argument_line_break.rb
@@ -63,13 +63,21 @@ module RuboCop
       #     }
       #   )
       #
+      # @example AllowedMethods: ['some_method']
+      #
+      #   # good
+      #   some_method(foo, bar,
+      #     baz)
       class FirstMethodArgumentLineBreak < Base
         include FirstElementLineBreak
+        include AllowedMethods
         extend AutoCorrector
 
         MSG = 'Add a line break before the first argument of a multi-line method argument list.'
 
         def on_send(node)
+          return if allowed_method?(node.method_name)
+
           args = node.arguments.dup
 
           # If there is a trailing hash arg without explicit braces, like this:
--- a/changelog/change_add_allowed_methods_to_first_method_argument_line_break.md
+++ b/changelog/change_add_allowed_methods_to_first_method_argument_line_break.md
@@ -1,0 +1,1 @@
+* [#13286](https://github.com/rubocop/rubocop/issues/13286): Add `AllowedMethods` configuration to `Layout/FirstMethodArgumentLineBreak`. ([@dvandersluis][])
--- a/spec/rubocop/cop/layout/first_method_argument_line_break_spec.rb
+++ b/spec/rubocop/cop/layout/first_method_argument_line_break_spec.rb
@@ -141,4 +141,15 @@ RSpec.describe RuboCop::Cop::Layout::FirstMethodArgumentLineBreak, :config do
       RUBY
     end
   end
+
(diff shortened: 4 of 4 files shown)


lib/rubocop/cop/layout/first_method_argument_line_break.rb
```diff
@@ -63,13 +63,29 @@ module Layout
       #     }
       #   )
       #
+      # @example AllowedMethods: ['method']
```
koic, line 66: That's just my two cents. Since Ruby has the `method` method, an example like `some_method` might be preferred.

dvandersluis: Sure, I just maintained the pattern from the previous examples. Should I update the other examples or just the new ones?

koic: I see. The new examples will likely be sufficient.

dvandersluis: Updated!

config/default.yml
```diff
@@ -860,7 +860,10 @@ Layout/FirstMethodArgumentLineBreak:
                  multi-line method call.
   Enabled: false
   VersionAdded: '0.49'
+  VersionChanged: '<<next>>'
```
koic, line 863: Since no configuration changes have been made that would alter existing behavior, `VersionChanged '<<next>>'` is likely unnecessary.

config/default.yml
```diff
@@ -860,7 +860,10 @@ Layout/FirstMethodArgumentLineBreak:
...
   AllowMultilineFinalElement: false
+  AllowedMethods: []
+  AllowedPatterns: []
```
koic, line 866: Perhaps only `AllowedMethods` should be provided, and `AllowedPattern` might not be necessary.

dvandersluis: `AllowedPattern` would allow the original issue to be neatly resolved by matching all 1-character method names.

config/default.yml
dvandersluis: Sure, removed.

config/default.yml
koic: Certainly, it is possible to patternize a single character, but since method names typically start with a lowercase letter and the number of single-character method names is limited, I think they can be listed in `AllowedMethods`.

dvandersluis: I think there are other cases that a regexp would be useful, and personally would prefer configuring `/^.$/` over a list of method names were I to have this requirement; however, I have removed the `AllowedPattern` configuration from the cop in the interest of getting this done.

PR #13322, opened 2024-10-11
[Fix #13318] Prevent modifying blocks with `Style/HashEachMethods` if the hash is modified within the block

### LINKED ISSUE #13318: Autocorrection of `Style/HashEachMethods` can break code
## Description

Adding items to a hash while looping through the hash itself results in an error in ruby: `can't add a new key into hash during iteration (RuntimeError)`.
But in the case ...
```ruby
x = { a: 1 }
x.keys.each do |key|
  x["#{key}_dup"] = 2
end
```
... rubocop wants to autocorrects this code to ...
```ruby
x = { a: 1 }
x.each_key do |key|
  x["#{key}_dup"] = 2
end
```
... which will raise an error.

## Expected behavior

Does not autocorrect (and lint maybe?) such code.

## Actual behavior

Auto-correction breaks ruby code.

```shell
echo 'x = { a: 1 }; x.keys.each { |key| x["#{key}_dup"] = 2 }; puts x' > example.rb
ruby example.rb # works fine
rubocop -A example.rb
ruby example.rb # throws an error
```

## RuboCop version

```
$  rubocop -V
1.66.1 (using Parser 3.3.3.0, rubocop-ast 1.32.3, running on ruby 3.2.2) [x86_64-linux]
```

dvandersluis:
If `keys.each` or `values.each` block that modifies the original hash is changed to `each_key` or `each_value`, it results in `RuntimeError: can't add a new key into hash during iteration`.

This change updates `Style/HashEachMethods` to not register an offense in these cases.

Fixes #13318.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_prevent_modifying_blocks_with.md
+++ b/changelog/fix_prevent_modifying_blocks_with.md
@@ -1,0 +1,1 @@
+* [#13318](https://github.com/rubocop/rubocop/issues/13318): Prevent modifying blocks with `Style/HashEachMethods` if the hash is modified within the block. ([@dvandersluis][])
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

PR #13342, opened 2024-10-16
[Fix #13341] Fix false positives for `Lint/SafeNavigationChain`

### LINKED ISSUE #13341: Infinite loop for `Lint/SafeNavigationChain` and `Lint/SafeNavigationConsistency`
```ruby
foo&.bar&.baz && foo&.bar&.zoo
```

```
Offenses:

example.rb:3:26: W: [Corrected] Lint/SafeNavigationChain: Do not chain ordinary method call after safe navigation operator.
foo&.bar&.baz && foo&.bar.zoo
                         ^^^^
example.rb:3:26: W: [Corrected] Lint/SafeNavigationConsistency: Use . instead of unnecessary &..
foo&.bar&.baz && foo&.bar&.zoo
                         ^^

0 files inspected, 2 offenses detected, 2 offenses corrected
Infinite loop detected in example.rb and caused by Lint/SafeNavigationConsistency -> Lint/SafeNavigationChain
Hint: Please update to the latest RuboCop version if not already in use, and report a bug if the issue still occurs on this version.
Please check the latest version at https://rubygems.org/gems/rubocop.
```

koic:
Fixes #13341.

This PR fixes false positives for `Lint/SafeNavigationChain` when a safe navigation operator is used with a method call as the RHS operand of `&&` for the same receiver.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/safe_navigation_chain.rb
+++ b/lib/rubocop/cop/lint/safe_navigation_chain.rb
@@ -38,6 +38,8 @@ module RuboCop
         PATTERN
 
         def on_send(node)
+          return unless require_safe_navigation?(node)
+
           bad_method?(node) do |safe_nav, method|
             return if nil_methods.include?(method) || PLUS_MINUS_METHODS.include?(node.method_name)
 
@@ -51,6 +53,13 @@ module RuboCop
         end
 
         private
+
+        def require_safe_navigation?(node)
+          parent = node.parent
+          return true unless parent&.and_type?
+
+          parent.rhs != node || parent.lhs.receiver != parent.rhs.receiver
+        end
 
         # @param [Parser::Source::Range] offense_range
         # @param [RuboCop::AST::SendNode] send_node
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/safe_navigation_chain.rb
```diff
@@ -52,6 +54,13 @@ def on_send(node)
 
         private
 
+        def require_safe_navigation?(node)
+          parent = node.parent
+          return true unless parent&.and_type?
+
+          parent.rhs != node || parent.lhs.receiver != parent.rhs.receiver
```
fatkodima, line 61: Shouldn't we also check if `parent.lhs` is a `send` node before calling `.receiver`?

koic: As far as I have verified, it seems there are no cases where this does not respond to `RuboCop::AST::Node#receiver`.

fatkodima: Ah, nice, was thinking `receiver` is only for send nodes.

PR #13345, opened 2024-10-16
[Fix #13343] Prevent `Layout/LineLength` from breaking up a method with arguments chained onto a heredoc delimiter

### LINKED ISSUE #13343: `Layout/LineLength` can fold code inside a heredoc
```ruby
str = <<-HTML.strip_heredoc.strip_heredoc.strip_heredoc.encode!('utf-8', invalid: :replace, replace: '').strip.force_encoding('binary')
  <html>
  </html>
HTML
```

## Expected behavior

_idk_

## Actual behavior

```ruby
str = <<-HTML.strip_heredoc.strip_heredoc.strip_heredoc.encode!('utf-8', invalid: :replace, 
replace: '').strip.force_encoding('binary')
  <html>
  </html>
HTML
```

## RuboCop version

```
$ bin/bundle exec rubocop -V
1.67.0 (using Parser 3.3.5.0, rubocop-ast 1.32.3, analyzing as Ruby 2.7, running on jruby 3.1.4) [java]
  - rubocop-performance 1.22.1
```

dvandersluis:
If a method chain starting with a heredoc goes over the line length limit, breaking it onto multiple lines is a syntax error. This prevents the autocorrection that would do so.

```ruby
str = <<~HEREDOC.do_something.with_args(foo: '', bar: '', baz: '')
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^ Line is too long. [66/40]
  text
HEREDOC
```

Fixes #13343.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/check_line_breakable.rb
+++ b/lib/rubocop/cop/mixin/check_line_breakable.rb
@@ -44,6 +44,8 @@ module RuboCop
     module CheckLineBreakable
       def extract_breakable_node(node, max)
         if node.send_type?
+          return if chained_to_heredoc?(node)
+
           args = process_args(node.arguments)
           return extract_breakable_node_from_elements(node, args, max)
         elsif node.def_type?
@@ -221,6 +223,14 @@ module RuboCop
         return node.first_line != node.last_argument.last_line if node.def_type?
 
         !node.single_line?
+      end
+
+      def chained_to_heredoc?(node)
+        while (node = node.receiver)
+          return true if (node.str_type? || node.dstr_type? || node.xstr_type?) && node.heredoc?
+        end
+
+        false
       end
     end
   end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/mixin/check_line_breakable.rb
```diff
@@ -222,6 +224,22 @@ def already_on_multiple_lines?(node)
...
+        while receiver
+          if (receiver.str_type? || receiver.dstr_type? || receiver.xstr_type?) && receiver.heredoc?
+            return true
+          end
+
+          receiver = receiver.receiver
+        end
+
+        false
+      end
```
koic, line 242: Can you refactor to the following?

```suggestion
      def chained_to_heredoc?(node)
        while (node = node.receiver)
          if (node.str_type? || node.dstr_type? || node.xstr_type?) && node.heredoc?
            return true
          end
        end

        false
      end
```

dvandersluis: Sure, updated.

koic:
Thanks!

PR #13346, opened 2024-10-16
[Fix #13324] Fix `--disable-uncorrectable` to not insert a comment inside a string continuation

### LINKED ISSUE #13324: --disable-uncorrectable creates syntax errors when violations occur within multi-line strings
`--disable-uncorrectable` should not attempt to inject a line-level `rubocop:todo` comment in the middle of a multi-line string.  This results in syntax errors.

This was filed in issue https://github.com/rubocop/rubocop/issues/7216 but root cause was not addressed in that issue.  The example cop was just modified to detect fewer violations to work around it.  The general issue of `--disable-uncorrectable` not working correctly with multi-line strings is reproducible with any non-autocorrectable cop that can detect a violation on a line that ends with a string.

Example:

```
# This code has a violation of Lint/DuplicateHashKey

          foo = {
            key1: "something",
            key1: "whatever"\
              "something else",
          }
```

--------

## Expected behavior

`--disable-uncorrectable` should wrap the multi-line string like this:

```
          foo = {
            key1: "something",
            # rubocop:todo Lint/DuplicateHashKey
            key1: "whatever"\
              "something else",
            # rubocop:enable Lint/DuplicateHashKey
          }
```

## Actual behavior

Instead it creates a syntax error

```
          foo = {
            key1: "something",
            key1: "whatever"\ # rubocop:todo Lint/DuplicateHashKey
              "something else",
          }
```

Create any code sample that has a non-autocorrectable violation of any cop on a line that terminates part-way through a multi-line string.

Run rubocop with `-A --disable-uncorrectable`

## RuboCop version

```
bundle exec rubocop -V
1.66.1 (using Parser 3.3.5.0, rubocop-ast 1.32.3, running on ruby 3.1.3) +server [arm64-darwin21]
  - rubocop-rails 2.10.1
  - rubocop-rspec 2.12.1
  - rubocop-sorbet 0.7.4
  - rubocop-thread_safety 0.5.1

```

dvandersluis:
With the following code and `--disable-uncorrectable` specified:

```ruby
{
  key1: "something",
  key1: "whatever"\
    "something else",
}
```

Previously RuboCop corrected to a syntax error:

```ruby
{
  key1: "something",
  key1: "whatever"\ # rubocop:todo Lint/DuplicateHashKey
    "something else",
}
```

With this correction, RuboCop uses the block form of `todo` instead:

```ruby
{
  key1: 'something',
  # rubocop:todo Lint/DuplicateHashKey
  key1: 'whatever'\
    'something else'
  # rubocop:enable Lint/DuplicateHashKey
}
```

Fixes #13324.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/autocorrect_logic.rb
+++ b/lib/rubocop/cop/autocorrect_logic.rb
@@ -49,7 +49,9 @@ module RuboCop
       private
 
       def disable_offense(offense_range)
-        range = surrounding_heredoc(offense_range) || surrounding_percent_array(offense_range)
+        range = surrounding_heredoc(offense_range) ||
+                surrounding_percent_array(offense_range) ||
+                string_continuation(offense_range)
 
         if range
           disable_offense_before_and_after(range_by_lines(range))
@@ -88,8 +90,26 @@ module RuboCop
         end
 
         percent_array.map(&:source_range).find do |range|
-          offense_range.begin_pos > range.begin_pos && range.overlaps?(offense_range)
+          range_overlaps_offense?(offense_range, range)
         end
+      end
+
+      def string_continuation(offense_range)
+        return nil if offense_range.empty?
+
+        string_continuation_nodes = processed_source.ast.each_descendant.filter_map do |node|
+          range_by_lines(node.source_range) if string_continuation?(node)
+        end
+
+        string_continuation_nodes.find { |range| range_overlaps_offense?(offense_range, range) }
+      end
+
+      def range_overlaps_offense?(offense_range, range)
+        offense_range.begin_pos > range.begin_pos && range.overlaps?(offense_range)
+      end
+
+      def string_continuation?(node)
+        (node.str_type? || node.dstr_type? || node.xstr_type?) && node.source.match?(/\\\s*$/)
       end
 
       def range_of_first_line(range)
--- a/changelog/fix_fix_disable_uncorrectable_to_not_insert_a_comment.md
+++ b/changelog/fix_fix_disable_uncorrectable_to_not_insert_a_comment.md
@@ -1,0 +1,1 @@
+* [#13324](https://github.com/rubocop/rubocop/issues/13324): Fix `--disable-uncorrectable` to not insert a comment inside a string continuation. ([@dvandersluis][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/autocorrect_logic.rb
```diff
@@ -88,10 +90,28 @@ def surrounding_percent_array(offense_range)
...
+
+        string_continuation_nodes = processed_source.ast.each_descendant.filter_map do |node|
+          range_by_lines(node.source_range) if string_continuation?(node)
+        end
+
+        string_continuation_nodes.find { |range| range_overlaps_offense?(offense_range, range) }
+      end
+
+      def range_overlaps_offense?(offense_range, range)
+        offense_range.begin_pos > range.begin_pos && range.overlaps?(offense_range)
```
koic, line 108: It seems the test passes even without `range.overlaps?(offense_range)`. Can you review the logic or the test?

lib/rubocop/cop/autocorrect_logic.rb
```diff
@@ -88,10 +90,28 @@ def surrounding_percent_array(offense_range)
...
+      end
+
+      def string_continuation?(node)
+        (node.str_type? || node.dstr_type? || node.xstr_type?) && node.source =~ /\\\s*$/
```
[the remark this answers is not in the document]
dvandersluis: Updated, thanks.

lib/rubocop/cop/autocorrect_logic.rb
dvandersluis: It is necessary. I added a second test that would fail without it:

```ruby
context 'and the offense is on a different line than a string continuation' do
  it 'adds a single one-line disable statement' do
    create_file('.rubocop.yml', <<~YAML)
      AllCops:
        DisabledByDefault: true
      Lint/DuplicateHashKey:
        Enabled: true
    YAML
    create_file('example.rb', <<~'RUBY')
      x = 'something'\
        'something else'
      {
        key1: 'foo',
        key1: 'bar'
      }
    RUBY
    expect(exit_code).to eq(0)
    expect(File.read('example.rb')).to eq(<<~'RUBY')
      x = 'something'\
        'something else'
      {
        key1: 'foo',
        key1: 'bar' # rubocop:todo Lint/DuplicateHashKey
      }
    RUBY
  end
end
```

PR #13347, opened 2024-10-16
Show Ruby version of the current dir for `rubocop -V`

Earlopain:
The rules are complex, even within the same project different ruby versions can be used for analysis. This makes `rubocop -V` output the ruby version of the current working directory.

The first commit is a bit of cleanup. It was more difficult than necessary to understand where the `version` method is being used and how. Most callers simply want `RuboCop::Version::STRING`, i've updated these places. I wanted to just remove the `debug` option since it is marked private but standardrb uses it and other places use `RuboCop::Version.version` as well anyways, regardless of intended visibility. Instead I've left a note, so that _if_ it does get changed in the future, it can be more coordinated.

-----------------

* [x] Added tests.


bbatsov:
> even within the same project different ruby versions can be used for analysis.

It never crossed my mind someone might want to do something like this, but I guess in makes sense in some mono-repo setups when you have multiple projects.

Earlopain:
Yeah, monorepos are a possible use-case. Rails uses `rubocop-md` and analyses code with a more recent version that what they themselves support so they can use more modern syntax in docs.

PR #13349, opened 2024-10-16
[Fix #13348] Ensure `Style/BlockDelimiters` autocorrection does not move other code between the block and comment

### LINKED ISSUE #13348: `Style/BlockDelimiters` inserts a misplaced newline when moving a trailing comment that follows non-block code
With 

```ruby
my_method { |x|
            x.foo } unless bar # comment
```

`Style/BlockDelimiters` autocorrects to:

```ruby
# comment
my_method do |x|
  x.foo
end
 unless foo
```

Which is a syntax error (`unless` is no longer attached to the previous statement).

References https://github.com/rubocop/rubocop/issues/13096#issuecomment-2417493491

## RuboCop version

  - rubocop-performance 1.22.1
  - rubocop-rake 0.6.0
  - rubocop-rspec 3.1.0

dvandersluis:
When `Style/BlockDelimiters` autocorrects `{...}` to `do...end` with a trailing comment, it previously would insert a newline after the end of the `block` range, but in the cases where additional code follows on that line, it could change semantics or even cause a syntax error. This change ensures that that code is retained on the same line as the end of the block.

Fixes #13348.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/block_delimiters.rb
+++ b/lib/rubocop/cop/style/block_delimiters.rb
@@ -303,11 +303,26 @@ module RuboCop
 
         def move_comment_before_block(corrector, comment, block_node, closing_brace)
           range = block_node.chained? ? end_of_chain(block_node.parent).source_range : closing_brace
+
+          # It is possible that there is code between the block and the comment
+          # which needs to be preserved and trimmed.
+          pre_comment_range = source_range_before_comment(range, comment)
+
           corrector.remove(range_with_surrounding_space(comment.source_range, side: :right))
-          remove_trailing_whitespace(corrector, range, comment)
-          corrector.insert_after(range, "\n")
+          remove_trailing_whitespace(corrector, pre_comment_range, comment)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/block_delimiters.rb
```diff
@@ -303,13 +303,28 @@ def whitespace_after?(range, length = 1)
 
         def move_comment_before_block(corrector, comment, block_node, closing_brace)
           range = block_node.chained? ? end_of_chain(block_node.parent).source_range : closing_brace
+
+          # it is possible that there is code between the block and the comment
+          # which needs to be preserved and trimmed
```
koic, line 308: That's just my two cents. Can you tweak it?

```suggestion
          # It is possible that there is code between the block and the comment
          # which needs to be preserved and trimmed.
```

dvandersluis: Sure happy to. Updated!

koic:
Good catch! Thanks!

PR #13350, opened 2024-10-16
Consistently spell "Ruby" in `rubocop -V`

Earlopain:
This is incredibly minor but since the running ruby version was added they are now cased in two different ways.
There doesn't seem to be a constant that contains the proper spelling, so I just added something for lookup.

-----------------

* [ ] Added tests.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -38,7 +38,7 @@ output by `rubocop -V`, include them as well. Here's an example:
 
 ```
 $ [bundle exec] rubocop -V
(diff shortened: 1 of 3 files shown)


koic:
Hm, the point is understood, but it cannot be definitively stated that editing the value is of significant value. There's also the perspective that "Ruby" refers to the specification, while "ruby" refers to the implementation. To respect the value of `RUBY_ENGINE`, so I'll close this PR. Thank you.

Earlopain:
Surely it should either be `ruby` or `Ruby` in both cases, why should analysing and running be capitalized differently? I don't see why it would make a difference where that string was aquired from. How about just downcasing the other "Ruby" then?

koic:
The analyzing follows the Ruby specification. The runtime follows the Ruby implementation. These are different things.

PR #13352, opened 2024-10-16
[Fix #13351] Add Lint/NoRescueErrorMessageChecking rubocop to discourage rescue error matching

### LINKED ISSUE #13351: New rubocop to discourage rescue error matching
We have been discussing internally in my company that rescue error matching is antipattern and should be discourage. I dont believe we can autocorrect but it would be useful to point it out. I have already written it for our internal repo and was encouraged to port it over to rubocop core.

```
# bad
begin
  something
rescue => e
  if e.message.match?(/Duplicate entry/)
    handle_error
  end
end

# bad
begin
  something
rescue => e
  unless e.message.match?(/Duplicate entry/)
    handle_error
  end
end

# good
begin
  something
rescue ActiveRecord::RecordNotUnique => e
  handle_error
end
```

simulk:
fixes https://github.com/rubocop/rubocop/issues/13351

We have been discussing internally in my company that rescue error matching is antipattern and should be discourage. I dont believe we can autocorrect but it would be useful to point it out. I have already written it for our internal repo and was encouraged to port it over to rubocop core. Proposing a new `Lint/NoRescueErrorMessageChecking` cop

```
# bad
begin
  something
rescue => e
  if e.message.match?(/Duplicate entry/)
    handle_error
  end
end

# bad
begin
  something
rescue => e
  unless e.message.match?(/Duplicate entry/)
    handle_error
  end
end

# good
begin
  something
rescue ActiveRecord::RecordNotUnique => e
  handle_error
end
```

-----------------

* [ ] Added tests.


dvandersluis:
Hi @simulk thanks for your contribution! I'm not sure if this makes sense as a built in cop or a custom one as you've done for your codebase. Opening a PR in rubocop/ruby-style-guide to collect feedback would likely be a good first step.

An issue I see with this in general is handling exceptions coming from external libraries or gems. While I agree that it would be preferable to define specific exception classes for different use cases rather than switching on message text, when the exceptions aren't under your control you have to do what you have to do.

That being said, if this was to be merged, it'd make sense at least to match code in the form `/Duplicate entry/.match(e.message)` to complement the inverse that you've included.

lib/rubocop/cop/lint/no_rescue_error_message_checking.rb
```diff
@@ -0,0 +1,58 @@
...
+        def on_rescue(node)
+          node.each_descendant(:if, :unless).each do |condition_node|
+            add_offense(condition_node) if message_check?(condition_node)
+          end
+        end
+
+        private
+
+        def message_check?(condition_node)
+          condition_node.condition&.type == :send &&
```
dvandersluis, line 50: It's preferable to use the built in node type matchers instead: 
```suggestion
          condition_node.condition.send_type? &&
```

Earlopain:
Yes, I do agree that there are situations where you can't help yourself and have to check the exception message. Other gems are an obvious source. Also external system like an api or postgres for example which have nothing to do with rails itself where you may just get the string and that's it.

Currently the cop is a bit too eager: 

```rb
begin
rescue => e
  raise if unrelated_var.match?(whatever)
end
```

```rb
begin
  foo
rescue NameError => e
  raise if e.name.match?(/(?:\A|::)#{name}\z/)
end
```

These should not be an offense I think.

simulk:
I opened issue https://github.com/rubocop/ruby-style-guide/issues/951 in ruby style guide. based on the feedback, if its deemed to be an useful built in cop, I can update PR based on suggestions otherwise close it.

pirj:
Would you consider the “bad” code just as bad if the conditional statement included an `else` branch?

What is considered bad here?
Using the message text as a distinction for handling flow, which is brittle?
Or swallowing exceptions implicitly, potentially inadvertently?

ngan:
Just wanted to add some more context (I work with @simulk, at Gusto). This cop was extracted from our multi-million line Rails monolith. From what we've seen internally, pretty much all of our existing violations were from cases where rescuing a more specific error makes more sense. For example, people were rescuing `ActiveRecord::ActiveRecordError` and checking the message for uniqueness error instead of rescuing `ActiveRecord::RecordNotUnique`. If there's an exception to this rule, we see it as an opportunity to fix/improve. If we're striving for "good," IMHO messages should be for human consumption, not machine.

lib/rubocop/cop/lint/no_rescue_error_message_checking.rb
```diff
@@ -0,0 +1,58 @@
...
+      #     handle_error
+      #   end
+      #
+      class NoRescueErrorMessageChecking < Base
+        MSG = 'Avoid checking error message while handling exceptions. ' \
+              'This is brittle and can break easily.'
+        METHODS_TO_CHECK = %i[match? include?].freeze
+
+        def on_rescue(node)
+          node.each_descendant(:if, :unless).each do |condition_node|
```
pirj, line 42: Case, ternary

lib/rubocop/cop/lint/no_rescue_error_message_checking.rb
```diff
@@ -0,0 +1,58 @@
...
+            add_offense(condition_node) if message_check?(condition_node)
```
pirj, line 43: What if the statement is separate?
```
non_unique = e.message == "Uniq error"
if non_unique
  # handle
```

spec/rubocop/cop/lint/no_rescue_error_message_checking_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      end
+    RUBY
+  end
+
+  it 'does not register an offense for other conditions' do
+    expect_no_offenses(<<~RUBY)
+      begin
+        # some code
+      rescue => e
+        if e.class == StandardError
```
pirj, line 81: Maybe use some other one as an example here? It is confusing because e is a standard exception, but potentially is an instance of a subclass.

pirj:
While obviously you’re scratching your own itch, can there be legitimate cases of such usage in the wild?
With this question I’m trying to aboid future frustration over RuboCop enforcing some style on the whole community without checking in advance what community prefers, this enforcement to become a new practice in some places, and be silenced in others, thus resulting in further style difference rather than consistency in the community overall.

simulk:
@pirj @dvandersluis apologies for late reply - I got caught up in few things at work.

I'm happy to make changes as suggested but want to confirm if we want to proceed with this cop. As mentioned, this is an outcome of our own frustration, but if its not that useful for broader community, I am happy to close the PR as well.

dvandersluis:
As mentioned earlier, I have unfortunately experienced many instances at work where a library we were using used a singular expection with different messages for different situations. Short of replacing the library with our own abstraction layer (which introduces coupling and brittleness), there is no way to deal with this situation without comparing strings.

> For example, people were rescuing `ActiveRecord::ActiveRecordError` and checking the message for uniqueness error instead of rescuing `ActiveRecord::RecordNotUnique`

I think this is the key point for me, and indicates that perhaps this cop is going the wrong direction. I'd suggest that rather than preventing checking strings, it'd make more sense to have a cop that looks for specific superclass exception classes and suggests using a more specific exception instead. I could even see a cop specifically for `rescue ActiveRecord::ActiveRecordError` fitting in rubocop-rails.

I ran it against real-world-ruby-apps and real-world-rails and while there wasn't a *ton* of offenses, it found 41 in the former and 165 in the latter.

In any case, thanks for your contribution but I don't think this is a good fit for rubocop.

dvandersluis:
BTW, if you are running your cop, you should know that there's a false positive (this test fails):

```ruby
it 'does not register an offense outside of a `rescue`' do
  expect_no_offenses(<<~'RUBY')
    begin
      foo if e.message.match?(/pattern/)
    rescue
      bar  
    end
  RUBY
end
```

You could fix it by replacing `on_rescue` with `on_resbody`.

corsonknowles:
Thank you @dvandersluis -- one of our engineers caught and patched that case by replacing `node.each_descendant` with `node.resbody_branches.last.each_descendant`

I will pass along that there's a better hook to use to speed that up.

PR #13366, opened 2024-10-21
[Fix #13365] Fix false positives for `Lint/SafeNavigationConsistency`

### LINKED ISSUE #13365: false positive `Lint/SafeNavigationConsistency`
## Expected behavior

`Lint/SafeNavigationConsistency` should not occur.

## Actual behavior

`Lint/SafeNavigationConsistency` occurs.

```
user&.freeze && user == current_user
```
https://gist.github.com/nna774/ca3957f172424ac8a1a7e1b32ea48373

```
test.rb:3:17: W: Lint/SafeNavigationConsistency: Use . instead of unnecessary &..
user&.freeze && user == current_user
                ^^^^^^^^^^^^^^^^^^^^

2 files inspected, 1 offense detected
```

Similar to the second pattern( `foo&.bar && foo.baz` ), I think this should be considered good.
https://docs.rubocop.org/rubocop/cops_lint.html#examples-lintsafenavigationconsistency

## RuboCop version

```
$ bundle exec rubocop -V
1.67.0 (using Parser 3.3.5.0, rubocop-ast 1.32.3, analyzing as Ruby 2.7, running on ruby 3.3.3) [arm64-darwin23]
```

koic:
Fixes #13365.

This PR fixes false positives for `Lint/SafeNavigationConsistency` when using safe navigation on the LHS with operator method on the RHS of `&&`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_false_positives_for_lint_safe_navigation_consistency.md
+++ b/changelog/fix_false_positives_for_lint_safe_navigation_consistency.md
@@ -1,0 +1,1 @@
+* [#13365](https://github.com/rubocop/rubocop/issues/13365): Fix false positives for `Lint/SafeNavigationConsistency` when using safe navigation on the LHS with operator method on the RHS of `&&`. ([@koic][])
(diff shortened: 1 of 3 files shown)


nna774:
Thanks for fixing https://github.com/rubocop/rubocop/issues/13365.

But, like below code
```
str = ''
str&.to_s && str + 'hello'
```
produces SafeNavigationConsistency too.
So I think this is not sufficient(but I don't know what to do...).

Thanks.

koic:
Ah, good catch! I've updated this PR.

PR #13370, opened 2024-10-21
[Fix #13364] Fix incorrect autocorrect with `Lint/UselessAssignment` a multiple assignment or `for` contains an inner assignment

### LINKED ISSUE #13364: `Lint/UselessAssignment` autofix completely trashes an expression
## Expected behavior

Describe here how you expected RuboCop to behave in this particular situation.

```ruby
def func(a, b, c = 0)
  [a, b, c]
end

a, = func(
  1,
  2,
  3,
)

a
```

## Actual behavior

```console
$ ./rubocop --except Naming/MethodParameterName -a t.rb 
/home/stdedos/Documents/RelexOy/planning-cloud/backend/bin/bundle:7: warning: unused literal ignored
t.rb:1:1: C: [Correctable] Style/FrozenStringLiteralComment: Missing frozen string literal comment.
def func(a, b, c = 0)
^
t.rb:8:3: W: [Corrected] Lint/UselessAssignment: Useless assignment to variable - c. Use _ or _c as a variable name to indicate that it won't be used.
  c = 3,
  ^
 1/1 file |======================= 100 =======================>| Time: 00:00:01 

1 file inspected, 2 offenses detected, 1 offense corrected, 1 more offense can be corrected with `rubocop -A`


You can opt out of this message by adding the following to your config (see https://docs.rubocop.org/rubocop/extensions.html#extension-suggestions for more options):
  AllCops:
    SuggestExtensions: false
$ cat t.rb 
def func(a, b, c = 0)
  [a, b, c]
end

a, = func(
  1,
  2,
  _,
)

a
```

## RuboCop version

```
$ [bundle exec] rubocop -V
  - rubocop-performance 1.22.1
  - rubocop-rspec 3.1.0
```

dvandersluis:
With code such as:

```ruby
a, b = func(c = 3)
[a, b]
```

`Lint/UselessAssignment` was previously considering the inner assignment (`c = 3`) to be part of the multiple assignment for autocorrection. This changes it to not consider assignment inside a method argument to be part of the containing `meta_assignment_node` for `masgn` and `for` nodes. 

Fixes #13364.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/variable_force/assignment.rb
+++ b/lib/rubocop/cop/variable_force/assignment.rb
@@ -102,6 +102,7 @@ module RuboCop
         end
 
         def multiple_assignment_node
+          return nil unless node.parent&.mlhs_type?
           return nil unless (grandparent_node = node.parent&.parent)
           if (node = find_multiple_assignment_node(grandparent_node))
             return node
@@ -119,7 +120,13 @@ module RuboCop
         end
 
         def for_assignment_node
-          node.ancestors.find(&:for_type?)
+          return unless (parent_node = node.parent)
+          return parent_node if parent_node.for_type?
+
+          grandparent_node = parent_node.parent
+          return grandparent_node if parent_node.mlhs_type? && grandparent_node&.for_type?
+
+          nil
         end
 
         def find_multiple_assignment_node(grandparent_node)
--- a/spec/rubocop/cop/variable_force/assignment_spec.rb
+++ b/spec/rubocop/cop/variable_force/assignment_spec.rb
@@ -144,6 +144,110 @@ RSpec.describe RuboCop::Cop::VariableForce::Assignment do
         expect(assignment.meta_assignment_node.type).to eq(:for)
       end
     end
+
+    context 'when it is an argument of a send node' do
+      # The `lvasgn` in question in this `context` is the one inside a call to `my_method`.
+      let(:lvasgn_node) { ast.each_node.to_a.reverse.find(&:lvasgn_type?) }
+
+      context 'when it is += operator assignment' do
+        let(:source) do
+          <<~RUBY
+            def some_method
+              foo += my_method(bar = 1)
+            end
+          RUBY
+        end
+
+        it 'returns nil' do
+          expect(assignment.meta_assignment_node).to be_nil
+        end
+      end
+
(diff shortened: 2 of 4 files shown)


lib/rubocop/cop/variable_force/assignment.rb
```diff
@@ -119,7 +120,15 @@ def rest_assignment_node
...
-          node.ancestors.find(&:for_type?)
+          parent_node = node.parent
+          return unless parent_node
+
+          grandparent_node = parent_node.parent
+
+          return parent_node if parent_node.for_type?
+          return grandparent_node if parent_node.mlhs_type? && grandparent_node&.for_type?
+
+          nil
```
koic, line 131: Can you tweak to the following? There is a possibility for an early return before searching for `grandparent_node`.

```suggestion
          return unless (parent_node = node.parent)
          return parent_node if parent_node.for_type?

          grandparent_node = parent_node.parent
          return grandparent_node if parent_node.mlhs_type? && grandparent_node&.for_type?

          nil
```

spec/rubocop/cop/variable_force/assignment_spec.rb
```diff
@@ -144,6 +144,110 @@ def some_method
         expect(assignment.meta_assignment_node.type).to eq(:for)
       end
     end
+
+    context 'when it is an argument of a send node' do
+      # The `lvasgn` in question in this `context` is the one inside a call to `my_method`
+      let(:lvasgn_node) { ast.each_node.select(&:lvasgn_type?).last } # rubocop:disable Performance/Detect
```
koic, line 150:
```suggestion
      # The `lvasgn` in question in this `context` is the one inside a call to `my_method`.
      let(:lvasgn_node) { ast.each_node.to_a.reverse.find(&:lvasgn_type?) }
```

lib/rubocop/cop/variable_force/assignment.rb
dvandersluis: Updated.

spec/rubocop/cop/variable_force/assignment_spec.rb
dvandersluis: Updated.

koic:
Thanks!

PR #13371, opened 2024-10-21
Fix a false positive for `Style/ZeroLengthPredicate` when using an Enumerator

punkstar:
We already ignore some stdlib classes like `File`, `Tempfile` and `StringIO` in the `#empty?` cop. This pull request adds [`Enumerator`](https://ruby-doc.org/core-2.6/Enumerator.html) to that list.

-----------------

* [x] Added tests.


koic:
Unlike `File`, `Tempfile`, and `StringIO`, `Enumerator` does not have an `open` method, so it cannot be resolved by simply adding it to a list. And the test cases `Enumerator.new(-> {0}).size` raises an `ArgumentError` and cannot be executed. As far as I know, the size of `Enumerator.new` returns nil unless explicitly implemented, which leads me to wonder whether `Enumerator.new` should be included here. So, I will close this proposal. Thank you.

PR #13373, opened 2024-10-22
[Fix #13372] Fix inconsistent cache directory when using --cache-root

### LINKED ISSUE #13372: `ResultCache.cleanup` inconsistency when using `--cache-root` vs ENV or config file
When using caching and the CLI flag `--cache-root`, with for example `tmp`, Rubocop attempts to clean up files within `tmp`, not just files within the expected subdirectory of `tmp` (`rubocop_cache`)

When using an environment variable or the YAML config directive to specify the cache root, the tool works as expected.

I believe this happens because `ResultCache.cleanup` is called [here](https://github.com/rubocop/rubocop/blob/8ac02bd715537e20422c3bb12586def0c40d822c/lib/rubocop/runner.rb#L130) with the raw `cache_root` option, and [that method implementation](https://github.com/rubocop/rubocop/blob/8ac02bd715537e20422c3bb12586def0c40d822c/lib/rubocop/result_cache.rb#L28) does not append `rubocop_cache`. 

This bug is similar to https://github.com/rubocop/rubocop/issues/10766 but limited to the cache cleanup portion.

--------

## Expected behavior

Using `--cache-root` would behave exactly like setting `RUBOCOP_CACHE_ROOT` or `AllCops: CacheRootDirectory` when cached files are cleaned up.  

## Actual behavior

The exact directory specified with`--cache-root flag` has its contents cleaned up, not the `rubocop_cache` subdirectory that other code uses.

- Set `MaxFilesInCache` config to a low number, 10
- Create a directory `tmp` and touch 10+ files in it with e.g. `touch tmp/01` etc
- Create a `lib` directory with several .rb files in it
- Run `rubocop --cache=true --cache-root=tmp`
- Observe that some of the files within `tmp` have been removed

## RuboCop version

```
danb@danb-macbook-air cleanup_bug_rubocop % rubocop -V
  - rubocop-performance 1.22.1
  - rubocop-rake 0.6.0
  - rubocop-rspec 3.1.0
```

Capncavedan:
Fixes #13372.

This PR fixes inconsistent cache directory calculation during cache pruning when the --cache-root option is used instead of ENV or YAML configuration.

<hr>

* [x]  The PR relates to _only_ one subject with a clear title and description in grammatically correct, complete sentences.
* [x]  Wrote [good commit messages](https://chris.beams.io/posts/git-commit/).
* [x]  Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x]  Feature branch is up-to-date with `master` (if not - rebase it).
* [x]  Squashed related commits together.
* [x]  Added tests.
* [x]  Ran `bundle exec rake default`. It executes all tests and runs RuboCop on its own code.
* [x]  Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

### DIFF
--- a/changelog/fix_cache_cleanup_path_when_cache_root_specified_with_cli_option.md
+++ b/changelog/fix_cache_cleanup_path_when_cache_root_specified_with_cli_option.md
@@ -1,0 +1,1 @@
+* [#13372](https://github.com/rubocop/rubocop/issues/13372): Add `rubocop_cache` to the path given by `--cache-root` when pruning cache ([@capncavedan][])
--- a/lib/rubocop/result_cache.rb
+++ b/lib/rubocop/result_cache.rb
@@ -28,6 +28,7 @@ module RuboCop
     def self.cleanup(config_store, verbose, cache_root = nil)
       return if inhibit_cleanup # OPTIMIZE: For faster testing
 
+      cache_root = File.join(cache_root, 'rubocop_cache') if cache_root
       cache_root ||= cache_root(config_store)
       return unless File.exist?(cache_root)
 
--- a/spec/rubocop/result_cache_spec.rb
+++ b/spec/rubocop/result_cache_spec.rb
@@ -15,6 +15,7 @@ RSpec.describe RuboCop::ResultCache, :isolated_environment do
   let(:options) { {} }
   let(:config_store) { instance_double(RuboCop::ConfigStore, for_pwd: RuboCop::Config.new) }
   let(:cache_root) { "#{Dir.pwd}/rubocop_cache" }
+  let(:cache_root_as_option) { Dir.pwd }
   let(:offenses) do
     [RuboCop::Cop::Offense.new(:warning, location, 'unused var', 'Lint/UselessAssignment')]
   end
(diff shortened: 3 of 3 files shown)


Earlopain:
Thanks for the fix, this makes sense to me. Since this logic is duplicated in the intialize method here https://github.com/rubocop/rubocop/blob/f3aa4fb7c3e3d3d91746fad37a233748aff0394e/lib/rubocop/result_cache.rb#L87-L89 could you move that to a common method? I'd just modify `cache_root(config_store)` to take an optional `cache_root` and return the appropriate result.

Capncavedan:
I think that does make sense @Earlopain.  Happy to do it.

What would you think of then also removing the optional `cache_root` argument on the  `ResultCache.cleanup` method?

Earlopain:
> What would you think of then also removing the optional cache_root argument

Can you explain what you mean with that? You need that argument, no?

Capncavedan:
@Earlopain nm, I see what you were driving at now - moving this logic 

```ruby
 cache_root ||= File.join(options[:cache_root], 'rubocop_cache') if options[:cache_root]
```

into the `ResultCache.cache_root` method

Capncavedan:
Closing in favor of https://github.com/rubocop/rubocop/pull/13405

PR #13380, opened 2024-10-23
[Fix #13378] Fix false positives for `Style/TernaryParentheses`

### LINKED ISSUE #13378: Style/TernaryParentheses reported incorrectly
cop incorrectly reports when using `&.` and comparators.

[Before `rubocop -a`]

code:

```ruby
value = 100
puts (value&.> 10) ? 1000 : 2000

value = nil
puts (value&.> 10) ? 1000 : 2000
```

result:

```
1000
2000
```

[After `rubocop -a`]

code:

```ruby
value = 100
puts value&.> 10 ? 1000 : 2000

value = nil
puts value&.> 10 ? 1000 : 2000
```

result:

```
false

```

--------

## Expected behavior

cop should not report Style/TernaryParentheses. Additionally, do not auto-correct.

## Actual behavior

```
C: [Correctable] Style/TernaryParentheses: Omit parentheses for ternary conditions.
puts (value&.> 10) ? 1000 : 2000
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

execute `rubocop -a`

## RuboCop version

```
```

koic:
Fixes #13378.

This PR fixes false positives for `Style/TernaryParentheses` when a ternary condition is a parenthesized safe navigation operator method and its argument is also parenthesized.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/ternary_parentheses.rb
+++ b/lib/rubocop/cop/style/ternary_parentheses.rb
@@ -70,7 +70,7 @@ module RuboCop
           condition = node.condition
 
           return if only_closing_parenthesis_is_last_line?(condition)
-          return if condition_as_parenthesized_one_line_pattern_matching?(condition)
+          return if condition_as_parenthesized_operator_or_one_line_pattern_matching?(condition)
           return unless node.ternary? && offense?(node)
 
           message = message(node)
@@ -86,15 +86,20 @@ module RuboCop
           condition.source.split("\n").last == ')'
         end
 
-        def condition_as_parenthesized_one_line_pattern_matching?(condition)
+        def condition_as_parenthesized_operator_or_one_line_pattern_matching?(condition)
           return false unless condition.parenthesized_call?
           return false unless (first_child = condition.children.first)
+          return true if parenthesized_operator_method_call?(first_child)
 
           if target_ruby_version >= 3.0
             first_child.match_pattern_p_type?
           else
             first_child.match_pattern_type? # For Ruby 2.7's one line pattern matching AST.
           end
+        end
+
+        def parenthesized_operator_method_call?(node)
+          node.call_type? && node.loc.dot && node.operator_method? && !node.parenthesized?
         end
 
         def autocorrect(corrector, node)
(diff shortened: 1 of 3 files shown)


dvandersluis:
@koic I opened #13381 before I realized you had also tackled this. I took a different approach, of adding parentheses around arguments when necessary. I also discovered more tangential cases which my PR handles.

lib/rubocop/cop/style/ternary_parentheses.rb
```diff
@@ -86,10 +88,14 @@ def only_closing_parenthesis_is_last_line?(condition)
...
         end
 
-        def condition_as_parenthesized_one_line_pattern_matching?(condition)
+        def condition_as_parenthesized_csend_operator_or_one_line_pattern_matching?(condition)
           return false unless condition.parenthesized_call?
           return false unless (first_child = condition.children.first)
 
+          if first_child.csend_type? && first_child.operator_method? && !first_child.parenthesized?
+            return true
+          end
```
dvandersluis, line 97: The same thing would happen without safe nav btw:

```ruby
# test.rb
(foo.> bar) ? 1000 : 2000

$ rubocop -A test.rb --only Style/TernaryParentheses
Inspecting 1 file
C

Offenses:

test.rb:1:1: C: [Corrected] Style/TernaryParentheses: Omit parentheses for ternary conditions.
(foo.> bar) ? 1000 : 2000
^^^^^^^^^^^^^^^^^^^^^^^^^

# test.rb becomes
foo.> bar ? 1000 : 2000
```

koic:
I'm sorry, but I believe overusing parentheses reduces readability, so I will go with the solution in this PR. Adding parentheses to change precedence should be a last resort. I think the code should rely as much as possible on the precedence rules provided by the syntax.

dvandersluis:
I don't follow how adding parentheses around the arguments changes precedence in this case. The parentheses around the entire send node encapsulates the arguments, and removing those would change precedence.

In any case, having a special case for operators used with a dot doesn't align with the numerous other test where an offense is raised for unparenthesized method calls (with or without corrections), and isn't consistent when the requested style is to remove parentheses around ternary conditions.

lib/rubocop/cop/style/ternary_parentheses.rb
koic: Good catch! I've updated this PR.

koic:
Ah, I see. I understand now. Let's go with #13381. I will close this PR, reopen #13381, and start the review. Thanks for the explanation!

dvandersluis:
Thanks @koic. Sorry if I wasn't explaining myself well enough!

PR #13381, opened 2024-10-23
[Fix #13378] When removing parens in `Style/TernaryParentheses` with a `send` node condition, ensure its arguments are parenthesized

### LINKED ISSUE #13378: Style/TernaryParentheses reported incorrectly

dvandersluis:
When the condition to a ternary is a send node with arguments, and `Style/TernaryParentheses` removes parentheses around that condition, ensure the arguments are themselves parenthesized, because otherwise removing the surrounding parens will change the semantics of the ternary.

This also allows a number of test cases to be autocorrected which previously did not have autocorrect.

Fixes #13378.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/ternary_parentheses.rb
+++ b/lib/rubocop/cop/style/ternary_parentheses.rb
@@ -171,9 +171,7 @@ module RuboCop
         end
 
         def unsafe_autocorrect?(condition)
-          condition.children.any? do |child|
-            unparenthesized_method_call?(child) || below_ternary_precedence?(child)
-          end
+          condition.children.any? { |child| below_ternary_precedence?(child) }
         end
 
         def unparenthesized_method_call?(child)
@@ -192,7 +190,7 @@ module RuboCop
         # @!method method_name(node)
         def_node_matcher :method_name, <<~PATTERN
           {($:defined? _ ...)
-           (send {_ nil?} $_ _ ...)}
+           (call {_ nil?} $_ _ ...)}
         PATTERN
 
         def correct_parenthesized(corrector, condition)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/ternary_parentheses.rb
```diff
@@ -203,16 +201,46 @@ def correct_parenthesized(corrector, condition)
...
+        end
+
+        def send_node_args_need_parens?(send_node)
+          return false unless send_node.call_type? || send_node.defined_type?
+
+          send_node.arguments.any? &&
+            !send_node.parenthesized? &&
+            (send_node.dot? ||
+              send_node.safe_navigation? ||
+              unparenthesized_method_call?(send_node))
```
koic, line 242: Can you refactor to the following?

```suggestion
          return false unless send_node.call_type? || send_node.defined_type?
          return false if send_node.arguments.none? || send_node.parenthesized?

          send_node.dot? || send_node.safe_navigation? || unparenthesized_method_call?(send_node)
```

lib/rubocop/cop/style/ternary_parentheses.rb
```diff
@@ -203,16 +201,46 @@ def correct_parenthesized(corrector, condition)
...
+        end
+
         def whitespace_after?(node)
           last_token = processed_source.last_token_of(node)
           last_token.space_after?
         end
+
+        def condition_send_node(condition)
+          return condition if condition.call_type? || condition.defined_type?
+          return nil unless condition.begin_type? && condition.child_nodes.length == 1
```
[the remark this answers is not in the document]
dvandersluis: Not necessary, I'll simplify. Thank you!

dvandersluis:
@koic thanks for the review! :heart: I've updated with your requested changes.

lib/rubocop/cop/style/ternary_parentheses.rb
```diff
@@ -203,16 +201,47 @@ def correct_parenthesized(corrector, condition)
...
+        end
+
         def whitespace_after?(node)
           last_token = processed_source.last_token_of(node)
           last_token.space_after?
         end
+
+        def condition_send_node(condition)
+          return condition if node_with_args?(condition)
+          return nil unless condition.begin_type?
```
koic, line 229: Is this condition (`return nil unless condition.begin_type?`) necessary as well?

lib/rubocop/cop/style/ternary_parentheses.rb
```diff
@@ -203,16 +201,44 @@ def correct_parenthesized(corrector, condition)
...
+        end
+
         def whitespace_after?(node)
           last_token = processed_source.last_token_of(node)
           last_token.space_after?
         end
+
+        def condition_send_node(condition)
+          send_node = condition.child_nodes.last
+          send_node if node_with_args?(send_node)
```
koic, line 229: Apologies for the repetition feedback, it seems the tests pass even without the `if node_with_args?` guard condition.

dvandersluis: Yup true, because `send_node_args_need_parens?` calls `node_with_args?` too. I refactored away the `condition_send_node` method now.

PR #13385, opened 2024-10-24
[Fix #11514] Add new `Lint/UnescapedBracketInRegexp` cop

### LINKED ISSUE #11514: Cop to catch unescaped `]` in RegExp

dvandersluis:
Adds new cop to emulate the following ruby warning:

```
$ ruby -e '/abc]123/'
-e:1: warning: regular expression has ']' without escape: /abc]123/
```

Fixes #11514.

-----------------

* [x] Added tests.


dvandersluis:
@rubocop/rubocop-core the cop parsing code that would register an offense causes ruby to print out the following warning in addition to the offense:

```
$ rubocop test.rb
Inspecting 1 file
/Users/daniel/.rbenv/versions/3.3.5/lib/ruby/gems/3.3.0/gems/parser-3.3.5.0/lib/parser/builders/default.rb:2266: warning: regular expression has ']' without escape: /abc]123/
W

Offenses:

test.rb:3:5: W: [Correctable] Lint/UnescapedBracketInRegexp: Regular expression has ] without escape.
/abc]123/
    ^
```

Any idea how to suppress it? I tried wrapping `on_regexp` with `RuboCop::Util.silence_warnings` but it didn't help. Even having an empty `on_regexp` method triggers the warning from parser.

Earlopain:
This isn't the only such code that would print such things. `FloatOutOfRange` suffers from the same thing, among a few other cops. These happen during parsing. I openend https://github.com/whitequark/parser/pull/1013 to supress these but there was some concern and I ended up not needing the PR myself anymore.

In theory you can wrap creation of the ProcessedSource in `rubocop-ast` to silence these warnings but I find that to be too broad.

dvandersluis:
@owst @bbatsov I missed that this was previously tackled in #12805 which was recently closed. 

Although there was previous discussion about this being too specific, I think it does make sense given this is a Lint cop emulating a specific ruby warning. If there are other similar warnings they could be handled too but I didn't find any. 

I also handled the unescaped bracket being the first character, as discussed in that cop.

bbatsov:
That's fair. I had kind of missed the point that this was emulating an existing Ruby warning, so being specific is fine here.

config/default.yml
```diff
@@ -2457,6 +2457,11 @@ Lint/UnderscorePrefixedVariableName:
   VersionAdded: '0.21'
   AllowKeywordBlockArguments: false
 
+Lint/UnescapedBracketInRegexp:
+  Description: 'Checks for unescaped literal `[` in Regexp.'
```
bquorning, line 2461: I think you mean `]` here.

lib/rubocop/cop/lint/unescaped_bracket_in_regexp.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Lint
+      # Checks for Regexpes (both literals and via `Regexp.new` / `Regexp.compile`)
+      # that contain unescaped `[` characters.
```
bquorning, line 7: I think you mean `]` here too.

bquorning (commented):
Instead of having a RuboCop cop for this, shouldn’t we just recommend that people use https://github.com/rubocop/rubocop-rubycw ?

config/default.yml
dvandersluis: 🤦 thank you!

dvandersluis:
@bquorning honestly I didn't know that existed, but I think it's still ok because this cop also autocorrects. In any case, there are quite a few existing cops that mirror ruby warnings.

bbatsov:
A bit of context for both of you - originally something very similar to `rubocop-rubycw` was part of RuboCop, but I've decided to emulate the built-in warnings one by one as cops, so that the warnings can be improved in one or another. (e.g. have more informative messages) I didn't want them to be carbon copy replica of the default Ruby behavior and we always have some leeway to do something a bit different/better than whatever's built-in.

bquorning:
Thank you for the extra context @bbatsov.

I was thinking for a minute that perhaps these cops solving `ruby -cw` problems belong better in `rubocop-rubycw`, but upon further thought no. In my (current) opinion, that gem is perfect the way it is, and the more specialized cops with autocorrection and whatnot belong just fine in our `Lint` namespace.

Earlopain:
`rubocop-rubycw` is nice but relies on mri-specific APIs that are unavailable on jruby and others. For those, it shells out once per source file which probably isn't the nicest of experience.

PR #13388, opened 2024-10-25
Fix another error for the documentation generator on Ruby 2.7

Earlopain:
Using `:isolated_environment` results in yard not finding the annotations which causes it to not run a few codepaths. So, just generate the docs into a manually created tempfile and add an option to generate it into that folder instead of `Dir.pwd`

-----------------

* [x] Added tests.


lib/rubocop/cops_documentation_generator.rb
```diff
@@ -158,8 +158,8 @@ def cop_subsection(title, cop)
   end
 
   def example_header(title, cop)
-    content = "[##{to_anchor(title)}-#{to_anchor(cop.cop_name)}]\n"
-    content << +"==== #{title}\n"
+    content = +"[##{to_anchor(title)}-#{to_anchor(cop.cop_name)}]\n"
```
Earlopain, line 161: This is the gist of it. Interpolated strings are frozen on ruby 2.7, following the magic literal comment.

PR #13398, opened 2024-10-29
[Fix #12140] Add new `Style/CombinableDefined` cop

### LINKED ISSUE #12140: New cop to detect unnecessary chains of `defined?` using `&&`
```ruby
# bad
defined?(Rails) && defined?(Rails::Railtie)

# good
defined?(Rails::Railtie) # works fine without the need to check for Rails first
```

Offenses found in the real world:
https://github.com/gregnavis/active_record_doctor/blob/f3b8ba515e400b9c36e1d117ed84f91eecfd78ad/lib/active_record_doctor.rb#L3
https://github.com/airbrake/airbrake/blob/6ca4b6023d56dba214b55f649743b41fa4934edf/lib/airbrake/rails/railties/action_controller_tie.rb#L60-L70 (3 offenses in that file)
https://github.com/discourse/discourse/blob/730f652255474f28e08f7cf442f6ffe2a68354cf/config/initializers/006-mini_profiler.rb#L19
https://github.com/ruby-i18n/i18n/blob/7cf09474b77fd41e65d979134b0525f67cf371b0/test/api/all_features_test.rb#L37
and more.

dvandersluis:
Detects instances of `&&`ed `defined?` calls that are redundant and can be combined.

```ruby
# bad
defined?(Foo) && defined?(Foo::Bar) && defined?(Foo::Bar::Baz)

# good
defined?(Foo::Bar::Baz)

# bad
defined?(foo) && defined?(foo.bar) && defined?(foo.bar.baz)

# good
defined?(foo.bar.baz)
```

Fixes #12140.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/combinable_defined.rb
```diff
@@ -0,0 +1,79 @@
...
+            if node.respond_to?(:namespace)
+              node.namespace
+            elsif node.respond_to?(:receiver)
+              node.receiver
+            end
+          end
+        end
+
+        def remove_clause(corrector, clause)
+          # Removes a `&&` clause including the `&&`.
```
koic, line 65: The comment’s position may not be appropriate, and the article for `&&` might be `an` instead of `a`.

dvandersluis: I've rewritten the comment to be clearer - my intention was to document that it was removing the `&&` operator as well as the `defined?` check.

spec/rubocop/cop/style/combinable_defined_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+      RUBY
+
+      expect_correction(<<~RUBY)
+        defined?(::Foo::Bar)
+      RUBY
+    end
+
+    it 'registers an offense for three `defined?`s with same nesting' do
+      expect_offense(<<~RUBY)
+        defined?(Foo) && defined?(Foo::Bar) && defined?(Foo::Bar::Baz)
```
bbatsov, line 54: Might be good to also cover `defined?(Foo::Bar) && defined?(Foo::Bar::Baz)` with a spec.

bbatsov:
Should we also add a spec for when those happen out of order? (e.g. `defined?(Foo::Bar) && defined?(Foo)` )

lib/rubocop/cop/style/combinable_defined.rb
```diff
@@ -0,0 +1,80 @@
...
+              node.receiver
+            end
+          end
+        end
+
+        def remove_clause(corrector, clause)
+          # Remove the redundant `defined?` node, as well as the subsequent
+          # `&&` operator.
+
+          source = @processed_source.buffer.source
```
koic, line 68: It seems my explanation wasn’t clear. If describing the behavior of the entire method, it would be better to place it as a method comment. For example, something like this.

```suggestion
        # Remove the redundant `defined?` node, as well as the subsequent `&&` operator.
        def remove_clause(corrector, clause)
          source = @processed_source.buffer.source
```

spec/rubocop/cop/style/combinable_defined_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+      RUBY
+
+      expect_correction(<<~RUBY)
+        defined?(foo.bar::BAZ)
+      RUBY
+    end
+
+    it 'registers an offense when there is another term between `defined?`s' do
+      expect_offense(<<~RUBY)
+        foo && defined?(Foo) && bar && defined?(Foo::Bar)
```
Earlopain, line 135: Seems like this would cause `bar` to run after correction where it would previously not, if `Foo` or `Foo::Bar` is not defined.

I don't think it should register an offense if there are terms in-between.

spec/rubocop/cop/style/combinable_defined_spec.rb
Earlopain: also `defined?(Foo) && defined?(Foo::Bar) && defined?(Foo::Baz)`

dvandersluis:
> @bbatsov: Should we also add a spec for when those happen out of order? (e.g. defined?(Foo::Bar) && defined?(Foo) )

Definitely. I'm expecting that isn't likely to be written but if it is the cop should work.

spec/rubocop/cop/style/combinable_defined_spec.rb
dvandersluis: Fair point. I'll change to not register if there are extra terms.

dvandersluis:
@koic @bbatsov @Earlopain thank you for the review. I have addressed all the raised points now.

I also realized that we need to handle using the `and` operator, so that is covered now as well.

bbatsov:
Great work!

PR #13399, opened 2024-10-29
Fix typos `and offense` instead of `an offense`

dvandersluis:
Noticed a few typos of `and offense` instead of `an offense`, so I corrected them. There's a few that were detected that I manually edited to remove some extra words.


spec/rubocop/cop/layout/first_array_element_line_break_spec.rb
```diff
@@ -36,7 +36,7 @@
   end
 
   context 'array nested in a method call' do
-    it 'registers an corrects the offense' do
+    it 'registers and corrects the offense' do
```
dvandersluis, line 39: Found one instance of the opposite so I figured I'd throw that in too.

PR #13404, opened 2024-10-31
Fix `Style/AccessModifierDeclarations` to register access modifiers with multiple symbols

dvandersluis:
Although `private :foo, :bar` is listed in the examples for `Style/AccessModifierDeclarations` (either good or bad depending on the value of the `AllowModifiersOnSymbols` configuration), it isn't actually registered as an offense when `AllowModifiersOnSymbols` is false. 

This also affects autocorrection of groupable modifiers because as-is these are not considered properly when checking siblings (the `right_siblings_same_inline_method?` check). I added autocorrect specs for these. 

This adds test cases for `AllowModifiersOnSymbols` true and false, and fixes the false negative when false. 

NOTE: this relies on rubocop/rubocop-ast#325, and will have failing tests until that's merged.

-----------------

* [x] The PR relates to *only* one subject with loop a clear title and description in grammatically correct, complete sentences.
* [x] Added tests.

### DIFF
--- a/changelog/fix_fix_style_access_modifier_declarations_to_register.md
+++ b/changelog/fix_fix_style_access_modifier_declarations_to_register.md
@@ -1,0 +1,1 @@
+* [#13404](https://github.com/rubocop/rubocop/pull/13404): Fix `Style/AccessModifierDeclarations` to register (as positive or negative, depending on `AllowModifiersOnSymbols` value) access modifiers with multiple symbols. ([@dvandersluis][])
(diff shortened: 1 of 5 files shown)


dvandersluis:
I also added some autocorrect specs and a fix for multiple modifiers mixed between methods and symbol lists.

dvandersluis:
rubocop/rubocop-ast 1.33.1 has been released now. I've updated the PR to use it and tests are passing now.

bbatsov (approved):

bbatsov:
Nice catch!

PR #13405, opened 2024-10-31
[Fix #13372] Fix inconsistent cache directory when using --cache-root

### LINKED ISSUE #13372: `ResultCache.cleanup` inconsistency when using `--cache-root` vs ENV or config file

Capncavedan:
Fixes #13372.

This PR fixes inconsistent cache directory calculation during cache pruning when the `--cache-root` CLI option is used instead of ENV or YAML configuration.

In this work, I renamed method arguments and local variables to try to better distinguish among "cache root override" (an optional value), "cache root" e.g. `.cache`, and "cache working directory" e.g. `.cache/rubocop_cache`.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/fix_cache_pruning_path_when_cache_root_specified_with_cli_option.md
+++ b/changelog/fix_cache_pruning_path_when_cache_root_specified_with_cli_option.md
@@ -1,0 +1,1 @@
+* [#13372](https://github.com/rubocop/rubocop/issues/13372): Add `rubocop_cache` to the path given by `--cache-root` when pruning cache. ([@capncavedan][])
--- a/lib/rubocop/result_cache.rb
+++ b/lib/rubocop/result_cache.rb
@@ -25,16 +25,16 @@ module RuboCop
     # cleaning should be done relatively seldom, since there is a slight risk
     # that some other RuboCop process was just about to read the file, when
     # there's parallel execution and the cache is shared.
-    def self.cleanup(config_store, verbose, cache_root = nil)
+    def self.cleanup(config_store, verbose, cache_root_override = nil)
       return if inhibit_cleanup # OPTIMIZE: For faster testing
 
-      cache_root ||= cache_root(config_store)
-      return unless File.exist?(cache_root)
+      rubocop_cache_dir = cache_root(config_store, cache_root_override)
+      return unless File.exist?(rubocop_cache_dir)
 
-      files, dirs = Find.find(cache_root).partition { |path| File.file?(path) }
+      files, dirs = Find.find(rubocop_cache_dir).partition { |path| File.file?(path) }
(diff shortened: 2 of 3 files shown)


bbatsov:
Looks good to me overall, although I'm not sure about the name choices. Perhaps we should at least document them somewhere in the code as before reading the description of the PR I wasn't sure what the `cache_working_directory` was. @jonas054 Any thoughts from you?

jonas054:
It looks good to me too. I agree that something could be done to improve the naming. @Capncavedan How about changing `cache_working_dir` to `rubocop_cache_dir`? Connecting the variable name to the actual directory name `rubocop_cache` should make it easier to grasp.

jonas054 (changes requested):
It looks good to me too. I agree that something could be done to improve the naming. @Capncavedan How about changing `cache_working_dir` to `rubocop_cache_dir`? Connecting the variable name to the actual directory name `rubocop_cache` should make it easier to grasp.

Capncavedan:
@jonas054 done!

PR #13406, opened 2024-10-31
[Fix #11393] Update `Lint/UnusedMethodArgument` to allow the class names for `IgnoreNotImplementedMethods` to be configured

### LINKED ISSUE #11393: About the NotImplementedError exception for Lint/UnusedMethodArgument
First of all, thanks for an awesome project! 🏅 

The current implementation will not complain about unused method arguments if the method raises a `NotImplementedError` (it was implemented in this PR: https://github.com/rubocop/rubocop/issues/7739).

However, many recommend against raising this exception in your code:

- https://oleg0potapov.medium.com/ruby-notimplementederror-dont-use-it-dff1fd7228e5
- https://blog.airbrake.io/blog/ruby-exception-handling/notimplementederror
- http://chrisstump.online/2016/03/23/stop-abusing-notimplementederror/

These three posts all boil down to:

1. Its actual meaning is different than the name implies – [according to RubyDoc](https://ruby-doc.org/core-3.0.0/NotImplementedError.html) it should be raised when a feature is not implemented on the current platform, think of low-level things like calling fork or fsync.
2. It's a descendant of `ScriptError` so a standard rescue block will not capture this, which may cause unexpected problems.

There is also ongoing discussion in this sibling repo to Rubocop:

- https://github.com/rubocop/ruby-style-guide/issues/458

### Downside

It's somewhat weird to have `def read_db(_name)`, where the real method name is `name`.

### Possible solution

- Add an option called `IgnoredExceptions` (array type), with the default value of `['NotImplementedError']`.
  - For anyone who hasn't configured this cop, it's 100% backwards compatible.
    - People who have set `IgnoreNotImplementedMethods` to `false` would need to update their config to use `IgnoredExceptions: []` intead.
  - Could then deprecate the current option called `IgnoreNotImplementedMethods` and remove it in the next major.
- Allow the yarddoc `@abstract` keyword to be used as another signal to disable this lint check on a method.
- Remove the exception case for `NotImplementedError` (if you think the reasons for doing so makes sense).
- Do nothing and consider this an edge case.

### Docs for Lint/UnusedMethodArgument

image

### Addendum

Also, I realize this is nuanced. So feel free to close this issue if you think it's irrelevant or that the current logic is "good enough".

dvandersluis:
Augments the update to `Lint/UnusedMethodArgument` in #7747 to allow the exception classes used to indicate a not implemented method to be customized (when `IgnoreNotImplementedMethods` is `true`). This allows for codebases that don't want to use `NotImplementedError` or want a custom exception to be handled.

Fixes #11393.

-----------------

* [x] Added tests.

### DIFF
--- a/changelog/change_update_lint_unused_method_argument_to_allow_the.md
+++ b/changelog/change_update_lint_unused_method_argument_to_allow_the.md
@@ -1,0 +1,1 @@
+* [#11393](https://github.com/rubocop/rubocop/issues/11393): Update `Lint/UnusedMethodArgument` to allow the class names for `IgnoreNotImplementedMethods` to be configured. ([@dvandersluis][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -2521,10 +2521,12 @@ Lint/UnusedMethodArgument:
   Enabled: true
   AutoCorrect: contextual
   VersionAdded: '0.21'
-  VersionChanged: '1.61'
+  VersionChanged: '<<next>>'
   AllowUnusedKeywordArguments: false
   IgnoreEmptyMethods: true
   IgnoreNotImplementedMethods: true
+  NotImplementedExceptions:
+    - NotImplementedError
 
 Lint/UriEscapeUnescape:
   Description: >-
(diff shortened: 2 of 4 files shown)


bbatsov:
I like making this more flexible. Thanks!

PR #13409, opened 2024-11-01
[Fix #13395] False positive for `Lint/UselessAssignment` when assigning in branch and block

### LINKED ISSUE #13395: False positive for `Lint/UselessAssignment` when assigning in branch and block
## Expected behavior
The following code warns about a unused variable:

```rb
changed = false

if Random.rand > 1
  changed = true
end

[].each do
  changed = true
end

puts changed
```

I expect no offense for this case. The first assignment is meaningful, removing it will change the output from `false` to `nil`. Changed with https://github.com/rubocop/rubocop/pull/13389, not released yet.

## Actual behavior

```
$ bundle exec rubocop test.rb --only Lint/UselessAssignment
Inspecting 1 file
W

Offenses:

test.rb:1:1: W: [Correctable] Lint/UselessAssignment: Useless assignment to variable - changed.
changed = false
^^^^^^^

```

## RuboCop version

```
Commit 8454d933
$ bundle exec rubocop -V
1.67.0 (using Parser 3.3.5.0, rubocop-ast 1.32.3, analyzing as Ruby 2.7, running on ruby 3.3.5) [x86_64-linux]
  - rubocop-performance 1.22.1
  - rubocop-rake 0.6.0
  - rubocop-rspec 3.2.0
```

pCosta99:
As demonstrated in the issue #13395, there is a false positive as such:
```ruby
changed = false
^^^^^^^ Useless assignment to variable - `changed`.

if Random.rand > 1
  changed = true
end

[].each do
  changed = true
end

puts changed
```

In the presence of a conditional, the cop `Lint/UselessAssignment` can't be certain if the variable will be reassigned or not, so, very much like the case for blocks, the best approach is simply to ignore it.

There appears to be no benefit in actually looking into all branches and checking if all branches reassign since even then the outer assignment defines the variable.

This also applies to case statements.

-----------------

* [X] Added tests.

### DIFF
--- a/lib/rubocop/cop/variable_force/variable.rb
+++ b/lib/rubocop/cop/variable_force/variable.rb
@@ -6,6 +6,8 @@ module RuboCop
       # A Variable represents existence of a local variable.
       # This holds a variable declaration node and some states of the variable.
       class Variable
+        extend NodePattern::Macros
+
         VARIABLE_DECLARATION_TYPES = (VARIABLE_ASSIGNMENT_TYPES + ARGUMENT_DECLARATION_TYPES).freeze
 
         attr_reader :name, :declaration_node, :scope, :assignments, :references, :captured_by_block
@@ -31,10 +33,20 @@ module RuboCop
         def assign(node)
           assignment = Assignment.new(node, self)
 
-          @assignments.last&.reassigned! unless captured_by_block?
+          mark_last_as_reassigned!(assignment)
 
           @assignments << assignment
         end
+
+        def mark_last_as_reassigned!(assignment)
+          return if captured_by_block?
+          return if candidate_condition?(assignment.node.parent)
+
+          @assignments.last&.reassigned!
+        end
+
+        # @!method candidate_condition?(node)
+        def_node_matcher :candidate_condition?, '[{if case case_match when}]'
 
         def referenced?
           !@references.empty?
--- a/changelog/fix_false_positive_for_useless_assignment.md
+++ b/changelog/fix_false_positive_for_useless_assignment.md
@@ -1,0 +1,1 @@
+* [#13395](https://github.com/rubocop/rubocop/issues/13395): Fix a false positive for `Lint/UselessAssignment` when assigning in branch and block. ([@pCosta99][])
--- a/spec/rubocop/cop/lint/useless_assignment_spec.rb
+++ b/spec/rubocop/cop/lint/useless_assignment_spec.rb
@@ -551,6 +551,51 @@ RSpec.describe RuboCop::Cop::Lint::UselessAssignment, :config do
     end
   end
 
+  context 'when assigning in branch' do
+    it 'accepts' do
+      expect_no_offenses(<<~RUBY)
+        def some_method
+          changed = false
+
+          if Random.rand > 1
+            changed = true
+          end
+
+          [].each do
+            changed = true
+          end
+
+          puts changed
+        end
+      RUBY
+    end
+  end
+
+  context 'when assigning in case' do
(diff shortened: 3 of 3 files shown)


lib/rubocop/cop/variable_force/variable.rb
```diff
@@ -31,11 +33,21 @@ def initialize(name, declaration_node, scope)
...
 
-          @assignments.last&.reassigned! unless captured_by_block?
+          mark_last_as_reassigned!(assignment)
 
           @assignments << assignment
         end
 
+        def mark_last_as_reassigned!(assignment)
+          return if captured_by_block?
+          return if candidate_condition?(assignment.node.parent)
```
pCosta99, line 43: @koic Can I make `parent` an attribute in `RuboCop::Cop::VariableForce::Assignment`? 
Seems like a worthy refactor, makes sense as an attribute and simplifies a bunch of accesses throughout the class

bbatsov:
> There appears to be no benefit in actually looking into all branches and checking if all branches reassign since even then the outer assignment defines the variable.

I don't quite get this part, as by looking at them you can actually determine whether some assignment is void or not.

pCosta99:
> > There appears to be no benefit in actually looking into all branches and checking if all branches reassign since even then the outer assignment defines the variable.
> 
> I don't quite get this part, as by looking at them you can actually determine whether some assignment is void or not.

Can you create an example? What I was trying to highlight is situations like this:

```ruby
changed = false

if Random.rand > 1
  changed = true
else 
  changed = false
end

[].each do
  changed = true
end

puts changed
```

Even considering that any branch will assign to `changed` the assignment above is not useless since it creates the variable. 
Out of the top of my head no examples come to mind that it matters to check all branches.

bbatsov:
> Even considering that any branch will assign to changed the assignment above is not useless since it creates the variable.

So, your point applies only if there's a usage outside the branches, right? (in your case the `puts`)

pCosta99:
> > Even considering that any branch will assign to changed the assignment above is not useless since it creates the variable.
> 
> So, your point applies only if there's a usage outside the branches, right? (in your case the `puts`)

I'd say so. And testing with, for instance:
```ruby
def some_method
  changed = false
  ^^^^^^^ Useless assignment to variable - `changed`.

  if Random.rand > 1
    changed = true
    ^^^^^^^ Useless assignment to variable - `changed`.
  else
    changed = false
    ^^^^^^^ Useless assignment to variable - `changed`.
  end
end
```

works as expected. However, one important detail, in the following scenario:

```ruby
def some_method
  changed = false

  if Random.rand > 1
    changed = true
  end

  [].each do
    changed = true
  end
end
```

the variable is not unused since it is captured by a block. This is respectful of the comments under `RuboCop::Cop::VariableForce::Variable::used?` that quote the following:

>         # This is a convenient way to check whether the variable is used
>        # in its entire variable lifetime.
>        # For more precise usage check, refer Assignment#used?.
>        #
>        # Once the variable is captured by a block, we have no idea
>        # when, where, and how many times the block would be invoked.
>        # This means we cannot track the usage of the variable.
>       # So we consider it's used to suppress false positive offenses.

@bbatsov

pCosta99:
It's been a long time, I've kinda forgotten about this, sorry, where do we stand on this @bbatsov ?

navels:
This is still needed . . . thanks

pCosta99:
@navels not sure how to proceed here to be honest

koic:
Let’s merge this proposal for now and proceed with resolving the issue. The different tasks can be addressed separately.

navels:
Thanks!

PR #13410, opened 2024-11-01
Grammar fixes

dvandersluis:
Using the `Regex` option of `Naming/InclusiveLanguage`, I found and corrected 4 types of English issues in the codebase. I didn't include the changes into .rubocop.yml because they make things pretty slow (and also still need to be manually reviewed for false positive because English grammar is not a great use case for regex).

For ease of review I have them each as a separate commit.


bbatsov:
Nicely done!

PR #13411, opened 2024-11-02
Fix `Style/BitwisePredicate` when using a method

d4be4st:
Fix `Style/BitwisePredicate` when using a method.

When having a regular method named `positive?` or `zero?` the cop would raise an error.

```ruby
def positive?
  some_ruby_code
end
```
Would throw a 
```
NoMethodError: undefined method `begin_type?' for nil:NilClass
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/bitwise_predicate.rb
+++ b/lib/rubocop/cop/style/bitwise_predicate.rb
@@ -71,6 +71,7 @@ module RuboCop
         PATTERN
 
         def on_send(node)
+          return unless node.receiver
           return unless node.receiver.begin_type?
           return unless (preferred_method = preferred_method(node))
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/bitwise_predicate.rb
```diff
@@ -71,6 +71,7 @@ class BitwisePredicate < Base
         PATTERN
 
         def on_send(node)
+          return unless node.receiver
           return unless node.receiver.begin_type?
```
dvandersluis, line 75: These two lines can be combined: 

```suggestion
          return unless node.receiver&.begin_type?
```

d4be4st: Sure, was not sure how big you all were on the safe navigation operator :)

koic:
Thanks!

PR #13413, opened 2024-11-03
Reduce node destructuring

bquorning:
Now that rubocop-ast has a `MasgnNode` and a `MlhsNode`, and `CasgnNode`, `AsgnNode` etc. has gotten some extra methods, we can remove some node destructuring from our code base.

See https://github.com/rubocop/rubocop-ast/pull/203 and https://github.com/rubocop/rubocop-ast/pull/326 for the implementations I depend on.

cc @dvandersluis 

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/redundant_self.rb
+++ b/lib/rubocop/cop/style/redundant_self.rb
@@ -66,14 +66,12 @@ module RuboCop
         # Assignment of self.x
 
         def on_or_asgn(node)
-          lhs, _rhs = *node
-          allow_self(lhs)
+          allow_self(node.lhs)
         end
         alias on_and_asgn on_or_asgn
 
         def on_op_asgn(node)
-          lhs, _op, _rhs = *node
-          allow_self(lhs)
+          allow_self(node.lhs)
         end
 
         # Using self.x to distinguish from local variable x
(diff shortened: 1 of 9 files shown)


lib/rubocop/cop/style/redundant_self.rb
```diff
@@ -92,13 +90,11 @@ def on_blockarg(node)
         end
 
         def on_masgn(node)
-          lhs, rhs = *node
-          add_masgn_lhs_variables(rhs, lhs)
+          add_masgn_lhs_variables(node.expression, node.lhs)
```
dvandersluis, line 93: You could use `node.rhs` for balance if you'd like (I added it as an alias with this in mind but up to you of course!)

dvandersluis (approved):
Looks good!

lib/rubocop/cop/style/redundant_self.rb
bquorning: Good idea, but… I can use `#rhs` here, but not in line 97. Do you think we need an `alias rhs expression` for `AsgnNode` too?

dvandersluis: I think it's ok not having `rhs` for your specific example because you're also not using `lhs`.

That being said, it _could_ help allow `masgn` to be treated similiarly to the other assignment nodes, or at least have a common nomenclature for assignments. If we were to do so we should add `lhs` too (I guess an alias for `name`). The obvious caveat there is that for a generic assignment `lhs` is either a string or a `mlhs`.

bquorning: > I'm happy to open a rubocop-ast PR tomorrow if you don't 

I will leave that to you 😄

dvandersluis: Sounds good, I opened rubocop/rubocop-ast#326.

bquorning:
@dvandersluis I have changed some more files to use the new methods you added in rubocop-ast. Please re-review.

dvandersluis (approved):
Good improvements @bquorning!

dvandersluis:
Thank you!

PR #13420, opened 2024-11-05
Update `Lint/RedundantSafeNavigation` to register an offense when the receiver is `self`

dvandersluis:
It is redundant to use safe-navigation on `self` because `self` always is the current object, and therefore not nil (other than if reopening `NilClass`).

```ruby
# bad
self&.foo

# good
self.foo
```

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/redundant_safe_navigation.rb
+++ b/lib/rubocop/cop/lint/redundant_safe_navigation.rb
@@ -6,7 +6,7 @@ module RuboCop
       # Checks for redundant safe navigation calls.
       # Use cases where a constant, named in camel case for classes and modules is `nil` are rare,
       # and an offense is not detected when the receiver is a constant. The detection also applies
-      # to literal receivers, except for `nil`.
+      # to `self`, and to literal receivers, except for `nil`.
       #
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/redundant_safe_navigation.rb
```diff
@@ -6,7 +6,7 @@ module Lint
       # Checks for redundant safe navigation calls.
       # Use cases where a constant, named in camel case for classes and modules is `nil` are rare,
       # and an offense is not detected when the receiver is a constant. The detection also applies
-      # to literal receivers, except for `nil`.
+      # to `self`, and literal receivers, except for `nil`.
```
koic, line 9: That's my two cents. In this case, the comma seems unnecessary.

```suggestion
      # to `self` and literal receivers, except for `nil`.
```

dvandersluis: Because there are two separate clauses that detection applies to (1. `self`; 2. literal receivers, except for `nil`), the comma is useful to emphasize that the "except for" clause only belongs to the second one. However, I've changed it slightly to further emphasize that (by adding a second "to"), thanks.

koic:
Thanks!

PR #13426, opened 2024-11-06
Where possible, use node extensions instead of `*splat`

bquorning:
This PR fixes most of the `InternalAffairs/NodeDestructuring` offenses. New features are constantly being added to Node subclasses in rubocop-ast, we just need to actually use them in rubocop.

-----------------

* [x] Added tests.
* [ ] ~Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.~


bquorning:
cc @dvandersluis 

Would you have time to look into this new RuboCop error?
```
An error occurred while Style/ConditionalAssignment cop was inspecting /home/runner/work/rubocop/rubocop/lib/rubocop/comment_config.rb:206:8.
```
I _think_ there may be a bug in rubocop-ast.

bquorning:
Also I’m a bit confused about when to use `SendNode#first_argument` vs `SendNode#last_argument`.

bbatsov:
> Fixing most InternalAffairs/NodeDestructuring offenses, but perhaps we should cherry-pick them a bit and not do it all in one big PR.

I think it'd be fine to do the change in bulk, as I don't see a pretty meaningful way to split it in chunks.

bquorning:
> I _think_ there may be a bug in rubocop-ast.

I opened https://github.com/rubocop/rubocop-ast/issues/331

bquorning:
I added one more commit, to work around the rubocop-ast bug.

lib/rubocop/cop/mixin/check_assignment.rb
```diff
@@ -17,27 +17,19 @@ def on_lvasgn(node)
...
-          _scope, _lhs, rhs = *node
-        elsif node.op_asgn_type?
-          _lhs, _op, rhs = *node
-        elsif node.call_type?
-          rhs = node.last_argument
-        elsif node.assignment?
-          _lhs, rhs = *node
+        if node.call_type?
+          node.last_argument
+        elsif node.casgn_type? || node.op_asgn_type? || node.assignment?
```
dvandersluis, line 30: `node.assignment?` hanldes `casgn_type?` and `op_asgn_type?`
```suggestion
        elsif node.assignment?
```

lib/rubocop/cop/layout/empty_lines_around_exception_handling_keywords.rb
```diff
@@ -128,7 +128,7 @@ def keyword_locations_in_rescue(node)
         end
 
         def keyword_locations_in_ensure(node)
-          ensure_body, = *node
+          ensure_body, = *node # rubocop:disable InternalAffairs/NodeDestructuring
```
dvandersluis, line 131: `EnsureNode#body` is available, so you should be able to use `node.body` here.

lib/rubocop/cop/layout/empty_lines_around_exception_handling_keywords.rb
```diff
@@ -72,7 +72,7 @@ def on_def(node)
         alias on_numblock on_def
 
         def on_kwbegin(node)
-          body, = *node
+          body, = *node # rubocop:disable InternalAffairs/NodeDestructuring
```
dvandersluis, line 75: rubocop/rubocop-ast#333 😄

lib/rubocop/cop/layout/empty_lines_around_exception_handling_keywords.rb
bquorning: `node.body` returns `node.children[1]`, but what’s called `ensure_body` here is actually `node.children[0]`. The wording confuses me.

I might instead change this to
```rb
rescue_body_without_ensure = node.children.first
```

dvandersluis: Yeah I opened rubocop/rubocop-ast#334 because I don't think `EnsureNode#body` is named consistently.

bquorning: Ah. I’ll wait for #333 and use `rescue_body = node.rescue_node` 😅

lib/rubocop/cop/layout/space_around_operators.rb
```diff
@@ -115,35 +113,32 @@ def on_send(node)
...
-          _, rhs, = *node
-
-          return unless rhs
+          return unless node.rhs
 
-          check_operator(:assignment, node.loc.operator, rhs)
+          check_operator(:assignment, node.loc.operator, node.rhs)
         end
 
         def on_casgn(node)
```
dvandersluis, line 121: `on_casgn` can be combined into `on_assignment`

lib/rubocop/cop/layout/space_around_operators.rb
```diff
@@ -115,35 +113,36 @@ def on_send(node)
...
 
           return unless rhs
 
           check_operator(:binary, node.loc.operator, rhs)
         end
 
         def on_special_asgn(node)
-          _, _, right, = *node
+          # node.first_argument - is this correct?
+          rhs = node.send_type? ? node.first_argument : node.rhs
```
bquorning, line 141: @dvandersluis Is this correct? When do I use `SendNode#first_argument` vs `SendNode#last_argument`?

dvandersluis: In this case yes it's correct because it's not possible to pass multiple arguments to a setter method. `x.y = 3, 5` acts as an implicit array (`[3,5]`), so it'll always be the first argument there. Probably would be good to add this to the tests of this cop.

FWIW I'd probably refactor `on_special_asgn` to `on_setter_method` to only deal with the send node case, and then the other case can be handled by `on_assignment` as well since it'll be the same at that point except for the type, which could be set in `on_assignment`.

lib/rubocop/cop/style/swap_values.rb
```diff
@@ -79,26 +79,15 @@ def replacement(x_assign)
...
-              "#{namespace.const_name}::#{name}"
-            else
-              name.to_s
-            end
+          if node.casgn_type? && node.namespace
+            "#{node.namespace.const_name}::#{node.name}"
           else
-            node.children[0].to_s
+            node.name.to_s
           end
```
dvandersluis, line 86: Now that rubocop/rubocop-ast#330 is released in v1.34.1, you could do this instead:

```suggestion
          if node.casgn_type?
            "#{'::' if node.absolute?}#{node.const_name}"
          else
            node.name.to_s
          end
```

lib/rubocop/cop/style/self_assignment.rb
```diff
@@ -82,13 +78,11 @@ def autocorrect(corrector, node)
         end
 
         def autocorrect_send_node(corrector, node, rhs)
-          _receiver, method_name, args = *rhs
```
bquorning, line 85: I guess this worked by accident before. `args` gets assigned the _first_ argument to the rhs `SendNode`.

lib/rubocop/cop/layout/space_around_operators.rb
bquorning: I didn’t quite follow that. Would you be up for pushing a commit to my branch?

bquorning:
> Also I’m a bit confused about when to use `SendNode#first_argument` vs `SendNode#last_argument`.

I added a commit with guard clauses like `return unless node.arguments.one?` to make it safer to use `SendNode#first_argument`. Instead, I guess I’d have to do e.g. `node.arguments.map(&:source).join(', ')` when auto-correcting 😬

lib/rubocop/cop/layout/space_around_operators.rb
dvandersluis: @bquorning I pushed a commit with what I meant.

lib/rubocop/cop/style/self_assignment.rb
dvandersluis: `args` was already the first argument here, any node children after 3 were just ignored.

dvandersluis:
@bquorning I pushed a second commit to handle your `arguments.one?` clauses in a slightly different way, plus adding tests. A couple of them shouldn't have registered an offense at all if there are more than one argument. 

I'm happy to extract this change to a separate PR if you'd prefer.

bquorning:
> I'm happy to extract this change to a separate PR if you'd prefer.

Not at all. Adding to this branch is perfect 👍🏼

I think there are a few more places (look for `first_argument` / `last_argument`) that could perhaps use a guard clause (and tests) in case people pass multiple argument with a SendNode.

dvandersluis:
@bquorning looks like caused some rubocop offenses with my change. Do you want me to fix it? I don't like force pushing to someone else's branch unexpectedly. 

Otherwise, is there anything else to add on here?

bquorning:
> @bquorning looks like I caused some rubocop offenses with my change. Do you want me to fix it? I don't like force pushing to someone else's branch unexpectedly.

I fixed it.

bquorning:
> Otherwise, is there anything else to add on here?

I was hoping for https://github.com/rubocop/rubocop-ast/pull/333 to be ready, but we can incorporate that into a later PR.

I’ll mark this PR as Ready for review now.

dvandersluis:
> I was hoping for https://github.com/rubocop/rubocop-ast/pull/333 to be ready, but we can incorporate that into a later PR.

rubocop-ast 1.35 is released now.

bquorning:
> rubocop-ast 1.35 is released now.

😄 Things are moving fast. I’ve moved this PR back to draft; will mark it as ready (again) when I’ve taken what I can use from that v1.35.

bquorning:
> … when I’ve taken what I can use from that v1.35.

I only think lib/rubocop/cop/layout/empty_lines_around_exception_handling_keywords.rb could potentially be changed, but I think it will be easier after https://github.com/rubocop/rubocop-ast/issues/334 is merged.

lib/rubocop/cop/layout/empty_lines_around_exception_handling_keywords.rb
```diff
@@ -72,7 +72,7 @@ def on_def(node)
...
           check_body(body, node.loc.line)
```
dvandersluis, line 76: My previous suggestion to use `KeywordBeginNode.body` won't actually work without heavy refactoring because this cop looks at the rescue and ensure too. I'd just change it like so now:

```suggestion
          check_body(node.children.first, node.loc.line)
```

dvandersluis (approved):
@bquorning I’ve gone over this PR a lot of times now and I think we’re good to go. There’s a conflict to be fixed but the longer it sits the more conflicts will arise so once you’re ready I think we should merge.

bquorning:
The PR is ready for review now. Thank you for your patience.

bquorning:
Thank you for all the help @dvandersluis

PR #13427, opened 2024-11-06
Show warning if cop configurations create an inconsistency that will cause unexpected behaviour

dvandersluis:
Follows https://github.com/rubocop/rubocop/pull/13228#issuecomment-2458766634

Since there is configuration for `Style/SafeNavigation` and `Style/SafeNavigationChainLength` that could cause an incompatibility, introduce a warning that will show so that users can correct their configuration.

The warning looks like this:

image

It only is shown if both cops are enabled and have non-nil values set. 

Similarly to `ConfigObsoletion`, I built this so that it could be extended by extension gems if desired. Right now it's only used for this pair of cops but others could easily be added as needed. 

-----------------

* [x] Added tests.


spec/rubocop/config_spec.rb
```diff
@@ -603,28 +628,46 @@
...
+    context 'when Include config only includes regular paths' do
+      let(:hash) do
+        { 'AllCops' => { 'Include' => ['**/Gemfile', 'config/unicorn.rb.example'] } }
+      end
 
-      expect(configuration).not_to be_possibly_include_hidden
+      it 'returns true' do
+        expect(configuration).not_to be_possibly_include_hidden
+      end
     end
```
dvandersluis, line 639: Because `InconsistentConfig` uses `@config.for_cop`, configuration is loaded prior to these tests running, so `configuration['AllCops']` doesn't actually change anything. I updated the cops in this file to set the config first which makes the cop run. I'm not sure if there's a way I can get around this but I also don't think this should affect actual rubocop running (ie. I think this is a quirk of how config is loaded in these tests).

spec/rubocop/cop/message_annotator_spec.rb
```diff
@@ -138,12 +138,22 @@
...
+            'Cop/Cop' => {
+              'StyleGuide' => '../rails-style-guide#target_based_url'
+            }
+          )
+        end
+
+        it 'can accept relative paths' do
+          config['Cop/Cop'] = { 'StyleGuide' => '../rails-style-guide#target_based_url' }
+          expect(annotate).to include('https://github.com/rubocop/rails-style-guide#target_based_url')
+        end
```
dvandersluis, line 156: Same thing here.

koic:
I’d like to consider a simpler solution through implementation rather than introducing a new configuration. Although it will be some time before I can begin, I should be able to work on this, including the design.

bbatsov:
> I’d like to consider a simpler solution through implementation rather than introducing a new configuration.

A solution that's simpler, but still generic? I don't mind the proposed approach, but I'd be curious to see what alternative you'd present when you find to the time to work on it.

dvandersluis:
As I mentioned in my comment, I don't think it really makes sense to have two different places where the max chain length is defined that could conflict in the first place. However, until such time as that is changed, I thought it'd be useful to at least warn the user if they've configured themselves into an inconsistent situation. 

If you don't think this is worthwhile as an interim solution, let's close this.

PR #13429, opened 2024-11-07
Fix spacing in Lint/NoReturnInBeginEndBlocks

Zopolis4:
This doesn't get caught by `Layout/ExtraSpacing` because `AllowForAlignment` is enabled, but I presume this is still something that can be manually fixed. I assume this is not the correct layout because `lint/self_assignment.rb` has three `alias` calls with the regular spacing from lines 53-55:
```ruby
          add_offense(node) if rhs.type == rhs_type && rhs.source == lhs.to_s
        end
        alias on_ivasgn on_lvasgn
        alias on_cvasgn on_lvasgn
        alias on_gvasgn on_lvasgn

        def on_casgn(node)
```

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/no_return_in_begin_end_blocks.rb
+++ b/lib/rubocop/cop/lint/no_return_in_begin_end_blocks.rb
@@ -45,8 +45,8 @@ module RuboCop
             end
           end
         end
-        alias on_or_asgn    on_lvasgn
-        alias on_op_asgn    on_lvasgn
+        alias on_or_asgn on_lvasgn
+        alias on_op_asgn on_lvasgn
       end
     end
   end


Zopolis4:
Also, because `on_or_asgn` and `on_op_asgn` have the same length and there are no other bordering lines to be aligned, there's no need for extra spaces to align them. Would it make sense to have that trigger `Layout/ExtraSpacing` even when `AllowForAlignment` is true, or would that frustrate attempts to start aligned lists?

bbatsov:
>  Would it make sense to have that trigger Layout/ExtraSpacing even when AllowForAlignment is true, or would that frustrate attempts to start aligned lists?

Probably working on this is not going to be the best use of our time, as the impact is tiny.

PR #13432, opened 2024-11-07
Fix false positive for `Lint/FloatComparison` against nil

lovro-bikic:
Hi! This PR fixes a false positive for Lint/FloatComparison against nil.

Here is an example when a false positive can occur:
```ruby
Float(user_submitted_value, exception: false) == nil
```

If `user_submitted_value` is a value that cannot be made a Float, `nil` is returned because of the `exception: false` param. 

-----------------

* [x] Added tests.


lovro-bikic:
Done, let me know if it should be more detailed.

dvandersluis:
Thank you!

PR #13436, opened 2024-11-08
Recognize percent literal array arg in safe navigation for `Lint/RedundantSplatExpansion`

lovro-bikic:
Fixes incorrect offense and auto-correction for `Lint/RedundantSplatExpansion` cop when percent literal array is used as an argument in a **safe navigation** method call.

Here's a (renamed) production example from a Rails controller:
```ruby
params[:param_which_may_not_exist]&.exclude(*%w[foo bar])
```

When `AllowPercentLiteralArrayArgument` option is `true` (by default), the cop shouldn't register an offense, but it does:
image
and auto-corrects it to (removes the `*`):
```ruby
params[:param_which_may_not_exist]&.exclude(%w[foo bar])
```
which is not correct (nor desired in this case).

When `AllowPercentLiteralArrayArgument` is `false`, the cop prints an offense as expected, but the message is wrong (I believe):
image

The correct auto-correction (only in the latter case), should be:
```ruby
params[:param_which_may_not_exist]&.exclude('foo', 'bar')
```

Note also that when safe navigation is not used, e.g.:
```ruby
params[:param_which_may_not_exist].exclude(*%w[foo bar])
```
the cop works as expected (no offense for the default cop option, and the expected auto-correct for the other option).

This patch contains tests which fail on master and a fix for the issue. I've done my best with my limited knowledge of rubocop-ast, let me know if there's a better way to fix this 🙏 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/redundant_splat_expansion.rb
+++ b/lib/rubocop/cop/lint/redundant_splat_expansion.rb
@@ -157,7 +157,7 @@ module RuboCop
         end
 
         def method_argument?(node)
-          node.parent.send_type?
+          node.parent.call_type?
         end
 
         def part_of_an_array?(node)
@@ -171,7 +171,7 @@ module RuboCop
           parent = node.parent
           grandparent = node.parent.parent
 
-          parent.when_type? || parent.send_type? || part_of_an_array?(node) ||
+          parent.when_type? || method_argument?(node) || part_of_an_array?(node) ||
             grandparent&.resbody_type?
         end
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/lint/redundant_splat_expansion.rb
```diff
@@ -157,7 +157,7 @@ def array_splat?(node)
         end
 
         def method_argument?(node)
-          node.parent.send_type?
+          node.parent.send_type? || node.parent.csend_type?
```
koic, line 160:
```suggestion
          node.parent.call_type?
```

lovro-bikic: Fixed, thanks!

koic:
Thanks!

PR #13440, opened 2024-11-09
[Fix #13437] Add new cop `Lint/UselessDefined`

### LINKED ISSUE #13437: Cop idea: Detect usages of `defined?` that will always be truthy

`defined?` is a bit of a weird one, both in what it accept and in what it returns.

Usually it is used to check if a constant is defined via `defined?(FooBar)`. But if the user is not very familiar with it they might write it as `defined?("FooBar")` or `defined?(:FooBar)` instead in order to not raise an exception during runtime when Ruby seemingly tries to access the constant. Ruby handles this of course. In fact you can write anything in there as long as it is not a syntax error and it will produce some kind of result.

Examples:
* https://github.com/sidekiq/sidekiq/pull/6474/files#diff-ebb94e42d8a2439ff96291e25860212cec96b8d9372a767a4f673255ea2c15c0R19
* https://github.com/sidekiq/sidekiq/pull/6430/files#diff-ebb94e42d8a2439ff96291e25860212cec96b8d9372a767a4f673255ea2c15c0R20
* https://github.com/ruby-amqp/kicks/pull/19/files#diff-f2f6997940a8dd633e0f0a95b2e66c427d578ddb62983a86523f253a17cb6a1dR5
* https://github.com/ruby-amqp/kicks/pull/12/files#diff-f2f6997940a8dd633e0f0a95b2e66c427d578ddb62983a86523f253a17cb6a1dR4

```rb
# bad
defined?("FooBar") # will always return "expression"
defined?(:FooBar) # will also always return "expression"
defined?("foo#{bar}") # you might have guessed it, also "expression"

# good - anything else
defined?(FooBar)
defined?(foo_bar)
defined?([foo, bar]) # ruby checks each entry for existance. You probably shouldn't write this as well but :shrug:
# etc. etc.
```

Earlopain:
Fix #13437.

To detect code such as this:
```rb
defined?("FooBar")
defined?(:FooBar)
defined?("foo#{bar}")
```

Interpolation is included. Ruby does not check individual parts, for example `defined?("foo#{bar}")` is truthy even if `bar` doesn't exist.

It will find code such as the following:
https://github.com/search?q=lang%3Aruby+%2F+defined%5C%3F%5C%28%5B%3A%22%27%5D%5Ba-zA-Z%5D%2F&type=code

Extra reading: https://github.com/ruby/spec/blob/bbcd077ac3ff6da3730f388c61d4fcbbf085f4ab/language/defined_spec.rb

While you could add a bunch more here, nobody actually writes `defined?(return)`. Detecting such code and similar feels like a waste of time.

-----------------

* [x] Added tests.


config/default.yml
```diff
@@ -1724,6 +1724,11 @@ Lint/Debugger:
       - debug/open
       - debug/start
 
+Lint/DefinedWithFixedResult:
```
vlad-pisanov, line 1727: feels like the name of the cop should convey that the result is not only fixed, but also unexpectedly truthy.

"always defined argument"? "ineffective defined argument"? 🤷‍♂️

Earlopain: I renamed it to `UselessDefined`, there are already quite a few `Useless*` cops

lib/rubocop/cop/lint/defined_with_fixed_result.rb
```diff
@@ -0,0 +1,48 @@
...
+      #   defined?(FooBar)
+      #   defined?(foo_bar)
+      #
+      #   # bad - interpolation
+      #   defined?("#{name}::Foo")
+      #   defined?(:"#{name}::Foo")
+      #
+      #   # good
+      #   defined?(::Full::Path::To::Foo)
+      #   const_defined?(:Foo)
```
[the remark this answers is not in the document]
Earlopain: I added a more complex example for the constant case and added a snippet about using the more specific methods instead

config/default.yml
bbatsov: I also prefer the name `UselessDefined`.

koic:
It looks like the cop name in the commit message and PR title is not updated.

lib/rubocop/cop/lint/useless_defined.rb
```diff
@@ -0,0 +1,56 @@
...
+      # Such calls will always return `'expression'`, you probably meant to
+      # check for the existence of a constant, method, or variable instead.
+      #
+      # `defined?` is part of the Ruby syntax and doesn't behave like normal methods.
+      # You can safely pass in what you are checking for directly, without encountering
+      # a `NameError`.
+      #
+      # When interpolation is used, oftentimes it is not possible to write the
+      # code with `defined?`. In these cases, switch to one of the more specific methods:
+      # * `class_variable_defined?`
```
koic, line 16: Please insert a blank line.

```suggestion
      # code with `defined?`. In these cases, switch to one of the more specific methods:
      #
      # * `class_variable_defined?`
```

lib/rubocop/cop/lint/useless_defined.rb
```diff
@@ -0,0 +1,56 @@
...
+      # * `const_defined?`
+      # * `method_defined?`
+      # * `instance_variable_defined?`
+      # * `binding.local_variable_defined?`
+      #
+      #   # bad
```
koic, line 22:
```suggestion
      # @example
      #
      #   # bad
```

lib/rubocop/cop/lint/useless_defined.rb
```diff
@@ -0,0 +1,59 @@
...
+        # @!method offense?(node)
+        def_node_matcher :offense?, <<~PATTERN
+          (defined (${#{TYPES.keys.join(' ')}} ...))
+        PATTERN
+
+        def on_defined?(node)
+          offense?(node) do |type|
+            add_offense(node, message: format(MSG, type: TYPES[type]))
+          end
+        end
```
dvandersluis, line 55: You could simplify this without a node pattern:

```suggestion
        def on_defined?(node)
          return unless node.first_argument.type?(*TYPES.keys)

          add_offense(node, message: format(MSG, type: TYPES[node.first_argument.type]))
        end
```

Earlopain: That is nicer. `ReturnNilInPredicateMethodDefinition` doesn't like it though, had to configure it

PR #13466, opened 2024-11-17
Fix a test failure on 3.4-dev with webmock

Earlopain:
`net/http` removed old constants that webmock relies on. This can be reverted when https://github.com/bblimke/webmock/pull/1081 is released


koic:
Thanks!

PR #13472, opened 2024-11-19
Improve syntax error output for `rake documentation_syntax_check`

dvandersluis:
Added context to the warning that there's a syntax error in `rake documentation_syntax_check`.

image


koic:
Nice 👍

PR #13473, opened 2024-11-19
Update `Lint/ItWithoutArgumentsInBlock` to not register an offense in Ruby 3.4

dvandersluis:
`it` as a block parameter should not register an offense in Ruby 3.4. Currently if using 3.4-dev and setting `TargetRubyVersion: 3.4` in `.rubocop.yml`, rubocop will give offenses when using `it`.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/target_ruby_version.rb
+++ b/lib/rubocop/cop/mixin/target_ruby_version.rb
@@ -8,12 +8,28 @@ module RuboCop
         @minimum_target_ruby_version
       end
 
+      def required_maximum_ruby_version
+        @maximum_target_ruby_version
+      end
+
       def minimum_target_ruby_version(version)
         @minimum_target_ruby_version = version
       end
 
+      def maximum_target_ruby_version(version)
+        @maximum_target_ruby_version = version
+      end
+
       def support_target_ruby_version?(version)
-        required_minimum_ruby_version <= version
+        # By default, no minimum or maximum versions of ruby are required
+        # to run any cop. In order to do a simple numerical comparison of
+        # the requested version against any requirements, we use 0 and
+        # Infinity as the default values to indicate no minimum (0) and no
+        # maximum (Infinity).
+        min = required_minimum_ruby_version || 0
+        max = required_maximum_ruby_version || Float::INFINITY
+
+        min <= version && max >= version
       end
     end
   end
--- a/lib/rubocop/cops_documentation_generator.rb
+++ b/lib/rubocop/cops_documentation_generator.rb
@@ -114,7 +114,15 @@ class CopsDocumentationGenerator # rubocop:disable Metrics/ClassLength
   def required_ruby_version(cop)
     return '' unless cop.respond_to?(:required_minimum_ruby_version)
 
-    "NOTE: Required Ruby version: #{cop.required_minimum_ruby_version}\n\n"
+    if cop.required_minimum_ruby_version
+      requirement = cop.required_minimum_ruby_version
+    elsif cop.required_maximum_ruby_version
+      requirement = "<= #{cop.required_maximum_ruby_version}"
+    else
+      return ''
+    end
+
+    "NOTE: Requires Ruby version #{requirement}\n\n"
   end
 
   # rubocop:disable Metrics/MethodLength
--- a/changelog/change_update_it_without_arguments_in_block_to_not.md
+++ b/changelog/change_update_it_without_arguments_in_block_to_not.md
@@ -1,0 +1,1 @@
+* [#13473](https://github.com/rubocop/rubocop/pull/13473): Update `Lint/ItWithoutArgumentsInBlock` to not register offenses in Ruby 3.4. ([@dvandersluis][])
--- a/lib/rubocop/cop/lint/it_without_arguments_in_block.rb
+++ b/lib/rubocop/cop/lint/it_without_arguments_in_block.rb
@@ -26,6 +26,9 @@ module RuboCop
       #
       class ItWithoutArgumentsInBlock < Base
         include NodePattern::Macros
+        extend TargetRubyVersion
+
+        maximum_target_ruby_version 3.3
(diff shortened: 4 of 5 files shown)


lib/rubocop/cop/mixin/target_ruby_version.rb
```diff
@@ -5,15 +5,24 @@ module Cop
     # Common functionality for checking target ruby version.
     module TargetRubyVersion
       def required_minimum_ruby_version
-        @minimum_target_ruby_version
+        @minimum_target_ruby_version || 0
```
Earlopain, line 8: Can you check the documentation task? I think this will cause it to display a wonky required ruby version for this cop.

dvandersluis: Oh good call, you're probably right.

dvandersluis: Fixed now.

lib/rubocop/cops_documentation_generator.rb
```diff
@@ -114,7 +114,15 @@ def safety_object(safety_objects, cop)
   def required_ruby_version(cop)
     return '' unless cop.respond_to?(:required_minimum_ruby_version)
 
-    "NOTE: Required Ruby version: #{cop.required_minimum_ruby_version}\n\n"
+    if cop.required_minimum_ruby_version.nonzero?
+      note = "Required Ruby version: #{cop.required_minimum_ruby_version}"
+    elsif cop.required_maximum_ruby_version.finite?
+      note = "Maximum Ruby version: #{cop.required_maximum_ruby_version}"
```
Earlopain, line 120: This reads better to me. WDYT?

```suggestion
      note = "Required Ruby version: <= #{cop.required_maximum_ruby_version}"
```

dvandersluis: Should minimum have a `>=` then?

Earlopain: For correctness, I guess. But I think it would just be visual clutter.

If you have both an upper and lower bound I'd do it like `>=2.7, <=3.3` but in the most common case I would not add it.

dvandersluis: Updated. I'm leaving out the min+max case for now but we might have to handle that in the future.

Earlopain: There are some cops where this will be useful, so this is a welcome addition. I'll dig them up after this is merged

lib/rubocop/cop/mixin/target_ruby_version.rb
koic, line 8: The value `0` for the name `minimum_target_ruby_version` seems inappropriate. How about setting it to `2.0`, the minimum Ruby version supported for analysis?

```suggestion
        @minimum_target_ruby_version || 2.0
```

lib/rubocop/cop/mixin/target_ruby_version.rb
```diff
@@ -5,15 +5,24 @@ module Cop
...
       end
 
+      def maximum_target_ruby_version(version)
+        @maximum_target_ruby_version = version
+      end
+
       def support_target_ruby_version?(version)
-        required_minimum_ruby_version <= version
+        required_minimum_ruby_version <= version &&
+          required_maximum_ruby_version >= version
```
koic, line 25: Nits.

```suggestion
        required_minimum_ruby_version <= version && required_maximum_ruby_version >= version
```

lib/rubocop/cop/mixin/target_ruby_version.rb
dvandersluis: @koic @bbatsov I used 0 here because the idea is to just have a fixed minimum to compare against. I can do 2.0 (I understand the logic) but then the change to [cops_documentation_generator.rb](https://github.com/rubocop/rubocop/pull/13473/files#diff-664aaca7a3829e6afc7275b3aebc2ee646de27f55ca01361d7fff64e0f7f8641) needs to hardcode 2.0 as a minimum, which means there's an edge case if someone actually was to set `minimum_target_ruby_version 2.0` (which `Style/Dir` and `Style/SymbolArray`) do. Also, if we're going to show the minimum as 2.0 instead, it'd only show up in the documentation when a cop class is extended with `TargetRubyVersion`, which is a minority of cops, so the documentation would be inconsistent.

This would only affect cops that have a maximum version set but no minimum (which at the moment is only this one).

So IMO we need a value that allows comparison but will not affect how the documentation is rendered. If 0 doesn't work for you, how about `-Infinity`?

lib/rubocop/cop/mixin/target_ruby_version.rb
dvandersluis: Updated.

lib/rubocop/cop/mixin/target_ruby_version.rb
bbatsov: I'm OK with 0 as well - just add comment explaining this.

lib/rubocop/cop/mixin/target_ruby_version.rb
```diff
@@ -8,12 +8,23 @@ def required_minimum_ruby_version
...
         @minimum_target_ruby_version = version
       end
 
+      def maximum_target_ruby_version(version)
+        @maximum_target_ruby_version = version
+      end
+
       def support_target_ruby_version?(version)
-        required_minimum_ruby_version <= version
+        min = required_minimum_ruby_version || -Float::INFINITY
```
bbatsov, line 24: I think we need to have some comment here explaining this bit of code as it's not super obvious what we're aiming for - no matter if it's using 0 or something else.

dvandersluis: Makes sense, I added a comment, please let me know if it's clear.

PR #13474, opened 2024-11-19
Add new `Style/ItAssignment` cop to detect assignments to `it` inside blocks

dvandersluis:
Adds a new `Style` cop to detect reassigning `it` inside a block where it could be the default block parameter in ruby 3.4. Similar to `Lint/NumberedParameterAssignment`, but this is not a `Lint` cop because Ruby does not warn for reassigning `it`. However, users still might wish to avoid it so that when `it` is used in a block as a lvar it has a singular meaning.

This cop is not limited to target ruby version >= 3.4 in order to allow users to correct offenses prior to upgrading ruby.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/it_assignment.rb
```diff
@@ -0,0 +1,46 @@
...
+      # cause confusion if `it` is used as a block parameter elsewhere.
+      #
+      # @example
+      #   # bad
+      #   foo { it = 5 }
+      #
+      #   # good
+      #   foo { bar = 5 }
+      #
+      #   # good - `it` is not a parameter in a block with explicit parameters
```
Earlopain, line 19: I think it makes sense to ban these as well for consistency, the second paragraph you wrote applies here too

dvandersluis: Yeah I thought about it, I'm not sure if it makes sense to have conservative/aggressive styles (right now it only affects `it` in blocks that it can be a parameter), or just do it across the board always.

At the very least these examples should likely be "okay" and not "good".

Earlopain: I would not add styles initially, can always be added later when there is demand for it. I'd choose aggressive as the only option for now but don't feel too strongly about it.

dvandersluis: You're right, I've updated the cop to consider numblocks and blocks with parameters as well. Thanks!

koic:
`_1` is a reserved name that results in a clear syntax error, which justifies the existence of a cop for it.

```console
$ ruby -e "_1 = 42"
-e: -e:1: syntax error found (SyntaxError)
> 1 | _1 = 42
    | ^~ _1 is reserved for numbered parameters
  2 |
```

On the other hand, even if the `it` block parameter is introduced, assigning to `it` represents an intentional naming choice by the programmer, and it’s questionable whether restricting this has any real value.

If such a cop were to be implemented, it would seem more appropriate as a `Naming` cop rather than a `Style` cop. However, there is also doubt as to whether it is reasonable to provide a core cop for a case that seems unlikely to occur in practice.

bbatsov:
I definitely think the cop is useful. I just need a bit of time to decide what's the right department for it.

bbatsov:
I think I'm leaning towards the `Naming` department, although this one could have easily been a `Style` or even a `Lint` cop as well. @rubocop/rubocop-core Let me know if someone thinks differently. I'll give everyone a few days to respond here about the department, before I merge the PR.

bbatsov:
Probably we should also introduce a matching style guide entry as well.

koic:
The `it` parameter differs from numbered parameters in that assignments to it are syntactically valid, so making `Lint` seem overly strict. If `Naming` is not appropriate, `Style` appears to be the more reasonable choice.

bbatsov:
> The it parameter differs from numbered parameters in that assignments to it are syntactically valid, so making Lint seem overly strict.

While technically true, I still we can still make the case that assigning to `it` results in somewhat confusing code. But I do agree that `Naming` and `Style` are probably more natural choices.

dvandersluis:
To me, Naming is "how to name something" vs "avoid naming something [x]", so Style still makes more sense to me. There are some existing Naming cops that probably should be in Style already IMO (primarily those that have multiple styles, like `Naming/BlockForwarding` or `Naming/HeredocDelimiterCase`).

That being said I think the line between the two namespaces for this cop is blurry and I'm happy having it in either.

lib/rubocop/cop/naming/it_assignment.rb
```diff
@@ -0,0 +1,45 @@
...
+      #   # good - use a different variable name
+      #   foo { var = 5 }
+      #   foo { |bar| var = bar }
+      #   foo { bar = _2 }
+      class ItAssignment < Base
+        extend TargetRubyVersion
+
+        MSG = '`it` is the default block parameter; consider another name.'
+
+        minimum_target_ruby_version 3.4
```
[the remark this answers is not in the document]
dvandersluis: Sounds good, removed.

lib/rubocop/cop/naming/it_assignment.rb
```diff
@@ -0,0 +1,43 @@
...
+      #   foo { it = 5 }
+      #   foo { |bar| it = bar }
+      #   foo { it = _2 }
+      #
+      #   # good - use a different variable name
+      #   foo { var = 5 }
+      #   foo { |bar| var = bar }
+      #   foo { bar = _2 }
+      class ItAssignment < Base
+        extend TargetRubyVersion
```
koic, line 25: This mixin appears to be no longer needed.

dvandersluis: Thanks for catching that, I removed it.

dvandersluis:
@bbatsov the namespace has been changed back to `Style` now.

bbatsov (approved):

PR #13478, opened 2024-11-20
Yamllint consecutive newlines

Earlopain:
Noticed in https://github.com/rubocop/rubocop/pull/13463/files#diff-3653213e3a153018a75cc7b2536b1d0217a6d8d75188d71cf45441812f789698L2385

First checking if that file is even used by default on the CI action...


Earlopain:
Looks ok: https://github.com/rubocop/rubocop/actions/runs/11939490398/job/33280120440

The `yamllint_comment` option seems pretty useless. It's supposed to post a comment on lint failure but forks don't have access to those.

.yamllint.yml
```diff
@@ -7,3 +7,5 @@ rules:
   line-length: disable
   truthy:
     check-keys: false
+  empty-lines:
+    max: 1
```
koic, line 11: That's just my two cents. Can you move the `empty-lines` config to its position in ASCII order, which would be right after the `document-start` config?
https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration

Earlopain: Yeah, of course

koic:
Thanks!

PR #13484, opened 2024-11-21
Add new `Style/FileTouch` cop

### LINKED ISSUE #13482: New cop to detect usage of File.open with empty block

I've encountered this piece of code recently
```ruby
File.open(filename, 'a') {}
```
which is used to create a file if it doesn't exist already. The problem with this is that:
- [Lint/EmptyBlock](https://docs.rubocop.org/rubocop/cops_lint.html#lintemptyblock) registers an offense, but removing the block is not desired because the file won't be closed then
- the intention of the code isn't entirely clear at first glance

I'd like to implement a new cop (I believe it would fit under `Style/` department) to register this as an offense and autocorrect to 
```ruby
FileUtils.touch(filename)
```
An offense would also be registered for access mode `a+`.

Note that the suggested autocorrection isn't a 1-for-1 replacement because unlike `File.open` it updates atime and mtime, so it would probably be unsafe autocorrection.

This cop would not apply for access modes `w` and `w+` since they truncate the file.

## Additional context

A [crude search](https://github.com/search?q=%2Ffile%5C.open.%2B%5Cba%5C%2B%3F.%5C%29+%5C%7B%5C%7D%2F+language%3ARuby+&type=code&p=1) on GH shows some results which would be registered as offenses with this cop. At least [one place](https://github.com/joeyates/imap-backup/blob/ff09698d25e7033b1a2d8c182e95ed330572772a/lib/imap/backup/serializer/mbox.rb#L109-L113) tries to use `File.open` to update the file's timestamp (which would work as expected with `FileUtils.touch`).

lovro-bikic:
Closes #13482.

Adds new `Style/FileTouch` cop to detect usage of `File.open` in append mode with empty block and autocorrect it to `FileUtils.touch`.

Before:
```ruby
File.open(filename, 'a') {}
```

After:
```ruby
FileUtils.touch(filename)
```

Recognized access modes are `a`, `a+`, `ab`, `a+b`, `at`, `a+t`.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/file_touch.rb
```diff
@@ -0,0 +1,59 @@
...
+      #
+      class FileTouch < Base
+        extend AutoCorrector
+
+        MSG = 'Use `FileUtils.touch(%<argument>s)` instead of `File.open` in ' \
+              'append mode with empty block.'
+
+        RESTRICT_ON_SEND = %i[open].freeze
+
+        APPEND_FILE_MODES = %w[a a+ ab a+b at a+t].to_set.freeze
```
lovro-bikic, line 28: Technically, append mode can be defined with e.g. `File::APPEND`, but I followed this example here: https://github.com/rubocop/rubocop/blob/e751b382ac83c622029ec5cb8ecf02893cb216ce/lib/rubocop/cop/style/file_write.rb#L46

lib/rubocop/cop/style/file_touch.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for usage of `File.open` in append mode with empty block.
+      #
+      # @safety
```
Earlopain, line 8: Just leaving this here as quick feedback. `FileUtils`, unlike `File`, is not loaded by default and must be required:

```
$ ruby -e 'FileUtils'
-e:1:in `<main>': uninitialized constant FileUtils (NameError)

FileUtils
^^^^^^^^^
Did you mean?  FileTest
```

Not sure if there is a cop that inserts a require somewhere.

lovro-bikic: True, but this then affects [Lint/NonAtomicFileOperation](https://docs.rubocop.org/rubocop/cops_lint.html#lintnonatomicfileoperation) as well 🤔 

I think the reason for its usage might be that in a bundler context `FileUtils` will always be required, so it's usually considered present.

lib/rubocop/cop/style/file_touch.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Style
```
lovro-bikic, line 5: I'm wondering if this should be a Lint cop instead. [In the wild](https://github.com/search?q=%2Ffile%5C.open.%2B%5Cba%5C%2B%3F.%5C%29+%5C%7B%5C%7D%2F+language%3ARuby+&type=code&p=1), I've seen `File.open` used to make sure a file exists or [update its timestamps](https://github.com/joeyates/imap-backup/blob/ff09698d25e7033b1a2d8c182e95ed330572772a/lib/imap/backup/serializer/mbox.rb#L109-L113) (much like [`touch`](https://www.gnu.org/software/coreutils/manual/html_node/touch-invocation.html) utility), the latter of which actually doesn't happen:
```ruby
FileUtils.touch('foo.tmp') # create a file

atime_1, ctime_1, mtime_1 = File.atime('foo.tmp'), File.ctime('foo.tmp'), File.mtime('foo.tmp')

File.open('foo.tmp', 'a') {}

atime_2, ctime_2, mtime_2 = File.atime('foo.tmp'), File.ctime('foo.tmp'), File.mtime('foo.tmp')

atime_1 == atime_2 && ctime_1 == ctime_2 && mtime_1 == mtime_2
# => true
```
so this could perhaps be considered erroneous code.

bbatsov:
Looks good to me. Thanks!

PR #13494, opened 2024-11-25
[Fix #13486] Add new `Style/DigChain` cop

### LINKED ISSUE #13486: New cop for chained `#dig` calls

Sometimes I see `#dig` method calls chained, either with or without safe navigation:
```ruby
x.dig(:foo).dig(:bar).dig(:baz)
x.dig(:foo, :bar).dig(:baz)
x.dig(:foo, :bar)&.dig(:baz)
```
all of which could be replaced with a single `#dig`:
```ruby
x.dig(:foo, :bar, :baz)
```

A new cop to register this as an offense and autocorrect to a single dig. It would register two or more chained digs as a single offense and autocorrect by splatting arguments of all digs into a single dig.

## Impact

A [crude search](https://github.com/search?q=%2F%5Cbdig%5Cb%5C%28.%2B%3F%5C%29%26%3F%5C.%5Cbdig%5Cb%2F+language%3ARuby&type=code) on GH (there are some false positives in the results) shows that there's code which would benefit from this cop. 

## Incompatibility with Style/SingleArgumentDig

This cop would be incompatible with [Style/SingleArgumentDig](https://docs.rubocop.org/rubocop/cops_style.html#stylesingleargumentdig) when there's a single argument. I think in such a case, this cop should have precedence because autocorrecting `x.dig(:foo).dig(:bar, :baz)` to `x.dig(:foo, :bar, :baz)` looks cleaner (in my opinion) than `x[:foo].dig(:bar, :baz)`.

## Performance benefit

This cop would also yield a performance benefit: a simple benchmark shows that using a single dig is faster than chained digs (the gap becoming bigger with the length of the chain and use of safe navigation):

<details>
  <summary>Benchmark</summary>

```ruby
require 'benchmark/ips'

hash = { foo: { bar: { baz: {} } } }

Benchmark.ips do |x|
  x.report('single') { hash.dig(:foo, :bar, :baz) }
  x.report('chained') { hash.dig(:foo).dig(:bar).dig(:baz) }
  x.report('chained safe nav') { hash.dig(:foo)&.dig(:bar)&.dig(:baz) }

  x.compare!
end
```

```
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [x86_64-darwin23]
              single     1.072M i/100ms
             chained   798.267k i/100ms
    chained safe nav   771.547k i/100ms
              single     10.702M (± 0.8%) i/s   (93.44 ns/i) -     53.578M in   5.006930s
             chained      7.988M (± 1.7%) i/s  (125.19 ns/i) -     40.712M in   5.098312s
    chained safe nav      7.586M (± 3.2%) i/s  (131.82 ns/i) -     38.577M in   5.091096s

Comparison:
              single: 10701554.1 i/s
             chained:  7987553.2 i/s - 1.34x  slower
    chained safe nav:  7586273.5 i/s - 1.41x  slower
```
</details>

## Additional context

Part of the reason I suggested this cop is because chained digs are sometimes broken up by a call to `first` (which I described [here](https://github.com/rubocop/rubocop/issues/13487)), so maybe it makes sense for it to be a single cop instead of two cops? I didn't suggest that initially because I'm not sure if it makes sense to have one more powerful cop, or separate cops which can work in tandem.

dvandersluis:
Adds a cop to check for subsequent `dig` calls within a method chain that could be combined into a single `dig` call. Only `dig` calls with valid parameters (not a hash, kwargs, anonymous kwargs splat, anonymous block). Works with `...` unless a syntax error would be added by combining calls (ie. `...` has to be the last argument and only appear once).

Also updates `Style/SingleArgumentDig` to not register on chained `dig` calls that will be registered by `Style/DigChain`.

Fixes #13486.

-----------------

* [x] Added tests.


bbatsov:
Looks solid to me!

PR #13495, opened 2024-11-25
Fix false positives for `Style/HashExcept` cop

lovro-bikic:
Fixes false positive for `Style/HashExcept` cop when using `reject`/`!include?`, `reject`/`!in?` or `select/!exclude?` combinations. Adds new test cases which fail on master.

The tests that have been changed in this patch weren't correct and actually produced the inverse result; for example, the first updated test in this patch expected that
```ruby
{foo: 1, bar: 2, baz: 3}.reject { |k, v| !%i[foo bar].include?(k) }
# => {:foo=>1, :bar=>2}
```
is auto-corrected to
```ruby
{foo: 1, bar: 2, baz: 3}.except(:foo, :bar)
# => {:baz=>3}
```
when this shouldn't have been an offense for this cop since the code is equivalent to:
```ruby
{foo: 1, bar: 2, baz: 3}.slice(:foo, :bar)
```
<br />

This PR fixes these combinations:
```ruby
h.reject { |k, v| !k.in?(X) }
h.reject { |k, v| !X.include?(k) }
h.select { |k, v| !X.exclude?(k) }
```
Previously they would register as offenses, now they don't since they're functionally equivalent to `h.slice(*X)`.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/hash_except.rb
+++ b/lib/rubocop/cop/style/hash_except.rb
@@ -23,9 +23,20 @@ module RuboCop
       #   {foo: 1, bar: 2, baz: 3}.reject {|k, v| k == :bar }
       #   {foo: 1, bar: 2, baz: 3}.select {|k, v| k != :bar }
       #   {foo: 1, bar: 2, baz: 3}.filter {|k, v| k != :bar }
+      #   {foo: 1, bar: 2, baz: 3}.reject {|k, v| k.eql?(:bar) }
+      #
+      #   # bad
       #   {foo: 1, bar: 2, baz: 3}.reject {|k, v| %i[bar].include?(k) }
       #   {foo: 1, bar: 2, baz: 3}.select {|k, v| !%i[bar].include?(k) }
       #   {foo: 1, bar: 2, baz: 3}.filter {|k, v| !%i[bar].include?(k) }
+      #
+      #   # bad
+      #   {foo: 1, bar: 2, baz: 3}.reject {|k, v| !%i[bar].exclude?(k) }
+      #   {foo: 1, bar: 2, baz: 3}.select {|k, v| %i[bar].exclude?(k) }
+      #
+      #   # bad
+      #   {foo: 1, bar: 2, baz: 3}.reject {|k, v| k.in?(%i[bar]) }
+      #   {foo: 1, bar: 2, baz: 3}.select {|k, v| !k.in?(%i[bar]) }
       #
       #   # good
       #   {foo: 1, bar: 2, baz: 3}.except(:bar)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/hash_except.rb
```diff
@@ -129,11 +129,15 @@ def semantically_except_method?(send, block)
         end
 
         def included?(negated, body)
-          body.method?('include?') || body.method?('in?') || (negated && body.method?('exclude?'))
+          if negated
+            body.method?('exclude?')
+          else
+            body.method?('include?') || body.method?('in?')
+          end
```
lovro-bikic, line 136: The error was here, `body.method?('include?') || body.method?('in?')` were evaluated regardless of `negated` param.

For `reject`/`!include?` combination the following happened in `semantically_except_method?` method (it's above this method):
1. `negated = true` (because of `!` in front of `include?`)
2. `:reject === send.method_name` (so line 123 is evaluated)
3. `included?` returns `true` because `body.method?('include?')` is `true`

spec/rubocop/cop/style/hash_except_spec.rb
```diff
@@ -390,6 +390,12 @@
           RUBY
         end
 
+        it 'does not register an offense when using `reject` and calling `!key.in?` method with symbol array' do
+          expect_no_offenses(<<~RUBY)
+            {foo: 1, bar: 2, baz: 3}.reject { |k, v| !k.in?(%i[foo bar]) }
+          RUBY
+        end
```
lovro-bikic, line 397: I added this test because `reject/!in?` combination wasn't tested before.

lib/rubocop/cop/style/hash_except.rb
```diff
@@ -104,8 +103,6 @@ def bad_method?(method_name, block)
             end
           else
             bad_method_with_poro?(block) do |key_arg, send_node|
-              return false if method_name == :reject && block.body.method?(:!)
```
lovro-bikic, line 107: This was removed because it's part of the semantic check (`semantically_except_method?`). Originally added in https://github.com/rubocop/rubocop/pull/13015.

spec/rubocop/cop/style/hash_except_spec.rb
```diff
@@ -524,6 +514,12 @@
           RUBY
         end
 
+        it 'does not register an offense when using `select` and calling `!exclude?` method with symbol array' do
+          expect_no_offenses(<<~RUBY)
+            {foo: 1, bar: 2, baz: 3}.select { |k, v| !%i[foo bar].exclude?(k) }
+          RUBY
+        end
```
lovro-bikic, line 521: This test is also failing on master.

bbatsov:
I think you also need to update the cop's examples.

lovro-bikic:
> I think you also need to update the cop's examples.

Updated!

bbatsov:
Thanks!

PR #13496, opened 2024-11-26
Remove unnecessary `include IgnoredNode`

dvandersluis:
`IgnoredNode` is already included in `RuboCop::Cop::Base`, so including it again inside a cop is redundant.


bbatsov:
Good catch!

PR #13497, opened 2024-11-26
Fix infinite loop error for `Style/IfWithSemicolon`

koic:
This PR fixes infinite loop error for `Style/IfWithSemicolon` when using nested if/;/end in if body.

```console
$ cat example.rb
if cond1; foo + if cond2; bar
  else
  end
end
$ bundle exec rubocop -a -d
(snip)
Parser::Source::TreeRewriter detected clobbering
/Users/koic/.rbenv/versions/3.4-dev/lib/ruby/gems/3.4.0+0/gems/parser-3.3.6.0/lib/parser/source/tree_rewriter.rb:427:
in 'Parser::Source::TreeRewriter#trigger_policy'
/Users/koic/.rbenv/versions/3.4-dev/lib/ruby/gems/3.4.0+0/gems/parser-3.3.6.0/lib/parser/source/tree_rewriter.rb:414:
in 'Parser::Source::TreeRewriter#enforce_policy'
/Users/koic/.rbenv/versions/3.4-dev/lib/ruby/gems/3.4.0+0/gems/parser-3.3.6.0/lib/parser/source/tree_rewriter/action.rb:234:
in 'Method#call'
```

The autocorrection for the nested conventional case is handled as a repeated autocorrection, so there is no impact on the user experience.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/if_with_semicolon.rb
+++ b/lib/rubocop/cop/style/if_with_semicolon.rb
@@ -23,6 +23,7 @@ module RuboCop
 
         def on_normal_if_unless(node)
           return if node.parent&.if_type?
+          return if part_of_ignored_node?(node)
 
           beginning = node.loc.begin
           return unless beginning&.is?(';')
@@ -32,6 +33,8 @@ module RuboCop
           add_offense(node, message: message) do |corrector|
             autocorrect(corrector, node)
           end
+
+          ignore_node(node)
         end
 
         private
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/if_with_semicolon.rb
```diff
@@ -14,6 +14,7 @@ module Style
       #   result = some_condition ? something : another_thing
       #
       class IfWithSemicolon < Base
+        include IgnoredNode
```
bbatsov, line 17: `IgnoredNode` is already included in `RuboCop::Cop::Base`, so including it again inside a cop is redundant.

koic: Ah, I see! I've updated this PR. Thank you.

PR #13499, opened 2024-11-26
[Fix #13096] Update `Style/BlockDelimiters` to not change braces when they are required for syntax

### LINKED ISSUE #13096: `Layout/LineLength` safe autocorrect results in invalid code
Sometimes the `Layout/LineLength` cop, which is marked as a safe autocorrect, corrects to invalid ruby code which cannot be parsed.

## Expected behavior

When running `rubocop -a foo.rb` where `foo.rb` is the following file:

```ruby
# frozen_string_literal: true

puts ['foo'] + ['bar'].map { |x| x.upcase }, 'ok'
```

And I have the following `.rubocop.yml`:

```yml
AllCops:
  NewCops: enable
  TargetRubyVersion: 3.3

Layout/LineLength:
  Max: 48
```

I expect it to not result in invalid code.

## Actual behavior

It produces the following file:

```ruby
# frozen_string_literal: true

puts ['foo'] + ['bar'].map do |x|
  x.upcase
end, 'ok'
```

Running `ruby foo.rb` produces the following output:
```
foo.rb: --> foo.rb
unexpected write target
> 3  puts ['foo'] + ['bar'].map do |x|
> 5  end, 'ok'

foo.rb:5: syntax error, unexpected ',', expecting end-of-input (SyntaxError)
end, 'ok'
```

## Steps to reproduce

Follow the above steps in the "Expected behavior section"

## RuboCop version

```
> rubocop -V
1.65.1 (using Parser 3.3.4.1, rubocop-ast 1.32.0, running on ruby 3.3.4) [arm64-darwin23]
```

dvandersluis:
When a method argument has a block, braces are required, otherwise additional arguments will be a syntax error. Additionally, even without additional arguments, the binding of the block will change when braces are changed to do.

For instance, this will no longer register an offense as the braces are required:

```ruby
puts ['foo'] + ['bar'].map { |x|
  x.upcase }, 'ok'
```

Fixes #13096.

-----------------

* [x] Added tests.


lib/rubocop/cop/style/block_delimiters.rb
```diff
@@ -342,16 +342,22 @@ def with_block?(node)
           node.respond_to?(:block_node) && node.block_node
         end
 
-        def get_blocks(node, &block)
+        def get_blocks(node, &block) # rubocop:disable Metrics/CyclomaticComplexity
```
koic, line 345: `Metrics/CyclomaticComplexity` may be resolved at some point in the future.  e.g.
https://github.com/rubocop/rubocop/commit/ac98713904b3e6f10605f00786e85bc3b3d0aa15

To avoid polluting `git blame` when that happens, it would be better to place annotations before and after the method, rather than inline.

```ruby
# rubocop:disable Metrics/CyclomaticComplexity
def get_blocks(node, &block)
  # impl...
end
# rubocop:enable Metrics/CyclomaticComplexity
```

dvandersluis: Updated.

koic:
Thanks!

PR #13507, opened 2024-11-27
Add `Style/HashSlice` cop

lovro-bikic:
Adds `Style/HashSlice` to complement [`Style/HashExcept`](https://docs.rubocop.org/rubocop/cops_style.html#stylehashexcept).

Since the cops are complements of each other and share most of the logic, I extracted common functionality of `Style/HashExcept` to a mixin that's reused in `Style/HashSlice`. Let me know if this should be split into two PRs.

This cop will register code such as:
```ruby
{foo: 1, bar: 2, baz: 3}.filter { |k, v| [:foo, :bar].include?(k) }
```
as an offense and autocorrect to:
```ruby
{foo: 1, bar: 2, baz: 3}.slice(:foo, :bar)
```

-----------------

* [ ] Added tests.


lib/rubocop/cop/style/hash_except.rb
```diff
@@ -42,162 +42,26 @@ module Style
       #   {foo: 1, bar: 2, baz: 3}.except(:bar)
       #
       class HashExcept < Base
-        include RangeHelp
+        include HashSliceExcept
         extend TargetRubyVersion
         extend AutoCorrector
 
         minimum_target_ruby_version 3.0
```
lovro-bikic, line 49: `Hash#slice` was [added in Ruby 2.5](https://www.ruby-lang.org/en/news/2017/12/25/ruby-2-5-0-released/), I don't know if I should add `minimum_target_ruby_version 2.5` since currently required Ruby version is >= 2.7?

lib/rubocop/cop/style/hash_slice.rb
```diff
@@ -0,0 +1,62 @@
...
+
+module RuboCop
+  module Cop
+    module Style
+      # Checks for usages of `Hash#reject`, `Hash#select`, and `Hash#filter` methods
+      # that can be replaced with `Hash#slice` method.
+      #
+      # For safe detection, it is limited to commonly used string and symbol comparisons
+      # when using `==`.
+      # And do not check `Hash#delete_if` and `Hash#keep_if` to change receiver object.
```
lovro-bikic, line 11: This part was copied from `HashExcept` docs, but I don't fully understand what this sentence means, it doesn't feel clear enough. Perhaps the docs should be updated?

lib/rubocop/cop/style/hash_except.rb
lovro-bikic: > RuboCop can still inspect code all the way down to Ruby 2.0

That's impressive. Thanks, updated.

lib/rubocop/cop/style/hash_slice.rb
lovro-bikic: 👍 I updated the wording in both cops, thanks.

lib/rubocop/cop/mixin/hash_slice_except.rb
```diff
@@ -0,0 +1,160 @@
...
+              _ {:== :!= :eql? :in? :include? :exclude?} _)
+            (send
+              $(send
+                _ {:== :!= :eql? :in? :include? :exclude?} _) :!)
+            })
+      PATTERN
+
+      def on_send(node)
+        raise NotImplementedError
+      end
```
dvandersluis, line 46: Can you remove this method? Mixins aren't abstract classes, and this exception doesn't have much value (and also if this mixin is later included in another cop class that doesn't define `on_send` it'll raise for every `send` node).

lib/rubocop/cop/mixin/hash_slice_except.rb
```diff
@@ -0,0 +1,160 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # Common functionality for Style/HashExcept and Style/HashSlice
+    module HashSliceExcept
```
dvandersluis, line 6: Maybe something like `HashSubset` would be a better name? Also could you elaborate on the top comment to explain what this mixin does?

lib/rubocop/cop/mixin/hash_slice_except.rb
```diff
@@ -0,0 +1,160 @@
...
+        when :reject
+          body.method?('==') || body.method?('eql?') || included?(negated, body)
+        when :select, :filter
+          body.method?('!=') || not_included?(negated, body)
+        end
+      end
+
+      def semantically_slice_method?(send)
+        semantically_except_method?(send) == false
+      end
```
dvandersluis, line 94: Is there any reason `semantically_except_method?(send)` would return nil? Can this just be `!semantically_except_method?(send)`?

In other words, at the point where this is called, can we guarantee that the code should either become a `slice` or `except` call?

lib/rubocop/cop/mixin/hash_slice_except.rb
```diff
@@ -0,0 +1,160 @@
...
+        return unless bad_method?(block)
+
+        except_key = except_key(block)
+        return if except_key.nil? || !safe_to_register_offense?(block, except_key)
+
+        [offense_range(node), except_key_source(except_key)]
+      end
+
+      # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      def bad_method?(block)
```
dvandersluis, line 61: I know this is copied from `HashExcept` but I'd like to try renaming `bad_method?` to something more descriptive. By extension, this goes for the two matchers as well (`bad_method_with_active_support?` and `bad_method_with_poro?`). 

Perhaps `extracts_hash_subset?`? @Earlopain any ideas?

lib/rubocop/cop/mixin/hash_slice_except.rb
lovro-bikic: Sure, but I added this because otherwise I get an offense for `InternalAffairs/UselessRestrictOnSend` and I'm not sure what's the best way to handle that?

image

dvandersluis: There are only two other mixins that use `RESTRICT_ON_SEND`, and both of them actually define `on_send` as well. I think there's a few options here:

1. Inline disable `InternalAffairs/UselessRestrictOnSend` here.
2. Update this class to implement `on_send` as well, and further refactor out `HashSlice` and `HashExcept`.  This should be plausible because other than `return unless semantically_slice_method?(node)` or `return unless semantically_except_method?(node)`, the `on_send` method in both cops are essentially the same.

lib/rubocop/cop/mixin/hash_slice_except.rb
dvandersluis: I noticed that as well but was going to tackle it in a separate refactor after this had gotten merged to keep from scope creep. If you want to tackle the refactor, I think it'd be best to actually combine the two pattern matchers which are almost identical except for which methods are allowed, and handle that within what is currently `bad_method?` instead. I think we need tests for the Rails methods when ActiveSupportExtensions is not enabled to be added first too.

lovro-bikic: Here's my attempt at this: https://github.com/rubocop/rubocop/pull/13517

lovro-bikic: That PR has been merged, and I think it addressed all comments in this thread, so I'm resolving this comment.

lib/rubocop/cop/mixin/hash_slice_except.rb
lovro-bikic: I can't remember anymore my exact reasoning for writing the method like this, but with the refactor done in https://github.com/rubocop/rubocop/pull/13517, everything that's not `semantically_except_method?` must be `semantically_slice_method?` (though that could have been the case previously as well), so I've updated this method to `!semantically_except_method?`.

lib/rubocop/cop/mixin/hash_slice_except.rb
lovro-bikic: Thanks, that's a better name indeed. I updated the top comment as well.

lovro-bikic:
The above PR has been merged, marking this as ready for review again.

lib/rubocop/cop/mixin/hash_slice_except.rb
lovro-bikic: I've tried implementing the second option, let me know what you think.

bbatsov (approved):
The PR looks good to me, but I'll leave it to @dvandersluis to merge it, as he was the one who gave the most feedback on it so far.

dvandersluis:
Nice refactoring and new cop, thank you!

Issue #13510, opened 2024-11-28
Robustly handle `csend` nodes across all cops


tdeo:
## Is your feature request related to a problem? Please describe.

Some cops don't properly support `csend` nodes, i.e. `Style/LambdaCall` reports no offense on `arg?.(1)` while it would on `arg.(1)`.

## Describe the solution you'd like

I believe we could leverage an `InternalAffairs/ForgottenCsendSupport` cop that checks that all cops that define `on_send` also define `on_csend` (through a `def`, `alias` or `alias_method` technique) to ensure `&.` calls directly bypass calls for regular send calls.

## Describe alternatives you've considered

Keep adding incremental support when detected, by manually auditing cops.

## Additional context

I came across https://github.com/rubocop/rubocop/pull/6749 which implement support for `csend` nodes for a large number of cops, but still doesn't guarantee exhaustivity.

My `InternalAffairs` cop suggestion will likely produce some false positives - typically I believe the `TopLevelMethodDefinition` has no scenario today where it would need to support `csend` node. This could cause some useless performance hit (although reasonable is the cop implements `RESTRICT_ON_SEND`) or code readability - the tradeoff with exhaustivity is obviously subjective.

dvandersluis:
I've looked into this before but haven't moved forwards with it because it is not guaranteed that a cop that implements `on_send` also should implement `on_csend`. For instance, `on_csend` would not make sense for `Lint/AmbiguousOperatorPrecedence`, since that cop is looking for method calls without receivers, so no `&.` call would ever be relevant.

That being said, you're right that it should be handled more consistently.

I've opened #13512 to add `csend` support to `Style/LambdaCall`. I've also opened #13513 to add support for `csend` to the new cop generator (since it defines `on_csend` might as well add an alias by default as well).

dvandersluis:
I've now gone through all the cops that have `on_send` defined without `on_csend`, and [updated them](https://github.com/rubocop/rubocop/pulls?q=is%3Apr+author%3Advandersluis+%22Add+support+for+safe+navigation+to%22+in%3Atitle).

I've also opened #13719 to add an InternalAffairs cop to prevent this from happening in the future.

tdeo:
Thanks - happy to hear it also made its way into InternalAffairs so it's hopefully easy to catch in plugins

PR #13513, opened 2024-11-28
Update the new cop generator to suggest supporting `on_csend` along with `on_send`

dvandersluis:
Follows #13510. 

In order to try to encourage supporting `csend` nodes along with `send` nodes, I updated the generator to add an alias for the provided `def on_send` to `on_csend`. Of course, the user can remove it if its unnecessary, but it might prompt new cop developers to be thinking about safe navigation when they're working on method calls.

### DIFF
--- a/lib/rubocop/cop/generator.rb
+++ b/lib/rubocop/cop/generator.rb
@@ -71,11 +71,17 @@ module RuboCop
                   (send nil? :bad_method ...)
                 PATTERN
 
+                # Called on every `send` node (method call) while walking the AST.
+                # TODO: remove this method if inspecting `send` nodes is unneeded for your cop.
+                # By default, this is aliased to `on_csend` as well to handle method calls
+                # with safe navigation, remove the alias if this is unnecessary.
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/generator.rb
```diff
@@ -71,11 +71,17 @@ class %<cop_name>s < Base
                   (send nil? :bad_method ...)
                 PATTERN
 
+                # Called on every `send` node (method call) while walking the AST.
+                # TODO: remove this method if inspecting `send` nodes is unneeded for your cop.
+                # By default, this is aliased to `on_csend` as well to handle method calls
+                # with safe navigation, remove the alias if this is unnecessary.
+                # If kept, ensure your tests cover safe navigation as well!#{'  '}
```
bbatsov, line 78: What's with the `#{}` here?

dvandersluis: It's `Layout/TrailingWhitespace` being over-zealous in a heredoc and I missed it. Updated now, thanks!

PR #13515, opened 2024-11-28
Specify `maximum_target_ruby_version` for a handful of cops, document it

Earlopain:
Followup to https://github.com/rubocop/rubocop/pull/13473

* `Lint/CircularArgumentReference`: Invalid syntax on modern rubies
* `Lint/NonDeterministicRequireOrder`: Skipped in node handlers
* `Lint/UselessElseWithoutRescue`: Invalid syntax on modern rubies
* `Security/YAMLLoad`: Skipped in node handler

I left `Lint/RefinementImportMethods` alone and just added the removal timeframe in the docs.
Probably it could get an upper bound but docs currently don't handle both min/max at the same time.

-----------------

* [ ] Added tests.

### DIFF
--- a/docs/modules/ROOT/pages/development.adoc
+++ b/docs/modules/ROOT/pages/development.adoc
@@ -378,12 +378,12 @@ Some cops apply changes that only apply in particular contexts, such as if the u
 
 ==== Requiring a minimum Ruby version
 
-If your cop uses new Ruby syntax or standard library APIs, it should only autocorrect if the user has the target Ruby version, which you set with https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/TargetRubyVersion#minimum_target_ruby_version-instance_metho[`TargetRubyVersion#minimum_target_ruby_version`].
+If your cop uses new Ruby syntax or standard library APIs, it should only register offenses if the user has the proper target Ruby version, which you can require with https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/TargetRubyVersion#minimum_target_ruby_version-instance_method[`TargetRubyVersion#minimum_target_ruby_version`].
 
 For example, the `Performance/SelectMap` cop requires Ruby 2.7, which introduced `Enumerable#filter_map`:
 
 ```ruby
-module RuboCop::Cop::Performance::SelectMap < Base
+class RuboCop::Cop::Performance::SelectMap < Base
   extend TargetRubyVersion
 
   minimum_target_ruby_version 2.7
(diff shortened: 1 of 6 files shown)


docs/modules/ROOT/pages/development.adoc
```diff
@@ -378,7 +378,7 @@ Some cops apply changes that only apply in particular contexts, such as if the u
 
 ==== Requiring a minimum Ruby version
 
-If your cop uses new Ruby syntax or standard library APIs, it should only autocorrect if the user has the target Ruby version, which you set with https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/TargetRubyVersion#minimum_target_ruby_version-instance_metho[`TargetRubyVersion#minimum_target_ruby_version`].
+If your cop uses new Ruby syntax or standard library APIs, it should only register offenses if the user has the proper target Ruby version, which you can require with https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/TargetRubyVersion#minimum_target_ruby_version-instance_method[`TargetRubyVersion#minimum_target_ruby_version`].
```
Earlopain, line 381: Typo in the docs link (missing `d` from `method` at the end.

I also slightly tweaked it. It seems weird to me that it would talk about autocorrect when the cop is just not doing anything at all. It's elaborated on further down but it seems unnecessary

lib/rubocop/cop/lint/useless_else_without_rescue.rb
```diff
@@ -25,8 +25,12 @@ module Lint
       #     do_something_else
       #   end
       class UselessElseWithoutRescue < Base
+        extend TargetRubyVersion
+
         MSG = '`else` without `rescue` is useless.'
 
+        maximum_target_ruby_version 2.5
```
dvandersluis, line 32: The comment on line 8 is probably not necessary since this will add a ruby version note to the docs.

dvandersluis:
Nice job! Thanks for adding the documentation section on `maximum_target_ruby_version`!

> Probably it could get an upper bound but docs currently don't handle both min/max at the same time.

Well that happened quickly :joy: We should probably handle that next.

lib/rubocop/cop/lint/useless_else_without_rescue.rb
Earlopain: They all have comments similar to this. I think those are fine to leave as rationale why the cop is now irrelevant.

dvandersluis: True, fair enough.

Earlopain:
> Re: Lint/RefinementImportMethods

I'm not 100% on that one. The cop can still detect the code it is looking for, it would just be a runtime error instead of a warning for Ruby. I guess there may be some sort of value in just letting it run? Copying code, bumping multiple version, that sort of thing. But that seems like a bit of a stretch.

docs/modules/ROOT/pages/development.adoc
```diff
@@ -392,7 +392,23 @@ module RuboCop::Cop::Performance::SelectMap < Base
...
+This cop won't register offenses on Ruby 2.6 or older.
+
+==== Requiring a maximum Ruby version
+
+Mirroring `minimum_target_ruby_version`, you can also specify a maximum Ruby version your cop should analyze.
+
+For example, the `Lint/CircularArgumentReference` cop only runs when analyzing code for Ruby before 2.7. The code it is looking for can never be written in more recent Rubies, it would be a syntax error:
+
+```ruby
+module RuboCop::Cop::Lint::CircularArgumentReference < Base
```
koic, line 404:
```suggestion
class RuboCop::Cop::Lint::CircularArgumentReference < Base
```

Earlopain: Bad copy-paste, I fixed both

PR #13517, opened 2024-11-28
Fix `Style/HashExcept` to recognize safe navigation with Active Support methods

lovro-bikic:
Fixes `Style/HashExcept` to recognize offenses when safe navigation is used and `ActiveSupportExtensionsEnabled` config option is enabled.

Previously, this code:
```ruby
{foo: 1, bar: 2, baz: 3}&.reject { |k, v| k.in?(%i[foo bar]) }
```
wouldn't register as offense (but without safe nav it would). Now it registers and autocorrects to:
```ruby
{foo: 1, bar: 2, baz: 3}&.except(:foo, :bar)
```

Additional context for the refactoring done in this PR: https://github.com/rubocop/rubocop/pull/13507#discussion_r1861491315

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/style/hash_except.rb
+++ b/lib/rubocop/cop/style/hash_except.rb
@@ -10,8 +10,10 @@ module RuboCop
       # (`Hash#except` was added in Ruby 3.0.)
       #
       # For safe detection, it is limited to commonly used string and symbol comparisons
-      # when used `==`.
-      # And do not check `Hash#delete_if` and `Hash#keep_if` to change receiver object.
+      # when using `==` or `!=`.
+      #
+      # This cop doesn't check for `Hash#delete_if` and `Hash#keep_if` because they
+      # modify the receiver.
       #
       # @safety
       #   This cop is unsafe because it cannot be guaranteed that the receiver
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/style/hash_except.rb
```diff
@@ -187,12 +175,11 @@ def decorate_source(value)
...
         def except_key(node)
-          key_argument = node.argument_list.first.source
-          body = extract_body_if_negated(node.body)
+          key_arg = node.argument_list.first.source
+          body, = extract_body_if_negated(node.body)
           lhs, _method_name, rhs = *body
-          return if [lhs, rhs].map(&:source).none?(key_argument)
 
-          [lhs, rhs].find { |operand| operand.source != key_argument }
+          lhs.source == key_arg ? rhs : lhs
```
lovro-bikic, line 182: Since the node pattern now ensures that key is used in the block body, it must be either the receiver or the argument.

spec/rubocop/cop/style/hash_except_spec.rb
```diff
@@ -225,6 +261,18 @@
       RUBY
     end
 
+    it 'does not register an offense when using more than two block arguments' do
+      expect_no_offenses(<<~RUBY)
+        {foo: 1, bar: 2, baz: 3}.reject { |k, v, o| k == :bar }
+      RUBY
+    end
```
lovro-bikic, line 268: Technically, this is something this cop could handle:
```ruby
{foo: 1, bar: 2, baz: 3}.reject { |k, v, o| k == :bar } == {foo: 1, bar: 2, baz: 3}.except(:bar)
# => true
```
but I think it's out of scope of this PR.

spec/rubocop/cop/style/hash_except_spec.rb
```diff
@@ -225,6 +261,18 @@
...
+
+    it 'does not register an offense when calling `include?` method without a param' do
+      expect_no_offenses(<<~RUBY)
+        {foo: 1, bar: 2, baz: 3}.reject { |k, v| %i[foo bar].include? }
+      RUBY
+    end
```
lovro-bikic, line 274: This kind of code will break at runtime with ArgumentError, but I've added this test because the [current code](https://github.com/rubocop/rubocop/blob/5b8a474ec19b6bd81c1991d9dbb43dca5706dd74/lib/rubocop/cop/style/hash_except.rb#L108-L118) uses safe nav for the receiver and the first argument.

However, I'm not sure if safe nav is required because of the way node pattern is written (i.e., there's always a receiver and an argument) and because when I remove safe nav no tests break, so this test might be useless if save nav provides no value in current code.

PR #13518, opened 2024-11-29
Improve warning when an invalid foldable type is given to `CodeLengthCalculator`

dvandersluis:
The warning message previously offered `send` and `csend` as possible options, but they are not in fact valid, and will raise a warning if used. `method_call` should be offered instead.

### DIFF
--- a/spec/rubocop/cop/metrics/utils/code_length_calculator_spec.rb
+++ b/spec/rubocop/cop/metrics/utils/code_length_calculator_spec.rb
@@ -467,9 +467,12 @@ RSpec.describe RuboCop::Cop::Metrics::Utils::CodeLengthCalculator do
         end
       RUBY
 
+      warning = 'Unknown foldable type: :send. ' \
+                'Valid foldable types are: array, hash, heredoc, method_call.'
+
       expect do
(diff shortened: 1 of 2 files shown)


spec/rubocop/cop/metrics/utils/code_length_calculator_spec.rb
```diff
@@ -467,9 +467,12 @@ def test
         end
       RUBY
 
+      warning = 'Unknown foldable type: :send. ' \
+                'Valid foldable types are: array, hash, heredoc, method_call.'
```
koic, line 471: Would it be better to standardize the notation to symbols, in line with `:send`?

```suggestion
                'Valid foldable types are: :array, :hash, :heredoc, :method_call.'
```

dvandersluis: I think no (and probably not for `:send` either), because these values are strings set in the `CountAsOne` config for `Metrics/BlockLength`, etc. 

But actually, that probably indicates that this check is happening in the wrong place because “foldable types” is a concept that the user doesn’t directly engage with.

koic:
Thanks!

PR #13522, opened 2024-11-30
Fix `Lint/UnescapedBracketInRegexp` failure in case of invalid regexp

viralpraxis:
The `Lint/UnescapedBracketInRegexp` cop does not handle exceptions raised by the `regexp_parser` gem. While raising exceptions (or issuing dedicated warnings) might be useful for runtime-parsed regular expressions (e.g., `Regexp.parse(<string-literal>)`), this behavior appears to fall outside the scope of this cop.

Proposed fix: ignore regular expressions that the gem cannot parse. Alternatively, we could emit a warning, either here or in a different cop.

I could not find any discussion on this topic in [the PR](https://github.com/rubocop/rubocop/pull/13385) where this cop was implemented.

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/lint/unescaped_bracket_in_regexp.rb
+++ b/lib/rubocop/cop/lint/unescaped_bracket_in_regexp.rb
@@ -58,6 +58,9 @@ module RuboCop
             Regexp::Parser.parse(text.value)&.each_expression do |expr|
               detect_offenses(text, expr)
             end
+          rescue Regexp::Parser::ParserError
+            # Upon encountering an invalid regular expression,
+            # we aim to proceed and identify any remaining potential offenses.
           end
         end
 
--- a/spec/rubocop/cop/lint/unescaped_bracket_in_regexp_spec.rb
+++ b/spec/rubocop/cop/lint/unescaped_bracket_in_regexp_spec.rb
@@ -229,6 +229,23 @@ RSpec.describe RuboCop::Cop::Lint::UnescapedBracketInRegexp, :config do
           RUBY
         end
       end
+
+      context 'invalid regular expressions' do
+        %w[+ * {42}].each do |invalid_regexp|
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/lint/unescaped_bracket_in_regexp.rb
```diff
@@ -58,6 +58,8 @@ def on_send(node)
             Regexp::Parser.parse(text.value)&.each_expression do |expr|
               detect_offenses(text, expr)
             end
+          rescue Regexp::Parser::ParserError
+            nil
```
koic, line 62: Can you replace `nil` with an intentional comment?

```suggestion
            # Does not handle exceptions raised by the regexp_parser gem.
```

spec/rubocop/cop/lint/unescaped_bracket_in_regexp_spec.rb
```diff
@@ -229,6 +229,22 @@
...
+          RUBY
+
+          expect_no_offenses(<<~RUBY)
+            Regexp.#{method}("+")
+          RUBY
+
+          expect_no_offenses(<<~RUBY)
+            Regexp.#{method}("{42}")
+          RUBY
+        end
```
koic, line 246: Can you write `it 'does not register an offense ...' do` for each `expect_no_offenses`?

lib/rubocop/cop/lint/unescaped_bracket_in_regexp.rb
viralpraxis: Done.

spec/rubocop/cop/lint/unescaped_bracket_in_regexp_spec.rb
viralpraxis: Done.

PR #13528, opened 2024-12-02
Fix `Style/RedundantParentheses` cop failure in case of splatted `case` node  without condition

viralpraxis:
The following ruby code leads to an error:

```shell
rubocop --stdin bug.rb --only Style/RedundantParentheses -d <<EOF
_ = *(case when rand > 0.5 then 1 end)
EOF
```

Backtrace:

```
An error occurred while Style/RedundantParentheses cop was inspecting /bug.rb:1:5.
undefined method `begin_type?' for nil
lib/rubocop/cop/style/redundant_parentheses.rb:247:in `only_begin_arg?'
lib/rubocop/cop/style/redundant_parentheses.rb:230:in `keyword_with_redundant_parentheses?'
lib/rubocop/cop/style/redundant_parentheses.rb:139:in `find_offense_message'
lib/rubocop/cop/style/redundant_parentheses.rb:130:in `check'
lib/rubocop/cop/style/redundant_parentheses.rb:37:in `on_begin'
```

```shell
ruby-parse --legacy -e 'case when rand > 0.5 then 1 end'
(case nil
  (when
    (send
      (send nil :rand) :>
      (float 0.5))
    (int 1)) nil)
```

-----------------

* [X] Added tests.

### DIFF
--- a/spec/rubocop/cop/style/redundant_parentheses_spec.rb
+++ b/spec/rubocop/cop/style/redundant_parentheses_spec.rb
@@ -723,6 +723,16 @@ RSpec.describe RuboCop::Cop::Style::RedundantParentheses, :config do
     expect_no_offenses('x = *(y || z)')
   end
 
+  it 'accepts parentheses around `case` expression in splat' do
+    # Parentheses are redundant, but respect user's intentions for readability.
+    expect_no_offenses('x = *(case true when true then false end)')
+  end
+
+  it 'accepts parentheses around `case` expression without condition in splat' do
+    # Parentheses are redundant, but respect user's intentions for readability.
+    expect_no_offenses('x = *(case when rand > 0.5 then 1 end)')
+  end
+
   it 'accepts parentheses around logical operator in double splat' do
     # Parentheses are redundant, but respect user's intentions for readability.
     expect_no_offenses('x(**(y || z))')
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/style/redundant_parentheses_spec.rb
```diff
@@ -723,6 +723,16 @@ def foo
     expect_no_offenses('x = *(y || z)')
   end
 
+  it 'accepts parentheses around `case` operator in splat' do
```
koic, line 726: Shouldn't it be "case expression" rather than "case operator"?

```suggestion
  it 'accepts parentheses around `case` expression in splat' do
```

spec/rubocop/cop/style/redundant_parentheses_spec.rb
```diff
@@ -723,6 +723,16 @@ def foo
...
+    # Parentheses are redundant, but respect user's intentions for readability.
+    expect_no_offenses('x = *(case true when true then false end)')
+  end
+
+  it 'accepts parentheses around `case` operator without condition in splat' do
```
koic, line 731: Ditto.

```suggestion
  it 'accepts parentheses around `case` expression without condition in splat' do
```

spec/rubocop/cop/style/redundant_parentheses_spec.rb
viralpraxis: Agreed, fixed

koic:
Thanks!

PR #13529, opened 2024-12-02
Update `rake new_cop` to handle `InternalAffairs` cops

dvandersluis:
`InternalAffairs` cops are not added to `lib/rubocop.rb` or `config.default.yml`. Instead, they are required via `lib/rubocop/cop/internal_affairs.rb`, so if generating an `InternalAffairs` cop, automatically update the correct file.

I added these changes to the rake task, not to `RuboCop::Cop::Generator`, so that they would not affect other rubocop extensions. As far as I can tell, the other extensions tend to define their own `new_cop` rake task so they should not be affected.

### DIFF
--- a/tasks/new_cop.rake
+++ b/tasks/new_cop.rake
@@ -9,12 +9,19 @@ task :new_cop, [:cop] do |_task, args|
     exit!
   end
 
+  badge = RuboCop::Cop::Badge.parse(cop_name)
   generator = RuboCop::Cop::Generator.new(cop_name)
 
   generator.write_source
   generator.write_spec
-  generator.inject_require
-  generator.inject_config
+
+  if badge.department == :InternalAffairs
+    # InternalAffairs cops are required separately, and not added to config/default.yml
+    generator.inject_require(root_file_path: 'lib/rubocop/cop/internal_affairs.rb')
+  else
+    generator.inject_require
+    generator.inject_config
+  end
 
   puts generator.todo
 end


lib/rubocop/cop/generator.rb
```diff
@@ -116,8 +116,8 @@ def on_send(node)
         '[modify] A configuration for the cop is added into ' \
         '%<configuration_file_path>s.'
 
-      def initialize(name, output: $stdout)
-        @badge = Badge.parse(name)
+      def initialize(name, output: $stdout, badge: Badge.parse(name))
```
koic, line 119: Is it necessary to change this interface? It seems redundant to make the client prepare a `Badge.parse(name)` object using the `name` argument's value.

dvandersluis: The idea was to avoid parsing the badge twice(once in the rake task, once in `Generator`). The interface doesn't change unless you choose to pass in a badge, otherwise one will be calculated when initializing `Generator` as before.

koic: Yeah, I understand the compatibility of behavior when the `badge` keyword argument is not specified. Based on that, I would like to raise the following concerns:

- The `badge` keyword argument becoming part of the published interface would require guaranteeing its support for clients in the future.
- There is a design concern regarding the redundancy between the first argument `name` and the `name` used in `badge: Badge.new(name)`. While one option might be to temporarily accept both `name` and `badge` as the first argument, I question whether it is necessary for the client to generate a `Badge` instance.
- This `Generator` is not used at runtime.

Considering these points, it is doubtful whether the added complexity is justified by the convenience it provides.

dvandersluis: @koic Fair points! I have reverted the changes to `Generator.new`.

koic:
Thanks!

PR #13533, opened 2024-12-02
[Fix #13532] Fix `Lint/NumericOperationWithConstantResult` cop not to complain about `% 1` operation

### LINKED ISSUE #13532: Lint/NumericOperationWithConstantResult incorrectly handles i % 1
--------

## Expected behavior
input: 
```ruby
if i % 1 == 0
```
output:
```ruby
if i % 1 == 0
```
## Actual behavior
input: 
```ruby
if i % 1 == 0
```
output:
```ruby
if 0 == 0
```

`i % 1 == 0` is a valid way to determine if a numeric value has a fractional value:
```ruby
> 1 % 1 == 0
=> true
> 1.0 % 1 == 0
=> true
> 1.1 % 1 == 0
=> false
```

but the autofix from `Lint/NumericOperationWithConstantResult` breaks this. 
```
$ cat foo.rb
if i % 1 == 0

end
$ rubocop -A foo.rb
Inspecting 1 file
W

Offenses:

foo.rb:3:4: W: Lint/BinaryOperatorWithIdenticalOperands: Binary operator == has identical operands.
if 0 == 0
   ^^^^^^

$ cat foo.rb
if 0 == 0

end
```

## RuboCop version

```
1.69.0 (using Parser 3.3.6.0, rubocop-ast 1.36.2, analyzing as Ruby 3.3, running on ruby 3.3.5) [arm64-darwin24]
  - rubocop-factory_bot 2.26.1
  - rubocop-performance 1.23.0
  - rubocop-rails 2.27.0
  - rubocop-rspec 3.2.0
  - rubocop-rspec_rails 2.30.0```

### PR DESCRIPTION
**Pull request by viralpraxis**
Fixes https://github.com/rubocop/rubocop/issues/13532

That's clearly not true that `%numeric == 0 (mod 1)` in general:

```
docker run -it --rm rubylang/all-ruby env ALL_RUBY_SINCE=ruby-2.2.0 ./all-ruby -e 'p(1.42 % 1)'
ruby-2.2.0          0.41999999999999993
...
ruby-3.4.0-preview2 0.41999999999999993
```

Thus it makes sense not to complain about `% 1` or `%= 1` at all.

-----------------

* [X] Added tests.

### DIFF
--- a/spec/rubocop/cop/lint/numeric_operation_with_constant_result_spec.rb
+++ b/spec/rubocop/cop/lint/numeric_operation_with_constant_result_spec.rb
@@ -23,14 +23,9 @@ RSpec.describe RuboCop::Cop::Lint::NumericOperationWithConstantResult, :config d
     RUBY
   end
 
-  it 'registers an offense when the modulo of variable and 1 is taken' do
-    expect_offense(<<~RUBY)
+  it 'does not register an offense when the modulo of variable and 1 is taken' do
+    expect_no_offenses(<<~RUBY)
       x % 1
-      ^^^^^ Numeric operation with a constant result detected.
-    RUBY
-
-    expect_correction(<<~RUBY)
-      0
     RUBY
   end
 
@@ -89,14 +84,9 @@ RSpec.describe RuboCop::Cop::Lint::NumericOperationWithConstantResult, :config d
     RUBY
   end
 
-  it 'registers an offense when the modulo of variable and 1 is taken via abbreviated assignment' do
-    expect_offense(<<~RUBY)
+  it 'does not register an offense when the modulo of variable and 1 is taken via abbreviated assignment' do
+    expect_no_offenses(<<~RUBY)
       x %= 1
-      ^^^^^^ Numeric operation with a constant result detected.
-    RUBY
-
-    expect_correction(<<~RUBY)
-      x = 0
     RUBY
   end
 
--- a/changelog/fix_lint_numeric_operation_with_constant_result_with_modulo_one.md
+++ b/changelog/fix_lint_numeric_operation_with_constant_result_with_modulo_one.md
@@ -1,0 +1,1 @@
+* [#13532](https://github.com/rubocop/rubocop/issues/13532): Fix `Lint/NumericOperationWithConstantResult` cop not to complain about `% 1` operation. ([@viralpraxis][])
--- a/lib/rubocop/cop/lint/numeric_operation_with_constant_result.rb
+++ b/lib/rubocop/cop/lint/numeric_operation_with_constant_result.rb
@@ -5,7 +5,7 @@ module RuboCop
     module Lint
       # Certain numeric operations have a constant result, usually 0 or 1.
       # Subtracting a number from itself or multiplying it by 0 will always return 0.
-      # Additionally, a variable modulo 0 or itself will always return 0.
+      # Additionally, a variable modulo itself will always return 0.
       # Dividing a number by itself or raising it to the power of 0 will always return 1.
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: spec/rubocop/cop/lint/numeric_operation_with_constant_result_spec.rb [hunk 1]
```diff
@@ -23,17 +23,6 @@
...
-  it 'registers an offense when the modulo of variable and 1 is taken' do
-    expect_offense(<<~RUBY)
-      x % 1
-      ^^^^^ Numeric operation with a constant result detected.
-    RUBY
-
-    expect_correction(<<~RUBY)
-      0
-    RUBY
-  end
```
koic, line 35: To prevent it from becoming an incorrect behavior again, can you replace `expect_offense` with `expect_no_offenses` instead of removing the test?

spec/rubocop/cop/lint/numeric_operation_with_constant_result_spec.rb
```diff
@@ -89,17 +78,6 @@
...
-  it 'registers an offense when the modulo of variable and 1 is taken via abbreviated assignment' do
-    expect_offense(<<~RUBY)
-      x %= 1
-      ^^^^^^ Numeric operation with a constant result detected.
-    RUBY
-
-    expect_correction(<<~RUBY)
-      x = 0
-    RUBY
-  end
```
koic, line 101: Ditto.

spec/rubocop/cop/lint/numeric_operation_with_constant_result_spec.rb
viralpraxis: Sure, done 👍🏻

viralpraxis:
@koic There's the same issue with `x % x`:

```
irb(main):001> f = 1; f % f
=> 0
irb(main):002> f = 1.1; f % f
=> 0.0
```

Should I address it the same way in a separate PR?

koic:
Which cop does that issue pertain to? It would be helpful if you could provide a bit more context.

viralpraxis:
For instance,

```ruby
def x
  1.618
end

p(x % x)
```

is corrected by `Lint/NumericOperationWithConstantResult` to

```ruby
def x
  1.618
end

p(0)
```

while the result is `0.0`

bbatsov:
//cc @Zopolis4

Earlopain:
> There's the same issue with x % x

The same would also be true for `x - x`. I can see why someone would write `x % 1` but is there a usecase for these two cases? Convoluted way to check if a number is a float, I guess. The result may be different but does it actually matter?

Zopolis4:
These changes make sense to me-- re `x % x`, that should still work when not applied to floats, right? I didn't think enough about floats when making the cop, evidently.

bbatsov:
> The same would also be true for x - x. I can see why someone would write x % 1 but is there a usecase for these two cases? Convoluted way to check if a number is a float, I guess. The result may be different but does it actually matter?

Same sentiment here. I'd actually discourage such convoluted checks and suggest some more readable alternatives instead.

viralpraxis:
So, what should we do here? Leave things as-is? Or disable autocorrection for *some* numeric operations? Or something else? 🙂 

Also, see related https://github.com/rubocop/rubocop/issues/13557

Earlopain:
I openend https://github.com/rubocop/rubocop/pull/13577 to discuss about it

viralpraxis:
> Sorry, I forgot about that you already made this PR when I was amending #13577. I removed both `-` and `%` handling in its entirety in that PR.

Alright, closing this one then

PR #13534, opened 2024-12-02
Fix `Layout/LineLength` cop failure in case of tabbed YARD-comment-like string

viralpraxis:
The following ruby code leads to an error:

```
rubocop -c <(echo "Layout/LineLength:\n AllowURI: true\n Max: 1") --stdin bug.rb --only Layout/LineLength -d <<EOF
  expect(some_exception_variable) {|e| e.url.should == 'http://host/path'}
EOF
```

(there's a `\t` in the beginning)

Backtrace:

```
An error occurred while Layout/LineLength cop was inspecting bug.rb.
undefined method `offset' for nil
lib/rubocop/cop/mixin/line_length_help.rb:76:in `extend_uri_end_position'
lib/rubocop/cop/mixin/line_length_help.rb:44:in `find_excessive_uri_range'
lib/rubocop/cop/layout/line_length.rb:359:in `check_uri_line'
lib/rubocop/cop/layout/line_length.rb:258:in `check_line'
lib/rubocop/cop/layout/line_length.rb:103:in `block in on_investigation_end'
lib/rubocop/cop/layout/line_length.rb:102:in `each'
lib/rubocop/cop/layout/line_length.rb:102:in `each_with_index'
lib/rubocop/cop/layout/line_length.rb:102:in `on_investigation_end'
```

-----------------

* [X] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/line_length_help.rb
+++ b/lib/rubocop/cop/mixin/line_length_help.rb
@@ -37,11 +37,12 @@ module RuboCop
         last_uri_match = match_uris(line).last
         return nil unless last_uri_match
 
-        begin_position, end_position = last_uri_match.offset(0).map do |pos|
-          pos + indentation_difference(line)
-        end
-
+        begin_position, end_position = last_uri_match.offset(0)
         end_position = extend_uri_end_position(line, end_position)
+
+        line_indentation_difference = indentation_difference(line)
+        begin_position += line_indentation_difference
+        end_position += line_indentation_difference
 
         return nil if begin_position < max_line_length && end_position < max_line_length
 
--- a/changelog/fix_layout_line_length_cop_failure_with_yard_like_string.md
+++ b/changelog/fix_layout_line_length_cop_failure_with_yard_like_string.md
@@ -1,0 +1,1 @@
+* [#13534](https://github.com/rubocop/rubocop/pull/13534): Fix `Layout/LineLength` cop failure in case of YARD-comment-like string. ([@viralpraxis][])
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/mixin/line_length_help.rb
```diff
@@ -37,12 +37,13 @@ def find_excessive_uri_range(line)
...
 
-        begin_position, end_position = last_uri_match.offset(0).map do |pos|
-          pos + indentation_difference(line)
-        end
-
+        begin_position, end_position = last_uri_match.offset(0)
         end_position = extend_uri_end_position(line, end_position)
 
+        line_indentation_difference = indentation_difference(line)
+        begin_position += line_indentation_difference
```
koic, line 44: Is this line of processing necessary?

viralpraxis: It does make a difference in some cases.

For example, this line `"\t\t# See additional infor at http://x.com/rr"` with config

```
'Layout/LineLength' => { 'Max' => 30, 'AllowURI' => true }
'Layout/IndentationWidth' => { 'Width' => 2 },
```

would be marked as an offense (`Line is too long. [45/30]`) if we keep `begin_position += line_indentation_difference` (otherwise it wouldn't)

In general I believe this is a relatively rare edge case which does not make much difference in practice.

koic: It seems that no failing tests exist even if this line is removed. Could you add a new test code for this?

viralpraxis: Sure; I added a test based on my comment above.

koic:
Thanks!

PR #13536, opened 2024-12-03
Specify newer repository revision for precommit integration in docs

viralpraxis:
I've tried to add rubocop to my `.pre-commit-config.yaml` using snippet from [docs](https://docs.rubocop.org/rubocop/integration_with_other_tools.html#git-pre-commit-hook-integration-with-pre-commit), but it seems there's no `.pre-commit-hooks.yaml` at 1.8.1:

> An error has occurred: InvalidManifestError: 
> =====> ~/.cache/pre-commit/repopxc490ng/.pre-commit-hooks.yaml is not a file

It worked fine after changing `rev` to 1.69.0

-----------------

* [X] Added tests.

### DIFF
--- a/docs/modules/ROOT/pages/integration_with_other_tools.adoc
+++ b/docs/modules/ROOT/pages/integration_with_other_tools.adoc
@@ -125,7 +125,7 @@ following to your `.pre-commit-config.yaml` file:
 [source,yaml]
 ----
 - repo: https://github.com/rubocop/rubocop
-  rev: v1.8.1
+  rev: v1.9.0
   hooks:
     - id: rubocop
 ----
@@ -136,7 +136,7 @@ entries in `additional_dependencies`:
 [source,yaml]
 ----
 - repo: https://github.com/rubocop/rubocop
-  rev: v1.8.1
+  rev: v1.9.0
   hooks:
     - id: rubocop
       additional_dependencies:


docs/modules/ROOT/pages/integration_with_other_tools.adoc
```diff
@@ -125,7 +125,7 @@ following to your `.pre-commit-config.yaml` file:
 [source,yaml]
 ----
 - repo: https://github.com/rubocop/rubocop
-  rev: v1.8.1
+  rev: v1.69.0
```
koic, line 128: How about indicating the version in which it was introduced (https://github.com/rubocop/rubocop/commit/ff19e9ad0f42f2f5405393a4b1f73a2b7b830cc3) , such as 1.9?

```suggestion
  rev: v1.9.0
```

viralpraxis: Makes sense 👍🏻 

I tested this revision and it works properly.

Earlopain: How does this work, is it fetching this file from that point in time from the repository/what does revision mean? There has been at least one change in https://github.com/rubocop/rubocop/pull/10682 that should be picked up.

koic: Yeah, it seems that warning is suppressed starting from version 1.30.1, but it should still be usable in earlier versions. In any case, it's not a critical matter.

koic:
Thanks!

docs/modules/ROOT/pages/integration_with_other_tools.adoc
viralpraxis: > How does this work, is it fetching this file from that point in time from the repository/

Yes, it uses both git repo URL and `rev` revision to fetch repository.

I usually run `pre-commit autoupdate` time to time which automatically fetches the latest revision of repository and updates local `.pre-commit-config.yaml` to point to the latest `rev` (currently v1.69.1).

> There has been at least one change in https://github.com/rubocop/rubocop/pull/10682 that should be picked up.

Oh, you're right -- I was too fast regarding v1.9.0 :(

> rubocop..................................................................Failed
> - hook id: rubocop
> - exit code: 1
> 
> --auto-correct is deprecated; use --autocorrect instead.

viralpraxis: @koic looks like we have to update `rev` once again to catch the change @Earlopain mentioned

Earlopain: I would just have this be changed automatically in the release task here: https://github.com/rubocop/rubocop/blob/9f2ffa2270bbc174b2a622350d120b05672a1ce5/tasks/cut_release.rake#L39-L50

You bump it manually once in the PR for that to current latest, and it'll always kept up-to-date

koic: Yeah, if there are any issues with the work, please open a PR at a suitable time.

koic: I have no strong preference between specifying the minimum supported version or always documenting the latest version. Ideally, the minimum working version could be documented, and while updates to the latest version will be handled automatically by the release task. @viralpraxis As a user, please make changes that you believe will provide better user experience 😃

